PR #130, opened 2018-03-06
Centralise the canonical import paths and more package docs

markmandel:
Put the canonical import path into the main directory, to centralise it, and make sure it's applied across everything.

As a side effect of this work, also added some more package docs.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,18 @@
+/*
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


markmandel:
.. or maybe it actually makes more sense to drop this into 0.2 and play the safe game. WDYT?

enocom:
@markmandel Yeah. Definitely a nice to have. It's probably just a matter of finding the convention and following it.

markmandel:
SGTM. Moving to 0.2, and this PR can be a thread on how we want to have a convention on this.

enocom:
Take a look at [google-cloud-go](https://github.com/GoogleCloudPlatform/google-cloud-go). The top level `cloud.go` file specifies an import restriction. That same restriction appears nowhere else in subpackages.

If that's any guide, then the single restriction in the `version.go` file is enough.

markmandel:
From going from your above code - maybe there should be an `import.go` (or a `doc.go`) at the top level of this repository (rather than in pkg) - which would cover `pkg`, `cmd` and `sdks/go`.

That would seem to make more sense?

enocom:
Putting the restriction in a `doc.go` file is perfect. Great idea!

markmandel:
Done! PTAL.

PR #136, opened 2018-03-09
Cleanup of grpc go generation code

markmandel:
I'd previously generated this in two places, and looking at it through Godoc, it seemed like a code smell.

Having the generated code in a single place seems like the cleaner option.


markmandel:
@enocom Curious if you see any issues here?

enocom:
@markmandel Yeah, definitely don't need two generated files. The C++ generated stuff might even be deleted too. 

In theory, the proto file is enough, but since we're providing a Go library, we might as well have the generated code for people consuming things within `pkg`.

markmandel:
Personally, I prefer to keep generated files in the repo - makes clear what is going on (the CRD client is all generated too) - as long as it's header'd appropriately.

Kinda like vendoring dependencies :smile:

markmandel:
Also makes it go-gettable

enocom:
SGTM.

PR #143, opened 2018-03-15
Update create_gameserver.md

royingantaginting:
Use Github raw content url as source for gameserver.yml


royingantaginting:
I signed it!

PR #149, opened 2018-03-21
Helm integration

fooock:
Related to #101

PR #152, opened 2018-03-27
Documentation on the CI build system.

markmandel:
From #149 - realised we needed at least a note explaining how the build system works.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,5 +31,16 @@ Participation in this project comes under the [Contributor Covenant Code of Cond
 - Once review has occurred, please rebase your PR down to a single commit. This will ensure a nice clean Git history.
 - Finally - *Thanks* for considering submitting code to Agones!
 
+## Continuous Integration
+
+Continuous integration is provided by [Google Cloud Container Builder](https://cloud.google.com/container-builder/),
+through the [cloudbuilder.yaml](./cloudbuild.yaml) file found at the root of the directory.
+
+Build success/failure with relevant details are pushed automatically to pull requests via the not (yet 😉) opensourced
+build system.
+
+See the [Container Builder documentation](https://cloud.google.com/container-builder/docs/) for more details on
+how to edit and expand the build process.
+
 ### Additional Resources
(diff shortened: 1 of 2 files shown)


pondohva:
There is nice example we can adopt for helm chart https://github.com/kubernetes/charts/tree/master/test/circle

For helm we need to have a:

1. CI system where we can run lint and package commands
2. storage (I prefer google cloud storage for this, also we can use github pages or smth else)

Also, we can contribute to kubernetes/charts repo agones helm chart from #149. This repo is a common place to look for helm charts, they provide CI system and 2 repos: https://kubernetes-charts.storage.googleapis.com and https://kubernetes-charts-incubator.storage.googleapis.com

CONTRIBUTING.md
```diff
@@ -31,5 +31,16 @@ Participation in this project comes under the [Contributor Covenant Code of Cond
 - Once review has occurred, please rebase your PR down to a single commit. This will ensure a nice clean Git history.
 - Finally - *Thanks* for considering submitting code to Agones!
 
+## Continuous Integration
```
pondohva: nice!

pondohva:
We need to add go lint/go vet and tests for golang code also.

markmandel:
@pondohva we run the go tests on every build, see [here](https://github.com/GoogleCloudPlatform/agones/blob/master/cloudbuild.yaml#L30) in the cloudbuild.yaml.

While [gometalinter](https://github.com/alecthomas/gometalinter) is part of the build image (and therefore shell), I'm reticent to make it part of the test/build pipeline, just because of the propensity for it to raise false positives, and also because it doesn't ignore generated files - of which there are quite a few in this project between gRPC and generated CRD apis. (Also see [this issue](https://github.com/alecthomas/gometalinter/issues/270)).

That being said, if you feel that there are places in which we can do a more on the testing front - or feel you have a way to work around the above issues, definitely  submit a PR to the cloudbuild.yaml and supporting test suite.

Kuqd:
I agree with @pondohva we should have a CI for the helm package when #149 is closed.

markmandel:
Any reason I can't get an approval on this PR though, so I can get it merged? :smile:

markmandel:
Thanks!

Issue #101, opened 2018-02-22
YAML packaging


Kuqd:
Looking at the big install.yaml file and all the configurations possible, I think it would help people to provide a way of packaging YAML files.

Possible variables:
- namespace
- service account names
- MIN_PORT
- MAX_PORT
- images version (for later)
- RBAC on/off ???

The default variable should reflect the current install.yaml.

We should also update the documentation to add installation step using helm on top of what we already have.

I don't think it's urgent for 0.1 but interesting to have.

/cc @rodcloutier

Kuqd:
https://github.com/SUSE/helm-certgen

markmandel:
Chat from today.
TL;DR - We will go with Helm.

```
Mark Mandel [1:12 PM]
@Cyril Tovena chatting about Helm - quick thought - if we wanted to do a quick spike, to see what a helm chart for Agones looked like - easy/hard ?


Rod Cloutier [1:12 PM]
Would be fairly easy for us to do.

Mark Mandel [1:12 PM]
Sounds like that may be a nice way forward? Just to try it out, see how it feels?
especially if it will handle the certs issues
and installing helm doesn't look particularly onerous

Rod Cloutier [1:13 PM]
We don`t need to use Helm on both the client and server side. You can generate a yaml from Helm charts and do a `kubectl apply` afterward.

Mark Mandel [1:13 PM]
OH REALLY?
this is why I ask you smart people the questions :smile:
I did not know this
Sounds like Helm is a winner winner chicken dinner?

Cyril Tovena [1:15 PM]
like I said you could try to even get a stable chart
which is good for visibility
I suggest that I give a go with helm without certs and move from there

Mark Mandel [1:17 PM]
SGTM
I'll copy paste this little chat into the ticket, just for posterity/transparency
```

Kuqd:
while reading the doc I found that you can run post installation test : https://docs.helm.sh/developing_charts/#chart-tests

That's just awesome, I'm thinking to use flag switch in the controller container to run a simple gs/pod creation and release control.

WDYT ? I'm planing to introduce this after a first version so probably create a new issue if we all think that's a boomer.

markmandel:
We do have a need for #37 (End to end tests) - so maybe this helps with both tickets!

rodcloutier:
Just be careful with this as you now need to install tiller in the cluster to make your tests. This is nice to use a sanity check for the user but I would probably not use this as end to end tests.

markmandel:
[...]

PR #154, opened 2018-03-28
Allow GameServers to be created in non-default namespace.

### LINKED ISSUE #146: Cannot run GameServer in non-default namespace
I want to start by explaining that this came by my fault by creating the gameserver in the agones-system namespace, but anyway, it should be able to run in any namespace.

It makes the gameserver not being able to get an ip and port where to connect to.
([Step 2 on simple-udp example](https://github.com/GoogleCloudPlatform/agones/blob/master/docs/create_gameserver.md#2-fetch-the-gameserver-status))

Command: `kubectl logs simple-udp-8l26l -c agones-gameserver-sidecar` 
Logs: https://gist.github.com/jmaeso/aa4562bd3ad4e5c91bd1e558e329f8a7

Workaround for this: 
```
After you install Agones - make sure you are back on the default namespace
```

dzlier-gcp:
Closes #146

### DIFF
--- a/build/install.yaml
+++ b/build/install.yaml
@@ -225,6 +225,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: agones-sdk
+  namespace: agones-system
 secrets:
 - name: agones-sdk-secret
 ---
@@ -240,7 +241,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: agones-sdk
-  namespace: agones-system
 rules:
 - apiGroups: ["stable.agones.dev"]
   resources: ["gameservers"]
@@ -250,10 +250,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: agones-sdk-access
-  namespace: agones-system
 subjects:
 - kind: User
-  name: system:serviceaccount:default:agones-sdk
+  name: system:serviceaccount:agones-system:agones-sdk
   apiGroup: rbac.authorization.k8s.io
 roleRef:
   apiGroup: rbac.authorization.k8s.io
--- a/install.yaml
+++ b/install.yaml
@@ -224,6 +224,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: agones-sdk
+  namespace: agones-system
 secrets:
 - name: agones-sdk-secret
 ---
@@ -239,7 +240,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: agones-sdk
-  namespace: agones-system
 rules:
 - apiGroups: ["stable.agones.dev"]
   resources: ["gameservers"]
@@ -249,10 +249,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: agones-sdk-access
-  namespace: agones-system
 subjects:
 - kind: User
-  name: system:serviceaccount:default:agones-sdk
+  name: system:serviceaccount:agones-system:agones-sdk
   apiGroup: rbac.authorization.k8s.io
 roleRef:
   apiGroup: rbac.authorization.k8s.io


markmandel:
Just took this for a spin - built a new image, and installed from `build` tools, but it didn't work for me :(

```json
{
    "error": "error creating Pod for GameServer cpp-simple-zx7m2: pods \"cpp-simple-zx7m2-\" is forbidden: error looking up service account default/agones-sdk: serviceaccount \"agones-sdk\" not found",
    "level": "error",
    "msg": "",
    "obj": "default/cpp-simple-zx7m2",
    "queue": "stable.agones.dev.GameServerController",
    "source": "*gameservers.Controller",
    "time": "2018-03-29T00:32:45Z"
}
```

```json
{
    "level": "error",
    "msg": "error creating Pod for GameServer cpp-simple-zx7m2: pods \"cpp-simple-zx7m2-\" is forbidden: error looking up service account default/agones-sdk: serviceaccount \"agones-sdk\" not found",
    "stack": [
        "agones.dev/agones/pkg/gameservers.(*Controller).syncGameServerCreatingState\n\t/go/src/agones.dev/agones/pkg/gameservers/controller.go:387",
        "agones.dev/agones/pkg/gameservers.(*Controller).syncGameServer\n\t/go/src/agones.dev/agones/pkg/gameservers/controller.go:273",
        "agones.dev/agones/pkg/gameservers.(*Controller).(agones.dev/agones/pkg/gameservers.syncGameServer)-fm\n\t/go/src/agones.dev/agones/pkg/gameservers/controller.go:111",
        "agones.dev/agones/pkg/util/workerqueue.(*WorkerQueue).processNextWorkItem\n\t/go/src/agones.dev/agones/pkg/util/workerqueue/workerqueue.go:97",
        "agones.dev/agones/pkg/util/workerqueue.(*WorkerQueue).runWorker\n\t/go/src/agones.dev/agones/pkg/util/workerqueue/workerqueue.go:73",
        "agones.dev/agones/pkg/util/workerqueue.(*WorkerQueue).(agones.dev/agones/pkg/util/workerqueue.runWorker)-fm\n\t/go/src/agones.dev/agones/pkg/util/workerqueue/workerqueue.go:115",
        "agones.dev/agones/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/go/src/agones.dev/agones/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133",
        "agones.dev/agones/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/src/agones.dev/agones/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134",
        "agones.dev/agones/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\t/go/src/agones.dev/agones/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88",
        "runtime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:2361"
    ],
    "time": "2018-03-29T00:32:45Z"
}
```

enocom:
It sounds like there are some possible improvements to how Agones handles namespaces.

For now, I think moving Agones into the non-default namespace is a nice change.

@markmandel Any objections to just merging this?

dzlier-gcp:
The result of merging this would be that all GameServer + Pods would have to be created in the agones-system namespace, so if people are already accustomed to creating them in default, it will break some things.

enocom:
Thinking about upgrade paths is a good point. Since we're in alpha, I think we can still make breaking changes, but we should provide a note in the release notes about it.

markmandel:
So it's not a _hard_ requirement that the Pod be created in the same namespace as the GameServer - but I think logically it makes sense, and it would be confusing to the end user to find them in different places.  If I make a Deployment, my Pods stay in the same Namespace as well.

Forcing people to create GameServers in a specific namespace (other than default) also feels very counter intuitive for any Kubernetes user.

On upgrades - this is alpha -- we can break things, as long as we document them. It's in the root README :smile: 

I feel like there has to be a way to do this - the kubectl user has access to all namespaces, so there has to be a way. May be time to reach out and see who we know who works on RBAC, and get them to review?

markmandel:
...and apparently I'm totally wrong. Chatting with a few people:

> service accounts are namespaced. pods are namespaced. podspecs can only references pods, secrets, service accounts that are local to the namespace.

So this sounds like something we should solve at the #101 (Yaml/Helm packaging) - wherein the user selects which namespaces `GameServers` can be used at install/update time, and all the service accounts that are needed are created at that point. WDYT?

markmandel:
Just to clean up - should we close this PR, and make a note to revisit this in #146 when #101 has the first part of it's work complete (a Helm install that works)

enocom:
Yep. I think that makes sense. @dzlier-gcp What do you think?

enocom:
I'm going to close this in favor of #101.

Issue #70, opened 2018-02-04
GameServer Fleets


markmandel:
# Design

## Description
Fleets are a group of warm servers that are available to be allocated to players when needed.

In Kubernetes parlance, they are the `Deployment`/`ReplicaSet` to Pods, but for `GameServer`s

## Features

- Be able to define a `Fleet`, with an attached `GameServerTemplate` (much like a `PodTemplate`)
- The Fleet ensures there is always `replicas` number of Healthy `GameServers` available (assuming resources exist)
- If a `GameServer` becomes Unhealthy, then delete it and create it anew (we may add more options at a later date).
- If you delete a `Fleet`, this also deletes the backing `GameServers` (this should be by default in Kubernetes now anyway)
- A mechanism (TBD) to be able to get an `allocated` GameServer out of the pool.
  - If a warm server is not available, then a cold one should be started (assuming resources)
  - An allocated `GameServer` is moved to a `Allocated` state on allocation.
- If the `replicas` are increased in the `Fleet`, the number of `GameServers` is increased to match that number (assuming resources)
 - If the `replicas` are decreased in the `Fleet`, the number of `GameServers` is decreased to match that number.
    - `GameServers` that are in an `Allocated` state will never be deleted during the decrease
- The `GameServer` `template` is changed, then we will mimic a `Deployment` in that we can do either a `Recreate` or a `RollingUpdate` to switch out the waiting warm servers.
  - GameServers that are in an Allocated state will never be deleted during the decrease
- Validation of the configuration and changes.

# Configuration
```yaml
apiVersion: "stable.agon.io/v1alpha1"
kind: Fleet
metadata:
  name: "fleet-example"
spec:
  # number of GameServers
  replicas: 10
  # deployment strategy for updating the image
  # Lifted directly from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#deploymentstrategy-v1-apps
  strategy:
    # Recreate or RollingUpdate. Default to RollingUpdate
    type: RollingUpdate
    rollingUpdateDeployment: # optional rolling update config
      maxSurge: "25%"
      maxUnavailable: "25%"
  # A GameServer template
  template:
    # Standard ObjectMeta
    metadata:
      labels:
        mylabel: myvalue
    # GameServer spec
    spec:
      containerPort: 7654
        template:
          spec:
            containers:
            - name: cpp-simple
              image: gcr.io/agon-images/cpp-simple-server:0.1
```
## Allocation

[...]

PR #156, opened 2018-04-03
GameServerSet Implementation

markmandel:
GameServerSets are the basic building block for Fleets. They allow for creating a set of GameServers that can be scaled up and down in size, while also being aware of `Allocated` GameServers and ensuring they don't get deleted. 

GameServerSets will be allow Fleet migrations to occur, similarly to how ReplicaSets allow Deployments to migrate one image type to another.

This has not been formally documented, as this will likely be an internal CRD, and not (widely) used externally.

Parent ticket: #70


markmandel:
/cc @ianlewis if you also want to take a look.

If anyone has a better name than `GameServerSet`, I'm also all ears. I can't reuse `ReplicaSet` because it would conflict with `kubectl` so :man_shrugging:

cmd/controller/main.go
```diff
@@ -158,10 +172,7 @@ func main() {
...
 	}()
 
-	err = c.Run(2, stop)
-	if err != nil {
-		logger.WithError(err).Fatal("Could not run gameserver controller")
-	}
-
-	logger.Info("Shut down gameserver controller")
+	<-stop
+	logger.Info("Shut down agones controller")
```
EricFortin: Nitpick: `controllers`

pkg/gameserversets/controller.go
```diff
@@ -0,0 +1,345 @@
...
+
+		if gs.Status.State != stablev1alpha1.Allocated {
+			err := c.gameServerGetter.GameServers(gs.Namespace).Delete(gs.ObjectMeta.Name, nil)
+			if err != nil {
+				return errors.Wrapf(err, "error deleting gameserver for gameserverset %s", gss.ObjectMeta.Name)
+			}
+			c.recorder.Eventf(gss, corev1.EventTypeNormal, "SuccessfulDelete", "Deleted GameServer: %s", gs.ObjectMeta.Name)
+			count++
+		}
+	}
```
EricFortin: What happens if they are all allocated? Will this re-queue the sync event to be processed later on?

cmd/controller/main.go
markmandel: Done.

pkg/gameserversets/controller.go
markmandel: If they are allocated, [the event handler above](https://github.com/GoogleCloudPlatform/agones/pull/156/commits/ef92839c3ac99bdf224ec0d16adf367d40098bc1#diff-3d3de0ef142666d0fe31e84b75d29d9cR119) will watch for when the `GameServer` is deleted for this `GameServerSet`, and when it catches that event, it will queue the owning `GameServerSet` to be processed again, and make sure it's at the right number.

cmd/controller/main.go
```diff
@@ -128,7 +129,8 @@ func main() {
 	agonesInformerFactory := externalversions.NewSharedInformerFactory(agonesClient, 30*time.Second)
 	kubeInformationFactory := informers.NewSharedInformerFactory(kubeClient, 30*time.Second)
 
-	c := gameservers.NewController(wh, health, minPort, maxPort, sidecarImage, alwaysPullSidecar, kubeClient, kubeInformationFactory, extClient, agonesClient, agonesInformerFactory)
+	gsController := gameservers.NewController(wh, health, minPort, maxPort, sidecarImage, alwaysPullSidecar, kubeClient, kubeInformationFactory, extClient, agonesClient, agonesInformerFactory)
+	gssController := gameserversets.NewController(wh, kubeClient, extClient, agonesClient, agonesInformerFactory)
```
enocom: @markmandel The one letter difference between variable names is easy to miss. How about `gs` and `gsSet` or something that is easier to parse quickly?

enocom:
Small nitpick on variable names in `main.go`. Otherwise, LGTM.

cmd/controller/main.go
markmandel: I like gsSet better. I'm actually going to change that across the whole codebase. I've had issues writing this mixing up `gs` and `gss` and it took me ages to fix the bugs!

markmandel:
Blocking #149 (Helm) before merging this - so as to not keep moving the goalposts on what the install.yaml should be.

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -0,0 +1,109 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# validation for a gameserver spec
+
+{{- define "gameserver.validation" }}
```
markmandel: @Kuqd @fooock - see any issues with me shifting this validation into a template like this?

Seems like the correct way to do it, but wanted to double check in case you could see anything obviously bad?

fooock: For me it's ok. Small things:

* I would add a small documentation above the `define`, line `{{ /*  Validation for a gameserver spec .... */ }}`
* Why `{{ end - }}` instead of `{{ - end }}` ??
* Normally these templates are inside the `_helpers.tpl` file, but by organization, I would leave it like that.

markmandel: Good call on the first two. Done!.

Re: 
> Normally these templates are inside the _helpers.tpl file, but by organization, I would leave it like that.

So I looked at https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/templates.md

> Template files should have the extension .yaml if they produce YAML output. The extension .tpl may be used for template files that produce no formatted content.

So I made it a .yaml file with an underscore, since it's a pretty big yaml block. WDYT?

PR #158, opened 2018-04-04
Add GKE Comic to explain Kubernetes to newcomers

markmandel:
Wanted an avenue for those who don't know Kubernetes and want somewhere to start. I love this comic, I think it makes Kubernetes super approachable.


markmandel:
If anyone has opinions on where in the README this should be - would definitely like to hear them. I wasn't sure if this is the right spot.

PR #160, opened 2018-04-05
Register liveness check in gameservers.Controller

### LINKED ISSUE #116: Add liveness check to `pkg/gameservers/controller`.
[Original Context](https://github.com/googleprivate/agones/pull/98#issuecomment-368082197).

> I don't know if this is possible, but what would determine if this controller is healthy is that the goroutines generated [in this code](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameservers/controller.go#L224) are still running and haven't exited in any way.

enocom:
The liveness check is based on the worker queue having all its worker
goroutines running. If one of those goroutines exits, the liveness check
reports an unhealthy status.

Fixes #116

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -109,6 +109,7 @@ func NewController(
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
 	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
+	health.AddLivenessCheck("game-server-worker-queue", healthcheck.Check(c.workerqueue.Healthy))
 
 	wh.AddHandler("/mutate", stablev1alpha1.Kind("GameServer"), admv1beta1.Create, c.creationMutationHandler)
 	wh.AddHandler("/validate", stablev1alpha1.Kind("GameServer"), admv1beta1.Create, c.creationValidationHandler)
--- a/pkg/util/workerqueue/workerqueue.go
+++ b/pkg/util/workerqueue/workerqueue.go
@@ -18,6 +18,8 @@
 package workerqueue
 
 import (
+	"fmt"
+	"sync"
 	"time"
 
 	"agones.dev/agones/pkg/util/runtime"
@@ -26,6 +28,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+)
+
+const (
+	workFx = time.Second
 )
 
 // Handler is the handler for processing the work queue
@@ -40,6 +46,10 @@ type WorkerQueue struct {
 	queue  workqueue.RateLimitingInterface
 	// SyncHandler is exported to make testing easier (hack)
 	SyncHandler Handler
+
+	mu      sync.Mutex
+	workers int
+	running int
 }
 
 // NewWorkerQueue returns a new worker queue for a given name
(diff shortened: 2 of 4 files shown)


pkg/gameservers/controller.go
```diff
@@ -82,7 +82,8 @@ func NewController(
 	kubeInformerFactory informers.SharedInformerFactory,
 	extClient extclientset.Interface,
 	agonesClient versioned.Interface,
-	agonesInformerFactory externalversions.SharedInformerFactory) *Controller {
+	agonesInformerFactory externalversions.SharedInformerFactory,
+) *Controller {
```
markmandel: For some reason, I find this being on the next line kinda weird. I'm assuming you did this on purpose :wink:

pkg/gameservers/controller.go
```diff
@@ -109,6 +110,12 @@ func NewController(
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
 	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
+	health.AddLivenessCheck("game-server-worker-queue", healthcheck.Check(func() error {
+		if !c.workerqueue.Healthy() {
```
markmandel: Thought - would it make sense to have c.workerqueue.Healthy() return an error when it's not healthy - rather than a bool?

Then we could just do: `health.AddLivenessCheck("game-server-worker-queue", c.workerqueue.Healthy)`

Not wedded to it, but thought it might be worth discussion?

pkg/util/workerqueue/workerqueue.go
```diff
@@ -40,6 +45,10 @@ type WorkerQueue struct {
 	queue  workqueue.RateLimitingInterface
 	// SyncHandler is exported to make testing easier (hack)
 	SyncHandler Handler
+
+	mu      sync.Mutex
```
markmandel: Should this be a RWLock? I suppose there will only be a single thread hitting the `Healthy()` function, so there isn't much benefit to allow for multiple concurrent `Healthy()` requests. WDYT?

pkg/util/workerqueue/workerqueue.go
```diff
@@ -106,15 +115,56 @@ func (wq *WorkerQueue) processNextWorkItem() bool {
...
+		go wq.run(stop)
 	}
 
 	<-stop
 	wq.logger.Info("...shutting down workers")
+	wq.queue.ShutDown()
+}
+
+func (wq *WorkerQueue) run(stop <-chan struct{}) {
+	wq.inc()
```
markmandel:
```go
wq.inc()
defer wq.dec()
wait.Until(wq.runWorker, workFx, stop)
``` 
?

pkg/util/workerqueue/workerqueue.go
```diff
@@ -106,15 +115,56 @@ func (wq *WorkerQueue) processNextWorkItem() bool {
...
+	wait.Until(wq.runWorker, workFx, stop)
+	wq.dec()
+
+}
+
+// Healthy reports whether all the worker goroutines are running.
+func (wq *WorkerQueue) Healthy() bool {
```
markmandel:
```go
wq.mu.Lock()
defer wq.mu.Unlock()
return wq.workers == wq.running
```
?
In this case, this shows why I prefer a lock/defer unlock pattern, as once you past it, it shows much easier exactly what the logic of the function is. Otherwise you have to mentally unwind what `want` and `got` actually mean. In this case it's pretty easy, but be good to set a standard. WDYT?

pkg/util/workerqueue/workerqueue.go
```diff
@@ -106,15 +115,56 @@ func (wq *WorkerQueue) processNextWorkItem() bool {
...
+}
+
+func (wq *WorkerQueue) RunCount() int {
+	wq.mu.Lock()
+	defer wq.mu.Unlock()
+	return wq.running
+}
+
+func (wq *WorkerQueue) setWorkerCount(n int) {
+	wq.mu.Lock()
```
markmandel:
```go
wq.mu.Lock()
defer wq.mu.Unlock()
wq.workers = n
```
? 
Really just being consistent in lock/unlock strategy through the code.

pkg/util/workerqueue/workerqueue.go
```diff
@@ -106,15 +115,56 @@ func (wq *WorkerQueue) processNextWorkItem() bool {
...
+	wq.workers = n
+	wq.mu.Unlock()
+}
+
+func (wq *WorkerQueue) inc() {
```
markmandel:
```go
wq.mu.Lock()
defer wq.mu.Unlock()
wq.running++
```
?

pkg/util/workerqueue/workerqueue.go
```diff
@@ -106,15 +115,56 @@ func (wq *WorkerQueue) processNextWorkItem() bool {
...
+	wq.mu.Lock()
+	wq.running++
+	wq.mu.Unlock()
+}
+
+func (wq *WorkerQueue) dec() {
```
markmandel:
```go
wq.mu.Lock()
defer wq.mu.Unlock()
wq.running--
```
?

pkg/gameservers/controller.go
enocom: That's a better idea. I'll make the change.

pkg/util/workerqueue/workerqueue.go
enocom: I'm happy to use the `defer` pattern.

pkg/util/workerqueue/workerqueue.go
enocom: I'm tempted to start with a plain mutex and leave open the possibility for improving performance of the `WorkerQueue` once we understand the bottle necks.

markmandel: SGTM!

PR #162, opened 2018-04-05
Fix some Go Report Card warnings

dvrkps:
Many of golint warnings are in autogenerated files.

### DIFF
--- a/examples/simple-udp/client/main.go
+++ b/examples/simple-udp/client/main.go
@@ -44,14 +44,20 @@ func main() {
 	if err != nil {
 		log.Fatalf("Could not connect: %v", err)
 	}
-	defer conn.Close()
+
+	defer func() {
+		err = conn.Close()
+		if err != nil {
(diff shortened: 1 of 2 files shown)


examples/simple-udp/client/main.go
```diff
@@ -44,14 +44,20 @@ func main() {
...
+		}
+	}()
 
 	go func() {
 		b := make([]byte, bufferSize)
 		for {
-			n, err := conn.Read(b)
-			if err != nil {
-				log.Fatalf("Could not read packet: %v", err)
+			n, errRead := conn.Read(b)
```
enocom: Catching these kinds of things are why the linters are so helpful.

markmandel: @enocom Does that mean you'll integrate `gometalinter` into the cloudbuild.yaml system? :smile: 

It's already in the build package.

enocom: Added #163 to track the work.

PR #164, opened 2018-04-06
Refactor of workerqueue health testing

markmandel:
- Moved the health check testing out of the gameserver controller, into the workerqueue test as the health check will be in place in multiple controllers
- Added the use of the `httptest` package, so we can run this test in parralel
- Updated more tests to be parralel.

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -109,7 +109,7 @@ func NewController(
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
 	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
-	health.AddLivenessCheck("game-server-worker-queue", healthcheck.Check(c.workerqueue.Healthy))
+	health.AddLivenessCheck("gameserver-workerqueue", healthcheck.Check(c.workerqueue.Healthy))
 
 	wh.AddHandler("/mutate", stablev1alpha1.Kind("GameServer"), admv1beta1.Create, c.creationMutationHandler)
 	wh.AddHandler("/validate", stablev1alpha1.Kind("GameServer"), admv1beta1.Create, c.creationValidationHandler)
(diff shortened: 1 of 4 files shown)


pkg/util/workerqueue/workerqueue_test.go
```diff
@@ -81,3 +91,50 @@ func TestWorkerQueueHealthy(t *testing.T) {
...
+		return nil
+	}
+	wq := NewWorkerQueue(handler, logrus.WithField("source", "test"), "test")
+	health.AddLivenessCheck("test", wq.Healthy)
+
+	server := httptest.NewServer(health)
+	defer server.Close()
+
+	stop := make(chan struct{})
+	go func() {
```
enocom: The call to `wq.Run` can be inlined to remove `func`.

pkg/util/workerqueue/workerqueue_test.go
```diff
@@ -81,3 +91,50 @@ func TestWorkerQueueHealthy(t *testing.T) {
...
+	t.Parallel()
+
+	health := healthcheck.NewHandler()
+	handler := func(string) error {
+		return nil
+	}
+	wq := NewWorkerQueue(handler, logrus.WithField("source", "test"), "test")
+	health.AddLivenessCheck("test", wq.Healthy)
+
+	server := httptest.NewServer(health)
```
enocom: This is a great choice, as all the port handling between different tests isn't our responsibility.

pkg/gameservers/controller.go
```diff
@@ -109,7 +109,7 @@ func NewController(
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
 	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
-	health.AddLivenessCheck("game-server-worker-queue", healthcheck.Check(c.workerqueue.Healthy))
+	health.AddLivenessCheck("gameserver-workerqueue", healthcheck.Check(c.workerqueue.Healthy))
```
enocom: Do you know of a convention for naming these liveness checks from Kubernetes?

markmandel: No idea. I just figured keep it closer to the names of the CRDs and objects that back them.

pkg/util/workerqueue/workerqueue_test.go
markmandel: Done!

PR #172, opened 2018-04-10
Centralise utilities for testing controllers

markmandel:
Found myself repeating the same test setup when I started doing the Fleet controllers, so moved this
into a single location (`pkg/testing`) to make it easier to reuse.


markmandel:
So this is an interesting question - I followed what Kubernetes does (and is that the best thing?), and then use package aliases to manage things.

https://godoc.org/k8s.io/client-go/testing
https://godoc.org/k8s.io/client-go/tools/cache/testing
https://godoc.org/k8s.io/client-go/util/testing
https://godoc.org/k8s.io/kubernetes/pkg/api/testing
(And so on)

I'm not aware of any other patterns, but I'm open!
WDYT?

PR #174, opened 2018-04-12
Creating a Fleet creates a GameServerSet

markmandel:
All this commit is doing is create a GameServerSet when a Fleet is created. Also, if the Replicas in
a Fleet are updated, then the owned GameServerSet's Replicas are also updated.

For this commit, we are not worrying about handling updates to the Spec (e.g. new image) to the Fleet,
and doing updates to a live Fleet, we are just concerned with scaling up and down.

Parent ticket: #70


fooock:
@markmandel can be the error in this [line](https://github.com/GoogleCloudPlatform/agones/blob/master/build/Makefile#L110)? `-mkdir` instead `mkdir`

markmandel:
@fooock - the `-` just says to `Make`, if this command fails, just ignore it.
Since the directory could already exist from previous runs, we just ignore the `mkdir` command

The test failed because I didn't run `gen-install` to push the changes I made to the Helm chart to the manual install. `diff` actually returns a non-zero exit code when it finds a difference :smile:

fooock:
ahhh ok, thanks for the explaination @markmandel . I didn't know it 😃

pkg/fleets/controller.go
```diff
@@ -0,0 +1,249 @@
...
+		activeGsSet, err = c.gameServerSetGetter.GameServerSets(fleet.ObjectMeta.Namespace).Create(activeGsSet)
+		if err != nil {
+			return errors.Wrapf(err, "error creating gameserverset for fleet %s", fleet.ObjectMeta.Name)
+		}
+
+		c.recorder.Eventf(fleet, corev1.EventTypeNormal, "CreatingGameServerSet",
+			"Created GameServerSet %s", activeGsSet.ObjectMeta.Name)
+
+	} else {
+		// for now, we're ignoring any change to the template - will handle on the next PR
```
EricFortin: You want to support multiple game server set to implement things like rolling update?

pkg/fleets/controller.go
```diff
@@ -0,0 +1,249 @@
...
+	// if the replica count has changed, then update the GameServerSet
+	if fleet.Spec.Replicas != activeGsSet.Spec.Replicas {
+		gsSetCopy := activeGsSet.DeepCopy()
+		gsSetCopy.Spec.Replicas = fleet.Spec.Replicas
+
+		if gsSetCopy, err = c.gameServerSetGetter.GameServerSets(fleet.ObjectMeta.Namespace).Update(gsSetCopy); err != nil {
+			return errors.Wrapf(err, "error updating replicas for gameserverset for fleet %s", fleet.ObjectMeta.Name)
+		}
+		c.recorder.Eventf(fleet, corev1.EventTypeNormal, "ScalingGameServerSet",
+			"Scaling GameServerSet %s to %d", gsSetCopy.ObjectMeta.Name, gsSetCopy.Spec.Replicas)
```
EricFortin: Could we add the old number to in the event so we know if it was scaling up or down?

pkg/fleets/controller.go
markmandel: Yep, exactly!

pkg/fleets/controller.go
markmandel: Nice! Done.

PR #184, opened 2018-04-23
Adding additional resources for CRDs and Controllers

markmandel:
Had several people asking about how to learn about Custom Resource Definitions and Controllers with Kubernetes.

Adding the best resources I've found on the above.


enocom:
@markmandel This just needs to be rebase on top of master and can be merged.

PR #193, opened 2018-05-01
Creating a FleetAllocation allocated a GameServer from a Fleet

markmandel:
This implements the FleetAllocation CRD, that when created looks at the requested Fleet, and marks one of the GameServers as Allocated (if available) and returns it's value in the FleetAllocation.

This also includes a quickstart and spec for reference as well.

Some refactoring was included in the PR (probably should have been a seperate PR for that).


pkg/fleets/controller.go
```diff
@@ -169,7 +168,7 @@ func (c *Controller) syncFleet(key string) error {
 		return errors.Wrapf(err, "error retrieving fleet %s from namespace %s", name, namespace)
 	}
 
-	list, err := c.listGameServerSets(fleet)
+	list, err := ListGameServerSetsByFleetOwner(c.gameServerSetLister, fleet)
```
markmandel: @enocom would love to get your opinion on this refactor ([implementation here](https://github.com/GoogleCloudPlatform/agones/pull/193/files#diff-91029c2ecd86ab36712b39724f333a94R28))

There were several places I needed a list type function like this, and wasn't quite sure the best way to do it.

pkg/apis/stable/v1alpha1/fleetallocation.go
```diff
@@ -0,0 +1,68 @@
...
+
+// FleetAllocationStatus will contain the
+// `GameServer` that has been allocated from
+// a Fleet
+type FleetAllocationStatus struct {
+	GameServer *GameServer
+}
+
+// Validateupdate validates when an update occurs
+func (fa *FleetAllocation) Validateupdate(new *FleetAllocation) (bool, []metav1.StatusCause) {
```
EricFortin: Nitpick: ValidateUpdate

EricFortin:
LGTM!

pkg/fleets/fleets.go
```diff
@@ -0,0 +1,66 @@
...
+	"agones.dev/agones/pkg/gameserversets"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// ListGameServerSetsByFleetOwner lists all the GameServerSets for a given
+// Fleet
+func ListGameServerSetsByFleetOwner(gameServerSetLister listerv1alpha1.GameServerSetLister, f *stablev1alpha1.Fleet) ([]*stablev1alpha1.GameServerSet, error) {
+	list, err := gameServerSetLister.List(labels.SelectorFromSet(labels.Set{stablev1alpha1.FleetGameServerSetLabel: f.ObjectMeta.Name}))
```
enocom: Based on a cursory reading of the code, I think you could even replace the concrete type `*stablev1alpha1.Fleet` with a `metav1.Object` interface. This way, the `ListGameServerSetsByFleetOwner` function might enjoy an even wider usage.

enocom: If you use the `metav1.Object` interface here, then the name lookup becomes `f.GetName()` as with the lookup below.

pkg/fleets/controller.go
enocom: See my response below.

pkg/fleets/fleets.go
enocom: I've seen that pattern in some object-oriented code, but I'm having a hard time remembering what the pattern might be called. It seems entirely reasonable to me.

markmandel: SGTM! Thanks for the sanity check.

enocom: I think it's basically a form of [double dispatch](https://en.wikipedia.org/wiki/Double_dispatch).

PR #197, opened 2018-05-03
Fixes for GKE issues with install/quickstart

markmandel:
Resolving some issues that new users found when trying to run through the quickstart with GKE and Cloud Shell.

Apparently Cloud Shell blocks UDP traffic, so dropped in a workaround.

### DIFF
--- a/docs/create_gameserver.md
+++ b/docs/create_gameserver.md
@@ -4,8 +4,8 @@ This guide covers how you can quickly get started using Agones to create GameSer
 
 ## Objectives
 
-- Create a GameServer in Kubernetes using Agones custom ressource.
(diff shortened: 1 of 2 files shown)


markmandel:
@dzlier-gcp did you run into the Cloud Shell issue when you wrote this? Wondering if it's new (more out of curiosity than anything else)

markmandel:
/cc @Kuqd I was totally wrong here, and thought @dzlier-gcp wrote the quickstart.

@Kuqd - do you see any issues with the changes I made? Had to include a workaround for cloud shell, because it looks like `nc -u` is blocked on Cloud Shell. Can always amend as a separate PR.

PR #199, opened 2018-05-08
Fleet update strategy: Replace

markmandel:
When the template for a Fleet is edited, a deployment strategy can be defined. At the moment, this is just the `Replace` strategy, which terminates all current non-allocated GameServers and immediately starts up new GameServers to replace them.

Includes documentation as well.


docs/create_fleet.md
```diff
@@ -302,6 +299,44 @@ If you run `kubectl describe gs | grep State` again - either the GameServer will
 Since we are running a `Fleet`, Agones will always do it's best to ensure there are always the configured number
 of `GameServers` in the pool in either a `Ready` or `Allocated` state. 
 
+### 7. Deploy a new version of the GameServer on the Fleet
+
+We can also change the configuration of the `GameServer` of the running `Fleet`, and have the changes
+roll out, without interrupting the currently `Allocated` `GameServers`.
+
+Let's take this for a spin! Run `kubectl edit fleet simple-udp` and set the `repicas` field to back to `5`.
```
EricFortin: Typo: `replicas` field back to `5`

pkg/fleets/controller.go
```diff
@@ -173,39 +224,102 @@ func (c *Controller) syncFleet(key string) error {
 		return err
 	}
 
-	var activeGsSet *stablev1alpha1.GameServerSet
+	activeGsSet, rest := c.filterGameServerSetByActive(fleet, list)
```
EricFortin: I want to make sure I understand this piece of code. This method returns the active gameserver set a a list of previous one. We then go on to set the replicas count to 0 on those and then deleting them. What happens if one of those replicas was allocated?

markmandel: Excellent question - any previously `Allocated` `GameServers` that part of the "inactive" `GameServerSet` are left alone, and the `GameServerSet` will stay alive until the `Allocated` `GameServer` shuts itself down or is otherwise terminated.

For reference you can see the code for scaling down `GameServerSet` [here](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameserversets/controller.go#L304), and how it ignores `Allocated` `GameServers`.

For a live demo, you can actually see it in action on [this twitch video](https://www.twitch.tv/videos/259103050) (only live for 13 more days, will push to YouTube at some point today).

markmandel: As an interesting note - this is why `deleteEmptyGameServerSets` ([code](https://github.com/GoogleCloudPlatform/agones/pull/199/files/980ad2dd64c2b1c28db2ae92d33cdf4aebd14c5e#diff-7c0697b565f7e02d6db88c116bb970f1R295)) only deletes a `GameServerSet` when the `gsSet.Status.Replicas == 0` - because this will never be 0 when there are still `Allocated` `GameServers`.

docs/create_fleet.md
markmandel: Done.

markmandel:
@enocom see anything that needs improvement in my Go code? :smile:

pkg/fleets/controller.go
```diff
@@ -173,39 +224,102 @@ func (c *Controller) syncFleet(key string) error {
...
+		gsSet.Spec.Replicas = replicas
+		gsSet, err := c.gameServerSetGetter.GameServerSets(gsSet.ObjectMeta.Namespace).Create(gsSet)
 		if err != nil {
 			return errors.Wrapf(err, "error creating gameserverset for fleet %s", fleet.ObjectMeta.Name)
 		}
 
 		c.recorder.Eventf(fleet, corev1.EventTypeNormal, "CreatingGameServerSet",
-			"Created GameServerSet %s", activeGsSet.ObjectMeta.Name)
+			"Created GameServerSet %s", gsSet.ObjectMeta.Name)
+	} else if replicas != gsSet.Spec.Replicas {
```
enocom: Rather than use an `else if` statement here, you could add a `return nil` to the `if` block above, and a `return nil` to the `else if` block. That way the control flow is easier to follow and more idiomatic.

markmandel: Updated. That should be better. Definitely seems more readable :+1:

markmandel:
@enocom @EricFortin gentle bump :smile: 

Please let me know if there is anything else to change, otherwise, would love an approval :blush: 

Thanks!

PR #205, opened 2018-05-15
Add formatting guidelines to CONTRIBUTING.md

### LINKED ISSUE #97: Add .editorconfig to ensure common formatting
See [here](http://editorconfig.org/).

Or, we could simply add details to `CONTRIBUTING.md`.

enocom:
Fixes #97

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,14 @@ Participation in this project comes under the [Contributor Covenant Code of Cond
 - Once review has occurred, please rebase your PR down to a single commit. This will ensure a nice clean Git history.
 - Finally - *Thanks* for considering submitting code to Agones!
 
+## Formatting
+
+When submitting pull requests, make sure to do the following:
+
+- Format all Go code with [gofmt](https://golang.org/cmd/gofmt/).
+- Remove trailing whitespace. Many editors will do this automatically.
+- Ensure any new files have [a trailing newline](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file)
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -31,6 +31,14 @@ Participation in this project comes under the [Contributor Covenant Code of Cond
 - Once review has occurred, please rebase your PR down to a single commit. This will ensure a nice clean Git history.
 - Finally - *Thanks* for considering submitting code to Agones!
 
+## Formatting
+
+When submitting pull requests, make sure to do the following:
+
+- Format all Go code with [gofmt](https://golang.org/cmd/gofmt/).
```
markmandel: `gofmt` and/or `goimport` ?

enocom: Hmm. Technically `gofmt`, but you'd be mad to write your own import statements and not use `goimports`.

enocom: How about another sentence after this one that says, "Many people use `goimports` which adds import statements automatically while also running `gofmt`"?

markmandel: SGTM.

markmandel:
Content LGTM - I'll repeat my earlier question, I think it got missed:

> Curious question - #97 does mention a .editorconfig but we don't have an .editorconfig , even though this is closing that ticket.
>
> Should we have a .editorconfig?

enocom:
@markmandel Since opening the issue, I have realized `gofmt` does what `.editorconfig` hopes to do. So, I think we're good without it.

Issue #177, opened 2018-04-14
Fix cyclomatic complexity in cmd/controller/main.go


EricFortin:
Remove the `// nolint` comment.

The main function has a cyclomatic complexity greater than 10. Refactor the function to lower it.

enocom:
There's a consistent pattern in `main.go` where we start a goroutine and call `Run` on something, reporting an error if the call fails.

With some small refactoring, we could do this:

```
var rs []runner = []runner{                                                                                                                
    wh, gsController, gsSetController, fleetController, faController, health,                                                              
}                                                                                                                                          
for _, r := range rs {                                                                                                                     
    go r.Run(stop)                                                                                                                         
}                                                                                                                                          
<-stop                                                                                                                                     
logger.Info("Shut down agones controllers")
```

Alternatively, if that seems too fancy, we could also simple extract a named function for each started goroutine.

Thoughts?

enocom:
I'll put together the PR with that pattern and we can revisit the question of making controllers implement a common interface.

PR #207, opened 2018-05-16
Shrink main func to resolve gocyclo warning

### LINKED ISSUE #178: Fix cyclomatic complexity in examples/simple-udp/server/main.go
Remove the // nolint comment.

The main function has a cyclomatic complexity greater than 10. Refactor the function to lower it.

enocom:
Fixes #178 

This is the simplest fix to the gocyclo linter error: breaking out functions of related code. I propose taking the same approach for #177.

### DIFF
--- a/examples/simple-udp/server/main.go
+++ b/examples/simple-udp/server/main.go
@@ -21,7 +21,6 @@ import (
 	"net"
 	"os"
 	"strings"
-
 	"time"
 
 	"agones.dev/agones/sdks/go"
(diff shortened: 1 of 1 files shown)


examples/simple-udp/server/main.go
```diff
@@ -50,7 +49,7 @@ func main() { // nolint: gocyclo
 	}
 
 	log.Print("Starting Health Ping")
-	stop := make(chan bool)
+	stop := make(chan struct{})
```
markmandel: Just for my own curiosity - why the change? Because it more closely matches `context.Context` and some of the Kubernetes constructs?

enocom: Since `stop` is used purely as a signally mechanism, and not as a way to transport meaningful information, I have followed the Go idiom of using `struct{}`, which indicates the value is of no meaning in and of itself. On the other hand, `bool` suggests we might send `true` or `false`, when in fact we don't really care about either.

markmandel: Thanks! That makes perfect sense!

PR #208, opened 2018-05-18
Testing functions for resource events

markmandel:
I kept writing the same two select statements a lot, so I pulled it into the central testing library for reuse.

### DIFF
--- a/pkg/testing/controller.go
+++ b/pkg/testing/controller.go
@@ -18,9 +18,12 @@ import (
 	"context"
 	"time"
 
+	gotesting "testing"
+
 	agonesfake "agones.dev/agones/pkg/client/clientset/versioned/fake"
 	"agones.dev/agones/pkg/client/informers/externalversions"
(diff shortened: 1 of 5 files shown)


pkg/testing/controller.go
```diff
@@ -84,3 +87,24 @@ func NewEstablishedCRD() *v1beta1.CustomResourceDefinition {
...
 	}
 }
+
+// AssertEventContains asserts that a k8s event stream contains a
+// value, and assert.FailNow() if it does not
+func AssertEventContains(t *gotesting.T, events <-chan string, contains string) {
+	select {
+	case e := <-events:
+		assert.Contains(t, e, contains)
+	case <-time.After(3 * time.Second):
```
enocom: Should this duration be configurable?

markmandel: I've never had a need to change it. The event action is usually (always?) synchronous in testing, 3 seconds is really a "just in case" - it could probably be dropped lower.

PR #209, opened 2018-05-18
Fix cyclomatic complexity in cmd/controller/main.go

### LINKED ISSUE #177: Fix cyclomatic complexity in cmd/controller/main.go

enocom:
Fixes #177.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"time"
 
@@ -43,13 +44,14 @@ import (
 )
 
 const (
-	sidecarFlag           = "sidecar"
-	pullSidecarFlag       = "always-pull-sidecar"
-	minPortFlag           = "min-port"
-	maxPortFlag           = "max-port"
-	certFileFlag          = "cert-file"
-	keyFileFlag           = "key-file"
-	controllerThreadiness = 2
+	sidecarFlag     = "sidecar"
+	pullSidecarFlag = "always-pull-sidecar"
+	minPortFlag     = "min-port"
+	maxPortFlag     = "max-port"
+	certFileFlag    = "cert-file"
+	keyFileFlag     = "key-file"
+	workers         = 2
+	defaultResync   = 30 * time.Second
 )
 
 var (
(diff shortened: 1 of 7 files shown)


EricFortin:
I would have merged it but it says the branch needs to be updated. There is also a button offering to update, should I do it or let @enocom do it himself?

markmandel:
Can we rebase this down to a single commit please?

enocom:
I'll update the branch and squash all the commits.

cmd/controller/main.go
```diff
@@ -101,95 +163,60 @@ func main() { // nolint: gocyclo
...
+		certFile:          certFile,
 	}
+}
 
-	kubeClient, err := kubernetes.NewForConfig(config)
-	if err != nil {
-		logger.WithError(err).Fatal("Could not create the kubernetes clientset")
-	}
+// ctlConfig stores all required configuration to create a game server controller.
+type ctlConfig struct {
```
markmandel: Super minor nit - should this just be `config` - it's not like we have any other configurations around? Happy if you ignore this, but thought I'd ask.

pkg/fleets/controller.go
```diff
@@ -49,6 +49,7 @@ import (
 // Controller is a the GameServerSet controller
 type Controller struct {
 	logger              *logrus.Entry
+	threadiness         int
```
markmandel: So looking at other controllers, such as [Kubernetes controllers](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/job_controller.go#L137) or the [sample controller](https://github.com/kubernetes/sample-controller/blob/master/controller.go#L156) they have a `Run()` signature of `Run(threadiness int, stopCh <-chan struct{})` (although some of the naming differs from place to place).

Should we be aligning with the Kubernetes convention here / are we losing anything by not doing that. Maybe we lose a bit of ability for people to transfer Controller building knowledge when moving to Agones? (although it's minimal)

This isn't something I'm _super_ worried about, but figured it was worth raising - and if we find it a real issue down the line, it's pretty easy to back out of.

WDYT?

enocom: That's a good discovery. I'll adjust to match it.

cmd/controller/main.go
enocom: Sure. I'll rename it.

markmandel:
LGTM - will hit approve & merge when it's rebased to a single commit :+1: 

Nice stuff!

enocom:
Just squashed the commits.

PR #210, opened 2018-05-18
helm namespaces

### LINKED ISSUE #146: Cannot run GameServer in non-default namespace

Kuqd:
Changes:
- Reworked helm parameters to be clearer.
- Update documentation and explain how to add/create new namespace to use
- Add a resources options for the controller (always good to have resources limit parameter)
- Scope down SDK permissions to it's destination namespace (RoleBinding from a ClusterRole with a namespace)

This fixes #146


install/helm/agones/values.yaml
```diff
@@ -14,28 +14,32 @@
 
 # Declare variables to be passed into your templates.
```
markmandel: oooh! I like this! :+1:

install/helm/README.md
```diff
@@ -27,6 +27,26 @@ The command deploys Agones on the Kubernetes cluster with the default configurat
...
+By default Agones is configured to work with game servers deployed in the default namespace. If you are planning to use other(s) namespace(s) you can configure Agones via the parameter `gameservers.namespaces`.
+
+For example to use `default` and `xbox` namespaces:
+
+```bash
+$ kubectl create namespace xbox
+$ helm install --set "gameservers.namespaces={default,xbox,ps4}"--name my-release agones
+```
+
+> You need to create your namespaces before installing agones.
```
markmandel: Interesting question - should we create the namespaces for them? Because we could - and it's idempotent, so if they already exist, nothing is going to break. WDYT?

install/helm/README.md
```diff
@@ -27,6 +27,26 @@ The command deploys Agones on the Kubernetes cluster with the default configurat
 
 > **Tip**: List all releases using `helm list`
 
+## Namespaces
+
+By default Agones is configured to work with game servers deployed in the default namespace. If you are planning to use other(s) namespace(s) you can configure Agones via the parameter `gameservers.namespaces`.
```
markmandel: > with game servers deployed in the `default` namespace

May want to add backticks around "default"

markmandel: > If you are planning to use other namespaces

I'd suggest dropping the (s) on both, I don't think they are needed, and I feel like it reads a bit better.

install/helm/README.md
Kuqd: I thought about this, wanted to ask on slack your opinion but it was late. I just need to make sure it doesn't block the deployment if the namespace exists.

markmandel: Yeah, good point - if it errors when the namespace already exists, then we shouldn't create the namespace - totally agreed.

Just thinking through some pros and cons otherwise:

Pros:
- Simpler install experience, since you just say what namespaces you want, and Helm will create them for you (if they don't already exist)

Cons:
- If you typo a namespace, you won't know it's broken, because the namespace will be created.

Can you think of any others? At least given the above, I'm leaning towards creating the namespaces, for the easier install experience. It's a bit more "turn key". WDYT?

markmandel: To capture our chat - keep things simple for now, and not implement this. Let's wait on feedback, as there is complexity with helm potentially owning the namespace.

Kuqd: the best option if we really want this is to use a pre-install and pre-upgrade hook helm. The hook will create namespace if they don't exist so the namespace is not tracked by helm (once not deleted when the release is deleted, could be dangerous), but that means the hook job needs permissions to create namespace, which complicate things, I also prefer to keep things simple and take action if a real need appears.

PR #213, opened 2018-05-18
Rolling updates for Fleets

### LINKED ISSUE #70: GameServer Fleets

markmandel:
This implements a configurable rolling update strategy for Fleets that also ensures that Allocated GameServers are not interrupted.

Also includes updates to documentation.

Closes #70


pkg/fleets/controller.go
```diff
@@ -309,16 +314,102 @@ func (c *Controller) deleteEmptyGameServerSets(fleet *stablev1alpha1.Fleet, list
...
+		}
+	}
+
+	return fleet.LowerBoundReplicas(fleet.Spec.Replicas - stablev1alpha1.SumStatusAllocatedReplicas(rest)), nil
+}
+
+// rollingUpdateDeployment will do the rolling update of the old GameServers
+// through to the new ones, based on the fleet.Spec.Strategy.RollingUpdate configuration
+// and return the replica count for the active GameServerSet
+func (c *Controller) rollingUpdateDeployment(fleet *stablev1alpha1.Fleet, active *stablev1alpha1.GameServerSet, rest []*stablev1alpha1.GameServerSet) (int32, error) {
```
markmandel: @enocom particularly love your eyes on this function, and the functions it calls - it's a bit complicated, so if you can see anything that could improve the code / readability, would definitely be interested.

pkg/fleets/controller.go
```diff
@@ -309,16 +314,102 @@ func (c *Controller) deleteEmptyGameServerSets(fleet *stablev1alpha1.Fleet, list
...
+	}
+	return replicas, nil
+}
+
+// rollingUpdateActive applies the rolling update to the active GameServerSet
+// and returns what its replica value should be set to
+func (c *Controller) rollingUpdateActive(fleet *stablev1alpha1.Fleet, active *stablev1alpha1.GameServerSet, rest []*stablev1alpha1.GameServerSet) (int32, error) {
+	replicas := active.Spec.Replicas
+
+	if active.Spec.Replicas < fleet.Spec.Replicas && active.Spec.Replicas == active.Status.Replicas {
```
enocom: I would invert this boolean expression such that the body becomes `return replicas, nil` as on line 381 below. That way the control flow of the function handles errors within if statements and happy paths outside. Does that make sense?

pkg/fleets/controller.go
```diff
@@ -309,16 +314,102 @@ func (c *Controller) deleteEmptyGameServerSets(fleet *stablev1alpha1.Fleet, list
...
+	}
+
+	r, err := intstr.GetValueFromIntOrPercent(fleet.Spec.Strategy.RollingUpdate.MaxUnavailable, int(fleet.Spec.Replicas), true)
+	if err != nil {
+		return errors.Wrapf(err, "error calculating scaling gameserverset: %s", fleet.ObjectMeta.Name)
+	}
+	unavailable := int32(r)
+
+	for _, gsSet := range rest {
+		if gsSet.Status.Replicas > 0 && gsSet.Status.Replicas == gsSet.Spec.Replicas {
```
enocom: Same thing here as above. If you invert the boolean check, your if statement can become much smaller with a continue in the body, and the code outside the if statement can be moved below, which makes understanding the flow easier.

pkg/fleets/controller.go
```diff
@@ -309,19 +314,119 @@ func (c *Controller) deleteEmptyGameServerSets(fleet *stablev1alpha1.Fleet, list
...
+	}
+	return replicas, nil
+}
+
+// rollingUpdateActive applies the rolling update to the active GameServerSet
+// and returns what its replica value should be set to
+func (c *Controller) rollingUpdateActive(fleet *stablev1alpha1.Fleet, active *stablev1alpha1.GameServerSet, rest []*stablev1alpha1.GameServerSet) (int32, error) {
+	replicas := active.Spec.Replicas
+
+	// if the active spec replicas are greater than or equal the fleet spec replicas, then we don't
```
markmandel: @enocom made the inversion here, and also put in some text to help explain the logic. WDYT?

pkg/fleets/controller.go
```diff
@@ -309,19 +314,119 @@ func (c *Controller) deleteEmptyGameServerSets(fleet *stablev1alpha1.Fleet, list
...
+	r, err := intstr.GetValueFromIntOrPercent(fleet.Spec.Strategy.RollingUpdate.MaxUnavailable, int(fleet.Spec.Replicas), true)
+	if err != nil {
+		return errors.Wrapf(err, "error calculating scaling gameserverset: %s", fleet.ObjectMeta.Name)
+	}
+	unavailable := int32(r)
+
+	for _, gsSet := range rest {
+		// if the status.Replicas are less than or equal to 0, then that means we are done
+		// scaling this GameServerSet down, and can therefore exit/move to the next one.
+		if gsSet.Status.Replicas <= 0 {
```
markmandel: @enocom the inversion here breaks this out into two sections, with slightly different logic - which I think is much easier to read :+1: and parse. WDYT?

pkg/fleets/controller.go
```diff
@@ -309,19 +314,119 @@ func (c *Controller) deleteEmptyGameServerSets(fleet *stablev1alpha1.Fleet, list
...
+			continue
+		}
+		// If the Status.Replicas does not equal the Status.Replicas for this GameServerSet, this means
```
EricFortin: If the `Spec.Replicas` does not equal ...

markmandel: Good catch. Done!

PR #214, opened 2018-05-20
Generate Certs for Mutation/Validatiion Webhooks

### LINKED ISSUE #169: Generate Certs for Mutation/Validatiion Webhooks
Now that we have #101 (Helm) we can use it to generate the cert for the validation and mutation webhook at installation time.

## Design

Looks like it should be relatively straight forward, as there is a `genSignedCert` function built into helm. See examples below.

We will need to change the controller so that it pulls the cert that it uses from a generated Secret.

This should be relatively straightforward as it already [reads the certs a file](https://github.com/GoogleCloudPlatform/agones/blob/master/cmd/controller/main.go#L65-L66) , we'll just need to not [add them in the Dockerfile](https://github.com/GoogleCloudPlatform/agones/blob/master/cmd/controller/Dockerfile#L3-L4) and make sure to mount the Secret in the [controller deployment](https://github.com/GoogleCloudPlatform/agones/blob/master/install/helm/agones/templates/controller.yaml)

## Research
https://github.com/lachie83/internallb-webhook-admission-controller/blob/master/charts/internallb-webhook-admission-controller/templates/admissionregistration.yaml
https://github.com/kubernetes/helm/blob/master/docs/charts_tips_and_tricks.md
http://masterminds.github.io/sprig/crypto.html

Kuqd:
- Add the ability to generate tls certs from helm for webhooks.
- By default it's on but you can pass your own certificates.
- Target gen-install  use our default certificate to stay idempotent
- certificates are passed by kubernetes secrets.
- Remove certs from controller container (Dockerfile)
- Add a target to remove gcloud deployment, usefull to reset
- Update helm chart documentation
- Generate new install.yaml

Tested on GKE !

Closes #169


install/helm/README.md
```diff
@@ -78,6 +78,7 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.controller.healthCheck.failureThreshold`       | Number of times before giving up (in seconds)                   | `3`                        |
 | `agones.controller.healthCheck.timeoutSeconds`         | Number of seconds after which the probe times out (in seconds)  | `1`                        |
 | `agones.controller.resources`  | Controller resource requests/limit | `{}`
+| `agones.controller.generateTls`  | Set to true to generate tls certificates or false to provide your own certificates in `certs/*` | `true`
```
markmandel: Small thing, but:
Tls => TLS, since it's an acronym?

markmandel:
Just ran through a quick e2e test - all worked great. So outside of the couple of questions above, this LGTM.

Kuqd:
Requested changes are in !

PR #215, opened 2018-05-22
Release process documentation

### LINKED ISSUE #203: Release cycle
As we head closer to what I think will be a 0.2 release - we should consider how we are doing releases going forward, and document that process.

In my mind there are two potential release cadences (although if there are more that people thing are appropriate, please feel free to put them forward).

1. We plan a set of priorities for each release, as we get close to completion of those priorities, the committers (and community) decide if we are ready for a release, slip anything we feel is worth slipping to the next release - go into feature freeze for a period (to be determined), only merge bug fixes and documentation PRs etc, and then release.
2. We have a regular cadence. For example, we will release every 2 months with whatever is in master. The last two weeks of the 2 month period we go into feature freeze and only accept bug fixes, and documentation submissions etc for that time. Then every priority issue (in that release's milestone) that isn't ready for that release rolls into the next release (the time periods I've outlined here are just arbitrary examples, we could do whatever). The only exception to this would be if there are critical bugs in the current release that would push a release out, or hotfixes/security issues that would need to come out post a release on an as-needed basis.

I think that while we are in alpha, I don't think we need to worry about doing release candidate builds/pre-release builds - as essentially everything is pre-release - but also happy to hear other opinions if people feel strongly otherwise (maybe we should still cut an rc to make user testing easier?). Once we hit a 1.0, I feel that this release cycle would change, but we can discuss that then.

I lean more towards option (2) because then users and contributors are always aware of what is coming, and when it is coming by and can plan appropriately,  but would love to hear thoughts on either of these approaches (or other ones if you think they will fit) - each has pros and cons. If we can choose a general approach, then we can look at narrowing down the specifics on that process.

markmandel:
Outlines a 6 week release process, with included templates for a release issue and github release.

The `make do-release` will need to be updated once this has been approved, and `make gen-changelog` will also need to implemented.

Closes #203

### DIFF
--- a/docs/governance/release_process.md
+++ b/docs/governance/release_process.md
@@ -1,0 +1,32 @@
+# Agones Release Management
+
+# Release Cadence
+
+- Versioned releases will occur every 6 weeks
+- Releases occur on a Tuesday.
+- 5 week development cycle, at the end of a which a Release Candidate (RC) will be released with the contents of master.
+- For the next week, the project is in "feature freeze" - i.e. only bug and documentation (.md and examples) fixes during this time.
+  - Any new PRs that are submitted during feature freeze, will be tagged with the label `merge-after-release` 
+    to delineate that they should only be merged after the full release is complete. 
+- At the end of the RC week, the complete version release will occur.
+
+## Release Calendar
+
+> Release Calendar forthcoming once the 0.2 release is complete, when the scheduled release cadence will start.
+
+# Release Process
+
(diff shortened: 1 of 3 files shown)


Kuqd:
Should we talk about versioning ? Are those major releases ?

markmandel:
I figured we'd outline a process for the while we are currently in alpha. When we are ready to go 1.0, then let's worry about semver and when major versions happen, etc. Start simple :smile:

PR #220, opened 2018-05-23
add the ability to turn off RBAC in helm and customize gcp test-cluster

### LINKED ISSUE #211: helm RBAC on/off
we should have an option in helm to skip rbac resources creation if they are not required by the cluster.

see:
- https://github.com/istio/istio/search?utf8=%E2%9C%93&q=rbacEnabled&type=
- https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/values.yaml#L51

Kuqd:
- You can remove RBAC using the helm chart. (`agones.rbacEnabled` default true)
- Added variable in the Makefile to customize your GCP test cluster (zone,name,machineType,size,)
- Update helm documentation

Tested on GKE with RBAC off.

Closes #211


build/Makefile
```diff
@@ -38,7 +38,11 @@ REGISTRY ?= gcr.io/agones-images
 # Where the kubectl configuration files are being stored
 KUBEPATH ?= ~/.kube
 # The (gcloud) test cluster that is being worked against
-CLUSTER_NAME ?= test-cluster
+GCP_CLUSTER_NAME ?= test-cluster
+GCP_CLUSTER_ZONE ?= us-west1-c
+GCP_CLUSTER_LEGACYABAC ?= false
+GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT ?= 3
+GCP_CLUSTER_NODEPOOL_MACHINETYPE ?= n1-standard-4
```
markmandel: Should we move
```
GCP_CLUSTER_LEGACYABAC ?= false
GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT ?= 3
GCP_CLUSTER_NODEPOOL_MACHINETYPE ?= n1-standard-4
```
To being target specific variables to `gcloud-test-cluster`, since they aren't reused across more than 1 target.

WDYT?

([here is an example](https://github.com/GoogleCloudPlatform/agones/blob/master/build/Makefile#L121), if you aren't familiar with them)

Kuqd: Good idea, will do !

build/README.md
```diff
@@ -133,13 +133,26 @@ we will need to authenticate out gcloud tooling against it. To do that run `make
...
+|---------------------------------------|-------------------------------------------------------------------------------|---------------|
+| `GCP_CLUSTER_NAME`                    | The name of the cluster                                                       | `test-cluster`  |
+| `GCP_CLUSTER_ZONE`                    | The name of the Google Compute Engine zone in which the cluster will resides. |  `us-west1-c`   |
+| `GCP_CLUSTER_LEGACYABAC`              | Enables or disables the [ABAC](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters#LegacyAbac) authorization mechanism on a cluster.            | `false`         |
+| `GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT`| The number of nodes to create in this cluster.                                |  `3`            |
+| `GCP_CLUSTER_NODEPOOL_MACHINETYPE`    | The name of a Google Compute Engine machine type.                             | `n1-standard-4` |
+
+If you would like to change more settings, feel free to edit the `deployment.yaml` file before running this command.
```
markmandel: Didn't you delete the deployment.yaml? :blush:

install/helm/README.md
```diff
@@ -47,6 +47,11 @@ $ kubectl create namespace ps4
 $ helm upgrade --set "gameservers.namespaces={default,xbox,ps4}" my-release agones
 ```
 
+## RBAC
+
+If role-based access control (RBAC) is enabled in your cluster, you must set `agones.rbacEnabled` to true.
```
markmandel: I'm going to suggest we change this to the following, as it sounds like I need to take action if I have RBAC enabled (which most people do), when in fact I don't. How does this sound:

---

By default, `agones.rbacEnabled` is set to true. This enable RBAC support in Agones and must be true if RBAC is enabled in your cluster.

The chart will take care of creating the required service accounts and roles for Agones.

If you have RBAC disabled, or to put it another way, ABAC enabled - you should set this value to `false`.

---

WDYT?

PR #222, opened 2018-05-24
Update tools - vetshadow is more aggressive

markmandel:
I noticed the latest vetshadow seems to be far more aggressive with the newer version.

So updating the build image, and fixing all the issues.


markmandel:
This may need to wait until #221 is merged, and can be updated into this branch. 100s may not be long enough.

PR #228, opened 2018-05-25
do-release release registry and upstream push

markmandel:
This PR does two things:
1) Ensure the built images always have the release registry
2) Ensure the release branch is pushed to upstream, rather than to a forked origin.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -26,6 +26,9 @@
 # base version target. This is usually the next release.
 base_version = 0.2.0
 
+# agones image release registry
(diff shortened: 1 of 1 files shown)


markmandel:
That's a good question - I've just been using `upstream` for years since I probably read it somewhere on the internet that is was a thing :smile: 

Looks like github also has the same thing - https://help.github.com/articles/configuring-a-remote-for-a-fork/

Is it actually part of some tooling that sets it up that way as well?

PR #245, opened 2018-06-06
Make Cloud Builder Faster

markmandel:
This should make Cloud Builder faster
- Request an 8 core machine
- Move the test and build steps to be parralel.


markmandel:
😲 Build time: 6 min 5 sec

PR #249, opened 2018-06-08
helm namespace

### LINKED ISSUE #212: helm namespace
We should follow helm convention and use --namespace (and .Release.Namespace) and drop the current `agones.namespace` parameter.

We should also update the installation documentation to explain this change.

```
helm install install/helm/agones --name agones --namespace agones-system
```

Kuqd:
The namespace is now created by helm and not our template. This is the convention helm recommend and the one istio uses so helm user won't be lost.

It's also more compatible for use cases when you already have the namespace created.

Tested on GKE !

Closes #212


Kuqd:
closing this I realized I've created the branch not on the fork !

PR #250, opened 2018-06-08
use the helm --namespace convention 

### LINKED ISSUE #212: helm namespace

Kuqd:
The namespace is now created by helm and not our template. This is the convention helm recommend and the one istio uses so helm user won't be lost.

It's also more compatible for use cases when you already have the namespace created.

Tested on GKE !

Closes #212


markmandel:
So my only question about this, is - aren't we also doing stuff in other namespaces - i.e. creating service accounts in the default namespace. Is there any issue about potential confusion/conflict there? Otherwise I'm personally not seeing any issue here.

Also, if we are going to do this, we need to also update the Helm install documentation (and maybe have two versions - the 0.2.0 install steps, and then the development install steps, with the standard warning)

Kuqd:
I'll check if it works when not supplying the --namespace but for sure it's less confusing since right now if you have the `agones-system` already existing it will fail.

markmandel:
SGTM - then the only outstanding thing for me is an update to the docs. Otherwise, SGTM.

install/helm/README.md
```diff
@@ -21,11 +21,13 @@ Download the latest `agones-install` zip from the [releases](https://github.com/
 $ wget https://github.com/GoogleCloudPlatform/agones/releases/download/v0.2.0/agones-install-0.2.0.zip
 $ unzip agones-install-0.2.0.zip
 $ cd install/helm/
-$ helm install --name my-release agones
+$ helm install --name my-release --namespace agones-system agones
```
markmandel: Running this against 0.2.0 results in:
```bash
root@markmandel:/go/src/agones.dev/agones# helm install --name test-release --namespace agones-system ./install/helm/agones/
Error: release test-release failed: namespaces "agones-system" already exists
```

I think we'll need to split this - keep the original version for 0.2.0, and then do a development version with the standard :warning: signs stating that it's development only, and we can switch when 0.3.0 comes out.

Kuqd: I feel like this start to be confusing having two type of install. Should we wait then for a release to get that PR in or hotfix 0.2.0 ?

markmandel: So I was thinking about something like the following:
https://gist.github.com/markmandel/a4468aa62c6b7793ce36bc398627d402

Two very separate sections -- then when we go to do the 0.3.0 release, we can do our usual documentation cleanup and remove the older install version.

WDYT?

Kuqd: Done, I modified your example to make it works with the master branch.

I'm wondering should we have a latest to make it simpler to use the last dev build ? I could then change my doc to use latest.

Kuqd: @markmandel

PR #265, opened 2018-06-13
Expose SDK-Server at HTTP+JSON

### LINKED ISSUE #240: Expose SDK Sidecar GRPC Server as HTTP+JSON
# Overview

Provide a REST / HTTP+JSON interface to the SDK gRPC Sidecar service.

# Motivation

1. The tooling around C++ on Windows is basically a deal breaker. It involved [installing 7 things](https://github.com/grpc/grpc/blob/v1.12.x/INSTALL.md#building-using-cmake-recommended) and deviated quite considerably from the tooling around installing on Linux/MacOS
1. gRPC on Unreal and Unity tends to be difficult to impossible. But both can do HTTP+JSON.  This will make integration with these commercial engines much easier.
1. Any other language that gRPC doesn't support out of the box, this provides an easy to use alternative.

# Design
I see that there are two options, each with it's own pros and cons:

## Envoy Proxy
- Pro: It's less tightly integrated with the SDK process.
- Pro: it has a lot of development and backing behind it.
- Con: It would be a separate process / separate Pod, which adds complexity.
- Pro: It also exposes metrics around usage and consumption. They may come in handy down the line.

We could do local development with a `docker run` command that runs both Envoy and the SDK in local mode.

## gRPC Gateway

- Pro: It looks like we can run this in the same process as the gRPC server. This means we can keep the single binary for the sidecar as well as for local development.
- Pro; it's simpler. It's built for just the one purpose that we want.
- Con: I'm not sure that the grpc-gateway has as much contribution / backing at Envoy.
  (but looks like releases are [fairly frequent](https://github.com/grpc-ecosystem/grpc-gateway/releases))
- Question: How well does the grpc-gateway support streaming rpcs?

# Research

## Envoy
- https://www.envoyproxy.io/docs/envoy/v1.6.0/configuration/http_filters/grpc_json_transcoder_filter
- https://www.envoyproxy.io/docs/envoy/v1.6.0/configuration/http_filters/grpc_http1_bridge_filter.html?highlight=grpc

## grpc-gateway
- https://grpc-ecosystem.github.io/grpc-gateway/docs/usage.html

markmandel:
Implement grpc-gateway in front of the gRPC based sdk-server, so that it can be access via HTTP+JSON.

This includes documentation and a swagger/openapi specification.

This also has been implemented such that the sdk-server is still a single binary, and as such, the HTTP+JSON interface can still be used for local development.

Closes #240


cmd/sdk-server/main.go
```diff
@@ -56,56 +59,33 @@ var (
 )
 
 func main() {
```
markmandel: @enocom I refactored this to look more like the controller main, so I could also do the grpc-gateway proxy.

Would love your eyes on it.

cmd/sdk-server/main.go
```diff
@@ -56,56 +59,33 @@ var (
...
-	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
-	runtime.Must(viper.BindEnv(localFlag))
-	runtime.Must(viper.BindEnv(gameServerNameEnv))
-	runtime.Must(viper.BindEnv(podNamespaceEnv))
-	runtime.Must(viper.BindEnv(healthDisabledFlag))
-	runtime.Must(viper.BindEnv(healthTimeoutFlag))
-	runtime.Must(viper.BindEnv(healthInitialDelayFlag))
-	runtime.Must(viper.BindEnv(healthFailureThresholdFlag))
-	runtime.Must(viper.BindPFlags(pflag.CommandLine))
+	ctlConf := parseEnvFlags()
```
enocom: This is a nice touch. Makes it easy to quickly read through main and stop on env flags only if it's necessary.

cmd/sdk-server/main.go
```diff
@@ -36,15 +38,16 @@ import (
...
+	httpPort = 59358
 
 	// specifically env vars
 	gameServerNameEnv = "GAMESERVER_NAME"
 	podNamespaceEnv   = "POD_NAMESPACE"
 
 	// Flags (that can also be env vars)
 	localFlag                  = "local"
-	addressFlag                = "address"
+	addressFlag                = "Address"
```
enocom: What's the rationale for this change?

cmd/sdk-server/main.go
```diff
@@ -128,28 +108,101 @@ func main() {
...
 			logger.WithError(err).Fatalf("Could not start sidecar")
 		}
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
 
 		go s.Run(ctx.Done())
 		sdk.RegisterSDKServer(grpcServer, s)
 	}
 
+	run(ctx, grpcServer, lis, grpcEndpoint, mux, httpServer)
```
enocom: More a matter of style, but I prefer making functions synchronous and then prefacing them with the `go` keyword.

cmd/sdk-server/main.go
```diff
@@ -128,28 +108,101 @@ func main() {
...
 		}
 	}()
 
-	<-stop
-	logger.Info("shutting down grpc server")
-	// don't graceful stop, because if we get a kill signal
-	// then the gameserver is being shut down, and we no longer
-	// care about running RPC calls.
-	grpcServer.Stop()
+	conn, err := grpc.DialContext(ctx, grpcEndpoint, grpc.WithBlock(), grpc.WithInsecure())
```
enocom: That `grpc.WithBlock` is a tricky API isn't it. Nice usage.

cmd/sdk-server/main.go
```diff
@@ -128,28 +108,101 @@ func main() {
...
 	}
 
+	run(ctx, grpcServer, lis, grpcEndpoint, mux, httpServer)
+
+	<-stop
+	logger.Info("shutting down sdk server")
+}
+
+// run runs all the services
+func run(ctx context.Context, grpcServer *grpc.Server, lis net.Listener, grpcEndpoint string,
```
enocom: One way to get around all the arguments to `run` would be to introduce a type, maybe `sdkServer` and attach `run` to it. But again, style. Not a big deal.

cmd/sdk-server/main.go
markmandel: I just copied the pattern in the other main.go :+1: no credit to me!

cmd/sdk-server/main.go
markmandel: whoops. Probably a refactor gone awry. Rolled back.

markmandel:
@enocom I think I fixed up most of your concerns - PTAL!

cmd/sdk-server/main.go
markmandel: Good call. splitting this out into two synchronous functions reads a lot better.

build/build-image/gen-grpc-cpp.sh
```diff
@@ -14,12 +14,32 @@
...
+protoc -I ${googlepais} -I . --cpp_out=./sdks/cpp sdk.proto ${googlepais}/google/api/annotations.proto  ${googlepais}/google/api/http.proto
+
+mkdir -p /tmp/cpp
+
+cd ./sdks/cpp
+header sdk.pb.h
+header sdk.grpc.pb.cc
+header sdk.pb.cc
+
+cd ./google/api/
```
Kuqd: what is this for ? I didn't know that adding the http gateway will affect the grpc client gen.

sdks/README.md
```diff
@@ -75,6 +79,24 @@ $ ./sidecar.linux.amd64 --local
...
+
+If client generation is well supported by [gRPC](https://grpc.io/docs/), then generate a client from the
+[sdk.proto](../sdk.proto), and look at the current [sdks](.) to see how the wrappers are implemented to make interaction
+with the SDK server simpler for the user.
+
+#### REST API Implementation
+
+If client generation is not well supported by gRPC, or if there are other complicating factors, implement the SDK through
+the [REST](../docs/sdk_rest_api.md) HTTP+JSON interface. This could be written by hand, or potentially generated from
+the [Swagger/OpenAPI Spec](../sdk.swagger.json).
```
Kuqd: Should we show an example on how to do that ?

```
docker run --rm -v ${PWD}:/local swaggerapi/swagger-codegen-cli generate -i /local/sdks/sdk.swagger.json  -l cpprest -o /local/out/cpp
```

Do you plan to add a SDK using the http gateway ?

build/build-image/gen-grpc-cpp.sh
markmandel: It does because the annotations.proto is included in the sdk.proto file, it generated .cc and .h files that are used in places.

Also it does because the gen is never just the client - it's the client AND the server parts, we just use the client parts -- I don't think there is a way to ask to just generate the client AFAIK. (Also, this grpc cpp stuff will go away eventually, so for now, making it "just work" is more of a priority so we can replace the whole thing, IMHO)

sdks/README.md
markmandel: Ooh. I like this idea. I'd like to point to a central documentation too - is this it?
https://swagger.io/docs/open-source-tools/swagger-codegen/

Re: to add an SDk using the gateway. I wasn't going to to do it with this PR, I figure Unreal + CPP would end up building on top of this work, and be the examples. We could likely follow the same pattern as the existing SDKs (light wrapper around the possibly generated client), but was waiting to see what an actual implementation would look like. WDYT?

Kuqd: Yes that's the link, the plan sounds good !

build/build-image/gen-grpc-cpp.sh
Kuqd: Thanks

PR #271, opened 2018-06-16
Move cluster node addition/removal out of "experimental"

### LINKED ISSUE #60: Support Cluster Node addition/deletion
Need to ensure that when the Cluster is grown / shrunk, this is handled gracefully.

- Port Allocation will need a workerqueue for node deletion, in case the master goes down.
- Fleets will solve the issue with non-allocatable gameservers due to port exhaustion/issues. They will be marked as Unhealthy, and then retried later.

markmandel:
Implemented workerqueue for node modifications in the PortAllocator, so that if the master happens to go
down, then the operations the PortAllocator has to do to keep things in check will retry and result in a correct state.

Closes #60


pkg/gameservers/portallocator.go
```diff
@@ -83,31 +91,27 @@ func NewPortAllocator(minPort, maxPort int32,
...
 
 	pa.gameServerInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		DeleteFunc: pa.syncDeleteGameServer,
 	})
 
 	// Experimental support for node adding/removal
 	pa.nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: pa.syncAddNode,
+		AddFunc: func(obj interface{}) {
+			node := obj.(*corev1.Node)
```
enocom: I don't know the semantics of the event handler, but is there a chance that `obj` isn't of `*corev1.Node` type? In other words, is this a safe type assertion?

markmandel: A good question - basically, there isn't really. But a good question!

If something other than a Node comes through a Node event handler, then something _very_ weird has happened -- at which point, it really should panic.

Here another example in the Kubernetes sample-controller doing similar things:
https://github.com/kubernetes/sample-controller/blob/master/controller.go#L129

enocom: Then panicking is exactly the right thing to do.

enocom:
@markmandel If you rebase onto master, I'll be happy to merge. 😄

PR #272, opened 2018-06-20
Added list of various libs used in code

mean-mango:
I've seen this practice in other open source git repos and find it rather useful so wanted to propose adding it :)

To me it serves two goals:
1) recognizing others' work that make this project possible and
2) quick cheat sheet for new contributors before diving into the code. For example it helps me skip over chasing lib code initially and focus on Agones' native code making it faster to learn what it's trying to achieve.


build/README.md
```diff
@@ -358,3 +359,12 @@ instead of `make shell` to start an interactive shell for development on Minikub
...
 Use TAG to specify the image to transfer into minikube
+
+## Built With
+
+- [healthcheck](https://github.com/heptiolabs/healthcheck)
+- [errors](https://github.com/pkg/errors)
+- [pflag](https://github.com/spf13/pflag)
+- [viper](https://github.com/spf13/viper)
+- [grpc](https://godoc.org/google.golang.org/grpc)
+- [context](https://golang.org/pkg/context/)
```
markmandel: I feel like context is a core library in go, so it's probably not worth listing here.

Do we want to just list everything in here:
https://github.com/GoogleCloudPlatform/agones/blob/master/Gopkg.toml

(Or conversely, just point people to it?)

markmandel: @enocom wdyt?

enocom: A link for people unaware of dep's convention of using `Gopkg.toml` might be nice.

markmandel: I don't have strong opinions on this one.

Maybe something like "This project uses `dep` (with link) as a dependency manager. You can see the list of dependencies here (link)"

That's probably the most valuable without having to (a) rewrite docs for another project (b) keep a list up to date in two places.

WDYT?

mean-mango: SGTM!

Issue #277, opened 2018-06-27
Session data propagation to dedicated server


defat:
This is a feature request.
There are two options to get dedicated server for game session using Agones: create or allocate it from fleet. Using creation option session data such as map/game mode/level etc. can be passed to dedicated server with environment variables listed in pod template. Allocation from fleet lacks such possibility.
It would be great to have Agones SDK calling some callback method _OnAllocation_ implemented on dedicated server with some _AllocationParams_ object containing _map<const char*, const char*> SessionData_ field. This session data can be passed through FleetAllocation object.
Another cool feature would be complting fleet allocation (through setting it status to Ready or smth like this) only after dedicated server loaded level and is fully ready to accept players' connections. Agones SDK could be extended with _Loaded()_ method for this.

markmandel:
So thanks for bringing this up - being able pass data to the `GameServer` has been front of mind for me recently as well (I'm going to limit this ticket to this one issue for now - if we want to talk about status management, let's start a different ticket)

My current thinking has been more along the following, although I think we end up in the same place - let me know what you think:

1) Be able to access the `GameServer` crd data via the SDK. This is useful for several reasons (game server processes could know their own IP) - but also, since GameServers have annotations and labels, this could be a mechanism for passing custom data.

2) Extend `FleetAllocation` to have some kind of configuration option to manipulate the label/annotation information on the GameServer before marking it as Allocated - this gives you a mechanism to basically pass arbitrary key value pairs down to the `GameServer` -- and should have some other nice other applications as well (and be introspectable by standard Kubernetes tooling)

Maybe something like:

```yaml
apiVersion: "stable.agones.dev/v1alpha1"
kind: FleetAllocation
metadata:
  generateName: fleet-allocation-example-
spec:
  fleetName: fleet-example
  metadata:
     labels:
       foo: bar
     annotations:
       thing: stuff   
```

Wherein the `metadata` section would apply the labels and annotations to the `GameServer` object before moving it to an `Allocated` state. (open to better names that just `metadata` if anyone can think of one)

The interesting question will be - having a callback such that the SDK will send a notification whenever a State changes on the GameServer process. I think we can do that - and we should do it for all State changes, to make it as applicable to as many things as possible.

How does all that sound to you?

defat:
Yeah, this sounds extremely convenient, especially if annotations will be able to contain some complex structures such as json objects (maybe quoted to be strings).
Thank you for such fast reply!

[...]

PR #288, opened 2018-07-15
SDK `GameServer()` function for retrieving backing GameServer configuration

markmandel:
Extended the gRPC SDK to include a `GameServer()` function for both Go, C++
and REST by extending the backing `sdk.proto` and implementing the functionality
in the sdk sidecar.

This is the start of work needed for #277.

### DIFF
--- a/docs/sdk_rest_api.md
+++ b/docs/sdk_rest_api.md
@@ -63,3 +63,46 @@ Call when the GameServer session is over and it's time to shut down
 ```bash
 $ curl -d "{}" -H "Content-Type: application/json" -X POST http://localhost:59358/shutdown
 ```
+
+### GameServer
+
+⚠️⚠️⚠️ **/gameserver is currently a development feature and has not been released** ⚠️⚠️⚠️
+
+Call when you want to retrieve the backing `GameServer` configuration details
+
+- Path: `/gameserver`
+- Method: `GET`
+
+```bash
+$ curl -H "Content-Type: application/json" -X GET http://localhost:59358/gameserver
+```
+
+Response:
+```json
+{
+    "object_meta": {
+        "name": "local",
+        "namespace": "default",
+        "uid": "1234",
+        "resource_version": "v1",
+        "generation": "1",
+        "creation_timestamp": "1531795395",
+        "annotations": {
+            "annotation": "true"
+        },
+        "labels": {
+            "islocal": "true"
+        }
+    },
+    "status": {
+        "state": "Ready",
+        "address": "127.0.0.1",
+        "ports": [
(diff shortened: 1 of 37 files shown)


docs/sdk_rest_api.md
```diff
@@ -62,4 +62,17 @@ Call when the GameServer session is over and it's time to shut down
...
+⚠️⚠️⚠️ **/gameserver is currently a development feature and has not been released** ⚠️⚠️⚠️
+
+Call when you want to retrieve the backing `GameServer` configuration details
+
+- Path: `/gameserver`
+- Method: `GET`
+
+```bash
+$ curl -H "Content-Type: application/json" -X GET http://localhost:59358/gameserver
+```
```
Kuqd: Should we show an expected response or link to the model?

pkg/gameservers/localsdk_test.go
```diff
@@ -48,4 +48,8 @@ func TestLocal(t *testing.T) {
 	close(stream.msgs)
 
 	wg.Wait()
+
+	gs, err := l.GetGameServer(ctx, e)
+	assert.Nil(t, err)
+	assert.Equal(t, "local", gs.GetObjectMeta().Name)
```
Kuqd: You can equal a full object with testity.

install/helm/agones/templates/serviceaccounts/sdk.yaml
```diff
@@ -38,7 +38,7 @@ metadata:
 rules:
 - apiGroups: ["stable.agones.dev"]
   resources: ["gameservers"]
-  verbs: ["get", "update"]
+  verbs: ["get", "list", "update"]
```
Kuqd: Do we still need get since you have a cache sdk side with the list ?

sdk.swagger.json
```diff
@@ -98,8 +115,125 @@
     }
   },
   "definitions": {
+    "GameServerObjectMeta": {
+      "type": "object",
```
Kuqd: Nice

docs/sdk_rest_api.md
markmandel: Good call. Added expected response.

pkg/gameservers/localsdk_test.go
markmandel: Good call. Changed.

install/helm/agones/templates/serviceaccounts/sdk.yaml
markmandel: Nicely spotted! No we do not. Removed `get`.

Issue #37, opened 2017-12-24
End to End test


markmandel:
Write a end to end test - possibly, using make and kubectl, or maybe use code - what is the best option?
- installation
- Creating a GameServer
- Connecting

Will likely need to both check that a connection works with the go sample (edit to send a message via a flag, rather than stdin), but also will need to check the C++ sdk works as well.

I din't think there are any e2e integration test libraries that exist for Kubernetes platforms - but do have a look first. May be worth looking at how Kubernetes itself does it.

addisonbair:
I am new to this project and haven't yet stood up the development environment yet, but it appears that Kubernetes itself uses Ginkgo, Gomega and the Kubernetes API itself for much of the end-to-end testing of a particular release.

https://github.com/kubernetes/kubernetes/tree/master/test/e2e

Here's a simpler example from the Prometheus Operator project:

https://github.com/coreos/prometheus-operator/tree/master/test/e2e

Kuqd:
Should I start by some simple e2e go test and may be bring this later into an image to support in-cluster config.

I quite like the prom-operator example.

markmandel:
The other option is using the Helm test tooling? Or maybe there is a combo option?

Or if I follow - you're saying, start with a simple binary we can run and point at a cluster, and then move forward from there? That would make sense (assuming that's what you are suggesting)

Kuqd:
yes, I want to add the helm integration after, using a container.

markmandel:
Sounds like a good approach to me!

markmandel:
Got another one for you - if health checking is disabled, then the game server shouldn't be marked as unhealthy if no health pings are sent (just found this as a bug, and it's only going to get covered by e2e test)

But yes - we should prioritise, and we can work through them slowly each as a separate PR. Also happy to defer to your opinion on the order.

markmandel:
Just updated this with what I completed on e2e tests.

thisisnotapril:
Repo health check - should we keep this issue open?

markmandel:
There are still a few left on the list (I cleaned things up a bit above) that would be good to have I think.

markmandel:
I think this is actually good to close now. If anyone disagrees, please let me know, otherwise I'll close this next Monday.

markmandel:
Closing!

PR #303, opened 2018-07-23
Adds the first e2e test

Kuqd:
This is a sample of one e2e tests for #37 . This first test create a gameserver, waits for readiness, assert some values then finally ping it (send/receive).

The idea of this PR is to see if the direction is correct, but also I'd like to plug it to cloud builder to see if we can make it works @markmandel I'll need your help on this. ( Can we use your account to provision a cluster for e2e ? How should we proceed ?)

I would like that we list all tests for gameservers/fleets/allocation, so I can tackled them one by one. (Let's use the ticket #37 )

Locally with a GKE cluster it works.
(I'm having sometimes some timeout with the k8s api, might be from my connection
`Post https://35.185.224.168/apis/stable.agones.dev/v1alpha1/namespaces/default/gameservers: dial tcp 35.185.224.168:443: i/o timeout`)

Have a spin, I hope you like it.


Kuqd:
I can confirm there is something weird with gcloud toolchain, if I use kubectl with my machine, it doesn't work anymore in the container.(MacOS)

markmandel:
Swamped with Cloud NEXT - so will likely take a look once things are done / calmed down a bit, so I may not respond until next week.

build/build-image/run-individual-e2e.sh
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cd ./build
+kubectl port-forward statefulset/consul-consul 8500:8500 &
+# waits for port-forward to setup
+sleep 5
+echo "Starting e2e tests"
+consul lock -timeout 5m -try 5m -verbose LockE2E /root/e2e.sh
+killall -q kubectl
```
Kuqd: I might need to add a newline here

Kuqd:
what a ride ! I'll need to cleanup now

Kuqd:
I'll open a new one for cleaner review !

PR #307, opened 2018-07-27
Session data propagation to dedicated server

victor-prodan:
This is the implementation of https://github.com/GoogleCloudPlatform/agones/issues/277

Add metadata to fleet allocation specification that is added to game server status in the moment of allocation. The server can retrieve this data using GameServer() call.


examples/fleetallocation.yaml
```diff
@@ -28,4 +28,9 @@ metadata:
 spec:
   # The name of the fleet to allocate from. Must be an existing Fleet in the same namespace
   # as this FleetAllocation
-  fleetName: fleet-example
\ No newline at end of file
+  fleetName: fleet-example
+  # Custom metadata that is added to game server status in the moment of allocation
+  # You can use this to tell the server necessary session data
+  metadata:
```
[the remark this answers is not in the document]
markmandel: I don't mind this - but I think it should be split into labels and annotations - so you have a choice which one you want to apply.

pkg/apis/stable/v1alpha1/gameserver.go
```diff
@@ -164,6 +164,7 @@ type GameServerStatus struct {
 	Ports    []GameServerStatusPort `json:"ports"`
 	Address  string                 `json:"address"`
 	NodeName string                 `json:"nodeName"`
+	AllocationMeta map[string]string  `json:"allocationMeta,omitempty"`
```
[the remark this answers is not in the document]
markmandel: I have to respectfully very much disagree with having a separate metadata here. I wrote a longer explanation on #277

victor-prodan:
Thanks for your feedback!

victor-prodan:
Sorry guys... I don't understand what is failing now 😐 Tests are ok on my side

pkg/fleetallocation/controller.go
```diff
@@ -279,7 +279,11 @@ func (c *Controller) allocate(f *stablev1alpha1.Fleet) (*stablev1alpha1.GameServ
 
 	gsCopy := allocation.DeepCopy()
 	gsCopy.Status.State = stablev1alpha1.Allocated
-
+    
+    if fam != nil {
```
markmandel: This will need formatting - if you `make shell` you can run: `goimports -w ./pkg/` and you will be good to go :+1:

pkg/fleetallocation/controller.go
```diff
@@ -288,3 +292,26 @@ func (c *Controller) allocate(f *stablev1alpha1.Fleet) (*stablev1alpha1.GameServ
 
 	return gs, nil
 }
+
+// patch the labels and annotations of an allocated GameServer with metadata from a FleetAllocation
+func (c *Controller) patchMetadata(gs *stablev1alpha1.GameServer, fam *stablev1alpha1.FleetAllocationMeta) () {
+	// patch ObjectMeta labels
+	if fam.Labels != nil {
+		if gs.ObjectMeta.Labels == nil {
+			gs.ObjectMeta.Labels = make(map[string]string, 1)
```
markmandel: @enocom should this be `make` or should this be `gs.ObjectMeta.Labels = map[string]string{}` ?

pkg/fleetallocation/controller.go
```diff
@@ -288,3 +292,26 @@ func (c *Controller) allocate(f *stablev1alpha1.Fleet) (*stablev1alpha1.GameServ
...
+		}
+		for key, value := range fam.Labels {
+			gs.ObjectMeta.Labels[key] = value
+		}
+	}
+
+	// apply annotations patch
+	if fam.Annotations != nil {
+		if gs.ObjectMeta.Annotations == nil {
+			gs.ObjectMeta.Annotations = make(map[string]string, 1)
```
markmandel: @enocom same as above. WDYT?

pkg/fleetallocation/controller.go
enocom: There almost synonymous. I say go with whatever you do elsewhere in the codebase. The use of `make` looks a little less common to my eyes.

pkg/fleetallocation/controller.go
enocom: Unless `Annotations` needs to be a small map, the object literal syntax `map[string]string{}` seems more idiomatic to me.

markmandel: We actually don't know the size of the map at all (use configurable).
I'd be inclined to do a `map[string]string{}` myself. 

Personally I only lean on make if I know the size of the map I'm already going to create.

So actually, if we want to keep the `make` we could do:
`gs.ObjectMeta.Annotations = make(map[string]string, len(fam.Annoatations))`
(and similar for the example above)

That would likely actually be the most efficient?

What do you all think?

enocom: It's a question of readability more than performance in my mind. So if one looks better in the context of Agones, then that's the right one, I think.

markmandel: I had to go back and look - and yep, that's how I did it :smile: 

So I say, let's go with:
`gs.ObjectMeta.Annotations = make(map[string]string, len(fam.Annoatations))`

Since we know the size, and can allocate it upfront, and we won't have to resize down the line.

victor-prodan:
Argh, git is such a nightmare for me :(
I think I need to close this PR and create a new one.

PR #312, opened 2018-07-31
Metadata propagation from fleet allocation to game server

victor-prodan:
In the fleet allocation configuration you can add optional custom metadata that will be added to the game server in the moment of allocation

Implements https://github.com/GoogleCloudPlatform/agones/issues/277
Contains fixes requested in https://github.com/GoogleCloudPlatform/agones/pull/307 (which I closed because git and I have a mutual hate relationship)


pkg/fleetallocation/controller.go
```diff
@@ -288,3 +292,25 @@ func (c *Controller) allocate(f *stablev1alpha1.Fleet) (*stablev1alpha1.GameServ
 
 	return gs, nil
 }
+
+// patch the labels and annotations of an allocated GameServer with metadata from a FleetAllocation
+func (c *Controller) patchMetadata(gs *stablev1alpha1.GameServer, fam *stablev1alpha1.FleetAllocationMeta) {
```
Kuqd: @markmandel do you think we should move this to `func(gs *stablev1alpha1.GameServer) MergeFleetMetadata(fam *stablev1alpha1.FleetAllocationMeta)` and use it 
```
if fam != nil {
		gsCopy.MergeFleetMetadata(fam)
	}
```

markmandel: Hmnnn. Interesting question.

I guess I'm more inclined to leave it - since it's functionality that is only used in this particular place. If it ends up being logic that needs to be shared, we could move it then.

Do you think it likely that we would need to have this logic be run in multiple places?

Kuqd: no but it could be a function though since `c` is not used

victor-prodan: We were talking of the game server being able to update its own metadata, right?
Could this function be reused for that?

victor-prodan: Yes, that's what I had in mind but I guess it's safer to wait until we have a definite need.

markmandel:
I almost hit approve, but I realised there is but one small thing missing!
https://github.com/GoogleCloudPlatform/agones/blob/master/docs/fleet_spec.md

The documentation for the spec needs updating as well - always be documenting! :smile: 

Probably just needs a section on the `metadata` which can be lifted from your `examples/fleetallocation.yaml`.

Although, please add a ":warning::warning::warning: This is currently a development feature and has not been released :warning::warning::warning:" above it, so that people are aware that it's not yet in a release (we remove these later).

This is just something we're doing until we get a website sorted out.

PR #315, opened 2018-08-01
Features/e2e

### LINKED ISSUE #173: Move Deployment in install script out of v1beta1
Somehow I managed to make the `Deployment` as a `extensions/v1beta1` deployment, which has since been deprecated. 

https://github.com/GoogleCloudPlatform/agones/blob/master/install/helm/agones/templates/controller.yaml

It should really be a `apps/v1` `Deployment` instead.
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

Kuqd:
This is a sample of one e2e tests for #37 . This first test create a gameserver, waits for readiness, assert some values then finally ping it (send/receive).

Everything is plugged to cloud builder and run against a real gke cluster, branch will run sequentially using a consul lock within the cluster.
Flow of the runner is : 

- port-forward the internal consul api locally
- wait for port-forward to establish
- take a lock
- install the current release via helm (agones images controller/sdk just built within the same pipeline)
- run the test as normal go test using builder service account.
- release the lock

That said there is a target if you want to run the e2e against your cluster (minikube/gke/....) from your machine.

The idea of this PR is to see if the direction is correct.

Ticket #37 has more details of remaining work to do.

Have a spin, I hope you like it.

PS: this also fixes #173 

<details><summary>e2e runner sample output</summary>

```
Already have image: e2e-runner
Fetching cluster endpoint and auth data.
kubeconfig entry generated for e2e-test-cluster.
Waiting consul port-forward to launch on 8500...
Forwarding from 127.0.0.1:8500 -> 8500
consul port-forward launched. Starting e2e tests...
Handling connection for 8500
Handling connection for 8500
Setting up lock at path: LockE2E/.lock
Attempting lock acquisition
Starting handler
installing current release
\
helm upgrade --install --recreate-pods --wait --namespace=agones-system \
--set agones.image.tag=0.4.0-9e1598b,agones.image.registry=gcr.io/agones-images,agones.image.controller.pullPolicy="Always",agones.image.sdk.alwaysPull=true \
Handling connection for 8500
Release "agones" has been upgraded. Happy Helming!
LAST DEPLOYED: Wed Aug 1 21:12:38 2018
NAMESPACE: agones-system
STATUS: DEPLOYED

RESOURCES:
==> v1/Secret
NAME TYPE DATA AGE
agones-cert Opaque 2 1h

==> v1/ServiceAccount
NAME SECRETS AGE
agones-controller 1 1h
agones-sdk 1 1h

==> v1/ClusterRole
NAME AGE
agones-controller 1h
agones-sdk 1h

==> v1/RoleBinding
NAME AGE
agones-sdk-access 1h

==> v1beta1/ValidatingWebhookConfiguration
NAME AGE
agones-validation-webhook 1h

==> v1/Pod(related)
NAME READY STATUS RESTARTS AGE
agones-controller-745568c64f-z8vwf 1/1 Running 0 5s

==> v1beta1/CustomResourceDefinition
NAME AGE
fleets.stable.agones.dev 1h
fleetallocations.stable.agones.dev 1h
gameservers.stable.agones.dev 1h
gameserversets.stable.agones.dev 1h

==> v1/ClusterRoleBinding
NAME AGE
agones-controller-access 1h

==> v1/Service
NAME TYPE CLUSTER-IP EXTERNAL-IP PORT(S) AGE
agones-controller-service ClusterIP 10.43.244.248 <none> 443/TCP 1h

==> v1/Deployment
NAME DESIRED CURRENT UP-TO-DATE AVAILABLE AGE
agones-controller 1 1 1 1 1h

==> v1beta1/MutatingWebhookConfiguration
NAME AGE
agones-mutation-webhook 1h


NOTES:
The Agones controller has been installed in the namespace agones-system.

You can watch the status by running 'kubectl --namespace agones-system get pods -o wide -w agones-controller'

Once ready you can create your first GameServer using our examples:

'kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/agones/master/examples/simple-udp/server/gameserver.yaml'

An example GameServer that makes use of the controller:

kind: GameServer
metadata:
name: "simple-udp"
spec:
portPolicy: "dynamic"
containerPort: 7654
template:
spec:
containers:
- name: simple-udp
image: gcr.io/agones-images/udp-server:0.2

Finally don't forget to explore our documentation and usage guides on how to develop and host dedicated game servers on top of Agones. :

- [Create a Game Server](https://github.com/GoogleCloudPlatform/agones/blob/master/docs/create_gameserver.md)
- [Integrating the Game Server SDK](https://github.com/GoogleCloudPlatform/agones/tree/master/sdks)
- [GameServer Health Checking](https://github.com/GoogleCloudPlatform/agones/blob/master/docs/health_checking.md)
- [Accessing Agones via the Kubernetes API](https://github.com/GoogleCloudPlatform/agones/blob/master/docs/access_api.md)
starting e2e test
go test -v agones.dev/agones/test/e2e/... \
--kubeconfig /root/.kube/config \
--gameserver-image=gcr.io/agones-images/cpp-simple-server:0.2
=== RUN TestCreateConnect
=== PAUSE TestCreateConnect
=== CONT TestCreateConnect
--- PASS: TestCreateConnect (8.31s)
PASS
ok agones.dev/agones/test/e2e	8.385s
? agones.dev/agones/test/e2e/framework	[no test files]
Child terminated, releasing lock
Cleanup succeeded
```
</details>


build/Makefile
```diff
@@ -85,6 +88,10 @@ include ./includes/$(osinclude)
 # personal includes, excluded from the git repository
 -include ./local-includes/*.mk
 
+ifdef DOCKER_RUN
```
markmandel: So I'm curious - what does this do, and why is it necessary?

build/build-image/Dockerfile
```diff
@@ -87,6 +92,19 @@ RUN wget -q https://static.rust-lang.org/rustup/archive/1.11.0/${RUST_ARCH}/rust
 RUN cargo install protobuf-codegen --vers 2.0.2
 RUN cargo install grpcio-compiler --vers 0.3.0
 
+# set up Consul.
```
markmandel: I take it we need consul locally to run e2e against local clusters?

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,134 @@
...
+	return readyGs, nil
+}
+
+// WaitForGameServerState Waits untils the gameserver reach a given state before the timeout expires
+func (f *Framework) WaitForGameServerState(gs *v1alpha1.GameServer, state v1alpha1.State,
+	timeout time.Duration) (*v1alpha1.GameServer, error) {
+	var pollErr error
+	var readyGs *v1alpha1.GameServer
+
+	err := wait.Poll(2*time.Second, timeout, func() (bool, error) {
```
markmandel: Should we `Poll` or `PollImmediate` ? If we `Poll` we always wait for 2 second, when maybe we don't need to?

test/e2e/gameserver_test.go
```diff
@@ -0,0 +1,69 @@
...
+// limitations under the License.
+
+package e2e
+
+import (
+	"fmt"
+	"testing"
+
+	"agones.dev/agones/pkg/apis/stable/v1alpha1"
+	agonesFramework "agones.dev/agones/test/e2e/framework"
```
markmandel: Any reason we can't use "framework"?

Pretty sure it's not canonical go to use camelCase here. @enocom I'll defer to you?

Looking at https://golang.org/doc/effective_go.html#package-names

"By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps"

"Another convention is that the package name is the base name of its source directory; the package in src/encoding/base64 is imported as "encoding/base64" but has name base64, not encoding_base64 and not encodingBase64."

But I couldn't find anything specific about import aliases.

test/e2e/gameserver_test.go
```diff
@@ -0,0 +1,69 @@
...
+	assert.Equal(t, readyGs.Status.State, v1alpha1.Ready)
+
+	reply, err := agonesFramework.PingGameServer("Hello World !", fmt.Sprintf("%s:%d", readyGs.Status.Address,
+		readyGs.Status.Ports[0].Port))
+
+	if err != nil {
+		t.Fatalf("Could ping GameServer: %v", err)
+	}
+
+	assert.Equal(t, reply, "ACK: Hello World !\n")
```
markmandel: :dancing_women: :dancing_women: :dancing_women: :dancing_women: :dancing_women: :dancing_women: :dancing_women:

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,134 @@
...
+	// required to use gcloud login see: https://github.com/kubernetes/client-go/issues/242
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// Framework is a testing framework
+type Framework struct {
+	KubeClient      kubernetes.Interface
+	AgonesClient    versioned.Interface
+	GameServerImage string
```
markmandel: So I'm trying to work out what this is used for? It doesn't look like it gets used?

build/e2e-image/entrypoint.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Copyright 2017 Google Inc. All Rights Reserved.
```
markmandel: 2018 :smile:

build/e2e-image/e2e.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# Copyright 2017 Google Inc. All Rights Reserved.
```
markmandel: 2018 :smile:

build/gke-test-cluster/cluster-e2e.yml
```diff
@@ -0,0 +1,44 @@
+# Copyright 2017 Google Inc. All rights reserved.
```
markmandel: 2018 :smile:

test/e2e/README.md
```diff
@@ -0,0 +1,16 @@
+# E2E Testing
```
markmandel: Should we add a section to our Development guide on this - at least we should add the new major `Make` targets?

I feel like there should be a "how to run e2e tests" section, or maybe section per deployment? Or something.

Then I can follow the guide, and make sure it all works :smile:

test/e2e/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Prerequisites:
+- a running k8s cluster and kube config. We will need to pass kube config as arguments.
+- Have kubeconfig file ready.
+
+e2e tests are written as Go test. All go test techniques apply, e.g. picking
+what to run, timeout length. Let's say I want to run all tests in "test/e2e/":
+
+```
+$ go test -v ./test/e2e/ --kubeconfig "$HOME/.kube/config" --gameserver-image=gcr.io/agones-images/cpp-simple-server:0.2
```
markmandel: Should we be running tests this way, or through `make e2e` ?

build/Makefile
```diff
@@ -102,11 +109,18 @@ build-images: build-controller-image build-agones-sdk-image
...
+test: $(ensure-build-image) test-go test-install-yaml
 
 # Run go tests
 test-go:
-	docker run --rm $(common_mounts) $(build_tag) go test -race $(agones_package)/...
+	docker run --rm $(common_mounts) $(build_tag) go test -race $(agones_package)/pkg/... \
+	$(agones_package)/sdks/...
+
+# Runs end-to-end tests on the current configured cluster
+test-e2e:
```
markmandel: We'll also need a `minikube-test-e2e` target, just because it has special mounting for .kube config and should run on host network. Should be able to lift an existing example, or I can write it if you want.

test/e2e/gameserver_test.go
enocom: Yep, @markmandel, you're right. Named imports should be all lowercase.

build/build-image/Dockerfile
Kuqd: No we don’t, I left it in case we would need it for testing but I’m happy to remove it, WDYT ?

markmandel: I'd say let's remove it - it's an added step we have then have to maintain, and keep up to date.

build/Makefile
Kuqd: I couldn't get a our `docker run` type of target get the gcloud service account of the cloud builder to access our kubernetes cluster, so I ended up using a custom 2e2  image which is a `FROM gcr.io/cloud-builders/gcloud`  and has a gcloud already configured automagically.

Since I have run without docker in the build system, I need to call the make target without doing a docker run and calling the go test command straight. **So this line make sure I don't run `ensure-build-image` if DOCKER_RUN is empty, which otherwise would fail.**

I wanted to keep the runner using our make targets (install and test-2e2) so we don't end up with too many targets. Another option would be to have special targets only for the runner and we would replicate the `go test` and the `helm install` command. 

I'm ok to shift toward a multiple makefile system if you think that is clearer, let me know.

test/e2e/framework/framework.go
Kuqd: yes you're right ! let me change that

test/e2e/gameserver_test.go
Kuqd: `framework` is used by a global variable, do you think I should not use a global variable ? I could renamed that import to `e2eframework` ?

test/e2e/framework/framework.go
Kuqd: you're right, it should be used here https://github.com/Kuqd/agones/blob/e4f921d744fc8117c97dc3efd71a0b3bdb8be247/test/e2e/gameserver_test.go#L45, do you want me to create a getter in the framework ? or should I move the creation to the framework ?

test/e2e/README.md
Kuqd: you should use `make e2e`, I will change.

test/e2e/README.md
Kuqd: yes but I think only make e2e would be sufficient no ? Do you see a need to talk about the e2e cluster ? Since people could just use their development cluster or minikube which is already well documented.

So far how to run e2e is pretty straightforward after you've setup your cluster, just run `make test-e2e`. I can add small sections about how just after talking about `Running a Test Minikube cluster` which would explain how to run against those 2 clusters and include the make target.

Do you think I should explain the e2e image and gcloud-e2e targets which are exclusively for cloud builder, I'm actually not sure that's why I ask ?

build/Makefile
Kuqd: let me see if I can get it working.

build/Makefile
markmandel: SGTM. I (think) I follow.

I can't think of a better way to do this :smile: so yeah, let's run with it for now :+1:

test/e2e/README.md
markmandel: No, I think you are right - we should document how to run the e2e tests, I don't think we need to document how they are built.

Will need to also have `make minikube-test-e2e` in the minikube section but that's about it.

:+1:

Kuqd: I've updated the build/ documentation, let me know.

build/build-image/Dockerfile
```diff
@@ -39,17 +39,23 @@ RUN wget -q https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.zip && un
 ENV PATH /usr/local/go/bin:/go/bin:/opt/google-cloud-sdk/bin:$PATH
 
 # RUN gcloud components update
-RUN gcloud components update && gcloud components install kubectl
+RUN gcloud components update
+
+# install kubectl without gcloud as we want the last version
+ENV KUBECTL_VER 1.11.0
```
markmandel: Super minor thing. Do we want 1.10? I don't think you can run 1.11 on anything yet?

test/e2e/README.md
markmandel: LGTM!

build/build-image/Dockerfile
Kuqd: Good question ! I was lazy to figured which version I need.

I updated kubectl because the version from gcloud doesn't support port-forward on stateful set and I needed it there : 

https://github.com/GoogleCloudPlatform/agones/blob/dc0e9dd3df062e58366320f44a932c1931b65d8a/build/e2e-image/entrypoint.sh#L29

it has been introduced by https://github.com/kubernetes/kubernetes/pull/59705 which seems to be 1.10, do you want me to downgrade ?

markmandel: I think so - it's super minor, bit it's nice to have our supported version line up with our Dev tooling, Just to make sure nothing odd happens.

Other than that - swash to a single commit, and I'll got approve!

PR #316, opened 2018-08-08
Implements SDK callback for GameServer updates

### LINKED ISSUE #277: Session data propagation to dedicated server

markmandel:
This implements the gRPC system to send messages up to a unidirectional
stream from the sdk-server to the SDK.

This has been implemented in the Go, C++ and REST SDK.

Closes #277


sdks/go/sdk.go
```diff
@@ -76,3 +87,43 @@ func (s *SDK) Health() error {
...
 }
+
+// WatchGameServer asynchronously calls the given GameServerCallback with the current GameServer
+// configuration when the backing GameServer configuration is updated.
+// This function can be called multiple times to add more than one GameServerCallback.
+func (s *SDK) WatchGameServer(f GameServerCallback) error {
+	s.watchCallbacks = append(s.watchCallbacks, f)
+	var err error
+
+	s.watchOnce.Do(func() {
```
markmandel: @enocom I've love to get your take on this approach. WDYT?

sdks/cpp/sdk.cc
```diff
@@ -59,6 +59,18 @@ namespace agones {
         return stub->GetGameServer(context, request, response);
     }
 
+    grpc::Status SDK::WatchGameServer(const std::function<void(stable::agones::dev::sdk::GameServer)> callback) {
```
markmandel: @Kuqd @EricFortin My C++ is terrible - I hope I haven't done anything totally silly here?

sdks/go/sdk.go
enocom: Instead of using a `sync.Once`, it might be easier to have a `RegisterCallback` method to register a new callback. Then `WatchGameServer` can run forever invoking the callbacks. 

Presently, there's a data race between the call to `append` above and reading the callbacks in the goroutine with `range` below. If you split out the two concerns, you can lock around the slice of callbacks, taking out the lock when appending or reading from the slice, and then releasing it afterwords.

Be careful about locking around the callback, though, as there is a chance the callback might never finish. You might consider passing a `content.Context` into the callback to provide a mechanism for cancelling the work and releasing the lock in failure cases.

markmandel: Reading all this, and your comments I think I've got too fancy with this. 

I think I'll take an approach that is similar to what I did with the C++ (although non-blocking). Each call to `WatchGameServer` is it's own call to the gRPC service, with it's own *singular* function callback - that will make things far simpler to run and test, rather than managing sets of function callbacks and the potential issues therein.

sdks/cpp/sdk.cc
Kuqd: looks good to me, may be @victor-prodan can confirm ?

victor-prodan: Yes, looks fine to me as well.

sdks/go/sdk.go
```diff
@@ -76,3 +83,33 @@ func (s *SDK) Health() error {
...
+// WatchGameServer asynchronously calls the given GameServerCallback with the current GameServer
+// configuration when the backing GameServer configuration is updated.
+// This function can be called multiple times to add more than one GameServerCallback.
+func (s *SDK) WatchGameServer(f GameServerCallback) error {
+	stream, err := s.client.WatchGameServer(s.ctx, &sdk.Empty{})
+	if err != nil {
+		return errors.Wrap(err, "could not watch gameserver")
+	}
+
+	go func() {
```
enocom: @markmandel Should this goroutine ever exit?

markmandel: Yes - on EOF of the gRPC stream. Basically when the stream closes on the sdk sidecar server shutdown.

I could have written the if statement much cleaner looking at it now.

PR #317, opened 2018-08-13
Fix bug: Disabled health checking not implemented

markmandel:
This commit fixes actually two bugs:

1) While we documented that you could disabled health checking we never
actually implemented it! This fixes that. Thankfully, the work that has been
done to retrieve `GameServer` details through the SDK makes this relatively
easy.

2) SDK Server sidecar never had the necessary RBAC permissions to send events
to the `GameServer` CRD. This is now fixed as well.

### DIFF
--- a/pkg/gameservers/sdkserver_test.go
+++ b/pkg/gameservers/sdkserver_test.go
@@ -15,10 +15,9 @@
 package gameservers
 
 import (
-	"net/http"
+		"net/http"
 	"sync"
 	"testing"
-
 	"time"
 
 	"agones.dev/agones/pkg/apis/stable/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/gameservers/sdkserver_test.go
```diff
@@ -157,7 +171,7 @@ func TestSidecarHealthLastUpdated(t *testing.T) {
 
 	sc, err := defaultSidecar(m)
 	assert.Nil(t, err)
-	sc.healthDisabled = false
+	sc.health = v1alpha1.Health{Disabled: true}
```
EricFortin: Doesn't it change the intent of the test?

markmandel: Fixed! Incoming shortly.

PR #321, opened 2018-08-14
Add Minikube instructions

slartibaartfast:
Add Minikube instructions to edit_first_game_server.md

### DIFF
--- a/docs/edit_first_game_server.md
+++ b/docs/edit_first_game_server.md
@@ -1,15 +1,20 @@
 # Getting Started
-The following guide is for developers without Docker or Kubernetes experience, that want to use the simple-udp example as a starting point for a custom game server. Since this guide is for Google Kubernetes Engine, we welcome a Pull Request to expand this to include Minikube as well.
+The following guide is for developers without Docker or Kubernetes experience, that want to use the simple-udp example as a starting point for a custom game server. This guide addresses Google Kubernetes Engine and Minikube.  We would welcome a Pull Request to expand this to include other platforms as well.
 
 ## Prerequisites
 
 1. Downland and install Golang from https://golang.org/dl/.
 2. Install Docker from https://www.docker.com/get-docker.
-3. Follow the install instructions (if you haven't already) at [Install and configure Agones on Kubernetes](../install/README.md). At least, you should have "Setting up a Google Kubernetes Engine (GKE) cluster", "Enabling creation of RBAC resources" and "Installing Agones" done.
+3. Install Agones on GKE or Minikube.  
+
+To install on GKE, follow the install instructions (if you haven't already) at
+[Setting up a Google Kubernetes Engine (GKE) cluster](../install/README.md#setting-up-a-google-kubernetes-engine-gke-cluster). Also complete the "Enabling creation of RBAC resources" and "Installing Agones" sets of instructions on the same page.
+
+To install locally on Minikube, read [Setting up a Minikube cluster](../install/README.md#setting-up-a-minikube-cluster).  Also complete the "Enabling creation of RBAC resources" and "Installing Agones" sets of instructions on the same page.
 
 ## Modify the code and push another new image
 
-### Modify the source code
+### Modify the simple-udp example source source code
 Modify the main.go file. For example:
 
 Change main.go line 92:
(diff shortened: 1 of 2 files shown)


slartibaartfast:
Signed the thing

docs/edit_first_game_server.md
```diff
@@ -1,11 +1,15 @@
...
 ## Prerequisites
 
 1. Downland and install Golang from https://golang.org/dl/.
 2. Install Docker from https://www.docker.com/get-docker.
-3. Follow the install instructions (if you haven't already) at [Install and configure Agones on Kubernetes](../install/README.md). At least, you should have "Setting up a Google Kubernetes Engine (GKE) cluster", "Enabling creation of RBAC resources" and "Installing Agones" done.
+3. Install Agones on GKE or Minikube.  
+
+To install on GKE, follow the install instructions (if you haven't already) at [Install and configure Agones on Kubernetes](../install/README.md). At least, you should have "Setting up a Google Kubernetes Engine (GKE) cluster", "Enabling creation of RBAC resources" and "Installing Agones" done.
+
+To install locally on Minikube, read [Developing, Testing, and Building Agones](../build/README.md) making sure to follow the instructions in the [Running A Test Minikube cluster](../build/README.md#running-a-test-minikube-cluster) section.
```
markmandel: Should we link to installing on Minikube instead?
https://github.com/GoogleCloudPlatform/agones/blob/master/install/README.md#setting-up-a-minikube-cluster

docs/edit_first_game_server.md
```diff
@@ -1,11 +1,15 @@
...
+The following guide is for developers without Docker or Kubernetes experience, that want to use the simple-udp example as a starting point for a custom game server. This guide addresses Google Kubernetes Engine and Minikube.  We would welcome a Pull Request to expand this to include other platforms as well.
 
 ## Prerequisites
 
 1. Downland and install Golang from https://golang.org/dl/.
 2. Install Docker from https://www.docker.com/get-docker.
-3. Follow the install instructions (if you haven't already) at [Install and configure Agones on Kubernetes](../install/README.md). At least, you should have "Setting up a Google Kubernetes Engine (GKE) cluster", "Enabling creation of RBAC resources" and "Installing Agones" done.
+3. Install Agones on GKE or Minikube.  
+
+To install on GKE, follow the install instructions (if you haven't already) at [Install and configure Agones on Kubernetes](../install/README.md). At least, you should have "Setting up a Google Kubernetes Engine (GKE) cluster", "Enabling creation of RBAC resources" and "Installing Agones" done.
```
markmandel: We can link to the GKE section directly:
https://github.com/GoogleCloudPlatform/agones/blob/master/install/README.md#setting-up-a-google-kubernetes-engine-gke-cluster

WDYT?

docs/edit_first_game_server.md
```diff
@@ -33,20 +37,31 @@ go get agones.dev/agones/pkg/sdk
...
 ### Create a new docker image
 ```bash
 docker build -t gcr.io/[PROJECT_ID]/agones-udp-server:0.2 .
 ```
 Note that: you can change the image name "agones-udp-server" to something else.
 
-### Push the image to GCP Registry
+### If using GKE, push the image to GCP Registry
 ```bash
 docker push gcr.io/[PROJECT_ID]/agones-udp-server:0.2
```
markmandel: Wondering if we should link to:
https://cloud.google.com/container-registry/docs/advanced-authentication

To point people in the right direction to authenticate against the GCR?

docs/edit_first_game_server.md
```diff
@@ -33,20 +37,31 @@ go get agones.dev/agones/pkg/sdk
...
 ```
 
-### Modify gameserver.yaml
+### If using Minikube, change directory to agones/build and push the image into Minikube
+```bash
```
markmandel: Rather than point people to development tools, we can just show them the command line trick:
(just concerned with Agones users relying on development tooling - in case it changes).

`docker save [PROJECT_ID]/agones-udp-server:0.2 | (eval $(minikube docker-env) && docker load)`

WDYT?

docs/edit_first_game_server.md
```diff
@@ -56,14 +71,32 @@ Modify the following line from gameserver.yaml to use the new configuration.
...
-### Deploy Server
+### For Minikube local development, modify gameserver.yaml
+Modify the following line from gameserver.yaml to use the new configuration.
+
+```
+    spec:
+      containers:
+      - name: agones-simple-udp
+        image: gcr.io/[PROJECT_ID]/agones-udp-server:0.2
+        imagePullPolicy: Never
```
markmandel: Do we need `imagePullPolicy: Never` ? it is the default.

docs/edit_first_game_server.md
slartibaartfast: Yes that makes sense.  It seems less complex.  I tried both ways and eval $(minikube docker-env) etc... worked fine provided the make tools weren't used.

docs/edit_first_game_server.md
slartibaartfast: I found I needed it or I would get an error as it failed to pull the image.  I'll try without it when not using make.

docs/edit_first_game_server.md
slartibaartfast: Do you mean add content along the lines of "More information is available at <link> "?

docs/edit_first_game_server.md
slartibaartfast: Yes, that would be more consistent.

docs/edit_first_game_server.md
slartibaartfast: Pros seem to be that it is less complex and requires less reading time.  
Cons are that the reader is left to discover build tools on their own, and they might need information about having the project on their gopath.

docs/edit_first_game_server.md
slartibaartfast: Committed some text adding this link.

docs/edit_first_game_server.md
slartibaartfast: This change has been made.

docs/edit_first_game_server.md
markmandel: So I would argue that if they are developing with Agones - they shouldn't need the build tools - they should install and use Agones as per normal.

I'm very hesitant to have the streams of using Agones and Developing Agones mix - they have very different targets, and it could get super messy in the long term.

docs/edit_first_game_server.md
markmandel: Perfect :+1:

docs/edit_first_game_server.md
markmandel: Hmn. That's weird. Actually, I lie - `Never` is not the default `IfNotPresent` - but if you have transferred the image to Minikube, it shouldn't pull it then.

docs/edit_first_game_server.md
markmandel: I'm thinking we shouldn't be pointing people towards the development tooling - it's very cross purpose, and developers shouldn't be coupled to our own development tooling.

docs/edit_first_game_server.md
slartibaartfast: imagePullPolicy: Never has been removed.  Now both GKE and Minikube share the same paragraph.

docs/edit_first_game_server.md
slartibaartfast: Per your suggestion, this now links to https://github.com/GoogleCloudPlatform/agones/blob/master/install/README.md#setting-up-a-minikube-cluster

docs/edit_first_game_server.md
slartibaartfast: The doc now uses docker save [PROJECT_ID]/agones-udp-server:0.2 | (eval $(minikube docker-env) && docker load) instead of make.

markmandel:
LGTM! Can we rebase down to a single commit, and I'll give it the stamp of approval and merge!

THANKS!

slartibaartfast:
It's cleaned up and rebased to a single commit.  Next time I will be more careful squashing :)

PR #323, opened 2018-08-22
Implement SDK SetLabel and SetAnnotation functionality

### LINKED ISSUE #279: Propagating game server process events to Agones system
Moving relevant items from #277 so they can be tackled separately

> Another cool feature would be complting fleet allocation (through setting it status to Ready or smth like this) only after dedicated server loaded level and is fully ready to accept players' connections. Agones SDK could be extended with Loaded() method for this.

> we were thinking that the game server should also be able answer back to let know that it is ready, let's say it will load a new map and we want to receive player only when the map is loaded.

I'm letting this peculate in my head a little before driving towards a opinion on a solution, but some initial thoughts and concerns:

1. This could possibly be a job for a SDK integrate that lets you set game specific Status key-value pairs. Then you could use the K8s watch api to look for these events. Then we aren't tied to just one event (like "map loaded", etc), and it can be specifically tailored to just about any game. the downside here, would be this pushes some work up to the layer that integrated into Agones for the sake of flexibility. But it would be super easy to do.
2. It's possible that we may have another replicated GameServer control plane other than Fleets at some point in the future (not sure what this would look like , but maybe something more MMO focused?) - I'm a little hesitant to tie the SDK to Fleet specific things (which would be `Allocated` state), at least right now for this reason, but maybe I'm being overtly cautious. Entirely possible -- maybe a solution here is to work out a way to have an optional blocking callback on any status change, although actually not 100% on how to do this technologically. I also get scared of "what happens if the callback fails somewhere".
3. I'm hesitant to block on a gameserver process on letting the GameServer move to Allocated, just because the MutationWebhook means there is a hard 30s timeout on the work that it does -- and if we hit that, weird stuff could potentially start to happen. Hence I would lead more towards option No.1. 

Outside of "wait until a map is loaded before letting players join" - are there any other scenarios that you would want to wait for certain things to happen? Exploring this space a bit more may open up more flexible solutions.

markmandel:
This implements new functions in the SDK:

- SetLabel(key, value) - that lets you set a label on the backing `GameServer`
- SetAnnotation(key, value) - that lets you set an annotation on the backing
`GameServer`

All keys are prefixed with "stable.agones.dev/sdk-" to maintain isolation.

Closes #279


test/e2e/framework/framework.go
```diff
@@ -70,6 +71,8 @@ func (f *Framework) CreateGameServerAndWaitUntilReady(ns string, gs *v1alpha1.Ga
 		return nil, fmt.Errorf("creating %v GameServer instances failed (%v): %v", gs.Spec, gs.Name, err)
 	}
 
+	logrus.WithField("name", newGs.ObjectMeta.Name).Info("GameServer created, waiting for Ready")
```
markmandel: @Kuqd I added some logging here, to make it easier to track down what the name of the GameServers are, if things go wrong.

test/e2e/main_test.go
```diff
@@ -31,8 +31,8 @@ func TestMain(m *testing.M) {
 	usr, _ := user.Current()
 	kubeconfig := flag.String("kubeconfig", filepath.Join(usr.HomeDir, "/.kube/config"),
 		"kube config path, e.g. $HOME/.kube/config")
-	gsimage := flag.String("gameserver-image", "gcr.io/agones-images/cpp-simple-server:0.2",
-		"gameserver image to use for those tests, gcr.io/agones-images/cpp-simple-server:0.2")
+	gsimage := flag.String("gameserver-image", "gcr.io/agones-images/udp-server:0.3",
```
markmandel: @Kuqd I updated this to make in IDE testing easier - and to match the `Makefile` - do you see any issue?

Kuqd: good catch it was a mistake from me.

test/e2e/framework/framework.go
Kuqd: make sense !

test/e2e/gameserver_test.go
```diff
@@ -67,3 +52,87 @@ func TestCreateConnect(t *testing.T) {
...
+func TestSDKSetAnnotation(t *testing.T) {
+	t.Parallel()
+	gs := defaultGameServer()
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(defaultNs, gs)
+	if err != nil {
+		t.Fatalf("Could not get a GameServer ready: %v", err)
+	}
+
+	assert.Equal(t, readyGs.Status.State, v1alpha1.Ready)
+	reply, err := e2eframework.PingGameServer("ANNOTATION", fmt.Sprintf("%s:%d", readyGs.Status.Address,
```
Kuqd: This feels like we're going to need a better client for interacting with the gameserver at some point. If we could received and send a variable for instance we would be able to assert that the right value is stored.

However I think for now this would do the trick !

test/e2e/gameserver_test.go
```diff
@@ -67,3 +52,87 @@ func TestCreateConnect(t *testing.T) {
...
+		}
+		return gs.ObjectMeta.Annotations != nil, nil
+	})
+
+	assert.Nil(t, err)
+	assert.NotEmpty(t, gs.ObjectMeta.Annotations["stable.agones.dev/sdk-timestamp"])
+}
+
+func defaultGameServer() *v1alpha1.GameServer {
+	return &v1alpha1.GameServer{ObjectMeta: metav1.ObjectMeta{GenerateName: "udp-server", Namespace: defaultNs},
```
Kuqd: I meant to do that in my previous PR :), totally forgot about it.

PR #326, opened 2018-08-24
Move Status > Address & Ports population to `Creating` state processing

### LINKED ISSUE #293: Move Status > Address & Status > Ports population to `Creating` state processing
Right now the IP and Ports of the `GameServer` is only populated when the `GameServer` is set to `Ready`.

Since this information may be needed by the `GameServer` process before the game server process actually wants to mark itself as `Ready`, we should make this available earlier. There doesn't seem to be a technical reason to wait until `Ready` - other than we need a `Pod` to know what the IP is - but we create one at the `Creating` state processing, so it shouldn't be an issue, so we can move the processing to there, and make it available at that point and time.

## Implementation
https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameservers/controller.go

Move the code that populates the `Status.Address` and `Status.Ports` from `syncGameServerRequestReadyState` to `syncGameServerCreatingState`

markmandel:
Previously the IP and Ports of the `GameServer` were populated only when the `GameServer` is set to Ready.

This information may be needed by the `GameServer` process before the process actually wants to mark itself as Ready.

Therefore, port and address population has been moved to the `Creating` state processing, as that is where the Pod is created, and therefore we have both address and port information.

`GameServer` events now look like this:

```
Events:
  Type    Reason          Age   From                   Message
  ----    ------          ----  ----                   -------
  Normal  PortAllocation  17s   gameserver-controller  Port allocated
  Normal  Creating        16s   gameserver-controller  Pod simple-udp-fcf44 created
  Normal  Starting        16s   gameserver-controller  Address and Port populated
  Normal  Ready           15s   gameserver-controller  SDK.Ready() executed
```

Closes #293


victor-prodan:
Looks ok to me but I'm no go guru, to put it mildly....

But this feature îs definitely helpful, am glad to see it.

steven-supersolid:
I spent about 30 minutes looking this morning and couldn't see anything obviously wrong but very new to this. Will also see if I can check out a couple of PRs over the weekend

PR #328, opened 2018-08-26
Ability to pass `GameServer` yaml/json to local sdk server

### LINKED ISSUE #296: When running the SDK sidecar in local mode, be able to specify the backing `GameServer` configuration
I am testing locally the game server with the sidecar launched independently in --local mode.
The thing is that in this case, the name of the port created by the sidecar ("default") does not match the name of the port from the yaml file used with a real deployment.
I would like to be able at least to change the "default" port name.

markmandel:
To be able to work locally, you need to be able to specify your local `GameServer` configuration, as it likely will have application specific configuration in it -- or maybe you want to specify what state it's in.

This commit allow you to specify the local resource as either yaml/json through a `-f` or `--file` flag.

Closes #296


markmandel:
@victor-prodan the comments you made are tied to the commit, so you have to go back to the commit to reply :smile: github shenanigans.

Totally fair call about 4096 (you also made me laugh out loud :rofl: )! At the very least, it should be a much clearer explanation of what that value is! I have made that change now.

> In this case, I believe that the writer of the function should have also exposed a default buffer size as a constant.

I agree on that point as well!

PR #337, opened 2018-09-01
Add Troubleshooting section to Build doc

victor-prodan:


build/README.md
```diff
@@ -375,3 +379,20 @@ Use TAG to specify the image to transfer into minikube
 ## Dependencies
 
 This project uses the [dep](https://github.com/golang/dep) as a dependency manager. You can see the list of dependencies [here](https://github.com/GoogleCloudPlatform/agones/blob/master/Gopkg.toml).
+
+## Troubleshooting
+
+Frequent issues and possible solutions
+
+#### $GOPATH/$GOROOT error when building
```
markmandel: Do we want to add something specific about how it should be mounted on `/c/` -- or that to check if the mount is occuring correctly (with some troubleshooting steps?)

@fooock did you have some suggestions here?

fooock: I would indicate that if the project is on a disk other than `c`, it would be necessary to mount the disk on which the project is located. 

Docker for windows automatically assembles disk `c`, but if our project is in `d`, we will have to mount disk `d`.

victor-prodan:
You can see the formatted version here: https://github.com/victor-prodan/agones/tree/master/build
(I hope it's public ?!). Let me know if I should change wording/content.

PR #339, opened 2018-09-03
Fix yaml file paths

oskoi:


markmandel:
This should probably actually point to the latest full release, rather than at master.

Master is always going to be volatile.

If you also want to make that change as well, that would also be appreciated. That being said, 0.4.0 will come out next Tuesday, so that could come next week.

oskoi:
I signed cla.

@markmandel i rebased branch onto latest release. Thanks for your help!

Issue #334, opened 2018-08-31
Horizontal Fleet Autoscaling


victor-prodan:
**What**

Horizontal Fleet Autoscaling means modifying in real time the number of replicas (game servers) of a fleet, according to demand (i.e. numbers of servers allocated) and user-defined settings (i.e. min/max buffer size).

Cluster autoscaling (adding/removing nodes) is a separate issue (#145) which, given the multi-tenancy nature of Kubernetes (a cluster may host multiple fleets), must be addressed at a central, higher layer.

**Why**

Currently the fleet specification includes a static number of replicas, which include game servers in all possible states - pending, ready, allocated, terminating etc. This means on one hand that you have a limited capacity that can be used and on the other hand that you pay for instances that are not used. Horizontal Fleet Autoscaling addresses these two issues by scaling up and down the number of replicas according to number of already allocated replicas or, in other words, scaling the fleet according to demand.

For example, instead of saying that the fleet must have 50 replicas, you say that the fleet may have between 5 and 100 total replicas, and keep a buffer of 5 ready game instances. The fleet will start with 5 game instances (the minimum number) and when one is allocated, the autoscaler will automatically add a new replica to the fleet to refill the "ready" buffer. 

This HFA system be useful in itself, even without cluster autoscaling, in the following scenarios:

- you are using a cluster to host multiple server versions for development or QA purposes (i.e. daily build produced automatically by a build farm); you don't know in advance how many developers will use each version, and their "popularity" will vary in time as well; if you set a low replica number you risk to block the developers if many of them will want to use the same version (i.e. on Mondays almost everyone will sync), and if you set a high low replica number you will reserve cluster capacity you will not actually use and this will lead to higher costs; 
- you have a stable (or limited - i.e. closed alpha, beta etc) player base but many game modes and you can't predict with accuracy the popularity of each mode; however, it is sure that one user will play only one game mode at a time; in this case, you can host all the fleets for the game modes in the same cluster (sized to cover the player limit) and use HFA to adjust them dynamically according to user preference; 

**How**

I think that this behavior may be implemented as part of the fleet controller, because:
- the fleet autoscaler is a fleet component as there is a one-to-one relation between fleet and autoscaler; each autoscaler is working on a specific fleet, it is created and destroyed together with the fleet
- rolling updates  are also handled in the fleet controller
- autoscaling settings can be part of the Fleet specification, it would make sense; each fleet must have a different autoscaling settins
[...]

Issue #329, opened 2018-08-27
Move to using CRD Subresources for all Agones CRDs


markmandel:
# Design
We would want to generate the client code and write the kubectl configuration for Subresources.

I'm not 100% positive if this is enabled by default / available on cloud providers in Kubernetes 1.10. We _may_ have to wait until Kubernetes 1.11 is available across the board.

That being said, this looks to break down to two major things:

### Status Subresource
This doesn't give us any extra functionality, but does give us more control RBAC control over what can be changed by what part of the system. For example #150 will only be possible once this is in effect. 

There are also some potential minor performance optimisations - but I think this will be minor. This is more of a security++ operation.

### Scale on Fleets
This may be the easier one to start with - but the endpoint being that it would enable `kubectl scale Fleet simple-udp --replicas=10` which would be super nice, and make custom autoscalers much easier.

As an implementation detail, this may also drive us to implement to scale on the backing `GameServerSet` as well.

## Research
- https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#subresources
- https://blog.openshift.com/kubernetes-custom-resources-grow-up-in-v1-10/
- https://kubernetes.io/docs/setup/release/notes/#sig-api-machinery

victor-prodan:
Subresource implementation example found here: https://github.com/kubernetes/sample-controller

markmandel:
We can do this now that #447 is merged!

aLekSer:
Started adding new Subresources into helm templates and it is working for Fleet for instance:
```
kubectl scale --replicas=15 fleet/simple-udp
```
I followed the way which is documented under `Subresources` section here: https://v1-11.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#advanced-topics

aLekSer:
Regarding adding Scale subresource: it makes Fleet able to be used with Horizontal Pod Autoscaling. And we can use metrics which was previously exported to stackdriver to control the scaling. This is just one of scale subresource gains.
https://cloud.google.com/blog/products/gcp/beyond-cpu-horizontal-pod-autoscaling-comes-to-google-kubernetes-engine
```
Allows systems like HorizontalPodAutoscaler and PodDisruptionBudget interact with your new resource
```
https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#api-server-aggregation

markmandel:
[...]

PR #340, opened 2018-09-04
FleetAutoScaler v0

victor-prodan:
Implements #334 
v1 adds policy-based autoscaling.
:warning:  Ready for review. :warning:


markmandel:
Wasn't sure - is this ready for review, or is more work coming? Figured I would ask.

victor-prodan:
It's ready for review, yes.

Pe 5 sept. 2018 23:51, "Mark Mandel" <notifications@github.com> a scris:

Wasn't sure - is this ready for review, or is more work coming? Figured I
would ask.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/GoogleCloudPlatform/agones/pull/340#issuecomment-418877260>,
or mute the thread

pkg/apis/stable/v1alpha1/fleetautoscaler.go
```diff
@@ -0,0 +1,112 @@
...
+	// Must be bigger than 0
+	// Note: by "ready" we understand in this case "non-allocated"; this is done to ensure robustness 
+	//       and computation stability in different edge case (fleet just created, not enough capacity in the cluster etc)
+	BufferSize int32 `json:"bufferSize"`
+}
+
+// FleetAutoScalerStatus needs doc
+type FleetAutoScalerStatus struct {
+	// Fleet is the copy of fleet we're attached to. It is updated each tick
+	Fleet *Fleet
```
markmandel: I'm not a fan of this - it's a lot of redundant information to be updating all the time.

It's also not necessary, since the Fleet is in local cache as well.

Looking at the HorizontalPodAutoscaler - the `desiredReplicas` and `currentReplicas` would likely be enough?
What extra information do we need here?

(As I write this, I realise I've got the full game server in the allocation - I'm now also questioning that decision!)

victor-prodan: Yes I agree we don't need the Fleet object, I just didn't know what to put in its stead.
How about a subresource compatible with 'kubectl scale'.... that could be used to scale bufferSize?

pkg/fleetautoscalers/controller.go
```diff
@@ -0,0 +1,347 @@
...
+	"encoding/json"
+	"fmt"
+
+	"agones.dev/agones/pkg/apis/stable"
+	stablev1alpha1 "agones.dev/agones/pkg/apis/stable/v1alpha1"
+	"agones.dev/agones/pkg/client/clientset/versioned"
+	getterv1alpha1 "agones.dev/agones/pkg/client/clientset/versioned/typed/stable/v1alpha1"
+	"agones.dev/agones/pkg/client/informers/externalversions"
+	listerv1alpha1 "agones.dev/agones/pkg/client/listers/stable/v1alpha1"
+	//	"agones.dev/agones/pkg/fleets"
```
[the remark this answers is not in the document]
victor-prodan: Oups!

examples/fleetautoscaler.yaml
```diff
@@ -0,0 +1,39 @@
...
+  # as this FleetAutoScaler
+  fleetName: fleet-example
+  # Size of a buffer of "ready" game server instances
+  # The FleetAutoScaler will scale the fleet up and down trying to maintain this buffer, as
+  # instances are being allocated or terminated
+  bufferSize: 5
+  # minimum fleet size to be set by this FleetAutoScaler. 
+  # if not specified, the actual minimum fleet size will be bufferSize
+  minReplicas: 10
+  # maximum fleet size that can be set by this FleetAutoScaler
```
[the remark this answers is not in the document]
victor-prodan: I can't think of a single case where this would be useful... but what would be a good default value? or should we make this value required?

pkg/apis/stable/v1alpha1/fleetautoscaler.go
markmandel: I think `desiredReplicas` and `currentReplicas` would be good to put in the status, it gives enough information to those need to look into what is going on with the Autoscaler (and we have precedent with the HorizontalPodAutoscaler)

We should probably also have some events - but wanted to take this for a spin first before making suggestions.

Once we have #329 going, then there's an interesting question on how 'scale' should work wiith autoscalers -- let's definitely keep that in mind!

examples/fleetautoscaler.yaml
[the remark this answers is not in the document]
markmandel: My thought was more we should make this value required, so that there is always _some_ top value.

Infinity scares me a little? :fearful:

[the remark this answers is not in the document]
victor-prodan: done

pkg/apis/stable/v1alpha1/fleetautoscaler.go
victor-prodan: done

pkg/fleetautoscalers/controller.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	// Attach to the fleet
+	// When a Fleet is deleted, the FleetAutoScaler should go with it
+	ref := metav1.NewControllerRef(fleet, stablev1alpha1.SchemeGroupVersion.WithKind("Fleet"))
+	fas.ObjectMeta.OwnerReferences = append(fas.ObjectMeta.OwnerReferences, *ref)
+
+	// Q: can we do this nicer?
+	c.recorder.Eventf(fleet, corev1.EventTypeNormal, string("CreatingFleetAutoScaler"), "Attached FleetAutoScaler %s to fleet %s", fas.ObjectMeta.Name, fas.Spec.FleetName)
+	c.recorder.Eventf(fas, corev1.EventTypeNormal, string("CreatingFleetAutoScaler"), "Attached FleetAutoScaler %s to fleet %s", fas.ObjectMeta.Name, fas.Spec.FleetName)
```
markmandel: Is there a reason we are doing this twice?

Also, I don't think "CreatingFleetAutoScaler" needs to be cast to string, it already is a string?

victor-prodan: I don't want to do it twice, hence the Q.
What I wanted is to push the event in both the fleet and fleetautoscaler queues.

Agreed with string conversion, don't know who put it there :D

markmandel: You know, I totally missed that you are pushing to both :man_facepalming: yeah. There is no special way to do that. You did the right thing.

victor-prodan: Added a comment explaining the reason for duplication

victor-prodan:
@markmandel I'm just fooling around waiting for your feedback. I'm not doing anything relevant....

markmandel:
@victor-prodan no worries - sorry, too many meetings in Japan. Will find time tomorrow!

Also trying to fix #354

victor-prodan:
@markmandel No worries, I can't merge until #354 is fixed anyway. No pressure :stuck_out_tongue:

pkg/apis/stable/v1alpha1/fleetautoscaler.go
```diff
@@ -0,0 +1,127 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []FleetAutoScaler `json:"items"`
+}
+
+// FleetAutoScalerSpec is the spec for a Fleet Scaler
+type FleetAutoScalerSpec struct {
+	FleetName string `json:"fleetName"`
+	// MaxReplicas is the maximum amount of replicas that the fleet may have.
+	// It must be bigger than both MinReplicas and BufferSize
```
EricFortin: I think it should be greater than the sum of `MinReplicas` and `BufferSize`

pkg/apis/stable/v1alpha1/fleetautoscaler.go
```diff
@@ -0,0 +1,127 @@
...
+			Message: "fleetName cannot be updated",
+		})
+	}
+
+	return new.ValidateAutoScalingSettings(causes)
+}
+
+//ValidateAutoScalingSettings validates the FleetAutoScaler scaling settings
+func (fas *FleetAutoScaler) ValidateAutoScalingSettings(causes []metav1.StatusCause) []metav1.StatusCause {
+	if fas.Spec.MinReplicas != 0 && fas.Spec.MinReplicas > fas.Spec.MaxReplicas {
```
EricFortin: I think the first check is useless as even if 0, `min` should be lower than `max`.

pkg/apis/stable/v1alpha1/fleetautoscaler.go
```diff
@@ -0,0 +1,127 @@
...
+		})
+	}
+	if fas.Spec.BufferSize <= 0 {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Field:   "bufferSize",
+			Message: "bufferSize must be bigger than 0",
+		})
+	}
+	if fas.Spec.MaxReplicas < fas.Spec.BufferSize {
```
EricFortin: `MaxReplicas` shouldn't be lower than the sum of `MinReplicas` and `BufferSize`.

pkg/fleetautoscalers/controller.go
```diff
@@ -0,0 +1,354 @@
...
+		AddFunc: c.workerqueue.Enqueue,
+		UpdateFunc: func(_, newObj interface{}) {
+			c.workerqueue.Enqueue(newObj)
+		},
+	})
+
+	return c
+}
+
+// Run runs this controller. This controller doesn't (currently)
```
EricFortin: It looks like there is a worker queue, is the comment still valid?

pkg/gameservers/controller.go
```diff
@@ -498,6 +498,27 @@ func (c *Controller) addGameServerHealthCheck(gs *v1alpha1.GameServer, pod *core
 	}
 }
 
+// applyGameServerAddressAndPortIfNeeded calls applyGameServerAddressAndPort if
+// the address and port is not already applied in game server status
+// NOTE THIS IS NOT PART OF AUTO SCALER
+func (c *Controller) applyGameServerAddressAndPortIfNeeded(gs *v1alpha1.GameServer) (*v1alpha1.GameServer, bool, error) {
```
EricFortin: Is this from Mark's fix? If not, might be better to remove this and rebase.

EricFortin:
Code looks OK. I am not convinced with the name though. How do we call the next autoscaler that scales differently? Is BufferBasedFleetAutoscaler too long? What is k8s convention here?

pkg/apis/stable/v1alpha1/fleetautoscaler.go
victor-prodan: right!

pkg/gameservers/controller.go
victor-prodan: `// NOTE THIS IS NOT PART OF AUTO SCALER` :smile:

pkg/apis/stable/v1alpha1/fleetautoscaler.go
victor-prodan: Can you share some reasoning behind this?
I don't see it.... :thinking:

victor-prodan:
> Code looks OK. I am not convinced with the name though. How do we call the next autoscaler that scales differently? Is BufferBasedFleetAutoscaler too long? What is k8s convention here?

I was thinking of adding some multiple Strategy support in this autoscaler instead of creating new one...

victor-prodan:
@EricFortin thanks for your feedback! I implemented all changes expect the maxreplicas relation to buffersize and minreplicas, whose reason I don't understand :(

victor-prodan:
> Code looks OK. I am not convinced with the name though. How do we call the next autoscaler that scales differently? Is BufferBasedFleetAutoscaler too long? What is k8s convention here?

@markmandel WDYT?

pkg/gameservers/controller.go
Kuqd: ready to rebase on master

victor-prodan:
@markmandel I had a discussion with Eric about future extensibility. A solution would be to allow the user to define a autoscaling function for maximum flexibility. This can be achieved using REST - use a POST to pass the current fleet status in json format and get back the desired replicas.

One other idea would be to host this user function in a second sidecar

Of course, we could support both.

To prepare the road to this, I could put add support for multiple strategies from this PR, with StaticBuffer being the only one for now.

WDYT?

Kuqd:
> One other idea would be to host this user function in a second sidecar

Why would we want to add scaling logic to gameserver pod ?

victor-prodan:
> > One other idea would be to host this user function in a second sidecar
> 
> Why would we want to add scaling logic to gameserver pod ? It seems that the cardinality is way different, there can be a huge amount of gameservers but only one autoscaler ?

Right, I was confusing the entities. It would be an independent pod (as the MMF pod is for open-match) and not a sidecar...

Kuqd:
Seems that GRPC is a great fit here

examples/fleetautoscaler.yaml
```diff
@@ -0,0 +1,39 @@
...
+  # https://v1-9.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#objectmeta-v1-meta
+  Name: fleet-autoscaler-example
+spec:
+  # The name of the fleet to attach to and control. Must be an existing Fleet in the same namespace
+  # as this FleetAutoScaler
+  fleetName: fleet-example
+  # Size of a buffer of "ready" game server instances
+  # The FleetAutoScaler will scale the fleet up and down trying to maintain this buffer, as
+  # instances are being allocated or terminated
+  bufferSize: 5
```
markmandel: Looking at our original [design ticket](https://github.com/GoogleCloudPlatform/agones/issues/334#issuecomment-417785367) we wanted to have `buffer` config with a `strategy` definition (not got huge feeling on `replicaBuffer vs bufferSize vs `size if it comes under the `buffer` key)

If we do that, then that should resolve @EricFortin 's issues re: naming of the CRD - since the original design for this CRD was meant to be extensible, so we didn't have a CRD+controller for all fleet autoscaler strategies.

WDYT? Then once this is complete, we could add a `strategy: custom` strategy after the fact -- which to be honest, I think needs it's own design ticket outside of this implementation and #334 (we could start that now, there's lots of prior art there too we could leverage).

How does everyone feel about that?

EricFortin: Sounds good to me.

victor-prodan:
@markmandel, @ericfortin I won't have the next few days to work on this but when I will I want to finish it up. Îs there anything else to do? How do You are integration with doc for instance?

markmandel:
@victor-prodan for docs, I would have an entry under 'Quickstart', `Reference` docs - then make sure there is a link to the full example (which I think you already have), and make sure there is a instance of a HFA for xonotic and simple-udp. How's that sound?

victor-prodan:
@markmandel @EricFortin Have a look at my last commit please. Am I on the right track?

markmandel:
This looks perfect - one comment, I don't think you need the `bufferType` so you have absolute vs relative.

(for some reason the airport lounge wifi has blocked godoc, so I'll link to the source instead) 
You can use [`GetValueFromIntOrPercent`](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/intstr/intstr.go#L153) to pull out either percentage or absolute values, without having to specify which one you want in another config value.

You can see it in effect in [Deployments Spec Strategy](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#deploymentstrategy-v1-apps), and in our code base, in [Fleet rolling updates as well](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/fleets/controller.go#L363)

:+1:

victor-prodan:
@markmandel : thanks for the hint!

victor-prodan:
@markmandel @EricFortin it's ready for review now
One thing I am not sure about... the name. What do you feel about changing FleetAutoScaler->FleetAutoscaler ?

markmandel:
> One thing I am not sure about... the name. How do you feel about changing FleetAutoScaler->FleetAutoscaler ?

https://en.wikipedia.org/wiki/Autoscaling -- Autoscaling is one word, so I'd be fine with Autoscaler.

`HorizontalPodAutoscaler` is also used in Kubernetes. Sounds like a good change to me :+1:

markmandel:
Oh that's a really fun one. Maybe a good reason to delete the helm chart after we are done? WDYT @Kuqd 

Cleaning it out now.

pkg/fleetautoscalers/controller.go
```diff
@@ -0,0 +1,363 @@
...
+			Message: "FleetAutoscaler update is invalid",
+			Reason:  metav1.StatusReasonInvalid,
+			Details: &details,
+		}
+	}
+
+	return review, nil
+}
+
+// syncFleetAutoscaler needs doc
```
markmandel: Mega minor nit ;) also amusing :smile:

test/e2e/fleetautoscaler_test.go
```diff
@@ -0,0 +1,223 @@
...
+			fleet.Status.ReadyReplicas == 1 &&
+			fleet.Status.Replicas == 1
+	})
+	assert.Nil(t, err)
+}
+
+// TestAutoscalerStressCreate creates many fleetautoscalers with random values
+// to check if the creation validation works as expected and if the fleet scales
+// to the expected number of replicas (when the creation is valid)
+func TestAutoscalerStressCreate(t *testing.T) {
```
markmandel: :+1:

pkg/fleetautoscalers/controller.go
```diff
@@ -0,0 +1,363 @@
...
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(c.logger.Infof)
+	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
+	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "fleetautoscaler-controller"})
+
+	kind := stablev1alpha1.Kind("FleetAutoscaler")
+	wh.AddHandler("/mutate", kind, admv1beta1.Create, c.creationMutationHandler)
+	wh.AddHandler("/validate", kind, admv1beta1.Update, c.mutationValidationHandler)
+
+	fasInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
markmandel: Took this for a spin - works perfectly :+1: 

Question for you - this is only going to run about every ~30 seconds (which is what the current default sync is). Was wondering if that was by design?

I know we talked about slowing down the dropping of the of `GameServers` - but we could watch for the `Allocation` of `GameServers` (watch for `GameServer` update events, where the new state is `Allocated`), and enqueue the FAS at that point - then we would get realtime scaling up, rather than up to 30 seconds wait time.

I'm very happy to drop in the change myself once this PR is merged as well -- not a blocking issue.

victor-prodan: Actually yes, the current behavior is intended.

This is how autoscalers work afaik, they sample the current state at regular intervals.
This is done to reduce noise and to decouple the workload. In a live game , if we link the autoscaler to pod events it will work non-stop.

I also don't think this is really needed. The buffer size you set is supposed to cover the load between scaling intervals.

But sure, if this is really needed maybe it can be added as an optional flag in the spec.
It should come in a package with a cooldown value as well.
So yes, I think it should be done in a separate PR as well.

pkg/fleetautoscalers/controller.go
victor-prodan: I am here to please :grinning: 
Fixed.

test/e2e/fleetautoscaler_test.go
victor-prodan: Glad you like it :smiley: 
I am also planning to add a stress test for scaling during rolling updates, but in a separate PR.

victor-prodan:
Fixed missing doc and rebased

Issue #309, opened 2018-07-28
Helm upgrade and SSL certificates


Kuqd:
When updating using `helm upgrade` the certificates for the admission controller gets regenerated but the controller pod doesn’t  get restarted and thus keep the old certificates which causes this error 

```
GameServer instances failed (): Internal error occurred: failed calling admission webhook 
"mutations.stable.agones.dev": Post https://agones-controller-service.agones-
system.svc:443/mutate?timeout=30s: x509:certificate signed by unknown authority 
(possibly because of "crypto/rsa: verification error" while trying to verify candidate 
authority certificate "admission-controller-ca")
```

using `helm upgrade --recreate-pods` fixes the issue but we needs to document it.

@markmandel are we ok in recreating the controller pod at each upgrade ? Or do you have a better idea ?

markmandel:
Copy pasting chat transcripts, for posterity:

```chat
Mark Mandel [8:13 AM]
Also need to respond to your comment re: the internal ssl cert

I've run into that before as well - I usually hand delete the pod - but it's something we should discuss overall

Cyril Tovena [8:46 AM]
for now in my MR I forced the recreation of the pods
and it fixes the issue
I don't think the fix is easy

Mark Mandel [8:47 AM]
it's really only a development issue
in prod, you'll get a new tagged image, so the original pod will get deleted and replaced

Cyril Tovena [8:48 AM]
true unless you just want to change something else

Mark Mandel [8:48 AM]
oh true!

Cyril Tovena [8:48 AM]
but it's minor you're right I would not spend a lot of time for now

Mark Mandel [8:48 AM]
Is there a way in helm to only generate the cert if it's not there already?

Cyril Tovena [8:48 AM]
that's why I went for a recreate
I did not search more info, they have a solution with sha

Mark Mandel [8:49 AM]
yeah, we could also add a faq/troubleshooting guide to the development guide for this

Cyril Tovena [8:49 AM]
exactly
I think we can live with this, let's just mention it somewhere

Mark Mandel [8:49 AM]
Actually - having a troubleshooting section would be good anyway

Cyril Tovena [8:49 AM]
definitively
```

Kuqd:
Yes indeed, but it’s fixed only if you use the make target to install!

Kuqd:
I'd like to explore this https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#automatically-roll-deployments-when-configmaps-or-secrets-change

[...]

PR #345, opened 2018-09-06
Guidelines on how to test developer changes to Agones

### LINKED ISSUE #308: Update "Developing, Testing and Building Agones" tutorial with how to push updates to your test cluster
I followed the Developing, Testing and Building Agones instructions and they worked pretty well to push and test the first build to GKE.

The problem occurred when I had made some changes locally (to the agones controller) and I wanted to push my changes to the cluster. 

After multiple trial and error I discovered that it is possible by:
- push the new images to gcr.io (make push)
- delete the current agones install by using the generated install yaml file (make shell + kubectl delete ./install/yaml/install.yaml
- re-install agones: exit the shell and make install

I suggest adding this explanation to the tutorial and maybe also add a "make install-update" or "make reinstall" build target.

markmandel:
The developer guide didn't cover what to do "next" after compiling Agones -- i.e. how to make changes and test them.

This PR is an effort to fix this. Since #309 is completed, you don't have to manually delete the agones-controller pod, so this also becomes much simpler.

Closes #308

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -234,6 +234,52 @@ $ make minikube-transfer-image TAG=myimage:0.1
 
 Running end-to-end tests on minikube is done via the `make minikube-test-e2e` target. This target use the same `make test-e2e` but also setup some prerequisites for use with a minikube cluster.
 
+### I Made A Change, Now What?
+
+Awesome! You took your first step to becoming a contributor!
+
+We have several things we can do at this point, some of which you have already seen!
+
+#### Rerun all the tests
+
+Check your code against the current test suite - run `make test` to see if your changes broke any of the current test
+suite.
+
+#### Lint your code
+
+Run `make lint` - this will run the linter over your code (the same as we do in CI), which will look for bad comments,
+errors that haven't been checked, and more. It's a great tool, but can take a while to complete, so you may want to
+run it sparingly.
+
+#### Build and install your changes on a cluster
+
+Assuming you already followed the above section, you should have some kind of test cluster already up, and running
+Agones. To build and install your changes on your cluster, you can repeat the steps you took above - but here's a nice
+single line command:
+
+##### GKE
+
+`make build-images gcloud-auth-docker push install`
+
+This will:
+1. Build the Agones images
+2. Authenticate against your project's GCR instance
+3. Push the image you built to it
+4. Install your version (replacing the current one, if there is one)
+
+You can now test your changes by hand, but also can run `make e2e-test` to run the automated tests as well
+
+##### minikube
+
+`make build-images minikube-push minikube-install`
+
+This will:
+1. Build the Agones images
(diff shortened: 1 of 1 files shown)


build/README.md
```diff
@@ -234,6 +234,52 @@ $ make minikube-transfer-image TAG=myimage:0.1
...
+
+#### Build and install your changes on a cluster
+
+Assuming you already followed the above section, you should have some kind of test cluster already up, and running
+Agones. To build and install your changes on your cluster, you can repeat the steps you took above - but here's a nice
+single line command:
+
+##### GKE
+
+`make build-images gcloud-auth-docker push install`
```
victor-prodan: this didn't work for me.
if you don't do any git operations the image label will remain the same and the pull policy of the controller image is "IfNotPresent", so nothing will happen.

victor-prodan:
I would also suggest adding a section "Running a custom test cluster" , explaining what must be done if you want to use a different repo/k8s provider:
- using REGISTRY env var
- adding "imagePullSecrets" to install.yaml
- ...

build/README.md
markmandel: This should now work, because @Kuqd made it so there is a new annotation that will force the reset on a new install. And if you installed through the `make install` the pull policy should be `Always`. I will confirm though!

markmandel:
@victor-prodan this is all covered here: https://github.com/GoogleCloudPlatform/agones/blob/master/build/README.md#running-a-test-google-kubernetes-engine-cluster

Is there anything missing?
(you definitely don't need to edit your install.yaml)

- We could also mention that you can use local-includes to set the REGISTRY env var.

markmandel:
@victor-prodan I don't run a custom cluster :smile: but this would be a good addition. Would love a PR from someone who has done this! Maybe it's something we can support better in our tooling?

build/README.md
victor-prodan: `Always` means that it will re-pull the image when recreating the controller. The question is: why would the controller be recreated if the spec file is identical?

markmandel: The annotation update _should_ force a redeploy. I will confirm though.

victor-prodan: FYI, when I'm doing 'kubectl apply -f agones.yaml' I get this output:
```
service "agones-controller-service" unchanged
deployment "agones-controller" unchanged
```

Don't know about helm though...

victor-prodan: Ah, so we go to "Running a clustom test cluster" section, as I'm not using GKE :)
Anyway, is this a confirmation that `make install` is always recreating the controller?

victor-prodan:
Yes, I'll look into that

build/README.md
markmandel: Confirmed! `make install` is always recreating the controller.

PR #346, opened 2018-09-07
Switch to golangci-lint

### LINKED ISSUE #187: Integrate a spell and/or grammar check into build system
It's super easy to miss misspelled words in our .md documentation files.

It would be great if there was a spellchecker (and maybe even a grammar / documentation helper?) integrated into the build system to ensure the documentation continues to be written well.

We could probably just install [aspell](http://aspell.net) and call it directly.

Or maybe something like [write-good](https://github.com/btford/write-good) - but I don't know if we want to add node to our build images. Maybe there is a go version somewhere? Or an online service? Possibly worth exploring.

### LINKED ISSUE #237: Use revive instead of gometalinter
Although it might not offer as complete coverage, [revive](https://blog.mgechev.com/2018/05/28/revive-golang-golint-linter/) seems significantly faster than gometalinter with a simpler API. Switching linters might help resolve the gometalinter timeouts we've been seeing in CI.

Kuqd:
I've made the switch to run golang-ci instead of gometalinter.

I've added all supported and not included by default linters of gometalinters and added:

- goimports
- misspell

Excluded files are now automatically handled, but examples/ was also skipped so I'm running the linter twice. See documentation examples below:

```
 # which dirs to skip: they won't be analyzed;
  # can use regexp here: generated.*, regexp is applied on full path;
  # default value is empty list, but next dirs are always skipped independently
  # from this option's value:
  #   	vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
  skip-dirs:
    - src/external_libs
    - autogenerated_by_my_lib

  # which files to skip: they will be analyzed, but issues from them
  # won't be reported. Default value is empty list, but there is
  # no need to include all autogenerated files, we confidently recognize
  # autogenerated files. If it's not please let us know.
  skip-files:
    - ".*\\.my\\.go$"
    - lib/bad.go
```

Found some interesting issues (typo and goimports) but also some false positives (dupl).

We should have this tools included in our IDE (there is vs-code support).

Closes #187 
Closes #237


examples/simple-udp/main.go
```diff
@@ -24,9 +24,10 @@ import (
 	"strings"
 	"time"
```
markmandel: Super mega nit: would love to see this blank line gone :)

pkg/gameservers/portallocator_test.go
```diff
@@ -120,7 +120,7 @@ func TestPortAllocatorAllocate(t *testing.T) {
 			// ports between 10 and 20
 			for i := 10; i <= 20; i++ {
 				var p int32
-				gs, err := pa.Allocate(fixture.DeepCopy()) // nolint: vetshadow
+				gs, err := pa.Allocate(fixture.DeepCopy())
```
markmandel: Just for curiosity - does this mean we aren't checking shadowing, or just that this vet shadow check is a bit more sophisticated?

sdks/go/sdk.go
```diff
@@ -19,11 +19,12 @@ import (
 	"fmt"
 	"time"
```
markmandel: Also mega nit - would like the blank line removed.

sdks/go/sdk_test.go
```diff
@@ -17,12 +17,13 @@ package sdk
 import (
 	"testing"
 
+	"time"
```
markmandel: blank line

examples/simple-udp/main.go
Kuqd: goimport force it, I guess for separetion of standard vs community

markmandel: I mean the one between "time" and "strconv"

pkg/gameservers/portallocator_test.go
Kuqd: Apparently it's true by default.

```
linters-settings:
  govet:
    check-shadowing: true
```

examples/simple-udp/main.go
Kuqd: I'll give a try

markmandel: I've noticed goimports will leave in random blank lines in places, and they just need some manual cleanup.

Kuqd: right ! I'm using goreturns though

pkg/gameservers/portallocator_test.go
Kuqd: So I think it's more sophisticated.

examples/simple-udp/main.go
markmandel: I'm willing to bet if you remove the blank line between it will stay removed. I'd put $10 on it using the same code as goimports to sort the import statements.

Kuqd: you were right

Issue #348, opened 2018-09-10
Improve support for developing in custom environments


victor-prodan:
Improve makefile support and documentation for working with:
- custom cluster providers
- custom image repositories
- multiple clusters

victor-prodan:
Impemented in #349

PR #349, opened 2018-09-10
Improve support for custom dev environments

victor-prodan:
Issue #348

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -38,8 +38,8 @@ release_registry = gcr.io/agones-images
 VERSION ?= $(base_version)-$(shell git rev-parse --short HEAD)
 # The registry that is being used to store docker images
 REGISTRY ?= $(release_registry)
-# Where the kubectl configuration files are being stored
-KUBEPATH ?= ~/.kube
+# kubectl configuration to use
+KUBECONFIG ?= ~/.kube/config
 # The (gcloud) test cluster that is being worked against
 GCP_CLUSTER_NAME ?= test-cluster
 GCP_CLUSTER_ZONE ?= us-west1-c
@@ -54,10 +54,13 @@ mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 build_path := $(dir $(mkfile_path))
 agones_path := $(realpath $(build_path)/..)
 
+kubeconfig_path := $(dir $(KUBECONFIG))
+kubeconfig_file := $(notdir $(KUBECONFIG))
+
 agones_package = agones.dev/agones
 mount_path = /go/src/$(agones_package)
 common_mounts = -v $(build_path)/.config/gcloud:/root/.config/gcloud \
-				-v $(KUBEPATH):/root/.kube \
+				-v $(kubeconfig_path):/root/.kube \
 				-v $(agones_path):$(mount_path)
 
 build_tag = agones-build:$(build_version)
@@ -66,7 +69,7 @@ controller_tag = $(REGISTRY)/agones-controller:$(VERSION)
 sidecar_tag = $(REGISTRY)/agones-sdk:$(VERSION)
 
 go_version_flags = -ldflags "-X agones.dev/agones/pkg.Version=$(VERSION)"
-DOCKER_RUN ?= docker run --rm $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag)
+DOCKER_RUN ?= docker run --rm $(common_mounts) -e "KUBECONFIG=/root/.kube/$(kubeconfig_file)" $(DOCKER_RUN_ARGS) $(build_tag)
 #    ___  ____    ___            _           _
 #   / _ \/ ___|  |_ _|_ __   ___| |_   _  __| | ___
 #  | | | \___ \   | || '_ \ / __| | | | |/ _` |/ _ \
@@ -76,7 +79,7 @@ DOCKER_RUN ?= docker run --rm $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag)
 
 uname := $(shell uname -s)
 ifneq ($(findstring Microsoft,$(shell uname -r)),)
-    osinclude := windows.mk
+	osinclude := windows.mk
 else ifeq ($(uname),Linux)
 	osinclude := linux.mk
 else ifeq ($(uname),Darwin)
@@ -89,7 +92,7 @@ include ./includes/$(osinclude)
 -include ./local-includes/*.mk
 
 ifdef DOCKER_RUN
-    ensure-build-image += ensure-build-image
+	ensure-build-image += ensure-build-image
 endif
 
 #   _____                    _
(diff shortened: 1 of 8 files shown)


build/Makefile
```diff
@@ -40,6 +40,8 @@ VERSION ?= $(base_version)-$(shell git rev-parse --short HEAD)
 REGISTRY ?= $(release_registry)
 # Where the kubectl configuration files are being stored
 KUBEPATH ?= ~/.kube
+# Name of the kubectl configuration file to use
+KUBECONFIGFILE ?= config
```
markmandel: Curious - should we just leverage `KUBECONFIG` rather than the file? So people can overwrite the entire path. Would that be more flexible? (and leverage the same env var that standard Kubernetes tooling uses anyway)

build/Makefile
```diff
@@ -410,3 +418,30 @@ minikube-transfer-image:
...
 minikube-test-e2e: minikube-agones-profile test-e2e 
+
+#   _____          _                  
+#  / ____|        | |                 
+# | |    _   _ ___| |_ ___  _ __ ___  
+# | |   | | | / __| __/ _ \| '_ ` _ \ 
+# | |___| |_| \__ \ || (_) | | | | | |
+#  \_____\__,_|___/\__\___/|_| |_| |_|
+ 
+setup-custom-test-cluster: $(ensure-build-image)
```
markmandel: Minor nit: for consistency sake, this should likely be called `custom-test-cluster` (custom as the prefix)

Looking back, we should be consistent with the prefixes for cluster operations - `gcloud` for GKE (or even `gke`),  `minikube` for minikube, etc. Which we _mostly_  are.

build/README.md
```diff
@@ -234,6 +236,41 @@ $ make minikube-transfer-image TAG=myimage:0.1
...
+
+This section is addressed to developers using a custom Kubernetes provider, a custom image repository and/or multiple test clusters.
+
+Prerequisites:
+- a(some) running k8s cluster(s)
+- Have kubeconfig file(s) ready
+- docker must be logged into the image repository you're going to use
+
+To begin, you need to set up the following environment variables:
+- `KUBEPATH` should point to the folder containing your kubeconfig file;
```
markmandel: I feel like `KUBEPATH` and `KUBECONFIGFILE` could be replaced by `KUBECONFIG` as a full path. WDYT?

markmandel:
@victor-prodan just checking in :) I assume you're busy on #340 - just wanted to make sure you weren't waiting on anything from me.

@Kuqd Did you get a chance to have a look - wanted to make sure I didn't miss anything.

build/Makefile
victor-prodan: I named it 'setup' on purpose because, opposed to the `gcloud-test-cluster` or 'minikube-test-cluster' it's not creating the cluster, but only preparing it for agones.

WDYT?

build/README.md
victor-prodan: I agree. @Kuqd WDYT?

Kuqd:
I will have look, sorry guys !

build/Makefile
Kuqd: I agree the convention is to use `KUBECONFIG` see https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/ default being `$HOME/.kube/config`. We should stick to that so it's more user-friendly and get rid of the KUBEPATH at the same time.

build/Makefile
Kuqd: you could make that target common to every provider and rename to reflect that and use it as a post-target in `gcloud-test-cluster` and  `minikube-test-cluster`

build/Makefile
```diff
@@ -410,3 +418,30 @@ minikube-transfer-image:
...
+	$(DOCKER_RUN) kubectl apply -f $(mount_path)/build/helm.yaml
+	$(DOCKER_RUN) helm init --service-account helm
+
+clean-custom-test-cluster: $(ensure-build-image) $(uninstall)
```
Kuqd: same here this could be common to every provider

build/Makefile
```diff
@@ -410,3 +418,30 @@ minikube-transfer-image:
...
+	docker run --rm -it $(common_mounts) -e "KUBECONFIG=/root/.kube/$(KUBECONFIGFILE)" $(DOCKER_RUN_ARGS) $(build_tag) helm reset
+
+install-custom-pull-secret:
```
Kuqd: same here I could use the same for gke or minikube I think you can make it common. So every provider benefits from it.

build/README.md
Kuqd: yes good call plus both ` KUBEPATH` and `KUBECONFIGFILE` are not standard, while `KUBECONFIG` is , so there is a greater chance user have this set.

build/Makefile
victor-prodan: ok, will do

build/Makefile
victor-prodan: ok, will try

build/README.md
victor-prodan: agreed.

victor-prodan:
@Kuqd: what do you mean by "splitting the makefile into multiple purpose makefile include" more exactly?  I don't know... :disappointed:

build/Makefile
victor-prodan: Done

build/Makefile
victor-prodan: Ok now?

victor-prodan:
I hope I havent forgot any point....

build/Makefile
victor-prodan: done, I hope

markmandel:
i think that was my fault, I was futzing with the cluster.

Kuqd:
@victor-prodan I might having multiple makefile one for build, one minikube, one for setup, etc... , using includes. But let's see that later, we should do another PR it's easier.

install/helm/agones/templates/controller.yaml
```diff
@@ -81,3 +81,7 @@ spec:
       - name: certs
         secret:
           secretName: {{ template "agones.fullname" . }}-cert
+{{- if .Values.agones.image.controller.pullSecret }}
```
markmandel: Just thought of one more thing - this should be documented here:
https://github.com/GoogleCloudPlatform/agones/blob/master/install/helm/README.md#configuration

Which I think will solve #358's issue as well - won't it?

build/Makefile
```diff
@@ -361,25 +389,23 @@ clean-gcloud-config:
 #
 # Use MINIKUBE_DRIVER variable to change the VM driver
 # (defaults virtualbox for Linux and macOS, hyperv for windows) if you so desire.
+minikube-test-cluster: DOCKER_RUN_ARGS="--network=host -v $(minikube_cert_mount) $(DOCKER_RUN_ARGS)"
```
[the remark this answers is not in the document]
victor-prodan: Thank you!

install/helm/agones/templates/controller.yaml
victor-prodan: Yes, indeed.

Kuqd:
You possibly hit #357 ?  @markmandel do you know why 

```
 	reply, err := e2eframework.PingGameServer("ANNOTATION", fmt.Sprintf("%s:%d", readyGs.Status.Address,
 		readyGs.Status.Ports[0].Port))
``
would be out of index

markmandel:
Yep, it's a race condition. Review and merge #357 and it will go away 👍

PR #350, opened 2018-09-10
Add allocator service, docs

slartibaartfast:
This is a service that allocates a game server from a fleet using the Agones API.  It then returns the IP address and port of the allocated game server.  I cleaned up the notes I took while making it, and maybe they might provide some useful documentation for people who are looking to programmatically allocate game servers.

Some things I'm not sure about:
- Is the document create_allocator_service.md more of a Quickstart or a Guide?
- The service in the example uses a container in my public Docker registry...I don't mind, but it might be better off in a repo controlled by this project.
- Step 12 and onward of create_allocator_service might be out of scope.
- The error handling in main.go seem inadequate to me...but maybe it is ok?

Please let me know if this would be useful, and if so, what changes should be made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,14 @@ This software is currently alpha, and subject to change. Not to be used in produ
 - Client SDKs for integration with dedicated game servers to work with Agones.
 
 ## Why does this project exist?
-For more details on why this project was written, read the 
+For more details on why this project was written, read the
 [announcement blog post](https://cloudplatform.googleblog.com/2018/03/introducing-Agones-open-source-multiplayer-dedicated-game-server-hosting-built-on-Kubernetes.html).
 
 ## Requirements
 - Kubernetes cluster version 1.9+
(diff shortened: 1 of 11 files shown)


slartibaartfast:
"could not import github.com/gin-gonic/gin (cannot find package \"github.com/gin-gonic/gin"

I guess there isn't any reason why that package would be available to the build system.

markmandel:
Haven't had a deep look, but any reason this can't be done with just the standard lib, rather than with gin?

slartibaartfast:
I can't think of any reason why not off the top of my head.  I'll give it a try.

markmandel:
You didn't have to close this, you could have update it at a later date?

But whatever works for your process! :smile: happy to have the contribution.

slartibaartfast:
I didn't want to clutter up the list of open pull requests.
I'll open it again or send a new one after replacing gin with standard lib packages.

PR #357, opened 2018-09-18
Fixes race condition bug with Pod not being scheduled before Ready()

### LINKED ISSUE #354: Fix loophole in game server initialization
The normal server init flow is creating->starting->requestready->ready

In creating state we do the following:
- createGameServerPod
- applyGameServerAddressAndPort 

If after creating a pod, we can't retrieve its address for some reason (func applyGameServerAddressAndPort fails), we stop and the game remains in creating state without status unpopulated. 

The pod is created though, and the game server starts and sets itself as ready
Therefore, we jump directly from creating to requestready state, without going through starting
The status is never populated again, because the prerequisite is to be in creating state.

The result is gameservers that look like this:

```
Status:
  Address:
  Node Name:
  Ports:      <nil>
  State:      Ready
```

/cc @markmandel

markmandel:
This fix does several things:

1. Breaks out ip and port setting of the `Status` into its own state, to make
    implementation and testing easier.
1. Listen for Pod updates, and if a Pod is a `GameServer` pod that has had
    its `NodeName` set, then queue the backing `GameServer`
1. Finally - if Ready() gets called before any of this, populate the `Status` ip
   and port values at this time as well.

Fixes #354

### DIFF
--- a/pkg/gameservers/controller_test.go
+++ b/pkg/gameservers/controller_test.go
@@ -15,10 +15,10 @@
 package gameservers
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"testing"
-	"time"
 
 	"agones.dev/agones/pkg/apis/stable"
 	"agones.dev/agones/pkg/apis/stable/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/gameservers/controller_test.go
```diff
@@ -109,7 +120,7 @@ func TestControllerSyncGameServer(t *testing.T) {
 
 		err = c.syncGameServer("default/test")
 		assert.Nil(t, err)
-		assert.Equal(t, 2, updateCount, "update reactor should twice")
+		assert.Equal(t, 3, updateCount, "update reactor should twice")
```
victor-prodan: hum.. thrice? :smile:

markmandel:
I think that was my fault, I was futzing with the cluster.

PR #361, opened 2018-09-21
Agones stable helm repository

### LINKED ISSUE #285: Agones helm repo
We could setup a chart repository to streamline installation.

See:
https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md

Possible options :
- Google Cloud Storage.
- Github pages.

Once this is done installation step would be:

```
helm repo add agones our_repo_url
helm install agones/agones
```

Kuqd:
I've added targets to push helm charts and use them in the `do-release`.

I've also updated the documentation.

Closes #285

### DIFF
--- a/docs/governance/templates/release.md
+++ b/docs/governance/templates/release.md
@@ -13,8 +13,15 @@ See [CHANGELOG](https://github.com/GoogleCloudPlatform/agones/blob/{release-bran
 This software is currently alpha, and subject to change. Not to be used in production systems.
 
 Images available with this release:
+
 - [gcr.io/agones-images/agones-controller:{version}](https://gcr.io/agones-images/agones-controller:{version})
 - [gcr.io/agones-images/agones-sdk:{version}](https://gcr.io/agones-images/agones-sdk:{version})
 - [gcr.io/agones-images/cpp-simple-server:{example-version}](https://gcr.io/agones-images/cpp-simple-server:{example-version})
 - [gcr.io/agones-images/udp-server:{example-version}](https://gcr.io/agones-images/udp-server:{example-version})
 - [gcr.io/agones-images/xonotic-example:{example-version}](https://gcr.io/agones-images/xonotic-example:{example-version})
+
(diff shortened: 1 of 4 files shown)


docs/governance/templates/release.md
```diff
@@ -18,3 +18,6 @@ Images available with this release:
 - [gcr.io/agones-images/cpp-simple-server:{example-version}](https://gcr.io/agones-images/cpp-simple-server:{example-version})
 - [gcr.io/agones-images/udp-server:{example-version}](https://gcr.io/agones-images/udp-server:{example-version})
 - [gcr.io/agones-images/xonotic-example:{example-version}](https://gcr.io/agones-images/xonotic-example:{example-version})
+- [`helm install agones/agones --version {example-version}][https://agones.dev/chart/stable/agones-{example-version}.tgz]
```
markmandel: Small issue - link should be in `()` not `[]` :smile: 

Should this be in it's own section above, rather in images?

install/helm/README.md
```diff
@@ -30,13 +26,12 @@ you can use the helm `--namespace` parameter to specify a different namespace._
...
 > **Tip**: List all releases using `helm list`
->
+
 
 > If you are installing a development build of Agones (i.e. not the 0.4.0 release), you will need to install Agones the following way:
 
 ```bash
-$ cd install/helm/
-$ helm install --name my-release --namespace agones-system agones --set agones.image.tag=0.4.0-481970d
+$ helm install --name my-release --namespace agones-system agones/agones --set agones.image.tag=0.4.0-481970d
```
markmandel: Should we leave this as "cd install/helm" - just because if you are installing the development build, you likely want to do it from github source (which we should probably also be clearer), as the install scripts for development may be different than what was in the previous release. WDYT?

markmandel: For now, should we make it a `git clone` then `cd` into the directory operation?

markmandel: ignore the above. I missed this was outdated.

Issue #368, opened 2018-10-02
Using the Cluster Autoscaler with Agones


markmandel:
## Problem

*   It is desirable to autoscale the Kubernetes cluster to account for increases in player count, and need for more game servers, in a way that will work across cloud
*   There currently exists a generic cluster autoscaler project -- but it is mostly targeted at stateless workloads

## Design

The overall design to use the [open source cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) revolves around using the cluster autoscaler
to remove empty nodes as they appear, but specifically disallows the autoscaler from attempting to evict `GameServer` Pods in an attempt to move them to a new Node.

This gives us the following benefits:

*   Implementations for multiple cloud providers are already written and community tested
*   Load testing (on GKE) has been done for us
*   Existing SLOs already exist for the autoscaler
*   Removes any possibility for the autoscaler to cause race conditions when allocating `GameServers`.

Since the autoscaler can be implement with Agones GameServers -- this essentially means that scaling and autoscaling can be essentially managed at a Fleet level.
If you want a bigger cluster, increase the size of your Fleet, and your cluster will adjust. If you want a smaller cluster, shrink your fleet, and the cluster will adjust.

### Write a scheduler that bin packs all our pods into as tight a cluster as possible

A custom scheduler will be built that will prioritise the scheduling of `GameServer` Pods onto Nodes that have the most `GameServer` pods.
This will ease scaling down, as it will mean the game servers aren't spread out of many Nodes, and there is wasted resource space.

(Unless there is a way to do this with the default scheduler, but I've not found one so far -- best I could find was [PreferredDuringSchedulingIgnoredDuringExecution on HostName](PreferredDuringSchedulingIgnoredDuringExecution))

### Prioritise Allocating GameServers from Nodes that already have Allocated GameServers

To also make it easier to scale down, we essentially want to bin-pack as many allocated game servers on a single node as much as possible.

To that end, the `allocate()` function will order the `GameServers` it is considering by the number of other `GameServers` that exist on the same node as it.

This ensures that we don't end up with (as much as possible) a "swiss cheese" problem, with Allocated game servers spread out across the cluster. Bin packing Allocated `GameServers`
makes it much easier Fleets to scale down in a way that will leave empty nodes for the autoscaler to delete.

### When Fleets get shrunk, prioritise removal from Nodes with the least number of GameServer Pods

[...]

PR #370, opened 2018-10-02
Prioritise Allocation from Nodes with Allocated/Ready GameServers

markmandel:
One of the first parts for Node autoscaling (#368) - make sure we essentially bin pack our allocated game servers.

This change makes allocation first prioritise allocation from `Nodes` that already have the most `Allocated` `GameServers`, and then in the case of a tie, to the `Nodes` that have the most `Ready` `GameServers`.

This sets us up for the next part, such that when we scale down a Fleet, it removes `GameServers` from `Nodes` that have the least `GameServers` on them.

### DIFF
--- a/docs/scheduling_autoscaling.md
+++ b/docs/scheduling_autoscaling.md
@@ -1,0 +1,113 @@
+# Scheduling and Autoscaling
+
+⚠️⚠️⚠️ **This is currently a development feature and has not been released** ⚠️⚠️⚠️
+
+> Autoscaling is currently ongoing work within Agones. The work you see here is just the beginning.
+
+
+Table of Contents
+=================
+
+* [Fleet Autoscaling](#fleet-autoscaling)
+* [Autoscalng Concepts](#autoscalng-concepts)
+   * [Allocation Scheduling](#allocation-scheduling)
+* [Fleet Scheduling](#fleet-scheduling)
+   * [Packed](#packed)
+      * [Allocation Scheduling Strategy](#allocation-scheduling-strategy)
+   * [Distributed](#distributed)
+      * [Allocation Scheduling Stategy](#allocation-scheduling-stategy)
+
+Scheduling and autoscaling go hand in hand, as where in the cluster `GameServers` are provisioned
+impacts how to autoscale fleets up and down (or if you would even want to)
+
+## Fleet Autoscaling
+
+Fleet autoscaling is currently the only type of autoscaling that exists in Agones. It is also only available as a simple
+buffer autoscaling strategy. Have a look at the [Create a Fleet Autoscaler](create_fleetautoscaler.md) quickstart,
+and the [Fleet Autoscaler Specification](fleetautoscaler_spec.md) for details.
+
+Node scaling, and more sophisticated fleet autoscaling will be coming in future releases ([design](https://github.com/GoogleCloudPlatform/agones/issues/368))
+
+## Autoscaling Concepts
+
+To facilitate autoscaling, we need to combine several piece of concepts and functionality, described below.
+
+### Allocation Scheduling
+
+Allocation scheduling refers to the order in which `GameServers`, and specifically their backing `Pods` are chosen
+from across the Kubernetes cluster within a given `Fleet` when [allocation](./create_fleet.md#4-allocate-a-game-server-from-the-fleet) occurs.
+
+## Fleet Scheduling
+
+There are two scheduling strategies for Fleets - each designed for different types of Kubernetes Environments.
+
+### Packed
+
+```yaml
+apiVersion: "stable.agones.dev/v1alpha1"
+kind: Fleet
+metadata:
+  name: simple-udp
+spec:
+  replicas: 100
+  scheduling: Packed
+  template:
+    spec:
+      ports:
+      - containerPort: 7654
+      template:
+        spec:
+          containers:
+          - name: simple-udp
+            image: gcr.io/agones-images/udp-server:0.4
+```
+
+This is the *default* Fleet scheduling strategy. It is designed for dynamic Kubernetes environments, wherein you wish 
+to scale up and down as load increases or decreases, such as in a Cloud environment where you are paying
+for the infrastructure you use.
+
+It attempts to _pack_ as much as possible into the smallest set of nodes, to make
+scaling infrastructure down as easy as possible.
+
+Currently, Allocation scheduling is the only aspect this strategy affects, but in future releases it will
+also affect `GameServer` `Pod` scheduling, and `Fleet` scale down scheduling as well.
+
+#### Allocation Scheduling Strategy
+
+Under the "Packed" strategy, allocation will prioritise allocating `GameServers` to nodes that are running on 
+Nodes that already have allocated `GameServers` running on them.
+
+### Distributed
+
+```yaml
+apiVersion: "stable.agones.dev/v1alpha1"
+kind: Fleet
+metadata:
+  name: simple-udp
+spec:
+  replicas: 100
+  scheduling: Distributed
+  template:
+    spec:
+      ports:
+      - containerPort: 7654
+      template:
+        spec:
+          containers:
+          - name: simple-udp
+            image: gcr.io/agones-images/udp-server:0.4
+```
+
+This Fleet scheduling strategy is designed for static Kubernetes environments, such as when you are running Kubernetes
+on bare metal, and the cluster size rarely changes, if at all.
+
+This attempts to distribute the load across the entire cluster as much as possible, to take advantage of the static
+size of the cluster.
+
+Currently, the only thing the scheduling strategy affects is Allocation scheduling, but in future releases it will
+also affect `GameServer` `Pod` scheduling, and `Fleet` scaledown scheduling as well.
+
+#### Allocation Scheduling Strategy
+
+Under the "Distributed" strategy, allocation will prioritise allocating `GameSerers` to nodes that have the least
+number of allocated `GameServers` on them.
--- a/pkg/apis/stable/v1alpha1/fleet.go
+++ b/pkg/apis/stable/v1alpha1/fleet.go
@@ -22,10 +22,26 @@ import (
 )
 
 const (
+	// Packed scheduling strategy will prioritise allocating GameServers
+	// on Nodes with the most Allocated, and then Ready GameServers
+	// to bin pack as many Allocated GameServers on a single node.
+	// This is most useful for dynamic Kubernetes clusters - such as on Cloud Providers.
+	// In future versions, this will also impact Fleet scale down, and Pod Scheduling.
+	Packed SchedulingStrategy = "Packed"
+
+	// Distributed scheduling strategy will prioritise allocating GameServers
+	// on Nodes with the least Allocated, and then Ready GameServers
+	// to distribute Allocated GameServers across many nodes.
+	// This is most useful for statically sized Kubernetes clusters - such as on physical hardware.
+	// In future versions, this will also impact Fleet scale down, and Pod Scheduling.
+	Distributed SchedulingStrategy = "Distributed"
+
 	// FleetGameServerSetLabel is the label that the name of the Fleet
 	// is set to on the GameServerSet the Fleet controls
 	FleetGameServerSetLabel = stable.GroupName + "/fleet"
 )
+
+type SchedulingStrategy string
 
 // +genclient
 // +genclient:noStatus
@@ -56,6 +72,8 @@ type FleetSpec struct {
 	Replicas int32 `json:"replicas"`
 	// Deployment strategy
 	Strategy appsv1.DeploymentStrategy `json:"strategy"`
+	// Scheduling strategy. Defaults to "Packed".
+	Scheduling SchedulingStrategy `json:"scheduling"`
 	// Template the GameServer template to apply for this Fleet
 	Template GameServerTemplateSpec `json:"template"`
 }
@@ -103,6 +121,10 @@ func (f *Fleet) GameServerSet() *GameServerSet {
 func (f *Fleet) ApplyDefaults() {
 	if f.Spec.Strategy.Type == "" {
 		f.Spec.Strategy.Type = appsv1.RollingUpdateDeploymentStrategyType
+	}
+
+	if f.Spec.Scheduling == "" {
+		f.Spec.Scheduling = Packed
 	}
 
 	if f.Spec.Strategy.Type == appsv1.RollingUpdateDeploymentStrategyType {
--- a/pkg/fleetallocation/controller.go
+++ b/pkg/fleetallocation/controller.go
@@ -20,7 +20,7 @@ import (
 	"sync"
 
 	"agones.dev/agones/pkg/apis/stable"
-	stablev1alpha1 "agones.dev/agones/pkg/apis/stable/v1alpha1"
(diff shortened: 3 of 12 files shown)


pkg/fleetallocation/count.go
```diff
@@ -0,0 +1,47 @@
...
+				count = &nodeCount{}
+			}
+
+			if gs.Status.State == v1alpha1.Allocated {
+				count.allocated++
+			} else if gs.Status.State == v1alpha1.Ready {
+				count.ready++
+			}
+
+			result[gs.Status.NodeName] = count
```
victor-prodan: Update only If changed

pkg/fleetallocation/controller.go
```diff
@@ -272,6 +273,29 @@ func (c *Controller) allocate(f *stablev1alpha1.Fleet, fam *stablev1alpha1.Fleet
 		return allocation, err
 	}
 
+	// for each node, count how many ready and allocated gameservers exist
+	nodeCounts := countReadyAndAllocatedPerNode(gsList)
+
+	// prioritise by nodes that already have allocated gameservers, and then by ready
+	sort.Slice(gsList, func(i, j int) bool {
```
victor-prodan: We are allocating just one server here so we dont need to actually sort the list (nlogn), just to find the best one (O(n)).

So we should combine this new logic with the iteration below.

victor-prodan:
IT should work, but I have some performance concerns.

pkg/fleetallocation/count.go
```diff
@@ -0,0 +1,47 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fleetallocation
+
+import "agones.dev/agones/pkg/apis/stable/v1alpha1"
+
+type nodeCount struct {
+	ready     int64
```
victor-prodan: Why int64 and not just int?

pkg/fleetallocation/controller.go
markmandel: Excellent catch. Working on it now.

pkg/fleetallocation/count.go
```diff
@@ -0,0 +1,75 @@
...
+	var bestGS *v1alpha1.GameServer
+	for nodeName, count := range counts {
+		update := false
+		// if there is no best GameServer, then this node & GameServer is the always the best
+		if bestGS == nil {
+			update = true
+		} else if count.allocated == bestCount.allocated && count.ready > bestCount.ready {
+			update = true
+		} else if count.allocated > bestCount.allocated {
+			update = true
```
victor-prodan: BestNode should have at least one ready gs.

pkg/fleetallocation/find.go
```diff
@@ -0,0 +1,76 @@
...
+		if bestGS == nil {
+			update = true
+		} else if count.allocated == bestCount.allocated && count.ready > bestCount.ready {
+			update = true
+		} else if count.allocated > bestCount.allocated {
+			update = true
+		}
+
+		if update {
+			bestCount = count
```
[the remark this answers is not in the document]
markmandel: I don't _think_ that it will - but, it definitely makes the code easier to understand with it together - so change incoming :+1:

pkg/fleetallocation/find_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"testing"
+
+	"agones.dev/agones/pkg/apis/stable/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestFindReadyGameServer(t *testing.T) {
```
victor-prodan: Add this test case: node1 with 4 allocated (first in list), then node 2 with 4 ready.

Node1 would be the trap I mentioned earlier.

Thing is that allocated is the primary sort key, but ready>0 is the veto condition. Actually you could also include this in the 'best' search above. 

Its quite late here so I am not sure, am I making sense? 😴

victor-prodan:
Together with this change  I think you should also include an allocation strategy param in the allocation spec: stacked or flat. In my case, runnning a static cluster, i want the load to be distributed evenly on the machines.

WDYT?

markmandel:
Ooh interesting - should this be on the allocation, or on the Fleet? I kinda feel like it should be at the Fleet level, because it likely also impacts downscaling/bin packing scheduling/etc as well.

pkg/fleetallocation/find_test.go
markmandel: I added the extra test - seems to be passing -- maybe I got it wrong? :man_shrugging: :smile:

victor-prodan: That's right, it was working before because you were setting `update=true` if bestGs=nil` as the first condition. If you were using 'bestCount=nil', it wouldn't have worked, and for someone who reads the code the reason is not easily seen as they should be an atomic pair. 

The rules (which you explained on Slack and in the design ticket) were very clear, but are not transposed into the code clearly imo. For example, the update logic is split into two places: first you set an `update` bool that is actually half of the truth and then you apply another condition. I will add a few notes inline for some specific things I would do differently.

pkg/fleetallocation/find.go
```diff
@@ -0,0 +1,76 @@
...
+	allocated int64
+}
+
+// findReadyGameServerForAllocation is a O(n) implementation to find a GameServer with priority
+// on Nodes with GameServers that are allocated, and then Nodes with the most
+// Ready GameServers
+func findReadyGameServerForAllocation(gsList []*v1alpha1.GameServer) *v1alpha1.GameServer {
+	counts := map[string]*nodeCount{}
+	// track potential gameservers, one for each node
+	gsOptions := map[string]*v1alpha1.GameServer{}
```
victor-prodan: I don't understand the name `gsOptions`. I would rename it to `gsInstance`, `gsAllocatableInstance` or `gsReadyInstance`, because that's what it will contain.

pkg/fleetallocation/find.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+		}
+	}
+
+	var bestCount *nodeCount
+	var bestGS *v1alpha1.GameServer
+	for nodeName, count := range counts {
+		update := false
+		// if there is no best GameServer, then this node & GameServer is the always the best
+		if bestGS == nil {
```
victor-prodan: I would change this to 'bestGS == nil && count.ready > 0`, because this is the valid condition to update bestGS imo

pkg/fleetallocation/find.go
```diff
@@ -0,0 +1,76 @@
...
+			update = true
+		} else if count.allocated == bestCount.allocated && count.ready > bestCount.ready {
+			update = true
+		} else if count.allocated > bestCount.allocated {
```
victor-prodan: I could change this condition to `count.allocated > bestCount.allocated && count.ready > 0` to be more correct.

pkg/fleetallocation/find.go
```diff
@@ -0,0 +1,76 @@
...
+			update = true
+		}
+
+		if update {
+			// we may not have any GameServers on this node, so check first!
```
victor-prodan: We shouldn't try to update if we know that the current node has no ready nodes. Therefore, I would `assert.NotNil(gsOptions[nodeName])`.

victor-prodan:
@markmandel 

> Ooh interesting - should this be on the allocation, or on the Fleet? I kinda feel like it should be at the Fleet level, because it likely also impacts downscaling/bin packing scheduling/etc as well.

Yes, indeed. Fleet is a better place.

> Previously, it was essentially a random distribution of allocated game servers (because listing tends to be random) - so that was not an official round-robin -- would you want actual round robin, or is random okay (as it was)?
> Ditto for scale down, it was essentially random.

Random is fine, I don't see any advantage in a real round-robin. What I would add would be a flat/evenly distributed allocation, in which you allocate a server on a machine that has the least amount of ready servers (to spread the load evenly).
 
> This is why I think it should be at the fleet level -- the fleet "optimisation" (is that a good name?) would determine how allocation, scaling down and what scheduler the backing Pods would get
> `stacked`: basically everything in #370
> `random`: What currently exists now, default scheduler
> Don't know if it's the right names but does that make sense?

Yes, it makes sense.  I don't agree with the `optimization` though, I see no relation to its meaning. Maybe `operation`, because it's related to how it operates? Or even `allocation`, because it's related to both allocating pods on nodes (aka scheduling) and allocating gs? Must think more on this.

We would have the following operation/allocation strategies:
- `stacked` and `random` as you said
- `flat/even/distributed` (must find a good name), which means - use the default scheduler for pod allocation and allocate a gs from the node with the least number of allocated gs.

And maybe it's not a subject for current PR, but I think we should also consider how to handle different node pools/different machine types. Both for `stacked` and `distributed` strategies we should actually care not for the absolute number of instances on machine, but for the fraction `instances/machine_size`. In case of 'stacked', 4 instances running of a 2-core machine is more than 8 instances running on a 16-core machine.

WDYT?

pkg/fleetallocation/find.go
markmandel: Go doesn't really have a runtime assert, all I can do is check if it exists, and return an error if it doesn't (or nil, which is caught later)

markmandel:
I'd like to keep it to at most 2 scheduling strategies (maybe we should call it `scheduling` ?), as with each strategy this can greatly increase our testing workload, and also increase the cognitive overhead for users. 

I think our target is (1) dynamic systems, i.e. cloud providers -- which would be `packed` and (2) static systems, in which we want more even distribution, so `distributed` (I think I like that the best) -- I say let's drop `random`, it doesn't sound like it's actually needed, as long as we have `even`

The other fun question - do you think this configuration option should be part of this PR? I'm leaning yes, but could be swayed the other way.

Re: fractions -- totally agreed, but I think let's not worry about this for this PR, we can come back to it.

pkg/fleetallocation/find.go
markmandel: I reworked this a bit, to try and make it more readable -- agreed with lots of what you wrote, tweaked a couple of things. Would love you to take another pass :+1: Thanks for taking all the time to do such a detailed review.

markmandel:
I am thinking though, that unless we have both strategies in place, we shouldn't push this PR into the next release -- it's a pretty big change, and without an option to backout - I think that would be bad.

I'm going to try and get the `distributed` option done this week though in time for RC next week though.

victor-prodan:
> I'd like to keep it to at most 2 scheduling strategies (maybe we should call it `scheduling` ?)

👍 

> The other fun question - do you think this configuration option should be part of this PR? 

Yes, for the same reason you mentioned yourself- it's a pretty big change, and it could cause surprising side effects. Some kind of off switch is needed in this case.

> Re: fractions -- totally agreed, but I think let's not worry about this for this PR, we can come back to it.

👍

markmandel:
This is ready for review again! We now have two `scheduling` strategies - `Packed` and `Distributed`, and documentation to go with it!

pkg/apis/stable/v1alpha1/fleet.go
```diff
@@ -22,11 +22,29 @@ import (
 )
 
 const (
+	// Packed scheduling strategy will prioritise allocating GameServers
+	// on Nodes with the most Allocated, and then Ready GameServers
+	// to bin pack as many Allocated GameServers on a single node.
+	// This is most useful for dynamic Kubernetes clusters - such as on Cloud Providers.
+	// In future versions, this will also impact Fleet scale down, and Pod Scheduling.
+	// TODO: example and document
```
victor-prodan: Remove todo

test/e2e/fleet_test.go
```diff
@@ -39,25 +39,34 @@ const (
...
 	t.Parallel()
 
-	fleets := framework.AgonesClient.StableV1alpha1().Fleets(defaultNs)
-	flt, err := fleets.Create(defaultFleet())
-	if assert.Nil(t, err) {
-		defer fleets.Delete(flt.ObjectMeta.Name, nil) // nolint:errcheck
-	}
+	fixtures := []v1alpha1.SchedulingStrategy{v1alpha1.Packed, v1alpha1.Distributed}
+
+	for _, strategy := range fixtures {
```
victor-prodan: Nice!

pkg/apis/stable/v1alpha1/fleet.go
markmandel: Well spotted!

docs/scheduling_autoscaling.md
```diff
@@ -0,0 +1,53 @@
...
+
+This is the *default* Fleet scheduling strategy. It is designed for dynamic Kubernetes environments, wherein you wish 
+to scale up and down as load increases or decreases, such as in a Cloud environment where you are paying
+for the infrastructure you use.
+
+It attempts to _pack_ as much as possible into the smallest set of nodes, to make
+scaling infrastructure down as easy as possible.
+
+Currently, Allocation scheduling is the only aspect this strategy affects, but in future releases it will
+also affect `GameServer` `Pod` scheduling, and `Fleet` scale down scheduling as well.
```
Kuqd: `allocation scheduling` is difficult to grasp, I do understand but I'm not sure everyone will. Should we explain or refer to what allocation is ? Or give concrete example.

pkg/fleetallocation/find.go
```diff
@@ -0,0 +1,88 @@
...
+	counts := map[string]*nodeCount{}
+	// track potential gameservers, one for each node
+	allocatableGameServers := map[string]*v1alpha1.GameServer{}
+
+	// count up the number of allocated and ready game servers that exist
+	// also, since we're already looping through, track one Ready GameServer
+	// per node, so we can use that as a short list to allocate from
+	for _, gs := range gsList {
+		if gs.DeletionTimestamp.IsZero() &&
+			(gs.Status.State == v1alpha1.Allocated || gs.Status.State == v1alpha1.Ready) {
```
Kuqd: I'm expecting a lot of allocations so we might want to find a way latter to cache that informations or have a single source of truth instead of traversing the list. Food for thoughts.

pkg/fleetallocation/find.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// track the best node count
+	var bestCount *nodeCount
+	// the current GameServer from the node with the most GameServers (allocated, ready)
+	var bestGS *v1alpha1.GameServer
+
+	for nodeName, count := range counts {
+		// count.ready > 0: no reason to check if we don't have ready GameServers on this node
+		// bestGS == nil: if there is no best GameServer, then this node & GameServer is the always the best
+		if count.ready > 0 && (bestGS == nil || comparator(bestCount, count)) {
```
Kuqd: I like this comparator, this is also nice for testing.

pkg/fleetallocation/find.go
markmandel: Yeah, I was debating that too - maybe we keep a running count for each node at all times - but I was concerned that would add complexity. But definitely food for thought, and potentially an optimisation as we build out.

docs/scheduling_autoscaling.md
markmandel: That's a fair comment - I'll take another pass, see if I can add some more to make it clearer.

markmandel: Updated - hopefully this is clearer.

PR #372, opened 2018-10-04
Add Amazon EKS Agones Setup Instructions

GabeBigBoxVR:
Added basic instructions to get Agones up and running in the EKS networking environment.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -15,6 +15,10 @@ In this quickstart, we will create a Kubernetes cluster, and populate it with th
    1. [Installing Minikube](#installing-minikube)
    1. [Creating an agones profile](#creating-an-agones-profile)
    1. [Starting Minikube](#starting-minikube)
+1. [Setting up an Amazon Web Services EKS cluster](#setting-up-an-amazon-web-services-eks-cluster)
+    1. [Create EKS Instance](#create-eks-instance)
+    1. [Ensure VPC CNI 1.2 is Running](#ensure-vpc-cni-12-is-running)
+    1. [Follow Normal Instructions to Install](#follow-normal-instructions-to-install)
 1. [Setting up an Azure Kubernetes Service (AKS) cluster](#setting-up-an-azure-kubernetes-service-aks-cluster)
     1. [Choosing your shell](#choosing-your-shell)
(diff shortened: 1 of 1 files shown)


GabeBigBoxVR:
I signed it!

install/README.md
```diff
@@ -174,6 +179,56 @@ minikube start --kubernetes-version v1.10.0 --vm-driver virtualbox \
...
+apiVersion: "stable.agones.dev/v1alpha1"
+kind: GameServer
+metadata:
+  name: "simple-udp"
+spec:
+  ports:
+  - name: default
+    portPolicy: "dynamic"
+    containerPort: 7654             # ADD THIS
+    hostPort: 7654                  # ADD THIS
```
markmandel: Why do you need to set the hostPort, when the port policy is dynamic?

Also, you shouldn't have to set it at the pod level either - Agones should take care of that for you -- was there issue without it?

GabeBigBoxVR:
I signed it!

install/README.md
GabeBigBoxVR: Good catch, I set it since in my case it's portPolicy: "static"
No, I haven't tested it only at the pod level, let me try that now and verify it works.

install/README.md
```diff
@@ -15,6 +15,11 @@ In this quickstart, we will create a Kubernetes cluster, and populate it with th
    1. [Installing Minikube](#installing-minikube)
    1. [Creating an agones profile](#creating-an-agones-profile)
    1. [Starting Minikube](#starting-minikube)
+1. [Setting up an Amazon Web Services EKS cluster](#setting-up-an-amazon-web-services-eks-cluster)
+    1. [Create EKS Instance](#create-eks-instance)
+    1. [Ensure VPC CNI 1.2 is Running](#ensure-vpc-cni-12-is-running)
+    1. [Follow Normal Instructions to Install](#follow-normal-instructions-to-install)
+    1. [Add HostPort to fleet or gameserver instructions](#add-hostport-to-fleet-or-gameserver-instructions)
```
markmandel: I believe this should go too? :smile: 

Also, please rebase down to a single commit - and once this header is gone, happy to approve and merge.

GabeBigBoxVR: 👍

PR #390, opened 2018-10-18
Specify CPU Request for the SDK Server Sidecar

### LINKED ISSUE #344: Agones sdk-server sidecar should have cpu and memory limits
Right now the sdk-server sidecar will default to 0.1 cpu usage (not sure on memory).

This feels like overkill for the sidecar server, as it doesn't do very much, and is a waste of good cpu!

We should measure this, and set appropriate limits.

We should also make this helm configurable.

## Implementation
Add resource limits here:
https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameservers/controller.go#L440

markmandel:
This provides the mechanism (and defaults) for being able to set both the CPU request, and CPU limits for the SDK Server `GameServer` sidecar.

I've only set the Request level, as it seems that the major issue is not CPU usage, but actually how the scheduler allots space for the sidecar (by default 100/0.1 vCPU is alloted to each container.

I've set the default request level to be 5m/0.005 vCPU -- while this is above what load tests have shown, I wanted to be conservative. Also, the controls exist to tweak this value yourself via the Helm chart.

I've not set a CPU limit, as I found when setting a low (<= 20m) CPU limit on the sidecar it mostly stopped working. But if people want to experiment with this, it is also configurable via the Helm chart.

Closes #344

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Documentation and usage guides on how to develop and host dedicated game servers
 
 ### Advanced
 - [Scheduling and Autoscaling](./docs/scheduling_autoscaling.md)
+- [Limiting CPU/Memory](./docs/limiting_resources.md)
 
 ## Get involved
 
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -40,20 +40,23 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	extclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
 const (
-	sidecarFlag     = "sidecar"
-	pullSidecarFlag = "always-pull-sidecar"
-	minPortFlag     = "min-port"
-	maxPortFlag     = "max-port"
-	certFileFlag    = "cert-file"
-	keyFileFlag     = "key-file"
-	workers         = 2
-	defaultResync   = 30 * time.Second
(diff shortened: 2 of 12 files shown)


markmandel:
Looking at the graphs, it never gets anywhere close to 20m/0.02 (the max is ~ 0.003) :man_shrugging: but maybe it's being sampled?

Looking at the [limits description](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-limits-are-run):

> The spec.containers[].resources.limits.cpu is converted to its millicore value and multiplied by 100. The resulting value is the total amount of CPU time that a container can use every 100ms. A container cannot use more than its share of CPU time during this interval.

Let's see if I can get this math right.
- 20m = 0.02
- 0.02 * 100 = 2
- So, the container gets 2ms every 100ms, which is only 20ms per 1s.
- Maybe this is actually too little to get things done? (But you likely have a better idea of lower level cpu cycles than I do)

For requests:

> The spec.containers[].resources.requests.cpu is converted to its core value, which is potentially fractional, and multiplied by 1024. The greater of this number or 2 is used as the value of the --cpu-shares flag in the docker run command.

Looking at [`cpu-shares`](https://docs.docker.com/engine/reference/run/#cpu-share-constraint), this is a CPU limiter, but:

> The proportion will only apply when CPU-intensive processes are running. When tasks in one container are idle, other containers can use the left-over CPU time. The actual amount of CPU time will vary depending on the number of containers running on the system.

That being said - this is also why I allowed both to configured, in case this is all totally wrong, it can be adjusted as necessary.

20m mostly worked -- 30m was more consistently good -- at least on the CPUs I was running in my cluster.

WDYT?

markmandel:
Not hearing any objections - so updated to 30m on the cpu request, and also wrote an advanced doc on cpu and memory limiting.

PR #375, opened 2018-10-04
Add crd-install helm hook to crds templates

smoya:
From Helm documentation:
> crd-install: Adds CRD resources before any other checks are run. This is used only on CRD definitions that are used by other manifests in the chart.

This PR aims to update agones helm charts crds adding the `crd-install` hook, ensuring then the installation of those to be done before other workloads.

This change also fixes third party charts with agones as dependency that deploy workloads related to those custom resources.

PR #396, opened 2018-10-23
Update Helm to 2.11.0

markmandel:
This will likely fail on first PR, as we will need to update the Helm install on the e2e cluster.

A good opportunity to delete and restart the e2e cluster anyway -- but will need to manage this around other PRs.

Should enable #375 to pass tests as well.


Kuqd:
Should we also update our e2e image ? it might cause issue though as the tiller won't match the right version.

markmandel:
Cool - this failed as expected now!

When this PR get's approval, I'll manage creating the new cluster / upgrading helm by hand before running this again.

PR #401, opened 2018-10-25
Packed: Fleet scaled down removes GameServers from least used Nodes

markmandel:
This implements the strategy such that when a Fleet is scaled down, and
has the "Packed" strategy, it removed GameServers from Nodes that have the
least GameServers running on them.

This also fixes up some issues with the OpenAPI validation as well.

(Sorry - there was a bunch of cleanup in here, lemme know if you want me to point out the most important sections - basically controller.go and `filterGameServersOnLeastFullNodes` implementation)

### DIFF
--- a/pkg/gameserversets/gameserversets.go
+++ b/pkg/gameserversets/gameserversets.go
@@ -15,22 +15,81 @@
 package gameserversets
 
 import (
-	stablev1alpha1 "agones.dev/agones/pkg/apis/stable/v1alpha1"
+	"sort"
+
+	"agones.dev/agones/pkg/apis/stable/v1alpha1"
 	listerv1alpha1 "agones.dev/agones/pkg/client/listers/stable/v1alpha1"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+// node is just a convenience data structure for
+// keeping relevant GameServer information about Nodes
+type node struct {
+	name  string
+	total int64
+	ready []*v1alpha1.GameServer
+}
+
+// filterGameServersOnLeastFullNodes returns a limited list of GameServers, ordered by the nodes
+// they are hosted on, with the least utilised Nodes being prioritised
+func filterGameServersOnLeastFullNodes(gsList []*v1alpha1.GameServer, limit int32) []*v1alpha1.GameServer {
+	if limit <= 0 {
+		return nil
+	}
+
+	nodeMap := map[string]*node{}
+	var nodeList []*node
+
+	// count up the number of allocated and ready game servers that exist
+	// also, since we're already looping through, track all the deletable GameServers
+	// per node, so we can use this as a shortlist to delete from
+	for _, gs := range gsList {
+		if gs.DeletionTimestamp.IsZero() &&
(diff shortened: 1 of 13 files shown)


pkg/gameserversets/gameserversets.go
```diff
@@ -15,22 +15,83 @@
...
+// keeping relevant GameServer information about Nodes
+type node struct {
+	name  string
+	total int64
+	ready []*v1alpha1.GameServer
+}
+
+// filterGameServersOnLeastFullNodes returns a limited list of GameServers, ordered by the nodes
+// they are hosted on, with the least utilised Nodes being prioritised
+func filterGameServersOnLeastFullNodes(gsList []*v1alpha1.GameServer, limit int32) []*v1alpha1.GameServer {
```
markmandel: @victor-prodan I took my best shot at optimising this, but you always have a good eye for algorithm optimisations - any thoughts?

victor-prodan: It had been a very busy week for me, I will have a look this weekend

markmandel: No dramas. End of year I'm not travelling much, so I have more time to write code and not be on airplanes. Take your time. :+1:

pkg/gameserversets/gameserversets.go
```diff
@@ -15,22 +15,83 @@
...
+	// also, since we're already looping through, track all the deletable GameServers
+	// per node, so we can use this as a shortlist to delete from
+	for _, gs := range gsList {
+		if gs.DeletionTimestamp.IsZero() &&
+			(gs.Status.State == v1alpha1.Allocated || gs.Status.State == v1alpha1.Ready) {
+			_, ok := nodeMap[gs.Status.NodeName]
+			if !ok {
+				node := &node{name: gs.Status.NodeName}
+				nodeMap[gs.Status.NodeName] = node
+				nodeList = append(nodeList, node)
```
victor-prodan: doing memory allocations during a loop is not a good idea.
memory allocations are generally expensive, with a non-constant cost.
it's better to reserve memory in advance even if it's more than you need.
in this case, I suppose that the total number of nodes in the cluster should be already known locally?

pkg/gameserversets/gameserversets.go
```diff
@@ -15,22 +15,83 @@
...
+		}
+	}
+
+	// sort our nodes, least to most
+	sort.Slice(nodeList, func(i, j int) bool {
+		return nodeList[i].total < nodeList[j].total
+	})
+
+	// we need to get Ready GameServer until we equal or pass limit
+	result := make([]*v1alpha1.GameServer, 0, limit)
```
victor-prodan: this is good :+1:

markmandel:
@EricFortin Oops - yep, changed the title!

pkg/gameserversets/gameserversets.go
```diff
@@ -15,22 +15,83 @@
...
+			(gs.Status.State == v1alpha1.Allocated || gs.Status.State == v1alpha1.Ready) {
+			_, ok := nodeMap[gs.Status.NodeName]
+			if !ok {
+				node := &node{name: gs.Status.NodeName}
+				nodeMap[gs.Status.NodeName] = node
+				nodeList = append(nodeList, node)
+			}
+
+			if gs.Status.State == v1alpha1.Allocated {
+				nodeMap[gs.Status.NodeName].total++
```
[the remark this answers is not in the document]
markmandel: Nice catch! Done.

PR #408, opened 2018-11-09
Mark GameServer Unhealthy if allocated HostPort isn't available

markmandel:
This marks a GameServer as unhealthy is the allocated port isn't available on a given Node for whatever reason.

Since it's marked as Unhealthy, the Fleet controller will delete it, and the new GameServer will be allocated a new random port that it deems open and available.

I've yet to see this in the wild,  but I had mild concerns, so figured I'd fix it now rather than later.


pkg/gameservers/health.go
```diff
@@ -126,7 +145,23 @@ func (hc *HealthController) syncGameServer(key string) error {
...
-	if gs.Status.State == v1alpha1.Ready {
+	var reason string
+	unhealthy := false
+
+	switch gs.Status.State {
+
+	case v1alpha1.Starting:
+		hc.logger.WithField("key", key).Info("GameServer cannot start on this port")
+		unhealthy = true
+		reason = "No more of the allocated ports free"
```
EricFortin: Your method and test works with `free ports` but your real code return a string with `ports free`. These need to be aligned.

markmandel: Done!

PR #416, opened 2018-11-15
FleetAutoscaler can be targeted at Non Existent Fleets

### LINKED ISSUE #406: Admission webhook "mutations.stable.agones.dev" errors with Invalid FleetAutoscaler
**Bare-metal k8s v1.10.4, helm v2.9.1, Agones (installed by helm) v0.5.0, 
Linux 4.4.159-1.el7.elrepo.x86_64, Centos 7.4**

Custom helm chart for my deployment with fleet and fleetAutoscaler
```
  policy:
    type: Buffer
    buffer:
      bufferSize: 2
      minReplicas: 0
      maxReplicas: 10
```
At **helm install** almost always successful, but sometimes(1:10):
```
Release "xxx" does not exist. Installing it now.
Error: release xxx failed: Internal error occurred: admission webhook 
  "mutations.stable.agones.dev" denied the request: Invalid FleetAutoscaler
```
Same error in kube-apiserver log:
```
net/http.HandlerFunc.ServeHTTP(0xc42a10c2e0, 0x7f4ba00d9160, 0xc4216cf320, 0xc44a31e200)
        /usr/local/go/src/net/http/server.go:1918 +0x44
k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/server/filters.(*timeoutHandler).ServeHTTP.func1(0xc42a10c380, 0xb1e9820, 0xc4216cf320, 0xc44a31e200, 0xc43dd541e0)
        /workspace/anago-v1.10.4-beta.0.68+5ca598b4ba5abb/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/server/filters/timeout.go:93 +0x8d
created by k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/server/filters.(*timeoutHandler).ServeHTTP
        /workspace/anago-v1.10.4-beta.0.68+5ca598b4ba5abb/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/server/filters/timeout.go:92 +0x1ab

logging error output: "{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"Internal error occurred: admission webhook \\\"mutations.stable.agones.dev\\\" denied the request: Invalid FleetAutoscaler\",\"reason\":\"InternalError\",\"details\":{\"causes\":[{\"message\":\"admission webhook \\\"mutations.stable.agones.dev\\\" denied the request: Invalid FleetAutoscaler\"}]},\"code\":500}\n"
 [[tiller/v0.0.0 (linux/amd64) kubernetes/$Format] 10.10.20.31:34754]
```

As background - 20-40 fleets with autoscalers in one namespace with frequent parallel redeploy (install/upgrade/delete via helm)

Which log can help resolve this critical issue?
Almost always sequential **./helm upgrade --force --install --wait --timeout 1200 ...** resolve issue.

Maybe race condition between fleet and correspond autoscaler creation/update?

Update 1: I try apply autoscaler without fleet - got same error
```
kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/agones/master/examples/simple-udp/fleetautoscaler.yaml

Error from server (InternalError): error when creating "https://raw.githubusercontent.com/GoogleCloudPlatform/agones/master/examples/simple-udp/fleetautoscaler.yaml": Internal error occurred: admission webhook "mutations.stable.agones.dev" denied the request: Invalid FleetAutoscaler
```
Update 2: https://github.com/helm/helm/issues/2994 - common CRD issue...

markmandel:
This removes some previous restrictions with the FleetAutoscaler,
as well as cleaning up a couple of items.

- FleetAutoscalers can now have a Fleet target that doesn't exit.
  - If the target doesn't exist, this will be recorded as a warning event of type `FailedGetFleet`, rather than `ScalingLimited` status item, as it allows a human readable message, as well as give us data on how often this occurs, and it is readable by stat/alerting packages.
- FleetAutoscalers can now have their Fleet target edited. This means that FleetAutoscalers can be transitioned from one Fleet to another. This may be useful in scenarios like red-green deployments, or if you wish to temporarily disable an Autoscaler, but not delete it.
- `FleetAutoscaler.Status.ScalingLimited` was also replaced by a `ScalingLimited` Event, as it allows a human readable message, as well as give us data on how   often this occurs, and it is readable by stat/alerting packages.

Closes #406

### DIFF
--- a/pkg/apis/stable/v1alpha1/fleetautoscaler.go
+++ b/pkg/apis/stable/v1alpha1/fleetautoscaler.go
@@ -114,29 +114,16 @@ type FleetAutoscalerStatus struct {
 	ScalingLimited bool `json:"scalingLimited"`
 }
 
-// ValidateUpdate validates when an update occurs
-func (fas *FleetAutoscaler) ValidateUpdate(new *FleetAutoscaler, causes []metav1.StatusCause) []metav1.StatusCause {
-	if fas.Spec.FleetName != new.Spec.FleetName {
-		causes = append(causes, metav1.StatusCause{
-			Type:    metav1.CauseTypeFieldValueInvalid,
-			Field:   "fleetName",
-			Message: "fleetName cannot be updated",
-		})
-	}
-
-	return new.ValidateAutoScalingSettings(causes)
-}
-
-// ValidateAutoScalingSettings validates the FleetAutoscaler scaling settings
-func (fas *FleetAutoscaler) ValidateAutoScalingSettings(causes []metav1.StatusCause) []metav1.StatusCause {
+// Validate validates the FleetAutoscaler scaling settings
+func (fas *FleetAutoscaler) Validate(causes []metav1.StatusCause) []metav1.StatusCause {
 	if fas.Spec.Policy.Type == BufferPolicyType {
-		causes = fas.Spec.Policy.Buffer.ValidateAutoScalingBufferPolicy(causes)
+		causes = fas.Spec.Policy.Buffer.ValidateBufferPolicy(causes)
 	}
 	return causes
 }
 
-// ValidateAutoScalingBufferPolicy validates the FleetAutoscaler Buffer policy settings
-func (b *BufferPolicy) ValidateAutoScalingBufferPolicy(causes []metav1.StatusCause) []metav1.StatusCause {
+// ValidateBufferPolicy validates the FleetAutoscaler Buffer policy settings
+func (b *BufferPolicy) ValidateBufferPolicy(causes []metav1.StatusCause) []metav1.StatusCause {
 	if b == nil {
 		return append(causes, metav1.StatusCause{
 			Type:    metav1.CauseTypeFieldValueInvalid,
--- a/install/helm/agones/templates/admissionregistration.yaml
+++ b/install/helm/agones/templates/admissionregistration.yaml
@@ -86,7 +86,6 @@ webhooks:
           - "gameservers"
           - "fleets"
           - "fleetallocations"
-          - "fleetautoscalers"
         apiVersions:
           - "v1alpha1"
         operations:
--- a/install/yaml/install.yaml
+++ b/install/yaml/install.yaml
@@ -925,7 +925,6 @@ webhooks:
           - "gameservers"
           - "fleets"
           - "fleetallocations"
-          - "fleetautoscalers"
         apiVersions:
           - "v1alpha1"
         operations:
--- a/pkg/apis/stable/v1alpha1/fleetautoscaler_test.go
+++ b/pkg/apis/stable/v1alpha1/fleetautoscaler_test.go
@@ -25,70 +25,10 @@ import (
 func TestFleetAutoscalerValidateUpdate(t *testing.T) {
 	t.Parallel()
 
-	t.Run("same fleet name", func(t *testing.T) {
-
-		fas := &FleetAutoscaler{
-			ObjectMeta: metav1.ObjectMeta{Name: "test"},
-			Spec: FleetAutoscalerSpec{
-				FleetName: "testing",
-				Policy: FleetAutoscalerPolicy{
-					Type: BufferPolicyType,
-					Buffer: &BufferPolicy{
-						BufferSize:  intstr.FromInt(1),
-						MaxReplicas: 10,
-					},
-				},
-			},
-		}
-
-		causes := fas.ValidateUpdate(fas.DeepCopy(), nil)
-		assert.Len(t, causes, 0)
-	})
-
-	t.Run("different fleet name", func(t *testing.T) {
-
-		fas := &FleetAutoscaler{
-			ObjectMeta: metav1.ObjectMeta{Name: "test"},
-			Spec: FleetAutoscalerSpec{
-				FleetName: "testing",
-				Policy: FleetAutoscalerPolicy{
-					Type: BufferPolicyType,
-					Buffer: &BufferPolicy{
-						BufferSize:  intstr.FromInt(1),
-						MaxReplicas: 10,
-					},
-				},
-			},
-		}
-		fasCopy := fas.DeepCopy()
-		fasCopy.Spec.FleetName = "notthesame"
-
-		causes := fas.ValidateUpdate(fasCopy, nil)
-
-		assert.Len(t, causes, 1)
-		assert.Equal(t, "fleetName", causes[0].Field)
-	})
-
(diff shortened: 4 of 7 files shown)


pkg/apis/stable/v1alpha1/fleetautoscaler.go
```diff
@@ -102,37 +102,17 @@ type FleetAutoscalerStatus struct {
...
+func (fas *FleetAutoscaler) Validate(causes []metav1.StatusCause) []metav1.StatusCause {
 	if fas.Spec.Policy.Type == BufferPolicyType {
 		causes = fas.Spec.Policy.Buffer.ValidateAutoScalingBufferPolicy(causes)
 	}
 	return causes
 }
 
-//ValidateAutoScalingSettings validates the FleetAutoscaler Buffer policy settings
+//Validate validates the FleetAutoscaler Buffer policy settings
 func (b *BufferPolicy) ValidateAutoScalingBufferPolicy(causes []metav1.StatusCause) []metav1.StatusCause {
```
EricFortin: I think you meant to change the name of the function too.

markmandel: Actually no - but that comment shouldn't be changed. Weird the linter didn't pick that up?

markmandel: Fixed comment. Intellij got a little aggressive with an automatic refactor. Thanks for the pickup!

markmandel: Or actually, are you saying this should be something like `ValidateBufferPolicy` ? (That would make more sense)

pkg/apis/stable/v1alpha1/fleetautoscaler.go
```diff
@@ -105,30 +105,10 @@ type FleetAutoscalerStatus struct {
 	// lastScaleTime is the last time the FleetAutoscaler scaled the attached fleet,
 	// +optional
 	LastScaleTime *metav1.Time `json:"lastScaleTime"`
-
-	// AbleToScale indicates that we can access the target fleet
-	AbleToScale bool `json:"ableToScale"`
```
victor-prodan: Why did you remove these flags?
I thought they should be pretty helpful for monitoring autoscaling state

markmandel: This is why I wanted you to look :smile: I made recorded events for these items, so that we could see how often they happened, have a timestamp for them, make it easy to see it in the `kubectl describe history` - and these events can also be used to power stats as well (they are globally collectable).

An example probably will help, so here is a FAS, which initially started with a Fleet that couldn't be found, and then was hitting the max limits for the autoscalers. You can see how often FailedGetFleet happened, and over what period, as well as when the last one was.

```
root@7ea86f2112be:/go/src/agones.dev/agones# kubectl describe fas
Name:         simple-udp-autoscaler
...
Spec:
  Fleet Name:  simple-udp
  Policy:
    Buffer:
      Buffer Size:   2
      Max Replicas:  6
      Min Replicas:  0
    Type:            Buffer
Status:
  Current Replicas:  5
  Desired Replicas:  6
  Last Scale Time:   2018-11-16T20:24:57Z
Events:
  Type     Reason            Age              From                        Message
  ----     ------            ----             ----                        -------
  Warning  FailedGetFleet    2m (x2 over 3m)  fleetautoscaler-controller  could not fetch fleet: simple-udp
  Normal   AutoScalingFleet  1m               fleetautoscaler-controller  Scaling fleet simple-udp from 2 to 3
  Normal   AutoScalingFleet  1m               fleetautoscaler-controller  Scaling fleet simple-udp from 3 to 5
  Normal   AutoScalingFleet  25s              fleetautoscaler-controller  Scaling fleet simple-udp from 5 to 6
  Warning  ScalingLimited    25s              fleetautoscaler-controller  Scaling fleet simple-udp was limited to maximum size of 6
```
Once i did this, the `AbleToScale` and `ScalingLimited` seemed redundant to me, so I removed them, as I didn't think they were adding any value at that point.

But like I said, I took some liberties :smile: , mainly as a sacrificial draft, to get the feedback to see if it made sense in the grand scheme, and be able to show an example of what I was thinking.

Does that make sense? Would love to hear your thoughts.

victor-prodan: I added those two flags following the example of HPA.

I personally believe that it makes sense for those to be part of status because... They are describing the current status.

The event log has a different purpose imo, to show the history.

I can also argue that If we follow this events logic than why have any status members at all? You can retrieve all of them from events anyway 😋

markmandel: That's a totally fair point, although I don't see them here:
https://github.com/kubernetes/api/blob/master/autoscaling/v1/types.go#L53

There really isn't any downside in having both. I'll put them back in then! :+1:

markmandel: Just digging deeper to see if maybe there is a better pattern - I see the above values (`AbleToScale`, `ScalingLimited`) in `HorizontalPodAutoscalerCondition` - but can't find if it's used anywhere. 

Pods do have a `[]PodCondition` in `PodStatus` - maybe the plan was to do something like that, but it never eventuated. This is probably overkill for our FAS though. WDYT?

victor-prodan: Look at HorizontalPodAutoscalerConditionType 😋

It's a slightly different implemention pattern. My go was not good enough, so I simplified things a bit 😁

markmandel: Yep - looks like we are in furious agreement 👍😀😁

I'll put it back and in and add the tests back. Be good to get this in for 0.6.0.rc next week.

victor-prodan: Furious agreement 🤣🤣🤣

markmandel: Done!

pkg/apis/stable/v1alpha1/fleetautoscaler.go
markmandel: SGTM! Will make the change. Thanks for the quick response.

Issue #414, opened 2018-11-12
CPU/MEMORY leak in agones controller container


Oleksii-Terekhov:
This is only a suspicion, but after restart agones-controller pod CPU usage constantly increased. This behavior independent to current count of fleets/gs/autoscalest. 

Looks like any fleet/gs, even after deleting, eat CPU(need garbage collector in data structs???).

![040](https://user-images.githubusercontent.com/13468544/48336013-033b3800-e668-11e8-94a0-953bc28fb0ae.png)
![mixed](https://user-images.githubusercontent.com/13468544/48336014-033b3800-e668-11e8-91f8-d0b1ae882bf0.png)

 
Reproduced with agones v0.4.0 and v0.5.0

markmandel:
Thanks for sharing this, and taking the time to graph it out. Do you have any specific steps we could we use to reproduce this? Looks like it happens over a few days?

Oleksii-Terekhov:
Nothing special - start/stop fleets
Up to 40 fleets with 5 gs on v0.4.0
Up to 40 fleets with autoscaler(min:0 buffer:2 max:10) v0.5.0
Agerage deploy rate for each fleet: ~4-5 per day

Main concern - heavy CPU usage after remove all fleets/gs/autoscalers

Oleksii-Terekhov:
STR:

0. empty k8s cluster (dashboard, agones, nginx-ingress, istio), 3 master nodes, 3 worker nodes, 8 Cores, Ram 16G 
1. Create easy start-wait-stop shell file
```
$ cat loop.sh
while :
do

    kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/agones/master/examples/simple-udp/fleet.yaml
    sleep 5
    kubectl get fleet
    sleep 5
    kubectl get gs -o=custom-columns=NAME:.metadata.name,STATUS:.status.state,CONTAINERPORT:.spec.ports[0].containerPort
    sleep 15
    kubectl delete fleet simple-udp
    sleep 5
done
```
2. restart controller pod
3. wait for 2 hours
4. start loop.sh fir 2 hours
5. stop loop.sh + delete all fleets
6. wait for 1 hour

Results:
```
Wall clock: 10:00 CPU: 0.008m, RAM: 14M - restart controller
Wall clock: 12:00 CPU: 0.009m, RAM: 15M - start loop.sh
Wall clock: 14:00 CPU: 0.447m, RAM: 26M - stop all after 2h burn fleets
[...]

PR #422, opened 2018-11-15
Development tools to enable pprof

markmandel:
Tools for building a controller with pprof enabed, and Make targets for accessing the http pprof endpoint while it is running on a Kubernetes cluster, with accompanying documentation.

This will help in diagnosing #414


markmandel:
@enocom does this look a good way to do this? Is there a better way?

PR #427, opened 2018-11-27
add fleet name to gameservers labels

### LINKED ISSUE #411: `GameServers` should have the Fleet name in a label for easy retrieval
Especially for those building their own integrations with Agones - its difficult to traverse `Fleet` -> `GameServerSet` -> `GameServers` if they want to get a list of `GameServers` for a given `Fleet`

So I suggest we add the fleet name as a label and value to the `GameServer` to make this easier.

Basically we should replicate what we do here for `GameServerSets` (although we may want to refactor some names):
https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/apis/stable/v1alpha1/fleet.go#L102

This will also mean that we can do a more efficient version of `ListGameServersByFleetOwner`:
https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/fleets/fleets.go#L48
as well.

Kuqd:
This adds fleet name to gameservers, I've renamed the label to be more general.

I don't think I've forgot anything let me know. I needed this also for metrics.

Closes #411

### DIFF
--- a/pkg/apis/stable/v1alpha1/fleet.go
+++ b/pkg/apis/stable/v1alpha1/fleet.go
@@ -22,9 +22,9 @@ import (
 )
 
 const (
-	// FleetGameServerSetLabel is the label that the name of the Fleet
-	// is set to on the GameServerSet the Fleet controls
-	FleetGameServerSetLabel = stable.GroupName + "/fleet"
+	// FleetNameLabel is the label that the name of the Fleet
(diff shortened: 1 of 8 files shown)


markmandel:
If you need this urgently, please feel free to merge - i wouldn't mind having #430 merge first, just for ordering, but I'm not strongly wedded to it.

PR #437, opened 2018-12-02
Delete crds, and fleets, gameservers etc on deletion of Helm chart

### LINKED ISSUE #426: Delete crds, and fleets, gameservers etc on deletion of Helm chart
Due to #375 - the crds are not managed by Helm, so they are left behind on deletion of the chart.

Also, if you delete the helm chart before removing your `Fleets` and `GameServers`, you can get stuck GameServers, and it can be a pain to cleanup.

We should write a `post-delete` hook (or whatever is appropriate) to do the cleanup when deleting the Helm chart to clean all the things up.

Context: https://github.com/helm/helm/blob/master/docs/charts_hooks.md#hook-resources-are-not-managed-with-corresponding-releases

EricFortin:
This implements helm hooks to delete all Agones resources and CRDs when uninstalling our chart.

Deletion of resources is done in a pre-delete hook while crds are deleted in a post-delete hook. Had to add a new service account and a few other resources to accomplish it.

Fixes #426

### DIFF
--- a/install/helm/agones/scripts/delete_agones_resources.sh
+++ b/install/helm/agones/scripts/delete_agones_resources.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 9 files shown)


install/helm/agones/templates/hooks/post_delete_hook.yaml
```diff
@@ -0,0 +1,49 @@
...
+        image: "lachlanevenson/k8s-kubectl:v1.10.9"
+        command: ["/bin/sh", "/scripts/delete_crds.sh"]
+        volumeMounts:
+        - name: script
+          mountPath: /scripts/
+      volumes:
+      - name: script
+        configMap:
+          name: delete-crds
+{{- end }}
```
Kuqd: nitpick: I think there is some missing carrier return in some file.

Kuqd:
You need to rebase and squash btw.

install/helm/agones/scripts/delete_agones_resources.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
markmandel: This file needs a licence header 👍

PR #440, opened 2018-12-07
Adding explicit length of git revision in Makefile and E2E Can't Allocate test 

aLekSer:
Output of git rev-parse short command could be different for different environments. For example on my laptop I got 8 chars which leads to errors on deployment to test-cluster using make install.
Namely:
```
ImagePullBackOff     agones-controller: gcr.io/agones-images/agones-controller:0.7.0-5e2ad5ee 
```
which should be gcr.io/agones-images/agones-controller:0.7.0-5e2ad5e.
Also adding separate test for "Can't allocate on fully sized fleet".

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -35,7 +35,7 @@ release_registry = gcr.io/agones-images
 #
 
 # Version defaults to the short hash of the latest commit
-VERSION ?= $(base_version)-$(shell git rev-parse --short HEAD)
+VERSION ?= $(base_version)-$(shell git rev-parse --short=7 HEAD)
 # The registry that is being used to store docker images
 REGISTRY ?= $(release_registry)
 # kubectl configuration to use
--- a/test/e2e/fleet_test.go
+++ b/test/e2e/fleet_test.go
@@ -24,16 +24,17 @@ import (
 	e2e "agones.dev/agones/test/e2e/framework"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
-	"k8s.io/api/apps/v1"
+	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


build/Makefile
```diff
@@ -35,7 +35,7 @@ release_registry = gcr.io/agones-images
 #
 
 # Version defaults to the short hash of the latest commit
-VERSION ?= $(base_version)-$(shell git rev-parse --short HEAD)
```
markmandel: This is good catch! Should this be in a separate PR? Happy to have it in 1, but we usually squash down to a single commit for each PR.

test/e2e/fleet_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	e2e "agones.dev/agones/test/e2e/framework"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
-	"k8s.io/api/apps/v1"
+	v1 "k8s.io/api/apps/v1"
```
markmandel: Minor thing - either we should drop the `v1` import alias (as it's the same thing), or let's make it `appsv1` (which I'm pretty sure what we call it elsewhere)

test/e2e/fleet_test.go
```diff
@@ -318,14 +366,16 @@ func scaleFleet(f *v1alpha1.Fleet, scale int32) (*v1alpha1.Fleet, error) {
 	return framework.AgonesClient.StableV1alpha1().Fleets(defaultNs).Patch(f.ObjectMeta.Name, types.JSONPatchType, []byte(patch))
 }
 
+const replicasCount = 3
```
markmandel: Let's move this up to the top with the rest of the consts - that's the usual go way.

test/e2e/fleet_test.go
aLekSer: This is VS Code thing. It was added automatically. Agree that appsv1 feets more natural.

PR #441, opened 2018-12-08
Remove crd-install hook, as it break CRD updates

markmandel:
This is a problem that Helm is going to solve going forward, but for now, if you use the crd-install hook, then you can only install CRDs, and not update them at any point during a chart lifecycle.

Also, prior to Helm 2.12, if you installed chart with a crd-install hook that did not have one previously, it deleted the CRDs.

Therefore, removing the crd-install hook, so that CRDs are again managed by the Helm charts.

Added a `agones.crd.install` parameter, in case someone wants to subchart this chart, then can set this to false, and copy the Agones CRDs into their own charts to be included in the right place for their chart lifecycle.

Also, since we have a `agones.crd` config section, moved `agones.enableHelmCleanupHooks` into `agones.crds.cleanupOnDelete`

Unfortunately, with this back and forth on the crd-install hook, if you are using the Helm chart, you will need to do a full Agones `helm delete --purge` and cleanup any remaining CRDs to upgrade.

More context on helm + crds:
- https://github.com/helm/helm/issues/4697
- https://github.com/istio/istio/issues/9604
- https://github.com/istio/istio/issues/7688
- https://github.com/helm/community/pull/64
- https://github.com/helm/helm/issues/4863
- https://github.com/helm/helm/pull/4709


smoya:
Thanks for taking care @markmandel ! And thanks for mentioning me as well. Good solution btw.

sdake:
@markmandel note Istio came up with a slightly different solution:

See: https://github.com/istio/istio/pull/10562

And specifically `tiller-hack` here:
https://github.com/istio/tools/pull/31

This tool works around the problem of CRDs being deleted.  Note, this tool is specific to Istio, but the solution could be generalized.

Just thought I'd share.

Cheers
-steve

markmandel:
@sdake I'm struggling to work out what they are doing here. Can you give a quick summary?

PR #442, opened 2018-12-10
GameServerSets: DeleteFunc could receive a DeletedFinalStateUnknown

markmandel:
This should have a check on it anyway, but I went hunting as I was wondering if this might be the cause for #414


markmandel:
/cc @Oleksii-Terekhov @aLekSer - don't know if this is the problem, but it _might_ be.

Was thinking about instances where things could go into perpetual eventual backoff, and thought of DeleteFunc's - and there was one instance where the type check was not in place.

Not 100% convinced this is the leak, because if this part of code failed, it would have panicked, which I would have thought we should have seen in the logs, or we should have seen the controller restart. But worth a shot - I could be wrong.

PR #444, opened 2018-12-13
CRD: added additionalPrinterColumns to GameServer for kubectl

jkowalski:
This only works on Kubernetes >=1.11 and is ignored by older versions.

```
$ kubectl get gameservers
NAME                     STATE       ADDRESS        PORT   NODE                                     AGE
simple-udp-lrw4k-6c8rg   Ready       35.247.xx.xx   7741   gke-test-cluster-default-405277ca-2fr0   15m
simple-udp-lrw4k-78td7   Ready       35.247.xx.xx   7378   gke-test-cluster-default-405277ca-2fr0   15m
simple-udp-lrw4k-8mcwk   Ready       35.247.xx.xx   7795   gke-test-cluster-default-405277ca-2fr0   15m
simple-udp-lrw4k-wqwsn   Allocated   35.247.xx.xx   7301   gke-test-cluster-default-405277ca-2fr0   15m
simple-udp-lrw4k-xxj2f   Ready       35.247.xx.xx   7052   gke-test-cluster-default-405277ca-2fr0   15m

$ kubectl get gameserversets
NAME               SCHEDULING   DESIRED   CURRENT   ALLOCATED   READY   AGE
simple-udp-lrw4k   Packed       10        10        1           9       16m

$ kubectl get fleets
NAME         SCHEDULING   DESIRED   CURRENT   ALLOCATED   READY   AGE
simple-udp   Packed       10        10        1           9       16m

$ kubectl get fleetallocations
NAME               STATE       ADDRESS        PORT   NODE                                     AGE
simple-udp-sf66j   Allocated   35.247.xx.xx   7301   gke-test-cluster-default-405277ca-2fr0   16m

### DIFF
--- a/install/helm/agones/templates/crds/fleetallocation.yaml
+++ b/install/helm/agones/templates/crds/fleetallocation.yaml
@@ -25,6 +25,22 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.gameServer.status.state
+    name: State
+    type: string
+  - JSONPath: .status.gameServer.status.address
+    name: Address
+    type: string
+  - JSONPath: .status.gameServer.status.ports[0].port
+    name: Port
+    type: string
+  - JSONPath: .status.gameServer.status.nodeName
+    name: Node
+    type: string
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
   group: stable.agones.dev
   version: v1alpha1
   scope: Namespaced
--- a/install/helm/agones/templates/crds/fleet.yaml
+++ b/install/helm/agones/templates/crds/fleet.yaml
@@ -25,6 +25,25 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .spec.scheduling
+    name: Scheduling
+    type: string
+  - JSONPath: .spec.replicas
+    name: Desired
+    type: integer
+  - JSONPath: .status.replicas
+    name: Current
+    type: integer
+  - JSONPath: .status.allocatedReplicas
+    name: Allocated
+    type: integer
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: install/helm/agones/templates/crds/fleetallocation.yaml [hunk 1]
```diff
@@ -25,6 +25,22 @@ metadata:
...
     heritage: {{ .Release.Service }}
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.gameServer.status.state
+    name: State
+    type: string
+  - JSONPath: .status.gameServer.status.address
+    name: Address
+    type: string
+  - JSONPath: .status.gameServer.status.ports[?(@.name=='default')].port
```
markmandel: What happens if the user doesn't have a "default" port - does nothing show?

jkowalski: That's correct.

markmandel: I'm wondering if we don't want to show the entire map (probably too wide?) - should we grab the first port in the list, rather than one with the name "default" - just in case? That way we always get a record?

@Kuqd what do you think?

jkowalski: Can we require GS to always have a 'default' port?

Kuqd: I'm also for either the first one or all of them. Don't have a strong opinions, if it's easier the first one would do for now.

markmandel: How about we start with the first one for now - as that is the most simple, and doesn't break anything for anyone. If we find that people really want all ports later on, we can adjust it, or see if there is a way to make it work in a `o  -wide` scenario.

Also, reorganising the order of ports in the list for output display is far less brittle for users, than forcing people into a naming convention that may not work with their systems.

WDYT?

jkowalski: Updated GS and FleetAllocation CRD to ports[0]

PR #447, opened 2018-12-17
Update to Kubernetes 1.11

markmandel:
Now we have >= 1.11 support across cloud providers, we can finally update client-go to provide extra features.

This updated client-go to 8.0.0 this includes Status API on the clients so we can take advantage of it - which should ideally give us a performance boost.

*Please note:* this will mean that our required Kubernetes version will now be 1.11+, I think this is acceptable now it has penetration across cloud providers, but feedback appreciated.

Also took the opportunity to upgrade logrus, errors, testify as well.

This makes #329 possible.


markmandel:
Things the move to 1.11 will make possible across the board:
- https://github.com/kubernetes-sigs/kind
- #444 
- probably other things I'll think of later ;)

Kuqd:
LGTM!

Issue #453, opened 2018-12-21
Improve build speed by refactoring Makefile


jkowalski:
There are currently two issues with Makefile which prevent it from being parallelizable:

1. Some dependencies are missing in the Makefile which prevent it from working with -j N where N > 0. 

For example if I do:

```
$ make -j2 build build-controller-binary push-controller-image
```
It will likely fail or be incorrect (because build and push will be running in parallel, it will either fail to push or push a previous version). A fix for this is to add a dependency on build to push. There are all kinds of issues like that in the Makefile.

Ideally I should be able to say `export MAKEFLAGS="-j 4"` in my `~/.bashrc` and my builds would be magically faster, but still correct regardless of targets that I specify.

2. Some targets (e.g. `build-agones-sdk-binary`) are sequential, and pretty slow (in particular when building on macOS, which has horrible filesystem mount performance). They might benefit from being broken into smaller pieces which can run in paralllel.

jkowalski:
I can work on fixing that. 

Also I discovered that we're not using any go build caching, which kills interactive (re)build performance. Will send a PR which fixes that...

markmandel:
> Also I discovered that we're not using any go build caching, which kills interactive (re)build performance. Will send a PR which fixes that...

YEESSSS - I've wanted this for a while, but never got around to it! :man_cartwheeling:

Kuqd:
> Some targets (e.g. build-agones-sdk-binary) are sequential, and pretty slow (in particular when building on macOS, which has horrible filesystem mount performance). They might benefit from being broken into smaller pieces which can run in paralllel.

YES !  building on macOS is slow.

wilsonwang371:
It looks like building agones on Linux still has problems when building with -j. Is there any bug tracking this issue?

PR #454, opened 2018-12-21
added incremental build option to Makefile to speed up rebuilds

jkowalski:
This partially addresses #453.

Before:
=======

```
$ time make -C build -j4 build-images
real    2m33.772s
user    0m1.302s
sys     0m0.889s
```
Incremental re-builds:
=======

```
$ time make -C build -j4 build-controller-image
real    0m1.837s
user    0m0.241s
sys     0m0.085s

$ time make -C build -j4 build-agones-sdk-image
real    0m5.352s
user    0m0.447s
sys     0m0.244s

$ time make -C build -j4 build-ping-image
real    0m1.318s
user    0m0.198s
sys     0m0.063s

$ time make -C build -j4 build-images INCREMENTAL_BUILD=1
real    0m5.506s
user    0m1.177s
sys     0m0.617s
```

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -101,6 +101,21 @@ ifdef DOCKER_RUN
 	ensure-build-image += ensure-build-image
 endif
 
+# When performing full build, always rebuild go packages and compress more
+ifdef FULL_BUILD
+	go_rebuild_flags = -a
+	zip_flags = -9
+endif
+
+ifndef FULL_BUILD
+	# keep a cache of files built by Go across docker invocations in a local directory.
+	common_mounts += -v $(build_path)/.gocache:/root/.cache/go-build
+	go_rebuild_flags =
+
+	# do not compress files in zips, slightly faster
+	zip_flags = -0
+endif
+
 #   _____                    _
 #  |_   _|_ _ _ __ __ _  ___| |_ ___
 #    | |/ _` | '__/ _` |/ _ \ __/ __|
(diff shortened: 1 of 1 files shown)


build/Makefile
```diff
@@ -101,6 +106,13 @@ ifdef DOCKER_RUN
 	ensure-build-image += ensure-build-image
 endif
 
+ifdef INCREMENTAL_BUILD
```
markmandel: I'm thinking that this should be defaulted on - and then turned off for builds in `do-release`. I'm not seeing any downsides. WDYT?

build/Makefile
```diff
@@ -101,6 +106,13 @@ ifdef DOCKER_RUN
...
+	# keep a cache of files built by Go across docker invocations in a local directory.
+	common_mounts += -v $(CURDIR)/.gocache:/root/.cache/go-build
```
markmandel: Rather than use CURDIR, which is the working directory, let's explicitly put it in the $(build_path), then it never moves, regardless of where people call the Makefile from.

```suggestion
 	common_mounts += -v $(build_path)/.gocache:/root/.cache/go-build
```

Then we can also add a `mkdir -p` to the `ensure-build-config` target, so that the directory always is created before being mounted (I thought it should throw an error if it's not there - has that stopped happening?)

jkowalski: Docker appears to be creating the directory automatically.

build/Makefile
jkowalski: switched default to incremental.

markmandel:
:fire: WOW THAT IS FAST NOW :fire:

Issue #144, opened 2018-03-16
Statistics collection and display


markmandel:
Stats around game server performance, as well as in Game statistics need to be easily exportable from the GameServer into Agon through the SDK and go to some kind of statistic data store (such as [Prometheus](https://prometheus.io) either directly, or through [OpenCensus](https://opencensus.io)). There will also need to be some kind of visualisation layer, and possibly alerting as well.

Discussion will need to cover (a) what types of stats they would want to capture (b) the SDK API to collect them, and (c) the visualisation and/or alerting requirements

# Statistics Ideas
- Number of game servers / ready / allocated / unhealthy / etc
- Game server CPU / Memory - aggregate and per game server
- Game server Frame rates
- Game server hitching
- Scale up time of nodes, and corresponding node resource sizes (we'll need this for node autoscaling)
- Visualization of game servers/nodes/vs current trend up/down scaling rates

# Research
- https://github.com/improbable-eng/thanos

kwiesmueller:
Wouldn't it be best to really use Prometheus directly and then us the included service discovery to ensure metrics from all nodes get scraped?
I can't really tell a reason why there should even be some way to send metrics through agones as it would only implement more complexity than necessary.
The only thing I could imagine is the need for access to the metrics from inside some components (including game servers) which could be solved as well.

For instance, there would already be a quite good integration with GCP and Stackdriver as well and as far as I know Prometheus is quite widely adopted too.

markmandel:
Coming back to this - there are some very widely used statistics that we should provide out of the box - it seems like duplication of work to have everyone do this themselves. We will also need some standard visualisations of what is happening inside Agones as well.

That being said (and one of the reasons I like OpenCensus) - if there is a way we can expose/leverage a project like that to allow stats to go anywhere without blocking stats collection.

Basically find a way to make the best of both worlds.

GabeBigBoxVR:
Wow, this would be amazing, especially for FPS games where tick rate is super important.

Kuqd:
I'm going to start things off very simple using opencensus and prometheus in the controller, then we can build from there and add more metrics.
I'll start with : Number of game servers / ready / allocated / unhealthy

markmandel:
@heartrobotninja just checking in - did you have any work here that @Kuqd could leverage?

Kuqd:
I'm planning to collect also:
- go metrics
- healtheck metrics
- process metrics

Kuqd:
Hey Guys,

[...]

PR #457, opened 2018-12-25
Adds OpenCensus metrics integration.

Kuqd:
Merry Christmas Agones,

This is my present :santa: :christmas_tree: :gift: .

This PR moves forward #144 

- First set of state metrics added using informers cache.
- Added flag for switching on/off metrics via helm and env vars.
- Added documentation section about metrics and usage.
- Go runtime metrics and Process Metrics are also included.
- controller healthchecks metrics.

Next steps should be:

- Create a template dashboard for users.
- Add more metrics (allocations per nodes, allocations success/failure rate)


cmd/controller/main.go
```diff
@@ -112,15 +136,19 @@ func main() {
...
 	stop := signals.NewStopChannel()
 
 	kubeInformationFactory.Start(stop)
 	agonesInformerFactory.Start(stop)
 
-	rs := []runner{
-		wh, gsController, gsSetController, fleetController, faController, fasController, healthServer{handler: health},
-	}
 	for _, r := range rs {
+		if r == nil {
```
markmandel: Curious - what is this for?

pkg/metrics/controller.go
```diff
@@ -0,0 +1,369 @@
...
+	if oldFa.Status.GameServer == nil && newFa.Status.GameServer != nil {
+		c.recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyFleetName, newFa.Spec.FleetName)},
+			fleetAllocationTotalStats.M(1))
+	}
+}
+
+// Run the Metrics controller. Will block until stop is closed.
+// Collect metrics via cache changes and parse the cache periodically to record resource counts.
+func (c *Controller) Run(workers int, stop <-chan struct{}) error {
+	err := crd.WaitForEstablishedCRD(c.crdGetter, "fleets.stable.agones.dev", c.logger)
```
markmandel: Since we already do this in the fleet controller (check for the established crd), should we remove it from here? We likely don't need to do this twice.

cmd/controller/main.go
Kuqd: Metrics controller can be nil, if it's switche off

pkg/metrics/controller.go
Kuqd: true ! I should get rid of this

Kuqd:
my tests are generating a race, I see how I can fix that but it got me thinking. Should I always `DeepClone` in event handler ?

pkg/metrics/gameservers.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, fleets := range c {
+		for fleet := range fleets {
+			fleets[fleet] = 0
+		}
+	}
+}
+
+// record counts the list of gameserver per status and fleet name and record it to OpenCensus
+func (c GameServerCount) record(gameservers []*stablev1alpha1.GameServer) error {
+	// Currently there is no way to remove a metric so we have to reset our values to zero
```
Kuqd: @markmandel Can we check with OpenCensus team if there is a way to tackle this ? I found this annoying but I understand why it happens. 

With Prometheus not recording a metrics with tags at scrape time, considered  it deleted/missing.

May be we could have a way to reset a gauge ? or deleted it ?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,130 @@
...
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+var (
+	keyName, _      = tag.NewKey("name")
+	keyFleetName, _ = tag.NewKey("fleet_name")
+	keyType, _      = tag.NewKey("type")
+
+	fleetsReplicasCountStats = stats.Int64("fleets/replicas_count", "The count of replicas per fleet", "1")
```
Kuqd: I hesitated a lot between global scope and controller variable, in the end it's easier to deal with global scope but also cleaner to arrange and declare.

Couldn't find any example not using global scope for OpenCensus, are we ok with this ?

@markmandel @enocom

Kuqd: I also like having all views/stats declaration in one file.

Kuqd: Good idea, so you’re fine with the rest of this file ?

pkg/metrics/gameservers.go
Kuqd: Ok I’ll report it, FYI I’m counting gameservers per fleet name and status in a map, then going through the map to record each results, the problem is that if there is no more gameservers for a given fleet I have to reset the value to zero but the stats stays, so it’s ever growing, may be unregister/register the view each time.

markmandel: I figure asking on the Gitter is the best bet - otherwise, I've no idea what I'm asking :smile:

pkg/metrics/metrics.go
markmandel: I'm not seeing any issues here I don't think. Is this data mutated, or is it just a declaration? Especially if it's just a declaration, I think it makes perfect sense. (I think it's just a declaration, if I read it correctly?)

Kuqd: Yes it’s declaration only

pkg/metrics/gameservers.go
Kuqd: went for an issue instead - https://github.com/census-instrumentation/opencensus-go/issues/1003

I think it's easier to keep track of.

cmd/controller/main.go
markmandel: Random thought now that I stare at this some more - it _could_ be:

```go
	for _, r := range rs {
		if r != nil {
			go func(rr runner) {
				if runErr := rr.Run(workers, stop); runErr != nil {
					logger.WithError(runErr).Fatalf("could not start runner: %s", reflect.TypeOf(rr))
				}
			}(r)
		}
	}
```
Not sure which is more readable, but figured it might be more worth considering, in case you hadn't.

Kuqd: I considered it, I thought my solution created less code indentation, but it seems confusing for you ? Do you want me to rework using your solution ?

markmandel: it took me a second to resolve in my head - but I don't have strong opinions either way. I don't think it's worth blocking on, just wanted to consider it.

pkg/metrics/gameservers.go
markmandel: Do you want to block on this until we get an answer from the OpenCensus team, or shall we iterate as we get more information?

Kuqd: I don’t expect a solution by 0.7 so let’s go with what we have.

install/helm/agones/templates/service.yaml
```diff
@@ -28,4 +28,6 @@ spec:
     stable.agones.dev/role: controller
   ports:
     - port: 443
-      targetPort: 8081
\ No newline at end of file
+      targetPort: 8081
```
markmandel: This needs a name - otherwise:

```
Error: release agones failed: Service "agones-controller-service" is invalid: spec.ports[0].name: Required value
Makefile:195: recipe for target 'install' failed
make: *** [install] Error 1
```

:crying_cat_face:

build/helm.yaml
```diff
@@ -31,4 +31,23 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: helm
-    namespace: kube-system
\ No newline at end of file
+    namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
markmandel: Curious - what is this for?

Or at least making sure this was deliberate :smiley:

markmandel:
Yeeps - yep, e2e actually didn't fail when it was supposed to!

```
# if IMAGE_PULL_SECRET_FILE is specified, create the agones-system namespace and install the secret
\
helm upgrade --install --wait --namespace=agones-system\
--set agones.image.tag=0.7.0-67a7c71,agones.image.registry=gcr.io/agones-images \
--set agones.image.controller.pullPolicy="Always",agones.image.sdk.alwaysPull=true \
--set agones.image.controller.pullSecret= \
Error: UPGRADE FAILED: Service "agones-controller-service" is invalid: spec.ports[0].name: Required value
Makefile:195: recipe for target 'install' failed
starting e2e test
```

install/helm/agones/templates/service.yaml
Kuqd: TIL: port needs name if there are multiple of them.

docs/metrics.md
```diff
@@ -0,0 +1,78 @@
...
+  endpoints:
+  - port: web
+```
+
+Finally include that `ServiceMonitor` in your [Prometheus instance CRD](https://github.com/coreos/prometheus-operator/blob/v0.17.0/Documentation/user-guides/getting-started.md#include-servicemonitors), this is usually done by adding a label to the `ServiceMonitor` above that is matched by the prometheus instance of your choice.
+
+### Stackdriver
+
+We don't yet support the [OpenCensus Stackdriver exporter](https://opencensus.io/exporters/supported-exporters/go/stackdriver/) but you can still use the Prometheus Stackdriver integration by following these [instructions](https://cloud.google.com/monitoring/kubernetes-engine/prometheus).
+Annotations required by this integration can be activated by setting the `agones.metrics.prometheusServiceDiscovery` to true (default) via the helm chart value.
```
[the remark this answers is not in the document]
Kuqd: done

install/helm/agones/templates/service.yaml
Kuqd: fixed

PR #459, opened 2018-12-26
Upgrade minikube to K8s 1.11

markmandel:
Whoops, missed this change in the 1.11 upgrade.


markmandel:
Ah - I think that's a Linux/OSX differencec. @EricFortin had some hack for setting the ip the same across minikube and the local machine (or something?)? Maybe we should bake that into the osx version of `minikube-test-cluster` ?

Issue #134, opened 2018-03-08
C++ SDK Cross compilation and testing


markmandel:
## Problem 

Our current support for low dependency, cross compilation for the C++ SDK is very poor. Things work on Linux, because of `make` and not much else. We also have no tests.

Right now, the C++ SDK is built on top of [gRPC](https://grpc.io/), which may be adding too many dependencies to be used in a valuable way across platforms?

Code is here: https://github.com/GoogleCloudPlatform/agones/tree/master/sdks/cpp

## Requirements
1. Review the [current state of cross complication and dependency with gRPC](https://github.com/grpc/grpc/blob/master/src/cpp/README.md#make). 
1. Review the SDK [REST api endpoints](https://github.com/GoogleCloudPlatform/agones/blob/master/docs/sdk_rest_api.md), and see if that would be a better fit than connecting via gRPC.
1. Depending on that, prepare a design doc (added here is fine), with the following things in mind:
   1. Ability to cross compile on Linux/Mac/Windows
   1. Add some kind of unit tests, build tests, etc ideally with automation
   1. Anything else that is important to the C++ ecosystem (dependencies?)

Caveat: This is written by @markmandel who has little to no idea about C++ and its ecosystem, so direction on the above is also appreciated.

## Research
https://github.com/grpc/grpc/blob/master/src/cpp/README.md#make
https://www.appveyor.com
https://circleci.com/build-environments/xcode/osx/
https://docs.travis-ci.com/user/reference/osx/

[History](https://gist.github.com/markmandel/7e0adbb8141d1c862acba14bad63ed0d)

SurferJeffAtGoogle:
[CMake](https://cmake.org/) is the defacto standard for building C and C++ code across platforms.

markmandel:
Totally happy to have this rewritten with CMake (or whatever is appropriate) - I totally bashed this together with the help of someone with (somewhat?) more C++ experience than me, and this is an ecosystem I am very much not familiar with.

dzlier-gcp:
[Notes doc](https://docs.google.com/document/d/1O7Je4cXMyNdUl_M28Qq3dcrWClHy3Vm-hdjglmsvyN4/edit) after meeting with rennie@, markmandel@ and myself.

markmandel:
@martin-z-google This is the current state of the C++ SDK situation - please let me know if this makes no sense :smile:

ortho:
[...]

PR #464, opened 2018-12-28
Cpp sdk cmake

dsazonoff:
This changeset provides CMake scripts for building C++ SDK that works across Windows, Linux and macOS.
Old Makefile will be removed after migrating examples/cpp-simple to CMake.
Issue: #134


dsazonoff:
CLA signed, please review.

sdks/cpp/README.md
```diff
@@ -115,34 +115,49 @@ For more information, you can also read the [SDK Overview](../), check out [sdk.
...
-- `argonsdk-$(VERSION)-dev-linux-arch_64.tar.gz`: This includes all the 
-headers as well as dynamic and static libraries that are needed for development and runtime.
-- `argonsdk-$(VERSION)-runtime-linux-arch_64.tar.gz`: This includes just the dynamic libraries that 
-are needed to run a binary compiled against the SDK and its dependencies.
-
-### Building the Libraries
+## Building From Source
+To build Agones C++ SDK from sources you need to set up your development environment and install following tools:
+- CMake (https://cmake.org/download/)
+- OpenSSL (https://github.com/openssl/openssl)
```
markmandel: Why does the user need to install OpenSSL? We actually don't need SSL termination in our SDK, so if we can remove this, that would be ideal. In the previous incarnation, we just built out grpc++_unsecure

sdks/cpp/README.md
```diff
@@ -115,34 +115,49 @@ For more information, you can also read the [SDK Overview](../), check out [sdk.
...
+- Git (https://git-scm.com/downloads)
+- gRPC (https://github.com/grpc/grpc)
```
markmandel: What does installing gRPC mean here? 

The cmake installation should be as turnkey as possible. If they user has to manually install gRPC themselves,  this defeats the purpose of the tool we're trying to make here. The user really shouldn't care what the communication protocol is.

The previous incarnation did this really well:
https://github.com/GoogleCloudPlatform/agones/pull/421/files#diff-8ffc59e90c0bca72cf7710844b68352dR28

sdks/cpp/README.md
```diff
@@ -115,34 +115,49 @@ For more information, you can also read the [SDK Overview](../), check out [sdk.
...
 
-## Building From Source
-If you wish to compile from source, you will need to compile and install the following dependencies:
+On Windows, it is possible to use Chocolatey package manager (https://chocolatey.org/docs/installation) to install dependencies:
 
-- [gRPC](https://grpc.io), v1.8.x - [C++ compilation guide](https://github.com/grpc/grpc/blob/v1.8.x/INSTALL.md)
-- [protobuf](https://developers.google.com/protocol-buffers/), v3.5.0 - [C++ compilation guide](https://github.com/google/protobuf/blob/master/src/README.md)
+    choco install git cmake activeperl golang yasm
 
+It is possible to use your own build of gRPC. In this case you need to make your **gRPC** and **protobuf** installation visible to CMake, so `find_package` command can access all necessary cmake scripts.
```
markmandel: I don't think we really want users to be doing this. They have no need to - and it's going to be a source of pain for us to maintain. We should dictate the gRPC installation version, and it's install process, and manage it for the user.

To make things simpler for everyone, I think we should remove all customisation options for being able to provide your own gRPC. Our build system should manage this for the user end to end.

sdks/cpp/sdk.h
```diff
@@ -12,13 +12,15 @@
...
+#include "sdk_global.h"
+
 #include <grpcpp/grpcpp.h>
 #include "sdk.grpc.pb.h"
 
 namespace agones {
 
     // The Agones SDK
-    class SDK {
+    class AGONES_EXPORT SDK {
```
markmandel: Out of curiosity - what does this do?

sdks/cpp/README.md
```diff
@@ -115,34 +115,49 @@ For more information, you can also read the [SDK Overview](../), check out [sdk.
...
+To build Agones C++ SDK from sources you need to set up your development environment and install following tools:
+- CMake (https://cmake.org/download/)
+- OpenSSL (https://github.com/openssl/openssl)
+- Git (https://git-scm.com/downloads)
+- gRPC (https://github.com/grpc/grpc)
 
-If you want to build the libraries from Agones source, 
-the `make` target `build-sdk-cpp` will compile both static and dynamic libraries for Debian/Linux
-for your usage, to be found in the `bin` directory inside this one.
+To build gRPC you may also need to install next dependencies:
```
markmandel: The previous PR's documentation covered this really nicely, and specifically went through each dependency:
https://github.com/GoogleCloudPlatform/agones/pull/421/files#diff-eeb2d8fcc3782771feb2798fb61b5979R135

markmandel:
Apparently, I cc'd the wrong person :blush: 
/cc @sylvainduchesne - would you like to take a pass? I trust your C++ way more than mine.

sdks/cpp/README.md
sylvainduchesne: External dependencies can be kind of tricky. If a project already uses grpc (or other) you can't expect them to have 2 copies of the same library at different versions. For this reason, its generally better to allow for this customization.

sylvainduchesne: That's not to say that we cannot provide a version of grpc if one isn't specified however.

sdks/cpp/sdk.h
sylvainduchesne: This probably answers the question better than anything I could write:
https://stackoverflow.com/questions/8863193/what-does-declspecdllimport-really-mean

Basically, windows specific code for DLL usage.

sdks/cpp/README.md
sylvainduchesne: If I recall correctly, grpc doesn't offer much in terms of customization and it isn't technically possible to compile it without openssl.

sylvainduchesne:
I think the PR looks good, but I would like to see a proper cmake package so that agones could also be consumed as a cmake package. Would it be possible to add this to the PR?

sdks/cpp/README.md
markmandel: If I'm reading the previous PR correctly, it looks you can compile it without SSL. But maybe I'm reading it wrong? 🤷‍♂️

sdks/cpp/README.md
markmandel: Then we should definitely provide a default if one isn't provided 👍

I'm just very scared of trying to support someone running a random version of gRPC, and then expecting us to support it.

markmandel: Then we should definitely provide a default if one isn't provided 👍

I'm just very scared of trying to support someone running a random version of gRPC, and then expecting us to support it.

sdks/cpp/README.md
sylvainduchesne: Which previous PR? 
But if you look at grpc cmake https://github.com/grpc/grpc/blob/master/CMakeLists.txt#L141
and https://github.com/grpc/grpc/blob/master/cmake/ssl.cmake
there isn't really any way to turn it off.
GRPC cmake (I believe) is generated, and really hard to follow. 
It is probably possible to build specifically the `grpc_unsecure` target, and set `gRPC_SSL_PROVIDER`  to `module`. 

We really would need an automated cmake build for this repo though. A lot of surprises will pop up since its currently untested,

markmandel: This previous PR: https://github.com/GoogleCloudPlatform/agones/pull/421

> It is probably possible to build specifically the grpc_unsecure target, and set gRPC_SSL_PROVIDER to module

Can we do that :) ?

> We really would need an automated cmake build for this repo though. A lot of surprises will pop up since its currently untested,

Yes - agreed. We should integrate this PR into the cloudbuild system. Next steps should be to make some tests, and also setup CI with windows (AppVeyor?) and OSX (Travis I think does this?) as well.

sdks/cpp/README.md
dsazonoff: Projects may use gRPC for their own needs and in common way it is not possible to mix different versions of gRPC in one binary.
It is possible to do it only with some restrictions: only if SDK was built as dynamic library with statically linked gRPC. It means that client will use its own instance of gRPC, and SDK will use its own.
But it is a danger way, because both instances of gRPC (application's and SDK's) may share same kernel objects (named mutex, shared memory etc.) which may cause unexpected problems. To know for sure it is necessary to know gRPC internals in depth.
My suggestion is to provide a choice to an end user how to link with gRPC. But we can show a warning if user tries to use unexpected version of gRPC to build SDK. Or even to disable the build unless user explicitly state that he wants to use different version of gRPC. In this case we can support only exact version of gRPC; in other cases user will be responsible for possible problems.

sdks/cpp/README.md
dsazonoff: Cpp SDK depends on gRPC. Unfortunately, we can't simply install gRPC for C++, as [mentioned here](https://grpc.io/blog/installation). So we need to do a manual build of gRPC which depends on OpenSSL and so on.
I think we should do next steps, to build SDK and avoid OpenSSL dependency:
1) Download gRPC repo (if path to gRPC is not specified)
2) Build only `grpc_unsecure` target (gRPC cmake package scripts will not be generated)
3) Use our own find package script that will have all references to include/lib directories of gRPC
4) Build SDK
3rd step is necessary, so end user can link with SDK, gRPC and all other dependencies without hardcoding. We need to support both Debug/Release builds for IDE-s such XCode and MSVS.

@sylvainduchesne what do you think?

sdks/cpp/README.md
markmandel: > My suggestion is to provide a choice to an end user how to link with gRPC. But we can show a warning if user tries to use unexpected version of gRPC to build SDK. 

This sounds like the perfect option to me :+1: 

gRPC is generally SUPER backward compatible, but we should show a warning just in case.

sdks/cpp/README.md
sylvainduchesne: When you add a git external project in cmake, it automatically updates submodules. In the case of the previous PR, this was the case. One of GRPC submodule happens to be boringSSL :)

This is probably what will end up happening in this PR when no GRPC path is specified.

sdks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
```
aLekSer: 2019 in Copyright lines

dsazonoff: Probably we should patch all sources with such headers? Not only new one.

markmandel: Only new files need the new date. The date only needs to be applied at file creation.

dsazonoff:
@sylvainduchesne, @markmandel Please, review new changes.
Currently gRPC is used inplace (with add_subdirectory) and only necessary dependencies are built. OpenSSL is not required anymore.
It is possible to provide custom path to gRPC, but CMake will stop the build if version number is not equal to the version in CMakeLists.txt (there is an option to switch this check off).
If no gRPC installation found, CMake will download it from github.
Agones can be build for inplace usage or there is an "install" step, which will generate necessary configs for CMake package. Package is relocatable (no absolute path will be hardcoded).
There is a CMakeLists.txt in cpp-simple example.
-----
Some used workarounds:
- gRPC is not compiled completely and necessary cmake files are not generated. Solution: provided custom build/installation scripts for gRPC/protobuf and other depencencies.
- CMake (3.12 and higher) produce some warnings on gRPC compilation. They may be suppressed only with "-Wno-dev" option. Those warnings may be fixed when gRPC team will add CMake 3.0 support. Or we found another way to link with gRPC. Those warnings appears only when gRPC is added with add_subdirectory. Nothing is broken, just text warnings. Related to CMP0048 policy.
- Users need to suppress some warnings (related to protobuf) in their projects, if agones sdk is built as .dll. [Details](https://github.com/protocolbuffers/protobuf/blob/master/cmake/README.md#notes-on-compiler-warnings)

markmandel:
I defer to @sylvainduchesne on this for sure -- my C++ is no good here.

sdks/cpp/build_scripts/msvs_2017_x64_release.bat
```diff
@@ -0,0 +1,22 @@
...
+REM
+REM     http://www.apache.org/licenses/LICENSE-2.0
+REM
+REM Unless required by applicable law or agreed to in writing, software
+REM distributed under the License is distributed on an "AS IS" BASIS,
+REM WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+REM See the License for the specific language governing permissions and
+REM limitations under the License.
+
+pushd ..
```
sylvainduchesne: Don't know if this is a major issue, but these scripts need to be run from the build_scripts dir.
```
agones\sdks\cpp
$ build_scripts\msvs_2017_x64_release.bat
```
won't work

sylvainduchesne:
I did a build, and everything worked nicely. I was a bit surprised to see that it built DLL by default. Is this intended? I am also not sure how to build a static lib.

sylvainduchesne:
Looking for the documentation on how to build DLL, I did come across a typo in the URL of the SDK doc.
https://raw.githubusercontent.com/GoogleCloudPlatform/agones/master/sdks/cpp/README.md
`agondes.dev`, not sure if you want to fix this in the current PR

sdks/cpp/build_scripts/msvs_2017_x64_release.bat
```diff
@@ -0,0 +1,22 @@
...
+if not exist ".build" md ".build"
+pushd ".build"
+cmake .. -G "Visual Studio 15 2017 Win64" -Wno-dev
+cmake --build . --config Release
```
sylvainduchesne: might want to add `--target install`

sdks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,241 @@
...
+# limitations under the License.
+
+cmake_minimum_required (VERSION 3.13.0)
+
+project(agones VERSION 0.8.0 HOMEPAGE_URL https://github.com/GoogleCloudPlatform/agones LANGUAGES C CXX)
+
+# Build options
+option(AGONES_BUILD_SHARED "Build Agones C++ SDK as dynamic library" ON)
+option(AGONES_FORCE_GRPC_VERSION "Build Agones C++ SDK only with officially supported gRPC version" ON)
+option(AGONES_CREATE_PACKAGE "Generate CMake installation step and package files for Agones C++ SDK" OFF)
```
sylvainduchesne: I just tried with this enabled, and I got an error:
` cmake .. -G "Visual Studio 15 2017 Win64" -Wno-dev -DAGONES_BUILD_SHARED=OFF -DAGONES_CREATE_PACKAGE=ON`
```
-- Performing Test HAVE_STEADY_CLOCK -- success
CMake Error: Could not open file for write in copy operation C:/Program Files/agones/cmake/agonesConfig.cmake.tmp
CMake Error: : System Error: No such file or directory
CMake Error at C:/Program Files/CMake/share/cmake-3.13/Modules/CMakePackageConfigHelpers.cmake:316 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  CMakeLists.txt:189 (configure_package_config_file)
```

markmandel:
@sylvainduchesne raises a good point - we'll need to have this documented in https://agones.dev/site/docs/guides/client-sdks/cpp/ :+1:

sdks/cpp/CMakeLists.txt
dsazonoff: Good catch. I'll fix it. I need to store package config to the output folder first. 
In your case, even if build will be successful, you can't do an install step, because your compiler has no write access to default `CMAKE_INSTALL_PREFIX` path (which is Program Files).

sdks/cpp/build_scripts/msvs_2017_x64_release.bat
dsazonoff: What do you think about adding something like `CMAKE_INSTALL_PREFIX=./agones_sdk`?
In that case default installation will be in `sdks/cpp/.build/agones_sdk`. I think that users will not be happy with default Program Files installation location.

sylvainduchesne: indeed, default install path is probably not ideal.
How about defaulting to `${PROJECT_SOURCE_DIR}/libs` (or something along those lines).

dsazonoff:
@sylvainduchesne I fixed build. Now static build is default and installation step is added to build scripts. Please review once again.

sdks/cpp/build_scripts/msvs_2017_x64_release.bat
dsazonoff: I think it depends on personal preferences. Those scripts mostly for default build and demonstration of options. In real project there may be a lot of parent options like static/dynamic CRT, platform/compiler (x86, x64) etc.

sylvainduchesne:
Just tested the latest code, looks good! Sorry took a while, I recently switched computers at work.
Built examples/cpp-simple in shared and static, both worked without issues.

dsazonoff:
@markmandel , I think we can merge?

markmandel:
The only minor tweak I would make - on the documentation page, rather than replace the docs, keep the old documentation, and wrap it in a `feature` shortcode, and put the new section in its own `feature`s shortcode, so that the old switches with the new automatically for 0.9.0.

Details on how to do this here: https://agones.dev/site/docs/contribute/

Other than that, LGTM!

dsazonoff:
> The only minor tweak I would make - on the documentation page, rather than replace the docs, keep the old documentation, and wrap it in a `feature` shortcode, and put the new section in its own `feature`s shortcode, so that the old switches with the new automatically for 0.9.0.
> 
> Details on how to do this here: https://agones.dev/site/docs/contribute/
> 
> Other than that, LGTM!

@markmandel done

site/content/en/docs/Guides/Client SDKs/cpp.md
```diff
@@ -153,3 +153,58 @@ If you wish to compile from source, you will need to compile and install the fol
 If you are building a server on Windows or macOS, and need a development build to run on
 that platform, at this time you will need to compile from source. Windows and macOS libraries
 for the C++ SDK for easier cross platform development are planned and will be provided in the near future.
+{{% /feature %}}
+{{% feature publishVersion="0.9.0" %}}
```
markmandel: @sylvainduchesne I think this is good to go, but can you give these docs a once over, and if you are good, I'll approve and merge this PR!

build/Makefile
```diff
@@ -215,7 +215,7 @@ push-release-chart: $(ensure-build-image)
 	CHART_DIR=install/.helm-$(RELEASE_VERSION)/agones $(MAKE) push-chart
 
 #build all the sdks
-build-sdks: build-sdk-cpp
```
markmandel: @dsazonoff should this have been removed? Should we build and/or test this SDK on each run, in case something fails?

dsazonoff: I just removed c++ related part. Probably yes, we should remove this step.

markmandel: I guess my point is - shouldn't building the SDK be part of the tests, or ensure there isn't some error in it?

markmandel:
@dsazonoff this looks great! Please squash to a single commit, and resolve the above conflicts, and I'll approve and merge (just in time for RC on Tuesday!)

dsazonoff:
> @dsazonoff this looks great! Please squash to a single commit, and resolve the above conflicts, and I'll approve and merge (just in time for RC on Tuesday!)

Done, merged. I'm not sure, why build is failing. Locally everything is built.

PR #465, opened 2019-01-01
GameServerAllocation implementation

### LINKED ISSUE #436: Allocation on GameServers rather than Fleets
Have had a couple of conversations about `FleetAllocation` being tied to `Fleets` specifically, and how it would be more flexible to decouple this.

Thinking back, I'm feeling that the original design decision to tie these two things together was less than ideal - so wanted to propose the below solution to rectify this, and a couple of other things (`FleetAllocation.Status` primarily) I would like to see improved as well.

Also, given that people are actively using the platform - we are rapidly approaching what I think will be a 1.0 (separate conversation 😊) - if we want to make big changes to solve foundational issues, we may want to make them now, before we settle on what becomes a "1.0",  and things become harder to change (but check my backward compatibility section below on this for some more details on this).

## Proposed Specification

The flow of allocation should not change (basically everything works that same as `FleetAllocation` does now), but I propose that game server allocations now occur using the following CRD instead:

```yaml
kind: GameServerAllocation
metadata:
  # We recommend using the following to generate a unique name when creating Allocations
  # This will need to be created with `kubectl create` if using the command line tooling
  generateName: fleet-allocation-example-
spec:
  # GameServer selector from which to choose GameServers from.
  # GameServers still have the hard requirement to be `Ready` to be allocated from
  # however we can also make available `matchExpressions` for even greater
  # flexibility.
  # Below is an example of a GameServer allocated against a given fleet.
  required:
    matchLabels:
      stable.agones.dev/fleet: fleet-example
  # ordered list of preferred allocations out of the `required` set.
  # If the first selector is not matched,
  # the selection attempts the second selector, and so on.
  # This also support `matchExpressions`
  preferred:
    - matchLabels:
        stable.agones.dev/fleet: green-fleet
    - matchLabels:
        stable.agones.dev/fleet: blue-fleet
  # Allocation strategy. Packed or Distributed.
  # Since we aren't tied to a Fleet, we have to declare this here.
  # (Packed is default as per usual)
  strategy: Packed
  # Optional custom metadata that is added to the game server at allocation
  # You can use this to tell the server necessary session data
  metadata:
    labels:
      mode: deathmatch
    annotations:
      map:  garden22
```

The GameServerAllocation will then use both `required` selector to make a list of `Ready` `GameServers` that match the selector, and allocate based on that list (instead of listing the `GameServers` for a given specific Fleet).

`GameServers` that match the `preferred` selectors will be allocated first out of the `required` set. The list is a preferred priority list - i.e. an attempt will be made to find a `GameServer` that matches the first selector,
and if that cannot be matched, and attempt will be made for the second, and so forth. If a `GameServer` cannot be found, allocation will occur from the wider `required` set.

This decouples `Allocations` from `Fleets` - giving users (and ourselves in the future) the ability to create their own `GameServer` management strategies -- while still being able to take advantage of `Allocations` -- while also maintaining the ability to do a more simple `GameServerAllocation` from a `Fleet` with the selector. This also puts us closer inline with Kubernetes strategies as a whole as well.

Having selectors makes `GameServerAllocation` far more flexible than its current incarnation. Also, this would also provide the ability to allocate on [matchexpressions (set based expressions)](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) - which is even more flexible.

For example, this makes it possible for a user to allocate across a subset of a `Fleet` (possibly in combination with SDK Label setting),
or allocating across multiple Fleets inside a cluster for smoke tests with preferred selectors - depending on needs - or scaling of Fleets.

Downside here being that there is more complexity here - because most people will likely 1:1 allocate from a Fleet - but I think with appropriate documentation, this shouldn't be a large issue.

## GameServerAllocation Status

In retrospect, I don't think it was a good idea to include the entire `GameServer` as part of the FleetAllocation state. It is far too much information (which also required extra processing to generate the json patch), and could be potentially confusing as it may be expected for that `GameServer` information to stay up to date with the `FleetAllocation`.

Also, it is currently difficult to determine if an GameServerAllocation has failed because of error, or because there weren't any GameServers available. To that end, we should include this information in the `Allocation.Status` value as `State`.

For example, this is the response you get when a `FleetAllocation` fails due to lack of `GameServer` inventory via curl:

```cmd
< HTTP/1.1 500 Internal Server Error
< Content-Length: 415
< Content-Type: application/json
< Date: Sun, 02 Dec 2018 04:36:40 GMT
<
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "Internal error occurred: admission webhook \"mutations.stable.agones.dev\" denied the request: NotFound",
  "reason": "InternalError",
  "details": {
    "causes": [
      {
        "message": "admission webhook \"mutations.stable.agones.dev\" denied the request: NotFound"
      }
    ]
  },
  "code": 500
```

You would need to hunt for the string "NotFound" in `message` values - it's not inherently clear.

Instead, I would recommend the following `Status` values:

```golang
const (
        // Allocation successful
        Allocated State = "Allocated"
        // Allocation unsuccessful
        UnAllocated State = "UnAllocated"
)

type State string

// GameServerAllocationStatus is the status for an GameServerAllocation resource
type GameServerAllocationStatuss struct {
        // State is the current state of an GameServerAllocation, e.g. Allocated, or UnAllocated
        State          State                           `json:"state"`
        GameServerName string                          `json:"gameServerName"`
        Ports          []v1alpha1.GameServerStatusPort `json:"ports,omitempty"`
        Address        string                          `json:"address,omitempty"`
        NodeName       string                          `json:"nodeName,omitempty"`
}

```

This was a value is always returned, and an easily understood and parsed value of `State` will define if an allocation occurred successfully or not.

When an `GameServerAllocation` is returned in an `UnAllocated` state, it should be marked for deletion / deleted in a timely manner after return, to ensure they don't end up remaining forever.

In this incarnation, if a HTTP 500 error occurs - this is a bug / server issue, and is correctly identified as such.

If a user needs more information about the allocated `GameServer` - the GameServer name is available, and details can be retrieved through the Kubernetes API.

## Implementation

The good thing here, is that the implementation of this is not drastically different from the current `FleetAllocation` implementation (in fact, I believe that 99% of the code and tests can be copy pasted across into an `GameServerAllocation` controller -- and probably should be separate because of backward compatibility. See below).

To provide the described functionality, the only major change would be to change [this line](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/fleetallocation/controller.go#L270), from fleetallocation/controller.go's `allocate()` function from  `gsList, err := fleets.ListGameServersByFleetOwner(c.gameServerLister, f)`
to instead search for `GameServers` via the provided `required` selector, and filter appropriately via the `preferred` selector list. Almost all other code should be able to be reused.

## Backward Compatibility

While we are alpha, we do have games that are in/close to production - so it's good to have a good deprecation policy here. Also FleetAllocation is such a core part of the system, changing this shouldn't be done lightly.

We can deprecate FleetAllocations, and keep the code, and everything there can maintain its functionality. I would suggest a removal of FleetAllocation in +2 releases from when the GameServerAllocation CRD is released, to give everyone 12 weeks to test and migrate to the new GameServerAllocation CRD.

Since the mutex for allocation is already shared amongst controllers in the system - there is also no reason users can't use both GameServerAllocation and FleetAllocation at the same time as they migrate.

As always - thoughts, feelings, questions and all feedback is appreciated.

markmandel:
This implements the `GameServerAllocation` as designed in #436, which decouples Allocation from Fleets, and provides much more flexible functionality.

This also moves FleetAllocation to deprecated, and includes notes that it will be removed in +2 releases.

Closes #436


markmandel:
This is a bigger one, with lots of refactoring in it - I'm guessing :smile: it won't make 0.7.0, but wrote it as if it does, but can adjust as needed if need be.

examples/gameserverallocation.yaml
```diff
@@ -0,0 +1,62 @@
...
+  # ordered list of preferred allocations out of the `required` set.
+  # If the first selector is not matched, the selection attempts the second selector, and so on.
+  # This is useful for things like smoke testing of new game servers.
+  # This also support `matchExpressions`
+  preferred:
+    - matchLabels:
+        stable.agones.dev/fleet: green-fleet
+    - matchLabels:
+        stable.agones.dev/fleet: blue-fleet
+  replicas: 2
```
Kuqd:
```suggestion
```

Kuqd: I guess you don't need replicas here.

pkg/gameserverallocations/controller.go
```diff
@@ -0,0 +1,394 @@
...
+
+	json, err := json.Marshal(patch)
+	if err != nil {
+		c.logger.WithError(err).Error("error creating the json for the patch")
+		return review, errors.Wrapf(err, "error creating json for patch for GameServerAllocation %s", gs.ObjectMeta.Name)
+	}
+
+	c.logger.WithField("gsa", gsa.ObjectMeta.Name).WithField("patch", string(json)).Infof("patch created!")
+
+	pt := admv1beta1.PatchTypeJSONPatch
```
Kuqd: I'm curious what this does ? Tells the webhook to patch the resource ?

pkg/gameserverallocations/controller.go
```diff
@@ -0,0 +1,394 @@
...
+		}
+		for key, value := range fam.Annotations {
+			gs.ObjectMeta.Annotations[key] = value
+		}
+	}
+}
+
+// syncDelete takes unallocated GameServerAllocations, and deletes them!
+func (c *Controller) syncDelete(key string) error {
+	c.logger.WithField("key", key).Info("Deleting gameserverallocation")
```
Kuqd: Does this means the `GameServerAllocation` is added with a status `GameServerAllocationUnAllocated` but deleted right after. Asking question mostly for metrics implementation.

pkg/gameserverallocations/controller.go
markmandel: So when you return the `AdmissionReview.Response`, you can apply a `Patch, which states what changes you want to make to the object that you were originally passed - that's how you make changes via the mutationWebHook.

https://medium.com/ibm-cloud/diving-into-kubernetes-mutatingadmissionwebhook-6ef3c5695f74 has a good explanation.

pkg/gameserverallocations/controller.go
markmandel: Yeah - I couldn't think of a better way to handle UnAllocated GameServerAllocations - since there is no GameServer that will eventually get deleted, it need to be returned in the webhook, but eventually deleted, otherwise it would hang around forever.

docs/create_fleet.md
```diff
@@ -146,14 +163,106 @@ simple-udp-xvp4n-i6bnm   36s
...
+> In production, you would likely do the following through a [Kubernetes API call](./access_api.md), but we can also
 do this through `kubectl` as well, and ask it to return the response in yaml so that we can see what has happened.
 
+#### GameServerAllocation
+
+⚠️⚠️⚠️ **This is currently a development feature and has not been released** ⚠️⚠️⚠️
+
+We can do allocation of a GameServer for usage through a `GameServerAllocation`, which will both 
+return to us the details of a `GameServer` (assuming one is available), and also move it to the `Allocated` state,
+which demarkates that it
```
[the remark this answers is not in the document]
markmandel: Good catch. Fixed!

PR #474, opened 2019-01-08
fix autoscaler cleanup on tests failure

### LINKED ISSUE #471: e2e tests don't cleanup fleetautoscalers
Since Fleet Autoscalers aren't tied to fleets anymore, they can get left behind.  We should add some code to the end of the testing suite to delete all remaining Fleet Autoscalers.

(/cc @Kuqd who found this)

Kuqd:
Closes #471

It seems to be happening only when tests are failing.


markmandel:
Thought: should we also put a delete all FleetAllocations in the TestMain() like we do for Fleets and GameServers?

Kuqd:
Sure I’ll add it

PR #475, opened 2019-01-08
docs: added game server state diagram

jkowalski:


docs/gameserver_spec.md
```diff
@@ -46,3 +46,15 @@ The `spec` field is the actual GameServer specification and it is composed as fo
   - `protocol` the protocol being used. Defaults to UDP. TCP is the only other option.
 - `health` to track the overall healthy state of the GameServer, more information available in the [health check documentation](./health_checking.md).
 - `template` the [pod spec template](https://v1-10.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#podtemplatespec-v1-core) to run your GameServer containers, [see](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/#pod-templates) for more information.
+
+## GameServer State Diagram
```
markmandel: I feel like this should be it's own doc - maybe under reference? Or maybe part of the SDK integration? Not sure.

Here is feels a bit hidden?

PR #486, opened 2019-01-16
Fix for the controller panic issue on metrics.enabled is false

aLekSer:
If we disable metrics in Helm config file, Agones controller will panic on Run() call which could not be executed on non-nil interface with nil value (and not empty type of metrics.Controller).

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -104,8 +104,8 @@ func main() {
 	kubeInformationFactory := informers.NewSharedInformerFactory(kubeClient, defaultResync)
 
 	server := &httpServer{}
+	var rs []runner
 	var health healthcheck.Handler
-	var metricsController *metrics.Controller
 
 	if ctlConf.Metrics {
 		registry := prom.NewRegistry()
@@ -115,8 +115,7 @@ func main() {
 		}
 		server.Handle("/metrics", metricHandler)
 		health = healthcheck.NewMetricsHandler(registry, "agones")
-		metricsController = metrics.NewController(kubeClient, agonesClient, agonesInformerFactory)
-
+		rs = append(rs, metrics.NewController(kubeClient, agonesClient, agonesInformerFactory))
 	} else {
 		health = healthcheck.NewHandler()
 	}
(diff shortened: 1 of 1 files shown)


markmandel:
Rather than use all this reflect (and generally we shouldn't need to) - and to avoid this messiness, why don't we try something more like this, where we dynamically append to `rs` rather than look for nil values.

WDYT?

```go
	server := &httpServer{}
	var rs []runner
	var health healthcheck.Handler

	if ctlConf.Metrics {
		registry := prom.NewRegistry()
		metricHandler, err := metrics.RegisterPrometheusExporter(registry)
		if err != nil {
			logger.WithError(err).Fatal("Could not create register prometheus exporter")
		}
		server.Handle("/metrics", metricHandler)
		health = healthcheck.NewMetricsHandler(registry, "agones")
		rs = append(rs, metrics.NewController(kubeClient, agonesClient, agonesInformerFactory))

	} else {
		health = healthcheck.NewHandler()
	}

	server.Handle("/", health)

	allocationMutex := &sync.Mutex{}

	gsController := gameservers.NewController(wh, health, allocationMutex,
		ctlConf.MinPort, ctlConf.MaxPort, ctlConf.SidecarImage, ctlConf.AlwaysPullSidecar,
		ctlConf.SidecarCPURequest, ctlConf.SidecarCPULimit,
		kubeClient, kubeInformationFactory, extClient, agonesClient, agonesInformerFactory)
	gsSetController := gameserversets.NewController(wh, health, allocationMutex,
	faController := fleetallocation.NewController(wh, allocationMutex,
	gasController := gameserverallocations.NewController(wh, health, allocationMutex, kubeClient,
		kubeInformationFactory, extClient, agonesClient, agonesInformerFactory)
	fasController := fleetautoscalers.NewController(wh, health,

	rs = append(rs, wh, gsController, gsSetController, fleetController, faController, fasController, gasController, server)

	stop := signals.NewStopChannel()

	kubeInformationFactory.Start(stop)
	agonesInformerFactory.Start(stop)

	for _, r := range rs {
		go func(rr runner) {
			if runErr := rr.Run(workers, stop); runErr != nil {
				logger.WithError(runErr).Fatalf("could not start runner: %s", reflect.TypeOf(rr))
			}
		}(r)
	}
```

Kuqd:
I think it's a great idea !

PR #487, opened 2019-01-16
specify resource limits on simple-udp/fleet.yaml

jkowalski:
This allows simple-udp to be used for scale testing of GS controller with fewer nodes.

### DIFF
--- a/examples/simple-udp/fleet.yaml
+++ b/examples/simple-udp/fleet.yaml
@@ -28,3 +28,10 @@ spec:
           containers:
           - name: simple-udp
             image: gcr.io/agones-images/udp-server:0.5
+            resources:
+              requests:
+                memory: "64Mi"
+                cpu: "20m"
+              limits:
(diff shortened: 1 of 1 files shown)


markmandel:
Why both requests and limits? i don't think it makes sense to use both?

Probably makes life easier to just use a request, so it can burst if need be, but will still give direction to the scheduler.

jkowalski:
> Why both requests and limits? i don't think it makes sense to use both?
> 
> Probably makes life easier to just use a request, so it can burst if need be, but will still give direction to the scheduler.

Take a look at:

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/resource-qos.md

Basically if you specify limit==request, Kubernetes will put the pod in the Guaranteed class and:

"Guaranteed pods are considered top-priority and are guaranteed to not be killed until they exceed their limits, or if the system is under memory pressure and there are no lower priority containers that can be evicted."

PR #500, opened 2019-01-23
improved isolation of Agones controllers using taints and priority

### LINKED ISSUE #489: Assign higher priority to Agones system pods
Currently when creating large cluster (say 10K GS), when agones controller dies for whatever reason, it will typically not be recreated immediately if lots of other pods are also pending.

Since we're on Kubernetes 1.11+, we should be able to leverage 

https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#pod-priority

I tried that and it actually helped - as soon as I added priority class and changed it on agones deployment, the controller started immediately.

jkowalski:
Added high priority class called `agones-system`, which defines the priority at which both controller and ping service run. This causes them to be scheduled before any game server pods are scheduled.
This fixes #489

Also added default affinity and tolerations to helm values:

Agones Controller and Ping will prefer (but not require) nodes labeled with `stable.agones.dev/agones-system: true`. They will also tolerate taint `stable.agones.dev/agones-system=true:NoExecute`.

With those two mechanisms in place, isolating Agones controller should be as simple as creating dedicated node pool with appropriate annotations/labels:

```
gcloud container node-pools create agones-system ... \
  --node-taints stable.agones.dev/agones-system=true:NoExecute \
  --node-labels stable.agones.dev/agones-system=true
```

Observe how pods are scheduled on `agones-system` node pool.

```
$ kubectl get pod -n agones-system -o wide
NAME                                READY   STATUS    RESTARTS   AGE
IP         NODE                                           NOMINATED NODE
agones-controller-cd857555b-zgpqw   1/1     Running   0          34s
10.8.5.6   gke-agones-scale-agones-system-b36e72f2-7ks0   <none>
agones-ping-76999c8cc9-9nghq        1/1     Running   0          42s
10.8.4.3   gke-agones-scale-agones-system-823b885e-15nz   <none>
agones-ping-76999c8cc9-jhjq6        1/1     Running   0          39s
10.8.5.5   gke-agones-scale-agones-system-b36e72f2-7ks0   <none>
```

### DIFF
--- a/build/gke-test-cluster/cluster.yml.jinja
+++ b/build/gke-test-cluster/cluster.yml.jinja
@@ -34,6 +34,21 @@ resources:
               - https://www.googleapis.com/auth/devstorage.read_only
               - https://www.googleapis.com/auth/logging.write
               - https://www.googleapis.com/auth/monitoring
+        - name: "agones-system"
+          initialNodeCount: 1
+          config:
+            machineType: n1-standard-2
+            oauthScopes:
+              - https://www.googleapis.com/auth/compute
+              - https://www.googleapis.com/auth/devstorage.read_only
+              - https://www.googleapis.com/auth/logging.write
+              - https://www.googleapis.com/auth/monitoring
+            labels:
+              stable.agones.dev/agones-system: "true"
+            taints:
+              - key: stable.agones.dev/agones-system
+                value: "true"
+                effect: "NO_EXECUTE"
       masterAuth:
         username: admin
         password: supersecretpassword
--- a/site/content/en/docs/Installation/helm.md
+++ b/site/content/en/docs/Installation/helm.md
@@ -28,6 +28,22 @@ $ helm install --name my-release --namespace agones-system agones/agones
 _We recommend to install Agones in its own namespaces (like `agones-system` as shown above)
 you can use the helm `--namespace` parameter to specify a different namespace._
 
+{{% feature publishVersion="0.8.0" %}}
+
+When running in production, Agones should be scheduled on a dedicated pool of nodes, distinct from where Game Servers are scheduled for better isolation and resiliency. By default Agones prefers to be scheduled on nodes labeled with `stable.agones.dev/agones-system=true` and tolerates node taint `stable.agones.dev/agones-system=true:NoExecute`. If no dedicated nodes are available, Agones will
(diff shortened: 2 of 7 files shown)


jkowalski:
How about we add both nodepools and makefile targets in separate PR? This way we'll know that Agones works with and without special nodepools.

jkowalski:
PTAL, added documentation and DM template

build/gke-test-cluster/cluster.yml.jinja
```diff
@@ -34,6 +34,21 @@ resources:
               - https://www.googleapis.com/auth/devstorage.read_only
               - https://www.googleapis.com/auth/logging.write
               - https://www.googleapis.com/auth/monitoring
+        - name: "agones-system"
+          initialNodeCount: 1
+          config:
+            machineType: n1-standard-4
```
markmandel: Do we want 4 cpu, or would 2 do?

site/content/en/docs/Installation/helm.md
```diff
@@ -28,6 +28,19 @@ $ helm install --name my-release --namespace agones-system agones/agones
 _We recommend to install Agones in its own namespaces (like `agones-system` as shown above)
 you can use the helm `--namespace` parameter to specify a different namespace._
 
+When running in production, Agones should be scheduled on a dedicated pool of nodes, distinct from where Game Servers are scheduled for better isolation and resiliency. By default Agones prefers to be scheduled on nodes labeled with `stable.agones.dev/agones-system=true` and tolerates node taint `stable.agones.dev/agones-system=true:NoExecute`. If no dedicated nodes are available, Agones will
```
markmandel: Final thing - wrap this is a `feature` shortcode with a 0.8.0 publish date, so it's hidden until that date.

Details: https://agones.dev/site/docs/contribute/

Otherwise, LGTM!

build/gke-test-cluster/cluster.yml.jinja
jkowalski: At some point for scaling cluster of 10Ks you want bigger machines. For this one 2 is fine.

site/content/en/docs/Installation/helm.md
jkowalski: Done

PR #501, opened 2019-01-23
Prometheus and grafana improvements based on load testing experience

jkowalski:
- made prometheus PVC size configurable on the command line
- moved majority of Prometheus config overrides to separate yaml file.
- removed scraping of container stats from prometheus config, otherwise
clusters of 10K pods are very quickly consuming tons of space
- added taints and tolerations to prometheus and grafana

They will now will prefer (but not require) to be scheduled on nodes labeled
with `stable.agones.dev/agones-metrics: true`. They will also tolerate
taint `stable.agones.dev/agones-metrics=true:NoExecute`.

Creating node pool dedicated for monitoring is as simple as:

```
gcloud container node-pools create agones-metrics ... \
  --node-taints stable.agones.dev/agones-metrics=true:NoExecute
  --node-labels stable.agones.dev/agones-metrics=true \
  --num-nodes=1
```

### DIFF
--- a/build/gke-test-cluster/cluster.yml.jinja
+++ b/build/gke-test-cluster/cluster.yml.jinja
@@ -47,6 +47,19 @@ resources:
               stable.agones.dev/agones-system: "true"
             taints:
               - key: stable.agones.dev/agones-system
+        - name: "agones-metrics"
+          initialNodeCount: 1
+          config:
+            machineType: n1-standard-4
+            oauthScopes:
+              - https://www.googleapis.com/auth/compute
+              - https://www.googleapis.com/auth/devstorage.read_only
+              - https://www.googleapis.com/auth/logging.write
+              - https://www.googleapis.com/auth/monitoring
+            labels:
+              stable.agones.dev/agones-metrics: "true"
+            taints:
+              - key: stable.agones.dev/agones-metrics
                 value: "true"
                 effect: "NO_EXECUTE"
       masterAuth:
--- a/site/content/en/docs/Guides/metrics.md
+++ b/site/content/en/docs/Guides/metrics.md
@@ -118,14 +118,27 @@ Prometheus is an open source monitoring solution, we will use it to store Agones
 Let's install Prometheus using the [helm stable](https://github.com/helm/charts/tree/master/stable/prometheus) repository.
 
 ```bash
-helm install --wait --name prom stable/prometheus --namespace metrics \
-  --set pushgateway.enabled=false \
-  --set kubeStateMetrics.enabled=false,nodeExporter.enabled=false
+helm upgrade --install --wait prom stable/prometheus --namespace metrics \
+    --set server.global.scrape_interval=30s \
+    --set server.persistentVolume.enabled=true \
+    --set server.persistentVolume.size=64Gi \
+    -f ./build/prometheus.yaml
 ```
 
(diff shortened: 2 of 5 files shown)


jkowalski:
How about we add both nodepools and makefile targets in separate PR? This way we'll know that Agones works with and without special nodepools.

markmandel:
At the very least, this should have documentation in the metrics page for best practices with prom.

I'd still be inclined to include the node pool in the deployment manager script (not for e2e), so we can test both?

For personal testing in always going to default to wanting to have this node pool. I assume everyone else will too?

jkowalski:
updated DM template and docs

build/gke-test-cluster/cluster.yml.jinja
```diff
@@ -34,6 +34,21 @@ resources:
               - https://www.googleapis.com/auth/devstorage.read_only
               - https://www.googleapis.com/auth/logging.write
               - https://www.googleapis.com/auth/monitoring
+        - name: "agones-metrics"
+          initialNodeCount: 1
+          config:
+            machineType: n1-standard-4
```
markmandel: Should this be 4 cpu?
(Should this be in the same nodepool as the controller?)

site/content/en/docs/Guides/metrics.md
```diff
@@ -119,6 +119,17 @@ helm install --wait --name prom stable/prometheus --namespace metrics \
 or `make kind-setup-prometheus` and `make minikube-setup-prometheus` for {{< ghlink href="/build/README.md#running-a-test-kind-cluster" branch="master" >}}Kind{{< /ghlink >}}
 and {{< ghlink href="/build/README.md#running-a-test-minikube-cluster" branch="master" >}}Minikube{{< /ghlink >}}.
 
+For resiliency it is recommended to run Prometheus on a dedicated node which is separate from nodes where Game Servers are scheduled. If you use `make setup-prometheus` to set up Prometheus, it will schedule Prometheus pods on nodes tainted with `stable.agones.dev/agones-metrics=true:NoExecute` and labeled with `stable.agones.dev/agones-metrics=true` if available.
```
markmandel: @kuqd remind me what's our standard here, for stating how we want to do this? Point to the yaml in GitHub?

Not sure if this should be a `feature` shortcode here to hide this until 0.8.0?

build/gke-test-cluster/cluster.yml.jinja
jkowalski: I think those will be different shapes of VMs, so probably not (prometheus may get really large, while the controller likely won't).

site/content/en/docs/Guides/metrics.md
Kuqd: can you update line 113 with the new command using the yaml and here just explain the same way that with a link to your yaml config.

Kuqd: new command being 
```
helm upgrade --install --wait prom stable/prometheus --namespace metrics \
		--set alertmanager.enabled=false,pushgateway.enabled=false \
		--set kubeStateMetrics.enabled=false,nodeExporter.enabled=false \
		--set pushgateway.enabled=false \ 
--set server.global.scrape_interval=30s,server.persistentVolume.enabled=true,server.persistentVolume.size=64Gi  -f ./build/prometheus.yaml
```

markmandel: Awesome. I think once this is there, this is good for approval :+1:

jkowalski: done

PR #505, opened 2019-01-24
add client-go metrics and grafana dashboards

Kuqd:
This adds metrics integration for client-go and some dashboards that goes along with. I've refactored also our metrics package after having some discussion with @jkowalski last week.

Documentation has been updated with new dashboards.

@jkowalski I think those will be very interesting for you to see our Kubernetes API consumption but also the current status of our workqueues. (Caches is also there but I doubt it will shed more light as for now). Especially when you are load testing our controller.

@aLekSer I would love that you take this for a spin on Stackdriver see if metrics are working well.

This works also helps https://github.com/GoogleCloudPlatform/agones/issues/144 we should at last add game server allocations and deprecate fleet allocations.

some dashboards screenshots:
screen shot 2019-01-24 at 11 16 02 am
screen shot 2019-01-24 at 11 16 59 am
screen shot 2019-01-24 at 11 19 02 am

Let me know !


site/content/en/docs/Guides/metrics.md
```diff
@@ -173,24 +178,3 @@ kubectl port-forward deployments/grafana 3000 -n metrics
 Open a web browser to [http://127.0.0.1:3000](http://127.0.0.1:3000), you should see Agones [dashboards](#grafana-dashboards) after login as admin.
 
 > Makefile targets `make grafana-portforward`,`make kind-grafana-portforward` and `make minikube-grafana-portforward`.
-
-## Adding more metrics
```
markmandel: Curious - reason for removing this? (Not saying we shouldn't, just wondering what the motivation was)

site/content/en/docs/Guides/metrics.md
```diff
@@ -89,6 +88,12 @@ We provide a set of useful [Grafana](https://grafana.com/) dashboards to monitor
 
 - {{< ghlink href="/build/grafana/dashboard-controller-usage.yaml" branch="master" >}}Agones Controller Resource Usage{{< /ghlink >}} displays Agones Controller CPU and memory usage and also some Golang runtime metrics.
 
+- {{< ghlink href="/build/grafana/dashboard-goclient-requests.yaml" branch="master" >}}Agones Controller go-client requests{{< /ghlink >}} displays Agones Controller Kubernetes API consumption.
```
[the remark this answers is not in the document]
Kuqd: sure I'll have a look

site/content/en/docs/Guides/metrics.md
Kuqd: I had a talk last week with @jkowalski and we both think opencensus is already an abstraction there is no issue in using metrics directly where it matter, for example in the gameserver package we could have a metric.go file.

Originally this documentation comes from our discussion and we thought it would be good to avoid spreading metric code in our code base, but ultimately I think this makes thing more difficult to write.

markmandel: Okily. :+1:

site/content/en/docs/Guides/metrics.md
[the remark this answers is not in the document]
markmandel: Other than wrapping this in a `feature` shortcode to hide it until 0.8.0, this LGTM. (Just looking at it again)

aLekSer:
Hello @Kuqd , my apologize for missing your request.
Overall new metrics looks good however I found an issue with `agones/k8s_client_workqueue_depth` which shows only `-1` value.

Attaching dashboard example:
controllermetrics

PR #506, opened 2019-01-24
Speed up builds by using local go/zip instead of dockerized ones.

### LINKED ISSUE #453: Improve build speed by refactoring Makefile

jkowalski:
This adds LOCAL_GO and LOCAL_ZIP which cause locally-installed Go toolkit and zip tool to be used when building binaries, greatly speeding up incremental rebuilds on a Mac where docker filesystem driver is very slow.

Before:

```
$ time make -j4 build-controller-binary build-agones-sdk-binary

real    1m11.369s
user    0m0.325s
sys     0m0.273s
```

After:

```
$ time make -j4 build-controller-binary build-agones-sdk-binary LOCAL_GO=1 LOCAL_ZIP=1

real    0m4.605s
user    0m3.480s
sys     0m4.527s
```

Note that when using LOCAL_GO, agones.dev must be checked out in a GOPATH and Go compiler must be installed at appropriate version.
Use at your own risk!

This pretty fixes #453 as the incremental build times are finally good.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -121,6 +121,26 @@ ifndef FULL_BUILD
 	zip_flags = -0
 endif
 
+GO_BUILD_LINUX_AMD64=docker run --rm -e "CGO_ENABLED=0" $(common_mounts) $(build_tag) go build
+GO_BUILD_DARWIN_AMD64=docker run --rm -e "GOOS=darwin" -e "GOARCH=amd64" $(common_mounts) $(build_tag) go build
+GO_BUILD_WINDOWS_AMD64=docker run --rm -e "GOOS=windows" -e "GOARCH=amd64" $(common_mounts) $(build_tag) go build
+
+go_build_base_path=$(mount_path)
+
+ifdef LOCAL_GO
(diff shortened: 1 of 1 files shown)


jkowalski:
nice speedup, although not as dramatic as what I can see, can you merge?

PR #497, opened 2019-01-23
controller: made QPS, burst QPS and number of workers externally configurable

jkowalski:
Also increased their default values significantly based on scale test results.

PR #513, opened 2019-01-26
Restructure cloudbuild.yaml

markmandel:
Aim to (a) make the cloudbuild.yaml much more readable so that you can see what is supposed to run in what order and (b) make more things run in parallel to attempt to make the build faster.


markmandel:
It's cleaner, but I can't seem to make it go faster :cry:

jkowalski:
to make it faster, first merge #497 then use something like -parallel 32

jkowalski:
Very nice! LGTM

PR #516, opened 2019-01-26
Fix for flaky TestSDKSetAnnotation

markmandel:
Pretty sure the new version annotation on GameServers made this test flaky.

This is a fix!

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -88,10 +88,12 @@ func TestSDKSetLabel(t *testing.T) {
 func TestSDKSetAnnotation(t *testing.T) {
 	t.Parallel()
 	gs := defaultGameServer()
+	annotation := "stable.agones.dev/sdk-timestamp"
 	readyGs, err := framework.CreateGameServerAndWaitUntilReady(defaultNs, gs)
 	if err != nil {
 		t.Fatalf("Could not get a GameServer ready: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/gameserver_test.go
```diff
@@ -109,12 +110,13 @@ func TestSDKSetAnnotation(t *testing.T) {
...
-		return gs.ObjectMeta.Annotations != nil, nil
+
+		_, ok := gs.ObjectMeta.Annotations[annotation]
+		return ok, nil
 	})
 
 	assert.Nil(t, err)
-	assert.NotEmpty(t, gs.ObjectMeta.Annotations["stable.agones.dev/sdk-timestamp"])
-
+	assert.NotEmpty(t, gs.ObjectMeta.Annotations[annotation])
```
jkowalski: I distinctly remember seeing failures where "sdk-timestamp" was set but version was not.

jkowalski: Can you try:

```go test -count=100 -timeout 90s agones.dev/agones/test/e2e -run TestSDKSetAnnotation```

markmandel: That would be strange, as the version annotation is set via a mutation webhook.

Ran this 50 times (in batches to fit under the timeout - also added a defer to delete the gameserver), no errors :man_shrugging: 

(Nice trick with `-count=100` didn't know that one, will steal it :+1: )

PR #520, opened 2019-01-28
fixed gcloud-test-cluster setup problem caused by bad merge between #500 and #501

jkowalski:


markmandel:
Going to run a quick test, and then will approve.

/cc @aLekSer

PR #525, opened 2019-01-29
Fix potential data race in allocation counter

markmandel:
Found a rare data race in tests for the allocation counter.

Could have potentially have leaked into production code, so here is a fix.


jkowalski:
LGTM

Issue #530, opened 2019-01-29
gameserver configurations have no validation on update.


jeremyje:
Right now there's only validation on creation of a gameserver. If you push a configuration that's invalid like static port with no port numbers via update. Agones will let that through.

jeremyje:
I have a fix for this already. Discovered when working on #469

roberthbailey:
@jeremyje - do you still have a fix for this floating around? It'd be nice to get it added in, since I recently ran into this in my testing.

jeremyje:
No I abandoned the fix. My PR is referenced just above but I doubt it's compatible with what's there now. There's also plenty of edge cases that it doesn't quite work.

markmandel:
If/when we come back to this, once Player Tracking is in, initialCapacity shouldn't be editable.

See: https://github.com/googleforgames/agones/pull/1324#discussion_r378634250 for more context

PR #532, opened 2019-01-30
Add basic gameserver validation on update.

jeremyje:
Addresses issue #530.


jeremyje:
> This seems particularly flaky. Did you want me to have a poke and see if I can see if the code change is making a difference?

Sure, also where's the docs to debug the e2e tests? I've been trying the run them locally but I don't know how to pull logs from the docker containers it's creating.

Is there a way to run these tests in vscode?

markmandel:
https://agones.dev/site/docs/guides/troubleshooting/ is the guide you want.

If you can run Go tests in vscode, then you should also be able to run the e2e tests (they are essentially just go tests).

If you are running the e2e tests on your own GKE instances, you can also see all the logs in stackdriver.

jeremyje:
I figured out what's causing the issue. It turns out that /validate runs on every mutation not just ones initiated by kubectl.

My code is rejecting the request on the transition from PortAllocation to Creating because the HostPort is set when the PortPolicy=Dynamic. 

I think we'll need to split up the Validate logic to prevent always bad configurations from bad initial configs. Either that or add caveats into the PortPolicy check which might complicate things later on, adding new states, etc.

  review: {
   apiVersion:  "admission.k8s.io/v1beta1"    
   kind:  "AdmissionReview"    
   request: {
    kind: {…}     
    name:  "udp-serverswfxj"     
    namespace:  "default"     
    object: {
     apiVersion:  "stable.agones.dev/v1alpha1"      
     kind:  "GameServer"      
     metadata: {…}      
     spec: {
      container:  "udp-server"       
      health: {…}       
      ports: [
       0: {
        containerPort:  7654         
        hostPort:  7181         
        name:  "gameport"         
        portPolicy:  "dynamic"         
        protocol:  "UDP"         
       }
      ]
      scheduling:  "Packed"       
      template: {…}       
     }
     status: {
      address:  ""       
      nodeName:  ""       
      ports:  null       
      state:  "Creating"       
     }
    }
    oldObject: {
     apiVersion:  "stable.agones.dev/v1alpha1"      
     kind:  "GameServer"      
     metadata: {…}      
     spec: {
      container:  "udp-server"       
      health: {…}       
      ports: [
       0: {
        containerPort:  7654         
        name:  "gameport"         
        portPolicy:  "dynamic"         
        protocol:  "UDP"         
       }
      ]
      scheduling:  "Packed"       
      template: {…}       
     }
     status: {
      address:  ""       
      nodeName:  ""       
      ports:  null       
      state:  "PortAllocation"       
     }
    }
    operation:  "UPDATE"     
    resource: {…}     
    uid:  "0a63a40c-24d9-11e9-874d-42010a8a0141"     
    userInfo: {…}     
   }
   response: {
    allowed:  false     
    status: {
     details: {
      causes: [
       0: {
        field:  "gameport.hostPort"         
        message:  "HostPort cannot be specified with a Dynamic PortPolicy"         
        reason:  "FieldValueInvalid"         
       }
      ]
      group:  "stable.agones.dev"       
      kind:  "GameServer"       
      name:  "udp-serverswfxj"       
     }
     message:  "GameServer configuration is invalid"      
     metadata: {…}      
     reason:  "Invalid"      
     status:  "Failure"      
    }
    uid:  ""     
   }
  }
  source:  "*gameservers.Controller"   
 }

markmandel:
Ah fun! In previous examples, we've use a seperate `ValidateUpdate()` function, [like so](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/apis/stable/v1alpha1/gameserverset.go#L76) to validate on changes from one state to another.

Could we use that pattern here as well?

jeremyje:
ping review

pkg/apis/stable/v1alpha1/gameserver.go
```diff
@@ -268,6 +259,45 @@ func (gs *GameServer) Validate() (bool, []metav1.StatusCause) {
...
+}
+
+// ValidateUpdate validates the GameServer configuration after the initial GameServer has been created.
+// If a GameServer is invalid there will be > 0 values in
+// the returned array
+func (gs *GameServer) ValidateUpdate() (bool, []metav1.StatusCause) {
+	var causes []metav1.StatusCause
+
+	// no host port when using dynamic PortPolicy only if we are still allocating a port.
+	if gs.Status.State == GameServerStatePortAllocation || gs.Status.State == "" {
```
markmandel: So I'm trying to work out how this passes, since, when we `PortAllocate`, we [actually set the `HostPort` values on the `Spec.Ports`](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameservers/portallocator.go#L147). (Which, maybe we shouldn't do? There is some precedent of editing `Spec` in `Pod` with `Spec.NodeName`)

But I think this leads to a larger question - Should we allow editing of a `GameServer.Spec` at all?
Editing the Spec should really drive restart/reschedule of the backing Pod, so we can actually accomplish what the declaration is asking for. Also, what if the `GameServer` is `Allocated` and then the Spec changes. This seems scary to me.

But I can't think of a use case in which we should be doing this. GameServers are deleted when they want to be updated, and new versions are rolled through a Fleet.

So I'm leaning towards not allowing Spec to be changed at all on Update.

Anyone else have other thoughts?

Kuqd: make sense, this specific line needed a better comment also.

jeremyje: Thanks for the feedback. The reason for the bug caught by the e2e tests is that Update validation runs anytime an update is done to the CRD either by kubectl or through state changes as the controller reconciles. I'm not sure if there's a way to just prevent manual updates. I couldn't find that.

markmandel: So I see how it passes :+1: I guess my question is - should we (a) remove that update to the Spec.Ports and then (b) lock down `Spec` so it can't be updated at all during the GameServer lifecycle.

I'm leaning this way, just because I can't think of any reason we would want to allow changes to GameServer.Spec once it's been created.

markmandel: I think this works for 80% of cases, let me know if it doesn't - and also doesn't hardcode status values.

Switch the `ValidateUpdate` signature to match the other ones (also will need to validate against the old object, not the new one), and make the following change.

The TL;DR - you can only change ports, when the status.State is in flux. Otherwise you can't.
Here is a [patch.txt](https://github.com/GoogleCloudPlatform/agones/files/2830193/patch.txt), in case you want a quick way to apply the function (I've not done the webhook side or tested it)

There may be some bugs in here, but I hope you get the intent, and there is enough here to take it through, if the solution makes sense.

```go
// ValidateUpdate validates the GameServer configuration after the initial GameServer has been created.
// If a GameServer is invalid there will be > 0 values in
// the returned array
func (gs *GameServer) ValidateUpdate(new *GameServer) (bool, []metav1.StatusCause) {
	var causes []metav1.StatusCause

	// these values can never change
	if gs.Spec.Container != new.Spec.Container ||
		!equality.Semantic.DeepEqual(gs.Spec.Health, new.Spec.Health) ||
		gs.Spec.Scheduling != new.Spec.Scheduling ||
		!equality.Semantic.DeepEqual(gs.Spec.Template, new.Spec.Template) {
			Type:    metav1.CauseTypeFieldValueInvalid,
			Field:   "spec",
			Message: "spec cannot be updated",
		})
	}

	// port can be changed while status is in transition
	if gs.Status.State == new.Status.State &&
		len(causes) == 0 &&
		!equality.Semantic.DeepEqual(gs.Spec.Ports, new.Spec.Ports) {
			Type:    metav1.CauseTypeFieldValueInvalid,
			Field:   "spec",
			Message: "spec cannot be updated",
		})
	}

	return len(causes) == 0, causes
}
```

WDYT?

markmandel:
Coming back to this - we may want to hold off on this, as @aLekSer is working on bringing a status subresource to GameServers, which will likely tear our approaches asunder.

aLekSer:
Hello @markmandel . I think this PR could be pulled to master first and then I will apply it to GS Status PR (which is currently in draft state).
Note that I paused GS status subresource development, it makes E2E tests flaky and there are not so many gains out of separating them for now.

markmandel:
Happy to do that -- in which case,  @jeremyje I had an alternative approach that I proposed a while back - have you had a chance to take a look?

markmandel:
Any objection to me closing this? Seems like it's way behind the , and we'll likely move GameServer to have subresource support before 1.0, which will break things in this yet again.

If I don't hear any objections in 4 days, I'll go ahead and close this to clean things up.

markmandel:
Closing for cleanup.

PR #534, opened 2019-01-30
This is how you write shortcode in hugo

markmandel:
Make the contributing example easier to understand, since now I now know to escape shortcode examples.


jkowalski:
Noice.

PR #542, opened 2019-01-31
Speed up creation/deletion of game servers in a set

jkowalski:
- parallelizing create/delete calls in GSS controller
- creating dedicated worker queues for creation/deletion requests
- replacing default rate limiters with ones that don't have global QPS limit

That, plus general cleanup of GSS controller for readability.

Also, introduced GS state cache, so that actions taken by GSS controller are more precise because they don't rely on stale informer data.

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -18,6 +18,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"sync"
+	"time"
+
+	"k8s.io/client-go/util/workqueue"
 
 	"agones.dev/agones/pkg/apis/stable"
 	"agones.dev/agones/pkg/apis/stable/v1alpha1"
@@ -73,6 +76,8 @@ type Controller struct {
 	portAllocator          *PortAllocator
 	healthController       *HealthController
 	workerqueue            *workerqueue.WorkerQueue
+	creationWorkerQueue    *workerqueue.WorkerQueue // handles creation only
+	deletionWorkerQueue    *workerqueue.WorkerQueue // handles deletion only
 	allocationMutex        *sync.Mutex
 	stop                   <-chan struct {
 	}
@@ -124,20 +129,24 @@ func NewController(
 	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
-	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
+	c.workerqueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerController", fastRateLimiter())
+	c.creationWorkerQueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerControllerCreation", fastRateLimiter())
+	c.deletionWorkerQueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerControllerDeletion", fastRateLimiter())
 	health.AddLivenessCheck("gameserver-workerqueue", healthcheck.Check(c.workerqueue.Healthy))
+	health.AddLivenessCheck("gameserver-creation-workerqueue", healthcheck.Check(c.creationWorkerQueue.Healthy))
+	health.AddLivenessCheck("gameserver-deletion-workerqueue", healthcheck.Check(c.deletionWorkerQueue.Healthy))
 
 	wh.AddHandler("/mutate", v1alpha1.Kind("GameServer"), admv1beta1.Create, c.creationMutationHandler)
 	wh.AddHandler("/validate", v1alpha1.Kind("GameServer"), admv1beta1.Create, c.creationValidationHandler)
 
 	gsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: c.workerqueue.Enqueue,
+		AddFunc: c.enqueueGameServerBasedOnState,
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			// no point in processing unless there is a State change
 			oldGs := oldObj.(*v1alpha1.GameServer)
 			newGs := newObj.(*v1alpha1.GameServer)
 			if oldGs.Status.State != newGs.Status.State || oldGs.ObjectMeta.DeletionTimestamp != newGs.ObjectMeta.DeletionTimestamp {
-				c.workerqueue.Enqueue(newGs)
+				c.enqueueGameServerBasedOnState(newGs)
 			}
 		},
 	})
@@ -166,6 +175,31 @@ func NewController(
 	})
 
 	return c
+}
+
+func (c *Controller) enqueueGameServerBasedOnState(item interface{}) {
+	gs := item.(*v1alpha1.GameServer)
+
+	switch gs.Status.State {
+	case v1alpha1.GameServerStatePortAllocation,
+		v1alpha1.GameServerStateCreating:
+		c.creationWorkerQueue.Enqueue(gs)
+
+	case v1alpha1.GameServerStateShutdown:
+		c.deletionWorkerQueue.Enqueue(gs)
+
+	default:
+		c.workerqueue.Enqueue(gs)
+	}
+}
+
+// fastRateLimiter returns a fast rate limiter, without exponential back-off.
+func fastRateLimiter() workqueue.RateLimiter {
+	const numFastRetries = 5
+	const fastDelay = 20 * time.Millisecond  // first few retries up to 'numFastRetries' are fast
+	const slowDelay = 500 * time.Millisecond // subsequent retries are slow
+
+	return workqueue.NewItemFastSlowRateLimiter(fastDelay, slowDelay, numFastRetries)
 }
 
 // creationMutationHandler is the handler for the mutating webhook that sets the
@@ -268,7 +302,21 @@ func (c *Controller) Run(workers int, stop <-chan struct{}) error {
 	// Run the Health Controller
 	go c.healthController.Run(stop)
 
-	c.workerqueue.Run(workers, stop)
+	// start work queues
+	var wg sync.WaitGroup
+
+	startWorkQueue := func(wq *workerqueue.WorkerQueue) {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			wq.Run(workers, stop)
+		}()
+	}
+
+	startWorkQueue(c.workerqueue)
+	startWorkQueue(c.creationWorkerQueue)
+	startWorkQueue(c.deletionWorkerQueue)
+	wg.Wait()
 	return nil
 }
 
--- a/pkg/gameserversets/controller.go
+++ b/pkg/gameserversets/controller.go
@@ -16,6 +16,7 @@ package gameserversets
 
 import (
 	"encoding/json"
+	"sort"
 	"sync"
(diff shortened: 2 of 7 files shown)


pkg/gameserversets/gameserver_state_cache_test.go
```diff
@@ -0,0 +1,199 @@
+package gameserversets
+
+import (
+	"sort"
+	"testing"
+
```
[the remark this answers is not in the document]
jkowalski: done

pkg/gameserversets/controller.go
```diff
@@ -238,135 +256,271 @@ func (c *Controller) syncGameServerSet(key string) error {
...
+		// we have to pick which GS to delete, let's delete the newest ones first.
+		sort.Slice(toDelete, func(i, j int) bool {
+			return toDelete[i].ObjectMeta.CreationTimestamp.After(toDelete[j].ObjectMeta.CreationTimestamp.Time)
+		})
+
+		toDelete = toDelete[0:maxDeletions]
+		partialReconciliation = true
+	}
+
+	return //nolint:nakedret
```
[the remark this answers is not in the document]
jkowalski: don't know linter does not like large functions, I guess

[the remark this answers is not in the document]
markmandel: Yeah, the linter will catch you with high Cyclomatic complexity.

pkg/gameservers/controller.go
```diff
@@ -168,6 +177,24 @@ func NewController(
...
+func (c *Controller) enqueueGameServerBasedOnState(item interface{}) {
+	gs := item.(*v1alpha1.GameServer)
+
+	switch gs.Status.State {
+	case v1alpha1.GameServerStatePortAllocation,
+		v1alpha1.GameServerStateCreating,
+		v1alpha1.GameServerStateStarting:
+		c.creationWorkerQueue.Enqueue(gs)
+
+	case v1alpha1.GameServerStateUnhealthy,
```
markmandel: Did you mean for this to be `Shutdown` rather than `Unhealthy`?

The gameserver controller doesn't do anything to GameServers that are `Unhealthy`

pkg/gameservers/controller.go
```diff
@@ -124,20 +128,25 @@ func NewController(
 	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
-	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
+	c.workerqueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerController",
```
markmandel: So they all have the same settings, and the same sync - I'm dreadfully curious - how does this improve things in the `GameServer` controller?

Is it just that creation and deletion get's pushed to the front of it's own queue, while there rest can happen at a separate priority? (or at the same time?)

pkg/gameserversets/controller.go
```diff
@@ -238,135 +256,273 @@ func (c *Controller) syncGameServerSet(key string) error {
...
-	if err := c.syncLessGameServers(gsSet, diff); err != nil {
-		return err
+	c.logger.
+		WithField("targetReplicaCount", gsSet.Spec.Replicas).
+		WithField("numServersToAdd", numServersToAdd).
+		WithField("numServersToDelete", len(toDelete)).
+		WithField("isPartial", isPartial).
+		WithField("status", status).
+		WithFields(fields).
+		Infof("Reconciling GameServerSet")
```
markmandel:
```suggestion
		Info("Reconciling GameServerSet")
```

pkg/gameserversets/controller.go
```diff
@@ -238,135 +256,273 @@ func (c *Controller) syncGameServerSet(key string) error {
...
+
+		case v1alpha1.GameServerStateError, v1alpha1.GameServerStateUnhealthy:
+			scheduleDeletion(gs)
+		default:
+			// unrecognized state, assume it's up.
+			handleGameServerUp(gs)
+		}
+	}
+
+	// log.Printf("%v/%v %v/%v", upCount, targetReplicaCount, podPendingCount, maxPending)
```
markmandel: we can drop this line

pkg/gameserversets/controller.go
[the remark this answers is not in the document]
markmandel: Can we remove it then? We can always add a ignore for cyclo if we feel it's necessary?

pkg/gameserversets/controller.go
```diff
@@ -238,135 +256,273 @@ func (c *Controller) syncGameServerSet(key string) error {
...
+	}
+
+	return //nolint:nakedret
 }
 
-// syncMoreGameServers adds diff more GameServers to the set
-func (c *Controller) syncMoreGameServers(gsSet *v1alpha1.GameServerSet, diff int32) error {
-	if diff <= 0 {
-		return nil
+func isAllocated(gs *v1alpha1.GameServer) bool {
```
markmandel: A bit simpler?

```go
return gs.ObjectMeta.DeletionTimestamp == nil && gs.Status.State == v1alpha1.GameServerStateAllocated
```

pkg/gameservers/controller.go
jkowalski: that's exactly correct. GSS can then add/delete lots of GS very quickly without clogging the processing of GS already in-flight.

pkg/gameservers/controller.go
```diff
@@ -124,20 +128,25 @@ func NewController(
...
+		workqueue.NewItemFastSlowRateLimiter(20*time.Millisecond, 500*time.Millisecond, 5))
+	c.creationWorkerQueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerControllerCreation",
```
ilkercelikyilmaz: Should we consider having separate method for each workerqueue instead of syncGameServer method? Each method can only focus on the possible events rather than trying to process all.

jkowalski: queue is just a "dirty" bit, it's totally possible that the state of a resource will change underneath between enqueue and processing, so we should not assume anything about the state of each GS.

pkg/gameserversets/controller.go
jkowalski: done

pkg/gameserversets/controller.go
jkowalski: done

pkg/gameserversets/controller.go
jkowalski: done

pkg/gameservers/controller.go
jkowalski: both, actually

pkg/gameservers/controller.go
ilkercelikyilmaz: I understand your concern about queue. However the new "enqueueGameServerBasedOnState" method should put the messages to the appropriate queue. I think except syncGameServerDeletionTimestamp handler, every other handler ignore unexpected events.

pkg/gameserversets/controller.go
```diff
@@ -238,135 +256,262 @@ func (c *Controller) syncGameServerSet(key string) error {
...
+	return c.syncGameServerSetStatus(gsSet, list)
 }
 
-// syncUnhealthyGameServers deletes any unhealthy game servers (that are not already being deleted)
-func (c *Controller) syncUnhealthyGameServers(gsSet *v1alpha1.GameServerSet, list []*v1alpha1.GameServer) error {
+// computeReconciliationAction computes the action to take to reconcile a game server set set given
+// the list of game servers that were found and target replica count.
+func computeReconciliationAction(list []*v1alpha1.GameServer, targetReplicaCount int, maxCreations int, maxDeletions int, maxPending int) (numServersToAdd int, toDelete []*v1alpha1.GameServer, partialReconciliation bool) {
+	var upCount int         // up == Ready or will become ready
+	var podPendingCount int // podPending == "up" but don't have a Pod running yet
```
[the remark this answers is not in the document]
jkowalski: done

pkg/gameserversets/controller.go
```diff
@@ -238,135 +256,262 @@ func (c *Controller) syncGameServerSet(key string) error {
...
+	// pass 2 - handle all other statuses
+	for _, gs := range list {
+		if isAllocated(gs) {
+			// already handled above
+			continue
+		}
+
+		// GS being deleted counts towards target replica count.
+		if !gs.ObjectMeta.DeletionTimestamp.IsZero() {
+			//handleGameServerUp(gs)
```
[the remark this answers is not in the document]
jkowalski: done

pkg/gameserversets/controller.go
```diff
@@ -109,9 +123,12 @@ func NewController(
 			oldGss := oldObj.(*v1alpha1.GameServerSet)
 			newGss := newObj.(*v1alpha1.GameServerSet)
 			if oldGss.Spec.Replicas != newGss.Spec.Replicas {
-				c.workerqueue.Enqueue(newGss)
+				c.workerqueue.EnqueueImmediately(newGss)
```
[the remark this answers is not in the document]
jkowalski: reverted, this does not need to be immediate.

pkg/gameservers/controller.go
```diff
@@ -124,20 +128,25 @@ func NewController(
 	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
-	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
+	c.workerqueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerController",
+		workqueue.NewItemFastSlowRateLimiter(20*time.Millisecond, 500*time.Millisecond, 5))
```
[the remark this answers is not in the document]
jkowalski: extracted a function with local consts

pkg/gameservers/controller.go
```diff
@@ -124,20 +129,22 @@ func NewController(
 	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
 	c.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "gameserver-controller"})
 
-	c.workerqueue = workerqueue.NewWorkerQueue(c.syncGameServer, c.logger, stable.GroupName+".GameServerController")
+	c.workerqueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerController", fastRateLimiter())
+	c.creationWorkerQueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerControllerCreation", fastRateLimiter())
+	c.deletionWorkerQueue = workerqueue.NewWorkerQueueWithRateLimiter(c.syncGameServer, c.logger, stable.GroupName+".GameServerControllerDeletion", fastRateLimiter())
 	health.AddLivenessCheck("gameserver-workerqueue", healthcheck.Check(c.workerqueue.Healthy))
```
markmandel: Should the health check here be tied to all 3 workqueues, rather than just the one?

pkg/gameservers/controller.go
```diff
@@ -18,6 +18,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"sync"
+	"time"
+
+	"k8s.io/client-go/util/workqueue"
```
markmandel: I wish goimport didn't leave these lines in between. Nit to remove.

Issue #547, opened 2019-02-01
Grafana: add graph of nodes in cluster


jkowalski:
* total # of nodes in a cluster (to see when node autoscaler kicks in)
* total # of nodes that have any game servers in a cluster
* total # of empty nodes in a cluster

Perhaps more general:

* distribution of # of game servers per node - to be able to see min, max, avg, 50th, 90th and 99th, percentile

PR #551, opened 2019-02-01
Adds gameservers per node count and distribution

Kuqd:
Addresses #547 

I've included gameservers per nodes (count and distribution) in our `Agones GameServers` dashboard and move everything related to allocations to `Agones Allocations`.

### DIFF
--- a/pkg/metrics/controller.go
+++ b/pkg/metrics/controller.go
@@ -20,6 +20,8 @@ import (
 	"sync"
 	"time"
 
+	v1 "k8s.io/client-go/listers/core/v1"
+
 	stablev1alpha1 "agones.dev/agones/pkg/apis/stable/v1alpha1"
 	"agones.dev/agones/pkg/client/clientset/versioned"
 	"agones.dev/agones/pkg/client/informers/externalversions"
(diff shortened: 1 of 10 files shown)


pkg/metrics/controller_metrics.go
```diff
@@ -111,6 +113,19 @@ var (
 			Aggregation: view.Count(),
 			TagKeys:     []tag.Key{keyType, keyFleetName},
 		},
+		&view.View{
```
jkowalski: not sure if this is needed, the distribution bucket #0 will have the same value.

Kuqd: Can I get the count of nodes currently in the cluster using the distribution ? I can get the average using _sum and _count but not sure about instant value.

jkowalski: You can’t since the sum is ever increasing. A gauge is required.

Kuqd: Ok so I’ll keep the view.

pkg/metrics/controller.go
```diff
@@ -331,9 +343,44 @@ func (c *Controller) collectGameServerCounts() {
...
+		c.logger.WithError(err).Warn("failed listing gameservers")
+		return
+	}
+	recordWithTags(context.Background(), []tag.Mutator{tag.Insert(keyEmpty, "true")},
+		nodesCountStats.M(int64(len(nodes)-len(gsPerNodes))))
+	recordWithTags(context.Background(), []tag.Mutator{tag.Insert(keyEmpty, "false")},
+		nodesCountStats.M(int64(len(gsPerNodes))))
+
+	for _, node := range nodes {
+		if c, ok := gsPerNodes[node.Name]; ok {
```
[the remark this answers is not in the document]
Kuqd: Interesting you found a bug here ! I was counting more than required, when I made the change metrics were different and correct.

PR #554, opened 2019-02-04
Fix stackdriver distribution without bucket bounds

aLekSer:
New k8s metrics results in error in Stackdriver exporter. In order to
fix it we need to define two infinite buckets with a delimiter of 0.
Fix was verified in Stackdriver dashboard.

Fix was proposed [here](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/issues/41#issuecomment-421154451)
```
The documentation is here: https://godoc.org/go.opencensus.io/stats/view#Distribution . 
The problem is although OpenCensus allows Distribution with no bucket bounds,
 Stackdriver doesn't support it.
```


aLekSer:
I have made a rebase and rebuild with the latest master, still able to see metrics in Stackdriver.

Issue #469, opened 2019-01-02
Register GameServers with local IP addresses


markmandel:
## Problem
For development/QA purposes, it is useful to be able to register a gameserver with Agones, but overwrite it with a specific local IP address, so that it can run through the same Agones system, but QA testers can connect to it, without having to do a full build + deploy pipeline.

## Design
We assume all port configuration should be `static`.
If the `GameServer` has an annotation of: `stable.agones.dev/dev-address: "192.168.0.1"` then we intercept this in the controller:
1. Don't create a Pod
2. Use this as the `Status.State` address.
3. Static port configurations are moved directly into `Status.Ports`
3. Automatically mark this `GameServer` as `Ready`.

## Example:

```yaml
apiVersion: "stable.agones.dev/v1alpha1"
kind: GameServer
metadata:
  name: "xonotic"
  annotations:
    stable.agones.dev/dev-address: "192.168.0.1"
spec:
  ports:
    - name: default
      portPolicy: static
      containerPort: 26000
      hostPort: 26000
  template:
    spec:
      containers:
        - name: xonotic
          image: gcr.io/agones-images/xonotic-example:0.4
```

## Questions
- Do we like the name of the annotation?

ReDucTor:
Looks good, would simplify some of our alternative approaches to the problem.

jeremyje:
I'm looking into this.

jeremyje:
I'm still working through the bug, I have a test and the logic implemented here, https://github.com/jeremyje/agones/tree/local-ip.

I have a question though,
When you say don't create a pod. Does that mean the QA server isn't managed by Agones/Kubernetes and that the goal is to mainly register the existence of the QA server? Bypass all those liveness checks and all that.

markmandel:
Exactly. The active game server likely sits on someone's machine at their desk, and this gives them a way to basically "fake" a Ready gameserver on an Agones cluster, that someone can then connect to within their corp-network.

Then all your other tooling (matchmaking, etc) can still work as per normal - you can even allocate this GameServer - but it makes life easier for development.

jeremyje:
Wouldn't you want this server to be marked as Allocated then?

ReDucTor:
Marking it as already allocated would be game specific

markmandel:
:+1: on what @ReDucTor said - I think in _most_ cases you would want it to be ready, so you can flow through from your matchmaker to your game server just like you would normally.

[...]

PR #558, opened 2019-02-04
Add support to create a development gameserver.

jeremyje:
Implementation of #469

### DIFF
--- a/examples/simple-udp/dev-gameserver.yaml
+++ b/examples/simple-udp/dev-gameserver.yaml
@@ -1,0 +1,34 @@
+# Copyright 2019 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: "stable.agones.dev/v1alpha1"
+kind: GameServer
+metadata:
+  name: local-simple-udp
+  generateName: "simple-udp-"
+  annotations:
+    # Causes Agones to register your local game server at 192.1.1.2, replace with your server's IP address.
+    stable.agones.dev/dev-address: "192.1.1.2"
+spec:
+  ports:
+  - name: default
+    portPolicy: "static"
+    hostPort: 17654
+    containerPort: 17654
+  # The following is ignored but required due to validation.
+  template:
+    spec:
+      containers:
+      - name: simple-udp
+        image: gcr.io/agones-images/udp-server:0.5
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -357,6 +357,9 @@ func (c *Controller) syncGameServer(key string) error {
 	if gs, err = c.syncGameServerRequestReadyState(gs); err != nil {
 		return err
 	}
+	if gs, err = c.syncDevelopmentGameServer(gs); err != nil {
+		return err
+	}
 	if err = c.syncGameServerShutdownState(gs); err != nil {
 		return err
 	}
@@ -434,6 +437,9 @@ func (c *Controller) syncGameServerCreatingState(gs *v1alpha1.GameServer) (*v1al
 	if !(gs.Status.State == v1alpha1.GameServerStateCreating && gs.ObjectMeta.DeletionTimestamp.IsZero()) {
 		return gs, nil
 	}
+	if _, isDev := gs.GetDevAddress(); isDev {
+		return gs, nil
+	}
 
 	c.logger.WithField("gs", gs).Info("Syncing Create State")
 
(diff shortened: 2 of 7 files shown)


examples/simple-udp/dev-gameserver.yaml
```diff
@@ -0,0 +1,39 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
```
markmandel: 2019

pkg/gameservers/controller.go
```diff
@@ -416,6 +422,33 @@ func (c *Controller) syncGameServerCreatingState(gs *v1alpha1.GameServer) (*v1al
...
+	gsCopy := gs.DeepCopy()
+	ports := []v1alpha1.GameServerStatusPort{}
+	for _, p := range gs.Spec.Ports {
+		ports = append(ports, p.Status())
+	}
+	gsCopy.Status.State = v1alpha1.GameServerStateReady
+	gsCopy.Status.Ports = ports
+	gsCopy.Status.Address = devIPAddress
+	gsCopy.Status.NodeName = devIPAddress
+	gs, err := c.gameServerGetter.GameServers(gs.ObjectMeta.Namespace).Update(gsCopy)
```
markmandel: Shall we take our new, fancy `UpdateStatus` for a spin here instead?
https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/client/clientset/versioned/typed/stable/v1alpha1/gameserver.go#L40

We'll want to move everything over eventually, so might as well do new work with it.

(Pretty sure you're only updating status here?)

markmandel: @aLekSer - sounds like we can't actually do this one? So I may have to rescind this suggestion.

We would have to wait for your status subresource updates?

jeremyje: I'll add a TODO for this.

aLekSer: Yes, I think this change could be performed as a separate PR. Cause we need to change Helm template and this could lead to changes in other places too (RBAC permissions for "gameservers/status").
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/customresources-subresources.md

site/content/en/docs/Guides/development-server.md
```diff
@@ -0,0 +1,38 @@
+---
```
markmandel: Let's add a `publishDate: 2018-02-19` to the top content so that it doesn't show up before the stable release comes out.

@ReDucTor can you take a look, I'm not sure the phrasing here is right. Maybe it should be something more like:
"Coordinate Agones with a local development server" or something? I'm not quite sure what the documentation should say -- I figured you might have some good words.

We aren't actually hosting it in Agones - we're more coordinating it with a local build?

site/content/en/docs/Guides/development-server.md
```diff
@@ -0,0 +1,38 @@
...
+
+To register your development game server you'll need to know the IP address of the machine running it and the port. With that you'll create a game server config like the one below.
+
+```yaml
+apiVersion: "stable.agones.dev/v1alpha1"
+kind: GameServer
+metadata:
+  name: my-development-server
+  annotations:
+    stable.agones.dev/dev-address: "192.1.1.2" # Replace with the IP address of your development game server.
```
markmandel: Let's explicitly call out that this annotation needs to be added - that this is the magic sauce that makes it all happen :+1:

site/content/en/docs/Guides/development-server.md
jeremyje: It also turns out that we need a bunch of the pod spec template stuff to make this all work I'm adding comments for now to say it's ignored. I've been trying to experiment with not requiring it but I have to rip out a lot of validation that's in the yaml files to inch towards that. I'm now experimenting with just putting placeholder values for now.

markmandel: Saying it's ignored seems reasonable :+1:

ReDucTor: The only thing that causes a bit of confusion I believe is the phrase "Development mode" which makes it seem like its a feature toggle within all of Agones (even though it is in some ways). An alternative might be "Development game server"

jeremyje: What should this feature be called in the docs?

* Development Mode
* Unmanaged Mode
* Local Mode

etc?

jeremyje: Good news: I got the PodSpecTemplate to be optional in my experimental branch.
Bad news: It requires a bug fix in jsonpatch to get it working. I'll see if I put up a PR for that project.

markmandel: This is because we define it as required in the OpenAPI spec, yeah?

markmandel: Could we call the document "Overwrite GameServer IP and Port" ? - which is literally what it is?

And then in the document, outline some use case(s) - such as running a development game server on your local machine, but exposing it to you QA Agones cluster by registering it in this special way - ? WDYT?

jeremyje: I'd put a vote for "unmanaged mode".
I think it's clearer to align with a use case (intent) rather than explain what the logic is doing.

Unmanaged is generic enough to cover these cases like the QA workflow.

jeremyje: On the topic of jsonpatching and clearing out all the extra fields.

Also I looked into the json patch stuff. It's actually a bit more complicated than originally thought and probably past being worth the time now. It looks like a subtle bug between the 2 different json libraries that we use. I think it's just better to move away from the mattbierd library and focus on using the evanphx one that Kubernetes already depends on.

markmandel: AFAIK evanphx will not generate patches. Unless I missed something?

jeremyje: https://github.com/evanphx/json-patch#create-and-apply-a-json-patch

markmandel: This is a "Guide" - which is a howto, essentially, so thinking of it as an action - i.e. what is the user trying to do here?

It doesn't need a "mode" per se - it just needs to describe what we are doing.

- "Inject local Game Server into Agones"
- "Point Agones at local game server"
- "Manage local game server with Agones"

Something like that?

jeremyje: Register Local Game Server with Agones?

markmandel: I like it! :+1:

jeremyje: Updated

markmandel:
Assuming this all passes, this LGTM. The only thing would be, to please rebase this down to a single commit.
https://github.com/GoogleCloudPlatform/agones/blob/master/CONTRIBUTING.md#submitting-code-via-pull-requests

:+1: :+1: :+1: :+1: :+1: :+1:

PR #560, opened 2019-02-05
CloudBuild for a "development" subdomain

markmandel:
Once this is live, we can setup a development.agones.dev that has the latest version of site from master live at all times.

This also makes documentation for release candidates easier, as we can point them at this development release site.


markmandel:
Not yet. I need the deployed `development` service (which this does), then I can create the subdomain to match up to the service, and then we should be good to go (and then I can add a dropdown to the site so you can flip between them!)

markmandel:
Don't merge this yet - I messed something up in the site/cloudbuild.yaml

PR #565, opened 2019-02-06
Changed how GameServer POD names are generated

### LINKED ISSUE #490: Investigate why we sometimes have multiple pods per gameserver
During large cluster turnup I ran into this:

```shell
$ kubectl get gs
NAME                     STATE          ADDRESS   PORT   NODE   AGE
simple-udp-sv722-gj985   RequestReady                           27m
simple-udp-sv722-m7rk2   RequestReady                           27m

$ kubectl get pod
NAME                           READY   STATUS    RESTARTS   AGE
simple-udp-sv722-gj985-csgr7   2/2     Running   0          27m
simple-udp-sv722-gj985-rw6p7   2/2     Running   0          27m
simple-udp-sv722-m7rk2-g8rn8   2/2     Running   0          27m
simple-udp-sv722-m7rk2-q44mm   2/2     Running   0          27m
```
In a large turnup (10K GS) this reproes quite often - typically one or two of the GS get stuck in this state.

jkowalski:
Instead of generating name based on GS prefix, we take GS name as-is.
This is safe, since Pods are owned by Gameservers, so their lifetime is a subset of that of GS, thus we won't have accidental pod name collisions.

This fixes #490

### DIFF
--- a/pkg/apis/stable/v1alpha1/gameserver.go
+++ b/pkg/apis/stable/v1alpha1/gameserver.go
@@ -333,10 +333,10 @@ func (gs *GameServer) Pod(sidecars ...corev1.Container) (*corev1.Pod, error) {
 
 // podObjectMeta configures the pod ObjectMeta details
 func (gs *GameServer) podObjectMeta(pod *corev1.Pod) {
-	// Switch to GenerateName, so that we always get a Unique name for the Pod, and there
-	// can be no collisions
-	pod.ObjectMeta.GenerateName = gs.ObjectMeta.Name + "-"
-	pod.ObjectMeta.Name = ""
+	pod.ObjectMeta.GenerateName = ""
+	// Pods inherit the name of their gameserver. It's safe since there's
+	// a guarantee that pod won't outlive its parent.
+	pod.ObjectMeta.Name = gs.ObjectMeta.Name
 	// Pods for GameServers need to stay in the same namespace
 	pod.ObjectMeta.Namespace = gs.ObjectMeta.Namespace
 	// Make sure these are blank, just in case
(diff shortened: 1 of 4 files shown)


pkg/apis/stable/v1alpha1/gameserver.go
```diff
@@ -333,10 +333,10 @@ func (gs *GameServer) Pod(sidecars ...corev1.Container) (*corev1.Pod, error) {
...
 // podObjectMeta configures the pod ObjectMeta details
 func (gs *GameServer) podObjectMeta(pod *corev1.Pod) {
-	// Switch to GenerateName, so that we always get a Unique name for the Pod, and there
-	// can be no collisions
-	pod.ObjectMeta.GenerateName = gs.ObjectMeta.Name + "-"
-	pod.ObjectMeta.Name = ""
+	pod.ObjectMeta.GenerateName = ""
+	// Pods inherit the name of their gameserver. It's safe since there's
+	// a guarantee that pod won't outlive its parent.
+	pod.ObjectMeta.Name = gs.ObjectMeta.Name
```
ilkercelikyilmaz: I tested this locally and it seems work.

markmandel:
We will need to add a section [here](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameservers/controller.go#L485) in the gameservers controller, such that if `k8serrors.IsAlreadyExists(err)` - then continue through, otherwise we're going to end up in an infinite backoff if the pod already exists, as the error will keep getting rethrown.

jkowalski:
> We will need to add a section [here](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameservers/controller.go#L485) in the gameservers controller, such that if `k8serrors.IsAlreadyExists(err)` - then continue through, otherwise we're going to end up in an infinite backoff if the pod already exists, as the error will keep getting rethrown.

Done.

PR #572, opened 2019-02-07
Remove the mutex usage for Delete GS in both GS and GSS controllers

ilkercelikyilmaz:
AllocationMutex usage is unnecessary for Delete GS since the Shutdown state of GS prevents race conditions. Also remove direct Delete of GS and change it with setting the status of GS to Shutdown and let the GS controller to take care of it.

### DIFF
--- a/pkg/gameserversets/controller.go
+++ b/pkg/gameserversets/controller.go
@@ -440,11 +440,12 @@ func (c *Controller) deleteGameServers(gsSet *v1alpha1.GameServerSet, toDelete [
 	c.logger.WithField("diff", len(toDelete)).WithField("gameserverset", gsSet.ObjectMeta.Name).Info("Deleting gameservers")
 
 	return parallelize(gameServerListToChannel(toDelete), maxDeletionParallelism, func(gs *v1alpha1.GameServer) error {
-		c.allocationMutex.Lock()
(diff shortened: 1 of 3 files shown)


pkg/gameserversets/controller.go
```diff
@@ -440,11 +440,12 @@ func (c *Controller) deleteGameServers(gsSet *v1alpha1.GameServerSet, toDelete [
 	c.logger.WithField("diff", len(toDelete)).WithField("gameserverset", gsSet.ObjectMeta.Name).Info("Deleting gameservers")
 
 	return parallelize(gameServerListToChannel(toDelete), maxDeletionParallelism, func(gs *v1alpha1.GameServer) error {
-		c.allocationMutex.Lock()
-		err := c.gameServerGetter.GameServers(gs.Namespace).Delete(gs.ObjectMeta.Name, nil)
-		c.allocationMutex.Unlock()
+		// We should not delete the gameservers directly buy set their state to shutdown and let the gameserver controller to delete
```
jkowalski: perhaps explain the Allocate/Delete race this is resolving

ilkercelikyilmaz: GS Allocate and Delete activities were being synced across all the controllers. This is/will adversely impacting the performance. Since the delete is only happening here for the GS with Shutdown state, those can never be allocated hence no need to sync.

jkowalski:
We need to squash all commits into one, but otherwise LGTM.

PR #575, opened 2019-02-08
Add status subresource to fleet and Gameserverset

aLekSer:
Switch to use `UpdateStatus` function for Fleets when it is necessary, add RBAC
permissions for "fleets/status" and "gameserversets/status".


install/yaml/install.yaml
```diff
@@ -48,13 +48,13 @@ rules:
   resources: ["customresourcedefinitions"]
   verbs: ["get"]
 - apiGroups: ["stable.agones.dev"]
-  resources: ["gameservers", "gameserversets"]
+  resources: ["gameservers", "gameserversets", "gameserversets/status"]
```
markmandel: Is there opportunity here to get more restrictive with our rbac permissions?

I mean - does the controller need to be able to update a gameserverset for example?

install/yaml/install.yaml
```diff
@@ -48,13 +48,13 @@ rules:
...
   verbs: ["create", "delete", "get", "list", "update", "watch"]
 - apiGroups: ["stable.agones.dev"]
   resources: ["gameservers"]
   verbs: ["patch"]
 - apiGroups: ["stable.agones.dev"]
-  resources: ["fleets", "fleetallocations", "fleetautoscalers"]
+  resources: ["fleets", "fleets/status", "fleetallocations", "fleetautoscalers"]
```
markmandel: Same as above - could we lock this down a little tighter?

install/yaml/install.yaml
aLekSer: I will try to restrict them

aLekSer: @markmandel I have restricted new subresource permissions.
However cannot restrict `gameserversets` permissions. In fleets and fleetautoscalers controllers we are using main Resource endpoint `gsSetCopy.Spec.Replicas` and Spec.Replicas of the fleet:
https://github.com/GoogleCloudPlatform/agones/blob/0e92c90b9f0da86ab34cae53e74382113f56bf6e/pkg/fleets/controller.go#L268
https://github.com/GoogleCloudPlatform/agones/blob/0e92c90b9f0da86ab34cae53e74382113f56bf6e/pkg/fleetautoscalers/controller.go#L220

aLekSer:
The small regression after applying this changes:
If the spec.Replicas in GSS was 0 initially then there are empty values on Status.Replicas , Ready and Allocated:
```
NAME                                                        SCHEDULING   DESIRED   CURRENT   ALLOCATED   READY     AGE                                                                                   
gameserverset.stable.agones.dev/scale-fleet-0-45ksj-5qrkp   Packed       30        30        0           30        25s 
gameserverset.stable.agones.dev/scale-fleet-1-szfqc-g5b8g   Packed       0                                         25s
```

markmandel:
@aLekSer I'm guessing this is because there is no default values for those status items on creation? I'd be tempted to ask \#sig-apimachinery to see why 0 values come as blank? (although not quite sure what the reason for this is -- feels like a k8s bug?)

I'm wondering if its something we need to worry about it?  WDYT?

aLekSer:
I think this is because status fields are not defined in yaml file and the only place Status are changed (inited) is here:
https://github.com/GoogleCloudPlatform/agones/blob/3a0e6d1c335413ee30dd8ebce555e18bd8620d75/pkg/gameserversets/controller.go#L557
I am looking on a fix, without it we would see empty printing columns if Spec.Replicas was always 0.

aLekSer:
Some new error on a test step:
rm: cannot remove './public': No such file or directory
includes/website.mk:35: recipe for target 'site-static' failed
mkdir /workspace/site/public
make[2]: [site-static] Error 1 (ignored)

markmandel:
> I am looking for a fix, without it we would see empty printing columns if Spec.Replicas was always 0.

Is the issue that the value is 0, or that status hasn't been set at all?

If you scale a fleet up to 5, and then back down to 0, does it go back to being blank?

I'm wondering if its because the status hasn't been created yet?

aLekSer:
Hello @markmandel , 
I have fixed the issue. As you can see I added one more updateStatus on creating `GameServerSet` from fleet controller.
>Is the issue that the value is 0, or that status hasn't been set at all?
Status were not set at all, but now it is set on creating GSS with aditional step (above).
>If you scale a fleet up to 5, and then back down to 0, does it go back to being blank?
Now it always set right after create of GSS.
>I'm wondering if its because the status hasn't been created yet?
Yes, that was the case, no all should be fine with yesterday fix.

pkg/fleets/controller.go
```diff
@@ -291,11 +291,24 @@ func (c *Controller) syncFleet(key string) error {
...
 			return errors.Wrapf(err, "error creating gameserverset for fleet %s", fleet.ObjectMeta.Name)
 		}
 
+		// extra step which is needed to set
+		// default values for GameServerSet Status Subresource
+		gsSetCopy := gsSet.DeepCopy()
+		gsSetCopy.Status.ReadyReplicas = 0
+		gsSetCopy.Status.Replicas = 0
+		gsSetCopy.Status.AllocatedReplicas = 0
+		_, err = gsSets.UpdateStatus(gsSetCopy)
```
aLekSer: Here is extra step which set default values.

PR #579, opened 2019-02-09
Removal of allocationMutex from controllers that don't need it.

markmandel:
Just a minor cleanup I saw when working on allocation processing.


ilkercelikyilmaz:
Thanks, I was planning to do the clean-up from all the controllers once removed mutex from allocate

PR #581, opened 2019-02-10
Add Node.js SDK and example - closes #538

### LINKED ISSUE #538: Create Node.js library
We are currently using the REST API which works well but it may simplify things for new users to have a library to use with Node.js. When deploying a new Agones version with an API change, a developer would only have to update their library version. Ideally this would be published as a node module as this is idiomatic for Node.js.

There has been some discussion before about creating a gRPC client in Node.js, or the library could make REST calls directly. The gRPC route should result in generated code so would be more maintainable.

steven-supersolid:
This includes
* Generated SDK pb files
* SDK wrapper
* Tests
* Example project that uses the SDK

The example project was tested using local SDK server

Site documentation has not been added yet


steven-supersolid:
I signed it!

sdks/nodejs/README.md
```diff
@@ -0,0 +1,24 @@
+### Generate the client library
+
+```
+npm install -g grpc-tools
```
markmandel: If I read this right - is this a developer tooling readme?

steven-supersolid: Yes, I just threw all my developer notes in one file so not sure if all required or if we should split documentation.
The actual usage instructions are not included, although the example shows usage it could perhaps benefit from some basic instructions, such as run `npm install` followed by `npm run`

sdks/nodejs/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/GoogleCloudPlatform/agones.git"
+  },
+  "scripts": {
+    "cover": "nyc jasmine",
+    "lint": "eslint --fix src",
+    "test": "jasmine"
+  },
+  "version": "0.8.0"
```
steven-supersolid: We should perhaps think about versioning. E.g. should the SDK version match the agones version? What happens when we release a new version of agones with no SDK change - do we bump the version number and republish to npm? How do we handle patches if a bug is discovered in the SDK?

sdks/nodejs/src/agonesSDK.js
```diff
@@ -0,0 +1,101 @@
...
+					resolve(response.toObject());
+				}
+			});
+		});
+	}
+
+	health() {
+		if (!this.healthStream) {
+			this.healthStream = this.client.health((error) => {
+				if (error) {
```
steven-supersolid: It wasn't clear to me whether this callback ever includes a response, and if it did how to handle that as the callback may be executed some time after the initial call to `agonesSDK.health()`. It's also unclear where this error could be caught if it did not happen in the call chain of the first call to `agonesSDK.health()`.
Another option I considered is to move the stream setup to the constructor.

sdks/nodejs/src/agonesSDK.js
```diff
@@ -0,0 +1,101 @@
...
+					resolve(response.toObject());
+				}
+			});
+		});
+	}
+
+	watchGameServer(callback) {
+		let request = new messages.Empty();
+		let emitter = this.client.watchGameServer(request);
+		emitter.on('data', (data) => {
```
steven-supersolid: I do not handle the 'end' event as not sure what to do with it and did not see it fired during testing

sdks/nodejs/README.md
steven-supersolid: This is the first time I used gRPC and feel I had to resort to a couple of hacks to get something that worked so would appreciate help on this.
Potential hacks:
1. Manually cloning the googleapis repo and including in path
2. Manually removing the lines from `_pb.js` files that import/require `annotations_pb.js` - I couldn't figure out how to generate these automatically and have in the correct location. Could probably have generated from the googleapis project and copied

What's weird is the import is not used in the generated js, but I cannot remove from the .proto as it is required there. This could also be a bug with js code generation

sdks/nodejs/package.json
steven-supersolid: Npm modules are the standard way to install and update SDKs in Node.js.

Publishing should be as simple as executing `npm publish`, so could be automated although you can never unpublish or republish over a version number. Publishing will require a npm account and publish token can be used by the build system if there is a way to keep this secret (this is not something I've automated as we push manually to internal libraries).

Perhaps we can get some pointers from other Google open source Node.js projects?
https://www.npmjs.com/package/googleapis (https://github.com/googleapis/google-api-nodejs-client)

sdks/nodejs/src/agonesSDK.js
steven-supersolid: Execution is single threaded so this method should be OK. I'm just concerned that although `this.client.health()` returns immediately, it might hang on to the callback and call with an error later. I can take another look and potentially remove

steven-supersolid:
Regarding it working, it works against the local dev server `sdk-server.windows.amd64 --local` but I've not tried it against our existing project so that would be worth doing. We could also mark as experimental until it has been battle tested.

sdks/nodejs/README.md
markmandel: Yeeep - yep, we don't want to be doing that :) (manually removing lines from generated code) - but also want to make sure this can continue to work as we upgrade grpc versions, etc.

grpc generation does generate both client and server side code -- so it's not surprising you are getting both.

You can see in our build image, we have scripting to pull in the googleapis information from the right place, and incorporate it where it needs to be:
https://github.com/GoogleCloudPlatform/agones/blob/master/build/build-image/gen-grpc-go.sh
https://github.com/GoogleCloudPlatform/agones/blob/master/build/build-image/gen-grpc-cpp.sh

Does that help?

sdks/nodejs/package.json
markmandel: Probably first step - let's get some source code that people can use - then we'll look at various options for packaging and installation. https://docs.npmjs.com/cli/install does have a few options there!

sdks/nodejs/src/agonesSDK.js
markmandel: I guess if that happens, then everything breaks - and well, the server becomes unhealthy - which is probably correct anyway?

markmandel:
Added some notes on build tooling - first step would be generation -- then probably adding the test suite to the `make test` command. If you have questions, pop into \#development.

We will also want to add some docs :smile: to the website as well :scroll: 

But this looks like a good first start!

sdks/nodejs/src/agonesSDK.js
```diff
@@ -0,0 +1,101 @@
+const grpc = require('grpc');
+
+const messages = require('../lib/sdk_pb');
+const services = require('../lib/sdk_grpc_pb');
+
+class AgonesSDK {
+	constructor() {
+		this.client = new services.SDKClient('localhost:50051', grpc.credentials.createInsecure());
```
Kuqd: shouldn't the port be 59357 ?

Kuqd:
This looks good I'll help you get that into our Makefile soon.

sdks/nodejs/src/agonesSDK.js
steven-supersolid: Doh! That explains why I had to keep changing it to test. The port in the example code I was using is 50051. Changed in next commit

sdks/nodejs/lib/sdk_grpc_pb.js
```diff
@@ -0,0 +1,160 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
steven-supersolid: Maybe don't need to call header() on this file in the build script as the boilerplate is already included?

Kuqd: it seems so. This is good, we just need to include make targets !

markmandel:
Just touching base - I'm guessing the things that are outstanding are (a) documentation and (b) integration with our build system?

sdks/nodejs/README.md
steven-supersolid: Have now removed instructions and manual steps, and added `gen-grpc-nodejs.sh` as suggested. This changed the generated code as the version of protoc changed but have successfully tested against the local SDK server.

sdks/nodejs/package.json
steven-supersolid: If a node module is not published to npm then another way to get this is via a git link, e.g `npm install --save GoogleCloudPlatform/agones#release-0.9.0` (or older equivalent `npm install --save git://github.com/GoogleCloudPlatform/agones.git#release-0.9.0` But `package.json` would have to be in the root of the project for that to work. We cannot include the path to the SDK from the repo root as there is no way to install from a git subdirectory: https://github.com/npm/npm/issues/2974

I think in order for someone to use this code they would have to copy at the nodejs sdk path to their own public repository and include that way (what we will do), or copy and paste the code inside their own project. Enough for now?

sdks/nodejs/src/agonesSDK.js
steven-supersolid: I did a test and removed the throw as this could never be caught by client code.

sdks/nodejs/spec/agonesSDK.spec.js
```diff
@@ -0,0 +1,246 @@
+const EventEmitter = require('events');
```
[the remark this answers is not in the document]
steven-supersolid: We need to run `npm test` while inside the `sdks/nodejs` folder. Should we copy the folder in to the image to achieve this?

steven-supersolid:
Getting closer now :) I've added some documentation and also had a go at updating `make test` but not sure how to proceed on this so not committed.

site/content/en/docs/Guides/Client SDKs/nodejs.md
```diff
@@ -0,0 +1,66 @@
+---
+title: "Node.js Game Server Client SDK"
+linkTitle: "nodejs"
```
steven-supersolid: Will change to Node.js

sdks/nodejs/package.json
markmandel: Yeah, I think that is enough for now :+1: people can copy paste for the moment.

sdks/nodejs/spec/agonesSDK.spec.js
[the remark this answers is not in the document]
markmandel: I think this is the last item. Since we already have node installed on the build image, this should hopefully be pretty straight forward.

We should have a `test-node` target in the Makefile.  Probably something like:

```
test-node: $(ensure-build-image)
	docker run --rm $(common_mounts) -w $(mount_path)/sdk/nodejs $(build_tag) npm list test
```
(not sure if I got the commands 100% right, but let me know if you get my gist?)

Then also add this target to `test`, and I think we are good to go.

markmandel:
This looks great to me! @Kuqd anything you wanted to touch on? (Since I know you were looking at this closely). Otherwise, I say this is good to merge.

markmandel:
@steven-supersolid - this will need to be squashed into a single commit please. But otherwise, I think we are good to merge this!

site/content/en/docs/Guides/Client SDKs/nodejs.md
```diff
@@ -0,0 +1,66 @@
+---
+title: "Node.js Game Server Client SDK"
+linkTitle: "Node.js"
+date: 2019-02-24T15:56:57Z
+weight: 50
```
markmandel: Gah! One more tiny thing.
Can you please add here:
`publishDate: 2019-04-1`

This will ensure that this page won't show up until the stable release day.

Sorry, I missed this on my initial review.

site/content/en/docs/Guides/Client SDKs/nodejs.md
```diff
@@ -0,0 +1,67 @@
...
+
+Similarly `shutdown()`, `setAnnotation(key, value)` and `setLabel(key, value)` are async methods that perform an action and return an empty result.
+
+To get [details of the backing GameServer]({{< relref "_index.md#gameserver" >}}) call the async method `getGameServer()`. The result will be an object representing `GameServer` defined in {{< ghlink href="sdk.proto" >}}`sdk.proto`{{< /ghlink >}}.
+
+```javascript
+let result = await agonesSDK.getGameServer();
+```
+
+To get [updates on the backing GameServer]({{< relref "_index.md#watchgameserverfunctiongameserver" >}}) as they happen, call `watchGameServer(callback)`. The callback will be called with a parameter matching the result of `getGameServer()`.
```
markmandel:
```suggestion
To get [updates on the backing GameServer]({{< relref "_index.md#watchgameserver-function-gameserver" >}}) as they happen, call `watchGameServer(callback)`. The callback will be called with a parameter matching the result of `getGameServer()`.
```

That will fix the test failure, since the anchor is now inline with master.

PR #583, opened 2019-02-11
Improve allocation performance

ilkercelikyilmaz:
Remove the call to sync the gameserver cache during each allocation request and replaced it with using local (ready gameservers) cache.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -70,6 +70,9 @@ const (
 	logSizeLimitMBFlag           = "log-size-limit-mb"
 	kubeconfigFlag               = "kubeconfig"
 	defaultResync                = 30 * time.Second
+	// topNGSForAllocation is used by the GameServerAllocation controller
+	// to reduce the contention while allocating gameservers.
+	topNGSForAllocation = 100
 )
 
 var (
@@ -187,8 +190,8 @@ func main() {
 	fleetController := fleets.NewController(wh, health, kubeClient, extClient, agonesClient, agonesInformerFactory)
 	faController := fleetallocation.NewController(wh, allocationMutex,
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
-	gasController := gameserverallocations.NewController(wh, health, allocationMutex, kubeClient,
-		kubeInformerFactory, extClient, agonesClient, agonesInformerFactory)
+	gasController := gameserverallocations.NewController(wh, health, kubeClient,
+		kubeInformerFactory, extClient, agonesClient, agonesInformerFactory, topNGSForAllocation)
 	fasController := fleetautoscalers.NewController(wh, health,
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
 
--- a/pkg/apis/stable/v1alpha1/gameserverallocation.go
+++ b/pkg/apis/stable/v1alpha1/gameserverallocation.go
@@ -26,6 +26,9 @@ const (
 	GameServerAllocationAllocated GameServerAllocationState = "Allocated"
 	// GameServerAllocationUnAllocated when the allocation is unsuccessful
 	GameServerAllocationUnAllocated GameServerAllocationState = "UnAllocated"
+	// GameServerAllocationContention when the allocation is unsuccessful
+	// because of contention
+	GameServerAllocationContention GameServerAllocationState = "Contention"
 )
 
 // GameServerAllocationState is the Allocation state
--- a/pkg/gameserverallocations/controller.go
+++ b/pkg/gameserverallocations/controller.go
@@ -16,7 +16,10 @@ package gameserverallocations
 
 import (
 	"encoding/json"
+	"math/rand"
+	"strconv"
 	"sync"
+	"time"
 
 	"agones.dev/agones/pkg/apis/stable"
 	"agones.dev/agones/pkg/apis/stable/v1alpha1"
@@ -39,7 +42,7 @@ import (
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -52,6 +55,8 @@ var (
 	// ErrNoGameServerReady is returned when there are no Ready GameServers
 	// available
 	ErrNoGameServerReady = errors.New("Could not find a Ready GameServer")
+	// ErrConflictInGameServerSelection is returned when the candidate gameserver already allocated
+	ErrConflictInGameServerSelection = errors.New("The Gameserver was already allocated")
 )
 
 // Controller is a the GameServerAllocation controller
@@ -65,9 +70,64 @@ type Controller struct {
 	gameServerAllocationSynced cache.InformerSynced
 	gameServerAllocationGetter getterv1alpha1.GameServerAllocationsGetter
 	stop                       <-chan struct{}
-	allocationMutex            *sync.Mutex
 	workerqueue                *workerqueue.WorkerQueue
+	gsWorkerqueue              *workerqueue.WorkerQueue
 	recorder                   record.EventRecorder
+	readyGameServers           gameServerCacheEntry
+	// Instead of selecting the top one, controller selects a random one
+	// from the topNGameServerCount of Ready gameservers
+	topNGameServerCount int
+}
+
+// gameserver cache to keep the Ready state gameserver.
+type gameServerCacheEntry struct {
+	mu    sync.RWMutex
+	cache map[string]*v1alpha1.GameServer
+}
+
+// Store saves the data in the cache.
+func (e *gameServerCacheEntry) Store(key string, gs *v1alpha1.GameServer) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	if e.cache == nil {
+		e.cache = map[string]*v1alpha1.GameServer{}
+	}
+	e.cache[key] = gs.DeepCopy()
+}
+
+// Delete deletes the data. If it exists returns true.
+func (e *gameServerCacheEntry) Delete(key string) bool {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	ret := false
+	if e.cache != nil {
+		if _, ok := e.cache[key]; ok {
+			delete(e.cache, key)
+			ret = true
+		}
+	}
+
+	return ret
+}
+
+// Load returns the data from cache. It return true if the value exists in the cache
+func (e *gameServerCacheEntry) Load(key string) (*v1alpha1.GameServer, bool) {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	val, ok := e.cache[key]
+
+	return val, ok
+}
+
+// Range extracts data from the cache based on provided function f.
+func (e *gameServerCacheEntry) Range(f func(key string, gs *v1alpha1.GameServer) bool) {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	for k, v := range e.cache {
+		if !f(k, v) {
+			break
+		}
+	}
 }
 
 // findComparator is a comparator function specifically for the
@@ -75,15 +135,23 @@ type Controller struct {
 // scheduling strategy
 type findComparator func(bestCount, currentCount NodeCount) bool
 
+var allocationRetry = wait.Backoff{
+	Steps:    5,
+	Duration: 10 * time.Millisecond,
+	Factor:   1.0,
+	Jitter:   0.1,
+}
+
 // NewController returns a controller for a GameServerAllocation
 func NewController(wh *webhooks.WebHook,
 	health healthcheck.Handler,
-	allocationMutex *sync.Mutex,
 	kubeClient kubernetes.Interface,
 	kubeInformerFactory informers.SharedInformerFactory,
 	extClient extclientset.Interface,
 	agonesClient versioned.Interface,
-	agonesInformerFactory externalversions.SharedInformerFactory) *Controller {
+	agonesInformerFactory externalversions.SharedInformerFactory,
+	topNGameServerCnt int,
+) *Controller {
 
 	agonesInformer := agonesInformerFactory.Stable().V1alpha1()
 	c := &Controller{
@@ -94,11 +162,13 @@ func NewController(wh *webhooks.WebHook,
 		gameServerLister:           agonesInformer.GameServers().Lister(),
 		gameServerAllocationSynced: agonesInformer.GameServerAllocations().Informer().HasSynced,
 		gameServerAllocationGetter: agonesClient.StableV1alpha1(),
-		allocationMutex:            allocationMutex,
+		topNGameServerCount:        topNGameServerCnt,
 	}
 	c.baseLogger = runtime.NewLoggerWithType(c)
 	c.workerqueue = workerqueue.NewWorkerQueue(c.syncDelete, c.baseLogger, logfields.GameServerAllocationKey, stable.GroupName+".GameServerAllocationController")
+	c.gsWorkerqueue = workerqueue.NewWorkerQueue(c.syncGameServers, c.baseLogger, logfields.GameServerKey, stable.GroupName+".GameServerUpdateController")
 	health.AddLivenessCheck("gameserverallocation-workerqueue", healthcheck.Check(c.workerqueue.Healthy))
+	health.AddLivenessCheck("gameserverallocation-gameserver-workerqueue", healthcheck.Check(c.gsWorkerqueue.Healthy))
 
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartLogging(c.baseLogger.Infof)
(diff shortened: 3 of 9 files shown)


pkg/gameserverallocations/controller.go
```diff
@@ -137,11 +168,22 @@ func (c *Controller) Run(workers int, stop <-chan struct{}) error {
...
+	err = c.syncReadyGSServerCache()
+	if err != nil {
+		return err
+	}
+
 	c.workerqueue.Run(workers, stop)
+
+	// we don't want mutiple workers refresh cache at the same time so one worker will be better.
+	// Also we don't expect to have too many failures when allocating
+	c.gsWorkerqueue.Run(1, stop)
```
jkowalski: how can it be? Run() is a blocking call.

ilkercelikyilmaz: If we have more than one worker, each of them consume messages from the queue (in parallel) which will effectively try to refresh the cache (the refresh method refreshes the whole cache)

pkg/gameserverallocations/controller.go
```diff
@@ -117,6 +122,32 @@ func NewController(wh *webhooks.WebHook,
 		},
 	})
 
+	agonesInformer.GameServers().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
markmandel: I'm a little confused by this - I'm not seeing how this is better than the Informer cache. We still have to wait for the event to come through, so it is still eventually consistent -- and unless I'm mistaken (which is possible), this is a mirror of the Informer cache, so we don't gain anything?

Would it be easier / faster as a solution to have a list of potential Allocatable gameservers, and attempt to each on as allocated, and stop and return when one is successful? (fail on n number of attempts).

WDYT?

jkowalski:
Removing the mutex has a potential to introduce very subtle bugs (leaks, double allocation, panics), can we add some allocation stress test before making this change? then we can be confident we're not regressing.

How about this e2e scenario:

- make a large-ish fleet (say 100 GSs), wait for ready
- start ~10 goroutines, each allocating game servers till we get to exhaustion
- wait for all to complete
- verify that:
  - all GS got allocated
  - each time we got a different GS
  - each GS was properly updated

I would perhaps add one more twist: while allocating, we should have game servers update themselves (via SDK.SetLabel()), which will test that we're handling update conflicts in the allocation routine.

markmandel:
Maybe that is something that can be added here:
https://github.com/GoogleCloudPlatform/agones/blob/master/test/e2e/fleet_test.go#L373 ?

pkg/gameserverallocations/controller.go
markmandel: @jkowalski WDYT on the above?

ilkercelikyilmaz: cache.WaitForCacheSync(c.stop, c.gameServerSynced)  call in allocate() method seem to slow down the allocation call. Replacing with local cache and keep updating seem to improve the performance. The mutex blocking is only on the local cache instead of WaitForCacheSync.
The performance difference is not that much is the number of Ready servers is not changing (cluster is in steady state) and there are not too many concurrent allocation calls.

We don't need to wait on the event to come (at least for the Allocated Gameservers) since we are deleting the allocated servers from the cache.

I was thinking of the solution you suggested (in the second paragraph). I even try to implement something and tested but there are few moving parts where GSA , GS, FLT can easily go out of sync.

I am inspired from the Jarek's GameSerSet allocation and use of local cache.

ilkercelikyilmaz:
> Removing the mutex has a potential to introduce very subtle bugs (leaks, double allocation, panics), can we add some allocation stress test before making this change? then we can be confident we're not regressing.
> 
> How about this e2e scenario:
> 
> * make a large-ish fleet (say 100 GSs), wait for ready
> * start ~10 goroutines, each allocating game servers till we get to exhaustion
> * wait for all to complete
> * verify that:
>   
>   * all GS got allocated
>   * each time we got a different GS
>   * each GS was properly updated
> 
> I would perhaps add one more twist: while allocating, we should have game servers update themselves (via SDK.SetLabel()), which will test that we're handling update conflicts in the allocation routine.

I agree with removing mutex will cause unexpected behaviors/bugs. That's why I am still keeping it but reducing it's scope and blocking it local cache.

I did few manual tests (with parallel allocations). I wanted to get quick feedback before spending more time on tests. I will add e2e test(s).

pkg/gameserverallocations/controller.go
markmandel: > cache.WaitForCacheSync(c.stop, c.gameServerSynced) call in allocate() method seem to slow down the allocation call.

Yep, makes perfect sense :+1: 

>  We don't need to wait on the event to come (at least for the Allocated Gameservers) since we are deleting the allocated servers from the cache.

But we aren't deleted gameservers from the cache when Fleets are scaled down - so we could  have race conditions as Fleets are scaled down.

> I was thinking of the solution you suggested (in the second paragraph). I even try to implement something and tested but there are few moving parts where GSA , GS, FLT can easily go out of sync.

Not sure what you mean here. Can you maybe explain it in a different way?

ilkercelikyilmaz: > But we aren't deleted gameservers from the cache when Fleets are scaled down - so we could have race conditions as Fleets are scaled down.

This is true. However when/if the system tries to allocate a GS from cache that is deleted, it will fail to change the GS's state to Allocated since the revisonversion of the GS in the cache will be old.  

> Not sure what you mean here. Can you maybe explain it in a different way?

Since the GS state update ( from Ready to Allocated) and GSA record creation is not transnational ,we can end up creating multiple GSA records for the same GS.

markmandel: > This is true. However when/if the system tries to allocate a GS from cache that is deleted, it will fail to change the GS's state to Allocated since the revisonversion of the GS in the cache will be old.

Ah! true! But now we would likely return a 500 from the service when Fleets are scaling down. I think we need some retry logic.

> Since the GS state update ( from Ready to Allocated) and GSA record creation is not transnational ,we can end up creating multiple GSA records for the same GS.

I don't follow. My vision of the process would be:
1. Get n list of potential allocatable GS's (where n is essentially the set of retries)
1. Loop through list of GS's
   1. attempt to move to Allocated
   1. if successful, break
1. Update GSA with allocated GS details

Does that make sense?

ilkercelikyilmaz: > I don't follow. My vision of the process would be:

1. Get n list of potential allocatable GS's (where n is essentially the set of retries)
2. Loop through list of GS's
   
   1. attempt to move to Allocated
   2. if not successful, continue to next GS
   3. if successful, break
3. Update GSA with allocated GS details

Does that make sense?

Makes sense. I would like to reproduce the condition and then try to implement what you are suggesting.

markmandel:
Looks like you're still actively working on this PR? Just wanted to touch base and check?

ilkercelikyilmaz:
I made quite a bit progress. I am planning to checkin once I merge my changes. I might have more improvements but I am planning to do them in another PR.

markmandel:
Okay cool - will hold off doing another review until you give the :+1: 

Excited to see the improvements!

ilkercelikyilmaz:
> Okay cool - will hold off doing another review until you give the 
> 
> Excited to see the improvements!

It will be great if you can review. I will like to check-in this version. I might have further improvements in the second round.

ilkercelikyilmaz:
@markmandel , should we create GSA crd for failed gs allocations (with UnAllocated state)? 
When I was doing load test, the system ended up creating lot of UnAllocated GSA records and most didn't seem to cleaned-up immediately. It probably slows down the overall processing.
I start returning error when the server can't be allocated but that is failing some tests. I wanted to check with you before I started to cleaning them up. Please let me know what you think.

markmandel:
> @markmandel , should we create GSA crd for failed gs allocations (with UnAllocated state)?
> When I was doing load test, the system ended up creating lot of UnAllocated GSA records and most didn't seem to cleaned-up immediately. It probably slows down the overall processing.
> I start returning error when the server can't be allocated but that is failing some tests. I wanted to check with you before I started to cleaning them up. Please let me know what you think.

#600 should solve this issue, as we totally remove storage.

ilkercelikyilmaz:
> > @markmandel , should we create GSA crd for failed gs allocations (with UnAllocated state)?
> > When I was doing load test, the system ended up creating lot of UnAllocated GSA records and most didn't seem to cleaned-up immediately. It probably slows down the overall processing.
> > I start returning error when the server can't be allocated but that is failing some tests. I wanted to check with you before I started to cleaning them up. Please let me know what you think.
> 
> #600 should solve this issue, as we totally remove storage.

Agreed. Just want to make sure. I have already removed the Unallocate GSA creation from the code.

pkg/gameserverallocations/controller.go
```diff
@@ -65,25 +71,100 @@ type Controller struct {
...
+}
+
+// gameserver cache to keep the Ready state gameserver.
+type gameServerCacheEntry struct {
+	mu    sync.RWMutex
+	cache map[string]*v1alpha1.GameServer
+}
+
+// LoadOrStore returns data if exists or stores if it is not. Return the value and true if the data already exists.
+func (e *gameServerCacheEntry) LoadOrStore(key string, gs *v1alpha1.GameServer) (*v1alpha1.GameServer, bool) {
```
markmandel: Is this here to make testing easier, I assume?

pkg/gameserverallocations/controller.go
```diff
@@ -386,18 +495,143 @@ func (c *Controller) findReadyGameServerForAllocation(gsa *v1alpha1.GameServerAl
 		}
 	}
 
+	bestGSList := []v1alpha1.GameServer{}
```
markmandel: My only concern here, is that we get more of a random pattern of gameservers - unless I'm reading this wrong (which I may be), we aren't grabbing an ordered list of GameServers based on comparators, we are grabbing N number of the gameserver that we come across (in fact, the comparator isn't really doing anything at this point).

So we're breaking the `Scheduling` contract here essentially. While I see we want to break it a little (pull a random one out of set), I feel like we should at least get the top N number of sorted gameservers by comparator.

The solution I think would be easiest (although probably not the fastest) - would be to sort the `allocationSet` based on the `nodeCount` values that the allocation comes from with the `comparator`, and then take the top N GameServers from that. WDYT? Or  is there some way we could pre-calculate this? (probably not with the combo of selectors).

pkg/gameserverallocations/controller.go
```diff
@@ -168,14 +280,21 @@ func (c *Controller) creationMutationHandler(review admv1beta1.AdmissionReview)
...
-	if err != nil && err != ErrNoGameServerReady {
+	var gs *v1alpha1.GameServer
+	err = Retry(allocationRetry, func() error {
+		gs, err = c.allocate(gsa)
+		return err
+	})
+
+	if err != nil {
+		if err != ErrNoGameServerReady {
+			// this will trigger syncing of the cache (assuming cache might not be up to date)
```
markmandel: Re: your earlier comment about removing UnAllocated. I would probably like the functionality to be returned, just so that we're not breaking things -- but I'd probably also be happy with a TODO:; in the comments stating that it will be returning in #600 (and I'll  make sure to update the code in #600 to fix it, assuming this PR goes through first).

How do you feel about that?

cmd/controller/main.go
```diff
@@ -65,6 +65,7 @@ const (
 	apiServerBurstQPSFlag        = "api-server-qps-burst"
 	kubeconfigFlag               = "kubeconfig"
 	defaultResync                = 30 * time.Second
+	topNGSForAllocation          = 100
```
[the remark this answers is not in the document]
ilkercelikyilmaz: I added a comment.
I thought about making it configurable but I prefer to keep it constant until we can provide a guidance how this parameter effects the performance.

pkg/gameserverallocations/controller.go
```diff
@@ -65,25 +71,100 @@ type Controller struct {
...
 func NewController(wh *webhooks.WebHook,
 	health healthcheck.Handler,
-	allocationMutex *sync.Mutex,
 	kubeClient kubernetes.Interface,
 	kubeInformerFactory informers.SharedInformerFactory,
 	extClient extclientset.Interface,
 	agonesClient versioned.Interface,
-	agonesInformerFactory externalversions.SharedInformerFactory) *Controller {
+	agonesInformerFactory externalversions.SharedInformerFactory,
+	topNGameServerCnt int,
```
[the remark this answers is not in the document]
ilkercelikyilmaz: Done

pkg/gameserverallocations/controller.go
```diff
@@ -65,25 +71,100 @@ type Controller struct {
...
+	}
 }
 
 // findComparator is a comparator function specifically for the
 // findReadyGameServerForAllocation method for determining
 // scheduling strategy
 type findComparator func(bestCount, currentCount NodeCount) bool
 
+var allocationRetry = wait.Backoff{
+	Steps:    5,
```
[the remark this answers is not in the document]
ilkercelikyilmaz: I am not sure making this configurable will be helpful without some guidance. Eventually it can be. 
Once everything is stable, we can try different values and find a good number.
I get this from [here](https://github.com/GoogleCloudPlatform/agones/blob/master/vendor/k8s.io/client-go/util/retry/util.go). I was trying different numbers. I agree that it should not be too high. I can set it to 3 for now. 
What do you think?

pkg/gameserverallocations/controller.go
ilkercelikyilmaz: I originally used [syncmap](https://godoc.org/golang.org/x/sync/syncmap#Map.Range). Later on I decided to use my own synced cache so instead of changing my code, I implemented the syncmap interface. This was I can go back using syncmap as well.

markmandel: Fair enough - although removing methods is not going to block you from using sync.Map :)

I just saw you used LoadOrStore in your tests, so figured it was easier for testing :+1:

pkg/gameserverallocations/controller.go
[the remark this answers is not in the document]
markmandel: That sounds fair - more than anything else, I just wanted to ask the question. I'm happy with 5 - seems like a legitimate number. I feel like I'd rather aim higher, and then maybe pull it down if need be,

@Kuqd curious - at some point, should we add a metric for this somewhere? How many retries?

cmd/controller/main.go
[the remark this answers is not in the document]
markmandel: Sounds fair. Let's open that up once we're at a happier place performance wise.

pkg/gameserverallocations/controller.go
markmandel: Ooh. That's an exceptionally interesting question. My biggest concern is that the calling system is aware that there was a contention, and tries again. Returning UnAllocated, might make it think that it has none left, rather than it just failed under high load.

Throwing an error would make sense, but it would be hard to know if it's an error because something horrible happened, rather than a (hopefully) rare contention issue.

Another idea might be to add another status, `Allocated`, `UnAllocated`, `Contention` ? (or something more meaningful? I can't find the right words). How does that sound?

pkg/gameserverallocations/controller.go
markmandel: I don't _think_ this gives us a sorted list, but it gives us a slightly better list than before (since list ordering out of the informer cache is random, and we aren't going through all gameservers before appending them).

The `Packed` scheduling is in place, because without it, you get the "swiss cheese" problem, wherein you can't shut down nodes because of sparse game servers -- which at scale is real money.

> I can check-in by removing else if & else (see above) so we don't break the contract and we can create another issue to improve this. WDYT?

I think that makes the most sense. I've also realise that the recent changes to gameserverset also break the contract -- so I'm wondering :thinking: if it makes sense to keep an ordered by node count list of gameservers in a cache somewhere, that can be used globally, and are updated at event time -- but agreed, that should be a separate PR.

So that sounds like a good way through :+1: to get these changes over the line.

ilkercelikyilmaz: Ok, I will remove the else if & else to not break the Scheduling contract.
Just FYI, the performance difference between breaking and not breaking the contract is 175 Allocations per sec vs 116 Allocations per sec (first load test is which breaks the contract).
image

pkg/gameserverallocations/controller.go
ilkercelikyilmaz: I like the idea of adding another status. I am OK with Contention :+1: . We can change to something else if we find a better name.

pkg/apis/stable/v1alpha1/gameserverallocation.go
```diff
@@ -26,6 +26,9 @@ const (
 	GameServerAllocationAllocated GameServerAllocationState = "Allocated"
 	// GameServerAllocationUnAllocated when the allocation is unsuccessful
 	GameServerAllocationUnAllocated GameServerAllocationState = "UnAllocated"
+	// GameServerAllocationContention when the allocation is unsuccessful
+	// because of contention
+	GameServerAllocationContention GameServerAllocationState = "Contention"
```
markmandel: Do we want to document this somewhere? Maybe add a short line to https://agones.dev/site/docs/getting-started/create-fleet/#gameserverallocation , after "The state value will tell if a GameServer was allocated or not. If a GameServer could not be found, this will be set to UnAllocated." ?

Maybe do a line after that about what Contention means, and mark it to publish at 0.9.0 ?

https://agones.dev/site/docs/contribute/

ilkercelikyilmaz: I am adding this line : 

> If there are too many concurrent requests ovewhelmed the system, some the `state`  will be set to 
> {{% feature publishVersion="0.9.0" %}}`Contention`{{% /feature %}} even though there are available `GameServer`.

Let me know what you think.

markmandel: That sounds perfect, I would put the `{{% feature publishVersion="0.9.0" %}}` block around the whole sentence. :+1:

markmandel:
LGTM. Needs to be squashed to a single commit, and then it's good to go!

Issue #573, opened 2019-02-07
Infrastructure for Agones performance reporting


jkowalski:
We need a solution for gathering and publishing/analyzing Agones performance over time to identify trends and performance regressions.

We need to solve 4 aspects of it:

1. How to write performance tests and instrument them for performance data collection?
2. How to store the results long-term?
3. How to present the performance results (to be able to deep dive into a test run, have comparison of two runs and see long-term trends)
4. How to run all of this periodically in a clean environment so we have comparable data over time?

Here's one proposal:

* Use https://fortio.org/ for visualization (used by Istio, stores performance metrics in very simple JSON which we could quite easily produce ourselves)

  Example visualizations: _[Single run](https://fortio-daily.istio.io/browse?url=qps_max-s2_to_s1-release-1.0.0-snapshot.0-20180623-09-15-2018-06-23-09-39.json), [Comparison of two runs](https://fortio-daily.istio.io/?xMin=&xMax=&xLog=false&yLog=false&sel=qps_max-s2_to_s1-release-1.0.0-snapshot.0-20180623-09-15-2018-06-23-09-39&sel=qps_max-s2_to_s1-release-1.0.0-snapshot.0-20180622-17-43-2018-06-22-18-30) and [trends over time](https://fortio-daily.istio.io/?s=qps_max-s2_to_s1-release-1.0.0)._
 Updated links: [Single run](https://fortio.istio.io/browse?url=qps_max-s2_to_s1-0.5.1-2018-04-05-22-40.json), [Comparison of two runs](https://fortio.istio.io/?xMin=&xMax=&xLog=false&yLog=false&sel=qps_max-s2_to_s1-0.5.1-2018-04-05-22-40&sel=qps_max-s1_to_s2-0.7.1-2018-04-05-22-04) and [Trends over time](https://fortio.istio.io/?s=qps_max-s2_to_s1)

* Instrument e2e tests (akin to #571) to emit Fortio-compatible metrics as JSON files, and store those in GCS bucket for long-term storage. We could have "official" flag for tests that run against clean cluster 

* Have a website (https://performance.agones.dev/ on appengine perhaps?) that hosts Fortio and periodically syncs data from GCS bucket for presentation purposes.

* Have some cron job, that starts a clean GKE cluster on a schedule (e.g. 4 times a day), launches appropriate tests and uploads results to GCS for long-term storage and pushes them to Fortio for presentation. Hoping we can have a Makefile-based solution that we could drive using Cloud Build or something like that.

[...]

PR #586, opened 2019-02-11
Emit stress test metrics in Fortio format.

jkowalski:
This addresses the instrumentation aspect of #573 which is emitting performance metrics from E2E tests and saving them to a folder in JSON format compatible with Fortio.

By default stress test results are saved in `build/.perf` directory.


markmandel:
Is this on hold until we're happy with #573 ?

jkowalski:
> Is this on hold until we're happy with #573 ?

This is a small component of the solution, I'm hoping not a controversial one, unless we don't want to do Fortio at all.

test/e2e/framework/perf.go
```diff
@@ -1,25 +1,35 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
+	"fortio.org/fortio/fhttp"
+	"fortio.org/fortio/stats"
 	"github.com/sirupsen/logrus"
+
```
markmandel: nit: blank line.

test/e2e/framework/framework.go
```diff
@@ -49,10 +49,11 @@ type Framework struct {
 	GameServerImage string
 	PullSecret      string
 	StressTestLevel int
+	PerfOutputDir   string
 }
 
 // New setups a testing framework using a kubeconfig path and the game server image to use for testing.
-func New(kubeconfig, gsimage string, pullSecret string, stressTestLevel int) (*Framework, error) {
+func New(kubeconfig string) (*Framework, error) {
```
markmandel: Curious reason for this - getting too wide?

aLekSer:
@jkowalski Possibly we need to add a small description of `make stress-test-e2e` in a build/README.md file:
https://github.com/GoogleCloudPlatform/agones/blob/4ee2602c77bad8293d5dc8a86486ddd5c74fd844/build/README.md#L41
It is hard to guess the suitable cluster size needed to run test with `STRESS_TEST_LEVEL=20`
Currently only stress test levels <= 2 can be executed on default gcloud-test-cluster.

Issue #455, opened 2018-12-22
Race condition in SDK.SetLabel and SDK.SetAnnotation


jkowalski:
I believe there's currently a race condition in SDK.SetLabel and SDK.SetAnnotation implementation which causes the labels to sometimes go back in time, because when SDK server performs read-modify-write on GameServer object, it does not read the fresh state of GameServer, but a cached one, thus it may overwrite labels/annotations back to old values.

Example: assuming game server sets the labels like that:

```
SDK.SetLabel(A,1)
SDK.SetLabel(B,1)
SDK.SetLabel(A,2)
SDK.SetLabel(B,2)
```

every observer (using GET/LIST) should always see value of A>=B because A increase happens before B increase.

Now imagine there's a propagation delay after `SetLabels(A,2)`. When `SetLabel(B,2)` executes, the cached GS still contains `{A=1,B=1}`, and SDK server will set labels: `{A=1,B=2}` which violates the invariant.

This is quite easy to reproduce in E2E test, even with very little load in the cluster.

jkowalski:
(BTW. I have a failing test and can work on a fix, can sb assign this to me?)

markmandel:
Does setting the labels not increase the generation - and thereby force a retry if the data has been changed?

jkowalski:
I don't think there's any mechanism to force a retry automatically that would work here. In this case the problem is because we're relying on stale data and trusting it as fresh.

BTW. One possible fix is to use Getter (always fresh, strongly consistent) instead of Lister (eventually consistent).

Perhaps an even better fix is to support coalescing multiple labels/annotations into a single read-modify-write against GS - this would be a good way of supporting retries and implementing #310 without excessive API server traffic.

markmandel:
Both SetLabel and SetAnnotation operations specifically go into a workerqueue, so they do retry of they fail. Or at least, they should!

jkowalski:
There is no failure anywhere here - all API server calls succeed. It's just a propagation delay.

Yingxin-Jiang:
I am working on the issue.

Yingxin-Jiang:
In my local test, I updated TestSDKSetLabel to set multiple labels on the game server, https://github.com/GoogleCloudPlatform/agones/blob/dcb7dfaf11266e5d39b6aaee297eb1a77603716b/test/e2e/gameserver_test.go#L56

I couldn't reproduce what jkowalski described. 

[...]

PR #588, opened 2019-02-12
sdkserver: fix race condition in SDK.SetLabel and SDK.SetAnnotation (issue #455)

Yingxin-Jiang:
Fix issue #455.

When synchronizing GameServer with requested operations, i.e. updateState, updateLabel, updateAnnotation, change the key format from {operation}/{data} to {operation}. The operation data is persisted in SDKServer to prevent old operation data from overwriting the new one.

### DIFF
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -78,6 +78,10 @@ type SDKServer struct {
 	connectedStreams   []sdk.SDK_WatchGameServerServer
 	stop               <-chan struct{}
 	recorder           record.EventRecorder
+	gsLabels           map[string]string
+	gsAnnotations      map[string]string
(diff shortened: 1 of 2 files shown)


pkg/sdkserver/sdkserver.go
```diff
@@ -107,6 +111,9 @@ func NewSDKServer(gameServerName, namespace string, kubeClient kubernetes.Interf
 		healthMutex:        sync.RWMutex{},
 		healthFailureCount: 0,
 		streamMutex:        sync.RWMutex{},
+		gsLabels:           map[string]string{},
+		gsAnnotations:      map[string]string{},
+		gsUpdateMutex:      sync.RWMutex{},
```
jkowalski: nit: no need to initialize mutex like that (default value is fine). feel free to remove all of those.

Issue #541, opened 2019-01-31
Labels referencing resources name can be too long


Kuqd:
In Kubernetes you can create a resource with a 253 characters long name :

> By convention, the names of Kubernetes resources should be up to maximum length of 253 characters and consist of lower case alphanumeric characters, -, and ., but certain resources have more specific restrictions.

However labels and annotations values are restricted to 63 characters :

> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

In Agones we're setting labels value for referencing parent resources, but if a parent resource has a long name this cause errors.

References:
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

/cc @victor-prodan

aLekSer:
This give an error here:
```
 "agones.dev/agones/pkg/fleets.(*Controller).upsertGameServerSet
	/go/src/agones.dev/agones/pkg/fleets/controller.go:258"
```
I am thinking of CreationValidationHandler on Create and Update the Fleet to check the Fleet Name length (and insist on no more than 63).

markmandel:
Can this now be closed, or do we need to do GameServer as well?

aLekSer:
Let me check other places were we use name in labels.

aLekSer:
@Kuqd I think this ticket could be closed now. All names which reside in Label covered with validation functions.

PR #598, opened 2019-02-18
Add Fleet and Gameserverset Validation handler

aLekSer:
Add Fleet and GSS Name length check on Create and Update in order to prevent GSS label overflowing.
For #541 .


pkg/apis/stable/v1alpha1/fleet.go
```diff
@@ -134,6 +137,24 @@ func (f *Fleet) ApplyDefaults() {
 	f.ObjectMeta.Annotations[stable.VersionAnnotation] = pkg.Version
 }
 
+// Validate validates the Fleet configuration.
+// If a Fleet is invalid there will be > 0 values in
+// the returned array
+func (f *Fleet) Validate() (bool, []metav1.StatusCause) {
```
markmandel: Something I'm fixing elsewhere, for the other validate functions (PR incoming) - I think we should file the return value, so the return should be: `([]metav1.StatusCause, bool)` - just because that's what we do in go for everything else.

Check a map:  `v, ok :=  map[key]` for example. So we should order it the same.

I've got the rest in an upcoming PR that also has new Validate functions as well.

How do you feel about that?

aLekSer: yes, that would be much more natural.

Issue #536, opened 2019-01-30
Use batching in GameServerAllocation controller to improve throughput.


jkowalski:
To get better throughput in GSA controller we could do batching: group together N allocation requests, assign GS to each of them and individually commit in parallel.

markmandel:
As a first step, I'm going to look into moving `GameServerAllocation` into a Aggregate API so we can have ore control over what happens with the API.

In theory, there should be no/minimal change to how GSA's work -- at least that's the plan :smile:

markmandel:
The good news is - I have (some cleanup to do) an API extension working to do gameserver allocations. 

I've only implemented and supported the CREATE (HTTP: Post) method on the API, as without storage, it's really the only one needed. If people request it, I could look into the Watch function as well, if people want to watch for create events.

The annoying news is - each create API call has 60s to provide a response (although can keep processing in the background) -- which makes one of the long term goals, having a SDK.Ack() function for blocking on Allocation return) -- a little trickier. Or at least, with a shorter timeout than I may have liked.

Asking the community for feedback on that aspect ([slack](https://agones.slack.com/archives/C9DGM5DS8/p1550311673053600)):
> I'm wondering if a 30 second timeout (to give everything else some buffer) for that ack() to come back is reasonable or not? I can't imaging you would want to keep your players waiting that long anyway -- or is that too short a time? (could probably bump that out to 40 or 50 second if need be, but 30 is super conservative)

> (Basically, we'd wait for 30 second, and if we didn't get the ack back, we'd either delete the gameserver, or maybe kick it back to ready - haven't decided. Probably delete -- it's cleaner)

Regardless, this will now also allow us to batch, skip storage for the GSA, etc. And also make it easier if we decide to also provide a gRPC interface as well for allocation.

ilkercelikyilmaz:
What is the number of per second allocations do we need to reach? I am running some load tests against the recent allocation changes I did (haven't been checked-in yet). I run 60 concurrent clients and the system can allocate around 60 gs per sec (Allocated 2999 gs in 50 secs). 

![image](https://user-images.githubusercontent.com/17522001/52987154-e624ac80-33ae-11e9-902e-fc2725363512.png)

markmandel:
@ilkercelikyilmaz that's a huge improvement over what we had previously :fire: @pm7h do you have those numbers on hand? I can't seem to find them). Once we also incorporate #600 - I wonder if we might be very close to what we might need to be.

pm7h:
[...]

Issue #597, opened 2019-02-15
Proposal: Multi-Cluster Allocation Policies


jkowalski:
Background
----

When operating multiple Agones clusters to support world-wide game launch it is often necessary to perform multi-cluster allocations (allocations from a set of clusters instead of just one) based on defined policies.

Examples of common policies include:

* **burst to cloud** - first allocate from a fleet in a cluster located on premises but when it's out of capacity, start allocating from a fleet hosted in the cloud.
* **provider preference** - prefer certain cloud providers over others based on factors such as cost or quality of service, with the ability to fall back to others when capacity is not available
* **load spreading** - distribute allocations across clusters from different providers to limit the blast radius of an outage of a single cluster (e.g. network fiber cut)

Proposal
----

This document proposes changing how `GameServerAllocation` API works by adding support for forwarding allocation requests to other clusters based on policies that can be applied on a per-request basis.

We will add a new CRD called `GameServerAllocationPolicy` that controls how multi-cluster allocations will be performed. The policy will contain a list of clusters to allocate from with corresponding priorities and weights. The credentials for accessing those clusters will be stored in secrets.

```yaml
apiVersion: "stable.agones.dev/v1alpha1"
kind: GameServerAllocationPolicy
metadata:
  name: my-policy
  namespace: default
spec:
  clusters:
  - name: on-prem
    priority: 1
    server: "https://some-endpoint"
    credentials: "some-secret-name"
  - name: another-cloud
    priority: 2
    weight: 100
    server: "https://some-other-endpoint"
    credentials: "some-other-secret-name"
  - name: this-cluster
    priority: 2
    weight: 300
    server: "local"
```

The name of the policy can be specified when creating `GameServerAllocation`.

```yaml
apiVersion: "stable.agones.dev/v1alpha1"
kind: GameServerAllocation
metadata:
  generateName: my-allocation-
spec:
  policy: my-policy
  required:
    matchLabels:
      game: my-game
  preferred:
    - matchLabels:
        stable.agones.dev/fleet: green-fleet
  scheduling: Packed
  metadata:
    ...
```

If `policy` is not specified, the allocation will be attempted from the local cluster as it is done today.
 
[...]

PR #600, opened 2019-02-19
[Breaking Change] Move GameServerAllocation to an API Extension Server.

markmandel:
This moves the implementation of GameServerAllocation (GSA) to a [Kubernetes API Extension](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) instead of using CRDs. This was essentially done for performance reasons, but to break it down:

1. GSA is now create only. Since we had no need for GSA storage, and don't want the performance hit.
1. This removes the mutation and validation webhooks, which have 30s timeout and are run in serial
    1. API Server still does cut off a response after 60s, but the api can continue processing (60s gives us enough time, I think for a SDK.Ack() on Allocate, which I don't think we had before)
    1. Validation now happens in the request.
1. We can now do batching of requests for higher throughput (#536), since we control the entire http request.
1. Sets us up if we decide we also want to have an alternative (http and/or gRPC) endpoint for allocation, basedn feedback from this implementation.

The breaking changes are:
1. GameServerAllocation's group is now `allocation.agones.dev` rather than `stable.agones.dev`,
because a CRD group can't overlap with a api server.
1. Since there is only the `create` verb for GSA, there is no get/list/watch options for GameServerAllocations - so no informers/listers either. But this could be added at a later date, if needed.

This also includes some libraries for building further api server extension points.


install/helm/agones/templates/admissionregistration.yaml
```diff
@@ -18,10 +18,40 @@
 {{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca }}
 {{- if .Values.agones.registerWebhooks }}
 ---
+apiVersion: apiregistration.k8s.io/v1beta1
```
markmandel: I put the APIService here, because it uses the same cert - but I don't know if "registerWebhooks" is still accurate, or if the file name "admissionregistration" is still accurate.

Thoughts on what should be done here?

Iinitial idea: rename files to "extensions.yaml" or something similiar. Maybe add an extra flag for the APIService that is seperate from the webhooks? agones.registerAPIService` ?

Kuqd: I'm wondering why `registerWebhooks` is a parameter, if you don't Agones won't work ?

I think if you just rename the file to `extensions.yaml` that would do it.

markmandel: basically - because people sometimes have weird requirements for installations, and need an escape hatch :man_shrugging: 

The escape hatch it to disable a certain section, and then manually install the thing that's been turned off in the special way that their platform requires.

That's really about it -- have run into it a few times with some users I've worked with 1:1 where they have needed this for things i wouldn't otherwise have thought of - so better to have and not want, than not have and be blocked until next release.

jkowalski:
This is really the only API of the entire Agones that's going to be invoked at scale on a regular basis (up to 100s of QPS for popular games) and I think it makes sense to decouple it from Kubernetes API serving infra as much as possible.

Given that this API is primarily going to be consumed by match makers and since we're already rewriting it, can we make it explicitly gRPC based instead of Kubernetes-based?

This would allow us to:

* host this API in dedicated pods for better throughput
* isolate allocation API traffic from all other (administrative) traffic
* put additional protections in place to limit the damage in case of overload (QPS quota, etc.)
* implement batching, pod affinity, prefetching, etc. for better allocation performance (#536 )
* put a load balancer and/or horizontal pod autoscaler in front of the service
* easily implement custom multi-cluster allocation topologies mentioned in #597 both in- and out-of cluster

I was under the impression that using K8s API serving infrastructure is not the most recommended approach, instead they suggest vanilla CRDs, which are not appropriate for us.

install/helm/agones/templates/admissionregistration.yaml
markmandel: Fixed! moved to extensions.yaml, and made a new flag for the apiservice.

markmandel:
@jkowalski these are really good points - I'm definitely not against having _both_ a k8s api (for nice in-development experience / people that want to reuse rbac, etc), and a grpc endpoint (for far higher throughput, wider scaling, etc).

It's my opinion that reusing the k8s api is the simpler option from an implementation perspective, and requires less work on our end (and less reinvention of the wheel, with security/rbac/etc) - so figure, start here, and see what kind of throughput we can get on it, and move forward from there.

We still get some of the things from above, but to break it down a little.

> * implement batching, pod affinity, prefetching, etc. for better allocation performance
We can definitely still do this - and share the logic between the gRPC implementation, and the k8s api one.

> * easily implement custom multi-cluster allocation topologies mentioned in #597 both in- and out-of cluster 

We could still do this as well, not seeing the blocker here either.

But like I said - start here, see how it goes, and iterate from there. I don't think we _lose_ anything by having the k8s api, as we are not locked out from doing a gRPC based API as well.

Thoughts?

markmandel:
Gentle bump on this :smiley:

markmandel:
I'm going to close this PR for now - and break it up into smaller chunks so it's easier to review. Stay tuned for newer, smaller PRs.

PR #601, opened 2019-02-19
Create a boolean to gate the creation of priority classes for controllers.

bbf:


markmandel:
Whoops, this went to the wrong branch - please resubmit a PR pointing to master.

(I wonder if we can make that a requirement some how)

PR #605, opened 2019-02-20
Release 0.8.0

markmandel:
First time running through a release with the new site and with the new site checker -- had to add a few spots in which we need to tell the link checker to ignore as those links won't exist until we merge the release.


markmandel:
If anyone approves this, I would suggest not merging -- I'll merge when I am ready to do the next steps, otherwise we'll have our website out of sync with our release for a few hours while, rather than 10 minutes.

PR #616, opened 2019-02-21
Unified logging of resource identifiers so that we can reliably get entire history of a resource in stack driver.

jkowalski:
Well-known identifiers of resources types are defined in new `logfields` package.


pkg/gameserverallocations/controller.go
```diff
@@ -145,16 +146,24 @@ func (c *Controller) Run(workers int, stop <-chan struct{}) error {
...
 func (c *Controller) creationMutationHandler(review admv1beta1.AdmissionReview) (admv1beta1.AdmissionReview, error) {
-	c.logger.WithField("review", review).Info("creationMutationHandler")
+	c.baseLogger.WithField("review", review).Info("creationMutationHandler")
 	obj := review.Request.Object
 	gsa := &v1alpha1.GameServerAllocation{}
 
 	err := json.Unmarshal(obj.Raw, gsa)
 	if err != nil {
-		c.logger.WithError(err).Error("error unmarchaslling json")
+		c.baseLogger.WithError(err).Error("error unmarchalling json")
```
markmandel:
```suggestion
		c.baseLogger.WithError(err).Error("error unmarshalling json")
```

pkg/gameservers/health.go
```diff
@@ -17,6 +17,8 @@ package gameservers
 import (
 	"strings"
 
+	"agones.dev/agones/pkg/util/logfields"
+
```
markmandel: nit: delete line

pkg/gameserversets/controller.go
```diff
@@ -252,34 +256,46 @@ func (c *Controller) gameServerEventHandler(obj interface{}) {
...
+			runtime.HandleError(c.baseLogger.WithField("gsKey", gs.ObjectMeta.Namespace+"/"+gs.ObjectMeta.Name).WithField("ref", ref),
 				errors.Wrap(err, "error retrieving GameServer owner"))
 		}
 		return
 	}
 	c.workerqueue.EnqueueImmediately(gsSet)
 }
 
+func loggerForGameServerSetKey(base *logrus.Entry, key string) *logrus.Entry {
+	return base.WithField(gameServerSetKeyFieldName, key)
```
markmandel: Should this not use `logfields.GameServerSetKey` ?

pkg/sdkserver/sdkserver.go
```diff
@@ -21,6 +21,8 @@ import (
 	"sync"
 	"time"
 
+	"agones.dev/agones/pkg/util/logfields"
+
```
markmandel: nit: empty line

pkg/util/workerqueue/workerqueue.go
```diff
@@ -77,7 +80,7 @@ func (wq *WorkerQueue) Enqueue(obj interface{}) {
 		runtime.HandleError(wq.logger.WithField("obj", obj), err)
 		return
 	}
-	wq.logger.WithField("key", key).Info("Enqueuing key")
+	wq.logger.WithField(wq.keyName, key).Info("Enqueuing")
```
markmandel: Should the workqueue have the keyname all the way through all it's logs?
(in which case, rather than store it, attach it as an extra `WithField` on `wq.logger` ?

pkg/fleetautoscalers/controller.go
```diff
@@ -48,7 +50,7 @@ import (
 
 // Controller is a the FleetAutoscaler controller
 type Controller struct {
-	logger                *logrus.Entry
+	baseLogger            *logrus.Entry
```
markmandel: Curious the motivation for the rename to `baseLogger`here - rather than leave it at `logger` ?

I kept looking for an inheritance / composition pattern when i saw "base" :smile: too much OO languages, I expect.

pkg/gameservers/controller.go
```diff
@@ -319,20 +328,20 @@ func (c *Controller) Run(workers int, stop <-chan struct{}) error {
 // syncGameServer synchronises the Pods for the GameServers.
 // and reacts to status changes that can occur through the client SDK
 func (c *Controller) syncGameServer(key string) error {
-	c.logger.WithField("key", key).Info("Synchronising")
+	c.loggerForGameServerKey(key).Info("Synchronising")
```
markmandel: Just thinking out loud, but any thoughts on whether to move these `loggerFor*` functions into logfields somehow? (centralised logger object maybe?).

May be overkill, because there doesn't seem to be much/any duplication here, but thought I'd ask the question.

pkg/fleetautoscalers/controller.go
jkowalski: `c.logger` is too easy to find and misuse :)

pkg/util/workerqueue/workerqueue.go
jkowalski: except the key is available to this method only, right?

pkg/gameservers/controller.go
jkowalski: I thought about it but could not find a nice pattern that works, you need to pass the base logger, the type and the value somehow. Two of those things are typically bound to a controller so it makes sense to have a method there.

pkg/sdkserver/sdkserver.go
jkowalski: done

pkg/gameserversets/controller.go
jkowalski: fixed

pkg/gameservers/health.go
jkowalski: fixed

pkg/gameserverallocations/controller.go
jkowalski: fixed

pkg/util/workerqueue/workerqueue.go
markmandel: OIC! This is the wq.keyName - got it!

PR #620, opened 2019-02-22
Added support for persisting logs in 'emptyDir' volume attached to agones controller.

jkowalski:
(This creates a secondary place to look at logs for debugging when controller is logging too fast and `fluentd` can't sync to stackdriver fast enough)

The logs are saved in 100MB chunks under `/home/agones/logs` up to a specified limit (10000 MB by default).

Example:

```shell
$  kubectl exec -n agones-system -it agones-controller-7d85b5c996-8c22b -- ls -lh /home/agones/logs
total 757764
-rw-r--r--    1 agones   agones    100.0M Feb 22 21:22 agones-controller-20190222_212001-2019-02-22T21-22-37.499.log
-rw-r--r--    1 agones   agones    100.0M Feb 22 21:26 agones-controller-20190222_212001-2019-02-22T21-26-31.904.log
-rw-r--r--    1 agones   agones    100.0M Feb 22 21:27 agones-controller-20190222_212001-2019-02-22T21-27-01.546.log
-rw-r--r--    1 agones   agones    100.0M Feb 22 21:27 agones-controller-20190222_212001-2019-02-22T21-27-42.934.log
-rw-r--r--    1 agones   agones    100.0M Feb 22 21:28 agones-controller-20190222_212001-2019-02-22T21-28-46.574.log
-rw-r--r--    1 agones   agones    100.0M Feb 22 21:32 agones-controller-20190222_212001-2019-02-22T21-32-31.224.log
-rw-r--r--    1 agones   agones    100.0M Feb 22 21:33 agones-controller-20190222_212001-2019-02-22T21-33-30.350.log
-rw-r--r--    1 agones   agones     40.0M Feb 22 21:41 agones-controller-20190222_212001.log
```


install/helm/agones/templates/controller.yaml
```diff
@@ -114,10 +120,19 @@ spec:
...
+{{- if .Values.agones.controller.persistentLogs }}
+        - name: logs
+          mountPath: /home/agones/logs
+          readOnly: false
+{{- end }}
       volumes:
       - name: certs
         secret:
           secretName: {{ template "agones.fullname" . }}-cert
+{{- if .Values.agones.controller.persistentLogs }}
```
markmandel: Curious - why use an emptyDir? Couldn't we write to the local disk regardless?

We don't need to share the volume between containers do we?

install/helm/agones/values.yaml
```diff
@@ -49,6 +49,8 @@ agones:
                 operator: Exists
     generateTLS: true
     safeToEvict: false
+    persistentLogs: true
```
markmandel: Thought: Should this be on by default? I'm thinking most people won't be doing load tests, and stackdriver/fluentd are likely fast enough. WDYT?

jkowalski: it's not just load testing, we currently emit enough logs to overwhelm fluentd. For any fleets of significant size there are dropped messages.

install/helm/agones/templates/controller.yaml
jkowalski: emptyDir survives container restarts, not a huge deal. I originally had a PVC here which would let you keep long-term logs on a persistent disk if you choose to. Would that be useful?

markmandel: Ah, gotcha. Yes - keeping that around for container restarts is :+1: that makes a lot of sense.

install/helm/agones/values.yaml
markmandel: Ah. In that case, yes - let's keep this on by default. I didn't realise we were getting dropped messages, that's unfortunate.

Sounds like something we should push upstream? (or should we be logging less?)

But otherwise, this makes sense, and looks good to me.

Only other thing I can think of - should we add a section in (to be released in 0.9.0) https://agones.dev/site/docs/guides/troubleshooting/ to let people know that this is an option?

jkowalski: sure, I'll submit this one now and make next PR with doc change and ability to use PVC for durable logs

PR #621, opened 2019-02-22
Fixed gameserverset overshooting the number of GameServers

### LINKED ISSUE #569: GameServerSet sometimes creates more GameServers than necessary
In particular under stress.

```
NAME                                                        SCHEDULING   DESIRED   CURRENT   ALLOCATED   READY   AGE
gameserverset.stable.agones.dev/scale-fleet-0-zqvvw-vwx2k   Packed       500       510       0           479     4m
```

jkowalski:
This happens typically during large turnups where some of the servers become unhealthy and get deleted. Now we treat `Shutdown` the same as deleted.

Also fixed how we count # of servers in GSS to exclude `Shutdown` and deleted ones.

Fixes #569


pkg/apis/stable/v1alpha1/gameserver.go
```diff
@@ -321,6 +321,16 @@ func (gs *GameServer) GetDevAddress() (string, bool) {
 	return devAddress, hasDevAddress
 }
 
+// IsAllocated returns true if the server is currently allocated.
+func (gs *GameServer) IsAllocated() bool {
```
markmandel: Thank you! I've been wanting to write this for weeks.

PR #630, opened 2019-03-01
Removes the sdk generation tooling from our main build image

Kuqd:
This removes all tooling related to building and generating code for each supported sdk.
We now support 3 commands per SDK (build,test,gen).
I updated the documentation on how to add a new SDK.


build/Makefile
```diff
@@ -345,10 +339,13 @@ push-ping-image: $(ensure-build-image)
 build-ping-image: $(ensure-build-image) build-ping-binary
 	docker build $(agones_path)/cmd/ping/ --tag=$(ping_tag) $(DOCKER_BUILD_ARGS)
 
+<<<<<<< HEAD
```
Kuqd: oupsy

Kuqd:
@markmandel this is ready to review.

build/includes/sdk.mk
```diff
@@ -0,0 +1,93 @@
...
+
+
+#         ____  ____   ____   _____           _ _	
+#    __ _|  _ \|  _ \ / ___| |_   _|__   ___ | (_)_ __   __ _	
+#   / _` | |_) | |_) | |       | |/ _ \ / _ \| | | '_ \ / _` |	
+#  | (_| |  _ <|  __/| |___    | | (_) | (_) | | | | | | (_| |	
+#   \__, |_| \_\_|    \____|   |_|\___/ \___/|_|_|_| |_|\__, |	
+#   |___/                                               |___/ 
+
+build_sdk_base_tag = agones-build-sdk-base:$(build_version)
```
markmandel: If the Dockerfile from the sdk tooling gets changed - does that force a rebuild? (should it?)

build/build-sdk-images/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Makefile targets run docker containers built from Dockerfile in each folder found in this directory. This means you don't need to change our Makefile
+
+Simply create a new directory with the name of your SDK. Then copy our [template](./tool/template) folder content you SDK folder.
+
+Edit the Dockerfile to install all dependencies you need to generate, build and test your SDK. (you should not need to change the base image)
+
+> As explained in our [SDK documentation](https://agones.dev/site/docs/guides/client-sdks/) you have 2 options HTTP or GRPC. If you are using HTTP you don't need to use our base SDK image so feel free to use the distribution of your choice.
+
+Then add your steps in `build.sh`, `test.sh` and `gen.sh` script files.
```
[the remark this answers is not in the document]
Kuqd: Those scripts can easily have 10 to 20 lines, so you'll still end up with scripts even with a Makefile.

What was your idea ?

build/includes/sdk.mk
Kuqd: it should, and it does not... I'm looking into it.

Kuqd:
I will look into the new CPP.

build/includes/sdk.mk
Kuqd: I went for the easy path, rebuilding images all the times.

build/build-sdk-images/README.md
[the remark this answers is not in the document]
markmandel: Was thinking it could be Makefiles rather than .sh files - but we already have .sh files for all the things right now.

I think you are right -- let's leave it as is, and if we can see places where things like common includes / etc that Makefiles would be useful we can revisit :+1:

build/includes/sdk.mk
markmandel: Sounds like a good first start, we can iterate as we need :+1:

markmandel:
Just letting you know, I have space again to review this. But there are a bunch of conflicts now :(

But I'll pull this down local and take it for a spin shortly!

PR #632, opened 2019-03-01
Switch to using default gke-cluster oauthScopes settings for clusters

### LINKED ISSUE #614: GKE scopes in installation and testing are overkill
[GKE scopes](https://cloud.google.com/sdk/gcloud/reference/container/clusters/create) could be dropped to the new(?) `gke-default` scope, which is far less than what we have in either the [documentation](https://github.com/GoogleCloudPlatform/agones/blob/master/site/content/en/docs/Installation/_index.md#creating-the-cluster), our [e2e cluster](https://github.com/GoogleCloudPlatform/agones/blob/master/build/gke-test-cluster/cluster-e2e.yml) or [development test clusters](https://github.com/GoogleCloudPlatform/agones/blob/master/build/gke-test-cluster/cluster.yml.jinja).

Likely only needs someone to test creating a cluster with the default scope, and making the changes in those three places.

aLekSer:
Reduce oauthScope settings with test-e2e.
Verified that metrics and logs works as before this change.
[Node Config details](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/NodeConfig).

Closes #614.


build/gke-test-cluster/cluster-e2e.yml
```diff
@@ -30,8 +30,6 @@ resources:
             oauthScopes:
               - https://www.googleapis.com/auth/compute
```
markmandel: Looking at https://cloud.google.com/sdk/gcloud/reference/container/clusters/create, `gke-default` is:

```
gke-default	
https://www.googleapis.com/auth/devstorage.read_only
https://www.googleapis.com/auth/logging.write
https://www.googleapis.com/auth/monitoring
https://www.googleapis.com/auth/service.management.readonly
https://www.googleapis.com/auth/servicecontrol
https://www.googleapis.com/auth/trace.append
```

Should we use the same?

aLekSer: Hello @markmandel ,

Yes, if we want to have consistent scopes in docs and yml.
By the way these two are enabled by default, if monitoring and logging API is enabled:
```
https://www.googleapis.com/auth/logging.write
https://www.googleapis.com/auth/monitoring
```
Last 3 scopes were not configured but they could be helpful.
The purpose of these oauthScopes could be found [here](https://developers.google.com/identity/protocols/googlescopes).
Let me check cluster creating with those 6 and omitting ` https://www.googleapis.com/auth/compute`.

PR #637, opened 2019-03-05
Adding Kubernetes API server requests metrics

### LINKED ISSUE #546: Grafana: add basic API Server graphs
QPS per resource type/verb:

`sum(rate(apiserver_request_count[5m])) by (resource,verb)`

Latency percentiles for key resource types:

`apiserver_request_latencies_summary{resource=~"gameservers|fleets|pods",verb!~"WATCH|LIST",quantile="0.5"}`

Error rates:

`sum(rate(apiserver_request_count{code!~"2.."}[5m])) by (code)`

Perhaps with a nice drop-down to allow narrowing it down by "resource" for common resource types (fleets, gameservers, pods, etc.)

aLekSer:
New dashboard with dropdown Agones CRD selector.
Contains 4 graphs of `apiserver_request_count` by verb and error code as proposed in the ticket as well as 3 quantiles (0.5, 0.9 and 0.99) for `apiserver_request_latencies_summary`.

Closes #546  .

### DIFF
--- a/build/grafana/dashboard-apiserver-requests.yaml
+++ b/build/grafana/dashboard-apiserver-requests.yaml
@@ -1,0 +1,888 @@
+# Copyright 2019 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# configs map used by grafana
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: agones-apiserver-requests
+  namespace: metrics
+  labels:
+     grafana_dashboard: "1"
+data:
+  dashboard-agones-apiserver-requests.json: |
+    {
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
+            "datasource": "-- Grafana --",
+            "enable": true,
+            "hide": true,
+            "iconColor": "rgba(0, 211, 255, 1)",
+            "name": "Annotations & Alerts",
+            "type": "dashboard"
+          }
+        ]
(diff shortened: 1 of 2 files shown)


jkowalski:
I tried the change, looks promising but we should make it easier to read:

* add legend "as table" and "to the right" and expose "current"
* set label units to requests/sec for throughput and milliseconds or microseconds for latency
* have separate graphs for 50th, 90th, 95th and 99th percentile latencies
* perhaps exclude dimensions with zero value (just add " != 0" to metric)
* perhaps use something like "{{ verb }} {{ resource }} {{ subresource }}"
* perhaps have separate graphs for each subresource (main resource, scale, status, could be a dropdown)

aLekSer:
Hello @jkowalski , 
Thanks for review, will apply your comments.
All this graphs should be on one dashboard and fit in one screen. am I right?
For some reason 0.9 percentile does not show me the plot, I need to understand why.

jkowalski:
Took a look again.

* Can you swap "Main Resource Request Count Per Second" and "Subresource Request Count Per Second"
* The metric in Main Resource Request Count Per Second needs to be `sum(rate(apiserver_request_count{resource=~"[[CustomResourceDefinition]]",verb!~"WATCH|LIST",subresource=""}[5m])) by (resource,verb)` with legend format `{{verb}} {{resource}}`
* Can you change the legend format for `Request Error Rate` to `{{resource}} {{subresource}} {{ code }}`
* TIP: When setting up axes units you can use `Units` dropdown instead of `Label` - this makes the axis able to use smarter labels
* Can you change the title `Request Latency - 0.5 quantile, milliseconds` to simply `Request Latency` and drop the other latency graph? Alternatively we can have 4 hard-coded graphs for all quantiles and remove the quantile dropdown completely. It's actually quite useful to see all quantiles at once even if it requires scrolling (you can make them as wide as the page so they align nicely)

aLekSer:
@jkowalski I have applied all your comments. I only can not make 0.95 metrics show up. When switching to value of 0.9 in query graph become depicted.

aLekSer:
Current look of the graphs (time period by default is 1hour):
screenshot 2019-03-07 at 00 01 46
screenshot 2019-03-07 at 00 02 00

aLekSer:
Checked that we have only 3 quantiles available (0.5, 0.9 and 0.99). So I updated the resulting dashboard file it now contains 7 plots.

jkowalski:
LGTM!

PR #638, opened 2019-03-06
[Regression] Fleet scale down didn't adhere to Packed Scheduling

markmandel:
The new Fleet scale up/scale down performance enhancements removed the functionality that the `Packed` strategy is documented to provide -- when scaling down, removing GameServers from least used Nodes first.

This change implements a library for GameServer sorting that doesn't use reflection, unlike sort.Slice (much faster), and can be used with any Comparator, as I expect we'll likely need this again.

This also pulls out the GameServerCounter controller - which keeps a running total of GameServers per Node, to make this kind of sorting much faster and easier as we don't have to compute this on every iteration.

To maintain performance as well, the set of GameServers only get sorted when scaling down, and only for Packed strategy.


pkg/apis/stable/v1alpha1/gameserversort.go
```diff
@@ -0,0 +1,52 @@
...
+
+import "sort"
+
+// GameServerSort sorts gameservers, based on a comparator
+type GameServerSort struct {
+	// List is the GameServer list to sort
+	List []*GameServer
+	// Less is the comparator function to use as a sorting function.
+	// returns true if a < b
+	Comparator func(a, b *GameServer) bool
```
jkowalski: nit: call it "Less" ?

pkg/apis/stable/v1alpha1/gameserversort.go
```diff
@@ -0,0 +1,52 @@
...
+func (g *GameServerSort) Less(i, j int) bool {
+	a := g.List[i]
+	b := g.List[j]
+
+	return g.Comparator(a, b)
+}
+
+// Swap is the swap function for sort.Interface
+func (g *GameServerSort) Swap(i, j int) {
+	tmp := g.List[j]
```
jkowalski: nit: canonical thing to do is to do multi-assignment on both sides:

`a, b = b, a`

pkg/apis/stable/v1alpha1/gameserversort.go
```diff
@@ -0,0 +1,52 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+import "sort"
+
+// GameServerSort sorts gameservers, based on a comparator
+type GameServerSort struct {
```
jkowalski: overkill perhaps? sort.Slice is really fast for all practical data sizes, have you measured it and found it not to be the case?

pkg/gameserversets/controller.go
```diff
@@ -353,33 +356,38 @@ func (c *Controller) syncGameServerSet(key string) error {
 
 // computeReconciliationAction computes the action to take to reconcile a game server set set given
 // the list of game servers that were found and target replica count.
-func computeReconciliationAction(list []*v1alpha1.GameServer, targetReplicaCount int, maxCreations int, maxDeletions int, maxPending int) (int, []*v1alpha1.GameServer, bool) {
-	var upCount int // up == Ready or will become ready
+func (c *Controller) computeReconciliationAction(gsSet *v1alpha1.GameServerSet, list []*v1alpha1.GameServer,
```
jkowalski: instead of binding to "c" please pass counts as an argument. easier to test this as a pure function vs function that can freely reach into "c".

pkg/gameserversets/controller.go
```diff
@@ -353,33 +356,38 @@ func (c *Controller) syncGameServerSet(key string) error {
...
+	targetReplicaCount int, maxCreations int, maxDeletions int, maxPending int) (int, []*v1alpha1.GameServer, bool) {
+	var upCount int     // up == Ready or will become ready
+	var deleteCount int // number of gameservers to delete
 
 	// track the number of pods that are being created at any given moment by the GameServerSet
 	// so we can limit it at a throughput that Kubernetes can handle
 	var podPendingCount int // podPending == "up" but don't have a Pod running yet
+
+	var potentialDeletions []*v1alpha1.GameServer
 	var toDelete []*v1alpha1.GameServer
```
jkowalski: not sure why you need this change? just sort accordingly where we were sorting before?

pkg/gameserversets/controller.go
jkowalski: one idea: pass "deleteOrdering func(a, b *v1alpha1.GameServer)" and have the caller pass either the one that sorts based on time or node fullness

pkg/apis/stable/v1alpha1/gameserversort.go
markmandel: That was my first attempt - Go complained that I had a function and a member called `Less`, and wasn't very happy with me. `Comparator` was the best thing I could come up with.

pkg/apis/stable/v1alpha1/gameserversort.go
markmandel: Nice! Done.

pkg/apis/stable/v1alpha1/gameserversort.go
markmandel: TL;DR - sort.Slice uses reflection. Reflection in go is _slow_. At small scale sort.Slice is probably fine, but sorting several thousand gameservers, I'll take all the performance gains we can.

Context:
https://stackoverflow.com/a/54276484

pkg/gameserversets/controller.go
markmandel: When running as `Distributed`, we're grabbing a random set of gameservers to delete to bring us back down under the declared amount we want. We can sort by time, but it didn't seem like very valuable an operation, given the relatively random nature of the selection.

pkg/gameserversets/controller.go
markmandel: If we took this away, we would need to push everything that could be deleted into `toDelete`, sort that and then trim down to min(`deletedCount`, `maxDeletions) from there.

Also, if there are GameServers that are in Error on Unhealthy states, they may get pushed back / would need to be specially sorted to the front.

TL;DR - I think it makes the code harder to read, and harder to keep separate and test to track everything in toDelete.

pkg/gameserversets/controller.go
markmandel: > instead of binding to "c" please pass counts as an argument. easier to test this as a pure function vs function that can freely reach into "c".

The only downside there, is that we need to grab a copy of Counts on every pass through, rather than just on scale down, when Packed. It's relatively small operation, but it does involve a lock and a datastructure copy. Are we happy with that? Doing it this way limited any overhead on scale up.

pkg/apis/stable/v1alpha1/gameserversort.go
jkowalski: sort.Slice uses reflection basically once or twice per call, not once per comparison (see the source code). This is only to figure out the length of the slice and generate a swapper function, after that it's not using reflection anymore.

pkg/gameserversets/controller.go
jkowalski: we're not CPU-bound at this point and this would be happening once per GSS, I would not worry about performance here

pkg/gameserversets/controller.go
jkowalski: that's exactly what we do today, I don't think performance is a problem. Sorting 10000 things on a modern CPU is basically nothing (they can do billions of things per second per core...)

pkg/gameserversets/controller.go
markmandel: Sounds good - I'll make the change :+1:

pkg/apis/stable/v1alpha1/gameserversort.go
markmandel: That's fair. I can make that change.

pkg/gameserversets/controller.go
markmandel: From my understanding of the code, we were getting a random sample of gameservers to delete (since and informer list is in random order) - and then sorting that random set after the fact - which doesn't.

Whereas this is tracking everything that could be deleted, and then sorting from there, which gives us a stronger guarantee of order of deletion by removing the random aspect.

Even if I did remove `potentialAllocations`, I'm just replacing it with `toDelete` but the code wouldn't be the same.

markmandel: Sorry and to be clear - this isn't a performance issue - it's that the current implementation provides us with a random set of gameservers to delete, there's no way to guarantee anything near to consistent delete ordering when running Packed.

markmandel:
As requested - #639 is now available - moving the refactor out from this PR, so it can be reviewed independently :+1:

pkg/gameservers/gameservercounter.go
```diff
@@ -44,19 +44,21 @@ type AllocationCounter struct {
...
+func NewGameServerCounter(
 	kubeInformerFactory informers.SharedInformerFactory,
-	agonesInformerFactory externalversions.SharedInformerFactory) *AllocationCounter {
+	agonesInformerFactory externalversions.SharedInformerFactory) *GameServerCounter {
 
 	gameServers := agonesInformerFactory.Stable().V1alpha1().GameServers()
 	gsInformer := gameServers.Informer()
 
-	ac := &AllocationCounter{
+	ac := &GameServerCounter{
```
aLekSer: nit name also could be changed
```suggestion
	gc := &GameServerCounter{
```

pkg/gameservers/gameservercounter.go
```diff
@@ -131,7 +133,8 @@ func NewAllocationCounter(
 }
 
 // Run sets up the current state GameServer counts across nodes
-func (ac *AllocationCounter) Run(stop <-chan struct{}) error {
+// non blocking Run function.
+func (ac *GameServerCounter) Run(_ int, stop <-chan struct{}) error {
```
aLekSer: Nit: same as above
```suggestion
func (gc *GameServerCounter) Run(_ int, stop <-chan struct{}) error {
```

pkg/gameservers/gameservercounter_test.go
```diff
@@ -29,10 +29,16 @@ import (
...
+	defaultNs = "default"
+	name1     = "node1"
+	name2     = "node2"
+)
+
+func TestGameServerCounterGameServerEvents(t *testing.T) {
 	t.Parallel()
 
-	ac, m := newFakeAllocationCounter()
+	ac, m := newFakeGameServerCounter()
```
aLekSer: Minor change
```suggestion
	gc, m := newFakeGameServerCounter()
```

markmandel:
Assuming this passes tests, this should be good for re-review. Made changes based on above, and I think this should be better now.

cmd/controller/main.go
```diff
@@ -201,7 +201,7 @@ func main() {
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
 
 	rs = append(rs,
-		httpsServer, gsCounter, gsController, gsSetController, fleetController, faController, fasController, gasController, server)
+		httpsServer, gsCounter, gsCounter, gsController, gsSetController, fleetController, faController, fasController, gasController, server)
```
aLekSer: Why do we need two gsCounter in one `rs` slice? Possibly we need to add a comment for that?
Also I think that comment in `perNodeCounter` Run() should be updated, because it is now blocking with mutex:
 https://github.com/GoogleCloudPlatform/agones/blob/b133e5280d2bec4bee98f78f59561cc2dc5a7171/pkg/gameservers/prenodecounter.go#L136

markmandel: whoops - that looks like a typo on my end. Let me remove that!

markmandel: Fixed. The comment was to say that the function doesn't block on `stop` (which many Run functions do) - I can change it to be more clear?

pkg/gameserversets/controller.go
markmandel: ...and done!

pkg/gameserversets/controller.go
markmandel: also updated back in the newest sort for Distributed :+1:

cmd/controller/main.go
aLekSer: Thanks for the clarification, I think no need to change comment, cause this file was not touched here. 🙇

PR #639, opened 2019-03-07
Refactor AllocationCounter to GameServerCounter

markmandel:
Precursor to #638 - this moves the AllocationCounter code to a more central GameServerCounter (which I think is a better name) that tracks the number of Ready and Allocated GameServers that are
available on each node.

These details are useful for sorting for `Packed` scheduling strategies.

Once this PR is completed, we can use this Count() values provided by this controller in the GameServerSet scale down logic, to ensure that GameServers on the least used nodes are removed first when using a Packed strategy.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -184,6 +184,8 @@ func main() {
 
 	allocationMutex := &sync.Mutex{}
 
+	gsCounter := gameservers.NewPerNodeCounter(kubeInformerFactory, agonesInformerFactory)
+
 	gsController := gameservers.NewController(wh, health,
 		ctlConf.MinPort, ctlConf.MaxPort, ctlConf.SidecarImage, ctlConf.AlwaysPullSidecar,
 		ctlConf.SidecarCPURequest, ctlConf.SidecarCPULimit,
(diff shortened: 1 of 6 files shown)


pkg/gameservers/gameservercounter.go
```diff
@@ -166,7 +169,7 @@ func (ac *AllocationCounter) Run(stop <-chan struct{}) error {
 }
 
 // Counts returns the NodeCount map in a thread safe way
-func (ac *AllocationCounter) Counts() map[string]NodeCount {
+func (ac *GameServerCounter) Counts() map[string]NodeCount {
```
aLekSer: Nit: same as in the previous PR
```suggestion
func (gc *GameServerCounter) Counts() map[string]NodeCount {
```

pkg/gameservers/gameservercounter.go
```diff
@@ -29,11 +29,11 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-// AllocationCounter counts how many Allocated and
+// GameServerCounter counts how many Allocated and
```
jkowalski: nit: can we call it PerNodeCounter or something likethat? this is to avoid stuttering "gameservers.GameServerCounter", instead "gameservers.PerNodeCounter"

markmandel: I like it. That reads much better. Making this change!

Issue #409, opened 2018-11-09
Documentation for the extended Kubernetes API


markmandel:
When we implement CRDs for Agones, we create new API endpoints for each CRD.

It would be great to document these, and publish it somewhere, so that people can more easily integrate, and we can reference it in our documentation.

I feel there are several steps:

1. [ ] Create a `make` target that generates the documentation
2. [ ] Determine where it can be hosted (App engine?), integrate it into our build system and domain controls (@markmandel would need to do this part, most likely)

## Resources
- Kubebuilder has a tool, maybe we can leverage that - https://book.kubebuilder.io/beyond_basics/generating_documentation.html

markmandel:
This is currently blocked by: https://github.com/kubernetes-sigs/kubebuilder/issues/483

Will wait for it to get fixed.

markmandel:
Another tool to try out: https://github.com/ahmetb/gen-knative-api-reference-docs

ahmetb:
Any interest? I can help you get this done in ~20 minutes.

markmandel:
Yep, definitely! Just haven't got to it yet!

Will need to integrate it into our build image,  and our Make targets so we can regenerate on model change though!

aLekSer:
I see that Kubebuilder has an issue and functionality "Generate docs" seems to be not supported yet.
https://github.com/kubernetes-sigs/kubebuilder/issues/483
Therefore this @ahmetb's project would be really helpful.
@markmandel I can help with this task. By the way should we integrate all these pages to Hugo ?

markmandel:
Yeah, if there is a way we can integrate this into agones.dev somehow, I think that would be ideal :+1: 

Maybe whenever we run `make site-static` is generates the API?

aLekSer:
I need to investigate more on how generated docs would look like and how to integrate them into `site-static` flow.

markmandel:
@aLekSer can we close this now?

markmandel:
Whoops - hit the wrong button, but @aLekSer should we reopon this?

aLekSer:
Agree with closing that, I think the main part is done, the rest is to polish current version, only some small warnings while parsing "./pkg/apis/stable/v1alpha1/" occurred:
```
I0401 07:45:21.445044       8 main.go:116] parsing go packages in directory ../../../agones.dev/agones/pkg/apis/stable/v1alpha1/
[...]

PR #645, opened 2019-03-12
Add Agones Kubernetes API docs generator

aLekSer:
Add [gen-crd-api-reference-docs](https://github.com/ahmetb/gen-crd-api-reference-docs) utility into build-image, generate Agones CRD reference html document, map this doc to existent docs structure.
Covers #409.

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -123,6 +123,11 @@ RUN mkdir /tmp/hugo && \
 RUN curl -sL https://deb.nodesource.com/setup_11.x | bash - && \
     apt-get install -y nodejs
 
+# install API reference docs generator
+RUN mkdir -p /go/src/github.com/ahmetb && \
+    cd /go/src/github.com/ahmetb && git clone https://github.com/ahmetb/gen-crd-api-reference-docs && \
+    cd ./gen-crd-api-reference-docs && go build
+
 # html checker
 RUN mkdir /tmp/htmltest && \
     wget -O /tmp/htmltest/htmltest.tar.gz https://github.com/wjdp/htmltest/releases/download/v0.10.1/htmltest_0.10.1_linux_amd64.tar.gz && \
--- a/build/includes/website.mk
+++ b/build/includes/website.mk
@@ -23,7 +23,7 @@
 #
 # Website targets
 #
-
+	
 # generate the latest website
 site-server: ARGS ?=-F
 site-server: ENV ?= RELEASE_VERSION="$(base_version)" RELEASE_BRANCH=master
@@ -59,6 +59,24 @@ site-deploy-preview: site-static-preview
 
 site-test:
 	# generate actual html and run test against - provides a more accurate tests
+	$(MAKE) test-gen-api-docs
 	$(MAKE) site-static-preview
 	docker run --rm -t -e "TERM=xterm-256color" $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag) bash -c \
 		"mkdir -p /tmp/website && cp -r $(mount_path)/site/public /tmp/website/site && htmltest -c $(mount_path)/site/htmltest.yaml /tmp/website"
+
+# Path to a file and docker command
+REL_PATH := content/en/docs/Reference/agones_crd_api_reference.html
+GEN_API_DOCS := docker run -e FILE="$(mount_path)/site/$(REL_PATH)" -e VERSION=${base_version} --rm -i $(common_mounts) $(build_tag) bash -c "/go/src/agones.dev/agones/site/gen-api-docs.sh"
+
+# generate Agones CRD reference docs
+gen-api-docs: ensure-build-image
+	$(GEN_API_DOCS)
+
+# test generated Agones CRD reference docs
+test-gen-api-docs: expected_docs := $(site_path)/$(REL_PATH)
+test-gen-api-docs: ensure-build-image
+	cp $(expected_docs) /tmp/generated.html
+	sort /tmp/generated.html > /tmp/generated.html.sorted
+	$(GEN_API_DOCS)
+	sort $(expected_docs) > /tmp/result.sorted
+	diff -bB /tmp/result.sorted /tmp/generated.html.sorted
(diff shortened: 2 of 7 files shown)


aLekSer:
This is an initial commit. Some changes to the templates could occur in the future. Current look and feel:


Agones Kubernetes CRD docs

aLekSer:
Issue with resulting file position in Cloud Build:
```
Output to a file /go/src/agones.dev/agones/site/content/en/docs/Reference/agones_crd_api_reference.html
diff /workspace/site/content/en/docs/Reference/agones_crd_api_reference.html /tmp/generated.html
```
Locally I have for `test-gen-api-docs: expected_docs := $(site_path)/$(REL_PATH)` :
```
diff /Users/alexander.apalikov/go/src/agones.dev/agones/site/content/en/docs/Reference/agones_crd_api_reference.html /tmp/generated.html
```

aLekSer:
Found an issue, I have added the special kubebuilder declaration in `/pkg/apis/stable/v1alpha1/gameserver.go`, which is redundant, like that:
```
// +kubebuilder:resource:path=gameservers
```

build/includes/website.mk
```diff
@@ -24,6 +24,23 @@
 # Website targets
 #
 
+REL_PATH := content/en/docs/Reference/agones_crd_api_reference.html
+API_DOC_PATH := "$(mount_path)/site/$(REL_PATH)"
+GEN_API_DOCS ?= docker run -e FILE=$(API_DOC_PATH) --rm -i $(common_mounts) $(build_tag) bash -c "/go/src/agones.dev/agones/site/gen-api-docs.sh"
```
markmandel: Curious as to the motivation to put all of this in variables? Seems like it wouldn't change?

build/includes/website.mk
```diff
@@ -59,6 +76,7 @@ site-deploy-preview: site-static-preview
 
 site-test:
 	# generate actual html and run test against - provides a more accurate tests
+	$(MAKE) test-gen-api-docs
```
markmandel: I don't have a good answer for this - but because the website works off master - if we change our CRD data structures, that will change the website as well, and be out of sync until next release.

Only idea I've got - that in the advent of that happening - rename the old page and put an `expiryDate` of the next release in the toml.... but how would we get a publishDate in the new one, and still pass tests?

I'm just thinking out loud here, but any ideas?

build/includes/website.mk
aLekSer: I have added those options to be able to split test html result and normal one. GEN_API_DOCS used twice, that's why it was extracted.

build/includes/website.mk
aLekSer: Tried to use custom short code in front-matter and it seems that variables are not supported there.
So currently I am thinking of combining one full generated html with `expiryVersion` shortcode. And new one with `publishVersion` shortcode in the same file.
And during test just compare the second half with `publishVersion`. And during release move bottom half to the top and generate new one at the bottom.

markmandel: That sounds fancy - but makes sense to me! I assume you can make that work? :smile:

aLekSer: I have updated `gen-api-docs.sh` script. It now handles `base_version` environment variable.
What it does is get all lines before `publishVersion` line. Then extract what in the middle of this feature shortcode. Compares with what was generated if diff then do version compare.
Then set `expiryVersion` of old part with new `base_version` fresh generated doc now have new `publishVersion`.

aLekSer:
Reworked generation script it now handles the version in `publishVersion` shortcode. 
Old code becomes expiredVersion with new `base_version`. Test now compares only what is in the middle of `publishVersion` with the output of go tool [gen-crd-api-reference-docs](https://github.com/ahmetb/gen-crd-api-reference-docs).
Two calls to `awk` and one `sed` do the job.
We don't need to care about switching the versions on release process - it is automated for this file.

aLekSer:
I think I need to add md instructions on how to use new make target `gen-api-docs`. It was created to work in a similar way to  `make gen-install` when we call this on each crd update and just `git add` the resulted `.../Reference/agones_crd_api_reference.html`. So basically two subsequent calls of `	$(MAKE) test-gen-api-docs` would result in pass the test, but on CI if someone forgot to commit this html than test would fail.

aLekSer:
Fixed one issue, there was a commit-id in a default template. So I removed it, so now different git commit would generate the same html.
https://github.com/GoogleCloudPlatform/agones/pull/645/files#diff-b75745dde8c5e52ecb85780a42ac3669

markmandel:
Apologies - looks like I never submitted those comments :/ they've been sitting there forever. But otherwise, this is good to go.

site/content/en/docs/Guides/access-api.md
```diff
@@ -10,7 +10,11 @@ description: >
 ---
 
 Installing Agones creates several [Custom Resource Definitions (CRD)](https://kubernetes.io/docs/concepts/api-extension/custom-resources),
-which can be accessed and manipulated through the Kubernetes API.
+which can be accessed and manipulated through the Kubernetes API. 
+
+{{% feature publishVersion="0.9.0" %}}
```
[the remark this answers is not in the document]
aLekSer: Yes, I can fix that version, but then I need to check that test still pass.

aLekSer:
Merged master, fixed publishVersion and Makefile parameter.

build/build-image/Dockerfile
```diff
@@ -123,6 +123,11 @@ RUN mkdir /tmp/hugo && \
 RUN curl -sL https://deb.nodesource.com/setup_11.x | bash - && \
     apt-get install -y nodejs
 
+# install API reference docs generator
+RUN mkdir -p /go/src/github.com/ahmetb && \
+    cd /go/src/github.com/ahmetb && git clone https://github.com/ahmetb/gen-crd-api-reference-docs && \
```
ahmetb: I think you should probably pin to a commit here.
Given there's a copy of the template in this repo, I don't want to accidentally break the downstream.

aLekSer: Thanks @ahmetb for the review. I was thinking about that, will provide an update in a separate PR.

PR #655, opened 2019-03-15
Fix feature shortcode for Hugo

### LINKED ISSUE #650: Feature shortcode does not behave correctly for versions > "0.10.0" (2 digit minor version)
Steps to reproduce:
Set `expiryVersion` to `0.10.0` and `base_version` to 0.9.0 in one of md files.
Run `make site-server`. 
We will not see the doc part inside the shortcode:
```
{{% feature expiryVersion="0.10.0" %}}
{{% /feature %}}
```

While `{{% feature expiryVersion="0.9.1" %}}` works as expected.
The reason seems to be lexicographic comparison between strings, we should split the version into parts and compare them separately.

aLekSer:
Algorithm for splitting version strings by dots and comparing major, minor versions one by one separately. There is no break statement for range loop in Hugo functions, that's why added special handling for that case.

Closes #650 .

### DIFF
--- a/site/layouts/shortcodes/feature.html
+++ b/site/layouts/shortcodes/feature.html
@@ -1,6 +1,36 @@
 {{- $version := getenv "RELEASE_VERSION" | default $.Page.Site.Params.release_version }}
 {{- $publishVersion := (.Get "publishVersion" ) | default "0.0.0" }}
 {{- $expiryVersion := (.Get "expiryVersion") | default "9999.0.0"}}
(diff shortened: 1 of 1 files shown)


aLekSer:
By default if we access element over the list length we would get 0 - and this would make shortcode work even for `version = 1.99.0` and `publishVersion = 1.9`. In this function `$p=0` (length of `publDigits` is 3):
```
    {{- $p := int (index $publDigits 4)}}
```

PR #656, opened 2019-03-15
This is an experimental e2e tests runner on kind

Kuqd:
Draft PR for my experiment with e2e kind on cloud builder.


Kuqd:
I'm going no where, we need more options from kind, specifically to bound the api server to something else than only localhost. I've tried --net host on GCB with no luck.

PR #659, opened 2019-03-16
Lightweight library for implementing APIServer extensions

markmandel:
The other API extension libraries were too opinionated, or not flexible enough to provide exactly what we need -- register a single noun in the API, not have storage, give full control over the http lifecycle.

This provides just enough functionality for what we need, while also providing extension points in case we need to expand the API surface area in the future - such as a more explict OpenAPI spec.

This is the foundational work that is required for converting GameServerAllocations into an API Extension - next step will be to use this library to implement that functionality.


markmandel:
It will likely make things easier to review, but here is a (slightly out of date) example - https://github.com/markmandel/agones/blob/feature/gsa-api-server/pkg/gameserverallocations/controller.go#L104

So you can see how the pieces end up fitting together.

(I need to update this to use go.mod instead as well)

pkg/util/apiserver/apiserver.go
```diff
@@ -0,0 +1,220 @@
...
+	s := &APIServer{
+		mux:          mux,
+		resourceList: map[string]*metav1.APIResourceList{},
+		swagger:      &spec.Swagger{SwaggerProps: spec.SwaggerProps{}},
+		delegates:    map[string]CRDHandler{},
+	}
+	s.logger = runtime.NewLoggerWithType(s)
+
+	// we don't *have* to have a swagger api, so just do an empty one for now, and we can expand as needed.
+	// kube-openapi could be a potetial library to look at for future if we want to be more specific.
```
Kuqd:
```suggestion
	// kube-openapi could be a potential library to look at for future if we want to be more specific.
```

pkg/util/apiserver/apiserver.go
```diff
@@ -0,0 +1,220 @@
...
+	return func(w http.ResponseWriter, r *http.Request) error {
+		namespace, resource, err := splitNameSpaceResource(r.URL.Path)
+		if err != nil {
+			https.FourZeroFour(as.logger.WithError(err), w, r)
+			return nil
+		}
+
+		delegate, ok := as.delegates[fmt.Sprintf("%s/%s", gv, resource)]
+		if !ok {
+			https.FourZeroFour(as.logger, w, r)
```
Kuqd: nitpick: I don't think 404 is meaningful here. 421 ? 501 ? 502 ?

pkg/util/apiserver/apiserver.go
```diff
@@ -0,0 +1,220 @@
...
+			return nil
+		}
+
+		err = delegate(w, r, namespace)
```
Kuqd:
```
		if err := delegate(w, r, namespace);err != nil {
		    return err
		}
```

pkg/util/apiserver/apiserver.go
```diff
@@ -0,0 +1,220 @@
...
+	if !ok {
+		return info, errors.Errorf("Could not find serializer for Accept: %s", header)
+	}
+
+	return info, nil
+}
+
+// splitNameSpaceResource returns the namespace and the type of resource
+func splitNameSpaceResource(path string) (namespace, resource string, err error) {
+	list := strings.Split(strings.Trim(path, "/"), "/")
```
Kuqd: interesting way to use default value with returned variable.

pkg/util/apiserver/apiserver.go
Kuqd: reading your tests actually it make sense.

pkg/util/apiserver/apiserver_test.go
```diff
@@ -0,0 +1,220 @@
...
+func TestSplitNameSpaceResource(t *testing.T) {
+	type expected struct {
+		namespace string
+		resource  string
+		isError   bool
+	}
+
+	fixtures := []struct {
+		path     string
+		expected expected
```
Kuqd: you could use an anonymous struct.

Kuqd:
Another thing but I'm ok with it, I don't think https is a good package name I understand why but I would prefer that you named it http and import it as https.

pkg/util/apiserver/apiserver.go
markmandel: Done

pkg/util/apiserver/apiserver.go
markmandel: You know, I can't remember if I did this deliberately :thinking:

pkg/util/apiserver/apiserver_test.go
markmandel: AFAIK you can't nest anonymous structs, so I've always used this approach.

markmandel:
> LGTM - What kind of operations do we want to support ? GET/POST for sure but have you anything else in mind ?

So, since we control the entire HTTP response - all HTTP verbs will be passed down to the Handler - then it's up to the handler to decide what to do with  them. For example, we do this here: https://github.com/markmandel/agones/blob/feature/gsa-api-server/pkg/gameserverallocations/controller.go#L153-L157

So I guess we already support all of them? :smile: 

> Another thing but I'm ok with it, I don't think https is a good package name I understand why but I would prefer that you named it http and import it as https.

As you may have guessed, this server _has_ to be https, so it felt like https was a good package name. Maybe it should just be something more like `web` or `httpsserver` - but that didn't feel very go like? `http` feels misleading, since it has to be https - maybe http/TLSServer ? I'm not sure :man_shrugging: Happy to do a refactor PR if we can come up with a better package name :+1:

pkg/util/apiserver/apiserver.go
Kuqd: I don't see any issue, it just gave me idea for my own projects :)

pkg/util/apiserver/apiserver_test.go
Kuqd: ho yes definitively.

Kuqd:
let's keep it like this, may be later we will have a better view.

Issue #625, opened 2019-02-26
Move to go modules


markmandel:
Apparently [Go Modules](https://github.com/golang/go/wiki/Modules) are a thing now :smile: and where Go is headingin the near future (1.13 I think they will be default).

We should convert over from dep to the modules at some point in the future.

heartrobotninja:
I will be working on this shortly.

aLekSer:
@heartrobotninja  Please don't forget about #414 which currently was manually fixed in /vendor
https://github.com/kubernetes/kubernetes/pull/70277 is merged, but not sure about target Kubernetes version.

aLekSer:
@heartrobotninja  also note that we can have use vendors with modules at the same time to mitigate mentioned above issue:
https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away

heartrobotninja:
Here is what we have so far: 

Moving straight to go modules is ideal, but not doable right now. We have code in vendor that contains a fix that will not be released until March 25th, and there is no guarantee when the platforms will update to 1.14 since they are still on 1.12. 

Given this, it might be best that we move to go modules while also keeping the vendor directory. This would require re-implementing the fix from #414. If we are ok with doing that, I can do both parts fairly quickly.

markmandel:
It sounds like a way forward, but what does "_re-implementing the fix_" entail? Just making sure it's pasted into the new vendor directory?

Question - with go modules,without a vendor directory, does that mean on each CI build, we would have to download all our dependencies every time?

heartrobotninja:
re-implementing would just be re-applying the fix. I know there is a cache for go modules, but I am currently trying to hunt down how that works with CI builds.

markmandel:
Gotcha. We'd likely want to store that in GCS and move it back and forth.

heartrobotninja:
As for CI:
We could either cache all dependencies prior to building. This still makes us download still, but we do it in advance of building. There might be some magic we can do there where we keep the cache warmed. 

The other option is doing module-aware vendoring. This is mostly what we do now, so will likely consist of the least changes. Both are supported and I would propose we just do module-aware vendoring for now.

markmandel:
> I would propose we just do module-aware vendoring for now.

SGTM 👍

heartrobotninja:
[...]

PR #661, opened 2019-03-17
Switch Agones to Go Modules

heartrobotninja:
Most of the files in this PR are in vendor

For a detailed explanation see the [Design Doc](https://goo.gl/uNhtJ2).

This is for issue #625


build/README.md
```diff
@@ -664,6 +668,14 @@ If you get this error when building Agones in WSL (`make build`, `make test` or
 - Are your project files on a different folder than C? If yes, then you should either move them on drive C or set up Docker for Windows to share your project drive as well
 - Did you set up the volume mount for Docker correctly? By default, drive C is mapped by WSL as /mnt/c, but Docker expects it as /c. You can test by executing `ls /c` in your linux shell. If you get an error, then follow the instructions for [setting up volume mount for Docker](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly#ensure-volume-mounts-work)
 
+#### Error: cluster-admin-binding already exists
+
```
markmandel: As a separate PR, should we make the name of `cluster-admin-binding` unique to the user? (short Hash of the account name or something?)

build/build-image/Dockerfile
```diff
@@ -90,18 +91,14 @@ RUN echo "source <(helm completion bash)" >> /root/.bashrc
 #   |___/                                               |___/
 
 # install go-proto-gen 1.1
```
markmandel: Since we're here, should we update this comment as well? :smile:

build/README.md
```diff
@@ -646,7 +650,7 @@ Cleans up your custom cluster by reseting Helm.
 
 ## Dependencies
 
-This project uses the [dep](https://github.com/golang/dep) as a dependency manager. You can see the list of dependencies [here](https://github.com/GoogleCloudPlatform/agones/blob/master/Gopkg.toml).
+This project uses the [go modules](https://github.com/golang/go/wiki/Modules) as its manager. You can see the list of dependencies [here](https://github.com/GoogleCloudPlatform/agones/blob/master/go.mod).
```
markmandel: (Assuming I understand this correctly) - do we want to put a section in here explaining that we are using vendoring, and that requires the use of `go mod vendor` to save dependency changes  (https://tip.golang.org/cmd/go/#hdr-Make_vendored_copy_of_dependencies)

..and maybe we should add something about how we have some vendored patches for apimachinery? :man_shrugging:  WDYT?

heartrobotninja: It wouldn't hurt and would be more clear for people reading through so I am all for it.

build/build-image/Dockerfile
heartrobotninja: Might as well :-)

build/README.md
heartrobotninja: This I am not sure about and would have to look into that a bit more. Maybe someone with more Kubernetes knowledge can comment.

build/README.md
markmandel: Fair enough. I'll write up a small ticket for it - was just thinking out loud. 

Otherwise the rest of this LGTM!

build/README.md
markmandel: Fair enough. I'll write up a small ticket for it - was just thinking out loud.

Otherwise the rest of this LGTM!

build/README.md
heartrobotninja: Done.

build/build-image/Dockerfile
heartrobotninja: Done.

markmandel:
Going to take this for a spin and if all works well, then I'll approve (it should, since it passed CI, but just want to get the human feel)

One thing - please squash this down to one commit, per [contributing guidelines](https://github.com/GoogleCloudPlatform/agones/blob/master/CONTRIBUTING.md#submitting-code-via-pull-requests), thanks!

Will try and get this done today, GDC forgiving :+1:

Issue #657, opened 2019-03-15
Terraform scripts for Agones


markmandel:
From my understanding,  Terraform can both create a Kubernetes cluster, and install Agones in one go.

Looking at the increase  in complexity of the K8s installation (taints for metrics and controller), doing this by hand seems fraught with peril.

I'm thinking we would
1. Create the cluster with the right sizes of nodes with the right taints
1. Sets up appropriate firewall rulels
1. Installs the right version of Agones
1. (Somehow?) works across providers? (I don't know enough about Terraform to know if this is possible).
1. Maybe also replace our development deployment manager scripts with this as well, so we're all using the same codebase.

And maybe more.

What do people think?

aLekSer:
I will investigate the process of Agones deployment with terraform. I would work on points 1-3 of the above.

Kuqd:
4 - need to be separated.
3 - should be supported with helm provider.

aLekSer:
Also I found that there exist also `Kubernetes` provider which possibly has more universal nature (cloud agnostic). But I have not tested it.
https://medium.com/devopslinks/how-to-setup-terraform-with-kubernetes-provider-356d0778e771

thisisnotapril:
@markmande fyil I am reaching out to someone in the k8s community that may be interested in working on this

aLekSer:
I think we should add here `kubectl` setup step.

aLekSer:
Next step as I see it is to modularize GKE cluster creation config and prepare such modules for other Cloud providers (can also be baremetal one or minikube to start with):
To accomplish next point:
>4. (Somehow?) works across providers? (I don't know enough about Terraform to know if this is possible).

roberthbailey:
@aLekSer - we have some terraform scripts now. Can you document what you think remains to close out this issue?

aLekSer:
As from the original message we probably going to skip next one, current gcloud utility way looks much smoother:
>5. Maybe also replace our development deployment manager scripts with this as well, so we're all using the same codebase.
I would figure out todo list for this topic tomorrow.

aLekSer:
I think what we should do next is to update Terraform we are using from 0.11.3 to [0.12](https://www.hashicorp.com/blog/announcing-terraform-0-12). Released this May. 
Other than that I think we can close this Ticket and create one for updating Terraform in our build system.

[...]

PR #670, opened 2019-03-25
Terraform support to generate test cluster

aLekSer:
E2E test passed. 
New make targets:
1. terraform-init
1. gcloud-terraform-cluster
1. gcloud-terraform-destroy-cluster

Need to add install step of agones.

For #657 .

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -379,6 +379,13 @@ The Kubernetes config file used to access the cluster. Defaults to `~/.kube/conf
 ### CLUSTER_NAME
 The (gcloud) test cluster that is being worked against. Defaults to `test-cluster`
 
+### GCP_PROJECT
+Your GCP project for deploying GKE cluster. Defaults to gcloud default project settings.
+
+### GKE_PASSWORD
+If specified basic authentication would be enabled for your cluster with username "admin".
+Empty string `""` would disable basic authentication.
+
 ### IMAGE_PULL_SECRET
 The name of the secret required to pull the Agones images, if needed.
 If unset, no pull secret will be used.
@@ -541,6 +548,27 @@ Pulls down authentication information for kubectl against a cluster, name can be
 #### `make gcloud-auth-docker`
 Creates a short lived access to Google Cloud container repositories, so that you are able to call
 `docker push` directly. Useful when used in combination with `make push` command.
+
+### Terraform
+
+Utilities for deploying a Kubernetes Engine cluster on Google Cloud Platform using `google` Terraform provider.
+
+#### `make terraform-init`
+Install `google` and `google-beta` terraform providers and authorize.
+
+#### `make gcloud-terraform-cluster`
+Run next command with your project ID specified:
+```
+[GKE_PASSWORD="<YOUR_PASSWORD>"] make gcloud-terraform-cluster
+```
+Where `<YOUR_PASSWORD>` should be at least 16 characters in length. You can omit GKE_PASSWORD and then basic auth would be disabled. Also you change `ports="7000-8000"` setting using tfvars file.
+Also you can define password `password=<YOUR_PASSWORD>` string in `build/terraform.tfvars`.
+
+#### `make gcloud-terraform-destroy-cluster`
+Run `terraform destroy` on your cluster.
+
+#### `make terraform-clean`
+Remove .terraform directory with configs as well as tfstate files.
 
 ### Minikube
 
--- a/build/cluster.tf
+++ b/build/cluster.tf
@@ -1,0 +1,189 @@
+# Copyright 2019 Google LLC All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+provider "google-beta" {
+  zone      = "${lookup(var.cluster, "zone")}"
+}
+
+
+# Password for the Kubernetes API.
+# Could be defined using GKE_PASSWORD env variable
+# or by setting `password="somepass"` string in build/terraform.tfvars
+variable "password" {default = ""}
+variable "username" {default = "admin"}
+
+# Ports can be overriden using tfvars file
+variable "ports" {default="7000-8000"}
+
+# Set of GKE cluster parameters which defines its name, zone
+# and primary node pool configuration.
+# It is crucial to set valid ProjectID for "project".
+variable "cluster"  {
+  description = "Set of GKE cluster parameters."
+  type = "map"
+  default = {
+      "zone"             = "us-west1-c"
+      "name"             = "test-cluster"
+      "machineType"      = "n1-standard-4"
+      "initialNodeCount" = "4"
+      "legacyAbac"       = false
+      "project"          = "agones"
+  }
+}   
+
+
+# echo command used for debugging purpose
+# Run `terraform taint null_resource.test-setting-variables` before second execution
+resource "null_resource" "test-setting-variables" {
+    provisioner "local-exec" {
+      command = "${"${format("echo Current variables set as following - name: %s, project: %s, machineType: %s, initialNodeCount: %s, zone: %s, legacyAbac: %s",
+      "${lookup(var.cluster, "name")}", "${lookup(var.cluster, "project")}",
+      "${lookup(var.cluster, "machineType")}", "${lookup(var.cluster, "initialNodeCount")}",
+      "${lookup(var.cluster, "zone")}", "${lookup(var.cluster, "legacyAbac")}")}"}"
+    }
+}
+
+
+locals {
+  username = "${var.password != "" ? var.username : ""}"
+}
+
+# assert that password has correct length
+# before creating the cluster to avoid 
+# unfinished configurations
+resource "null_resource" "check-password-length" {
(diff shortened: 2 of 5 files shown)


build/gke-test-cluster/cluster.tf
```diff
@@ -0,0 +1,153 @@
...
+  name     = "${lookup(var.cluster, "name")}"
+  zone     = "${lookup(var.cluster, "zone")}"
+  project = "${lookup(var.cluster, "project")}"
+  provider = "google-beta"
+
+  initial_node_count = "${lookup(var.cluster, "initialNodeCount")}"
+
+  master_auth {
+    username = "admin"
+    password = "supersecretpassword"
```
bbf: This should also be externalized and perhaps even set to fail if not changed from the default (perhaps we shouldn't even have a default for this).

build/gke-test-cluster/cluster.tf
```diff
@@ -0,0 +1,153 @@
+# Copyright 2019 Google Inc. All Rights Reserved.
```
markmandel: Should this be in "build" - given that this will likely be our 'here's our opinionated way to install a cluster + Agones' - should this be in a more central place? What do you all think?

aLekSer: Yes, I can move this up to a "build" folder. Thinking on how to make this method of creating cluster universal (cloud platform agnostic), but this definitely would be separate change.

build/gke-test-cluster/cluster.tf
aLekSer: I made a clone of what we have in jinja template:
https://github.com/GoogleCloudPlatform/agones/blob/7368392ed6f2ca4cb42c34f53ef6d6a0730227c3/build/gke-test-cluster/cluster.yml.jinja#L75

aLekSer: I added env variable for that, also we can use `password = "${random_string.password.result}"`:
https://medium.com/@tsadoklf/cloud-recipes-use-terraform-to-create-kubernetes-cluster-and-memorystore-redis-instance-on-gcp-d51bb53df46a

build/cluster.tf
```diff
@@ -0,0 +1,154 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+provider "google-beta" {
+  zone      = "${lookup(var.cluster, "zone")}"
+}
+
+variable "cluster" {
+  type = "map"
```
Kuqd: you could add a small description.

build/cluster.tf
```diff
@@ -0,0 +1,154 @@
...
+  default = {
+      "zone"             = "us-west1-c"
+      "name"             = "test-cluster"
+      "machineType"      = "n1-standard-4"
+      "initialNodeCount" = "4"
+      "legacyAbac"       = false
+      "project"          = "agones"
+      "password"         = ""
```
Kuqd: I would totally remove this password field and use a another variable not declared here.

You can drop terraform.tfvars file and git ignore it.

That way this is more shareable.

build/cluster.tf
```diff
@@ -0,0 +1,154 @@
...
+# by using certificate-based authentication.
+
+resource "google_compute_firewall" "default" {
+  name    = "game-server-firewall-firewall"
+  project = "${lookup(var.cluster, "project")}"
+  network = "${google_compute_network.default.name}"
+
+  allow {
+    protocol = "udp"
+    ports    = ["7000-8000"]
```
Kuqd: should this be configurable ?

build/cluster.tf
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  source_tags = ["game-server"]
+}
+
+resource "google_compute_network" "default" {
+  project = "${lookup(var.cluster, "project")}"
+  name    = "test-network"
```
Kuqd: agones network ?

build/includes/google-cloud.mk
```diff
@@ -35,6 +35,36 @@ gcloud-test-cluster: $(ensure-build-image)
...
+	rm ./gke-test-cluster/terraform.tfstate*
+
+
+gcloud-terraform-cluster: GCP_PROJECT ?= agones
+gcloud-terraform-cluster: GCP_CLUSTER_LEGACYABAC ?= false
+gcloud-terraform-cluster: GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT ?= 4
+gcloud-terraform-cluster: GCP_CLUSTER_NODEPOOL_MACHINETYPE ?= n1-standard-4
+gcloud-terraform-cluster:
+	test -n "$(GCP_PASSWORD)"  # $$GCP_PASSWORD is undefined
+	docker run --rm -it $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag) bash -c ' \
```
Kuqd:
```suggestion
	$(DOCKER_RUN) bash -c ' \
```

build/includes/google-cloud.mk
```diff
@@ -35,6 +35,36 @@ gcloud-test-cluster: $(ensure-build-image)
...
+	cd $(mount_path)/build  && terraform plan -var "cluster={zone=\"$(GCP_CLUSTER_ZONE)\"}" \
+	-var "cluster={machineType=\"$(GCP_CLUSTER_NODEPOOL_MACHINETYPE)\"}" -var "cluster={name=\"$(GCP_CLUSTER_NAME)\"}" \
+	-var "cluster={project=\"$(GCP_PROJECT)\"}" \
+	-var "cluster={password=\"$(GCP_PASSWORD)\"}" \
+	-var "cluster={initialNodeCount=\"$(GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT)\"}" -var "cluster={legacyABAC=\"$(GCP_CLUSTER_LEGACYABAC)\"}" -out tfplan'
+	docker run --rm -it $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag) bash -c ' \
+	cd $(mount_path)/build && terraform apply -input=false tfplan'
```
Kuqd: Curious - tfplan contains all previous variable defined ?

Are we ok having that file with our password left behind ? if you can mix tfplan file and env var I'd like the password to be pass after at the apply time so nothing is left behind.

build/includes/google-cloud.mk
```diff
@@ -35,6 +35,36 @@ gcloud-test-cluster: $(ensure-build-image)
...
+	$(MAKE) gcloud-auth-cluster
+	$(MAKE) setup-test-cluster
+
+gcloud-terraform-destroy-cluster:
+	docker run --rm -it $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag) bash -c ' \
```
Kuqd:
```suggestion
	$(DOCKER_RUN) bash -c ' \
```

build/includes/google-cloud.mk
aLekSer: I am currently investigating what we can do, all variables are stored inside tfplan so on the apply step we can omit them.

build/cluster.tf
aLekSer: I added port var for this

build/includes/google-cloud.mk
aLekSer: added -auto-approve

aLekSer:
Reworked password setting.
Note that I want to add helm support (3. Installs the right version of Agones) in a separate PR.

markmandel:
Question: Should this replace `gcloud-test-cluster` entirely? (Basically delete the deployment manage sccripts?) Or do both have value?

Just considering keeping both in sync is going to be painful.

aLekSer:
> Should this replace `gcloud-test-cluster` entirely? (Basically delete the deployment manage sccripts?) Or do both have value?

I think both should be used.
I cannot prove now that them are equal, also `gcloud-test-cluster` much more stable, faced the issue that I could not use cluster right after creating using `make gcloud-terraform-cluster`. So terraform could has its own limitations/issues.

aLekSer:
Rebased latest master to this branch. Currently not sure on CI fail reason.

aLekSer:
New version contains a one resource for `google-container-cluster` with 3 node pools nested in it.
Test which is failing two times in this PR is:
```
--- FAIL: TestFleetRecreateGameServerOnPodDeletion (73.82s)
fleet_test.go:844: 
Error Trace:	fleet_test.go:844
```

aLekSer:
@bbf I made password basic authentication optional. Reflected in `README.md`. 
If user omit password or use `""` empty kubernetes would disable basic auth.

build/README.md
```diff
@@ -379,6 +379,13 @@ The Kubernetes config file used to access the cluster. Defaults to `~/.kube/conf
 ### CLUSTER_NAME
 The (gcloud) test cluster that is being worked against. Defaults to `test-cluster`
 
+### GCP_PROJECT
```
markmandel: Does this default to the gcp project that is configured in `make gcloud-init`?

build/README.md
```diff
@@ -542,6 +549,27 @@ Pulls down authentication information for kubectl against a cluster, name can be
...
+
+Utilities for deploying a Kubernetes Engine cluster on Google Cloud Platform using `google` Terraform provider.
+
+#### `make terraform-init`
+Install `google` and `google-beta` terraform providers and authorize.
+
+#### `make gcloud-terraform-cluster`
+Run next command with your project ID specified:
+```
+GCP_PROJECT=<YOUR_PROJECT_ID> [GKE_PASSWORD="<YOUR_PASSWORD>"] make gcloud-terraform-cluster
```
markmandel: There is a default project in `gcloud` - I take it that these two things are not linked?

aLekSer: They are linked, but I think here is better to have an option to explicitly choose, which project you are going to use.

aLekSer: Finally removed the GCP_PROJECT as prerequisites , I am setting it using `gcloud` command if not set.

PR #682, opened 2019-03-31
Implement GameServerAlocation as API Extension 

markmandel:
~~Dependent on #659 which will be merged after [0.9.0](https://github.com/GoogleCloudPlatform/agones/milestone/9) is released~~

This moves the implementation of GameServerAllocation (GSA) to a
[Kubernetes API Extension](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) instead of using CRDs. This was essentially done for performance reasons, but to break it down:

1. GSA is now create only. Since we had no need for GSA storage, and don't want the performance hit.
1. This removes the mutation and validation webhooks, which have 30s timeout and are run in serial
    1. API Server still does cut off a response after 60s, but the api can continue processing (60s gives us enough time, I think for a SDK.Ack() on Allocate, which I don't think we had before)
    1. Validation now happens in the request.
1. We can now do batching of requests for higher throughput (#536), since we control the entire http request.

The breaking changes are:
1. GameServerAllocation's group is now `allocation.agones.dev` rather than `stable.agones.dev`,
because a CRD group can't overlap with a api server.
1. Since there is only the `create` verb for GSA, there is no get/list/watch options for GameServerAllocations - so no informers/listers either. But this could be added at a later date, if needed.


pkg/gameserverallocations/controller.go
```diff
@@ -254,119 +243,140 @@ func (c *Controller) loggerForGameServerAllocation(gsa *v1alpha1.GameServerAlloc
...
 
-// creationMutationHandler will intercept when a GameServerAllocation is created, and allocate it a GameServer
-// assuming that one is available. If not, it will reject the AdmissionReview.
-func (c *Controller) creationMutationHandler(review admv1beta1.AdmissionReview) (admv1beta1.AdmissionReview, error) {
-	c.baseLogger.WithField("review", review).Info("creationMutationHandler")
-	obj := review.Request.Object
-	gsa := &v1alpha1.GameServerAllocation{}
+// allocationHandler CRDHandler for allocating a gameserver. Only accepts POST
+// commands
+func (c *Controller) allocationHandler(w http.ResponseWriter, r *http.Request, namespace string) error {
```
markmandel: /cc @pooneh-m @ilkercelikyilmaz for your review and testing (also, I don't think I broke anything you did @ilkercelikyilmaz , but would love for you to give it a look over to make sure) :+1:

cmd/controller/main.go
```diff
@@ -196,8 +199,8 @@ func main() {
 	fleetController := fleets.NewController(wh, health, kubeClient, extClient, agonesClient, agonesInformerFactory)
 	faController := fleetallocation.NewController(wh, allocationMutex,
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
-	gasController := gameserverallocations.NewController(wh, health, gsCounter, topNGSForAllocation,
-		kubeClient, extClient, agonesClient, agonesInformerFactory)
+	gasController := gameserverallocations.NewController(api, health, gsCounter, topNGSForAllocation,
```
ilkercelikyilmaz: This is OK for now but at some point we should run the allocation controller and api server in it's own Pod(s).

markmandel: Yeah totally agreed - but we can do that as a separate PR :+1:

pkg/gameserverallocations/controller.go
ilkercelikyilmaz: LGTM

install/helm/agones/templates/extensions.yaml
```diff
@@ -16,12 +16,44 @@
 {{- $altName1 := printf "agones-controller-service.%s"  .Release.Namespace }}
 {{- $altName2 := printf "agones-controller-service.%s.svc" .Release.Namespace }}
 {{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca }}
+---
+{{- if .Values.agones.registerApiService }}
```
jkowalski: is there a reason you don't want that?

also perhaps "registerAllocationApiService" ?

pkg/client/clientset/versioned/typed/allocation/v1alpha1/fake/fake_gameserverallocation.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
jkowalski: nit: 2019 everywhere

examples/simple-udp/gameserverallocation.yaml
```diff
@@ -12,14 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+#
+# A GameServerAllocation against a Fleet named 'simple-udp'
 # (GameServerAllocation is currently experimental, and likely to change in upcoming releases)
+#
 
-apiVersion: "stable.agones.dev/v1alpha1"
+apiVersion: "allocation.agones.dev/v1alpha1"
```
jkowalski: general question - is allocation the only thing that will be in this API going forward? Perhaps we need a broader name to indicate "services that external consumers are going to call".

cmd/controller/main.go
jkowalski: +1 to its own set of pods with higher number of replicas, so that this service is never disrupted when we upgrade Agones itself. This is going to be critical for production use.

install/helm/agones/templates/extensions.yaml
markmandel: GCP Marketplace usually requires a different install step, so there are switches to turn various pieces off if people need to break the glass and copy things into their own install process.

@bbf to confirm.

pkg/client/clientset/versioned/typed/allocation/v1alpha1/fake/fake_gameserverallocation.go
markmandel: Good catch!

examples/simple-udp/gameserverallocation.yaml
markmandel: That's a really interesting question. Groups are pretty broad, and we could definitely expand this - and anything that is "external" would likely need to be managed by multiple pods to ensure HA (as discussed above). Which would give us an internal delineation - anything under this group needs to be managed by multiple pods, anything in the `stable` group, is managed by the controller (which self heals).

Maybe `external.agones.dev` ?

It also begets a question of is "stable" the right name for the original group. I'm starting to wonder if `core.agones.dev` would have been better? Or something else?

Definitely open to suggestions here.

pkg/client/clientset/versioned/typed/allocation/v1alpha1/fake/fake_gameserverallocation.go
markmandel: Done!

markmandel:
This should be good for review now :+1:

site/content/en/docs/Reference/agones_crd_api_reference.html
```diff
@@ -2219,6 +2219,10 @@ <h3 id="WebhookPolicy">WebhookPolicy
 
 
 
+
+
+
```
markmandel: @aLekSer I can't work out why I'm not getting the allocation group for the `GameServerAllocation` on the docs? Do you have any ideas?

examples/simple-udp/gameserverallocation.yaml
markmandel: Looking at #703 - I'm feeling like `allocation` is actually the right choice here for the group name.

site/content/en/docs/Reference/agones_crd_api_reference.html
aLekSer: @markmandel, working on that

aLekSer: Fixed the issue using `go mod edit --replace=agones.dev/agones@latest=../../../agones.dev/agones/`

Resulted webpage could be found here:
http://35.237.42.204:1313/docs/reference/agones_crd_api_reference/

install/helm/agones/templates/extensions.yaml
markmandel: Regarding the rename comment - we have `registerWebhooks` for all our webhooks - I don't think we need to be that granular. If there ever end up being more APIServices, they could also go here.

markmandel:
@Kuqd @jkowalski any blockers from approving and merging this? I know it's slowing down @pooneh-m not having this merged?

Just updated it to master (and assuming it passes) I'm hoping this is good to go? :smile:

ilkercelikyilmaz:
I am running the load test on this change and so far it is looking good. I don't see any reason not to merge this change.
P.S. I am working on few more improvements on top of this change so it will be great to have this change merged.

PR #694, opened 2019-04-07
Move GameServer to Unheathy when Pod Deleted

### LINKED ISSUE #678: Agones controller does not remove deleted pod from game server list
I had a fleet of 2 game servers running, one of the game server was allocated, I deleted allocated server by running `kubectl delete pod pod-name-xyz` this deleted the pod but when retrieving fleet information it still showed that fleet had two game servers running and kept returning allocated server IP and PORT to which we couldn't connect because game server pod has been deleted. Is there any reason why agones does not detect deleted pods?

markmandel:
There was only implementation of GameServer's being moved to Unhealthy on very specific Pod events (container crash, being unschedulable), but never on Pod removal as a whole.

Also it was locked down to specific states, which made it very fragile.

Reworked this so that any of the above now triggers a GameServer being moved to an Unhealthy state.

Closes #678

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -499,7 +499,7 @@ func (c *Controller) syncDevelopmentGameServer(gs *v1alpha1.GameServer) (*v1alph
 	}
 
 	gsCopy := gs.DeepCopy()
-	ports := []v1alpha1.GameServerStatusPort{}
+	var ports []v1alpha1.GameServerStatusPort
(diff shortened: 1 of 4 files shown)


markmandel:
I am going to run this a bunch of times, to make sure it's actually working.

markmandel:
One more. If it passes again, I'm going to merge it.

markmandel:
That was the site deploy that failed. That's a new one, but a red herring.

PR #696, opened 2019-04-09
Helm support for Terraform

aLekSer:
Deploying agones using `agones.dev` chart. Version is specified in `helm.tf` config.

Two options of installation:
- ability to install current version of sourcecode like `make install` does.
- install release version from helm repo.
This Pull Request adds an ability to install Agones along with creating the cluster using Terraform.

Contains part of #670 Pull Request, so leave it as draft until rebased with new master.

For #657.

### DIFF
--- a/build/includes/google-cloud.mk
+++ b/build/includes/google-cloud.mk
@@ -33,7 +33,6 @@ gcloud-test-cluster: $(ensure-build-image)
 		--properties cluster.zone:$(GCP_CLUSTER_ZONE),cluster.name:$(GCP_CLUSTER_NAME),cluster.nodePool.initialNodeCount:$(GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT),cluster.nodePool.machineType:$(GCP_CLUSTER_NODEPOOL_MACHINETYPE),cluster.legacyAbac:$(GCP_CLUSTER_LEGACYABAC)\
 		--template=$(mount_path)/build/gke-test-cluster/cluster.yml.jinja
 	$(MAKE) gcloud-auth-cluster
-	$(MAKE) setup-test-cluster
 
 clean-gcloud-test-cluster: $(ensure-build-image)
 	docker run --rm -it $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag) gcloud \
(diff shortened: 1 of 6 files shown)


build/includes/google-cloud.mk
```diff
@@ -48,23 +47,52 @@ terraform-clean:
 	rm -r ./.terraform
 	rm ./terraform.tfstate*
 
-
+# Creates a cluster and install release version of Agones controller
+# Version could be specified by AGONES_VERSION
 gcloud-terraform-cluster: GCP_CLUSTER_LEGACYABAC ?= false
 gcloud-terraform-cluster: GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT ?= 4
 gcloud-terraform-cluster: GCP_CLUSTER_NODEPOOL_MACHINETYPE ?= n1-standard-4
+gcloud-terraform-cluster: AGONES_VERSION ?= 0.9.0
```
markmandel: I can't think of a better param than this - I just see it going out of date as releases continue forward.

Can we default to '' which will install the latest chart, and then have an optional override?

build/includes/google-cloud.mk
```diff
@@ -48,23 +47,52 @@ terraform-clean:
...
 	$(MAKE) gcloud-auth-cluster
-	$(MAKE) setup-test-cluster
+
+# Creates a cluster and install current version of Agones controller
+# Set all necessary variables as `make install` does
+gcloud-terraform-install: GCP_CLUSTER_LEGACYABAC ?= false
+gcloud-terraform-install: GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT ?= 4
+gcloud-terraform-install: GCP_CLUSTER_NODEPOOL_MACHINETYPE ?= n1-standard-4
+gcloud-terraform-install: ALWAYS_PULL_SIDECAR := true
+gcloud-terraform-install: IMAGE_PULL_POLICY := "Always"
```
markmandel: Given that we're installing a production version of Agones - probably don't need to be always pulling images, wdyt?

examples/terraform-submodule/module.tf
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+
+// Run:
+//  terraform apply -var project="<YOUR_GCP_ProjectID>" -var agones_version="0.9.0"
+variable "password" {
+   default= ""
+}
+variable "agones_version" {
+  default="0.9.0"
```
markmandel: I'm just hesitant to set the version as default. I already have a long list of things to update on release - this just adds another one. Can we use the helm default of installing the latest, with the ability to override?

site/content/en/docs/Installation/terraform.md
```diff
@@ -0,0 +1,65 @@
+---
+title: "Install Agones using Terraform"
```
markmandel: Do we want to be clear this is GKE only somehow?

site/content/en/docs/Installation/terraform.md
```diff
@@ -0,0 +1,65 @@
...
+
+- Terraform v0.11.13
+- [Helm](https://docs.helm.sh/helm/) package manager 2.10.0+
+- Access to Google Cloud Kubernetes Engine
+- `gcloud` utility installed
+- Git
+
+## Installing the Agones as Terraform submodule
+
+You can use Terraform to provision your GKE cluster and install agones on it using Helm Terraform provider.
```
markmandel: Can we add a quick summary of what this does? I.e. creates  a cluster with n nodes, sets up rbac rules, install Agones, etc.

Also, do we need a list of configurable params?

aLekSer: Yes, I think it is neccessary to add those params.

build/includes/google-cloud.mk
aLekSer: This make target will install not production, but development version as oppose to `make gcloud-terraform-cluster` (install production versions)

aLekSer:
Hello @markmandel ,
I have applied all comments above, please check.
Basically I removed `"agones_version"` default value.
`"agones.image.tag"` becomes optional.
So we have two separate make targets:
1. `make gcloud-terraform-cluster` will install production version.
1. `make gcloud-terraform-install` will install current development version which is equivalent of running `make gcloud-test-cluster` and `make install` at the same time.

PR #704, opened 2019-04-13
Change  the website theme  and add Ubisoft logo

Kuqd:
- Changes website theme to be more in line with Agones.
- Adds Ubisoft logo.
- Adds asciinema demo of a fleet.


site/content/en/_index.html
```diff
@@ -4,22 +4,22 @@
 
 +++
 
-{{< blocks/cover title="Host, run and scale dedicated game servers on Kubernetes" image_anchor="top" height="full" color="orange" >}}
+{{< blocks/cover title="" image_anchor="top" height="full" color="white" >}}
```
markmandel: I'm not sure how, but was thinking to remove the `td-cover-block--height-full` css class from this section in the redered html looks nicer (so it doesn't take up the full page)

thisisnotapril:
I love it!! This looks awesome @Kuqd. Same comment re: white space that @markmandel had.

site/content/en/_index.html
Kuqd: changed it td-cover-block--height-min and it much better !

markmandel:
LGTM too. Cyril, looks like we have some CLA issues (also not sure why this keeps failing to build - I think hugo is being a pain)

Also - this made me realise the app engine cleanup script were cleaning from last->first deployed, not first->last.. which is why previews only lasted ~12 hours :confused:

PR #709, opened 2019-04-16
Add proto directory and update tooling.

heartrobotninja:
Fixes one part of #695 and #630 where proto files are no longer in the vendor directory.


markmandel:
Should the proto files also be included in this PR?

heartrobotninja:
Ugh, I thought they were on this. Added and squashed.

PR #710, opened 2019-04-16
We no longer need gen-grpc-cpp.sh

markmandel:
So deleting it!


dsazonoff:
@markmandel , we should keep this file until we will have custom gRPC version support in cmake (PR for that will be ready at 22-23 April).
Currently in case of updating gRPC version we need to regenerate and commit new files to repo. This file is not used in a build because it is not a part of a build. It's just a helper with necessary arguments for protobuf and with scripts to insert license text.

markmandel:
But isn't this covered by this? https://github.com/GoogleCloudPlatform/agones/blob/master/build/build-sdk-images/cpp/gen.sh ?

dsazonoff:
Sorry, it is. I'm wrong. I thought that this file was renamed to gen.sh by Cyril Tovena, when he moved out sdk build from main makefiles.

PR #712, opened 2019-04-16
Mount go mod cache

markmandel:
To speed up things when the mod cache is required.


markmandel:
@heartrobotninja do you see any issues with this?

heartrobotninja:
We should keep an eye on future Go releases which may change the functionality of the cache. LGTM.

PR #715, opened 2019-04-16
Abstract build image ensuring and building

markmandel:
This is setup work, to enable SDK build images to also take advantage of the same scaffolding we use to only rebuild build images once they have changed.

Functionality is exactly the same!


markmandel:
@Kuqd I think this will come in handy for the SDK build images. Let me know if you think there are further improvements I can do to make things better :+1:

Issue #660, opened 2019-03-17
1st Party / Distributed Matchmaker support


markmandel:
# Objective

To be able to support matchmaking systems that require a gameserver to register themselves with the matchmaker -- such as first party matchmakers.

# Background

Many matchmakers that console and others companies provide as hosted services, have a matchmaker workflow in which:

1. A gameserver will register itself with the matchmaker when ready (usually for a specific period of time), with their IP and port(s) for gameplay
2. The matchmaker will choose the gameserver out of the pool that have registered themselves
3. The matchmaker will communicate to the game server that the matchmaker has chosen it for gameplay
4. Players will then play on the game server.

This would be an alternative flow to `FleetAllocation` or `GameServerAllocation`, which will remain the preferred method of `GameServer` allocation, so that Agones can retain fine control of scheduling within the cluster, but since this is quite a prevalent workflow, Agones should also support it, with appropriate documentation on the tradeoffs.

# Requirements and Scale
The design and implementation must ideally have no potential race conditions, and actively prevent the user from incurring race conditions in their usage as well.

# Design Ideas

To support this within Agones, we will need to add three enhancements:

1. A new `Reserved` `GameServer` state
2. A new SDK function `Reserve(seconds)`
3. A new SDK function `Allocate()`

## Reserved GameServer State

Reserved state is to signify that the GameServer cannot be deleted, as it may move to allocated in a given time frame. Therefore:

1. When scaling down a Fleet, `Reserved` instances will not be deleted
2. When autoscaling a fleet, they will be counted towards the current buffer, and therefore a change of state to `Reserved` will not incur an increase in GameServers in the Fleet.

This will mean that if a GameSerer is not demarcated for a game session by the matchmaker, it can move back to Ready in a timely manner, and is able to be scaled down as needed.

## SDK Function: Reserve(seconds)

This new SDK function, `Reserve` will set the `GameServer` record to the `Reserved` state for the given number of seconds. (0 indicating forever). When that time period has ended, the `GameServer` shall revert back to `Ready`. 

It is assumed that when working with a matchmaker, the developer will mark the GameServer as `Reserved` for slightly longer than it is registered with the matchmaker, so as to avoid scale down race conditions.

### Technical Details

Rather than implementing this with a queue, this should be a synchronous call to the Kubernetes API, with in-built retry and a timeout (30s) on failure. Otherwise there is potential for race conditions between calling the SDK function, and the `GameServer` being moved to `Reserved` state.

[...]

PR #721, opened 2019-04-17
Implementation of SDK.Allocate()

markmandel:
Now GameServers can self Allocate!

This is just the implementation of the GO SDK at this stage (although the gRPC libraries have been regenerated). Other languages can come in later PRs.

This is the first part of 1st Party MatchMaking support (#660)


pkg/sdkserver/sdkserver.go
```diff
@@ -330,6 +331,46 @@ func (s *SDKServer) Ready(ctx context.Context, e *sdk.Empty) (*sdk.Empty, error)
 	return e, nil
 }
 
+// Allocate set the GameServer to Allocate, as longs as it's not in UnHealthy,
+// Shutdown or has a DeletionTimeStamp(). Times out after 30 seconds if it cannot
+// complete the operation due to contention issues.
+func (s *SDKServer) Allocate(context.Context, *sdk.Empty) (*sdk.Empty, error) {
+	s.logger.Info("Received self Allocate request")
+	err := wait.PollImmediate(500*time.Millisecond, 30*time.Second, func() (done bool, err error) {
```
jkowalski: wait.ExponentialBackoff() is more appropriate to handle contention issues

having said that, I don't see why this would ever be contending - the GS is running so unlikely to be Unhealthy and not in Shutdown and not deleted

pkg/sdkserver/sdkserver.go
```diff
@@ -330,6 +331,46 @@ func (s *SDKServer) Ready(ctx context.Context, e *sdk.Empty) (*sdk.Empty, error)
...
+			return true, errors.New("cannot Allocate a Shutdown GameServer")
+		}
+
+		gsCopy := gs.DeepCopy()
+		gsCopy.Status.State = stablev1alpha1.GameServerStateAllocated
+		_, err = s.gameServerGetter.GameServers(s.namespace).Update(gsCopy)
+
+		// try again, with some jitter
+		if k8serrors.IsConflict(err) {
+			jitter := wait.Jitter(500*time.Millisecond, 1)
```
jkowalski: exponential back-off will do the sleep for you

pkg/sdkserver/sdkserver.go
markmandel: YES! You are right. 

I was worried about the Exponential backoff potentially going for a really long time, but I can provide a [Backoff](https://godoc.org/k8s.io/apimachinery/pkg/util/wait#Backoff) with a `Cap`, and I get the same thing. I will make this change. Nice catch - I was looking for this, but didn't find it previously :+1:

PR #716, opened 2019-04-17
Clang-formatted the C++ SDK files.

devjgm:
This addresses part of  #713. This seemed fairly uncontroversial. This PR makes zero semantic changes. It's only a whitespace/formatting change applied by `clang-format -style=Google ...`. 

While I expect no tests to fail from this change, I ran all the tests with `cd build; make test` as the [instructions](https://github.com/GoogleCloudPlatform/agones/tree/master/build#testing-and-building) suggest and they did pass. However, I also introduced an intentional syntax error in the C++ code, and the tests still passed, so I'm not convinced that they're running. Is there some other way that I should/could run the C++ build/test on my local machine before sending a PR?

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/agones/716)

PR #724, opened 2019-04-18
Adds a .clang-format file making Google style the default

devjgm:
This change should probably have been part of PR #716, but I forgot.

`.clang-format` is a standard file that the `clang-format` tool(s) look for. This file will tell all clang-format invocations to by-default use the "Google" style. This means that someone running `clang-format -i foo.cc` within the `sdks/cpp/...` directory hierarchy will automatically use the correct style.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/agones/724)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 !.gcloudignore
 !.eslintrc.*
 !.nycrc
+!.clang-format
 *.iml
 bin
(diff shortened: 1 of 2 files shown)


devjgm:
fyi, @markmandel

PR #730, opened 2019-04-19
Add status subresource for FleetAutoscaler

aLekSer:
Add E2E test of FleetAutoscaler status update.
Update serviceaccounts permissions accordingly. 

For #329 .

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -124,6 +124,7 @@ func (f *Framework) WaitForGameServerState(gs *v1alpha1.GameServer, state v1alph
 }
 
 // WaitForFleetCondition waits for the Fleet to be in a specific condition or fails the test if the condition can't be met in 5 minutes.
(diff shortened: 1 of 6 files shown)


test/e2e/framework/framework.go
```diff
@@ -124,6 +124,7 @@ func (f *Framework) WaitForGameServerState(gs *v1alpha1.GameServer, state v1alph
 }
 
 // WaitForFleetCondition waits for the Fleet to be in a specific condition or fails the test if the condition can't be met in 5 minutes.
+// nolint: dupl
```
markmandel: Curious - why is there `nolint: dupl` here?

aLekSer: We get two methods which are similar WaitForFleetCondition and WaitForFleetAutoscalerCondition. So `make tests` fails without it. You can see the output in the BuildFailed above.

PR #737, opened 2019-04-21
Intial tool vendoring commit.

heartrobotninja:
Gets most tools into tools.go, but only sets vendoring for gen script in go
sdk build directory.

Part 1 for #695


tools.go
```diff
@@ -0,0 +1,21 @@
+// +build tools
+
+package agones
+
+/*
```
markmandel: We shouldn't leave this lying around :smile:

go.mod
```diff
@@ -7,8 +7,8 @@ require (
 	contrib.go.opencensus.io/exporter/stackdriver v0.8.0
 	fortio.org/fortio v1.3.1
 	github.com/BurntSushi/toml v0.3.1 // indirect
+	github.com/ahmetb/gen-crd-api-reference-docs v0.1.1
```
markmandel: Just to clarify - these are now in here because we imported them in tools.go?

go.mod
```diff
@@ -47,6 +48,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.0.0-20181126195113-57b8dbfcc51a
 	k8s.io/apimachinery v0.0.0-20181126123124-70adfbae261e
 	k8s.io/client-go v8.0.0+incompatible
+	k8s.io/code-generator v0.0.0-20190419212335-ff26e7842f9d
```
markmandel: Should code generator be tied to the same version as: https://github.com/GoogleCloudPlatform/agones/blob/master/build/build-image/Dockerfile#L68 ?

(Should it be in the Dockerfile at all now?)

go.mod
heartrobotninja: Correct.

tools.go
heartrobotninja: The intention for documentation. That way when we add or remove tools, we can easily know where they are used.

go.mod
heartrobotninja: So this got really weird. Its not used right now since we still have the download of the pinned version in the dockerfile. There is an added issue which doesn't make sense: If I add a replace statement, it causes the vendoring to not work and it deletes the code out of vendor. The immediate option would be to download the code generator to the vendor_fixes directory and replace it that way. Do you have any thoughts on that?

markmandel: Potentially silly question, why not pin in this file to the same version that we have in the Docker container?

heartrobotninja: Because it deletes itself out of the vendor file because when I pin it in the go.mod file it says that its unused and deletes it out of vendor, but if I don't pin it and let it use that version I keeps it just fine. I am mainly using this as a place holder until I get that figured out since the module version is only imported and in vendor. The version in the docker file is still the same that was used before.

markmandel: How odd!

I assume there is value in keeping this then, since it doesn't line up with the Dockerfile version?

Just wondering if we should drop the import for now if its behaiving badly?

heartrobotninja: Removed it until we can figure out why this is borked.

PR #738, opened 2019-04-23
CI Speedup: Cache Build SDK between builds

markmandel:
This tags the base sdk image with the SHA of it's own Dockerfile, such that it will be rebuilt on Dockerfile change.

This also pushes the base SDK image up to GCR between builds, so that it doesn't get rebuilt every time.


markmandel:
After building the initial image - this drops build times to ~15 min (from 20-25min) :+1:

PR #739, opened 2019-04-24
We don't need CMake in the base build image

markmandel:


markmandel:
@dsazonoff see any issue with removing this?

dsazonoff:
> @dsazonoff see any issue with removing this?

Everything looks good.

PR #740, opened 2019-04-24
Updated the filtering condition on GameServerShutdown to include the undeleted Unhealthy GSs

ilkercelikyilmaz:
Fixed the filtering condition on GameServerShutdown to include the undeleted Unhealthy GSs.
After the change the Unhealthy servers are (eventually) being deleted(see the graph below).

Also remove the WaitForCacheSync call during the GS creation. This improves the allocation and overall performance.

image


ilkercelikyilmaz:
> Oops, sorry - can we squish this to one commit plz?

Done

PR #742, opened 2019-04-24
Switch to parrallel execution of SDK commands

aLekSer:
Use subshell execution of the `run-all-sdk-command` make target.


aLekSer:
I have added wait right after the loop, if one process fail we would get the error and a message.
Tested with `make test-sdks` it spawns two docker containers for go and node. 

And tested errors scenario for instance if we add `exit 1` to `test.sh` script. That would result in:
```
17 specs, 0 failures
Finished in 0.087 seconds
Failed to run SDK command
make: *** [run-all-sdk-command] Error 255
```

build/includes/sdk.mk
```diff
@@ -55,17 +55,22 @@ gen-all-sdk-grpc: run-all-sdk-command
 gen-sdk-grpc: COMMAND := gen
 gen-sdk-grpc: run-sdk-command
 
-# Runs a command on all supported languages, use COMMAND variable to select which command.
+# Runs a command on all supported languages in parallel, use COMMAND variable to select which command.
```
jkowalski: how about using native parallelism built into make:

```
run-all-sdk-command: run-sdk-command-go run-sdk-command-rust run-sdk-command-cpp

run-sdk-command-cpp:
  $(MAKE) run-sdk-command SDK_FOLDER=cpp

run-sdk-command-rust:
  $(MAKE) run-sdk-command SDK_FOLDER=rust

run-sdk-command-go:
  $(MAKE) run-sdk-command SDK_FOLDER=go
```

Then the user can do `make -j4 run-all-sdk-command COMMAND=gen`

markmandel: ooh that's a really interesting idea. I wonder if we could also come up with a pattern rule so we don't have to define each rule individually?

aLekSer: That native way looks more natural and simple, updating this Pull Request, removing initial and next waiting for loop:
```
	for job in `jobs -p` ; \
	do \
		wait $$job || let "FAIL+=1"; \
	done ; \
```

aLekSer: @jkowalski
The only thing that we don't have build.sh in each directory, that's why we have such check that file exists:
`if [ "$${d%?}" != "tool" ] &&  [ -f $$d/$(COMMAND).sh ];`

aLekSer:
Updated cloudbuild-yaml also. Now if I run:
`make -j 5 test`
I would see 5 parallel running images as expected (`docker ps --format "table {{.Image}}\t{{.Command}}"`):
```
IMAGE                              COMMAND                                                                                                                  
agones-build-sdk-go:c168fcc53f     "/root/entrypoint.sh…"                                                                                             
agones-build-sdk-node:c168fcc53f   "/root/entrypoint.sh…"                                                                                             
agones-build:c168fcc53f            "bash -c /go/src/ago…"                                                                                             
agones-build:c168fcc53f            "go test -mod=vendor…"                                                                                             
agones-build:c168fcc53f            "bash -c /go/src/ago…"
```

PR #746, opened 2019-04-25
Convert C++ Example to Docker Build Pattern

markmandel:
While this does make it a longer build time, it makes it easier to get started with the C++ example because there is no tooling dependencies other than Docker (Make is even optional, it just sets the default arguments up).

Also provides an end to end example of compiling the SDK, installing it and then using it with another C++ project.

Co-authored-by: @dsazonoff


markmandel:
/cc @dsazonoff Please confirm you are good with these changes being applied :smile: 

Also do you see any issues here?

dsazonoff:
Looks good.

PR #747, opened 2019-04-26
Cleanup - no longer need to list Pods for GameServers

markmandel:
Since we changes Pod names for GameServers to be the same name as the GameServer, we no longer need to list Pods when syncing's GameServers since we can look it up directly.


ilkercelikyilmaz:
> /cc @ilkercelikyilmaz curious is this has any performance improvements at large scale.

Yes it was. Especially for allocation

ilkercelikyilmaz:
I am not sure if this particular change or other change, but the allocation-shutdown-create-allocation cycle under load seems to change. It takes longer for Agones to clean-up shutdown servers. See the graph below:

image

PR #750, opened 2019-04-30
Deprecate Fleet Allocation.

markmandel:


pooneh-m:
Since FleetAllocation is in alpha version, according to [this](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning), you don't need to deprecate it, but just remove it.
> The API may change in incompatible ways in a later software release without notice.

markmandel:
It's true - but I like giving people some warning, as I know people are using FleetAllocation in production right now :smile:

PR #753, opened 2019-04-30
Implement multicluster allocation for local cluster allocation.

pooneh-m:
- Introduce allocation service API for enabling multi-cluster allocations.
- Implementing the logic to choose between local or remove allocation to be applied. (remote allocation is todo and will be done in the next PR)
- Enable the RBAC policies for allocation policies.

### DIFF
--- a/pkg/gameserverallocations/controller.go
+++ b/pkg/gameserverallocations/controller.go
@@ -25,11 +25,13 @@ import (
 
 	"agones.dev/agones/pkg/apis"
 	"agones.dev/agones/pkg/apis/allocation/v1alpha1"
+	multiclusterv1alpha1 "agones.dev/agones/pkg/apis/multicluster/v1alpha1"
 	"agones.dev/agones/pkg/apis/stable"
 	stablev1alpha1 "agones.dev/agones/pkg/apis/stable/v1alpha1"
 	"agones.dev/agones/pkg/client/clientset/versioned"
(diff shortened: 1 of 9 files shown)


pkg/gameserverallocations/controller.go
```diff
@@ -69,13 +71,14 @@ type Controller struct {
...
-	gameServerGetter    getterv1alpha1.GameServersGetter
-	gameServerLister    listerv1alpha1.GameServerLister
-	stop                <-chan struct{}
-	workerqueue         *workerqueue.WorkerQueue
-	recorder            record.EventRecorder
+	topNGameServerCount    int
+	gameServerSynced       cache.InformerSynced
+	gameServerGetter       getterv1alpha1.GameServersGetter
+	gameServerLister       listerv1alpha1.GameServerLister
+	allocationPolicyGetter multiclustergetterv1alpha1.GameServerAllocationPoliciesGetter
```
markmandel: I'm  wondering if it's better to use a Lister so we can take advantage of the its backing cache?

Rather than hit the k8s api every time for configuration I expect won't change very often?

pooneh-m: Thanks Mark! Exactly, the policies will not change very often and backing cache is more performant.

pooneh-m: Done

PR #756, opened 2019-05-02
Add AKS, GKE and Helm terraform modules

aLekSer:
Add configuration for deploying cluster and installing Agones on AKS.

Next step would be to move common modules into `/build/modules/` from `/build` and change `terraform.mk` make targets to support this. As you can see there is a small difference between `helm.tf` for GKE and AKS. So need to create common output to combine them into one helm.tf file.

Tested that this configuration works on Azure free account. AWS does not support EKS on free tier.

For #657 .


build/modules/aks/aks.tf
```diff
@@ -0,0 +1,115 @@
+# Copyright 2019 Google LLC All Rights Reserved.
```
roberthbailey: Have you thought about breaking this into multiple files to separate inputs / resources / outputs as per https://www.terraform-best-practices.com/code-structure#getting-started-with-structuring-of-terraform-configurations?

aLekSer:
I will break down tf config files and update this PR. Also I want to create prod and dev configurations, `dev` for current changes, and `prod` for stable releases.

aLekSer:
Add more natural variables and outputs which make me able to have one helm terraform configuration used by both GKE and AKS.
Also note that I added gke-local example which is using local module files. https://github.com/GoogleCloudPlatform/agones/blob/9259d65a9accb6a63a775e0a3b099585929a4eba/examples/terraform-submodules/gke-local/main.tf

build/modules/gke/cluster.tf
```diff
@@ -0,0 +1,155 @@
...
+  project  = "${lookup(var.cluster, "project")}"
+  provider = "google-beta"
+
+  # Setting an empty username and password explicitly disables basic auth
+  master_auth {
+    username = "${local.username}"
+    password = "${var.password}"
+  }
+  enable_legacy_abac = "${lookup(var.cluster, "legacyAbac")}"
+  node_pool = [
```
roberthbailey: After reading the docs about how terraform handles updates to node pools, I think it would be better to make each node pool a separate resource.

From https://www.terraform.io/docs/providers/google/r/container_cluster.html#node_pool:

> **Warning**: node pools defined inside a cluster can't be changed (or added/removed) after cluster creation without deleting and recreating the entire cluster. Unless you absolutely need the ability to say "these are the *only* node pools associated with this cluster", use the [google_container_node_pool](https://www.terraform.io/docs/providers/google/r/container_node_pool.html) resource instead of this property.

examples/terraform-submodules/gke-local/main.tf
```diff
@@ -0,0 +1,67 @@
...
+      "project"          = "${var.project}"
+      "zone"             = "us-west1-c"
+      "name"             = "test-cluster"
+      "machineType"      = "n1-standard-4"
+      "initialNodeCount" = "4"
+      "legacyAbac"       = false
+  }
+}
+
+/*
```
roberthbailey: Delete these commented out lines?

site/content/en/docs/Installation/terraform.md
```diff
@@ -12,10 +12,10 @@ description: >
 - Terraform v0.11.13
 - [Helm](https://docs.helm.sh/helm/) package manager 2.10.0+
 - Access to Google Cloud Kubernetes Engine
```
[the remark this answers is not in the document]
aLekSer: Yes, that would be much better description, I will split this into two parts, one for GKE and one for AKS

build/modules/gke/cluster.tf
aLekSer: That would make cluster creating longer, about 10 minutes compared to 3 minutes now.
But regarding this warning it seems to be inevitable.

aLekSer: Note that after this change I faced issue with cluster starts upgrading after I attach 2 additional node_pools to it.
https://github.com/terraform-providers/terraform-provider-google/issues/3385

aLekSer: Fixed the issue as proposed in link I posted above:
  remove_default_node_pool = true
And now we have 3 node pools and helm would install Agones right after creation of all node pools.

roberthbailey: I agree that cluster creation is much slower. It's really annoying that the terraform implementation for GKE requires jumping through these hoops. :/

site/content/en/docs/Installation/terraform.md
```diff
@@ -12,10 +12,10 @@ description: >
...
-- `gcloud` utility installed
+- `gcloud` or `az` utility installed
```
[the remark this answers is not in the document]
aLekSer: I reworded this lines and split specific requirements for subsequent sections.

aLekSer:
@roberthbailey, thanks for the review. Applied all your comments, in a separate PR I am going to remove old not module version of tf config which is left in terraform.mk.

roberthbailey:
Thanks!

@markmandel / @jkowalski - can one of you take a quick look for approval?

PR #757, opened 2019-05-03
Implementing cross cluster allocation request

pooneh-m:
1. Get secret that stores certificates.
2. Call remote allocation server.

### DIFF
--- a/pkg/gameserverallocations/controller.go
+++ b/pkg/gameserverallocations/controller.go
@@ -15,6 +15,10 @@
 package gameserverallocations
 
 import (
+	"bytes"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
@@ -48,9 +52,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1lister "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 )
@@ -61,6 +67,12 @@ var (
 	ErrNoGameServerReady = errors.New("Could not find a Ready GameServer")
 	// ErrConflictInGameServerSelection is returned when the candidate gameserver already allocated
 	ErrConflictInGameServerSelection = errors.New("The Gameserver was already allocated")
+)
+
+const (
+	secretClientCertName = "client-cert"
+	secretClientKeyName  = "client-key"
+	secretCaCertName     = "ca-cert"
 )
 
 // Controller is a the GameServerAllocation controller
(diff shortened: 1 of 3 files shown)


pkg/gameserverallocations/controller.go
```diff
@@ -75,6 +85,7 @@ type Controller struct {
 	gameServerGetter       getterv1alpha1.GameServersGetter
 	gameServerLister       listerv1alpha1.GameServerLister
 	allocationPolicyLister multiclusterlisterv1alpha1.GameServerAllocationPolicyLister
+	secretGetter           typedcorev1.SecretsGetter
```
jkowalski: should we use a lister here? Getter will always go to the API server, while lister will use cached data

pkg/gameserverallocations/controller.go
```diff
@@ -328,8 +342,95 @@ func (c *Controller) applyMultiClusterAllocation(gsa *v1alpha1.GameServerAllocat
...
+
+	data, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+	if response.StatusCode >= 400 {
+		return nil, errors.New(string(data))
+	}
+
+	err = json.Unmarshal(data, &gsaResult)
```
jkowalski: nit:

this can help avoid having to explicitly read `data`:

```
if err := json.NewUnmarshaler(response.Body).Decode(&gsaResult); err != nil {
  return nil, fmt.Errors("unable to parse request: %v", err)
}
```

pkg/gameserverallocations/controller.go
```diff
@@ -328,8 +342,95 @@ func (c *Controller) applyMultiClusterAllocation(gsa *v1alpha1.GameServerAllocat
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &gsaResult, nil
+}
+
+// createRemoteClusterRestClient creates a rest client with proper certs to make a remote call.
+func (c *Controller) createRemoteClusterRestClient(namespace, secretName string) (*http.Client, error) {
```
jkowalski: nit: this may need some caching at some poing, since I think this has some CPU cost.

pkg/gameserverallocations/controller.go
```diff
@@ -328,8 +342,95 @@ func (c *Controller) applyMultiClusterAllocation(gsa *v1alpha1.GameServerAllocat
...
+	secret, err := c.secretGetter.Secrets(namespace).Get(secretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if secret == nil || len(secret.Data) == 0 {
+		return nil, nil, nil, fmt.Errorf("secert %s does not have data", secretName)
+	}
+
+	// Create http client using cert
+	clientCert = secret.Data[secretClientCertName]
```
jkowalski: perhaps:

```
if clientCert == nil {
  return .. fmt.Erroff("missing client certificate in secret ...")
}
```

for all 3 fields

pkg/gameserverallocations/controller.go
pooneh-m: good catch!

pkg/gameserverallocations/controller.go
pooneh-m: The problem is this would fail for error cases that does not decode to gsa. So it actually need the two step deserialization.

pkg/gameserverallocations/controller.go
pooneh-m: Great point, thanks! Added a comment to cache the client in a later iteration.

pkg/gameserverallocations/controller.go
pooneh-m: X509KeyPair returns an error if any of the key pair is not conforming with the accepted format including null. I can add an extra error message for clarification.

pkg/gameserverallocations/controller.go
jkowalski: not sure what's going on here, then

is it important to return data in error string? if so, can you add some comments?

pooneh-m: The data is returned as an error string only for http status 4xx and 5xx. In that case the response body is expected to contain error and not GSA message. Therefore Decode(&gsaResult) would fail to parse the response body as GSA.
I'll add comment in the next PR since this is already merged.

PR #758, opened 2019-05-06
Flaky: Fix test for TestGameServerUnhealthyAfterDeletingPod

markmandel:
Change in behaviour on Unhealthy GameServers that are not in Fleets (deleted vs just being marked as Unhealthy), caused this e2e test to be flaky.

Added a TODO to review if we are still happy with this, since release is coming up, and having stable tests is most important right now.

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -21,6 +21,7 @@ import (
 	"agones.dev/agones/pkg/apis/stable"
 	"agones.dev/agones/pkg/apis/stable/v1alpha1"
 	e2eframework "agones.dev/agones/test/e2e/framework"
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -192,6 +193,8 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
 		t.Fatalf("Could not get a GameServer ready: %v", err)
 	}
 
+	logrus.WithField("gsKey", readyGs.ObjectMeta.Name).Info("GameServer Ready")
+
 	gsClient := framework.AgonesClient.StableV1alpha1().GameServers(defaultNs)
 	podClient := framework.KubeClient.CoreV1().Pods(defaultNs)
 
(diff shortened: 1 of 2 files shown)


markmandel:
/cc @ilkercelikyilmaz just a heads up about this - we might want to review this behaviour at some point. Originally GameServers not in fleet that dropped into `Unhealthy` did not get deleted (sorry I didn't pick up on this in review).

But maybe this is fine, and we can leave it as is :man_shrugging: 

Just wanted to put it on your radar, in case you had thoughts. In _theory_ the gameserverset controller should have taken care of this for the issue you were facing, but for some strange reason it didn't :cry:

test/e2e/gameserver_test.go
```diff
@@ -205,7 +208,27 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
+		gs, err := framework.AgonesClient.StableV1alpha1().GameServers(readyGs.Namespace).Get(readyGs.ObjectMeta.Name, metav1.GetOptions{})
+
+		// just in case
+		if k8serrors.IsNotFound(err) {
+			return true, nil
+		}
+
+		if err != nil {
+			logrus.WithError(err).Warn("error retrieving gameserver")
+			return false, nil
```
roberthbailey: Should this be `return false, err`? The func returns `(bool, error)` but none of the cases actually return a non-nil error.

roberthbailey:
lgtm overall, just one nit on error handling.

test/e2e/gameserver_test.go
markmandel: That's a good question - looking at the code, if any error is returned from the function, then `PollImmeadiate` will fail, and return an error - in this case, I don't think we actually want that to happen, we want it to retry and try again if something weird happens - because it may be a timing issue or self resolve?

ilkercelikyilmaz:
> /cc @ilkercelikyilmaz just a heads up about this - we might want to review this behaviour at some point. Originally GameServers not in fleet that dropped into `Unhealthy` did not get deleted (sorry I didn't pick up on this in review).
> 
> But maybe this is fine, and we can leave it as is 
> 
> Just wanted to put it on your radar, in case you had thoughts. In _theory_ the gameserverset controller should have taken care of this for the issue you were facing, but for some strange reason it didn't

I was hoping that GameserverSet controller to take care of this but it wasn't. Actually it was taking care of it when the Pod restarted.

PR #763, opened 2019-05-11
Revert UnHealthy standalone GameServers to not be deleted

markmandel:
This PR does several things:

- Revert back to standalone GameServers that are unhealthy to not be deleted. This should be a better debugging experience when trying to determine why a live GameServer is moving to Unhealthy - as it stays persistent outside a Fleet lifecycle
- Attempt to fix the bug wherein Unhealthy GameServers where leaking when shutting down GameServers that where in fleets - especially  at scale
- Add more testing around shutdown scenarios, such as Shutdown,  Unhealthy, pod deletion and more.

### DIFF
--- a/pkg/gameserversets/controller.go
+++ b/pkg/gameserversets/controller.go
@@ -467,7 +467,7 @@ func computeReconciliationAction(strategy apis.SchedulingStrategy, list []*v1alp
 		toDelete = append(toDelete, potentialDeletions[0:deleteCount]...)
 	}
 
-	if deleteCount > maxDeletions {
+	if len(toDelete) > maxDeletions {
 		toDelete = toDelete[0:maxDeletions]
 		partialReconciliation = true
 	}
--- a/pkg/gameserversets/controller_test.go
+++ b/pkg/gameserversets/controller_test.go
@@ -157,6 +157,44 @@ func TestComputeReconciliationAction(t *testing.T) {
 			wantNumServersToAdd: 0,
 			wantIsPartial:       true,
 		},
+		{
+			desc: "DeletingUnhealthyGameServers",
+			list: []*v1alpha1.GameServer{
+				gsWithState(v1alpha1.GameServerStateReady),
+				gsWithState(v1alpha1.GameServerStateUnhealthy),
+				gsWithState(v1alpha1.GameServerStateUnhealthy),
+			},
+			targetReplicaCount:     3,
+			wantNumServersToAdd:    2,
+			wantNumServersToDelete: 2,
+		},
+		{
+			desc: "DeletingErrorGameServers",
+			list: []*v1alpha1.GameServer{
+				gsWithState(v1alpha1.GameServerStateReady),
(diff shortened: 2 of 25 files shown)


markmandel:
PR left in draft because we're in feature freeze.

@ilkercelikyilmaz would love your review/testing on this - I _think_ I may have solved the Unhealthy leak issue (as I found a bug, and tweaked health checking in the sidecar) - also added more unit and e2e tests for shutdown scenarios - but would definitely like to confirm that this works at scale, which you have the keys to testing on right now :smile:

pkg/gameserversets/controller.go
```diff
@@ -467,7 +467,7 @@ func computeReconciliationAction(strategy apis.SchedulingStrategy, list []*v1alp
 		toDelete = append(toDelete, potentialDeletions[0:deleteCount]...)
 	}
 
-	if deleteCount > maxDeletions {
+	if len(toDelete) > maxDeletions {
```
markmandel: @jkowalski wanted to get your set of eyes on the bug fix above

pkg/gameserversets/controller_test.go
```diff
@@ -157,6 +157,44 @@ func TestComputeReconciliationAction(t *testing.T) {
 			wantNumServersToAdd: 0,
 			wantIsPartial:       true,
 		},
+		{
+			desc: "DeletingUnhealthyGameServers",
```
markmandel: @jkowalski added these extra tests, which resulted in the bug fix above - wanted to do more testing for shutdown scenarios, to try and catch anything that would cause that leaky `Unhealthy` gameservers to rear its head.

Any issues with these tests? And are there other scenarios you can think of that we should test that I didn't cover?

ilkercelikyilmaz:
> PR left in draft because we're in feature freeze.
> 
> @ilkercelikyilmaz would love your review/testing on this - I _think_ I may have solved the Unhealthy leak issue (as I found a bug, and tweaked health checking in the sidecar) - also added more unit and e2e tests for shutdown scenarios - but would definitely like to confirm that this works at scale, which you have the keys to testing on right now

I synced to this change and start running tests. Will let you know how it goes.

ilkercelikyilmaz:
> > PR left in draft because we're in feature freeze.
> > @ilkercelikyilmaz would love your review/testing on this - I _think_ I may have solved the Unhealthy leak issue (as I found a bug, and tweaked health checking in the sidecar) - also added more unit and e2e tests for shutdown scenarios - but would definitely like to confirm that this works at scale, which you have the keys to testing on right now
> 
> I synced to this change and start running tests. Will let you know how it goes.
The test results look OK (See below). I think it is safe to do proceed with this change.
image

markmandel:
Awesome - this makes me very happy :+1: thanks for testing! Will get this ready after the 0.10.0 stable release!

pkg/gameserversets/controller_test.go
jkowalski: LGTM

PR #766, opened 2019-05-14
Add Reserved GameServer State

markmandel:
This adds the `Reserved` state to the GameServer state system.

- Ensures that `Reserved` cannot be deleted on scale down, rolling   updates, etc
- Enforcing the correct rules for the autoscaling system (i.e don't  auto scale up on Reserve)
- Adds ReservedReplicas to GameServerSet and Fleet Status values

This does not implement the sdk.Reserve() functionality, that will come in a later PR.

### DIFF
--- a/pkg/apis/stable/v1alpha1/gameserver.go
+++ b/pkg/apis/stable/v1alpha1/gameserver.go
@@ -55,6 +55,8 @@ const (
 	GameServerStateError GameServerState = "Error"
 	// GameServerStateUnhealthy is when the GameServer has failed its health checks
 	GameServerStateUnhealthy GameServerState = "Unhealthy"
+	// GameServerStateReserved is for when a GameServer is reserved and therefore can be allocated but not removed
+	GameServerStateReserved GameServerState = "Reserved"
 	// GameServerStateAllocated is when the GameServer has been allocated to a session
 	GameServerStateAllocated GameServerState = "Allocated"
 
@@ -337,9 +339,14 @@ func (gs *GameServer) GetDevAddress() (string, bool) {
 	return devAddress, hasDevAddress
 }
 
-// IsAllocated returns true if the server is currently allocated.
-func (gs *GameServer) IsAllocated() bool {
-	return gs.ObjectMeta.DeletionTimestamp == nil && gs.Status.State == GameServerStateAllocated
+// IsDeletable returns false if the server is currently allocated/reserved and is not already in the
+// process of being deleted
+func (gs *GameServer) IsDeletable() bool {
+	if gs.Status.State == GameServerStateAllocated || gs.Status.State == GameServerStateReserved {
+		return !gs.ObjectMeta.DeletionTimestamp.IsZero()
+	}
+
+	return true
 }
 
 // IsBeingDeleted returns true if the server is in the process of being deleted.
--- a/pkg/apis/stable/v1alpha1/gameserver_test.go
+++ b/pkg/apis/stable/v1alpha1/gameserver_test.go
@@ -460,6 +460,27 @@ func TestGameServerGetDevAddress(t *testing.T) {
 	assert.Equal(t, "", devAddress, "dev-address IP address should be 127.1.1.1")
 }
 
+func TestGameServerIsDeletable(t *testing.T) {
+	gs := &GameServer{Status: GameServerStatus{State: GameServerStateStarting}}
+	assert.True(t, gs.IsDeletable())
+
+	gs.Status.State = GameServerStateAllocated
+	assert.False(t, gs.IsDeletable())
+
+	gs.Status.State = GameServerStateReserved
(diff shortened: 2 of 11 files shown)


pkg/apis/stable/v1alpha1/gameserver.go
```diff
@@ -330,16 +332,23 @@ func (gs *GameServer) GetDevAddress() (string, bool) {
...
+		(gs.Status.State == GameServerStateAllocated || gs.Status.State == GameServerStateReserved))
 }
 
 // IsBeingDeleted returns true if the server is in the process of being deleted.
 func (gs *GameServer) IsBeingDeleted() bool {
 	return !gs.ObjectMeta.DeletionTimestamp.IsZero() || gs.Status.State == GameServerStateShutdown
 }
 
+// IsAllocatable is this GameServer in a state that allows it to be allocated?
+func (gs *GameServer) IsAllocatable() bool {
```
jkowalski: CanBeAllocated perhaps?

pkg/apis/stable/v1alpha1/gameserver.go
```diff
@@ -330,16 +332,23 @@ func (gs *GameServer) GetDevAddress() (string, bool) {
 	return devAddress, hasDevAddress
 }
 
-// IsAllocated returns true if the server is currently allocated.
-func (gs *GameServer) IsAllocated() bool {
-	return gs.ObjectMeta.DeletionTimestamp == nil && gs.Status.State == GameServerStateAllocated
+// IsDeletable returns false if the server is currently allocated/reserved and should not be deleted
+func (gs *GameServer) IsDeletable() bool {
+	return !(gs.ObjectMeta.DeletionTimestamp.IsZero() &&
```
jkowalski: Not sure I understand the condition here, how about this for readability:

```
switch {
  case !gs.ObjectMeta.DeletionStamp.IsZero(): return false // already deleted
  case gs.Status.State == GameServerAllocated: return false
  case gs.Status.State == GameServerReserved: return false
  default: return true
}
```

pkg/apis/stable/v1alpha1/gameserver_test.go
```diff
@@ -474,6 +474,28 @@ func TestGameServerGetDevAddress(t *testing.T) {
...
+
+	gs.Status.State = GameServerStateReady
+	assert.True(t, gs.IsAllocatable())
+
+	gs.Status.State = GameServerStateReserved
+	assert.True(t, gs.IsAllocatable())
+}
+
+func TestGameServerIsDeletable(t *testing.T) {
+	gs := &GameServer{Status: GameServerStatus{State: GameServerStateStarting}}
```
jkowalski: add tests with DeletionTimestamp?

site/content/en/docs/Reference/agones_crd_api_reference.html
```diff
@@ -3823,6 +3824,17 @@ <h3 id="FleetStatus">FleetStatus
 </tr>
 <tr>
 <td>
+<code>reservedReplicas</code></br>
+<em>
+int32
+</em>
+</td>
+<td>
+<p>ReservedReplicas are hte number of Reserved GameServer replicas</p>
```
jkowalski: s/hte/the/

perhaps explain what Reserved is?

pkg/apis/stable/v1alpha1/gameserver.go
markmandel: That reads much better :+1: 

`if gs.CanBeAllocated() { ... }`

markmandel: Actually,I just realised this never gets used _anywhere_ :man_facepalming:  so I'm going to delete it.

pkg/apis/stable/v1alpha1/gameserver_test.go
markmandel: Good catch - added!

pkg/apis/stable/v1alpha1/gameserver.go
markmandel: This got a little more nuanced, to ensure it kept passing the tests for `computeReconciliationAction` - PTAL, let me know if this new format is more readable though.

markmandel:
These changes have been done - is this good to go now?

markmandel:
Sure! I we can also do that via github these days too. But I'll update and squash! :+1:

ilkercelikyilmaz:
> Sure! I we can also do that via github these days too. But I'll update and squash!

I am curious to learn more about it.

markmandel:
@ilkercelikyilmaz when you go to merge, we only have the option to "Merge and Squash" enabled - so anytime you merge, you have the opportunity to squash as well.

https://help.github.com/en/articles/about-pull-request-merges#squash-and-merge-your-pull-request-commits

PR #768, opened 2019-05-15
Adding an allocator service that acts as a reverse proxy.

pooneh-m:
Allocator service will solve the problem of sending allocation request from outside the kubernetes cluester as discussed in this issue https://github.com/GoogleCloudPlatform/agones/issues/597#issuecomment-486884886. The service has an external IP and authenticate using client cert.


cmd/allocator/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+	h := httpHandler{
+		agonesClient: agonesClient,
+		namespace:    os.Getenv("NAMESPACE"),
+	}
+
+	// TODO: add liveness probe
+	http.HandleFunc("/v1alpha1/gameserverallocation", h.postOnly(h.allocateHandler))
+
+	caCertPool, err := getCaCertPool(certDir)
```
jkowalski: nit: Ca => CA (it's recommended to maintain proper capitalization in Go)

cmd/allocator/main.go
```diff
@@ -0,0 +1,165 @@
...
+	// and all client certs should be added.
+	caCertPool := x509.NewCertPool()
+	filesInfo, err := ioutil.ReadDir(path)
+	if err != nil {
+		return nil, fmt.Errorf("error reading certs from dir %s: %s", path, err.Error())
+	}
+
+	for _, file := range filesInfo {
+		if strings.HasSuffix(file.Name(), ".crt") || strings.HasSuffix(file.Name(), ".pem") {
+			certFile := path + file.Name()
```
jkowalski: filepath.Join() instead of "+" is more explicit.

cmd/allocator/main.go
```diff
@@ -0,0 +1,165 @@
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	}
+
+	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(h.namespace)
+	allocatedGsa, err := allocation.Create(&gsa)
+	if err != nil {
+		http.Error(w, err.Error(), httpCode(err))
+	}
+	w.Header().Set("Content-Type", "application/json")
+	result, _ := json.Marshal(allocatedGsa)
```
jkowalski: json.NewEncoder(w).Encode(allocatedGsa) is a one-line shortcut you can use.

cmd/allocator/main.go
```diff
@@ -0,0 +1,165 @@
...
+	_, err = io.WriteString(w, string(result))
+	if err != nil {
```
jkowalski: at this point returning http.Error will probably have no effect, since we already sent the headers, I would just log a warning.

cmd/allocator/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+type httpHandler struct {
+	agonesClient versioned.Interface
+	namespace    string
+}
+
+func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
+	gsa := allocationv1alpha1.GameServerAllocation{}
+	if err := json.NewDecoder(r.Body).Decode(&gsa); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
jkowalski: probably better not to return err.Error() to the user - just return something vague like "invalid request"

cmd/allocator/main.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(h.namespace)
+	allocatedGsa, err := allocation.Create(&gsa)
+	if err != nil {
+		http.Error(w, err.Error(), httpCode(err))
```
jkowalski: you need to return after http.Error()

cmd/allocator/main.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return nil, errors.New("Could not create the agones api clientset")
+	}
+
+	return agonesClient, nil
+}
+
+func getCaCertPool(path string) (*x509.CertPool, error) {
+	// Add all certificates under client-certs path because there could be multiple clusters
+	// and all client certs should be added.
```
jkowalski: Normally this should be one CA certificate regardless of # of clusters, clients will use leaf certificates signed by this CA, so you technically only need one PEM file with all CA certificate bundle, but not all the client certs.

jkowalski: IOW - we need make sure we have the certificates example set up correctly, because people will be cargo-culting this a lot.

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -0,0 +1,189 @@
...
+  tls.crt: {{ b64enc $cert.Cert }}
+  tls.key: {{ b64enc $cert.Key }}
+{{- else }}
+  tls.crt: {{ .Files.Get "certs/allocator/server.crt" | b64enc }}
+  tls.key: {{ .Files.Get "certs/allocator/server.key" | b64enc }}
+{{- end }}
+
+---
+# Allocation client certs
+{{- $ca := genCA "allocation-client-ca" 3650 }}
```
jkowalski: pretty sure you want one CA and not separate CA for client and server

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -0,0 +1,189 @@
...
+kind: Secret
+metadata:
+  name: allocator-client-certs
+  labels:
+    app: {{ template "agones.fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+data:
+{{- if .Values.agones.allocator.generateTLS }}
```
jkowalski: generateTLS?

cmd/allocator/main.go
pooneh-m: This acts as a revocation list for certificates as well. If a secret is compromised, its CA should be revoked without impacting other clients calling allocation service. Having this pattern is valuable for match making and not for cluster to cluster calls. If a cluster secret is compromised, then all secrets to talk to another cluster is compromised. However, matchmaker is an independent entity and this solution helps securing system from matchmakers bad actors.

cmd/allocator/main.go
pooneh-m: Done. Thanks!

cmd/allocator/main.go
pooneh-m: This also returns the http status and body with error message.

install/helm/agones/templates/service/allocation.yaml
pooneh-m: Yes. Generate service TLS and adds its CA to the CA list.

cmd/allocator/main.go
pooneh-m: Thanks!

install/helm/agones/templates/service/allocation.yaml
pooneh-m: Server CA is needed if the server is self-signed. Otherwise, client can confirm the server's validity by consulting with CAs signed by valid certificate authorities e.g. DigiCert. Client CA list indicates a list of clients that a server accepts and acts as an authN.

cmd/allocator/main.go
```diff
@@ -0,0 +1,169 @@
...
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+	}
+}
+
+type httpHandler struct {
+	agonesClient versioned.Interface
+	namespace    string
+}
+
+func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
```
markmandel: Quite probably overkill - but in theory you could pull out the deseriasation logic from [here](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameserverallocations/https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameserverallocations/controller.go#L445) -- but quite likely total overkill :smile:

cmd/allocator/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+func main() {
+	agonesClient, err := getAgonesClient()
+	if err != nil {
+		logger.WithError(err).Fatal("could not create agones client")
+	}
+
+	h := httpHandler{
+		agonesClient: agonesClient,
+		namespace:    os.Getenv("NAMESPACE"),
```
markmandel: Any particular reason to tie this to a specific namespace? We could grab the namespace from the `GameServerAllocation.ObjectMeta` on deserialisation -- might be more flexible?

cmd/allocator/main.go
```diff
@@ -0,0 +1,169 @@
...
+// A handler for the web server
+type handler func(w http.ResponseWriter, r *http.Request)
+
+func main() {
+	agonesClient, err := getAgonesClient()
+	if err != nil {
+		logger.WithError(err).Fatal("could not create agones client")
+	}
+
+	h := httpHandler{
```
markmandel: Do we think we'll do a gRPC endpoint as well? Just curious what your take was.

cmd/allocator/main.go
pooneh-m: I agree [allocationDeserialization](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameserverallocations/controller.go#L445) is overkill at this level. The server is acting as a reverse proxy and pass whatever that is passed to it to the API server.

cmd/allocator/main.go
pooneh-m: I am introducing gRPC interface and that will not support k8s APIs including ObjectMeta. The idea is allocator service that is deployed to a namespace is responsible for serving allocation in that namespace. Then MatchMaking does not need to be aware of internal structure of k8s and namespaces and only calls the allocator's endpoint.

cmd/allocator/main.go
pooneh-m: Yes. In the next iteration, I will add the gRPC.

cmd/allocator/main.go
markmandel: But if a user wants to run GameServers in more than one namespace, then they have to run two different services - which brings us back to the same place - the matchmaker being namespace aware (and we have to run twice the infrastructure, and potentially redeploy if we want to be dynamic about the namespaces being used).

Seems simpler to me to allow the endpoint to take a namespace as an agreement? (Maybe the "default" namespace if the default?)

Wdyt?

pooneh-m: If there are two allocator services deployed to k8s in different namespaces, matchmaker only needs to know the endpoints. However, if we want to have one allocator service deployed to one cluster and handle allocation requests for two namespaces for potentially two different purposes, then there will not be enough isolation between traffics for the two namespaces. IOW QPS for one namespace may impact allocator service performance for another namespace. So there is a trade off.

I would prefer not to expose namespace unless we for sure know that it is needed. Because adding additional fields are easy to do but removing them in future is hard as it will be a breaking change. WDYT?

markmandel: > IOW QPS for one namespace may impact allocator service performance for another namespace. So there is a trade off.

Everything comes back to the k8s api anyway, so we'll always have that bottleneck?

> I would prefer not to expose namespace unless we for sure know that it is needed. Because adding additional fields are easy to do but removing them in future is hard as it will be a breaking change. WDYT?

I can't disagree with that statement :+1: 

The only other few devil's advocate statement I can make, is that I think that this makes things for the end user a tad more complicated. Up until this point, everything is installed in the `agones-system` namespace - now we have Agones system components bleeding into other areas of Kubernetes, whereas before they were pretty tightly contained in the `agones-system` namespace.

The other thing is - we're saying we're a reverse proxy for this CRD, but we are changing the expected behaviour of that CRD with the reverse proxy.  So it might be a bit confusing for users.

But given your excellent point above - I think we'll be okay to have the namespace defined in the env var -- and see how users like it. Much easier to add later :+1:

pooneh-m: > > IOW QPS for one namespace may impact allocator service performance for another namespace. So there is a trade off.
> 
> Everything comes back to the k8s api anyway, so we'll always have that bottleneck?
> 
Good point. I think you are right.
> > I would prefer not to expose namespace unless we for sure know that it is needed. Because adding additional fields are easy to do but removing them in future is hard as it will be a breaking change. WDYT?
> 
> I can't disagree with that statement 
> 
> The only other few devil's advocate statement I can make, is that I think that this makes things for the end user a tad more complicated. Up until this point, everything is installed in the `agones-system` namespace - now we have Agones system components bleeding into other areas of Kubernetes, whereas before they were pretty tightly contained in the `agones-system` namespace.
> 
> The other thing is - we're saying we're a reverse proxy for this CRD, but we are changing the expected behaviour of that CRD with the reverse proxy. So it might be a bit confusing for users.

Can you please explain more? How are we changing the expected behavior?
> 
> But given your excellent point above - I think we'll be okay to have the namespace defined in the env var -- and see how users like it. Much easier to add later

markmandel: > Can you please explain more? How are we changing the expected behavior?

Sure - basically, through the k8s api I can pass through what namespace I'm working with as part of the URL path - so I can access whatever namespace I want. That being said, in this instance, we aren't doing that - so it's probably not that huge a deal.

The only potentially confusing thing I see is if a user sets the namespace in `ObjectMeta.Namespace` and it doesn't translate through to the applied namespace in the service.

But I don't see either of these things as blocking issues. As you say above, we can add this functionality later if we need it.

(Also, if someone wants to work on a new namespace, we have to provision a service account and rbac rules anyway, so it's not like you can dynamically add/remove namespace suport that quickly)

pooneh-m: In the gRPC interface that I will introduce next, the namespace is not exposed. So ObjectMeta.Namespace will not be relevant.

Issue #765, opened 2019-05-14
If GameServer webhook validation fails, it doesn't raise up to Fleet


markmandel:
**What happened**:
Create a Fleet with an invalid GameServer specification - no error was returned, and no GameServer was created.

For example (container is in the wrong spot):

```yaml
apiVersion: "stable.agones.dev/v1alpha1"
kind: Fleet
metadata:
  name: zoba-fleet
spec:
  container: 1st
  replicas: 1
  scheduling: Packed
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
  template:
    metadata:
      labels:
        foo: bar
    spec:
      ports:
      - name: default
        portPolicy: Dynamic
        containerPort: 7788
      - containerPort: 3434
        protocol: TCP
        name: tcp
      health:
        initialDelaySeconds: 30
        periodSeconds: 10
      template:
        spec:
          containers:
          - name: 1st
            image: nginx
          - name: 2nd
            image: debian
            volumeMounts:
            - name: html
              mountPath: /html
            command: ["/bin/sh", "-c"]
            args:
              - while true; do
                  date >> /html/index.html;
                  sleep 1;
                done
```

**What you expected to happen**:

Should return a validation error as soon as the Fleet is created - i.e. the Fleet shouldn't get created at all, and the user should get a message as to why.

**How to reproduce it (as minimally and precisely as possible)**:

copy paste the above into bad-fleet.yaml, and `kubectl apply -f bad-fleet.yaml`

**Anything else we need to know?**:

**Environment**:
- Agones version: 0.9.0
- Kubernetes version (use `kubectl version`): 1.11.x
- Cloud provider or hardware configuration: ???
- Install method (yaml/helm): ???
- Troubleshooting guide log(s): ???
- Others:

aLekSer:
I will make a pull request to fix this issue.

PR #771, opened 2019-05-15
Add validation of the fleet underlying gameserver

aLekSer:
Perform additional validation on a Fleet Gameserver specification .
For #765.

### DIFF
--- a/pkg/apis/stable/v1alpha1/fleet.go
+++ b/pkg/apis/stable/v1alpha1/fleet.go
@@ -153,6 +153,14 @@ func (f *Fleet) Validate() ([]metav1.StatusCause, bool) {
 		})
 	}
 
+	// check gameserver specification in a fleet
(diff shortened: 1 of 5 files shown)


markmandel:
So this is great for a quick and easy fix (which is nice :+1:) - but makes me wonder if we should move some/all the validation functions down to the *Spec implementations - and then it would be far easier to pass this kind of things down?

Or maybe another way - 

Add a Validate() to `GameServerTemplate` that calls a `GameServerSpec.Validate()`, which is also called by `GameServer.Validate()`

WDYT? Would that be cleaner / easier to maintain?

aLekSer:
Agree with you, need to update this PR.

markmandel:
Should we write a e2e test for this? WDYT?

aLekSer:
Yes, I will prepare E2E test for this change

PR #776, opened 2019-05-16
Fixing SDK proto file according to style guide

aLekSer:
Use underscore for message field names.
Run linter results for proto file, namely protoc-gen-lint.

[Procol buffers style guide](https://developers.google.com/protocol-buffers/docs/style#message-and-field-names)


markmandel:
This would break everything - wouldn't it?
- sidecar
- the code for the clients
- etc?

aLekSer:
Yes, I was also thinking that it would break every SDK. For Golang field names was not changed.
For NodeJS they become more readable:
was `getInitialdelayseconds`, become `getInitialDelaySeconds`.
Same for C++: was `clear_periodseconds`.
So I think this change is needed at some point before 1.0 release. Now proto file contains mixture of two styles.
>This would break everything - wouldn't it?
    sidecar
    the code for the clients
    etc?
The code for the clients which depends on `Health` message should be change after this. Re sidecars I am going to check the impact.
That's why this PR intentionally left draft.

markmandel:
Got it - you are right, if we are going to do this, we need to do it before 1.0

We should also put the linter as part of our test suit, so we don't end up here again :+1:

markmandel:
Where did we end up with this? Sounds like something we should round off?

aLekSer:
As further steps which I need to add to this or to a new PR:
 > We should also put the linter as part of our test suit

1. So I am going to add this to the tests at least as Make target.
1. There is one more identifier which treated as error:
```
sdk.proto:23:9: 'SDK' - Use CamelCase (with an initial capital) for service names.
```

aLekSer:
Another project which is easier to use is `protolint`, however it does not handle `option (google.api.http) = {` well.
https://github.com/yoheimuta/protolint/issues/21
it does not treat `SDK`

markmandel:
Fair enough. Will just have to be diligent about it then.

I actually wonder if this will work fine with backware compatabiltiy. We're just changing the names, but not the data positions or the types of data. Protobuf I think will handle it fine.

PR #780, opened 2019-05-17
Prevent race conditions by syncing cache on new Allocation elements

markmandel:
Added cache.InformerSynced members for the new resources that are worked on within this code for the multicluster allocations, to ensure they aren't empty values on controller startup.

Also applying these at the test level, so we don't get random test flakiness due to cache's not being populated in time for the test to run.


markmandel:
/cc @pooneh-m fyi.

Apologies for not picking this up in code review :disappointed: 

Please let me know if you need to me explain what is going on here :+1:

pkg/gameserverallocations/controller_test.go
```diff
@@ -75,7 +75,7 @@ func TestControllerAllocationHandler(t *testing.T) {
 			return true, gs, nil
 		})
 
-		stop, cancel := agtesting.StartInformers(m)
+		stop, cancel := agtesting.StartInformers(m, c.gameServerSynced)
```
pooneh-m: if the controller has the WaitForCacheSync, is it really needed to call StartInformers that does the same thing (e.g. calling [WaitForCacheSync](https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/testing/controller.go#L73))

pooneh-m:
LGTM

pkg/gameserverallocations/controller_test.go
markmandel: The `WaitforcacheSync` is called in the controller `.Run()` command - so just instantiating the controller doesn't fire the `WaitforcacheSync`.

And in tests (I've found anyway), you often don't want everything that happens inside the `Run()` command to fire.

This was following the  patterns of the [sample controller](https://github.com/kubernetes/sample-controller/blob/master/controller.go#L159)

pkg/gameserverallocations/controller_test.go
```diff
@@ -893,7 +893,7 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
 				return true, getTestSecret(secretName, server.TLS.Certificates[0].Certificate[0]), nil
 			})
 
-		stop, cancel := agtesting.StartInformers(m)
+		stop, cancel := agtesting.StartInformers(m, c.secretSynced)
```
pooneh-m: agtesting.StartInformers(m, c.allocationPolicySynced, c.gameServerSynced, c.secretSynced) here and in other places?

markmandel: Nicely spotted! Yep - exactly. Pushing up a change now. :100:

PR #784, opened 2019-05-20
Packed Strategy Gameserverallocation fix

### LINKED ISSUE #783: Packed Allocation is very not packed
**What happened**:

* Cluster with 20 nodes for gameservers
* Create a Fleet with 100 GameServers
* Created 6 Allocations, and as you can see, they are almost all on different nodes:

```shell
root@30dacd3ba219:/go/src/agones.dev/agones# kubectl get gs | grep Alloc | sort
xonotic-27wqm-82mxs   Allocated   34.85.93.181     7607      gke-test-cluster-default-ace71fa2-gpw3   1m
xonotic-27wqm-dr5ql   Allocated   34.85.103.79     7341      gke-test-cluster-default-ace71fa2-1193   2m
xonotic-27wqm-jsbld   Allocated   34.85.115.151    7155      gke-test-cluster-default-ace71fa2-jzql   2m
xonotic-27wqm-v79zp   Allocated   34.85.103.79     7768      gke-test-cluster-default-ace71fa2-1193   1m
xonotic-27wqm-vbcqw   Allocated   35.243.109.91    7555      gke-test-cluster-default-ace71fa2-9f80   1m
```

Gameservers should have been allocated to the same node (or at least as close as we can get given some performance constraints)

This is far less than ideal, as this will cost our users quite large $$$ at large scale as they will be unable to scale down resources that may be mostly empty. It also defies the "Packed" contract.

Create any of the fleets we have, push it up to a scale such that it will encompass multiple nodes (xonotic is good for this, as it has a larger memory footprint).

Then allocate several gameservers, and `kubectl get gs` to see the results.

**Environment**: Linux
- Agones version: 1.10.0
- Kubernetes version (use `kubectl version`): 1.11.x
- Others: N/A

:warning: **I am marking this as a 1.0 blocker, because of the large cost it could incur on our users. Please feel free to disagree in comment below** :warning:

aLekSer:
Add `NodeName` compare for Packed Strategy when draw in Ready and Allocated Gameservers count per `Node`.

When we are performing node search using the Packed Strategy and we have all gameservers spread equally between nodes, we will select node alphabetically. In that case simultaneous calls to allocate will result in first nodeName in lexicographic order. 
Clear the list of best GameServers when Packed.
As we go through GS list we were adding Gameservers from different nodes.

Verified on GKE with `simple-udp` server that issue does not reproduce on Packed strategy.

Closes #783

### DIFF
--- a/pkg/gameserverallocations/controller.go
+++ b/pkg/gameserverallocations/controller.go
@@ -604,10 +604,16 @@ func (c *Controller) findReadyGameServerForAllocation(gsa *v1alpha1.GameServerAl
 	}
 
 	var bestGSList []stablev1alpha1.GameServer
+	bestNode := ""
 	for nodeName, gs := range allocationSet {
 		count := counts[nodeName]
-		// bestGS == nil: if there is no best GameServer, then this node & GameServer is the always the best
+		// bestGS == nil: if there is no best GameServer, then this node & GameServer is always the best
 		if bestGS == nil || comparator(*bestCount, count) {
+			if gsa.Spec.Scheduling == apis.Packed && bestNode != gs.Status.NodeName {
+				bestNode = gs.Status.NodeName
+				// Clear the list if we found new node for Packed strategy
+				bestGSList = bestGSList[:0]
+			}
 			bestCount = &count
 			bestGS = gs
 			bestGSList = append(bestGSList, *gs)
--- a/pkg/fleetallocation/find.go
+++ b/pkg/fleetallocation/find.go
@@ -23,6 +23,7 @@ import (
 type nodeCount struct {
 	ready     int64
 	allocated int64
+	nodeName  string
 }
 
 // findReadyGameServerForAllocation is a O(n) implementation to find a GameServer with priority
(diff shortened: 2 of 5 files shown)


markmandel:
Not sure this is going to work, because of the random selection in: https://github.com/GoogleCloudPlatform/agones/blob/master/pkg/gameserverallocations/controller.go#L620

pkg/gameserverallocations/controller.go
```diff
@@ -604,10 +604,16 @@ func (c *Controller) findReadyGameServerForAllocation(gsa *v1alpha1.GameServerAl
...
+	bestNode := ""
 	for nodeName, gs := range allocationSet {
 		count := counts[nodeName]
-		// bestGS == nil: if there is no best GameServer, then this node & GameServer is the always the best
+		// bestGS == nil: if there is no best GameServer, then this node & GameServer is always the best
 		if bestGS == nil || comparator(*bestCount, count) {
+			if gsa.Spec.Scheduling == apis.Packed && bestNode != gs.Status.NodeName {
+				bestNode = gs.Status.NodeName
+				// Clear the list if we found new node for Packed strategy
+				bestGSList = bestGSList[:0]
```
aLekSer: Here I clean the list right before 
```
bestGS = c.getRandomlySelectedGS(gsa, bestGSList)
```

markmandel: That's going to ruin throughput though - it's going to be highly contentious.

markmandel:
Should we close this PR? I'm going to work on the batching.

aLekSer:
Yes, closing this pull request, because another approach was selected

PR #788, opened 2019-05-22
Docs: FleetAllocation example fix

rorygarand:
The example in the docs assumes a fleet with two game servers, then tries to allocate 3 game servers.

This results with a cryptic "NotFound" error resulting from trying to allocate a resource that doesn't exist.

A better solution may be to replace this example with GSA instead of FA.


aLekSer:
@markmandel Yes, that is correct, because `examples/simple-udp/fleet.yaml` has 2 replicas in it.

PR #789, opened 2019-05-23
Flaky: TestAllocator

markmandel:
Attempt to fix flakiness in this e2e test.

Theory is that the allocation deployment is taking a little
while to come online.

### DIFF
--- a/build/e2e-image/entrypoint.sh
+++ b/build/e2e-image/entrypoint.sh
@@ -31,5 +31,5 @@ echo "Waiting consul port-forward to launch on 8500..."
 timeout 60 bash -c 'until printf "" 2>>/dev/null >>/dev/tcp/$0/$1; do sleep 1; done' 127.0.0.1 8500
 echo "consul port-forward launched. Starting e2e tests..."
 consul lock -child-exit-code=true -timeout 30m -try 30m -verbose LockE2E /root/e2e.sh
-killall -q kubectl
-
+killall -q kubectl || true
+echo "successfully killed kubectl proxy"
--- a/test/e2e/allocator_test.go
+++ b/test/e2e/allocator_test.go
@@ -22,12 +22,15 @@ import (
 	"io/ioutil"
 	"net/http"
 	"testing"
+	"time"
 
 	"agones.dev/agones/pkg/apis/allocation/v1alpha1"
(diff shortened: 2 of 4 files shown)


build/e2e-image/entrypoint.sh
```diff
@@ -31,5 +31,5 @@ echo "Waiting consul port-forward to launch on 8500..."
 timeout 60 bash -c 'until printf "" 2>>/dev/null >>/dev/tcp/$0/$1; do sleep 1; done' 127.0.0.1 8500
 echo "consul port-forward launched. Starting e2e tests..."
 consul lock -child-exit-code=true -timeout 30m -try 30m -verbose LockE2E /root/e2e.sh
-killall -q kubectl
-
+killall -q kubectl || true
```
markmandel: @Kuqd for interest's sake - this was the fix for the e2e tests. even with `-q` this was always returning a non zero status code. :man_shrugging:

test/e2e/allocator_test.go
```diff
@@ -70,9 +73,23 @@ func TestAllocator(t *testing.T) {
 	if !assert.Nil(t, err) {
 		return
 	}
-	response, err := client.Post(requestURL, "application/json", bytes.NewBuffer(body))
-	if !assert.Nil(t, err) {
-		return
+
+	// wait for the allocation system to come online
+	var response *http.Response
```
markmandel: @pooneh-m for interest's sake - this was what (seems to have) solved the flakiness issue. My theory is that there was a race condition wherein the backing pods for the service would come up after this test was run, but only sometimes (probably depending on e2e test order).

test/e2e/allocator_test.go
```diff
@@ -70,9 +73,23 @@ func TestAllocator(t *testing.T) {
...
+			return false, nil
+		}
+
+		return true, nil
+	})
+
+	if !assert.NoError(t, err) {
+		assert.FailNow(t, "Http test failed")
 	}
 	defer response.Body.Close() // nolint: errcheck
```
pooneh-m: remove?

test/e2e/allocator_test.go
pooneh-m: Thanks for the fix. Makes sense.

PR #793, opened 2019-05-25
Fix publishDate on unreal docs

markmandel:
Oops, this actually blocked the site updating. Surprised it didn't show up in CI actually.

This should unblock it!


roberthbailey:
lgtm

PR #811, opened 2019-06-05
Fixed missing apiGroup field in gameserver-allocator ServiceAccount role binding

structurefall:


markmandel:
/cc @pooneh-m for visibility

I believe this is to solve the issue you had when trying to install via install.yaml (which didn't work anyway, because master is always broken, since it's development)

```
error: error validating "agones-install.yaml": error validating data: ValidationError(RoleBinding.roleRef): missing required field "apiGroup" in io.k8s.api.rbac.v1.RoleRef; if you choose to ignore these errors, turn validation off with --validate=false
```

Change will need to be made in the relevant helm template, then run `make gen-install` to generate the install.yaml

@structurefall what Kubernetes version was this error received on?

pooneh-m:
https://github.com/googleforgames/agones/pull/825

PR #812, opened 2019-06-06
Flaky: TestAutoscalerWebhook

markmandel:
Replace flaky elements with polling strategy.


aLekSer:
Nice fix, it is needed to write tests without `time.Sleep()`. Will be creating new with this idea in mind.

markmandel:
Thank you! :tada:

Issue #809, opened 2019-06-05
Allocator service needs to get the namespace from input and not environment.


pooneh-m:
The allocator service is installed in the agones-system namespace. The current implementation is to get the [namespace](https://github.com/googleforgames/agones/blob/master/cmd/allocator/main.go#L56) from environment for creating gameserverallocation. It needs to be change to get the namespace from the request.

PR #814, opened 2019-06-06
Remove env namespace dependency from allocator service.

pooneh-m:
issue https://github.com/googleforgames/agones/issues/809

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"os"
 	"path/filepath"
 	"strings"
 
@@ -53,7 +52,6 @@ func main() {
 
 	h := httpHandler{
 		agonesClient: agonesClient,
-		namespace:    os.Getenv("NAMESPACE"),
 	}
 
 	// TODO: add liveness probe
@@ -133,7 +131,6 @@ func (h *httpHandler) postOnly(in handler) handler {
 
 type httpHandler struct {
 	agonesClient versioned.Interface
-	namespace    string
 }
 
 func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
@@ -143,7 +140,7 @@ func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(h.namespace)
+	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(gsa.ObjectMeta.Namespace)
 	allocatedGsa, err := allocation.Create(&gsa)
 	if err != nil {
 		http.Error(w, err.Error(), httpCode(err))
--- a/test/e2e/allocator_test.go
+++ b/test/e2e/allocator_test.go
@@ -60,6 +60,9 @@ func TestAllocator(t *testing.T) {
 	}
 	framework.WaitForFleetCondition(t, flt, e2e.FleetReadyCount(flt.Spec.Replicas))
 	gsa := &v1alpha1.GameServerAllocation{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: defaultNs,
+		},
 		Spec: v1alpha1.GameServerAllocationSpec{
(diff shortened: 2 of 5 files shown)


cmd/allocator/main.go
```diff
@@ -143,7 +140,7 @@ func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(h.namespace)
+	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(gsa.Namespace)
```
markmandel: Super minor nit - we've standardised on `{foo}.ObjectMeta.{var}` rather than going the shortcut version. 

```suggestion
	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(gsa.ObjectMeta.Namespace)
```

No major reason, just picked a version arbitrarily at the beginning, would be nice to keep it through :+1: 

Everything else looks great - weird that it's failing.

pooneh-m: I figured. I am getting 403 because I am trying to create gameserverallocation from a different namespace. It should be agones-system namespace. Actually this PR is not needed. The allocator service is correctly forwarding the allocation request to gameserverallocation api server extension.

pooneh-m: To simplify I will pass the agones-system as a namespace constant.

cmd/allocator/main.go
```diff
@@ -143,7 +141,7 @@ func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(h.namespace)
+	allocation := h.agonesClient.AllocationV1alpha1().GameServerAllocations(agonesSystemNamespace)
```
markmandel: I'm pretty sure you were right the first time - this should be `gsa.ObjectMeta.Namespace` (we could also default this to "default" if wanted to?, since that is the default for everything else) 

What we have here is going to create a GameServerAllocation that (should) searches in the agones-system namespace - because that is the namespace it is coming from. This works right now because the current GSA logic is buggy.

pooneh-m: Yes. Confusion over confusion ;) for a second I thought gsa is tight to agones-system because it is an API server extension, but it actually receives the namespace that the request is sent to as part of the request. I am defining cluster role to let allocator create gs allocation requests in multiple namespaces.

test/e2e/allocator_test.go
```diff
@@ -80,7 +83,9 @@ func TestAllocator(t *testing.T) {
 		response, err = client.Post(requestURL, "application/json", bytes.NewBuffer(body))
 
 		if err != nil {
-			response.Body.Close() // nolint: errcheck
+			if response != nil {
+				response.Body.Close() // nolint: errcheck
```
markmandel: Ah! Small Go gotcha. response.Body.Close() should always get run if the response not nil. Usually `defer` is a nice way to do it.

Details, and examples: https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body-in-golang

pooneh-m: response.Body.Close() is happening after the poll. So this one was not necessary.

markmandel: Aha! Nice!

markmandel:
So this is consistently failing on this flaky test:

```
--- FAIL: TestAutoscalerWebhook (88.75s)
    fleetautoscaler_test.go:326: 
        	Error Trace:	fleetautoscaler_test.go:326
        	Error:      	Expected nil, but got: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:""}, Status:"Failure", Message:"Operation cannot be fulfilled on fleetautoscalers.stable.agones.dev \"simple-fleet-vc2z8-autoscaler\": the object has been modified; please apply your changes to the latest version and try again", Reason:"Conflict", Details:(*v1.StatusDetails)(0xc000992600), Code:409}}
        	Test:       	TestAutoscalerWebhook
```

I'll rerun this once #812 has been merged. That should solve this problem and let this pass.

PR #817, opened 2019-06-08
PortPolicy of Passthrough - Same Port for Container and Host

### LINKED ISSUE #294: In case of dynamic port allocation, offer the option to set the container port to the same value as the host port
Currently dynamic port allocation assumes that the game server listens on a static port (i.e. 7000) and it allocates dynamically an "external" port (i.e. 7123) that the client will actually use to connect.

I would like to have the option to set the container port the same as the host port. This could be configured for example by not specifying any container port or by setting it to a special value (0 or *).

Why is this a useful feature:
- it would help debugging, cross-matching client and server logs; it's harder to identify the correct server or follow the logs if the client and server are logging different ports
- better compatibility with the multi-tenancy offered by other hosting providers; in that case, each server instance running on the same machine (virtual or physical) has a different port it listens to; the same port is exposed to clients
- it's easier to understand by people that have no experience with docker containers

markmandel:
This implements a new PortPolicy, such that the containerPort that is specified for the game server is dynamically set to the same random value that the hostPort is.

This is useful for game servers that have already been configured to broadcast the port that they originally started on as their connection port, with no overrides.

Closes #294


markmandel:
Question for all - are we happy with the name of the Port Policy being `Passthrough` ? Was the best thing I could come up with, but I'm not wedded to it. If you have a better name, please say so.

/cc @victor-prodan just to show you that I finally did this :smile:

examples/simple-udp/main.go
```diff
@@ -36,17 +36,15 @@ func main() {
...
 	flag.Parse()
 	if ep := os.Getenv("PORT"); ep != "" {
 		port = &ep
 	}
-
-	log.Printf("Starting UDP server, listening on port %s", *port)
-	conn, err := net.ListenPacket("udp", ":"+*port)
-	if err != nil {
-		log.Fatalf("Could not start udp server: %v", err)
+	if epass := os.Getenv("PASSTHROUGH"); epass != "" {
```
markmandel: @ilkercelikyilmaz side thought - we could use flags like this to do your "shutdown after n minutes" load test - rather than you having to reload all your changes in on every change. WDYT?

ilkercelikyilmaz: Agreed. That's what I had in mind as well.

pkg/gameservers/portallocator.go
```diff
@@ -265,7 +271,7 @@ func (pa *PortAllocator) registerExistingGameServerPorts(gameservers []*v1alpha1
 
 	for _, gs := range gameservers {
```
ilkercelikyilmaz: should we move this below 
if p.PortPolicy == v1alpha1.Dynamic || p.PortPolicy == v1alpha1.Passthrough {

markmandel: Sorry - not 100% sure I follow?

ilkercelikyilmaz: I was thinking that this will be little more efficient :

	for _, p := range gs.Spec.Ports {
		if p.PortPolicy == v1alpha1.Dynamic || p.PortPolicy == v1alpha1.Passthrough {
	             for _, gs := range gameservers {

markmandel: How do we get access to `gs.Spec.Ports`, if we are not within the  `for _, gs := range gameservers` section?

ilkercelikyilmaz: My bad, I didn't realize that (without IDE) that it is looping over the gs.Spec.Ports.

Issue #246, opened 2018-06-08
Unity Plugin SDK


samsgates:
Do you have unity plugin client sdk for accessing game server?

markmandel:
Not as of yet, but once we have #240 complete, this will be much easier. Are you actively creating dedicated game servers in Unity right now? I'd love to hear more about how you are doing that and your experience.

markmandel:
Now that #240 is complete - A SDK for Unity dedicated servers should be relatively trivial to implement.

(Although I'm noting that the request is for a client sdk -- there is no need for anything specific client side for Unity games to run on Agones - it's just an ip and port to communicate with)

GabeBigBoxVR:
Going to start some work on this to get a proof of concept working for Unity game servers client. Mainly to get the health checks working first.

rezam90:
Hi, any plans to have the Unity client?

GabeBigBoxVR:
Yes, I'm working on it now.

markmandel:
Just merged a SDK 🎉. I think we should keep this open until we have an example in the repository as well. Sound good?

markmandel:
@whisper0077 would you be willing to also submit your Unity example using the SDK to this project?

whisper0077:
@markmandel Ok, I will do it in a few days.

markmandel:
I think we can close this now! Fire any objections you may have, otherwise, I will close this tomorrow.

Great stuff!

markmandel:
No objections! Closing!

PR #818, opened 2019-06-08
Add Unity SDK

whisper0077:
Unity SDK for Agones. (issue : #246, closed pr : #395)

I have a sample repository (https://github.com/whisper0077/AgonesUnityServerSample)
I would love to hear feedback.

I'm not good at English sorry.
Please review it.

### DIFF
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -1,0 +1,239 @@
+﻿// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using UnityEngine;
+using UnityEngine.Networking;
+
+namespace Agones
+{
+    /// <summary>
+    /// Agones SDK for Unity.
+    /// </summary>
+    public class AgonesSdk : MonoBehaviour
+    {
+        [Range(1, 5)]
+        public float healthIntervalSecond = 5.0f;
+        public bool healthEnabled = true;
+        public bool logEnabled = false;
+
+        private const string sidecarAddress = "http://localhost:59358";
+        private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+
+        private float CurrentHealthTime { get; set; } = 0;
+
+        private struct KeyValueMessage
+        {
+            public string key;
+            public string value;
+            public KeyValueMessage(string k, string v) => (key, value) = (k, v);
+        }
+
+        #region Unity Methods
+        // Use this for initialization
+        void Start()
+        {
+            CurrentHealthTime = 0;
+        }
+
+        // Update is called once per frame
+        void Update()
+        {
+            if (!healthEnabled) { return; }
+
+            CurrentHealthTime += Time.unscaledDeltaTime;
+            if (CurrentHealthTime >= healthIntervalSecond)
+            {
+                Health();
+                CurrentHealthTime = 0;
+            }
+        }
+
+        void OnApplicationQuit()
+        {
+            cancellationTokenSource.Dispose();
+        }
+        #endregion
+
+        #region AgonesRestClient Public Methods
+        /// <summary>
+        /// Marks this Game Server as ready to receive connections
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The task result contains that the request is success or failure.
+        /// </returns>
+        public async Task<bool> Ready()
+        {
+            return await SendRequestAsync("/ready", "{}");
+        }
+
+        /// <summary>
+        /// Marks this Game Server as ready to shutdown
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The task result contains that the request is success or failure.
+        /// </returns>
+        public async Task<bool> Shutdown()
+        {
+            return await SendRequestAsync("/shutdown", "{}");
+        }
+
+        /// <summary>
+        /// Marks this Game Server as Allocated
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The task result contains that the request is success or failure.
+        /// </returns>
+        public async Task<bool> Allocate()
+        {
+            return await SendRequestAsync("/allocate", "{}");
+        }
+
+        /// <summary>
+        /// Set a metadata label that is stored in k8s
+        /// </summary>
+        /// <param name="key">label key</param>
+        /// <param name="value">label value</param>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The task result contains that the request is success or failure.
+        /// </returns>
+        public async Task<bool> SetLabel(string key, string value)
+        {
+            string json = JsonUtility.ToJson(new KeyValueMessage(key, value));
+            return await SendRequestAsync("/metadata/label", json, UnityWebRequest.kHttpVerbPUT);
+        }
+
+        /// <summary>
+        /// Set a metadata annotation that is stored in k8s
+        /// </summary>
+        /// <param name="key">annotation key</param>
+        /// <param name="value">annotation value</param>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The task result contains that the request is success or failure.
+        /// </returns>
+        public async Task<bool> SetAnnotation(string key, string value)
+        {
+            string json = JsonUtility.ToJson(new KeyValueMessage(key, value));
+            return await SendRequestAsync("/metadata/annotation", json, UnityWebRequest.kHttpVerbPUT);
+        }
+        #endregion
+
+        #region AgonesRestClient Private Methods
+        void Health()
+        {
+            _ = SendRequestAsync("/health", "{}");
+        }
+
+        async Task<bool> SendRequestAsync(string api, string json, string method = UnityWebRequest.kHttpVerbPOST)
+        {
+            // To prevent that an async method leaks after destroying this gameObject
+            cancellationTokenSource.Token.ThrowIfCancellationRequested();
+
+            var req = new UnityWebRequest(sidecarAddress + api, method)
+            {
+                uploadHandler = new UploadHandlerRaw(Encoding.UTF8.GetBytes(json)),
+                downloadHandler = new DownloadHandlerBuffer()
+            };
+            req.SetRequestHeader("Content-Type", "application/json");
+
+            await new AgonesAsyncOperationWrapper(req.SendWebRequest());
+
+            bool ok = req.responseCode == 200;
+
+            if (ok)
+            {
+                Log($"Agones SendRequest ok: {api} {req.downloadHandler.text}");
+            }
+            else
+            {
+                Log($"Agones SendRequest failed: {api} {req.error}");
+            }
+
+            return ok;
+        }
+
+        void Log(object message)
+        {
+            if (!logEnabled) { return; }
+
+#if UNITY_EDITOR
+            Debug.Log(message);
+#else
+            Console.WriteLine(message);
+#endif
+        }
+        #endregion
+
+        #region AgonesRestClient Nested Classes
+        class AgonesAsyncOperationWrapper
+        {
+            public UnityWebRequestAsyncOperation AsyncOp { get; }
+            public AgonesAsyncOperationWrapper(UnityWebRequestAsyncOperation unityOp)
+            {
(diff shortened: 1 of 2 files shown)


aLekSer:
Hello @whisper0077 ,

Thanks for the Pull Request. I could not be a reviewer for CS code. But some general comments:
1. You post a link to your project which uses this code. But it is hard to understand how to build Dockerfile. It just copies `COPY Builds/Server/ ./`. Do I need to have Unity editor or some other dependencies should be met.
1. It seems that you can add some details from your project [Unity Sample](https://github.com/whisper0077/AgonesUnityServerSample/blob/awaitable/README.md) to this `unity.md` to have more clear documentation.
1. Is there a way to add Dockerfile here to this PR also could be with `build.sh` script. As we do here for instance:
https://github.com/googleforgames/agones/tree/release-0.10.0/build/build-sdk-images/go

whisper0077:
Hi @aLekSer, thanks for comments.

My Sample is just a Dedicated Server (Echo Server) that is built by Unity Editor and uses Agones Rest SDK.

> 1. You post a link to your project which uses this code. But it is hard to understand how to build Dockerfile. It just copies COPY Builds/Server/ ./. Do I need to have Unity editor or some other dependencies should be met.

Yes, You need to have Unity Editor for building a dedicated server.
A `Builds/Server` folder is created after a build.
See [Building a Dedicated Server](https://github.com/whisper0077/AgonesUnityServerSample/blob/awaitable/README.md#building-a-dedicated-server) section, I have fixed a README.md.

> 2. It seems that you can add some details from your project Unity Sample to this unity.md to have more clear documentation.

ok, but I have no idea. 
I thought that people who use Unity Editor would understand.
I would appreciate if you sent specific comments.

> 3. Is there a way to add Dockerfile here to this PR also could be with build.sh script. As we do here for instance:

I saw [this document](../tree/master/build/build-sdk-images#adding-support-for-a-new-language).
This SDK uses Agones Rest APIs, so I think that there is no need to add those files.

markmandel:
@dzlier-gcp you free to do a review? You know Unity way better than me :+1:

dzlier-gcp:
Yes I will take a look, I could make a case for this being a Fixit week thing right?

markmandel:
SGTM :1st_place_medal:

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        void Log(object message)
+        {
+            if (!logEnabled) { return; }
+
+#if UNITY_EDITOR
+            Debug.Log(message);
+#else
+            Console.WriteLine(message);
```
dzlier-gcp: I believe Debug.Log works in built players, I don't think you have to redirect to Console.WriteLine.

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+#endif
+        }
+        #endregion
+
+        #region AgonesRestClient Nested Classes
+        class AgonesAsyncOperationWrapper
```
dzlier-gcp: Please add `<summary>` class documentation to these classes and their public fields/methods.

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+            private UnityWebRequestAsyncOperation asyncOp;
+            private Action continuation;
+
+            public AgonesAsyncOperationAwaiter(AgonesAsyncOperationWrapper wrapper)
+            {
+                asyncOp = wrapper.AsyncOp;
+                asyncOp.completed += OnRequestCompleted;
+            }
+
+            public bool IsCompleted => asyncOp.isDone;
```
dzlier-gcp: Fields and attributes should be placed before constructor methods.

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            public void GetResult()
+            {
+                asyncOp.completed -= OnRequestCompleted;
+
+                // remove references
+                asyncOp = null;
```
dzlier-gcp: Please document why this is necessary.

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            public AgonesAsyncOperationAwaiter(AgonesAsyncOperationWrapper wrapper)
+            {
+                asyncOp = wrapper.AsyncOp;
+                asyncOp.completed += OnRequestCompleted;
+            }
+
+            public bool IsCompleted => asyncOp.isDone;
+
+            public void GetResult()
```
dzlier-gcp: This method's name implies that it returns the result of the operation, when it actually just seems to unsubscribe the completed callback and unset its fields. Please either change the name or have the function return an expected value, in addition to adding `<summary>` descriptions.

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+                asyncOp.completed -= OnRequestCompleted;
+
+                // remove references
+                asyncOp = null;
+                continuation = null;
+            }
+
+            public void OnCompleted(Action continuation)
+            {
+                this.continuation = continuation;
```
dzlier-gcp: This class doesn't appear to work if the async operation has already completed by the time this Awaiter is created.

Unless this is impossible, this function should check if the operation is complete and if so, run the callback immediately.

And whenever the continuation function is called, the `continuation` field should be nulled out so as not to be repeated. 

Please also include some sort of `lock` statement on the `continuation` field itself to prevent race conditions.

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+        {
+            CurrentHealthTime = 0;
+        }
+
+        // Update is called once per frame
+        void Update()
+        {
+            if (!healthEnabled) { return; }
+
+            CurrentHealthTime += Time.unscaledDeltaTime;
```
dzlier-gcp: I think this health check would be more appropriate as a Coroutine as opposed to being checked every Update. This would cut out `CurrentHealthTime` entirely and avoid running code on every Update cycle.

In this example, the `Health()` method would also need to also be an async function that returns `true` if the health was successful. That part is recommended, but not required.

```csharp
private void Create() 
{
  if (healthEnabled) {
    StartCoroutine(HealthCheck());
  }
}

private IEnumerator HealthCheck()
{
  while (true) 
  {
    yield return new WaitForSeconds(healthIntervalSecond);
    if (await Health() == false) 
    {
      Debug.Log("Health check failed.");
    }
  }
}
```

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+namespace Agones
+{
+    /// <summary>
+    /// Agones SDK for Unity.
+    /// </summary>
+    public class AgonesSdk : MonoBehaviour
+    {
+        [Range(1, 5)]
+        public float healthIntervalSecond = 5.0f;
+        public bool healthEnabled = true;
```
dzlier-gcp: You could reduce the number of fields here if this was a property that checked if `healthIntervalSecond` is > 0 (and document accordingly).

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+        public bool logEnabled = false;
+
+        private const string sidecarAddress = "http://localhost:59358";
```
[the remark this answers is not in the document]
markmandel: I can confirm - host and port are always the same :+1:

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+        private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+
+        private float CurrentHealthTime { get; set; } = 0;
+
+        private struct KeyValueMessage
```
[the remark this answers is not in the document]
whisper0077: I tried the following code, but it returned `"{}"`;
So `KeyValuePair<K,V>` is not usable for this case.

```csharp
var kv = new KeyValuePair<string, string>("name", "hoge");
// output "{}"
Debug.Log(JsonUtility.ToJson(kv));
```

sdks/unity/AgonesSdk.cs
```diff
@@ -0,0 +1,239 @@
...
+            public string key;
+            public string value;
+            public KeyValueMessage(string k, string v) => (key, value) = (k, v);
+        }
+
+        #region Unity Methods
+        // Use this for initialization
+        void Start()
+        {
+            CurrentHealthTime = 0;
```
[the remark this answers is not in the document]
whisper0077: It is unnecessary, I will remove it.

sdks/unity/AgonesSdk.cs
whisper0077: Oh, It is unnecessary, I will remove it.

sdks/unity/AgonesSdk.cs
whisper0077: `await` can not be used in a not async function, can I use `async`?
like the below code.
```csharp
private async void HealthCheckAsync()
{
    while (healthEnabled)
    {
        await Task.Delay(TimeSpan.FromSeconds(healthIntervalSecond));

        try
        {
            await SendRequestAsync("/health", "{}");
        }
        catch (ObjectDisposedException)
        {
        }
    }
}
```

And if `LogEnabled` checked, logs will be output when `Health()` fails.
Should I still check the return value and output logs?

sdks/unity/AgonesSdk.cs
whisper0077: C# Awaiter Pattern requires that the return type of the `GetAwaiter` method has `GetResult()`,
and `AgonesAsyncOperationAwaiter` does not return a value in this case.

https://github.com/googleforgames/agones/pull/818/files#diff-f35409fa73f6901e4704a2951810b018R159

So I can not change this function name and do not need to return a value.

sdks/unity/AgonesSdk.cs
whisper0077: I can not explain in English well, so I paste the IL code of `SendRequestAsync` that is decompiled with [dnSpy](https://github.com/0xd4d/dnSpy).
```csharp
void IAsyncStateMachine.MoveNext()
{
    int num = this.<>1__state;
    AgonesSdk agonesSdk = this.<>4__this;
    bool result;
    try
    {
        AgonesSdk.AgonesAsyncOperationAwaiter agonesAsyncOperationAwaiter;
        if (num != 0)
        {
            agonesSdk.cancellationTokenSource.Token.ThrowIfCancellationRequested();
            this.<req>5__2 = new UnityWebRequest("http://localhost:59358" + this.api, this.method)
            {
                uploadHandler = new UploadHandlerRaw(Encoding.UTF8.GetBytes(this.json)),
                downloadHandler = new DownloadHandlerBuffer()
            };
            this.<req>5__2.SetRequestHeader("Content-Type", "application/json");
            agonesAsyncOperationAwaiter = new AgonesSdk.AgonesAsyncOperationWrapper(this.<req>5__2.SendWebRequest()).GetAwaiter();
            if (!agonesAsyncOperationAwaiter.IsCompleted)
            {
                this.<>1__state = 0;
                this.<>u__1 = agonesAsyncOperationAwaiter;
                // this methods means like OnCompleted(()=>{ this.MoveNext(); })
                this.<>t__builder.AwaitOnCompleted<AgonesSdk.AgonesAsyncOperationAwaiter, AgonesSdk.<SendRequestAsync>d__14>(ref agonesAsyncOperationAwaiter, ref this);
                return;
            }
        }
        else
        {
            agonesAsyncOperationAwaiter = (AgonesSdk.AgonesAsyncOperationAwaiter)this.<>u__1;
            this.<>u__1 = null;
            this.<>1__state = -1;
        }
        agonesAsyncOperationAwaiter.GetResult();
        bool flag = this.<req>5__2.responseCode == 200L;
        if (flag)
        {
            agonesSdk.Log("Agones SendRequest ok: " + this.api + " " + this.<req>5__2.downloadHandler.text);
        }
        else
        {
            agonesSdk.Log("Agones SendRequest failed: " + this.api + " " + this.<req>5__2.error);
        }
        result = flag;
    }
    catch (Exception exception)
    {
        this.<>1__state = -2;
        this.<>t__builder.SetException(exception);
        return;
    }
    this.<>1__state = -2;
    this.<>t__builder.SetResult(result);
}
```

I think that if the async operation completed immediately, this code works well. 
And `OnCompleted` is called on the main thread, so the `lock` statement is not necessary.

Please check it.

sdks/unity/AgonesSdk.cs
dzlier-gcp: Makes sense, just please add a method comment explaining such. Thanks

sdks/unity/AgonesSdk.cs
dzlier-gcp: Unity supports async and await from at least 2017 and beyond, so you can indeed use async function. The only thing to verify (because I'm not sure off hand) is how to properly trigger the function on a separate thread instead of holding up the main thread. It might be as simple as calling it from start as normal (without the `await` keyword).

markmandel:
Quick question - do we want to move the Unity example into our [examples repo](https://github.com/googleforgames/agones/tree/master/examples)? (Doesn't have to be in this PR either, can be a later one).

That's the only feedback I have. I'm happy to approve when @dzlier-gcp is happy with the code :+1:

whisper0077:
Thanks @markmandel 

Please re-review this @dzlier-gcp 
(but I can not explain well an unresolved conversation, please see the IL code in this unresolved conversation)

markmandel:
Nice! Just in time for the 0.11 Release Candidate going out tomorrow!

PR #826, opened 2019-06-12
snap

mikeroyal:
Includes  snapcraft.yaml to build [snap](https://snapcraft.io/).


markmandel:
So this is super interesting :+1: but I'm curious where you think this could be used?

Since you need a K8s cluster to run Agones, I'm not sure if having the binaries installed locally help with anything? But maybe I'm missing a use case here?

mikeroyal:
hi @markmandel , the nice thing about snaps is they are supported across 41 Linux distros. Snaps auto-update to the latest version and can also rollback to the previous version if something breaks.  [Helm](https://snapcraft.io/helm)is provided as a snap and [microk8s](https://microk8s.io/) for deploying Kubernetes cluster. I made the snap to see if it may be useful for you.

markmandel:
Snaps are nice - as an avid Linux user, I'm quite familiar.

I'm just trying to work out how this could ever be applied to a Kubernetes cluster. I mean you can't? :man_shrugging: (or can you?) :thinking: 

I guess we could do a snap for the local sidecar?

I'm just trying to work out the applicability here.

How would/did you see it working?

mikeroyal:
hi @markmandel, I think it is possible there would be two snaps involved. A snap for agones that does the local install of binaries for this repo and I would use a snap of either MicroK8s or MiniKube, which installs Kubernetes locally. I will more testing MicroK8s and MiniKube to see which one will work best for the Kubernetes cluster.

markmandel:
Interesting :+1: 

If we are looking at just the local binaries - we could download them from the release page, we don't have to compile them. Maybe that would be "snap install agones-local" ? WDYT?

Having a snap that was just "snap install agones-cluster" would be pretty awesome.

markmandel:
So I think for the sake of cleanup, I'm going to close this PR, as it's currently not doing anything we can use at this stage. But I love the ideas that came up in the discussion.

Unless you have major objections?

mikeroyal:
Hi @markmandel, I agree on closing for now and I will make a new pull request once I've made some more progress. I enjoyed the discussion we had.

PR #830, opened 2019-06-14
Changed AllocationEndpoint to array of endpoints

pooneh-m:


pkg/gameserverallocations/controller.go
```diff
@@ -398,26 +398,32 @@ func (c *Controller) allocateFromRemoteCluster(gsa v1alpha1.GameServerAllocation
...
-	}
-	if response.StatusCode >= 400 {
-		// For error responses return the body without deserializing to an object.
-		return nil, errors.New(string(data))
-	}
+		data, err := ioutil.ReadAll(response.Body)
+		if err != nil {
+			return nil, err
+		}
+		if response.StatusCode >= 500 && (i+1) < len(connectionInfo.AllocationEndpoints) {
```
aLekSer:
```suggestion
		if response.StatusCode >= 500 {
```
I think that this check is redundant for loop will check not continue if we are out of `range connectionInfo.AllocationEndpoints`.

aLekSer:
Thanks for the pull request. You should run `make lint` to check that go code is ok:
```
pkg/gameserverallocations/controller_test.go:911:18: string `remotecluster` has 3 occurrences, make it a constant (goconst)
		clusterName := "remotecluster"
```

pkg/gameserverallocations/controller.go
pooneh-m: Thanks. This condition says if the there is any other endpoint try sending the request with the new endpoint. Otherwise return the error. If I remove the condition, no error will be returned

test/e2e/gameserverallocation_test.go
```diff
@@ -85,9 +85,9 @@ func TestMultiClusterAllocationOnLocalCluster(t *testing.T) {
 					Priority: 1,
 					Weight:   100,
 					ConnectionInfo: multiclusterv1alpha1.ClusterConnectionInfo{
-						AllocationEndpoint: "localhost",
-						ClusterName:        "multicluster1",
-						SecretName:         "sec1",
+						AllocationEndpoints: []string{"localhost"},
```
ilkercelikyilmaz: Should we have test with multiple AllocationEndpoints as well?

pkg/gameserverallocations/controller.go
```diff
@@ -398,26 +398,32 @@ func (c *Controller) allocateFromRemoteCluster(gsa v1alpha1.GameServerAllocation
...
+			// If there is a server error try a different endpoint
+			continue
+		}
+		if response.StatusCode >= 400 {
```
markmandel: Curious - why stop at a 400, why not continue to the next endpoint like a 500?

pkg/gameserverallocations/controller.go
markmandel: Should we do a `Warn` level logging here, so we get some visibility into when things are failing?

pkg/gameserverallocations/controller.go
pooneh-m: 4xx means there is an error from client. We only retry with new endpoint if there is an error from server e.g. 5xx.

pkg/gameserverallocations/controller.go
pooneh-m: Good idea! Done.

test/e2e/gameserverallocation_test.go
pooneh-m: Thanks. I have added component testing for this case. Adding end-to-end testing requires fault injection. I'll add it in future, when the testing infrastructure is more mature to allow for that. For now we should be fine.

pkg/gameserverallocations/controller.go
```diff
@@ -398,26 +398,33 @@ func (c *Controller) allocateFromRemoteCluster(gsa v1alpha1.GameServerAllocation
...
-		// For error responses return the body without deserializing to an object.
-		return nil, errors.New(string(data))
-	}
+		data, err := ioutil.ReadAll(response.Body)
+		if err != nil {
+			return nil, err
+		}
+		if response.StatusCode >= 500 && (i+1) < len(connectionInfo.AllocationEndpoints) {
+			// If there is a server error try a different endpoint
+			c.baseLogger.Warnf("The request sent to %s failed: %s\n trying next endpoint", endpoint, err)
```
markmandel:
```suggestion
			c.baseLogger.WithError(err).WithField("endpoint", endpoint).Warn("The request sent failed, trying next endpoint")
```

WDYT?

pkg/gameserverallocations/controller.go
markmandel: My concern would be, that if after I set everything up, and all is well - if something weird happens to one of my clusters, such that they return a 404, that could potentially stop all/a large chunk of multicluster allocations.

pkg/gameserverallocations/controller.go
```diff
@@ -398,26 +398,33 @@ func (c *Controller) allocateFromRemoteCluster(gsa v1alpha1.GameServerAllocation
 	}
 
 	// TODO: Retry on transient error --> response.StatusCode >= 500
-	response, err := client.Post(connectionInfo.AllocationEndpoint, "application/json", bytes.NewBuffer(body))
-	if err != nil {
-		return nil, err
-	}
-	defer response.Body.Close() // nolint: errcheck
+	for i, endpoint := range connectionInfo.AllocationEndpoints {
```
markmandel: Should we be randomising the endpoints? (or this going to be policy driven in the future?)

pkg/gameserverallocations/controller.go
pooneh-m: 404 means the client is asking gameserverallocation for a resource that is not available. It is a client error and should be handled by client. If a cluster is out of gameservers to allocate, then it returns 500. Does it make sense?

markmandel: Oh yeah totally :+1: I'm just thinking through worst case scenarios.

Like someone accidentally puts a different service in place of where the allocator used to be setup :fearful: 

Not incredibly likely, but just thinking through it.

pkg/gameserverallocations/controller.go
pooneh-m: This is requesting based on the order of endpoints set in the policy. I wouldn't randomize it to have a consistent behavior, but I also see the value of choosing the endpoint that works first. It is good for now. 

For future, we can update policies endpoints based on working endpoints, but it needs more thinking. WDYT?

pkg/gameserverallocations/controller.go
pooneh-m: Thanks!

pkg/gameserverallocations/controller.go
markmandel: SGTM :+1:

PR #834, opened 2019-06-15
E2E test for Ready->Allocated->Ready

markmandel:
We said that using SDK.Ready() to go back to being ready once you were Allocated would work.

Wrote a test to prove it actually works! 

Also relevant for #660 so people can exit out of Reserved state if they need to manually do so.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -103,14 +103,14 @@ func (f *Framework) CreateGameServerAndWaitUntilReady(ns string, gs *stable.Game
 // WaitForGameServerState Waits untils the gameserver reach a given state before the timeout expires
 func (f *Framework) WaitForGameServerState(gs *stable.GameServer, state stable.GameServerState,
 	timeout time.Duration) (*stable.GameServer, error) {
-	var pollErr error
 	var readyGs *stable.GameServer
 
 	err := wait.PollImmediate(2*time.Second, timeout, func() (bool, error) {
-		readyGs, pollErr = f.AgonesClient.StableV1alpha1().GameServers(gs.Namespace).Get(gs.Name, metav1.GetOptions{})
+		var err error
+		readyGs, err = f.AgonesClient.StableV1alpha1().GameServers(gs.Namespace).Get(gs.Name, metav1.GetOptions{})
 
-		if pollErr != nil {
-			logrus.WithError(pollErr).Warn("error retrieving gameserver")
+		if err != nil {
+			logrus.WithError(err).Warn("error retrieving gameserver")
 			return false, nil
 		}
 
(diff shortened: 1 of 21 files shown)


test/e2e/framework/framework.go
```diff
@@ -120,7 +120,10 @@ func (f *Framework) WaitForGameServerState(gs *v1alpha1.GameServer, state v1alph
 		return false, nil
 	})
 	if err != nil {
-		return nil, errors.Wrapf(pollErr, "waiting for GameServer to be %v %v/%v: %v",
+		if pollErr != nil {
```
pooneh-m: Should the line 113 be updated to return the poll error instead? There is no where that PollImmediate is returning err.

markmandel: If we do that, then the retry will stop, and we could get race conditions. We may get an error because the GS isn't created yet (most likely option), but we want to wait for it to show up, and then wait for it to be `Ready`.

My change here allows for if the err != nil (which will happen on timeout), but the pollErr could also be nil, which means you end up with both a nil gameserver and a nil error, and very little idea what happened :smile: 

(This should probably have been it's own PR admittedly. I got a little lazy on this one)

pooneh-m: I see. Can you please add documentation on why err is replaced with pollErr?

markmandel: So I looked at this code again, and made it much simpler. I don't think we don't need the pollErr  juggling - we have a warn log at that point, so we have visibility, so I got rid of it.

I think this makes the code far cleaner, and easier to read. PTAL :+1:

test/e2e/framework/framework.go
```diff
@@ -103,14 +103,14 @@ func (f *Framework) CreateGameServerAndWaitUntilReady(ns string, gs *stable.Game
 // WaitForGameServerState Waits untils the gameserver reach a given state before the timeout expires
 func (f *Framework) WaitForGameServerState(gs *stable.GameServer, state stable.GameServerState,
 	timeout time.Duration) (*stable.GameServer, error) {
-	var pollErr error
 	var readyGs *stable.GameServer
 
 	err := wait.PollImmediate(2*time.Second, timeout, func() (bool, error) {
-		readyGs, pollErr = f.AgonesClient.StableV1alpha1().GameServers(gs.Namespace).Get(gs.Name, metav1.GetOptions{})
+		var err error
+		readyGs, err = f.AgonesClient.StableV1alpha1().GameServers(gs.Namespace).Get(gs.Name, metav1.GetOptions{})
```
pooneh-m: readyGs, err := ...
and remove var err error?

test/e2e/framework/framework.go
```diff
@@ -120,11 +120,9 @@ func (f *Framework) WaitForGameServerState(gs *stable.GameServer, state stable.G
...
 		return false, nil
 	})
-	if err != nil {
-		return nil, errors.Wrapf(pollErr, "waiting for GameServer to be %v %v/%v: %v",
-			state, gs.Namespace, gs.Name, err)
-	}
-	return readyGs, nil
+
+	return readyGs, errors.Wrapf(err, "waiting for GameServer to be %v %v/%v",
+		state, gs.Namespace, gs.Name)
```
pooneh-m: I don't think this would replace your previous change. pollErr is replaced with err which is local to the PollImmediate method. 
err := wait.PollImmediate --> this error will still be timeout as it is different that the err inside PollImmediate.

markmandel: Yeah, that was my intent with this change. I'm saying I don't think we need to do replace err with pollErr at all.

The original code was in place to make the pollErr error visible, in case it is the cause of the test failing -- but we already have logging in place to show that error, so it seems redundant to do do both logging _and_ bubble the error up.

If we remove the bubble up, the code is cleaner, and we don't sacrifice the ability to discover the reasons for issues.

At least - that is my intent :smile:

PR #839, opened 2019-06-18
Flaky: TestAllocator

markmandel:
Move creatRestClient into the polling code, as its possible the e2e test could use the old secret, if called early enough in the flow.


markmandel:
Well it passed, so hopefully this sticks!

Issue #672, opened 2019-03-26
SDK conformance harness


markmandel:
A way for developers who want to develop / bring their own SDK to know that their SDK implementation is compliant with what we need it to do.

Ideally, I think it would _not_ involve a full Kubernetes install, or e2e test, but likely could be run on it's own - maybe as a binary / docker container? Something simple.

We could then integrate this into our CI system to run against existing SDKs, and ensure that they always pass. Right now, this could fail for anything other than our Go sdk.

Would need to have a way to skip tests as well - since we often have SDKs that are partial implementations.

Open to suggestions on what this should look like?

aLekSer:
As I understand we should generate single GRPC/RestAPI client which would be calling SDK implementations one by one (each with unique hostnames). Nice approach which I found is to write golang declarative tests of the API:
[example](https://github.com/verdverm/pomopomo/tree/master/test/api)

aLekSer:
I make a draft proposal on how the system of harness test could look like:
https://docs.google.com/document/d/1Qjy-wXA1wsZB8PE_CP_GCgnyWktySm_ytZgRUZ2wtCo/edit?usp=sharing

markmandel:
Added a bunch of notes - PTAL

aLekSer:
I found an absence of new `Allocate()` method in Rust (fixed in a PR), CPP and Unreal Engine SDKs.
https://agones.dev/site/docs/guides/client-sdks/#allocate

markmandel:
Can we close this now? We have the harness after all.

aLekSer:
Yes, we can close this PR now.

markmandel:
Closing!

PR #845, opened 2019-06-19
Initial version of SDK conformance Testing framework for Golang, NodeJS and others

aLekSer:
PoC for SDK conformance testing.
Proposal Document could be found [here](https://docs.google.com/document/d/1Qjy-wXA1wsZB8PE_CP_GCgnyWktySm_ytZgRUZ2wtCo/edit?usp=sharing).
Test engine performs two tasks: code generation for Golang, Node SDKs and verification that SDK sends and sidecar receives all GRPC messages as intended.
YAML file with test definition in Formal language -> translates and concatenates from excerpts in SDK target language.
Code generator use code templates from sdk_client use test Case
scenarios from harness folder and produce go file which would run the
scenario.
`make build-conformance-tests` - build docker image and binaries of sdk-server (sidecar) and `test_engine`
`make run-conformance-tests` - build binaries and perform verify step (also `run-conformance-jstests`):
>`./sdk --verify=true --sdk=nodejs`  - runs sdk-server and binaries of target SDK client generated before.
Added --timeout for sdk-server (Sidecar) local run.

For #672.

### DIFF
--- a/build/build-sdk-conformance/Dockerfile
+++ b/build/build-sdk-conformance/Dockerfile
@@ -1,0 +1,36 @@
+# Copyright 2019 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+FROM golang:latest
+
+RUN apt-get update && \
+    apt-get install -y wget jq && \
+    apt-get clean
+
+# install go
+WORKDIR /usr/local
+ENV GO_VERSION=1.12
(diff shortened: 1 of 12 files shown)


markmandel:
So I'm digging through this, and while it seems cool - I'm wondering if it's over-engineered?

In the model in my head, there would be a Docker container, with a builder pattern for a language that gets built and run on each test for each SDK.

Alongside it we could run the standard local sdk server, with some variation of:
`docker run --rm gcr.io/agones-images/sdk-server:0.11.0 --test-mode=true, --test-list=init,get,watch,allocate`

The test mode expects those tests to pass (they have documented requirements) within,1 minute (or whatever time period - 30s may be enough)

Then we don't need code generation, or yaml files, or a go test hardness, or most of the complicated pieces.

WDYT?

aLekSer:
Thanks for the review, Mark.
Yes, my solution seems to be more complicated than it is needed. At least I understood SDK a lot better. 
One thing which we could not control when we just have the acceptance test on sdk-server side is the parsing the results on SDK client side and need to change more on sidecar side. But for the burden on SDK developer seems to be the same in both options of realisation.
I am going to develop proposed solution by you. Let's leave this PR for a while till I prepare new one.

aLekSer:
This one was superseded by #848 as it would be easier to maintain.

aLekSer:
Remove this PR as we switched to another version without code generation.

PR #847, opened 2019-06-21
Update CPP Example Readme and yaml files

markmandel:
Update incorrect instructions for building the examples, and
also update fleet.yaml and gameserver.yaml to the correct image
tag.


roberthbailey:
lgtm

PR #848, opened 2019-06-25
SDK Conformance testing

aLekSer:
Add test mode for SDK-server.
Local SDK server (Sidecar) now could record client requests.
Added test clients and make targets to test Golang and NodeJS SDKs.

For #672.

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -34,6 +34,7 @@ Table of Contents
         * [make build-images](#make-build-images)
         * [make build-sdks](#make-build-sdks)
         * [make build-sdk-cpp](#make-build-sdk-cpp)
+        * [make run-sdk-conformance-tests](#make-run-sdk-conformance-tests)
         * [make test](#make-test)
         * [make push](#make-push)
         * [make install](#make-install)
@@ -415,6 +416,10 @@ Build all the sdks required for Agones
 
 #### `make build-sdk-cpp`
 Build the cpp sdk static and dynamic libraries (linux libraries only)
+
+#### `make run-sdk-conformance-tests`
+Run SDK conformance test.
+Run SDK server (sidecar) in test mode (which would record all GRPC requests) versus all SDK test clients which should generate those requests. All methods are verified.
 
 #### `make test`
 Run the linter and tests
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -108,3 +108,25 @@ ensure-build-sdk-image-base:
 # create the build image sdk if it doesn't exist
 ensure-build-sdk-image:
 	$(MAKE) build-build-sdk-image SDK_FOLDER=$(SDK_FOLDER)
+
+# Run SDK conformance Sidecar server in docker in order to run
+# SDK client test against it. Useful for test development
+SECONDS ?= 30
+run-sdk-conformance-local:
+	docker run -e "ADDRESS=" -p 59357:59357 \
+	 -e "TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" -e "TIMEOUT=$(SECONDS)" $(sidecar_tag)
+
+# Run SDK conformance test for a specific SDK_FOLDER
+SECONDS ?= 30
+run-sdk-conformance-test: SDK_FOLDER=$(SDK_FOLDER)
+run-sdk-conformance-test: ensure-build-sdk-image
+	sleep 2s && DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
+	docker run -p 59357:59357 -e "ADDRESS=" \
+	 -e "TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" -e "TIMEOUT=$(SECONDS)" --net=host  $(sidecar_tag)
+
+# Run a conformance test for all SDKs supported
+run-sdk-conformance-tests: ensure-agones-sdk-image
+	$(MAKE) run-sdk-command SDK_FOLDER=node COMMAND=build
+	$(MAKE) run-sdk-conformance-test SDK_FOLDER=node
+	$(MAKE) run-sdk-conformance-test SDK_FOLDER=go
+	$(MAKE) run-sdk-command SDK_FOLDER=node COMMAND=clean
--- a/cmd/sdk-server/main.go
+++ b/cmd/sdk-server/main.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"agones.dev/agones/pkg"
 	"agones.dev/agones/pkg/client/clientset/versioned"
@@ -50,7 +51,9 @@ const (
 	// Flags (that can also be env vars)
 	localFlag   = "local"
 	fileFlag    = "file"
+	testFlag    = "test"
 	addressFlag = "address"
+	timeoutFlag = "timeout"
 )
 
 var (
@@ -69,6 +72,7 @@ func main() {
 		logger.WithField("grpcPort", grpcPort).WithField("Address", ctlConf.Address).Fatalf("Could not listen on grpcPort")
 	}
 	stop := signals.NewStopChannel()
+	timedStop := make(chan struct{})
 	grpcServer := grpc.NewServer()
 	// don't graceful stop, because if we get a kill signal
 	// then the gameserver is being shut down, and we no longer
@@ -85,9 +89,30 @@ func main() {
 	defer cancel()
 
 	if ctlConf.IsLocal {
-		err = registerLocal(grpcServer, ctlConf)
+		localSDK, err := registerLocal(grpcServer, ctlConf)
 		if err != nil {
 			logger.WithError(err).Fatal("Could not start local sdk server")
+		}
+		defer localSDK.Close()
+
+		if ctlConf.Timeout != 0 {
+			go func() {
+				time.Sleep(time.Duration(ctlConf.Timeout) * time.Second)
+				close(timedStop)
+			}()
+		}
+	} else if ctlConf.Test != "" {
+		localSDK, err := registerTestSdkServer(grpcServer, ctlConf)
+		if err != nil {
+			logger.WithError(err).Fatal("Could not start test sdk server")
+		}
+		defer localSDK.Close()
+
+		if ctlConf.Timeout != 0 {
+			go func() {
+				time.Sleep(time.Duration(ctlConf.Timeout) * time.Second)
+				close(timedStop)
+			}()
 		}
 	} else {
 		var config *rest.Config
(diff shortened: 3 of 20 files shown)


build/README.md
```diff
@@ -34,6 +34,7 @@ Table of Contents
         * [make build-images](#make-build-images)
         * [make build-sdks](#make-build-sdks)
         * [make build-sdk-cpp](#make-build-sdk-cpp)
+        * [make run-conformance-test](#make-run-conformance-test)
```
markmandel:
```suggestion
        * [make run-sdk-conformance-tests](#make-sdk-run-conformance-tests)
```

WDYT?

build/includes/sdk.mk
```diff
@@ -108,3 +109,22 @@ ensure-build-sdk-image-base:
...
+
+# Run SDK conformance test for a specific SDK_FOLDER
+SECONDS ?= 30
+run-sdk-conformance-test:
+	DOCKER_PARAMETERS="--net host" COMMAND=sdktest $(MAKE) run-sdk-command & \
+	docker run -p 59357:59357 -e "ADDRESS=" -e "LOCAL=true" \
+	 -e "TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" -e "TIMEOUT=$(SECONDS)" --net=host  $(sidecar_tag)
+
+# Run a conformance test for all SDKs supported
+run-conformance-test:
```
markmandel:
```suggestion
run-sdk-conformance-tests:
```

?? WDYT? also looks similar to other ones.

cmd/sdk-server/main.go
```diff
@@ -85,10 +89,22 @@ func main() {
 	defer cancel()
 
 	if ctlConf.IsLocal {
-		err = registerLocal(grpcServer, ctlConf)
+		localSdk, err := registerLocal(grpcServer, ctlConf)
```
markmandel: Random thought - should this be integrated with the local mode, or should it be it's own --test mode?

(and maybe be it's own separate implementation, rather than mixed in with the local implementation?)

aLekSer: Yes, I will try to separate test mode but them looks identical for now.

markmandel: If it doesn't work out, it doesn't work out - was just wondering if it would make things cleaner.

aLekSer: I think test mode should have a bit different constructor but other parts would look the same. So I am going to reuse Local SDK server code, not to duplicate code.

markmandel: Sounds good! We can also split at a later date if we decide there is need.

aLekSer:
@markmandel  Regarding functionality, here we just enumerate all functions and check that queries occur. Do we need something more or that is fine for conformance test?

markmandel:
> @markmandel Regarding functionality, here we just enumerate all functions and check that queries occur. Do we need something more or that is fine for conformance test?

That a good question - maybe we should be setting specific values in things like label and annotation?

Could we return some kind of value in `GameServer()` and expect that value to be passed back to the value of a hard-coded label and annotation key?

I'm not sure how to test Watch. Maybe something similar?

I think we can have a prescribed test path for these things, that people have to conform to.

WDYT?

aLekSer:
Thanks for sharing ideas on that, that would definitely work. That way we would just check that client sends the data in a correct format. More descriptive test scenarios should involve real cluster anyway.

aLekSer:
I have added check that SDK client received get `GameServer()` results `CreationTimestamp` would be checked in `setLabel()` value on the sidecar side. The similar thing could be done to test Watch results of them could be sent in `setAnnotation()`.
Note what I have found is that we could pass any data to JS SDK for label value. I think we should limit it to be string only.
```
	async setLabel(key, value) {
		const request = new messages.KeyValue();
		request.setKey(key);
		request.setValue(value);
```
I mean we could add some check like that:
```
if (typeof value == 'string' || value instanceof String) 
```

markmandel:
This looks really good! :smile: keen to take it for a spin!

Couple of questions:

- This isn't currently running in our cloudbuild.yaml? should it?
- Your approach for WATCH seems reasonable. Did you want to do that in this PR?

aLekSer:
Thanks for the response )

>This isn't currently running in our cloudbuild.yaml? should it?

We could add this target to `cloudbuild.yaml`. The only problem could be missing sdk-server image.
I could inject this step and see how it goes.

>Your approach for WATCH seems reasonable. Did you want to do that in this PR?

Was not able to receive those watch messages in JS, so added only `GetGameserver()` results feedback.

markmandel:
> > Your approach for WATCH seems reasonable. Did you want to do that in this PR?
> 
> Was not able to receive those watch messages in JS, so added only `GetGameserver()` results feedback.

So can we skip it for the node API? (looks like we have that capability, since we have the list), then we can do the test in Go (did we do this? I might have missed it)

aLekSer:
Yes, we can skip Node Watch check. Current set of checks:
`"TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" `
 I need to add watch feedback for both SDKs soon.

By the way, investigating how to create similar solution for other SDKs in form of docker image + bash script. For instance, started to use Unreal Editor plugin recently, not sure how Dockerfile would look like for `Unreal` case. Unreal Editor itself is too big - over 6 GB - and needs authentication to install.

markmandel:
Yes, and there are also EULA concerns with both of the platforms as well. We may need to leverage some internal resources if we can even do automated build steps with Unity/Unreal.

aLekSer:
Added SDK to cloud build. Step is fine: + 1 minute execution for 2 SDKs.
https://console.cloud.google.com/cloud-build/builds/f95dff85-c440-4365-a586-0d1b7c1e145c?project=agones-images
 `Watch` on client side now should get random `GS.ObjectMeta.UID` and return as Annotation.

aLekSer:
Now E2E test is failing:
```
Step #17: --- FAIL: TestGameServerPassthroughPort (305.73s)
Step #17: gameserver_test.go:356: 
Step #17: Error Trace:	gameserver_test.go:356
Step #17: Error: Received unexpected error:
Step #17: waiting for {udp-server [{ Passthrough 0 0 }] {false 0 0 0} {{ 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] nil [] } {[] [] [{udp-server gcr.io/agones-images/udp-server:0.11 [] [] [] [] [{PASSTHROUGH TRUE nil}] {map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>} BinarySI}] map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>} BinarySI}]} [] [] nil nil nil IfNotPresent nil false false false}] <nil> <nil> map[] <nil> false false false <nil> nil [] nil [] [] <nil> nil []}}} GameServer instance readiness timed out (): waiting for GameServer to be Ready default/udp-server6xpjj: timed out waiting for the condition
Step #17: Test: TestGameServerPassthroughPort
Step #17: gameserver_test.go:357: 
Step #17: Error Trace:	gameserver_test.go:357
Step #17: Error: Could not get a GameServer ready
```

markmandel:
This is a flaky test (I just dropped some details in chat), retry, it should pass.

I'm not quite sure what's causing it.

markmandel:
Just checking - this good for another review?

aLekSer:
@markmandel  Yes, this PR is ready for review. Added what you have requested. Also added random UID generation for local SDK server. `Watch` request now verified from the client test code.

markmandel:
So I just ran `make run-sdk-conformance-tests`, and the first time I ran it (no image cache), it failed.

Is it possible there local sidecar timedout while the images where building?

```
+ npm install
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN nodejs No description
npm WARN nodejs No repository field.
npm WARN nodejs No license field.

added 1 package from 1 contributor and audited 153 packages in 1.98s
found 0 vulnerabilities

+ npm run testSDK

> @ testSDK /go/src/agones.dev/agones/test/sdk/nodejs
> node ./testSDKClient.js

(node:30) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
{ Error: 14 UNAVAILABLE: Connect Failed
    at Object.exports.createStatusError (/go/src/agones.dev/agones/sdks/nodejs/node_modules/grpc/src/common.js:91:15)
    at Object.onReceiveStatus (/go/src/agones.dev/agones/sdks/nodejs/node_modules/grpc/src/client_interceptors.js:1204:28)
    at InterceptingListener._callNext (/go/src/agones.dev/agones/sdks/nodejs/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/go/src/agones.dev/agones/sdks/nodejs/node_modules/grpc/src/client_interceptors.js:618:8)
    at callback (/go/src/agones.dev/agones/sdks/nodejs/node_modules/grpc/src/client_interceptors.js:845:24)
  code: 14,
  metadata: Metadata { _internal_repr: {} },
  details: 'Connect Failed' }
make[2]: Leaving directory '/home/markmandel/workspace/agones/src/agones.dev/agones/build'
```

Running it a second time worked fine though for nodejs, but not for go:

```
make[4]: Leaving directory '/home/markmandel/workspace/agones/src/agones.dev/agones/build'
make[3]: Leaving directory '/home/markmandel/workspace/agones/src/agones.dev/agones/build'
+ /root/sdktest.sh
+ GO111MODULE=off
+ cd /go/src/agones.dev/agones/test/sdk/go
+ go run sdk-client-test.go
sdk-client-test.go:28: Client is starting
sdk-client-test.go:32: Could not connect to sdk: could not connect to localhost:59357: context deadline exceeded
exit status 1
includes/sdk.mk:76: recipe for target 'run-sdk-command' failed
make[2]: *** [run-sdk-command] Error 1
make[2]: Leaving directory '/home/markmandel/workspace/agones/src/agones.dev/agones/build'
```

markmandel:
I see you just pushed up some changes, would that fix the above?

aLekSer:
Latest changes was just rebasing master, let me check.

aLekSer:
Fail in E2E unrelated to the change:
```
Step #17: time="2019-07-08 14:24:30.950" level=info msg="fleet scale-fleet-1-g7fz8 has 10/0 ready replicas"
Step #17: time="2019-07-08 14:24:31.310" level=info msg="gameserver 3/10 not found" 
Step #17: time="2019-07-08 14:24:31.394" level=info msg="fleet scale-fleet-0-w74lv has 10/0 ready replicas" 
Step #17: --- FAIL: TestAutoscalerStressCreate (90.39s)
Step #17: fleetautoscaler_test.go:169: 
Step #17: Error Trace:	fleetautoscaler_test.go:169 
Step #17: fleetautoscaler_test.go:191 Step #17: Error: Should be true 
Step #17: Test: TestAutoscalerStressCreate 
Step #17: Messages: FleetAutoscaler created even if the parameters are NOT valid: 4 3 5 Step #17: fleetautoscaler_test.go:169:  
Step #17: Error Trace:	fleetautoscaler_test.go:169
```

aLekSer:
I cannot reproduce E2E test fail on another GKE cluster. For some reason cloud-build keeps failing at the same point.

aLekSer:
Thanks Mark E2E tests should now be ok, after `helm delete`

build/includes/sdk.mk
```diff
@@ -77,7 +78,7 @@ run-sdk-command:
 		cd - ; \
 		$(MAKE) ensure-build-sdk-image SDK_FOLDER=$(SDK_FOLDER) ; \
 		docker run --rm $(common_mounts) -e "VERSION=$(VERSION)" \
-			$(DOCKER_RUN_ARGS) $(build_sdk_prefix)$(SDK_FOLDER):$(build_version) $(COMMAND) ; \
+			$(DOCKER_RUN_ARGS) $(DOCKER_PARAMETERS) $(build_sdk_prefix)$(SDK_FOLDER):$(build_version) $(COMMAND) ; \
```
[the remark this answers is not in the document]
aLekSer: That is really strange I have added a clean step it would run as last `make run-sdk-conformance-tests` step.
I was checking that all is right, with `docker rmi ID`

aLekSer:
That is really strange I have added a clean step it would run as last make run-sdk-conformance-tests step.
I was checking that all is right, with docker rmi ID

aLekSer:
The issue with npm install could be caused by "Random issues", refer to [npm docs](https://docs.npmjs.com/common-errors#random-errors) so I added additional step which will run `npm install` again after cleaning node_modules in `test` and `sdks/nodejs` folders as well as cache.
So now we have additional build step as precondition before test run. Also we need those build.sh command for rust, because it could take more than 30 seconds to finish.

aLekSer:
The problem caused by invoking at the exact same time:
`make run-sdk-command-node COMMAND=test`
and
`make rund-sdk-command-node COMMAND=build`
Will add this dependency to a `cloudbuild.yaml`

aLekSer:
Now my PR is ready for review, but CLA bot is not so responsive.

PR #803, opened 2019-06-03
Cpp prerequisities cmake

dsazonoff:

Issue #713, opened 2019-04-16
C++ SDK should follow Google Style 


devjgm:
The [C++ SDK](https://github.com/GoogleCloudPlatform/agones/tree/master/sdks/cpp) should following [Google's C++ style guide](https://google.github.io/styleguide/cppguide.html) and generally recommended coding conventions, such as https://abseil.io/tips/

Concrete next steps that I recommend:

1. Fix all formatting of .cc and .h files with `clang-format -style=Google ...`. Ideally, have a CI build that verifies new PRs are properly formatted.

2. Misc. C++ code cleanups, like more use of `unique_ptr` (I think there are several leaks of `grpc::ClientContext` currently), remove unnecessary std::string copies, etc.

If changes like this sound acceptable to the project owners, I might be able to either (a) make these changes, or (b) review these changes if someone else wants to make them.

roberthbailey:
/cc @dsazonoff 

I'm partial to using Google's C++ style guide, but I would be happy if we specified any style guide and stuck with it (using one even if it isn't perfect is better than not using one). And if we want a style guide, picking one that has auto-formatting seems like a big win.

roberthbailey:
As per the second point, the [C++ SDK docs](https://agones.dev/site/docs/guides/client-sdks/cpp/#usage) say:

The C++ SDK is specifically designed to be as simple as possible, and deliberately doesn’t include any kind of singleton management, or threading/asynchronous processing to allow developers to manage these aspects as they deem appropriate for their system.

but I think it would be nice to not leak resources even if we don't provide sophisticated threading support.

markmandel:
I personally don't have strong opinions here regarding C++, I'm happy to defer to those that do and have actual real experience with C++ (which I definitely do not have!)

devjgm:
@roberthbailey thanks. I did see the C++ SDK docs. Keeping the implementation simple SGTM. I'm not suggesting anything to the contrary. I suggest fixing the memory leaks, but also there are generally accepted C++ best practices (e.g., http://abseil.io/tips, and the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md)) that I think the code should adhere to as well.

roberthbailey:
Adhering to best practices sgtm. I just want to make sure that we don't force any particular usage pattern upon the client. 

@dsazonoff is in a different timezone so I think we should give him a chance to chime in before settling anything since he's been working on the C++ SDK.

[...]

PR #855, opened 2019-06-26
Cpp clang-format

dsazonoff:
PR for #713 issue.
Added support of clang-format as a pre-build step (optional, useful for SDK contributors)
Added clang-format code style verification as a build step. Build will fail if SDK code is not following google code style (not applied for generated files).
Waiting for #803 to be merged in...

### DIFF
--- a/build/build-sdk-images/cpp/build.sh
+++ b/build/build-sdk-images/cpp/build.sh
@@ -17,4 +17,4 @@
 set -ex
 
 cd ./sdks/cpp
-make build install archive VERSION=$VERSION
+make build verify install archive VERSION=$VERSION
--- a/build/build-sdk-images/cpp/Dockerfile
+++ b/build/build-sdk-images/cpp/Dockerfile
@@ -15,25 +15,16 @@ ARG BASE_IMAGE=agones-build-sdk-base:latest
 FROM $BASE_IMAGE
 
 RUN apt-get update && \
-    apt-get install -y zip wget && \
+    apt-get install -y zip wget clang-format && \
     apt-get clean
 
-ADD https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh /cmake-3.14.1-Linux-x86_64.sh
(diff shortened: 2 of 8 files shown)


dsazonoff:
Example of output (#900 ):
```
CMakeLists.txt:72 (message)
Call Stack (most recent call first):
clang-format code style check failed for: sdks/cpp/src/agones/sdk.cc
```

@markmandel here is PR for #713 issue.

roberthbailey:
I'm not an expert on cmake (so if anyone is I'd appreciate their review) but I don't see any obvious issues with this change, and from #900 it appears to be validating that formatting has been successfully applied to the source files.

dsazonoff:
@sylvainduchesne, could you do a code review, please?

build/build-sdk-images/cpp/build.sh
```diff
@@ -17,4 +17,4 @@
 set -ex
 
 cd ./sdks/cpp
-make build install archive VERSION=$VERSION
\ No newline at end of file
+make build verify install archive VERSION=$VERSION
```
markmandel: Should verify be part of a new test.sh step?
Context: https://github.com/googleforgames/agones/tree/master/build/build-sdk-images

WDYT?

roberthbailey: As someone who was trying to run sdk tests yesterday, I think it's better as part of the build step. I think if a test step just runs formatting checks and not any actual tests, it'll make people _think_ that tests are running when in fact they aren't.

markmandel: Devil's advocate.

If we had unit tests, would you want verify to run in the same test.sh target as the unit tests?

roberthbailey: If we had tests I'd be fine with it in either the build or test target. You could argue the build target is still the right place, and you always have to build before testing anyway, so even if you just ran the test target you'd validate formatting. Or you could argue that checking formatting is a 'test' against the correctness of the code's layout, so it should be a test. 

But, right now we don't have tests, so I think it should be in the build target.

markmandel: Sounds good to me - just wanted to talk it through :+1: 

I'm going to approve this!

PR #863, opened 2019-07-02
Flaky: TestGameServerPassthroughPort

markmandel:
TestGameServerPassthroughPort exposed a pretty nasty race condition in the sdk sidecar server, in which a request for GameServer information could come in just before it had cached its details, resulting in a
"Not found" response.

This uses a WaitGroup to requests for GameServer information until such time as it is initially loaded.


pkg/sdkserver/sdkserver.go
```diff
@@ -70,6 +70,7 @@ var (
 
 // SDKServer is a gRPC server, that is meant to be a sidecar
 // for a GameServer that will update the game server status on SDK requests
+// nolint: maligned
```
Kuqd: hehe

Issue #491, opened 2019-01-17
Write a guide for setting up Agones with taints and tolerations


markmandel:
We should write a guide for setting up Agones with taints and toleration, so that Agones (and likely prom + grafana, if you are running them) are on their own machine(s), and gameservers run on their on machines -- and never the two shall meet.

This is how people should run it in production, so we should make it easy with a tool.

Stretch Goal: Make our `build` tooling do this by default on GKE.

markmandel:
Flip side - our `build` tooling already does this, and it's part of our helm template. But we just need to write the documentation.

@jkowalski I've never played with this stuff, from memory, you did this work, yes?

markmandel:
We have this section here:
https://agones.dev/site/docs/installation/helm/#installing-the-chart

Does that mean we can close this ticket?

@jkowalski you wrote this section - wdyt?

markmandel:
Closing this now. We can reopen / create a new ticket if this comes up again.

roberthbailey:
I've moved the instructions to the installation page along with the example of how to do it on a GKE cluster. I don't have an easy way to test EKS / AKS and the directions don't apply to minikube which has only a single node. 

How do folks feel about closing this (or moving it out of the 1.0 milestone)? Or if anyone can provide similar updates to configure EKS / AKS prior to the next release that'd be even better.

markmandel:
I was thinking about this - I think I would feel comfortable closing this, if the section 

> By default Agones prefers to be scheduled on nodes labeled with stable.agones.dev/agones-system=true and tolerates the node taint stable.agones.dev/agones-system=true:NoExecute. If no dedicated nodes are available, Agones will run on regular nodes, but that’s not recommended for production use.

Was hosted up in the install section (or somewhere near it) at the top, rather than buried in the GKE section.

At least then we are explaining what needs to be done for all installations, but not providing examples (yet).

WDYT?

roberthbailey:
That sounds like a great idea. I'll send a PR over to move it up and slightly reword the GKE section to close this out.

PR #870, opened 2019-07-03
Add instructions about taints and tolerations to the install instructions

roberthbailey:
And a specific example for GKE. Remove some of the duplicate verbiage from the Helm instructions.

Ref: https://github.com/googleforgames/agones/issues/491

### DIFF
--- a/site/content/en/docs/Installation/_index.md
+++ b/site/content/en/docs/Installation/_index.md
@@ -8,6 +8,10 @@ description: >
 
 In this quickstart, we will create a Kubernetes cluster, and populate it with the resource types that power Agones.
 
+_When running in production, Agones should be scheduled on a dedicated pool of nodes, distinct from where Game Servers
+ are scheduled for better isolation and resiliency. See the sections below for instructions on how to do this in your
+ preferred environment._
+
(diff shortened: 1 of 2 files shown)


roberthbailey:
/cc @markmandel 

I'm not sure how to verify if I made the relative link correctly locally.

markmandel:
Looks like updates need to be made to for the website: https://github.com/googleforgames/agones/blob/master/build/README.md

`make site-server` will start the local server, so you can see what is happening.

If you want to run the site tests `make site-test` to run them locally.

markmandel:
We also have some docs instructions here: https://agones.dev/site/docs/contribute/

Which is actually a better spot to be updated.

roberthbailey:
> $ make site-test
xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun

roberthbailey:
Seems like I need to look into that a bit more....

roberthbailey:
Looks like I somehow borked my xcode install, which I'm in the process of fixing. It should be easier to verify the next one.

PR #829, opened 2019-06-13
Move FleetAutoscaling to autoscaling.agones.dev group

markmandel:
This is a pure refactor, to move FleetAutoscaling to the group outlined in #703, where it will stay.

Did this one first, since it seemed like the easiest one to tackle.

Issue #703, opened 2019-04-12
Proposal: Split up the api group stable.agones.dev


markmandel:
This has been rolling around in my head for a while, so wanted to throw it out there.

Basically watching the changes coming to allocation (#682), and how that will _have_ to be in a different group, and then the multicluster work (#698) feels like it should be separated in someway from everything else - I'm wondering if having `stable.agones.dev` as the core group was a less than ideal decision.

If we made this change, I a sacrificial draft of grouping could be:

core.agones.dev:
 - GameServers
 - Fleets
 - GameServerSets

scaling.agones.dev
 - FleetAutoscaling

allocation.agones.dev
 - GameServerAllocation

multicluster.agones.dev
 - AllocationPolicy (or whatever it ends up being called)

(FleetAllocation would probably have to stay in core, until it goes away, just because you can't mix CRDs and APIServices)

Looking at Kubernetes - they also do something similar.

This is obviously a breaking change - so wanted to run it past everyone, see how people felt.

roberthbailey:
There is a great comment [here](https://github.com/kubernetes-sigs/cluster-api/pull/645#issuecomment-448655658) from Clayton about what an apigroup should encompass (I'll inline the important bits so people don't get side tracked by that issue):

> historically, a Kube API group is intended to roughly encompass a lifecycle. That lifecycle has sometimes been SIG defined (autoscaling) but has also been related to orthogonal features (batch and v2beta1 being created for cronjob specifically, or the split between the various parts of what sig-auth controls). When boundaries can exist between lifecycles, we strongly recommend those be created. Federation v1 is a cautionary tale around coupling too closely between "high level API" (the workloads) and the low level api (the workload APIs on each cluster) - we erred a bit too far on putting everything all together, while federation v2 has focused on a much lower level primitive succeeding.

I think that the lifecycles for the pieces you listed above will likely be separable. The "core" bits will probably stabilize and not change much as we iterate on allocations and especially multicluster support. 

Is there historical context for why `stable.agones.dev` was initially chosen vs just `agones.dev`? Was the idea to have both stable and unstable api groups? Why different groups instead of just different api versions within the same group?

[...]

PR #874, opened 2019-07-05
Promote autoscaling to v1

roberthbailey:
This is part of https://github.com/googleforgames/agones/issues/703. 

I didn't update [agones_crd_api_reference.html](https://github.com/googleforgames/agones/blob/master/site/content/en/docs/Reference/agones_crd_api_reference.html) because I was hoping that there is an automated way to do it that I haven't yet found (although the presence of `{{% feature expiryVersion="0.10.0" %}}` at the top of the file makes me think I may have unfounded hope...). 

I also fixed some import naming as suggested by @pooneh-m in https://github.com/googleforgames/agones/pull/829 to try and clean up some tech debt while I was at it. 

/assign @markmandel 

/label kind/breaking


aLekSer:
Hello @roberthbailey let me check what have happened with `agones_crd_api_reference.html`.

aLekSer:
Verified on your branch - all is fine with `make gen-api-docs`. However I think we could remove outdated versions with `expiryVersion` < current version - 1 minor version.

roberthbailey:
Thanks @aLekSer! Once #877 goes in I will re-generate API docs. In the mean time I'll try to get all of the presubmit tests passing.

site/content/en/docs/Reference/agones_crd_api_reference.html
```diff
@@ -7319,7 +7320,7 @@ <h3 id="MultiClusterSetting">MultiClusterSetting
 <hr/>
 <h2 id="autoscaling.agones.dev">autoscaling.agones.dev</h2>
 <p>
-<p>Package v1alpha1 is the v1alpha1 version of the API.</p>
+<p>Package v1 is the v1 version of the API.</p>
```
roberthbailey: Should I wrap these changes in the `{{% feature ... %}}` blocks like I did for the other documentation updates? It's unclear since this change was auto-generated.

aLekSer: No, they should be wrapped automatically. I mean we wrap the whole block, and have current and previous version as a whole piece so that's easier to compare. I have commented with the algorithm in my PR, fixed gen_api_docs a bit.

roberthbailey:
@markmandel - it looks like the e2e test uses `helm upgrade` which as expected fails since I've changed api groups. Doing a fresh helm install with `helm install --name autoscaling-v1 --namespace agones-system agones --set agones.image.tag=0.12.0-097a5ea,agones.image.registry=gcr.io/agones-images` on a new cluster worked for me, so I'm guessing that someone with admin access to the `agones-images` will need to modify the cluster used by cloud build for e2e testing to delete the existing helm install so a fresh one will succeed. 

It looks like in https://github.com/googleforgames/agones/pull/829 you did a `helm delete agones -purge` at one point, but it isn't clear if that was necessary for tests to pass or not.

site/content/en/docs/Reference/agones_crd_api_reference.html
roberthbailey: I've rebased on top of your PR and will re-generate this file.

roberthbailey: I re-ran `make gen-api-docs` after rebasing on top of your PR but this file doesn't look any different.... So my original question still stands about whether I should be wrapping this in the `{{% feature ... %}}` blocks.

markmandel: I just tested this via `make site-server` and then `make site-server ENV=RELEASE_VERSION=0.11.0` - and the former shows the v1 version, and the latter shows the v1alpha1 version - so I'm not 100% sure what the issue here is?

Issue #875, opened 2019-07-05
Add Java SDK


felixklauke:
It would be awesome to add a Java SDK for Java based Games like Minecraft.

markmandel:
We should probably do this at some point.

PR #876, opened 2019-07-05
Add Java SDK.

FelixKlauke:
Add a Java SDK like mentioned in #875


sdks/java/src/main/java/com/google/agones/java/AgonesJavaSDK.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.agones.java.SDKGrpc.SDKBlockingStub;
+import com.google.agones.java.SDKGrpc.SDKStub;
+import com.google.agones.java.observer.EmptyObserver;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.stub.StreamObserver;
+
+public class AgonesJavaSDK {
+
+    private static final String DEFAULT_HOST = "127.0.0.1";
```
aLekSer: As I see in other SDKs we are using localhost here. Example from Node SDK:
```
		this.client = new services.SDKClient('localhost:59357', grpc.credentials.createInsecure());
```

FelixKlauke: Changed.

sdks/java/src/main/java/com/google/agones/java/observer/EmptyObserver.java
```diff
@@ -0,0 +1,21 @@
+package com.google.agones.java.observer;
```
aLekSer: Please add license lines here.

sdks/java/src/test/java/com/google/agones/java/AgonesJavaSDKTest.java
```diff
@@ -0,0 +1,14 @@
+package com.google.agones.java;
```
aLekSer: Also lacks license text.

FelixKlauke: Done.

sdks/java/src/main/java/com/google/agones/java/observer/EmptyObserver.java
FelixKlauke: Done.

FelixKlauke:
As visible in the newest commit message, I also applied the Google Java Style.

sdks/java/src/main/proto/agones/sdk.proto
```diff
@@ -0,0 +1,141 @@
+// Copyright 2017 Google LLC All Rights Reserved.
```
markmandel: I don't suppose there is anything we can do here to not have a copy of the file here -- I'm concerned with remembering to keep them in sync?

Maybe we can use a symlink or copy it on every gen.sh? (and ignore it otherwise?)

FelixKlauke: I managed to use a symlink. Some changed were needed as the sdk.proto in root isn't java specific. Compiles like a charme now. I also added some missing dependencies, it should be possible to write tests now.

FelixKlauke:
@markmandel I added the needed scripts. They're pretty much straight forward as the compilation and generation is already managed by Maven.

FelixKlauke:
@markmandel Could you also specify where to add those docs?

aLekSer:
@FelixKlauke Please put your docs into this folder:
[Guides -> Client SDKs](https://github.com/googleforgames/agones/tree/master/site/content/en/docs/Guides/Client%20SDKs)

markmandel:
Hi! Is this being worked on anymore? Otherwise we might close this and pass the work on to someone else.

FelixKlauke:
I might continue the work soon :)

markmandel:
I'm going to close this PR for now, as it's been almost a month.

Feel free to reopen it / start a new one if the work starts again.

stevefan1999-personal:
I would like to take this a little over in the name of #1731, although I work in Kotlin

theminecoder:
I uh... may have done most of the work for this last night https://github.com/theminecoder/agones-java-sdk. it currently doesn't have a watch function for the server as it was out of scope of the projects I need the rest of the lib for, however it's a decently good base.

stevefan1999-personal:
@theminecoder you should use the gRPC generator for Java, which is much better than handcrafting and matching via REST interface

theminecoder:
#For my use cases it makes sense to be using http as the rest of our internal api already exists under http. I am also on a time crunch with one of my projects so I had to weigh getting this out vs playing around grpc settings all day.

stevefan1999-personal:
@theminecoder I've generated an SDK for Kotlin already and I'm testing it, might release soon, and then I'd try to merge the conclusion to Agones.

The problem is only Java is directly compatible with Kotlin, but not the other way round. I'm looking for a workaround without writing disgusting Java trash code

markmandel:
@stevefan1999-personal 

A reminder that our [code of conduct](https://github.com/googleforgames/agones/blob/master/code-of-conduct.md) states that our standards are "welcoming and inclusive language".

Could you please edit you comment to reflect that?  May I suggest something like:

> I prefer not to write Java, so I'm looking for a workaround so that I can implement this all in Kotlin

LucaDev:
Are there any news on this? @stevefan1999-personal it seems like you have discontinued your implementation, am I right?

PR #877, opened 2019-07-05
Fix for CRD API docs generator

aLekSer:
Fix CRD API docs generation process.
Leave only one expiryVersion and one publishVersion tag in the resulting doc.


aLekSer:
Wait a moment before merging, found one issue with awk.

aLekSer:
Fixed an issue with title extracting. It is needed to have a blank line after the title part:
```
+++
title="Agones Kubernetes API"
description="Detailed list of Agones Custom Resource Definitions available"
+++

```
In order to fast check the work of merging part of the script I have commented out generation line and change `base_version` to 1.0 in `Makefile`. 
Proven that updates on new releases goes smoothly.

aLekSer:
1. Algorithm will extract `expiryVersion` with future Release version say 1.0 (could be empty), `publishVersion` with any version as it exist in one instance. 
1. All other `expiryVersion`s would be ignored.
1. Then if something changes in output compared to the previous docs and if version was updated 
    1. we mark the old `publishVersion=0.12` part as `expiryVersion` with future Release version 1.0 
    1. and all new generated docs goes to `publishVersion` as 1.0.
1. If version was not updated:
   1. get the old `expiryVersion=1.0` and put as it was
   1. put new docs in `publishVersion=1.0` and rewriting all previous contents of the same block
1. If nothing was changed - no output

PR #878, opened 2019-07-05
Fix ensure SDK image

aLekSer:
Use bash scripts hash instead of build image version.
Does not rebuild SDK images if image was found, previously rebuild was performed every time.

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -22,12 +22,16 @@
 
 build_sdk_base_version = $(call sha,$(build_path)/build-sdk-images/tool/base/Dockerfile)
 build_sdk_base_tag = agones-build-sdk-base:$(build_sdk_base_version)
+
+# Calculate sha hash of sha hashes of all files in a specified SDK_FOLDER
+build_sdk_version = $(call sha_dir,$(build_path)/build-sdk-images/$(SDK_FOLDER)/*)
 build_sdk_base_remote_tag = $(REGISTRY)/$(build_sdk_base_tag)
 build_sdk_prefix = agones-build-sdk-
 grpc_release_tag = v1.16.1
 sdk_build_folder = build-sdk-images/
 SDK_FOLDER ?= go
 COMMAND ?= gen
+SDK_IMAGE_TAG=$(build_sdk_prefix)$(SDK_FOLDER):$(build_sdk_version)
 
 .PHONY: test-sdks test-sdk build-sdks build-sdk gen-all-sdk-grpc gen-sdk-grpc run-all-sdk-command run-sdk-command 
 
@@ -77,7 +81,7 @@ run-sdk-command:
 		cd - ; \
 		$(MAKE) ensure-build-sdk-image SDK_FOLDER=$(SDK_FOLDER) ; \
 		docker run --rm $(common_mounts) -e "VERSION=$(VERSION)" \
-			$(DOCKER_RUN_ARGS) $(build_sdk_prefix)$(SDK_FOLDER):$(build_version) $(COMMAND) ; \
+			$(DOCKER_RUN_ARGS) $(SDK_IMAGE_TAG) $(COMMAND) ; \
 	else \
 		echo "Command $(COMMAND) not found - nothing to execute" ; \
 	fi
@@ -90,7 +94,7 @@ build-build-sdk-image-base:
 # Builds the docker image used by commands for a specific sdk
 build-build-sdk-image: DOCKER_BUILD_ARGS= --build-arg BASE_IMAGE=$(build_sdk_base_tag)
 build-build-sdk-image: ensure-build-sdk-image-base
-		docker build --tag=$(build_sdk_prefix)$(SDK_FOLDER):$(build_version) $(build_path)build-sdk-images/$(SDK_FOLDER) $(DOCKER_BUILD_ARGS)
+		docker build --tag=$(SDK_IMAGE_TAG) $(build_path)build-sdk-images/$(SDK_FOLDER) $(DOCKER_BUILD_ARGS)
 
 # attempt to pull the image, if it exists and rename it to the local tag
 # exit's clean if it doesn't exist, so can be used on CI
@@ -107,7 +111,7 @@ ensure-build-sdk-image-base:
 
 # create the build image sdk if it doesn't exist
 ensure-build-sdk-image:
-	$(MAKE) build-build-sdk-image SDK_FOLDER=$(SDK_FOLDER)
+	$(MAKE) ensure-image IMAGE_TAG=$(SDK_IMAGE_TAG) BUILD_TARGET=build-build-sdk-image SDK_FOLDER=$(SDK_FOLDER)
 
 # Run SDK conformance Sidecar server in docker in order to run
 # SDK client test against it. Useful for test development
@@ -120,6 +124,7 @@ run-sdk-conformance-local: ensure-agones-sdk-image
 # Run SDK conformance test, previously built, for a specific SDK_FOLDER
 run-sdk-conformance-no-build: TIMEOUT ?= 30
 run-sdk-conformance-no-build: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch
+run-sdk-conformance-no-build: ensure-agones-sdk-image
 run-sdk-conformance-no-build: ensure-build-sdk-image
 	sleep 2s && DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
 	docker run -p 59357:59357 -e "ADDRESS=" \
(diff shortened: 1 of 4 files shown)


aLekSer:
E2E test failed, not connected with this PR:
```
Step #17: time="2019-07-05 20:22:03.977" level=info msg="Scaling fleet" fleet=preferred-gngb4 patch="[{ \"op\": \"replace\", \"path\": \"/spec/replicas\", \"value\": 120 }]" scale=120
Step #17: time="2019-07-05 20:22:04.575" level=info msg="fleet simple-fleet-8mlkz has 5/3 ready replicas"
Step #17: --- FAIL: TestAutoscalerStressCreate (68.99s)
Step #17: fleetautoscaler_test.go:169: 
Step #17: Error Trace:	fleetautoscaler_test.go:169
Step #17: fleetautoscaler_test.go:191
Step #17: Error: Should be true
Step #17: Test: TestAutoscalerStressCreate
Step #17: Messages: FleetAutoscaler created even if the parameters are NOT valid: 4 3 5
Step #17: fleetautoscaler_test.go:169: 
Step #17: Error Trace:	fleetautoscaler_test.go:169
Step #17: fleetautoscaler_test.go:191
Step #17: Error: Should be true
Step #17: Test: TestAutoscalerStressCreate
Step #17: Messages: FleetAutoscaler created even if the parameters are NOT valid: 3 1 2
```

build/includes/sdk.mk
```diff
@@ -28,6 +28,7 @@ grpc_release_tag = v1.16.1
 sdk_build_folder = build-sdk-images/
 SDK_FOLDER ?= go
 COMMAND ?= gen
+SDK_IMAGE_TAG=$(build_sdk_prefix)$(SDK_FOLDER):$(VERSION)
```
markmandel: We have a cross platform SHA function we can use, for example:
https://github.com/googleforgames/agones/blob/master/build/includes/sdk.mk#L23

This will mean the image only gets recreated on Dockerfile change, rather than on every commit.

aLekSer: Yes that's true, but we should add sha of bash scripts also. Will update this PR.

markmandel: We tend to fudge that a little with the `# forceUpdate` comment in the main build Dockerfile - it works well enough for now :man_shrugging: maybe there is a better way?

aLekSer: I will investigate options tomorrow.
shasum function can work on multiple files, we can make equivalent variables for other platforms:
```
shasum ./build-sdk-images/cpp/* | shasum
```
would compute a hash of hashes

markmandel: QQ - does this mean we'll have two tags for the sdk image, one by hash, ,and one by commit?

(Was working on conformance tests for Reserve, and I think I ran into this issue)

aLekSer: Hello, I think we can have only one tag - hash of all sh files as an ideal solution without hash of commit.
There was an issue with `make run-sdk-conformance-test SDK_FOLDER=rust`

aLekSer:
Added crossplatform sha_bash function verified on Linux and MacOs. Currently on both it gives a result of 5e38cff19f.

build/includes/sdk.mk
markmandel: Sorry - this is for the build image - not the sdk-server image. All good. Ignore me.

build/includes/sdk.mk
```diff
@@ -22,12 +22,14 @@
 
 build_sdk_base_version = $(call sha,$(build_path)/build-sdk-images/tool/base/Dockerfile)
 build_sdk_base_tag = agones-build-sdk-base:$(build_sdk_base_version)
+build_sdk_version = $(call sha_bash,$(build_path)/build-sdk-images/*/*.sh)
```
markmandel: A couple of questions on this:

1. What if I change the `/build-sdk-images/go/Dockerfile` then this won't rebuild, since it's only looking at .sh files, no?
1. If i change `/build-sdk-images/go/gen.sh` - won't that mean that all images will get rebuilt?

Should this be (or something close to):
```suggestion
build_sdk_version = $(call sha_bash,$(build_path)/build-sdk-images/$(SDK_FOLDER)/*)
```

?

aLekSer: Nice questions, I don't expect these scripts to be changed too often.
Currently we have rebuild every time. So this version is better, but not absolutely. Let me check both questions in practise. But I think:
1. We would have another base-image so rebuild would be triggered.

markmandel: Sorry I'm confused :confused: , if I _only_ change `/build-sdk-images/go/Dockerfile` - then the image won't rebuild, it would only change if I changed the .sh files. This could be really confusing for the developer.

I can't see a path that would cause that to do anything else? But I could be missing something.

aLekSer: You are right definitely, let's concat those two hashes. Will resend new version in a moment.

markmandel: I think we might be saying two different, but important things.

I think you are referencing when the `/build-sdk-images/tool/base/Dockerfile` changes.

I'm talking about when the language specific Dockerfile changes, for example `/build-sdk-images/go/Dockerfile` 

You are right - both are actually issues!

build/includes/sdk.mk
```diff
@@ -22,12 +22,14 @@
 
 build_sdk_base_version = $(call sha,$(build_path)/build-sdk-images/tool/base/Dockerfile)
 build_sdk_base_tag = agones-build-sdk-base:$(build_sdk_base_version)
+build_sdk_version = $(call sha_bash,$(build_path)/build-sdk-images/$(SDK_FOLDER)/*.sh)$(call sha,$(build_path)/build-sdk-images/$(SDK_FOLDER)/Dockerfile)
```
markmandel:
```suggestion
build_sdk_version = $(call sha_bash,$(build_path)/build-sdk-images/$(SDK_FOLDER)/{*.sh,Dockerfile})
```

`sha256sum {*.sh,Dockerfile} | cut -d' ' -f1 | sha256sum | head -c 10` - works in bash!

aLekSer:
Thanks, let's use just `$(SDK_FOLDER)/*` because suggested version leads to :  
```
sha256sum: '"/home/UserName/go/src/agones.dev/agones/build//build-sdk-images/go/{*.sh': No such file or directory
```

build/includes/sdk.mk
aLekSer: Thanks, let's use just `$(SDK_FOLDER)/*` because suggested version leads to :  
```
sha256sum: '"/home/UserName/go/src/agones.dev/agones/build//build-sdk-images/go/{*.sh': No such file or directory
```

aLekSer: I mean we could even more simplify to:
```
sha256sum {*} | cut -d' ' -f1 | sha256sum | head -c 10
```

markmandel: Yes - we could use everything in the directory. We may end up with a README.md in each one, but that's probably fine if we happen to catch that. WDYT?

build/includes/sdk.mk
markmandel: Just looking at this again, does this bring us back to the original problems?

build/includes/sdk.mk
aLekSer: Yes, the proper wildcard should fix README.md case in the future.

markmandel: I think it's fine if we end up pickup up a README - also it's not an issue _right now_ so, let's just use a solution that simplest for now (use `build_sdk_version = $(call sha_bash,$(build_path)/build-sdk-images/$(SDK_FOLDER)/*`) and we can always revisit it when we actually need to.

WDYT?

aLekSer: Yes, that would work, for now we don't have README files in any of `build-sdk-images/$(SDK_FOLDER)/`, I will update PR with suggested naming.

aLekSer:
Build failed because I mistakenly added `reserve` check into all tests.

build/includes/sdk.mk
```diff
@@ -22,12 +22,14 @@
 
 build_sdk_base_version = $(call sha,$(build_path)/build-sdk-images/tool/base/Dockerfile)
 build_sdk_base_tag = agones-build-sdk-base:$(build_sdk_base_version)
+build_sdk_version = $(call sha_dir,$(build_path)/build-sdk-images/$(SDK_FOLDER)/*.sh)
```
markmandel:
```suggestion
build_sdk_version = $(call sha_dir,$(build_path)/build-sdk-images/$(SDK_FOLDER)/*)
```

Yes?

aLekSer: yes, sorry, fixing

aLekSer:
Updated comment to :
`# Get the sha of all files in a directory using wildcard in $(1)`

aLekSer:
Strange flaky test not connected with this changes:
```
--- FAIL: TestControllerApplyGameServerAddressAndPort (0.15s)
controller_test.go:916: 
Error Trace:	controller_test.go:916
Error: Expected nil, but got: error getting external address for GameServer test: error retrieving node node1 for Pod test: node "node1" not found
Test: TestControllerApplyGameServerAddressAndPort
panic: runtime error: index out of range [recovered]
panic: runtime error: index out of range
```

PR #882, opened 2019-07-09
Last touches to the install guide regarding taints and tolerations

roberthbailey:
Move the bit about the node labels and tolerated taints up to the top of the installation page, as it applies to all environments.

Fixes https://github.com/googleforgames/agones/issues/491.


markmandel:
@googlebot rescan

PR #883, opened 2019-07-09
C# SDK and .NET Core Example

TimoSchmechel:
I added a full implementation of the Agones REST API in C#. I saw there was already an existing Unity SDK however it was using specific Unity webrequests which wont work in other C# codebases. 
This SDK works in Unity as well as .NET Core and Framework. I includes the visual studio files, if you want me to remove them and just keep it as code files only I am happy to remove them. 
I have included a basic .NET Core example which uses all the features of the API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ build/local-includes/*
 !build/local-includes/README.md
 /release
 debug.test
+/sdks/csharp/AgonesClient/obj
+/sdks/csharp/AgonesClient/Properties
+/sdks/csharp/packages
+/sdks/csharp/AgonesTestCore/obj
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,5 @@ build/local-includes/*
 /release
 debug.test
 /sdks/csharp/AgonesClient/obj
-/sdks/csharp/AgonesClient/Properties
 /sdks/csharp/packages
 /sdks/csharp/AgonesTestCore/obj
--- a/sdks/csharp/AgonesClient/AgonesClient.cs
+++ b/sdks/csharp/AgonesClient/AgonesClient.cs
@@ -1,0 +1,225 @@
+﻿using Agones.Models;
+using Agones.Utility;
+using Newtonsoft.Json;
+using System;
+using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+
+namespace Agones
+{
+	/// <summary>
+	/// The delegate type used to receive info when watching a game server
+	/// </summary>
+	/// <param name="serverInfo"></param>
+	public delegate void GameServerInfoHandler(GameServerInfo serverInfo);
+
+	/// <summary>
+	/// Agones Client used to interface with the agones sidecar
+	/// </summary>
+	public static class AgonesClient
+	{
+		const string BaseUrl = "http://localhost:59358/";
+
+		static HttpClient httpClient;
+		
+		/// <summary>
+		/// The logger that is used for all agones client logging
+		/// </summary>
+		public static ILogger Logger { get; set; } = new ConsoleLogger()
+		{
+			Prefix = "[AGONES] - "
+		};
+
+		static AgonesClient()
+		{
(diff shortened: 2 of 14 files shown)


FelixKlauke:
Nice work! Can you please squash your commits into a single one? You may also take a look at [build-sdk-images](https://github.com/googleforgames/agones/tree/master/build/build-sdk-images). Even if code generation isn't needed, the tests should be run.

AFAIK gRPC supports C#, is there any reason you cannot use it? It's the preferred way of implementation.

sdks/csharp/AgonesClient/AgonesClient.cs
```diff
@@ -0,0 +1,225 @@
+﻿using Agones.Models;
```
FelixKlauke: We need a license header here.

sdks/csharp/AgonesClient/ConsoleLogger.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
```
FelixKlauke: We need a license header here.

sdks/csharp/AgonesClient/GameServerInfo.cs
```diff
@@ -0,0 +1,107 @@
+﻿using Agones.Utility;
```
FelixKlauke: We need a license header here.

sdks/csharp/AgonesClient/ILogger.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
FelixKlauke: We need a license header here.

sdks/csharp/AgonesTestCore/Program.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Agones;
```
FelixKlauke: This is just a local manual test class. Consider replacing this with real (unit) tests.

sdks/csharp/AgonesClient/CustomUnixDateTimeConverter.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Newtonsoft.Json;
```
FelixKlauke: We need a license header here.

TimoSchmechel:
@FelixKlauke The main reason I didnt use gRPC generation is because I had never used it before and have decent experience writing a REST client manually. As I am actively using this client in the game Im developing I felt it was faster for me to go down that route. Happy to look at generation though. 
I'll look through the testing, license header and other issues you mentioned over the next few days.

TimoSchmechel:
Just created an Issue regarding this here #884

markmandel:
I would agree with @FelixKlauke - C# has pretty good gRPC support, and where possible, we should aim for a gRPC implementation. It will be easier to maintain, has more features, and have better performance.

We only go REST when we have to/its a better solution.

markmandel:
@googlebot rescan

markmandel:
@googlebot rescan

.gitignore
```diff
@@ -1,34 +1,37 @@
-# Copyright 2017 Google LLC All Rights Reserved.
```
aLekSer: I think this file showed up just because you changed line endings from \n to \r\n, from Linux style to Win.

aLekSer:
Also regarding VS proj files in the PR. Cmake supports C # project files generation. Would be useful if multiple IDE would be used with this SDK.
Makefile as in `build-sdk-images` would be useful to understand how to build the project if possible from CLI.

TimoSchmechel:
Alright seems like you're pretty adamant about going for gRPC, which makes sense as its easier to maintain going forward. Happy to reimplement using that some time next week when I get time.

How well does it support streams? Specifically the watch gameserver case

markmandel:
> How well does it support streams? Specifically the watch gameserver case

About a million times easier :smile: single/biderectional streaming is wher gRPC really shines (among other things)

markmandel:
Should we close this PR, and reopen for a new gRPC based one?

TimoSchmechel:
Yup makes the most sense

Issue #717, opened 2019-04-17
Upgrade to Kubernetes 1.12


heartrobotninja:
The version of client-go we vendor is 8.0 and is deprecated per the k8s client-go [compatability matrix](https://github.com/kubernetes/client-go#compatibility-matrix). They use [semver rules](https://github.com/kubernetes/client-go#versioning) which means that we may have breaking changes in 9.0+ (but we also might not). We should investigate what it would take to bring us up to current, or as close to current as possible.

heartrobotninja:
Another part of this is that if I vendor k8s.io/code-generator at its latest incarnation, I get an exception when running `make gen-crd-client` because the old client-go version uses glog, and code-generator uses klog (which is what should be used) and they use the same log_dir flag which results in this:
```
Generating deepcopy funcs
/go/bin/deepcopy-gen flag redefined: log_dir
panic: /go/bin/deepcopy-gen flag redefined: log_dir

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc00006c180, 0x78cc20, 0x9c8770, 0x703a10, 0x7, 0x73b82b, 0x2f)
        /usr/local/go/src/flag/flag.go:850 +0x4af
flag.(*FlagSet).StringVar(...)
        /usr/local/go/src/flag/flag.go:753
k8s.io/klog.InitFlags(0x0)
        /go/src/k8s.io/klog/klog.go:411 +0xa4
main.main()
        /go/src/k8s.io/code-generator/cmd/deepcopy-gen/main.go:59 +0x3f
Makefile:364: recipe for target 'gen-crd-client' failed
make: *** [gen-crd-client] Error 2
```

markmandel:
So the unwritten rule I've been following (we should make some kind of written rule), is that we've been updating the client as the next version of K8s has cloud penetration (which I've taken at available on EKS, GKE and AKS).

Usually there is breaking changes between each client-go version, but they tend to be relatively trivial to fix.

heartrobotninja:
Can you comment on what backwards compatibility usually is like with the major version increases?

markmandel:
Usually we upgrade because we want a latest feature - so backward compatibility has been less of a concern for us.

Moving forward, it's a tough question, as we only have https://github.com/kubernetes/client-go#compatibility-matrix to go by, which is not the most reliable thing in the world.

Realistically, we should run our own tests.

[...]

PR #887, opened 2019-07-09
Update GKE development tooling to 1.12

markmandel:
Updated to 1.12:
- GKE dev cluster deployment manager template
- e2e dev cluster deployment manager template
- kubectl version

Work on #717

### DIFF
--- a/build/gke-test-cluster/cluster-e2e.yml
+++ b/build/gke-test-cluster/cluster-e2e.yml
@@ -19,10 +19,10 @@ resources:
     cluster:
       name: e2e-test-cluster
       description: End to end tests cluster for Agones
-      initialClusterVersion: "1.11"
+      initialClusterVersion: "1.12"
       nodePools:
         - name: "default"
-          initialNodeCount: 4
+          initialNodeCount: 8
           config:
             machineType: n1-standard-4
             tags:
@@ -34,10 +34,10 @@ resources:
               - https://www.googleapis.com/auth/service.management.readonly
               - https://www.googleapis.com/auth/servicecontrol
               - https://www.googleapis.com/auth/trace.append
-- name: game-server-firewall
+- name: gke-game-server-firewall
   type: compute.beta.firewall
   properties:
-    name: game-server
+    name: gke-game-server-firewall
     description: "Firewall to allow game server udp traffic"
     targetTags:
       - "game-server"
--- a/build/gke-test-cluster/cluster.yml.jinja
+++ b/build/gke-test-cluster/cluster.yml.jinja
@@ -21,7 +21,7 @@ resources:
     cluster:
       name: {{ properties["cluster.name"] }}
       description: Test cluster for Agones
-      initialClusterVersion: "1.11"
+      initialClusterVersion: "1.12"
       nodePools:
         - name: "default"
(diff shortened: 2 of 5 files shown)


markmandel:
/cc @heartrobotninja do the GKE parameters look good to you here?

I ran most of the e2e tests on my local development cluster, and things look good!

build/gke-test-cluster/cluster-e2e.yml
```diff
@@ -19,10 +19,10 @@ resources:
     cluster:
       name: e2e-test-cluster
       description: End to end tests cluster for Agones
-      initialClusterVersion: "1.11"
+      initialClusterVersion: "1.12"
       nodePools:
         - name: "default"
-          initialNodeCount: 4
+          initialNodeCount: 8
```
roberthbailey: 8*4 = 32 which i believe is larger than the default quota in a zone for new projects. Do we need to bump this for individuals to run e2e tests? I know the shared e2e cluster has 8 nodes, but in my very limited testing with my own e2e cluster 4 seemed sufficient.

build/gke-test-cluster/cluster.yml.jinja
```diff
@@ -71,10 +71,7 @@ resources:
                 value: "true"
                 effect: "NO_EXECUTE"
       masterAuth:
-        username: admin
-        password: supersecretpassword
-      legacyAbac:
-        enabled: {{ properties["cluster.legacyAbac"] }}
+        username: ""
```
roberthbailey: You should remove the masterAuth block. From gcloud `WARNING: Starting in 1.12, new clusters will have basic authentication disabled by default.`

build/gke-test-cluster/cluster-e2e.yml
markmandel: So this is for the agones-images project, not really for individuals (although they are welcome to use it too). We run a higher parallel rate (32 from memory), and need the extra room.

In fact, given that e2e now takes 8 minutes, we could bump that up a bit too - hence the extra space.

build/gke-test-cluster/cluster.yml.jinja
markmandel: > The username to use for HTTP basic authentication to the master endpoint. For clusters v1.6.0 and later, basic authentication can be disabled by leaving username unspecified (or setting it to the empty string).

([link](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.MasterAuth
)) thought it might be good to explicitly turn it off. But removing it will do the same thing, so I'll remove it.

build/gke-test-cluster/cluster-e2e.yml
roberthbailey: Ah, I didn't realize that this was only used for e2e. The e2e cluster already has 8 nodes, so this just makes that consistent.

build/gke-test-cluster/cluster.yml.jinja
roberthbailey: I did a test last night and removing the block for master auth in terraform does the right thing in 1.12 (no basic auth, no client cert). It sounds like deployment mgr works the same way, which is nice.

PR #888, opened 2019-07-10
Explicitly disable creation of the client certificate on GKE

roberthbailey:
This is the default behavior starting with 1.12, which we are switching over to soon. Also remove the output variables for client key and client certificate as they aren't used. 

Part of #717. 

/assign @aLekSer


aLekSer:
Let me check the resulted terraform configuration files.

aLekSer:
Found one issue with small timeout in `./build/helm.tf` on 
`resource "helm_release" "agones" ` should be extended to 300 seconds (3 minutes now is not enough):
```
  timeout = 300
```

PR #889, opened 2019-07-10
Move the stable api group and promote it to v1 (WIP)

### LINKED ISSUE #703: Proposal: Split up the api group stable.agones.dev

roberthbailey:
I wanted to get the first part of this out for folks to look at before I go through all of the docs and install bits and get them updated. I've done the package move and version change and fixed enough of the code to get the unit tests to pass. 

Fixes #703.


roberthbailey:
Closing so that I can open a PR in draft status.

PR #891, opened 2019-07-10
Reserve() SDK implementation

markmandel:
This implements Reserve, with an implementation with the Go SDK.

It was also convenience to switch Allocate back to the standard async
implementation here, so that all the SDK commands are consistent in
their concurrency patterns.

Still to come:
- e2e tests
- sdk conformance test
- updates to gameserver lifecycle documentation

### DIFF
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -56,9 +56,7 @@ const (
 	updateAnnotation Operation = "updateAnnotation"
 )
 
-var (
-	_ sdk.SDKServer = &SDKServer{}
-)
+var _ sdk.SDKServer = &SDKServer{}
 
 // SDKServer is a gRPC server, that is meant to be a sidecar
 // for a GameServer that will update the game server status on SDK requests
@@ -200,7 +198,9 @@ func (s *SDKServer) Run(stop <-chan struct{}) error {
 	s.initHealthLastUpdated(time.Duration(gs.Spec.Health.InitialDelaySeconds) * time.Second)
 
 	if gs.Status.State == stablev1alpha1.GameServerStateReserved && gs.Status.ReservedUntil != nil {
+		s.gsUpdateMutex.Lock()
 		s.resetReserveAfter(context.Background(), time.Until(gs.Status.ReservedUntil.Time))
+		s.gsUpdateMutex.Unlock()
 	}
 
 	// start health checking running
@@ -259,13 +259,13 @@ func (s *SDKServer) updateState() error {
 		return err
 	}
 
-	// if we are currently in shutdown/being deleted, there is no escaping
+	// If we are currently in shutdown/being deleted, there is no escaping.
 	if gs.IsBeingDeleted() {
 		s.logger.Info("GameServerState being shutdown. Skipping update.")
 		return nil
 	}
 
-	// if the state is currently unhealthy, you can't go back to Ready
+	// If the state is currently unhealthy, you can't go back to Ready.
 	if gs.Status.State == stablev1alpha1.GameServerStateUnhealthy {
 		s.logger.Info("GameServerState already unhealthy. Skipping update.")
 		return nil
@@ -274,7 +274,7 @@ func (s *SDKServer) updateState() error {
 	s.gsUpdateMutex.RLock()
 	gs.Status.State = s.gsState
 
-	// if we are setting the Reserved status, check for the duration, and set that too
+	// If we are setting the Reserved status, check for the duration, and set that too.
 	if gs.Status.State == stablev1alpha1.GameServerStateReserved && s.gsReserveDuration != nil {
 		n := metav1.NewTime(time.Now().Add(*s.gsReserveDuration))
 		gs.Status.ReservedUntil = &n
@@ -295,16 +295,12 @@ func (s *SDKServer) updateState() error {
 	case stablev1alpha1.GameServerStateUnhealthy:
 		level = corev1.EventTypeWarning
 	case stablev1alpha1.GameServerStateReserved:
-		s.gsUpdateMutex.RLock()
-		hasDuration := s.gsReserveDuration != nil
-		if hasDuration {
-			message += fmt.Sprintf(", for %v", s.gsReserveDuration.String())
-		}
-		s.gsUpdateMutex.RUnlock()
-		// unfortunate complexity due to mutex locking, resetReserveAfter has it's own mutex
-		if hasDuration {
+		s.gsUpdateMutex.Lock()
+		if s.gsReserveDuration != nil {
+			message += fmt.Sprintf(", for %s", s.gsReserveDuration)
 			s.resetReserveAfter(context.Background(), *s.gsReserveDuration)
 		}
+		s.gsUpdateMutex.Unlock()
 	}
 
 	s.recorder.Event(gs, level, string(gs.Status.State), message)
@@ -474,40 +470,35 @@ func (s *SDKServer) Reserve(ctx context.Context, d *sdk.Duration) (*sdk.Empty, e
 
 	e := &sdk.Empty{}
 
-	s.logger.Info("Received Reserve request, adding to queue")
-	s.enqueueState(stablev1alpha1.GameServerStateReserved)
-
 	// 0 is forever.
-	if d.Seconds == 0 {
-		return e, nil
+	if d.Seconds > 0 {
+		duration := time.Duration(d.Seconds) * time.Second
+		s.gsUpdateMutex.Lock()
+		s.gsReserveDuration = &duration
+		s.gsUpdateMutex.Unlock()
 	}
 
-	duration := time.Duration(d.Seconds) * time.Second
-	s.gsUpdateMutex.Lock()
-	s.gsReserveDuration = &duration
-	defer s.gsUpdateMutex.Unlock()
+	s.logger.Info("Received Reserve request, adding to queue")
+	s.enqueueState(stablev1alpha1.GameServerStateReserved)
 
 	return e, nil
 }
 
-// resetReserveAfter will run a function after duration to move the GameServer
-// back to being Ready
+// resetReserveAfter will move the GameServer back to being ready after the specified duration.
+// This function should be wrapped in a s.gsUpdateMutex lock when being called.
 func (s *SDKServer) resetReserveAfter(ctx context.Context, duration time.Duration) {
-	s.gsUpdateMutex.Lock()
-	defer s.gsUpdateMutex.Unlock()
 	if s.reserveTimer != nil {
 		s.reserveTimer.Stop()
 	}
 
 	s.reserveTimer = time.AfterFunc(duration, func() {
-		_, err := s.Ready(ctx, &sdk.Empty{})
-		if err != nil {
+		if _, err := s.Ready(ctx, &sdk.Empty{}); err != nil {
 			s.logger.WithError(errors.WithStack(err)).Error("error returning to Ready after reserved")
 		}
 	})
 }
 
-// stopReserveTimer stops the reserve timer, if it is not nil
+// stopReserveTimer stops the reserve timer. This is a no-op and safe to call if the timer is nil
 func (s *SDKServer) stopReserveTimer() {
 	s.gsUpdateMutex.Lock()
 	defer s.gsUpdateMutex.Unlock()
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -15,6 +15,7 @@
 package sdkserver
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
@@ -35,7 +36,6 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	corev1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/clock"
@@ -58,14 +58,6 @@ const (
 
 var (
 	_ sdk.SDKServer = &SDKServer{}
-
-	defaultTimeout = 30 * time.Second
-	defaultBackoff = wait.Backoff{
-		Duration: 100 * time.Millisecond,
-		Factor:   1,
-		Jitter:   0.1,
-		Steps:    5,
-	}
 )
 
 // SDKServer is a gRPC server, that is meant to be a sidecar
@@ -96,6 +88,8 @@ type SDKServer struct {
 	gsState            stablev1alpha1.GameServerState
 	gsUpdateMutex      sync.RWMutex
 	gsWaitForSync      sync.WaitGroup
+	reserveTimer       *time.Timer
+	gsReserveDuration  *time.Duration
 }
 
 // NewSDKServer creates a SDKServer that sets up an
@@ -205,6 +199,10 @@ func (s *SDKServer) Run(stop <-chan struct{}) error {
 	s.healthTimeout = time.Duration(gs.Spec.Health.PeriodSeconds) * time.Second
 	s.initHealthLastUpdated(time.Duration(gs.Spec.Health.InitialDelaySeconds) * time.Second)
 
+	if gs.Status.State == stablev1alpha1.GameServerStateReserved && gs.Status.ReservedUntil != nil {
+		s.resetReserveAfter(context.Background(), time.Until(gs.Status.ReservedUntil.Time))
+	}
+
 	// start health checking running
 	if !s.health.Disabled {
 		s.logger.Info("Starting GameServer health checking")
@@ -275,6 +273,14 @@ func (s *SDKServer) updateState() error {
 
 	s.gsUpdateMutex.RLock()
 	gs.Status.State = s.gsState
+
+	// if we are setting the Reserved status, check for the duration, and set that too
+	if gs.Status.State == stablev1alpha1.GameServerStateReserved && s.gsReserveDuration != nil {
+		n := metav1.NewTime(time.Now().Add(*s.gsReserveDuration))
+		gs.Status.ReservedUntil = &n
+	} else {
+		gs.Status.ReservedUntil = nil
+	}
 	s.gsUpdateMutex.RUnlock()
 
 	_, err = gameServers.Update(gs)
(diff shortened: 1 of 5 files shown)


roberthbailey:
Looks like this is part of https://github.com/googleforgames/agones/issues/660.

markmandel:
> Looks like this is part of #660.

Yes! Thanks, forgot to put that on here.

pkg/sdkserver/sdkserver.go
```diff
@@ -58,14 +58,6 @@ const (
 
 var (
```
roberthbailey: This can go back to a single line instead of a block now that there is only one variable.

pkg/sdkserver/sdkserver.go
```diff
@@ -275,21 +273,41 @@ func (s *SDKServer) updateState() error {
 
 	s.gsUpdateMutex.RLock()
 	gs.Status.State = s.gsState
+
+	// if we are setting the Reserved status, check for the duration, and set that too
```
roberthbailey: nit: Comments should be proper sentences, starting with a capital letter and ending with a period. 

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

pkg/sdkserver/sdkserver.go
```diff
@@ -275,21 +273,41 @@ func (s *SDKServer) updateState() error {
...
 	case stablev1alpha1.GameServerStateUnhealthy:
 		level = corev1.EventTypeWarning
+	case stablev1alpha1.GameServerStateReserved:
+		s.gsUpdateMutex.RLock()
+		hasDuration := s.gsReserveDuration != nil
+		if hasDuration {
+			message += fmt.Sprintf(", for %v", s.gsReserveDuration.String())
+		}
+		s.gsUpdateMutex.RUnlock()
+		// unfortunate complexity due to mutex locking, resetReserveAfter has it's own mutex
```
roberthbailey: same nit about comment sentences.

pkg/sdkserver/sdkserver.go
```diff
@@ -275,21 +273,41 @@ func (s *SDKServer) updateState() error {
...
 	level := corev1.EventTypeNormal
 	// post state specific work here
 	switch gs.Status.State {
 	case stablev1alpha1.GameServerStateUnhealthy:
 		level = corev1.EventTypeWarning
+	case stablev1alpha1.GameServerStateReserved:
+		s.gsUpdateMutex.RLock()
+		hasDuration := s.gsReserveDuration != nil
+		if hasDuration {
+			message += fmt.Sprintf(", for %v", s.gsReserveDuration.String())
```
roberthbailey: Does this output nicely? The new state is passed as a separate argument than the message, so it isn't clear to me how this will actually appear in the event output. Also, should the message contain units for the duration (or are those appended automatically)?

pkg/sdkserver/sdkserver.go
```diff
@@ -486,9 +469,53 @@ func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServer
...
+	s.logger.Info("Received Reserve request, adding to queue")
+	s.enqueueState(stablev1alpha1.GameServerStateReserved)
+
+	// 0 is forever.
+	if d.Seconds == 0 {
+		return e, nil
+	}
+
+	duration := time.Duration(d.Seconds) * time.Second
+	s.gsUpdateMutex.Lock()
```
roberthbailey: should this be done before the enqueueState call above? It seems racey to set the new state and the duration in two separate locks regardless of the order....

pkg/sdkserver/sdkserver.go
```diff
@@ -486,9 +469,53 @@ func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServer
...
+	s.gsReserveDuration = &duration
+	defer s.gsUpdateMutex.Unlock()
+
+	return e, nil
+}
+
+// resetReserveAfter will run a function after duration to move the GameServer
```
roberthbailey: `// resetReserveAfter will move the GameServer back to being ready after the specified duration.`

pkg/sdkserver/sdkserver.go
```diff
@@ -486,9 +469,53 @@ func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServer
...
+// back to being Ready
+func (s *SDKServer) resetReserveAfter(ctx context.Context, duration time.Duration) {
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+	if s.reserveTimer != nil {
+		s.reserveTimer.Stop()
+	}
+
+	s.reserveTimer = time.AfterFunc(duration, func() {
+		_, err := s.Ready(ctx, &sdk.Empty{})
```
roberthbailey: nit: The if should be "cuddled":

```go
if _, err := s.Ready(...); err != nil {
  s.logger.WithError(...)
}
```

to reduce the scope of the `err` variable.

pkg/sdkserver/sdkserver.go
```diff
@@ -486,9 +469,53 @@ func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServer
...
+		if err != nil {
+			s.logger.WithError(errors.WithStack(err)).Error("error returning to Ready after reserved")
+		}
+	})
+}
+
+// stopReserveTimer stops the reserve timer, if it is not nil
```
roberthbailey: maybe better as `// stopReserveTimer stops the reserve timer. This is a no-op and safe to call if the timer does not exist.`

pkg/sdkserver/sdkserver_test.go
```diff
@@ -112,7 +109,18 @@ func TestSidecarRun(t *testing.T) {
...
+				state:      v1alpha1.GameServerStateAllocated,
+				recordings: []string{string(v1alpha1.GameServerStateAllocated)},
+			},
+		},
+		"reserved": {
+			f: func(sc *SDKServer, ctx context.Context) {
+				_, err := sc.Reserve(ctx, &sdk.Duration{Seconds: 10})
+				assert.NoError(t, err)
+			},
+			expected: expected{
```
roberthbailey: The expected struct doesn't have a way to check for the expected timer existing. Is there a way to add that check in?

From what I can tell, this is only testing one part of the new functionality: it doesn't test passing in zero and it doesn't verify that the timer properly moves the game server back to ready once the passed in time has elapsed. Will those be covered in tests elsewhere?

What happens if you set seconds to a negative number?

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
-		case <-time.After(5 * time.Second):
-			assert.Fail(t, "should be updated")
-		}
+		return true, gs, nil
 	})
 
-	t.Run("contention on first build", func(t *testing.T) {
-		m, sc, stop := setup(t, defaultGs.DeepCopy())
-		defer close(stop)
+	sc, err := NewSDKServer("test", "default", m.KubeClient, m.AgonesClient)
```
roberthbailey: Why doesn't this function use `sc, err := defaultSidecar(m)`

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
-	})
+	}
+	noAdditional := func(status v1alpha1.GameServerStatus) {}
 
-	t.Run("unhealthy gameserver", func(t *testing.T) {
-		gs := defaultGs.DeepCopy()
-		gs.Status.State = v1alpha1.GameServerStateUnhealthy
-		_, sc, stop := setup(t, gs.DeepCopy())
-		defer close(stop)
+	// 3 seconds
```
roberthbailey: This comment doesn't add much value, since the 3 seconds is obvious from the line below. Is there something worth describing here? If not, please remove.

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
-	t.Run("Shutdown gameserver", func(t *testing.T) {
-		gs := defaultGs.DeepCopy()
-		gs.Status.State = v1alpha1.GameServerStateShutdown
-		_, sc, stop := setup(t, gs.DeepCopy())
-		defer close(stop)
-
-		_, err := sc.Allocate(context.Background(), &sdk.Empty{})
-		assert.EqualError(t, err, "cannot Allocate a Shutdown GameServer")
+	// wait for the game server to go back to being Ready
+	assertStateChange(v1alpha1.GameServerStateRequestReady, 4*time.Second, func(status v1alpha1.GameServerStatus) {
```
roberthbailey: Unit tests are a lot more reliable if you don't rely on the system clock and hard coding times to wait for. Can you cause this to happen immediately if you mock the time and just advance the clock by 3 seconds immediately?

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
 
-	t.Run("Deleting gameserver", func(t *testing.T) {
-		gs := defaultGs.DeepCopy()
-		now := metav1.Now()
-		gs.ObjectMeta.DeletionTimestamp = &now
-		_, sc, stop := setup(t, gs.DeepCopy())
-		defer close(stop)
+	// now test for being able to escape the Reserved state when requesting another state
+
+	// test ready
```
roberthbailey: What does this comment mean?

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
+	_, err = sc.Reserve(context.Background(), &sdk.Duration{Seconds: 3})
+	assert.NoError(t, err)
+	assertStateChange(v1alpha1.GameServerStateReserved, 2*time.Second, noAdditional)
```
roberthbailey: above the check after calling reserve had an assert, why not do one here too?

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
+	assert.NoError(t, err)
+	assertStateChange(v1alpha1.GameServerStateRequestReady, 2*time.Second, func(status v1alpha1.GameServerStatus) {
+		assert.Nil(t, status.ReservedUntil)
 	})
 
-	t.Run("Unexpected error", func(t *testing.T) {
-		m, sc, stop := setup(t, defaultGs.DeepCopy())
-		defer close(stop)
-		done := make(chan struct{}, 10)
+	select {
```
roberthbailey: Is this waiting 4 seconds to verify that the timer has been disabled? Can you just check that directly instead of looking for a side effect?

pkg/sdkserver/sdkserver_test.go
roberthbailey: I think I see after reading below. Please change to a full sentence, something like

```go
// Test the transition from reserved to the ready state. 
```

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
+	case current := <-state:
+		assert.Failf(t, "Should not get update:", string(current.State))
+	case <-time.After(4 * time.Second):
+	}
 
-		m.AgonesClient.AddReactor("update", "gameservers", func(action k8stesting.Action) (bool, runtime.Object, error) {
-			defer func() {
-				done <- struct{}{}
-			}()
+	// test allocated
```
roberthbailey: Similar to above, make this a full sentence.

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
+		assert.Nil(t, status.ReservedUntil)
+	})
+
+	select {
+	case <-state:
+		assert.Fail(t, "Should not go back to RequestReady")
+	case <-time.After(4 * time.Second):
+	}
+
+	// test shutdown
```
roberthbailey: Here too.

pkg/sdkserver/sdkserver_test.go
```diff
@@ -678,176 +685,185 @@ func TestSDKServerUpdateEventHandler(t *testing.T) {
...
-		_, err := sc.Allocate(context.Background(), &sdk.Empty{})
-		assert.Error(t, err)
+	_, err = sc.Shutdown(context.Background(), &sdk.Empty{})
+	assert.NoError(t, err)
+	assertStateChange(v1alpha1.GameServerStateShutdown, 2*time.Second, func(status v1alpha1.GameServerStatus) {
+		assert.Nil(t, status.ReservedUntil)
 	})
+
+	select {
+	case current := <-state:
```
roberthbailey: With 3 test cases all waiting 4+ seconds, this change adds a considerable amount of time to unit tests (which should always be blazingly fast). It would be nice to see if there is a way to do this that completes nearly instantly instead (e.g. my comment about clock manipulation above).

pkg/sdkserver/sdkserver.go
Kuqd: String returns a string representing the duration in the form "72h3m0.5s".

Kuqd: The final message will be 'SDK state change, for 10m'. I also think it lacks of information even though the reason is `Reserved`. I would prefer `SDK state change to Reserved for ...' or even better `SDK state change to Reserved until utc date` since when looking at the event you might not know if it is gone.

Also you can drop the `.String()`, %v use the `Stringer` interface if implemented which is your case.

pkg/sdkserver/sdkserver.go
Kuqd: Its not clear to me why you need to lock that mutex.

pkg/sdkserver/sdkserver_test.go
Kuqd: it crashes, been there done that.

pkg/sdkserver/sdkserver_test.go
Kuqd: Interesting idea I always had this problem,.

pkg/sdkserver/sdkserver_test.go
Kuqd: Agreed, there are in parallel though.

Kuqd:
Looks good very good. Ping me back when you've gone through all.

markmandel:
Thanks for all the comments - will review tomorrow - got lots of time for coding tomorrow :+1:

site/content/en/docs/Guides/Client SDKs/_index.md
```diff
@@ -123,6 +123,25 @@ For those scenarios, this SDK functionality exists.
...
+With some matchmaking scenarios and systems it is important to be able to ensure that a `GameServer` is unable to be deleted,
+but doesn't trigger a FleetAutoscaler scale up. This is where `Reserver()` is useful.
+
+`Reserve(seconds)` will move the `GameServer` into the Reserved state for the specified number of seconds (0 is forever), and then it will be
+moved back to `Ready` state. While in `Reserved` state, the `GameServer` cannot be deleted on scale down or `Fleet` update,
+and also cannot be Allocated. 
+
+This is often used wherein a game server process must self register itself with an external system, such as a matchmaker,
+that requires it to designate itself as available for a game session for a certain period. Once a game session has started,
+it can call `SDK.Allocate` to designate that players are currently active on it.
```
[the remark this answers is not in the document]
markmandel: I switched it to `should` - just because you don't _have_ to, but you should. :+1:

pkg/sdkserver/sdkserver.go
markmandel: I've cleaned this up, much simpler locking now. PTAL.

pkg/sdkserver/sdkserver.go
markmandel: > Also you can drop the .String(), %v use the Stringer interface if implemented which is your case.

Done!

> The final message will be SDK state change, for 10m. I also think it lacks of information even though the reason is Reserved. I would prefer SDK state change to Reserved for ... or even better SDK state change to Reserved until utc date since when looking at the event you might not know if it is gone.

To give an example of the full event output, that might help with context on this:

```
Events:
  Type    Reason          Age               From                   Message
  ----    ------          ----              ----                   -------
  Normal  PortAllocation  53s               gameserver-controller  Port allocated
  Normal  Creating        53s               gameserver-controller  Pod simple-udp-567rr created
  Normal  Scheduled       53s               gameserver-controller  Address and port populated
  Normal  RequestReady    47s               gameserver-sidecar     SDK state change
  Normal  Reserved        10s               gameserver-sidecar     SDK state change, for 10s
  Normal  Ready           0s (x2 over 47s)  gameserver-controller  SDK.Ready() complete
```

So you can see the State in the `Reason` column, so I figure it's redundant to include it again in the Message.

I would also leave the time period as `for 10s` - reason being, from the `Age` column, you can actually see how long its been since the `Reserved` state was set, and therefore know what the difference is going to be, and can calculate the time left.

How do you both feel about this now?

pkg/sdkserver/sdkserver_test.go
roberthbailey: Here you pass in the `noAdditional` function as the third argument. Above you did

```go
assertStateChange(v1alpha1.GameServerStateReserved, 2*time.Second, func(status v1alpha1.GameServerStatus) {
		assert.Equal(t, time.Now().Add(3*time.Second).Round(time.Second), status.ReservedUntil.Time.Round(time.Second))
})
```

So the first time you verify that the state changed via the callback but in subsequent times you don't.

pkg/sdkserver/sdkserver_test.go
roberthbailey: I'm ok with this for now, but as you said in slack the presubmit tests are getting longer and longer (30+ minutes) and need some optimization. Maybe this isn't the most important place to start, but optimizing unit tests should also be on that list.

As for changing the clock interface... there's a cost of taking on new dependencies (check out https://research.swtch.com/deps if you haven't seen it) and I'm not sure that it's worth it in this case. It might be worth asking apimachinery to update their interface, but even if they did it today we wouldn't see the benefit for quite a while.

pkg/sdkserver/sdkserver_test.go
roberthbailey: I saw the reset test after I'd written this comment. I think the additional tests below also do a sufficient job making sure the timer exists and fires correctly.

pkg/sdkserver/sdkserver_test.go
markmandel: Yeah, I had both of those thoughts as well. Okay cool - so agreed on leaving this as is. Maybe this is something we can add to the apimachinery mock down the road.

pkg/sdkserver/sdkserver_test.go
markmandel: Oh! I see it now. I figured it might be redundant to test it again, but I see your point. I'll make the change!

markmandel: And done!

pkg/sdkserver/sdkserver.go
```diff
@@ -474,40 +470,35 @@ func (s *SDKServer) Reserve(ctx context.Context, d *sdk.Duration) (*sdk.Empty, e
 
 	e := &sdk.Empty{}
 
-	s.logger.Info("Received Reserve request, adding to queue")
-	s.enqueueState(stablev1alpha1.GameServerStateReserved)
-
 	// 0 is forever.
-	if d.Seconds == 0 {
-		return e, nil
+	if d.Seconds > 0 {
```
roberthbailey: This check actually means that 0 _or a negative value_ means forever.

markmandel: Yeah, that was deliberate. I wanted to handle the error condition in a sane way.

Do you feel I should change the docs?

markmandel:
Going to push this through, as I know you are waiting on it for some other PRs. @roberthbailey if you think I should change the docs, I'll do so in a separate PR.

Issue #892, opened 2019-07-10
Update to opencensus v0.22


pjayara-g:
This is from #831 . @markmandel suggested updating the version of opencensus. This bug is to track that.

markmandel:
This still would be great - just need to split things in the commit -- you can see details in the #893 PR.

markmandel:
gRPC has been updated, so we can now do this work! :fire:

aLekSer:
This should fix the following error on allocation Metrics:
`Failed to export to Prometheus: inconsistent label cardinality: expected 1 label values but got 0 in []string(nil)`
https://github.com/googleforgames/agones/issues/1330#issuecomment-584737095

aLekSer:
We have updated OpenCensus to the most recent release:
https://github.com/census-instrumentation/opencensus-go/releases/tag/v0.22.3

PR #893, opened 2019-07-10
Update to opencensus v0.22

pjayara-g:
The changes updates the version to v0.22. It also uses the newish ReadAndExport method in opencensus instead of using the prometheus exporter for validation in unit tests. Addresses #892.


Kuqd:
I'll take a look this week.

go.mod
```diff
@@ -34,14 +35,13 @@ require (
...
-	golang.org/x/sync v0.0.0-20181221193216-37e7f081c4d4 // indirect
+	go.opencensus.io v0.22.0
+	golang.org/x/net v0.0.0-20190501004415-9ce7a6920f09
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2
 	golang.org/x/tools v0.0.0-20190328211700-ab21143f2384
 	google.golang.org/api v0.0.0-20190117000611-43037ff31f69 // indirect
-	google.golang.org/genproto v0.0.0-20190111180523-db91494dd46c
-	google.golang.org/grpc v1.17.0
+	google.golang.org/genproto v0.0.0-20190425155659-357c62f0e4bb
+	google.golang.org/grpc v1.20.1
```
Kuqd: grpc bump and protobuf bump. @markmandel is that ok ? does it requires tooling update ?

Kuqd:
This include a fix for Prometheus tags, which was happening in my branch, @pjayara-g it would be great to also remove any tag value "none" as empty was not supported before.

go.mod
markmandel: Good catch. That's a big jump. If we do this, yes - we should regenerate things, and confirm 

Do we need this update to update opencensus as well? Are they linked dependencies?

I'm almost wondering if this should be split into some separate PRs to make this easier to review?
1. The update to OpenCensus
2. The updates to the test code
3. Anything else?

Would that be possible?

Chrome actually crashes on this page :disappointed: I should fire up Firefox, it usually works.

I'll also dig through this this week too.

pjayara-g: I suspect they are linked dependencies. The only change I had made was . to upgrade to 0.22. The rest came from running go mod vendor and go mod tiny.

I cannot split the test changes. One consequence of moving to 0.22 is that @Kuqd 's hack of unregistering/registering no longer works reliably.

Kuqd: I'll would like to have first a grpc/protobuf bump to the version you need.

markmandel: That seems like a good strategy to me :+1: grpc is annoyingly backward compatible, so i expect nothing will break even if we update only one side.

pjayara-g: FWIW, if grpc or genproto are not backwards compatible than a lot of things would break. Both libraries need to maintain backwards compatibility to be useful since clients and servers cannot be updated at the same time.

pkg/metrics/controller_test.go
```diff
@@ -15,21 +15,94 @@
...
+func serialize(args []string) string {
+	return strings.Join(args, "|")
+}
+
+type expectedMetricData struct {
+	labels []string
+	val    interface{}
+}
+
+func verifyMetricData(exporter *metricExporter, metricName string, expected []expectedMetricData) error {
```
markmandel: this could be:
```
func assertMetricData(t *test.T, exporter *metricExporter, metricName string, expected []expectedMetricData) bool
```

And then you could use the `assert` library all the way through, and wouldn't need to use reflect, and if there is a difference between deep structs, you will get nicer reporting out of the test. WDYT?

pkg/metrics/controller_test.go
```diff
@@ -40,29 +113,32 @@ func TestControllerGameServerCount(t *testing.T) {
...
-	report()
-
-	assert.Nil(t, testutil.GatherAndCompare(registry, strings.NewReader(gsCountExpected), "agones_gameservers_count"))
+	reader.ReadAndExport(exporter)
+	err := verifyMetricData(exporter, "gameservers_count", []expectedMetricData{
+		{labels: []string{"test-fleet", "Ready"}, val: int64(0)},
+		{labels: []string{"test-fleet", "Shutdown"}, val: int64(1)},
+		{labels: []string{"none", "PortAllocation"}, val: int64(2)},
+	})
+	assert.Nil(t, err)
```
markmandel: If you keep returning an error, `assert.NoError` is better, since if there is an error, it returns the error's message.

go.mod
markmandel: So I would agree with @Kuqd - I think the right step would be:

1. Update gRPC to the version required here (one PR)
1. Then (another PR) one of us can regenerate the SDK toolchain, and make sure that all work appropriately.
1. Then we can apply the opencensus update and changes (another PR)

That should make it a nicer, and safer process. WDYT?

pjayara-g: Sure thing.

PR #894, opened 2019-07-10
Move the stable api group and promote it to v1

### LINKED ISSUE #703: Proposal: Split up the api group stable.agones.dev

roberthbailey:
I wanted to get the first part of this out for folks to look at before I go through all of the docs and install bits and get them updated. I've done the package move and version change and fixed enough of the code to get the unit tests to pass. 

Part of #703.


roberthbailey:
The newest push has the a bunch of the remaining bits updated (I think all that remains is the docs). I've done a first pass on the SDKs, but I still need to figure out how to test them more thoroughly, since it seems like `make test-sdks` only tests the go and node SDKs.

markmandel:
Had a quick look through - approach looks good. Not seeing any red flags on my end :+1:

site/static/agones.cast
```diff
@@ -29,7 +29,7 @@
 [5.890398, "o", "\u001b[?1l\u001b>\u001b[?2004l\r\r\n"]
 [5.891151, "o", "\u001b]2;kubectl apply -f fleet.yaml\u0007\u001b]1;kubectl\u0007"]
 [5.891306, "o", "\u001b]133;C;\u0007"]
-[6.329093, "o", "fleet.stable.agones.dev/simple-udp created\r\n"]
+[6.329093, "o", "fleet.agones.dev/simple-udp created\r\n"]
```
roberthbailey: I'm not sure whether the feature flag things work in this file....

site/content/en/docs/Guides/access-api.md
```diff
@@ -236,7 +293,6 @@ $ curl http://localhost:8001/apis/stable.agones.dev/v1alpha1/namespaces/default/
 
 # allocate a gameserver from a fleet named 'simple-udp', with GameServerAllocation
 
-{{% feature expiryVersion="0.12.0" %}}
```
roberthbailey: This looks a bit funny, but I split all of the new stuff into a different block so that it's easier to see what changed.

site/static/agones.cast
Kuqd: this file is interpreted by javascript to render the cast.

roberthbailey: I wasn't sure how to fix it properly. One option is just put this into a separate PR and merge it right before we cut 0.12.0 so that the website flips is only different from the latest release for a short period of time.

site/content/en/docs/Tutorials/allocator-service-go.md
```diff
@@ -4,7 +4,12 @@ linkTitle: "Create an Allocator Service (Go)"
 date: 2019-01-03T03:15:34Z
 description: >
   This tutorial describes how to interact programmatically with the 
+{{% feature expiryVersion="0.12.0" %}}
```
roberthbailey: It looks like I can't use the curly braces in the yaml part here, since it causes the tests to fail. So I'm going to leave this with the v1alpha1 version and we'll need to change it around the time we cut the 0.12.0 release.

site/static/agones.cast
roberthbailey: I've moved this change into https://github.com/googleforgames/agones/pull/911.

Kuqd: awesome.

roberthbailey:
> LGTM. Should we merge tomorrow ?

I'm wondering if we should get @markmandel's PR for the Reserve() SDK in first, since this is going to conflict and I'm happy to do the resolving on my end. Once that one is in I don't see nearly as many open PRs that will conflict so it'd probably be a good time to do it.

roberthbailey:
Locally, `make test` is failing with

```
...
site/docs/examples/index.html
  Non-OK status: 404 --- site/docs/examples/index.html --> https://godoc.org/agones.dev/agones/pkg/client/clientset/versioned/typed/agones/v1
✘✘✘ failed in 5.7732241s
1 errors in 73 documents
make: *** [site-test] Error 1
```

But since that URL won't exist until later, I'm not sure how to get it to pass.

markmandel:
> Locally, `make test` is failing with
> 
> ```
> ...
> ========================================================================
> site/docs/examples/index.html
>   Non-OK status: 404 --- site/docs/examples/index.html --> https://godoc.org/agones.dev/agones/pkg/client/clientset/versioned/typed/agones/v1
> ========================================================================
> ✘✘✘ failed in 5.7732241s
> 1 errors in 73 documents
> make: *** [site-test] Error 1
> ```
> 
> But since that URL won't exist until later, I'm not sure how to get it to pass.

Add `data-proofer-ignore` attribute to the link, and it will be skipped.

https://github.com/wjdp/htmltest

roberthbailey:
I can reproduce the e2e failures locally. The issue is that `gcr.io/agones-images/udp-server:0.11` needs to be updated to a version that uses the new sdk. I've push a recompiled version into my own project based on the code in my branch and verified that all e2e tests pass for me. 

The question is: How should I updated the one in the `agones-images` project? Should I just build 0.12 from my branch? Should we only build and push new images to the shared project after code has been merged (in which case, we can temporarily change the parameter to use my image instead of the official one, and then toggle it back after pushing an official one once this PR merges)?

How has this been handled in the past?

Also, this points to the fact that this is a majorly breaking change. It isn't just that the api groups are changing; anyone that has compiled in the game server sdk will need to recompile to upgrade to this version of agones, meaning that it will be nearly impossible to upgrade from 0.11 to 0.12 in the same k8s cluster.

markmandel:
I'd love to know why this is needing the SDK update? (I haven't dug into this myself yet).

In the past, I've manually recompiled and pushed new versions of udp-server up into agones-images on an as needed basis (in fact, there may be a 0.12 there now from some other work I was doing, feel free to overwrite). Sounds like we need a more formal process (or make it part of our cloud build?)

I think for now to unstick us, create a 0.12 manually and push it up, and we'll come up with a more formal process later. WDYT?

roberthbailey:
> I'd love to know why this is needing the SDK update? (I haven't dug into this myself yet).

I updated the package name in the sdk proto along with the metadata prefix so that the naming conventions are consistent across the project. We can choose not to do this if it's important to keep the sdk stable, but I think we either break it now or never.

markmandel:
Ah of course. Then I advocate we break it now. :+1:

roberthbailey:
I added a commit that I think will make e2e pass, but I should also run through the other examples and see if they need a recompile / image update as well.

markmandel:
Good point - they probably do.

roberthbailey:
The latest build failed during the helm upgrade

```
Error: UPGRADE FAILED: MutatingWebhookConfiguration.admissionregistration.k8s.io "agones-mutation-webhook" is invalid: webhooks[0].clientConfig.url: Required value: exactly one of url or service is required && ValidatingWebhookConfiguration.admissionregistration.k8s.io "agones-validation-webhook" is invalid: webhooks[0].clientConfig.url: Required value: exactly one of url or service is required
make: *** [install] Error 1
Makefile:277: recipe for target 'install' failed
```

I'm going to restart it and see if we can get a signal from the sdk conformance tests.

aLekSer:
Regarding conformance tests I don't see any issues. Let see if they pass on Cloud Build.

roberthbailey:
The conformance tests are passing locally for me; hopefully they also pass in CI.

markmandel:
xug5ger

roberthbailey:
Regarding:

> I think for now to unstick us, create a 0.12 manually and push it up, and we'll come up with a more formal process later. WDYT?

I've created https://github.com/googleforgames/agones/issues/917. And in the meantime, I've rebuilt the container image from the master branch and pushed it.

PR #896, opened 2019-07-10
Update kind dev tooling to 1.12

roberthbailey:
Part of #717.

A couple of notes:
1. The `go get` install instructions didn't work for me, so instead of trying to keep our instructions up to date I just replaced it with a link to the Kind docs (it turns out they switched to go modules, which was why). 
1. The current version of kind (v0.4.0) doesn't successfully create clusters at 1.11.3 any longer so that was broken. 
1. I tried to use 1.12.10 (the latest 1.12 release) but kind only seems to support 1.12.9.

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -298,7 +298,7 @@ Running end-to-end tests on minikube is done via the `make minikube-test-e2e` ta
 Because Kind runs on a docker on the host, some of the standard build and development Make targets
 need to be replaced by kind specific targets.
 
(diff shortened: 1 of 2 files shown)


markmandel:
I'll take this for a spin, but @Kuqd did you also want to look, since KinD was your baby?

roberthbailey:
Kind does not appear to have support for type loadbalancer (see https://github.com/kubernetes-sigs/kind/issues/411) so it'd be great to get your change upstreamed.

PR #897, opened 2019-07-10
Update GKE documentation to 1.12

roberthbailey:
Part of #717.

Notes:
1. As in https://github.com/googleforgames/agones/pull/895, the role binding is no longer required with Kubernetes 1.12+ so I've removed the part. 
1. I've removed the flag to disable basic auth now that it is just setting the default behavior.


markmandel:
Not sure if for this PR, but I assume this should also mean it should also be removed from here?
https://github.com/googleforgames/agones/blob/master/build/includes/google-cloud.mk#L68

roberthbailey:
> Not sure if for this PR, but I assume this should also mean it should also be removed from here?
https://github.com/googleforgames/agones/blob/master/build/includes/google-cloud.mk#L68

Might as well remove it since we are switching to 1.12. If we don't do it now, it'll probably be forgotten. I've removed it and pushed an updated branch.

PR #898, opened 2019-07-11
Fix minikube developer experience

markmandel:
There were some missing attributes and images from the minikube
targets so allocator was not starting up.

This puts them in, and now everything works again!


roberthbailey:
lgtm

Issue #718, opened 2019-04-17
sdk/cpp cmake build fails on Linux


devjgm:
This issue came up during #713, but I'm moving the relevant details into an issue so it can be tracked separately.

My system:
*  Debian Linux, rolling release `#1 SMP Debian 4.19.28-2rodete1 (2019-03-18 > 2018)`
* cmake version 3.12.1
* grpc is not installed locally on my machine

Here are the commands I ran in this repo and the output:
```shell
$ cd sdks/cpp/
$ cd build_scripts/
$ bash build.sh 
-- git Version: v0.0.0-dirty
-- Version: 0.0.0
-- Performing Test HAVE_STD_REGEX
-- Performing Test HAVE_STD_REGEX -- success
-- Performing Test HAVE_GNU_POSIX_REGEX
-- Performing Test HAVE_GNU_POSIX_REGEX -- failed to compile
-- Performing Test HAVE_POSIX_REGEX
-- Performing Test HAVE_POSIX_REGEX -- success
-- Performing Test HAVE_STEADY_CLOCK
-- Performing Test HAVE_STEADY_CLOCK -- success
CMake Error at CMakeLists.txt:177 (install):
  install TARGETS given target "address_sorting" which does not exist in this
  directory.


CMake Error at CMakeLists.txt:195 (install):
  install TARGETS given target "address_sorting" which does not exist in this
  directory.


-- Configuring incomplete, errors occurred!
See also "/home/jgm/github/devjgm/agones/sdks/cpp/.build/CMakeFiles/CMakeOutput.log".
See also "/home/jgm/github/devjgm/agones/sdks/cpp/.build/CMakeFiles/CMakeError.log".
make: *** No rule to make target 'install'.  Stop.
```

markmandel:
Seem like that should be called from within the Makefile that we use for testing?
https://github.com/GoogleCloudPlatform/agones/blob/master/sdks/cpp/Makefile#L25

@dgkanatsios  ?

dsazonoff:
For testing we are using sdks/cpp/Makefile, but it is very similar. It looks like gRPC was downloaded by CMake from git without updating submodules. But I'm not sure, take a look a bit later.
Tested both scripts on Mac, it works. Same as raw CMake.

@devjgm thanks for the feedback. Please, also add a version of git that you use.

devjgm:
`git version 2.21.0`

aLekSer:
I have tested that following set of commands works on Debian linux:
```
[...]

PR #903, opened 2019-07-13
Changed cmake required version to 3.5 to work on more machines

devjgm:
The current cmake files don't actually *require* 3.13, so lowering the version they claim to require will allow the cmake files to work on more machines. Related to #718


dsazonoff:
I don't think that cmake version should be downgraded. It is very easy to upgrade it. And SDK supports docker build, so it doesn't matter for a machine what version to have.
There was a request from @markmandel to silent CMake output and there was a bug in older versions of CMake when output was not silenced for some platforms. If necessary, I can find an issue in cmake bugtracker. But it is not critical.

Don't forget to patch 'cpp/cmake/clang-verify.in' file and Dockerfile that are responsible for SDK builds (installation of CMake).

devjgm:
> I don't think that cmake version should be downgraded. It is very easy to upgrade it.

My distro is using a debian mirror that's slightly behind, and the newest cmake I can get is 3.12.1. So my only choice is to either: (A) download cmake from kitware.com (yuck), or (B) edit the "cmake_minimum_required" version to something slightly lower, which works fine.

FTR, we choose to require cmake 3.5 for https://github.com/googleapis/google-cloud-cpp because it's *easily* available on a large variety of important Linux distros, and it actually has most of the major modern features that people want/need with cmake. 
 
> And SDK supports docker build, so it doesn't matter for a machine what version to have.

Are there instructions for that? Because running `make`, as I did, used my locally installed cmake and did not use docker.

> Don't forget to patch 'cpp/cmake/clang-verify.in' file and Dockerfile that are responsible for SDK builds (installation of CMake).

I'm sorry, but I don't see any mention of `clang-verify.in`. Can you please provide a few more details or point me at the documentation? Thanks.

dsazonoff:
Now I agree about downgrading cmake version.
>Are there instructions for that? Because running make, as I did, used my locally installed cmake and did not use docker.

You may look [here](https://github.com/googleforgames/agones/tree/master/build/build-sdk-images). Or [here](https://github.com/googleforgames/agones/blob/master/examples/cpp-simple/Makefile). But I usually build everything locally, without docker.
I was talking about this [Dockerfile](https://github.com/googleforgames/agones/blob/master/build/build-sdk-images/cpp/Dockerfile#L21). I was wrong in my previous comment. I think we should downgrade cmake version in CMakeLists.txt files, but keep use v3.14 in docker build. So minimal supported version will be 3.5, but cloud build will use 3.14.

>I'm sorry, but I don't see any mention of clang-verify.in

I noticed that you created your PR before #855 was merged.

Please, update from master and fix cmake version in [clang-verify](https://github.com/googleforgames/agones/blob/master/sdks/cpp/cmake/clang-verify.in#L15) too.
Other things are fine.

devjgm:
Updated. Requiring 3.5 but using 3.14 in Docker sounds like a good idea to me.

roberthbailey:
@devjgm - should we also update https://github.com/googleforgames/agones/blob/master/examples/cpp-simple/CMakeLists.txt to specify cmake 3.5?

devjgm:
I think so, yes. I didn't know "examples/cpp-simple" existed, otherwise I'd have updated it. But I see no reason for it to be different. Thanks for pointing this out.

PR #904, opened 2019-07-13
fix: Fix gRCP context leaks.

devjgm:
The calls to gRPC methods were each allocating new `ClientContext`
object, then leaking them. ClientContext ownership is retained by the
caller, so it's the caller's responsibility to delete them. The normal
way this is done is to declare a local stack-allocated context, and pass
a pointer to that.

I also did a few small C++-y cleanups, such as the following:

*  Removed the public declaration for `SDKImpl` by making it a private
   internal class. Also, declare this as a struct rather than class
   since it had all public fields.

*  Removed the kPort int. It was never needed as an int, only a string.
   And it was only ever used in one place. The Google style guide says
   that data should be declared in the smallest scope possible, so
   declaring that at file-scope was too broad. In this case, the best
   solution is to just declare the grpc "target" as a string inline
   since it's needed in exactly one place.

*  Using `std::move()` to efficiently move (not copy) the string value
   function arguments.

*  Fixed formatting in .cc file to match .h file by putting the pointer
   asterisk adjacent to the type (this matches the .h file).

I verified that all the code still compiles by running `make`, but I
have not run any tests because AFAICT, there are no unit tests.


markmandel:
@dsazonoff any issues?

sdks/cpp/src/agones/sdk.cc
```diff
@@ -47,84 +43,85 @@ bool SDK::Connect() {
...
-  context->set_deadline(gpr_time_add(gpr_now(GPR_CLOCK_REALTIME),
-                                     gpr_time_from_seconds(30, GPR_TIMESPAN)));
+  grpc::ClientContext context;
+  context.set_deadline(gpr_time_add(gpr_now(GPR_CLOCK_REALTIME),
+                                    gpr_time_from_seconds(30, GPR_TIMESPAN)));
 
   stable::agones::dev::sdk::KeyValue request;
-  request.set_key(key);
-  request.set_value(value);
+  request.set_key(std::move(key));
```
dsazonoff: std::move is redundant here and in next calls of `set_key` and `set_value` because it accepts argument via const reference.

sdks/cpp/src/agones/sdk.cc
```diff
@@ -47,84 +43,85 @@ bool SDK::Connect() {
...
+  request.set_value(std::move(value));
```
dsazonoff: Remove std::move

sdks/cpp/src/agones/sdk.cc
```diff
@@ -47,84 +43,85 @@ bool SDK::Connect() {
...
-  context->set_deadline(gpr_time_add(gpr_now(GPR_CLOCK_REALTIME),
-                                     gpr_time_from_seconds(30, GPR_TIMESPAN)));
+  grpc::ClientContext context;
+  context.set_deadline(gpr_time_add(gpr_now(GPR_CLOCK_REALTIME),
+                                    gpr_time_from_seconds(30, GPR_TIMESPAN)));
 
   stable::agones::dev::sdk::KeyValue request;
-  request.set_key(key);
-  request.set_value(value);
+  request.set_key(std::move(key));
```
dsazonoff: Remove std::move

sdks/cpp/src/agones/sdk.cc
```diff
@@ -47,84 +43,85 @@ bool SDK::Connect() {
...
+  request.set_value(std::move(value));
```
dsazonoff: Remove std::move

sdks/cpp/src/agones/sdk.cc
devjgm: Since C++11, protos also have setters that take rvalues to support efficient moves. To quote from [this doc](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated):

> `void set_foo(string&& value)` (C++11 and beyond): Sets the value of the field, moving from the passed string. After calling this, foo() will return a copy of value.

So I think we want to do one of the following:

A: Keep these moves, or ...
B: Change the function's argument types to be `const std::string&` and not move. 

I recommend (A).

sdks/cpp/src/agones/sdk.cc
devjgm: See previous reply.

sdks/cpp/src/agones/sdk.cc
devjgm: See previous reply.

sdks/cpp/src/agones/sdk.cc
devjgm: See previous reply.

devjgm:
I *think* the build failure is an unrelated flake.

dsazonoff:
@markmandel could you restart a build, please. Code is Ok.

PR #911, opened 2019-07-15
Documentation updates to apply just prior to cutting the 0.12.0 release.

roberthbailey:


roberthbailey:
Marking as ready for review so that this can be merged when appropriate (not sure if that is with the release candidate or the actual release).

markmandel:
I was thinking to merge it before the stable release goes through - since it matches the experience and documentation that will exist once once stable has landed. WDYT?

roberthbailey:
sgtm.

markmandel:
Updating, so we can merge and release this :+1:

PR #908, opened 2019-07-15
Add ensure-sdk-image when run single SDK test

aLekSer:
Fixing issue of SDK conformance make targets.

PR #913, opened 2019-07-16
Fix Rust SDK conformance flakiness, add clean step

### LINKED ISSUE #912: Building the rust sdk leaves untracked files
**What happened**:  Running `make build` results in

```
$ git st
...
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	test/sdk/rust/Cargo.lock
	test/sdk/rust/target/
```

**What you expected to happen**: No untracked files after running a build

**How to reproduce it (as minimally and precisely as possible)**: `make build`

I'm wondering if these files should be inside of the agones-build-sdk-rust docker image instead of on my host machine. Alternatively, they could be added to the `.gitignore`. 

**Environment**:
- Agones version: master branch
- Kubernetes version (use `kubectl version`): _n/a_
- Cloud provider or hardware configuration: Building locally on Mac OS X 10.14.5
- Install method (yaml/helm): _n/a_
- Troubleshooting guide log(s): _n/a_
- Others:

/cc @aLekSer

aLekSer:
Add build and clean step even if we run tests not in a test suite, but
as a single test.
Add additional target `run-sdk-conformance-no-build` which would be useful for debugging with prebuilt SDKs.
`make build` now does not trigger Rust and NodeJS SDK builds.

Closes #912 . 
Contains #908 .


markmandel:
Pretty sure I know why this `stat` error keeps rearing its head.

Working on it, and will create a PR shortly.

markmandel:
Once #919 is working, this should solve this problem.

Basically this test was deleting files, while the e2e process was running `cp -r` and it caused all kind of issues.

This only made worse a previously more rare issue (gocache moving under our feet during the same thing), that caused the same error.

PR #920, opened 2019-07-17
Compliance tests for Reserve(seconds).

markmandel:
Currently only running in the golang test, as that's the only one that implements SDK.Reserve(seconds).

Added a couple of cleanup items in the conformance test makefile as well.

Work on #660

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -111,18 +111,18 @@ ensure-build-sdk-image:
 
 # Run SDK conformance Sidecar server in docker in order to run
 # SDK client test against it. Useful for test development
-SECONDS ?= 30
+run-sdk-conformance-local: TIMEOUT ?= 30
 run-sdk-conformance-local: ensure-agones-sdk-image
 	docker run -e "ADDRESS=" -p 59357:59357 \
-	 -e "TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" -e "TIMEOUT=$(SECONDS)" $(sidecar_tag)
+	 -e "TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" -e "TIMEOUT=$(TIMEOUT)" $(sidecar_tag)
 
 # Run SDK conformance test, previously built, for a specific SDK_FOLDER
-SECONDS ?= 30
-run-sdk-conformance-no-build: ensure-agones-sdk-image
+run-sdk-conformance-no-build: TIMEOUT ?= 30
+run-sdk-conformance-no-build: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch
 run-sdk-conformance-no-build: ensure-build-sdk-image
 	sleep 2s && DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
 	docker run -p 59357:59357 -e "ADDRESS=" \
-	 -e "TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" -e "TIMEOUT=$(SECONDS)" --net=host  $(sidecar_tag)
+	 -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" --net=host  $(sidecar_tag)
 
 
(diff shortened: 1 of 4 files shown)


markmandel:
@aLekSer you happy with these changes?

test/sdk/go/sdk-client-test.go
```diff
@@ -90,8 +90,11 @@ func main() {
 	if err != nil {
 		log.Fatalf("Could not set annotation: %s", err)
 	}
+	if err = sdk.Reserve(5 * time.Second); err != nil {
```
roberthbailey: nit: should be `if err := sdk.Reserve(...); ...`

build/includes/sdk.mk
```diff
@@ -111,25 +111,25 @@ ensure-build-sdk-image:
...
-	 -e "TEST=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch" -e "TIMEOUT=$(SECONDS)" --net=host  $(sidecar_tag)
+	 -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" --net=host  $(sidecar_tag)
 
 # Run a conformance test for all SDKs supported
 run-sdk-conformance-tests: ensure-agones-sdk-image
 	$(MAKE) run-sdk-command SDK_FOLDER=node COMMAND=build
 	$(MAKE) run-sdk-command SDK_FOLDER=rust COMMAND=build
 	$(MAKE) run-sdk-conformance-test SDK_FOLDER=node
-	$(MAKE) run-sdk-conformance-test SDK_FOLDER=go
+	$(MAKE) run-sdk-conformance-test SDK_FOLDER=go TESTS=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch,reserve
```
roberthbailey: Does the reserve test currently fail for the node & rust sdks? If so, are you going to fix that in a follow-up PR?

test/sdk/go/sdk-client-test.go
markmandel: I would respectfully disagree. If I do that, `err` becomes a shadow variable of the previously defined `err` variable.

build/includes/sdk.mk
markmandel: Reserve is not implemented in the other SDKs (yet). I can take on the languages that I know how to write. 

On a broader scope, there is missing functionality from some of the SDKs. In the past, I've started with the Go SDK, and the community has written the updates to the other languages - but we might want to take a review, and fill in any gaps we find.

test/sdk/go/sdk-client-test.go
roberthbailey: My experience with go idioms is to pretty much always scope variables as narrowly as possible, which often causes `err` (since it is a common name) to be shadowed in longer functions. I find this to be perfectly normal, but I can see in a quick google search that other people find it super confusing so I'm ok if you want to leave this as-is.

build/includes/sdk.mk
roberthbailey: So... to answer my original question, it sounds like the conformance test do fail in the other sdks (indicating that they haven't implemented Reserve), right?

Can you open up an issue to audit the existing SDKs for gaps and we can then prioritize getting them up to snuff prior to the 1.0 release.

test/sdk/go/sdk-client-test.go
markmandel: Also my editor complains when I do this :smile:

build/includes/sdk.mk
markmandel: Yes, 100% correct.

I'll do a review, and write up a ticket right now :+1:

markmandel: Done - #927

test/sdk/nodejs/node_modules/agones
```diff
@@ -0,0 +1 @@
+../../../../sdks/nodejs
```
roberthbailey: Should this file (and the next one) but added? They appear when I run tests but I'm not sure whether they should be checked in.

markmandel: oops! probably not. Let me go back and remove them.

Should I .gitignore them now, since I'm in there?

roberthbailey: That sounds like a good idea.

aLekSer:
Thanks for improving make targets.

PR #922, opened 2019-07-18
Add new GameServer lifecycle diagrams for Reserved

markmandel:
Shows how the workflow can exist for 1st party/distributed matchmakers.

Work on #660


build/Makefile
```diff
@@ -392,8 +392,6 @@ gen-crd-client: $(ensure-build-image)
 	docker run --rm $(common_mounts) -w $(workdir_path) $(build_tag) /root/gen-crd-client.sh
 	docker run --rm $(common_mounts) -w $(workdir_path)/pkg $(build_tag) goimports -w .
 
-site-images: $(site_path)/static/diagrams/gameserver-states.dot.png $(site_path)/static/diagrams/gameserver-lifecycle.puml.png
-
 # generate pngs from dot files
```
roberthbailey: Should this part of the makefile move as well? Are there places outside of the website where we generate pngs from dot files and puml files?

site/static/diagrams/gameserver-reserved.puml
```diff
@@ -0,0 +1,74 @@
...
+  available for, so it will not be
+  scaled down
+end note
+Binary -> Matchmaker: Register()
+note right
+  Register with the Matchmaker as
+  available to host a game session
+end note
+note left
+match maker registration
```
roberthbailey: Start with a capital to be consistent with the other notes.

roberthbailey: Is match maker one work (e.g. matchmaker)?

site/static/diagrams/gameserver-reserved.puml
```diff
@@ -0,0 +1,74 @@
...
+  the configured threshold.
+end note
+GameServer <-- SDK: Maintains Healthy status
+
+== Match Maker Registration ==
+group loop
+Binary -> SDK: SDK.Reserve(duration)
+GameServer <-- SDK: Set state to <i>Reserved</i> for duration
+note right
+  <i>duration</i> should be longer
```
roberthbailey: Start with a capital to be consistent with the other notes.

build/Makefile
markmandel: That is true that we don't - but it seemed general enough that maybe leave it where it was.

Honestly, I've no strong opinion on this, so if you feel strongly, I'll move it.

site/static/diagrams/gameserver-reserved.puml
markmandel: Yep, good catch.

markmandel:
Changes made, per comments :+1:

build/Makefile
roberthbailey: If we are still thinking about moving the website to a different repo after 1.0 I'd move it to the site makefile now to make that simpler.

markmandel: Makes perfect sense. Will make the move :+1:

PR #925, opened 2019-07-18
E2E tests for SDK.Reserve(seconds)

markmandel:
Pushes a new simple-udp version (0.13)

Work on #660


roberthbailey:
This should go in before https://github.com/googleforgames/agones/pull/928 since that bumps the image number to 0.14.

PR #926, opened 2019-07-18
Flaky: Make preview --no-promote to avoid promotion race conditions

markmandel:
This sometimes breaks CI, when two app engine apps try and
be promoted at the same time.


markmandel:
More context for anyone not familiar with app engine:

PR #928, opened 2019-07-19
Update K8s Code Generation Tooling to 1.12

markmandel:
Also regenerated the CRD clients with the updated tooling.

Needed to build a udp-server:0.14

Work for #717

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -44,12 +44,18 @@ RUN go get -u golang.org/x/tools/cmd/goimports
 # RUN gcloud components update
 RUN gcloud components update && gcloud components install app-engine-go
 
+# the kubernetes version for the file
+ENV KUBERNETES_VER 1.12.10
+
 # overwrite kubectl as we want a specific version
-ENV KUBECTL_VER 1.12.10
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VER}/bin/linux/amd64/kubectl && \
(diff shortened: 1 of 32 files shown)


build/build-image/Dockerfile
```diff
@@ -44,13 +44,19 @@ RUN go get -u golang.org/x/tools/cmd/goimports
...
 # overwrite kubectl as we want a specific version
-ENV KUBECTL_VER 1.12.10
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VER}/bin/linux/amd64/kubectl && \
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBERNETES_VER}/bin/linux/amd64/kubectl && \
     chmod go+rx ./kubectl && \
     mv ./kubectl /usr/local/bin/kubectl
 RUN echo "source <(kubectl completion bash)" >> /root/.bashrc
 
+# install the release branch of the code generator tools
+RUN mkdir -p /go/src && cd /go/src && mkdir -p k8s.io && cd k8s.io && \
```
roberthbailey: I realize it was written this way before, but I don't understand the multiple `mkdir` commands, `mkdir -p` will create intermediate directories so a single `mkdir -p /go/src/k8s.io && cd /go/src/k8s.io \` should do the trick.

markmandel: Oooh! Good point. I'll make this change and then merge.

PR #932, opened 2019-07-20
With the extra whitespace the rendered output has undesirable new paragraphs.

roberthbailey:
Fixes a bug I introduced in https://github.com/googleforgames/agones/pull/930. 

Part of #703.


markmandel:
Changes on the preview branch look like this:
image

I assume that wasn't quite what you were going for? :smile:

roberthbailey:
Uh, yeah that didn't work as I'd expected. Shows what happens when I don't try it locally....

I tried a few different ways to do the whitespace and couldn't figure one out that kept the paragraphs together so the newer version is better (doesn't have the weird formatting) but isn't really what I'd like it to look like either. But it's better than what is live so I think we should commit it unless you have some tricks for fighting with the hugo markdown.

PR #933, opened 2019-07-22
Update GameServer state diagram with Reserved

markmandel:
Update dot and generated PNG with the Reserved state, and resultant
flow from there.

Tried to keep it as simple as possible, while still representing
potential state changes.

Should be final work on #660 except for missing SDK functions.


roberthbailey:
There are a couple of places in gameserver.md where I mis-typed `{% feature publishversion="0.12.0" %}}` (note the lowercase 'v' in version) which is causing the currently live site to look wrong. Can you fix those while you are in that file anyway?

site/static/diagrams/gameserver-states.dot
```diff
@@ -38,6 +44,7 @@ digraph {
     RequestReady -> Unhealthy [ color="purple" ]
```
roberthbailey: The Scheduled -> Unhealthy arrow should be purple, right? It's on line 43 in the new file.

site/static/diagrams/gameserver-states.dot
```diff
@@ -23,9 +24,14 @@ digraph {
...
     Creating -> Starting [ label="created pod", color="blue" ]
     Starting -> Scheduled [ label="we have a pod, fetch its address", color="blue" ]
+    Scheduled -> Reserved [ label="SDK.reserved(seconds)", color="purple" ]
     Scheduled -> RequestReady [ label="SDK.ready()", color="purple" ]
     RequestReady -> Ready [ label="ready to be allocated", color="blue" ]
+    Reserved -> Ready [label="if seconds > 0  \nand failed to call  \nSDK.allocate()  " color="purple"]
+
     Ready -> Allocated [ label="allocated for use", color="orange" ]
+    Ready -> Allocated [ label="SDK.allocate()   ", color="purple" ]
+    Reserved -> Allocated [color="purple"]
```
roberthbailey: There isn't a line from Reserved -> Shutdown. It seems like there is a state transition from every state to both Unhealthy and to Shutdown, but this one is missing (shutdown only has 4 incoming arrows and unhealthy has 5).

site/static/diagrams/gameserver-states.dot
```diff
@@ -23,9 +24,14 @@ digraph {
     PortAllocation -> Creating [ label="allocated unused port", color="blue" ]
     Creating -> Starting [ label="created pod", color="blue" ]
     Starting -> Scheduled [ label="we have a pod, fetch its address", color="blue" ]
+    Scheduled -> Reserved [ label="SDK.reserved(seconds)", color="purple" ]
     Scheduled -> RequestReady [ label="SDK.ready()", color="purple" ]
     RequestReady -> Ready [ label="ready to be allocated", color="blue" ]
+    Reserved -> Ready [label="if seconds > 0  \nand failed to call  \nSDK.allocate()  " color="purple"]
```
roberthbailey: It's difficult to tell from the diagram which arrow this note belongs to. Maybe there isn't any avoiding that though...

site/static/diagrams/gameserver-states.dot
markmandel: Yeah, I wondered about that too. Fixed.

site/static/diagrams/gameserver-states.dot
markmandel: Tweaked some alignment, think it looks a bit better. PTAL.

PR #934, opened 2019-07-22
Updates to the C++ SDK along with the simple example that exercises it.

roberthbailey:
Part of https://github.com/googleforgames/agones/issues/703.

I ran `clang-format` on `examples/cpp-simple/server.cc` which accounts for many of the changes. The only code changes were modifying `stable::agones::...` to `agones::...`. 

Both https://github.com/googleforgames/agones/pull/855 and https://github.com/googleforgames/agones/pull/904 broke the cpp-simple example code (which points to the need for us to have sdk conformance tests for C++):

- In the former, `include/agones/sdk.grpc.pb.h` was no longer being installed with the sdk, which caused the sample code to no longer compile. I've restructured the cmake files a bit to make sure that it's now put into the expected place. /cc @dsazonoff

- In the latter, the client context for the healthcheck calls was allocated on the stack in the constructor, but the healthchecks were continuously being called via the `Health` SDK function, which caused the program to abort abnormally due to a check failure. The code is now leaking a single client context, which we could clean up in the destructor if we wanted to be especially thorough (but in generally, since it has the same lifetime of the SDK and we expect the SDK to have the lifetime of the game server, this seemed not so bad compared to leaking one on each SDK call). /cc @devjgm 

Note that I haven't pushed the 0.6 version of the container yet, since I have some questions about whether the updates I made to the SDK are correct.


examples/cpp-simple/README.md
```diff
@@ -8,9 +8,24 @@ It will
 - Every 10 seconds, write a log saying "Hi! I'm a Game Server"
 - After 60 seconds, call `SDK::Shutdown()` to shut the server down.
 
+## Running
+
+If you want to run the example code without building the container yourself, you can start a single gameserver by running
+
+```
+kubectl create -f https://raw.githubusercontent.com/googleforgames/agones/master/examples/cpp-simple/gameserver.yaml
```
devjgm: Will this stop the gameserver when it's finished as well?

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,115 @@
...
-    while (true) {
-        if (!sdk->Health()) {
-            std::cout << "Health ping failed" << std::endl;
-        } else {
-            std::cout << "Health ping sent" << std::endl;
-        }
-        std::this_thread::sleep_for(std::chrono::seconds(2));
+  while (true) {
+    if (!sdk->Health()) {
+      std::cout << "Health ping failed" << std::endl;
```
devjgm: It's best to avoid `std::endl` and just use "\n" instead. See https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rio-endl

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,115 @@
...
-    }
-
-    std::cout << "Setting an annotation" << std::endl;
-    status = sdk->SetAnnotation("test-annotation", "test value");
-    if (!status.ok()) {
-        std::cout << "Could not run SetAnnotation(): "+ status.error_message() + ". Exiting!" << std::endl;
+  std::cout << "C++ Game Server has started!" << std::endl;
+
+  std::cout << "Getting the instance of the SDK!" << std::endl;
+  agones::SDK *sdk = new agones::SDK();
```
devjgm: I don't think there's any need to heap-allocate the SDK. I recommend stack allocating it like `agones::SDK sdk;`

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,115 @@
...
+
+  std::cout << "Attempting to connect..." << std::endl;
+  bool connected = sdk->Connect();
+  if (!connected) {
+    std::cout << "Could not connect to the sidecar. Exiting!" << std::endl;
+    return -1;
```
devjgm: Note: With a heap-allocated SDK, this return would leak the SDK. Even though this is main() and we may not care about this leak, leak checkers will care and would report it. So it's best to avoid if possible.

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,115 @@
...
 
-#include <iostream>
-#include <thread>
 #include <agones/sdk.h>
 #include <grpc++/grpc++.h>
+#include <iostream>
+#include <thread>
 
 // send health check pings
 void doHealth(agones::SDK *sdk) {
```
devjgm: I realize this function name existed prior to your PR, but it's casing doesn't follow the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html). This should be spelled `DoHealth`, and similarly it should be `WatchUpdates`.

If you want to fix these now, that's fine. If you'd rather not touch them, that's fine too. I'll let you decide.

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,115 @@
...
+  }
+
+  std::cout << "...marked Ready" << std::endl;
+
+  std::cout << "Getting GameServer details..." << std::endl;
+  agones::dev::sdk::GameServer gameserver;
+  status = sdk->GameServer(&gameserver);
+
+  if (!status.ok()) {
+    std::cout << "Could not run GameServer(): " + status.error_message() +
```
devjgm: Since this code is already using an output stream, avoid the string concatenation (`op+`) and instead just stream each piece, like `std::cout << "Could not run GameServer(): " << status.error_message() << " Exiting!\n";`

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,115 @@
...
+                     ". Exiting!"
+              << std::endl;
+    return -1;
+  }
+
+  std::cout << "GameServer name: " << gameserver.object_meta().name()
+            << std::endl;
+
+  for (int i = 0; i < 10; i++) {
```
devjgm: Is this loop needed? The alternative would be something like:

```cc
// No loop
std::cout << "Sleeping for 1 minute... \n";
std::this_thread::sleep_for(std::chrono::minutes(1));
std::cout << "Shutting down...\n";
auto status = sdk.Shutdown();
...
```

Thoughts?

sdks/cpp/src/agones/sdk.cc
```diff
@@ -43,8 +43,7 @@ bool SDK::Connect() {
 
   // make the health connection
   agones::dev::sdk::Empty response;
-  grpc::ClientContext context;
-  pimpl_->health_ = pimpl_->stub_->Health(&context, &response);
+  pimpl_->health_ = pimpl_->stub_->Health(new grpc::ClientContext(), &response);
```
devjgm: Thanks for fixing this.

If this context needs to live as long as the SDK, I think it should be stored as a SDK data member wrapped in a std::unique_ptr.

examples/cpp-simple/server.cc
roberthbailey: I've fixed them.

examples/cpp-simple/server.cc
roberthbailey: I've replaced all of them with "\n"

examples/cpp-simple/server.cc
roberthbailey: done.

examples/cpp-simple/server.cc
roberthbailey: ack

examples/cpp-simple/server.cc
roberthbailey: I thought the same thing when I first saw it. There might still be a simpler way to re-write it, but what the loop does it to print out a message every 10 seconds (which is useful when you are debugging) and call shutdown after 60 seconds. The loop keeps running until 100 seconds, so you can see the difference between when shutdown is executed and when agones actually cleans up the container. I'm happy to rewrite it, but I think we should discuss in a different PR. Also note that other simple examples (rust at least) basically copied this code and if we change this one we should update the others as well.

sdks/cpp/src/agones/sdk.cc
roberthbailey: My C++ is a bit rusty, so while I _think_ I've done what you suggested, please check it closely to make sure I've done it correctly.

examples/cpp-simple/README.md
roberthbailey: Yes. If you use the fleet example, you will get a constant churning of new game servers as each one exits after ~1m and is replaced by a new one.

examples/cpp-simple/server.cc
roberthbailey: I fixed this here (and elsewhere) while I was updating the newline change.

examples/cpp-simple/README.md
devjgm: Ok, so there's no "shutdown" procedure I'd need to do? I don't know much about kubectl, so I'd run this command and leave it running, so as long as that's OK, great :-)

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,109 @@
...
-    status = sdk->SetAnnotation("test-annotation", "test value");
-    if (!status.ok()) {
-        std::cout << "Could not run SetAnnotation(): "+ status.error_message() + ". Exiting!" << std::endl;
+  std::cout << "C++ Game Server has started!\n";
+
+  std::cout << "Getting the instance of the SDK!\n";
+  agones::SDK sdk;
+
+  std::cout << "Attempting to connect...\n";
+  bool connected = sdk->Connect();
```
devjgm: This won't compile. Since `sdk` is no longer a pointer, you'll need to use `sdk.Connect()`

examples/cpp-simple/server.cc
devjgm: I still think we can make this quite a bit simpler, but it's not urgent.

sdks/cpp/src/agones/sdk.cc
```diff
@@ -41,10 +42,10 @@ bool SDK::Connect() {
 
   pimpl_->stub_ = agones::dev::sdk::SDK::NewStub(pimpl_->channel_);
 
-  // make the health connection
+  // Make the health connection.
   agones::dev::sdk::Empty response;
-  grpc::ClientContext context;
-  pimpl_->health_ = pimpl_->stub_->Health(&context, &response);
+  pimpl_->health_context_ = new grpc::ClientContext();
```
devjgm: How about:

```cc
pimpl_->health_context_ = std::unique_ptr<grpc::ClientContext>(new grpc::ClientContext);
```

examples/cpp-simple/README.md
roberthbailey: I've updated the instructions to explain this, since if you have these questions other folks are likely to as well.

sdks/cpp/src/agones/sdk.cc
roberthbailey: fixed.

examples/cpp-simple/server.cc
roberthbailey: Thanks for catching this.

examples/cpp-simple/server.cc
roberthbailey: So after doing this I ran into an issue where I need to pass the sdk as a pointer into the two functions that run in separate threads. If I try to pass by value/reference, I get this compile error:

```
In file included from /project/cpp-simple/server.cc:22:
/usr/local/include/c++/8.3.0/thread: In instantiation of 'std::thread::thread(_Callable&&, _Args&& ...) [with _Callable = void (&)(agones::SDK*); _Args = {agones::SDK&}; <template-parameter-1-3> = void]':
/project/cpp-simple/server.cc:61:35:   required from here
/usr/local/include/c++/8.3.0/thread:120:17: error: static assertion failed: std::thread arguments must be invocable after conversion to rvalues
  static_assert( __is_invocable<typename decay<_Callable>::type,
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           typename decay<_Args>::type...>::value,
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/include/c++/8.3.0/thread: In instantiation of 'struct std::thread::_Invoker<std::tuple<void (*)(agones::SDK*), agones::SDK> >':
/usr/local/include/c++/8.3.0/thread:132:22:   required from 'std::thread::thread(_Callable&&, _Args&& ...) [with _Callable = void (&)(agones::SDK*); _Args = {agones::SDK&}; <template-parameter-1-3> = void]'
/project/cpp-simple/server.cc:61:35:   required from here
/usr/local/include/c++/8.3.0/thread:250:2: error: no matching function for call to 'std::thread::_Invoker<std::tuple<void (*)(agones::SDK*), agones::SDK> >::_M_invoke(std::thread::_Invoker<std::tuple<void (*)(agones::SDK*), agones::SDK> >::_Indices)'
  operator()()
  ^~~~~~~~
...
```

Now scoping the lifetime of the sdk to the main function makes less sense, since it's being shared by the main thread, the health checking thread, and the thread watching for updates to the gameserver definition. If this were a regular program and this was a function, you'd deallocate the object while retaining a reference and end up in a bad state (since this is main, the program exits so it doesn't really matter). 

What do you think about leaving it this way vs. putting it back to heap allocation? I realize that the `delete sdk` call often isn't called (if the function has any non-success exit) but how much should we care?

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,107 @@
...
-        } else {
-            std::cout << "Health ping sent" << std::endl;
-        }
-        std::this_thread::sleep_for(std::chrono::seconds(2));
+void DoHealth(std::shared_ptr<agones::SDK> sdk) {
+  while (true) {
+    if (!sdk->Health()) {
+      std::cout << "Health ping failed\n";
+    } else {
+      std::cout << "Health ping sent\n";
```
devjgm: (optional) You might consider writing this like:

```cc
while (true) {
  book ok = sdk->Health();
  std::cout << "Health ping " << (ok ? "sent" : "failed) << "\n";
  std::this_thread::sleep_for(...);
}
```

Or similar. The reason being that often times less nesting results in more readable code.

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,107 @@
...
 // watch GameServer Updates
-void watchUpdates(agones::SDK *sdk) {
-    std::cout << "Starting to watch GameServer updates..." << std::endl;
-    sdk->WatchGameServer([](stable::agones::dev::sdk::GameServer gameserver){
-        std::cout << "GameServer Update, name: " << gameserver.object_meta().name() << std::endl;
-        std::cout << "GameServer Update, state: " << gameserver.status().state() << std::endl;
-    });
+void WatchUpdates(std::shared_ptr<agones::SDK> sdk) {
+  std::cout << "Starting to watch GameServer updates...\n";
+  sdk->WatchGameServer([](agones::dev::sdk::GameServer gameserver) {
```
devjgm: Is it intentional to pass a `GameServer` object by value? I don't know if this is an expensive type to copy, but it looks like accepting this object by const-reference may be worthwhile.

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,107 @@
...
+    std::cout << "GameServer Update, name: " << gameserver.object_meta().name()
```
devjgm: (optional) clang-format tip: You an influence the way clang-format wraps lines by using an end-of-line comment. This can be useful when you want to group semantically meaning things on a single line, whereas clang-format will only care about syntax, not semantics. For example, you might consider formatting this output like:

```cc
std::cout << "GameServer update:\n"                                 //
          << "\tname: " << gameserver.object_meta().name() << "\n"  // 
          << "\tstate: " << gameserver.status().state() << "\n";
```

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,107 @@
...
-    std::thread watch (watchUpdates, sdk);
-
-    std::cout << "Setting a label" << std::endl;
-    grpc::Status status = sdk->SetLabel("test-label", "test-value");
-    if (!status.ok()) {
-        std::cout << "Could not run SetLabel(): "+ status.error_message() + ". Exiting!" << std::endl;
+  std::cout << "C++ Game Server has started!\n";
+
+  std::cout << "Getting the instance of the SDK!\n";
+  std::shared_ptr<agones::SDK> sdk = std::make_shared<agones::SDK>();
```
devjgm: There's no need to repeat the type on both sides. Prefer instead `auto sdk = std::make_shared<agones::SDK>();`

examples/cpp-simple/server.cc
```diff
@@ -16,101 +16,107 @@
...
+
+  std::cout << "Attempting to connect...\n";
+  bool connected = sdk->Connect();
+  if (!connected) {
```
devjgm: Since `connected` is only needed in one place, you may want remove that variable and inline the Connect() call in the if. `if (!sdk->Connect()) ...` That removes a variable from the scope of main(), which is nice.

roberthbailey:
@devjgm - this is ready for review.

examples/cpp-simple/server.cc
roberthbailey: done.

examples/cpp-simple/server.cc
roberthbailey: I'm not sure. I've opened https://github.com/googleforgames/agones/issues/941 to discuss outside of this PR.

examples/cpp-simple/server.cc
roberthbailey: Neat trick. Done.

examples/cpp-simple/server.cc
roberthbailey: done.

examples/cpp-simple/server.cc
roberthbailey: done.

roberthbailey:
Comments addressed; new version pushed.

roberthbailey:
There was a compile error in the version where the build just succeeded, which makes it pretty clear that our CI doesn't compile the example code.

examples/cpp-simple/README.md
```diff
@@ -8,9 +8,50 @@ It will
 - Every 10 seconds, write a log saying "Hi! I'm a Game Server"
 - After 60 seconds, call `SDK::Shutdown()` to shut the server down.
 
+## Running
+
+If you want to run the example code without building the container yourself, you can start a single gameserver by running
+```
+kubectl create -f https://raw.githubusercontent.com/googleforgames/agones/master/examples/cpp-simple/gameserver.yaml
```
markmandel: Should this point to a release branch?

examples/cpp-simple/README.md
```diff
@@ -8,9 +8,50 @@ It will
...
+The gameserver should exit automatically after about a minute. If it doesn't you can remove it
+by running:
+```bash
+$ GAMESERVER_NAME=$(kubectl get gs -o go-template --template '{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')
+$ kubectl delete gs ${GAMESERVER_NAME}
+```
+
+You can also start a fleet of gameservers by running
+```
+kubectl create -f https://raw.githubusercontent.com/googleforgames/agones/master/examples/cpp-simple/fleet.yaml
```
markmandel: Same thing, should this point to a release branch?

examples/cpp-simple/README.md
```diff
@@ -8,9 +8,50 @@ It will
...
+```
+
+The fleet will continuously replace gameservers that exit, so when you are finished you should remove the fleet by running
+```bash
+$ FLEET_NAME=$(kubectl get fleets -o go-template --template '{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')
```
markmandel: Curious - do we not know the fleet name here?

examples/cpp-simple/README.md
roberthbailey: Is that possible in the markdown files? I thought it only worked in the website directory?

examples/cpp-simple/README.md
roberthbailey: These configs use generate-name and `kubectl create` so we do not know the name. I could change them to have a fixed name, but instead I left them as-is which has been nice for testing b/c it's easy to keep creating new ones (and with image pulling set to always you can see the differences as you change the code).

markmandel: Aah cool. Thanks - was on my phone so hard for me to see.

examples/cpp-simple/README.md
markmandel: oooh. good point.

Maybe something  like:

```suggestion
kubectl create -f https://raw.githubusercontent.com/googleforgames/agones/<release-branch>/examples/cpp-simple/gameserver.yaml
```

With a note like: "Where `<release-branch>` is the Agones release you are working on, such as `release-0.11`"

WDYT?

???

roberthbailey: The nice thing about pointing to the master branch is that you can copy/paste and it should just work (obviously it wouldn't right now if you installed 0.11, but if we keep compatibility going forward it should work). I agree that it would be better to keep it tied to the release branch, which maybe argues for moving this over to the website....

For things in markdown in the git repository, I'm ok with them pointing at the master branch and not trying to keep them synchronized with releases. But I'll defer to you if you think it'd be better to instruct users towards a release branch.

markmandel: Yeah, I totally get the copy/paste impetus 👍

I've just had this conversation a lot:

Them: I tried the example thing, and it's broken
Me: Which version did you install?
Them: I grabbed the code from master
Me: I just updated the example with new features - it won't work against the current release.
Them: How am I supposed to know that?
Me: There is really no way, sorry. Use the release branch

So just trying to find away from that, since master will inevitably break at some point.

Maybe we should point them at release source zip downloads? Is that easier?

roberthbailey: I think you've convinced me that we should just move this over to the website so that it can be automatically linked to the right source. 

Both https://github.com/googleforgames/agones/tree/master/examples/allocator-service and https://github.com/googleforgames/agones/tree/master/examples/autoscaler-webhook do a good job of having a README but also having the main documentation that explains how to interact on the website where it's versioned.

markmandel: I think you've convinced me to move this over to the website 👍 

I like this a lot. I'd be happy with the website documentation to be a separate PR(s), unless you feel strongly you want to make the change in this PR?

roberthbailey: I've started working on it for this PR, so I'll finish that and send it your way today.

examples/cpp-simple/server.cc
roberthbailey: After doing some more testing on my PR I noticed that after switching from using `std::endl` to just inserting `"\n"` that my output (to stdout) is now being buffered and never gets flushed while the container is running. 

Looking at https://stackoverflow.com/questions/6214629/stdout-and-need-to-flush-it-c it seems like i should either

1. Use `std::cerr` instead of `std::cout`
1. Switch back to passing `std::endl`
1. Insert a flush call after each line

examples/cpp-simple/README.md
roberthbailey: I've pushed a new version that adds a tutorial to the website which should be published at about the same time as the 0.12 release.

examples/cpp-simple/server.cc
roberthbailey: Nothing is written to stdout _at all_ until the container exits. If I use `std::cerr` or leave in the `std::endl` instead of "\n" then each line is flushed as it is written.

roberthbailey:
I've fixed the output buffering issues in the latest version. This should now be ready for another review pass.

roberthbailey:
I've built `gcr.io/agones-images/cpp-simple-server:0.9` from the new master branch and pushed it.

PR #937, opened 2019-07-22
Updates to the rust simple example.

### LINKED ISSUE #703: Proposal: Split up the api group stable.agones.dev

roberthbailey:
Fixes #703.

I re-ordered things into what I thought was a more useful ordering (less work -- cloud -- to more work -- local dev environment) and made it so that we could push an "official" image to the `agones-images` project so that you can run the example without needing to compile any code. 

One thing that is weird is that while the running locally with docker instructions work for me, when I run the sample in an actual GKE cluster the game server never moves to Ready -- even though the sidecar is getting health pings the game server never transitions out of the unhealthy state. Maybe @aLekSer has an idea of why this might be?

### DIFF
--- a/examples/rust-simple/README.md
+++ b/examples/rust-simple/README.md
@@ -8,62 +8,36 @@ It will
 - Every 10 seconds, write a log saying "Hi! I'm a Game Server"
 - After 60 seconds, call `SDK::Shutdown()` to shut the server down.
 
-## Running locally (without Docker)
+To learn how to deploy this example service to GKE, please see the tutorial [Build and Run a Simple Gameserver (Rust)](https://agones.dev/site/docs/tutorials/simple-gameserver-rust/).
 
-This will build example server which will run for 100 seconds:
-```
-make build
-```
+## Building
 
-In a separate terminal run next command:
-```
-cd ../../build; make run-sdk-conformance-local TIMEOUT=120 TESTS=ready,watch,health,gameserver
-```
-This will start an SDK-server in a docker, which will be running for 120 seconds.
+If you have a local rust developer environment installed locally, you can run `make build` to compile the code and
+`make run` to execute the resulting binary.
 
-Run the Rust Simple Gameserver binary:
-```
-make run
-```
+If you want to build an updated container image or want to build the source code without installing the rust developer
+tools locally, run `make build-image` to run the `docker build` command with the correct context.
 
-You will see the following output:
-```
-Rust Game Server has started!
-Creating SDK instance
-Setting a label
-Starting to watch GameServer updates...
-Health ping sent
-Setting an annotation
-...
+This example uses the [Docker builder pattern](https://docs.docker.com/develop/develop-images/multistage-build/) to
+build the SDK, example and host it inside a container.
+
+## Testing locally with Docker
+
+If you want to run the example locally, you need to start an instance of the SDK-server. To run an SDK-server for
+120 seconds, run
+```bash
+$ cd ../../build; make run-sdk-conformance-local TIMEOUT=120 TESTS=ready,watch,health,gameserver
 ```
 
-Clean the resulting `sdk` directory and `target` folder:
-```
-make clean
-```
-
-
-## Running locally with Docker
-
-Build the container locally:
-```
-make build-image
-```
-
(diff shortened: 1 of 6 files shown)


aLekSer:
Re GKE cluster I cannot find any images under
https://console.cloud.google.com/gcr/images/agones-images/GLOBAL/rust-simple-server?gcrImageListsize=30

roberthbailey:
@aLekSer - there aren't any images there (yet). I've been testing with images in my own project so that I only push images to the official project once the code has been reviewed. 

From https://github.com/googleforgames/agones/blob/c1dd21fb90fc7b9c8ee9515789c1e2dd51e00830/examples/rust-simple/README.md:

If you want to build your own image

```bash
$ REPOSITORY=<your-repository> # e.g. gcr.io/agones-images
$ make build-image REPOSITORY=${REPOSITORY}
$ docker push ${REPOSITORY}/rust-simple-server:0.3
```

To run your own image, make a copy of
[examples/rust-simple/gameserver.yaml](https://raw.githubusercontent.com/googleforgames/agones/master/examples/rust-simple/gameserver.yaml),
modify the image name to the image you pushed, and run `kubectl create -f gameserver.yaml`.

aLekSer:
@roberthbailey thanks for clarification, just want to let you note that the original gcr.io/agones-images/rust-simple-server:0.3 should be pushed at some point for others to follow `README.md`.

roberthbailey:
Definitely. I just want to make sure that it works first (right now it doesn't -- see https://github.com/googleforgames/agones/issues/938).

aLekSer:
Yes, I am new to Rust language that's why I can not provide a fix for #938 . Seems that it is needed to add some loop with sleep at the beginning on SDK connect stage.

examples/rust-simple/README.md
```diff
@@ -8,62 +8,70 @@ It will
...
-```
-cd ../../build; make run-sdk-conformance-local TIMEOUT=120 TESTS=ready,watch,health,gameserver
-```
-This will start an SDK-server in a docker, which will be running for 120 seconds.
 
-Run the Rust Simple Gameserver binary:
+To run the example server:
 ```
-make run
+kubectl create -f https://raw.githubusercontent.com/googleforgames/agones/master/examples/rust-simple/gameserver.yaml
```
markmandel: Same as the C++ one, should we make this 
```suggestion
kubectl create -f https://raw.githubusercontent.com/googleforgames/agones/<release-branch>/examples/rust-simple/gameserver.yaml
```

With a note saying "Where <release-branch> should match the Agones version you have deployed, e.g `release-0.12`"

Sound good?

roberthbailey:
> I assume we aren't going to merge this until we fix #938 ?

Yes. Unless it looks like it won't be fixed before the 0.12.0 release in which case I'll cut out the part about running in a cluster and replace it with a warning that the rust SDK is still experimental.

markmandel:
Just to confirm - this is building now, but just isn't blocking? If that;s the case, I'll approve so we at least have an example that works.

roberthbailey:
It builds and runs following the docs. But because the container doesn't block, when running the example on GKE the pod restarts and that means that the gameserver never becomes healthy.

markmandel:
Got it - thanks! I'll take a look, see if I can work out a quick solution as well. I've used rust before in a previous life, so hopefully it will come back to me 😄 

Locally. I've got some changes to the conformance tests that will also check for the blocking functionality going forward.

roberthbailey:
And running locally works just fine.

markmandel:
Once #953 is merged, this should be good to go.

roberthbailey:
I'm going to re-work this to create a new tutorial like the cpp/nodejs ones. Changes should be ready shortly.

site/content/en/docs/Tutorials/simple-gameserver-rust.md
```diff
@@ -0,0 +1,203 @@
+---
```
markmandel: I love these :blue_heart:

PR #947, opened 2019-07-25
Gameserver status subresource

markmandel:
This is a draft implementation. I'm not convinced this is a good idea
as it removes the essentially atomic/transactional nature of being able
to manipulate GameServer resources -- which because of Allocation, is
something that is ideal to have.

I expect that e2e tests will fail.


markmandel:
Closing this PR in favour of #959 - see #329 for details.

Issue #927, opened 2019-07-18
Missing SDK Functionality across languages/engines


markmandel:
We have a variety of SDK functionality, but it isn't implemented across all languages and engines.

This is a work-list of SDK functionality that still needs to be implemented, so we can track it in a single place.

## C++
- [x] Allocate()
- [x] Reserve(seconds)
- [x] Conformance tests

## Node.js
- [x] Reserve(seconds)

## Rust
- [ ] Reserve(seconds)

## Unity
- [x] GameServer()
- [x] Reserve(seconds)
- [x] WatchGameServer(callback)
- [ ] Conformance tests

## Unreal
- [x] Allocate()
- [x] GameServer()
- [x] Reserve(seconds)
- [x] SetAnnotation(key, value)
- [x] SetLabel(key, value)
- [ ] WatchGameServer(callback)
- [ ] Conformance tests
- [ ] Simple example showing how to use the SDK

roberthbailey:
Looking at https://github.com/googleforgames/agones/blob/master/build/includes/sdk.mk only node, go, and rust have conformance tests, so I've added a checkbox to each of the other SDKs to add them.

roberthbailey:
I have a PR for the missing C++ functions.

steven-supersolid:
I'll take a look at Node.js

markmandel:
Updated C++ checklist, due to #948

roberthbailey:
@whisper0077 - do you have time to add the remaining functions to the Unity SDK?

roberthbailey:
@YannickLange - do you have time to add the remaining functions to the Unreal SDK?

roberthbailey:
Updated node.js checklist, due to #955.

markmandel:
Came up in Slack - does Unity & Unreal need some kind of blocking `Connect()` function basically so they can also block to wait for the sidecar to come up?

We could do a polling operation on the `/healthz` endpoint, or even do what I did on the Rust SDK, and poll `GameServer()` until we get a valid response (since it's read only, and a cached value).

WDYT?

markmandel:
Should we add a conformance test for REST api?

aLekSer:
If conformance tests for REST API is needed I can add the support for existent 3: Go, Rust and NodeJS.

markmandel:
I don't think we need REST conformation for languages we don't support REST for. Still waiting on some kind of licence/legal review to see what is possible with Unity/Unreal.

markmandel:
[...]

PR #948, opened 2019-07-25
Add the missing functions to the C++ SDK (Allocated & Reserve)

roberthbailey:
Part of https://github.com/googleforgames/agones/issues/927.

Since the C++ SDK doesn't have any automated tests, I tested this manually by modifying the simple-cpp example to call Allocate / Reserve instead of Ready when starting up, and then verified that the gameserver made the expected state transitions.

### DIFF
--- a/sdks/cpp/include/agones/sdk.h
+++ b/sdks/cpp/include/agones/sdk.h
@@ -39,6 +39,13 @@ class SDK {
   // Marks the Game Server as ready to receive connections
   AGONES_EXPORT grpc::Status Ready();
 
+  // Self marks this gameserver as Allocated.
+  AGONES_EXPORT grpc::Status Allocate();
+
+  // Marks the Game Server as Reserved for a given number of seconds, at which
+  // point it will return the GameServer to a Ready state.
+  AGONES_EXPORT grpc::Status Reserve(std::chrono::seconds seconds);
+
   // Send Health ping. This is a synchronous request.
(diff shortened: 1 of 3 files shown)


roberthbailey:
@devjgm

sdks/cpp/include/agones/sdk.h
```diff
@@ -39,6 +39,13 @@ class SDK {
   // Marks the Game Server as ready to receive connections
   AGONES_EXPORT grpc::Status Ready();
 
+  // Self marks this gameserver as Allocated.
```
devjgm: What is "Self"?

roberthbailey: "self marks" is a phrase -- basically instead of having the gameserver be allocated through the allocation service, the gameserver does the allocation itself. The comment matches the [comment in the go sdk](https://github.com/googleforgames/agones/blob/master/sdks/go/sdk.go#L69) (as do all of the other comments in this file).

roberthbailey:
> Can we please also update the docs here? https://agones.dev/site/docs/guides/client-sdks/cpp/

Done.

markmandel:
Looks like the reserve link will need to be done as a <a href="..." data-proofer-ignore >...</a>

See: https://github.com/wjdp/htmltest#see_no_evil-ignoring-content

roberthbailey:
Fixed the link for reserve.

Issue #941, opened 2019-07-23
C++ SDK: Use const-reference in WatchGameServer


roberthbailey:
From https://github.com/googleforgames/agones/pull/934:

> Is it intentional to pass a GameServer object by value? I don't know if this is an expensive type to copy, but it looks like accepting this object by const-reference may be worthwhile.

We should consider changing the function signature of

```c++
grpc::Status SDK::WatchGameServer(
    const std::function<void(agones::dev::sdk::GameServer)>& callback) { ... }
```

Since this would be a breaking change to the CPP SDK, I'm marking this for milestone 0.12.0 and we should resolve to do this (or leave it) ASAP. 

@devjgm

roberthbailey:
@Kuqd / @markmandel - thoughts?

markmandel:
I've no Singh opinions. I basically wrote the C++ SDK as a sacrificial draft. Very happy to defer to opinions that have more C++ experience than me.

devjgm:
A couple questions:

1. Does the `GameServer` argument to the callback need to be mutated? Or could it be `const`?
2. Is `GameServer` a small and cheap type that is most efficiently passed by value?

Depending on the answers to the above, I'd suggest a signature like:

```cc
grpc::Status SDK::WatchGameServer(
    const std::function<void(const agones::dev::sdk::GameServer&)>& callback);
```

roberthbailey:
1. My understanding is that it should not need to be mutated as changes to it would not be reflected back into k8s (so mutating it would only be confusing). 

1. I'm not sure where the cut-off is for small/cheap, but `GameServer` is a generated proto type, so I'm guessing the answer would be no.

markmandel:
I can't think of any reason you would want to mutate the gameserver

roberthbailey:
If there aren't any dissenting opinions by the end of the day tomorrow I'll send a PR to change the function signature.

PR #951, opened 2019-07-26
C++ SDK: Update the function signature of WatchGameServer to use a const-reference

roberthbailey:
Also update the simple cpp example so that it still compiles.

Fixes https://github.com/googleforgames/agones/issues/941.


markmandel:
@devjgm - I'm assuming this is correct? If so, I'm happy to commit and push through 👍

PR #952, opened 2019-07-26
Update the container image version for the simple unity sdk.

roberthbailey:
I went through the instructions at https://github.com/googleforgames/agones/blob/master/examples/unity-simple/README.md to verify that the new container image works with the updated SDK / apigroup rename.


roberthbailey:
I've pushed `gcr.io/agones-images/unity-simple-server:0.2`.

PR #953, opened 2019-07-27
Implement block on connect with Rust+Node.js SDK

### LINKED ISSUE #938: Rust SDK does not wait for connection to be ready
**What happened**: rust-simple doesn't work on a GKE cluster

**What you expected to happen**: gameserver to transition to Ready

**How to reproduce it (as minimally and precisely as possible)**: Run the example code from my pending PR: https://github.com/googleforgames/agones/pull/937

**Environment**:
- Agones version: master branch (pre-0.12.0)
- Install method (yaml/helm): helm (via `make install`)
- Others:

When launching the simple rust server, the gameserver starts up really fast, tries to connect to the sidecar and fails, and then restarts. This immediate restart causes the gameserver to become unhealthy, even though the second time around it successfully connects to the sidecar and is sending health() messages every 2 seconds. 

It seems like the underlying issue is that the rust SDK isn't blocking when initially trying to connect to the sidecar. 

In the C++ SDK, we have
```c++
  if (!pimpl_->channel_->WaitForConnected(
          gpr_time_add(gpr_now(GPR_CLOCK_REALTIME),
                       gpr_time_from_seconds(30, GPR_TIMESPAN)))) {
    return false;
  }
```

The Rust SDK has a comment that makes me think it's doing the same thing:

```
    /// Starts a new SDK instance, and connects to localhost on port 59357.
    /// Blocks until connection and handshake are made.
    /// Times out after 30 seconds.
```

but I don't see anything in the code that blocks, nor do I see anything in the Rust docs for [Channel](https://docs.rs/grpcio/0.1.1/grpcio/struct.Channel.html) that matches the [WaitForConnected](https://grpc.github.io/grpc/cpp/classgrpc_1_1_channel_interface.html#a31d360cd218614814d68e01e07ce55cc) function in the C++ gRPC docs.

markmandel:
## Rust
- Uses a simple polling mechanism to confirm the connection is live
- Add `ready` to the SDK documentation.

## Node.js
- Implements a sdk.connect() function

## Make
- Tweaks the conformance tests to add a random delay to ensure future
  testing of SDK blocking.

Fixes #938

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -122,14 +122,16 @@ run-sdk-conformance-local: ensure-agones-sdk-image
 	 -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" $(sidecar_tag)
 
 # Run SDK conformance test, previously built, for a specific SDK_FOLDER
+# Sleeps the start of the sidecar to test that the SDK blocks on connection correctly
 run-sdk-conformance-no-build: TIMEOUT ?= 30
+run-sdk-conformance-no-build: RANDOM := $(shell bash -c 'echo $$RANDOM')
+run-sdk-conformance-no-build: DELAY ?= $(shell bash -c "echo $$[ ($(RANDOM) % 5 ) + 1 ]s")
 run-sdk-conformance-no-build: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch
 run-sdk-conformance-no-build: ensure-agones-sdk-image
 run-sdk-conformance-no-build: ensure-build-sdk-image
-	sleep 2s && DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
-	docker run -p 59357:59357 -e "ADDRESS=" \
+	DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
+	sleep $(DELAY) && docker run -p 59357:59357 -e "ADDRESS=" \
 	 -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" --net=host  $(sidecar_tag)
-
 
 # Run SDK conformance test for a specific SDK_FOLDER
 run-sdk-conformance-test:
--- a/sdks/nodejs/src/agonesSDK.js
+++ b/sdks/nodejs/src/agonesSDK.js
@@ -24,6 +24,18 @@ class AgonesSDK {
 		this.emitters = [];
 	}
 
+	async connect() {
+		return new Promise((resolve, reject) => {
+			this.client.waitForReady(Date.now() + 30000, (error) => {
+				if (error) {
+					reject(error);
+				} else {
+					resolve();
+				}
+			})
+		});
+	}
+
 	async close() {
 		if (this.healthStream !== undefined) {
 			this.healthStream.destroy();
--- a/sdks/rust/src/sdk.rs
+++ b/sdks/rust/src/sdk.rs
@@ -12,15 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::{Arc, Mutex};
+use std::thread::sleep;
+use std::time::Duration;
+
 use futures::{Future, Sink, Stream};
 use grpcio;
-use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use grpcio::CallOption;
+use protobuf::Message;
 
 use errors::*;
 use grpc::sdk;
 use grpc::sdk_grpc;
-use protobuf::Message;
 use types::*;
 
 const PORT: i32 = 59357;
@@ -34,7 +37,7 @@ pub struct Sdk {
 impl Sdk {
     /// Starts a new SDK instance, and connects to localhost on port 59357.
     /// Blocks until connection and handshake are made.
-    /// Times out after 30 seconds.
+    /// Times out after ~30 seconds.
     pub fn new() -> Result<Sdk> {
         let addr = format!("localhost:{}", PORT);
         let env = Arc::new(grpcio::EnvBuilder::new().build());
@@ -43,7 +46,24 @@ impl Sdk {
             .connect(&addr);
         let cli = sdk_grpc::SdkClient::new(ch);
         let req = sdk::Empty::new();
-        let _ = cli.ready(&req).map(Box::new)?;
+
+        // Unfortunately there isn't a native way to block until connected
+        // so we had to roll our own.
+        let mut counter = 0;
+        loop {
+            counter += 1;
+            match cli.get_game_server(&req) {
+                Ok(_) => break,
+                Err(e) => {
+                    if counter > 30 {
+                        return Err(ErrorKind::Grpc(e).into());
+                    }
+                    sleep(Duration::from_secs(1));
+                    continue;
+                }
+            }
+        }
+
         let (sender, _) = cli.health()?;
         Ok(Sdk {
             client: Arc::new(cli),
@@ -58,14 +78,12 @@ impl Sdk {
         Ok(res)
     }
 
-
-    /// Allocate the Game Server 
+    /// Allocate the Game Server
     pub fn allocate(&self) -> Result<()> {
         let req = sdk::Empty::default_instance();
         let res = self.client.allocate(req).map(|_| ())?;
         Ok(res)
     }
-
 
     /// Marks the Game Server as ready to shutdown
     pub fn shutdown(&self) -> Result<()> {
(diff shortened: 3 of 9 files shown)


build/includes/sdk.mk
```diff
@@ -122,15 +122,17 @@ run-sdk-conformance-local: ensure-agones-sdk-image
...
 run-sdk-conformance-no-build: TIMEOUT ?= 30
+run-sdk-conformance-no-build: RANDOM := $(shell bash -c 'echo $$RANDOM')
+run-sdk-conformance-no-build: DELAY ?= $(shell bash -c "echo $$[ ($(RANDOM) % 5 ) + 1 ]s")
 run-sdk-conformance-no-build: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch
 run-sdk-conformance-no-build: ensure-agones-sdk-image
 run-sdk-conformance-no-build: ensure-build-sdk-image
-	sleep 2s && DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
-	docker run -p 59357:59357 -e "ADDRESS=" \
+	DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
+	sleep $(DELAY) && docker run -p 59357:59357 -e "ADDRESS=" \
```
markmandel: @aLekSer wdyt of this approach?

sdks/rust/src/sdk.rs
```diff
@@ -43,7 +46,24 @@ impl Sdk {
             .connect(&addr);
         let cli = sdk_grpc::SdkClient::new(ch);
         let req = sdk::Empty::new();
-        let _ = cli.ready(&req).map(Box::new)?;
+
```
markmandel: @thara would love your review of my Rust code. I'm a little.... rusty 🙄 🤣 🦀

site/content/en/docs/Guides/Client SDKs/rust.md
```diff
@@ -46,6 +46,15 @@ if sdk.health().is_ok() {
 }
 ```
 
+{{% feature publishVersion="0.12.0" %}}
+
+To mark the [game session as ready]({{< relref "_index.md#ready" >}}) call `sdk.ready()`.
+
+```rust
+if sdk.ready()?
```
roberthbailey: I don't know much about rust, but this looks incomplete.

markmandel: Ah good catch. Will fix.

markmandel:
Hmnn. Not sure why we are failing. Will look into it. I _think_ it's failing on the nodejs version - not sure I'm following the logs well.

markmandel:
Yep, can confirm - it's the node client - I assumed it blocked on connect, but it doesn't look like it. I'm hunting it down by @steven-supersolid if you have the code needed to block on connect, that would be awesome.

markmandel:
https://grpc.github.io/grpc/node/grpc.Client.html#waitForReady__anchor looks like what I need, but now trying to work out how to apply it in an async language :D

markmandel:
Here's my attempt:
https://github.com/googleforgames/agones/pull/953/files#diff-a7b304ad3eaf389b05bfeb118f316989R27

But I keep getting:
```
attempting to connect
Error: Failed to connect before the deadline
    at checkState (/go/src/agones.dev/agones/sdks/nodejs/node_modules/grpc/src/client.js:833:16)
```

But it doesn't look like it actually waiting for the deadline. Not quite sure why it isn't waiting.

markmandel:
Worked it out - but I should write some docs on it 👍  (hopefully it passes). Would still love a review of my node.js though. I'll update my PR description too.

sdks/nodejs/src/agonesSDK.js
```diff
@@ -23,7 +23,20 @@ class AgonesSDK {
 		this.healthStream = undefined;
 		this.emitters = [];
 	}
-	async close(){
+
+	async connect() {
```
steven-supersolid: We should add a test for this to maintain code coverage. Please let me know if you'd like me to add, e.g. separate PR?

sdks/nodejs/src/agonesSDK.js
```diff
@@ -23,7 +23,20 @@ class AgonesSDK {
...
+		return new Promise((resolve, reject) => {
+			this.client.waitForReady((new Date).getTime() + 30000, (error) => {
```
steven-supersolid: I would be tempted to add the timeout to the constructor with a default value. E.g.
```
constructor(timeout = 30000) {
    this.timeout = timeout
...
use later as Date.now() + this.timeout
```
Or if you like pass in as seconds and multiply by 1000 in the constructor

sdks/nodejs/src/agonesSDK.js
steven-supersolid: We should also update the docs.
I will look at generating API docs automatically for Node.js soon (maybe next weekend)

markmandel: Not sure how we are going to make a unit test for this, when there is nothing to connect to, but I will have a look, see if I can work out a way.

sdks/nodejs/src/agonesSDK.js
markmandel: All other SDKs timeout by 30 seconds - we control the SDK server, so not personally seeing a good reason to make this configurable. I can't think of a use case in which someone would want to make this larger/smaller?

sdks/nodejs/src/agonesSDK.js
markmandel: Yep - on my todo list.

markmandel: Aha, got it - there is a mocking library you use to mock out the client. On it!

site/content/en/docs/Guides/Client SDKs/rust.md
markmandel: Done!

sdks/nodejs/src/agonesSDK.js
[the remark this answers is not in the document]
markmandel: Done!

sdks/nodejs/src/agonesSDK.js
markmandel: Done! PTAL.

markmandel: Done!. PTAL.

build/includes/sdk.mk
aLekSer: yes, this one much better

site/content/en/docs/Guides/Client SDKs/nodejs.md
```diff
@@ -28,14 +28,24 @@ Add the agones dependency to `package.json`, replacing with the download locatio
...
 
 ```javascript
 const AgonesSDK = require('agones');
 
 let agonesSDK = new AgonesSDK();
 ```
 
+{{% feature publishVersion="0.12.0" %}}
+To connect to the SDK server, either local or when running on Agones, run the `sdk.connect()` method.
+This will block for up to 30 seconds if the SDK server has not yet started. If the connection cannot be made,
```
steven-supersolid: Minor but would call out that it's an `async` method that will `resolve` once connected or `reject` on error or if no connection can be made after 30 seconds. Technically it will block execution of the calling code if that uses `await` but does not block the Node.js event loop. I try to stray away from using the term block for async code and would reserve that for synchronous code that does block the event loop e.g. heavy calculation/loop

steven-supersolid:
I just realised we will also need to update the example in `examples/nodejs-simple/src/index.js`

site/content/en/docs/Guides/Client SDKs/nodejs.md
markmandel: Thanks! Would definitely like to use more canonical language for node.js - will make this change.

markmandel:
> I just realised we will also need to update the example in `examples/nodejs-simple/src/index.js`

Yep - I think we can do this in a separate PR though - I think this one has enough in it already.

site/content/en/docs/Guides/Client SDKs/nodejs.md
markmandel: Done!

sdks/nodejs/src/agonesSDK.js
```diff
@@ -24,8 +24,20 @@ class AgonesSDK {
...
+				} else {
+					resolve();
+				}
+			})
+		});
+	}
+
 	async close() {
-		if (this.healthStream !== undefined) {
+		if (this.healthStream !== undefined){
```
roberthbailey: nit: you seem to have lost a space on this line

markmandel: ARG! Fixing.

roberthbailey:
@steven-supersolid & @thara - I'm going to merge this so that it gets in before the 0.12.0 release candidate cut. If you have any other review comments, feel free to send a followup PR to fix things up.

PR #955, opened 2019-07-27
Add Reserve to Node.js SDK

steven-supersolid:
Add wrapper method
Add tests
Update docs
Add missing doc entry for `Allocate`

Part of #927


roberthbailey:
Can you also update https://github.com/googleforgames/agones/blob/master/build/includes/sdk.mk#L142 to add the conformance test for the new function?

steven-supersolid:
I've added but noticed the conformance test does not check anything, just executes methods. I'm not really familiar with this kind of test but perhaps better if the test is written more like the existing unit tests but using the running server instead of mocking, so an integration test.

If so I can create a separate PR which will use the current unit test system to test all methods of the SDK. It might also be useful to move this to the SDK folder so it can be easily run and updated during SDK development e.g. with `npm run conformance`.

roberthbailey:
> I've added but noticed the conformance test does not check anything, just executes methods.

Interesting. I hadn't actually looked at them in detail, I just knew that the Go test was exercising different methods than the others. I guess maybe they are just validating that the methods have been implemented (e.g. that the SDK "conforms" to the proper set of functions) and we rely on individual SDK tests to make sure that the methods actually work properly?

@markmandel @aLekSer

markmandel:
Yeah - the conformance test is to make sure the gRPC / REST calls get made -- and we get the results from the client that is expected.

We have e2e and unit results to test the actual results - but if we know the SDK clients conform to the API surface of the grpc / reset server, then we are fairly sure that the SDKs actually work.

Does that make sense?

aLekSer:
Actually we also check that we that SDK client able to read UID and CreationTimestamp and set them properly as Label and Annotation. So we have a simple feedback mechanism here.

markmandel:
Weird. These are go tests that are failing. Lemme retry the test.

markmandel:
Oh man, looks like some GitHub urls got moved under you. I'll look into getting a PR in place, otherwise this will block everything

markmandel:
This should unblock once #961 is pushed through.

PR #960, opened 2019-07-29
Fix the capitalization for 'publishVersion' in the website docs.

roberthbailey:


roberthbailey:
I noticed while browsing some of the website docs that a bunch didn't look right, and realized that it was because I'd mis-capitalized `publishVersion` in a bunch of places.

PR #962, opened 2019-07-30
Fix Installation Docs and example GS configuration

aLekSer:
Now we have 6 pods in agones-system namespace, not one as per docs.
There was an issue in PortPolicy and image tag.

### DIFF
--- a/examples/gameserver.yaml
+++ b/examples/gameserver.yaml
@@ -41,7 +41,7 @@ spec:
     # - "Passthrough" dynamically sets the `containerPort` to the same value as the dynamically selected hostPort.
     #      This will mean that users will need to lookup what port has been opened through the server side SDK.
     # port is available. When static is the policy specified, `hostPort` is required to be populated
-    portPolicy: Dynamic
+    portPolicy: Static
     # the port that is being opened on the game server process
     containerPort: 7654
     # the port exposed on the host, only required when `portPolicy` is "Static". Overwritten when portPolicy is "Dynamic".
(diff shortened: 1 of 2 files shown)


roberthbailey:
According to https://agones.dev/site/docs/examples/ these aren't supposed to be working examples:

"These examples are for reference only. They are not backed by working images."

I'd be happy to make them workable, but if we do we should update the docs too (and make them all work).

aLekSer:
@roberthbailey 
Regarding the configuration, someone as I did could think they are working.
The most problem is that in comments section we read that Dynamic and HostPort could not defined at the same time. And then we define them both - which is mutually exclusive lines and bad configuration.
Note that I added fix of documentation here. PTAL again.

aLekSer:
By the way I can leave not working image and correct only PortPolicy.

roberthbailey:
> Regarding the configuration, someone as I did could think they are working.

I agree. I found it confusing as well.

markmandel:
I've no strong opinions on leaving them as broken images - if you both feel strongly they should be working images, then lets make them all working configurations :+1: not just gameservers, but fleet.yaml, etc.

Sound good?

roberthbailey:
I think it'd be nice to have them all working. It should be relatively easy to do so, and to update the docs to remove the note about them being non-functional.

markmandel:
Sounds good to me. Do we want to include it in this PR, or make a new one?

roberthbailey:
There isn't a good reason not to split it among multiple PRs so we can get this one in. I've created https://github.com/googleforgames/agones/issues/969 with a checklist for the files in question (I think some of them are already ok, so we can just verify those and fix the others).

roberthbailey:
Part of https://github.com/googleforgames/agones/issues/969.

PR #968, opened 2019-07-31
Flaky: TestGameServerUnhealthyAfterDeletingPod

markmandel:
1 minute may not be enough time. Let's find out.

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -207,7 +207,7 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
 	err = podClient.Delete(pod.ObjectMeta.Name, nil)
 	assert.NoError(t, err)
 
-	_, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateUnhealthy, time.Minute)
+	_, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateUnhealthy, 3*time.Minute)
 	assert.NoError(t, err)
 }
 


roberthbailey:
Is it easy to run this test say 100 times before and after this change to a) see if you can reproduce the error and b) verify that this fixes it?

Also, I'm always a bit leery of changes to tests that just twiddle timeouts as they can mask actual issues that should be investigated.

markmandel:
I can do that locally for sure!

In this case,  I figured that the gameserver being removed is an event that either happens, or it doesn't. 

Given that the default terminationGracePeriodSeconds is 30 seconds, and controller resync occurs every 30 seconds - I'm wondering if there is a chance that sometimes the pod takes ~ 30 seconds to actually be deleted, and then if the event gets missed by the controller for some reason, then it'll take another 30 seconds to resync the event and ping that it is missing - which hits the 60 second timeout on this test -- but otherwise it would still actually pass. So I figured a bump in timeout may be appropriate here.

If my theory is off, the test will remain being flaky, in which case I'm totally wrong, and something else is afoot.

roberthbailey:
Your theory makes sense. It would be awesome to have some supporting evidence (e.g. 2/100 tests failed before, now 0/100 fail).

markmandel:
Love it! :+1: 

https://varunksaini.com/blog/run-go-test-multiple-times/

Makes it very easy to do. I will do this! (just not right now, I have to run off)

roberthbailey:
Since this only changes test code, I'm ok merging it between the 0.12.0 RC and 0.12.0 release.

markmandel:
I have run it around 50 time so far, without an error - but hit the default go test timeout of 10 minutes :confused:  will rerun it with more soon.

markmandel:
Ran 100 times, with the 3m timeout, no errors.

For reference (i.e. in case I need to find it again), here is the command I used:
go test -mod=vendor -v -timeout=1h -count=100 agones.dev/agones/test/e2e -run ^TestGameServerUnhealthyAfterDeletingPod$

Running it 100 times with the 1m timeout, also passed... so now I have no idea.

I figured this was a good first educated guess at a solution.

Issue #958, opened 2019-07-29
Local SDK Server should update internal GameServer on Ready(), Allocate() etc.


markmandel:
**Is your feature request related to a problem? Please describe.**

When using the Local SDK Server, `Ready()`, `Allocate()`, `Shutdown()`, `Reserve(seconds)`, do not change the state of the local `GameServer`, and do not fire watch actions.

It would be really nice if they did - especially since `SetLabel(...)` and `SetAnnotation(...)` do do this - so there is precedent.

**Describe the solution you'd like**

`Ready()`, `Allocate()`, `Shutdown()`, `Reserve(seconds)` should update the underlying `GameServer` record as appropriate for their logic, and fire `WatchGameServer` events as well that can be intercepted by the game server SDK.

**Describe alternatives you've considered**
Leaving it as is - but that does leave an inconsistent user experience - and one that I believe is less smooth. 

Users can edit a YAML file directly to make the same thing happen, but its far less intuitive and smooth an experience. 

**Additional context**

[Code](https://github.com/googleforgames/agones/blob/master/pkg/sdkserver/localsdk.go)

aLekSer:
You can assign this to me, that's really frustrating fo developers who want to test their code versus local sidecar.

markmandel:
Assigned!

markmandel:
@aLekSer can we close this ticket now?

aLekSer:
Yes, I think we can close this as long as we don't need to implement Unhealthy state logic for local GameServer.

markmandel:
Oh that's an interesting question. I would say - since that isn't covered by this ticket, we can close this ticket. If we have users that want this functionality as part of the local SDK, we can include it there as well. 

How does that sound?

That also being said - should we add documentation to: https://agones.dev/site/docs/guides/client-sdks/local/ ? I'm thinking probably. WDYT?

aLekSer:
> If we have users that want this functionality as part of the local SDK, we can include it there as well.

For health check it could be helpful and we can turn it off (so could support both options).
> That also being said - should we add documentation to: https://agones.dev/site/docs/guides/client-sdks/local/ ?

Definitely yes, I will explain current behavior of Local SDK in upcoming PR.

markmandel:
All of this sounds good! Okay, let's not close this ticket then :+1:

aLekSer:
Returning back to this ticket, will add documentation to https://agones.dev/site/docs/guides/client-sdks/local/ .

[...]

PR #979, opened 2019-08-05
Local SDK Server: Add proper GS state handling

aLekSer:
Update the state of local SDK GS according to calls which were made
by SDK client. Local sidecar now keeps track of the current state.
Part of #958 fix.
Added UTs for `localsdk.go`.

Need to add transition to Unhealthy state in a separate PR.

### DIFF
--- a/pkg/sdkserver/localsdk.go
+++ b/pkg/sdkserver/localsdk.go
@@ -58,13 +58,16 @@ var (
 // is being run for local development, and doesn't connect to the
 // Kubernetes cluster
 type LocalSDKServer struct {
-	gsMutex          sync.RWMutex
-	gs               *sdk.GameServer
-	update           chan struct{}
-	updateObservers  sync.Map
-	requestSequence  []string
-	expectedSequence []string
-	testMode         bool
+	gsMutex           sync.RWMutex
+	gs                *sdk.GameServer
+	update            chan struct{}
+	updateObservers   sync.Map
+	requestSequence   []string
+	expectedSequence  []string
+	gsState           agonesv1.GameServerState
+	gsReserveDuration *time.Duration
+	reserveTimer      *time.Timer
+	testMode          bool
 }
 
 // NewLocalSDKServer returns the default LocalSDKServer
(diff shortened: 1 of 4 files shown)


roberthbailey:
Does this fix #958?

aLekSer:
@roberthbailey Yes, I think it is.

pkg/sdkserver/localsdk.go
```diff
@@ -58,24 +58,64 @@ var (
...
+	reserveTimer          *time.Timer
+	testMode              bool
+	validStateTransitions map[string]bool
+}
+
+func insertStateTransition(m map[string]bool, from, to agonesv1.GameServerState) {
+	m[string(from)+string(to)] = true
+}
+
+func initValidStateTransitions() map[string]bool {
```
markmandel: I'm wondering if we really care about this?

What are we trying to prevent here? I'm just wondering if this is going put some limitations on what's possible with the local test server, and add extra maintenance down the road for us.

What's the worst that happens if we don't validate state transitions locally?

aLekSer: Main intention was to prevent developers from using wrong assumptions while debugging their game servers locally.
Additionally this would be a place in go code where we state all possible edges of state diagram.
I imagine that we can add a flag to validate this transitions or not so that it would be optional.

markmandel: I'm just trying to think of what could be a bad transition?

Moving from Shutdown -> Ready/Allocated/Reserved ?

Even then - I'm not sure if we should stop people. If the developer wants to go in that loop, they would have to restart the gameserver and sdk every time, and that could slow them down. Fast development iteration loops are a very good thing, so should we be putting extra friction in people's way?

If people really want to move from Shutdown -> Ready they can change the backing file anyway, and it (should) overwrite what's in the local sdk. So I'm not sure this is necessary?

aLekSer: Yes, that's right. This is redundant for local development and those steps are well documented in GS State transitions schema.

PR #983, opened 2019-08-06
Release 0.12.0

markmandel:


site/content/en/blog/releases/0.12.0-rc.md
```diff
@@ -14,9 +14,7 @@ We've also worked to complete functionality across GameServer SDKs, but we still
 
 We are on track for a 1.0 release in September, with only a [few features left to implement](https://github.com/googleforgames/agones/milestone/14).
 
-Check the <a href="https://github.com/googleforgames/agones/tree/release-0.12.0-rc" data-proofer-ignore>README</a> for details on features, installation and usage.
-
-**Implemented enhancements:**
+Check the <a href="https://github.com/googleforgames/agones/tree/release-0.12.0-rc" >README</a> for details on features, installation and usage.
```
roberthbailey: seems like this shouldn't have the rc extension in it?

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+## [v0.12.0](https://github.com/googleforgames/agones/tree/v0.12.0) (2019-08-06)
```
roberthbailey: On the blog page it lists all of the RC changes but here it look like the release doesn't have much different (because they are all captured in the release candidate). Is that how it's been done in the past?

site/content/en/blog/releases/0.12.0-rc.md
markmandel: That's the release candidate post - so, yes it should? (the change here is removing the `data-proofer-ignore` tag from the link)

CHANGELOG.md
markmandel: Yeah, I tend to copy new features, etc through from RC->Stable for release notes, I figured users will be looking at either and/or stable to see what the new features where.

The changelog is auto generated, and its all in one list, so I figured it was easy to peruse down, so its left as is.

Happy to take other suggestions on how to do this though.

site/content/en/docs/Getting Started/create-fleet.md
```diff
@@ -76,56 +76,6 @@ Let's wait for the two `GameServers` to become ready.
 watch kubectl describe fleet simple-udp
 ```
 
-{{% feature expiryVersion="0.12.0" %}}
-```
-Name:         simple-udp
-Namespace:    default
-Labels:       <none>
-Annotations:  kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"stable.agones.dev/v1alpha1","kind":"Fleet","metadata":{"annotations":{},"name":"simple-udp","namespace":"default"},"spec":{"replicas":2,...
-API Version:  stable.agones.dev/v1alpha1
```
pooneh-m: Why removing the examples? Shouldn't them be replaced with v1 version of the same example?

markmandel: yep - that's the next line down after this. This is removing and cleaning up the pre-v1 example

roberthbailey: The v1 versions are live. See https://agones.dev/site/docs/getting-started/create-fleet/

PR #995, opened 2019-08-13
Flaky: Allocator bad tls certificate

markmandel:
This should fix the bad tls e2e test failures, as it will force a reload of the allocator deployment - which will load the most up to date generated ssl certificate.

This is the same implementation that is in place for the controller as well for its webhook ssl certificates.

### DIFF
--- a/install/helm/agones/templates/service/allocation.yaml
+++ b/install/helm/agones/templates/service/allocation.yaml
@@ -57,6 +57,10 @@ spec:
       heritage: {{ .Release.Service }}
   template:
     metadata:
+      annotations:
+{{- if .Values.agones.allocator.generateTLS }}
+        revision/tls-cert: {{ .Release.Revision | quote }}
+{{- end }}
       labels:
         multicluster.agones.dev/role: allocator
         app: {{ template "agones.name" . }}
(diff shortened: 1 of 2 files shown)


install/helm/agones/templates/service/allocation.yaml
```diff
@@ -57,6 +57,10 @@ spec:
       heritage: {{ .Release.Service }}
   template:
     metadata:
+      annotations:
+{{- if .Values.agones.allocator.generateTLS }}
+        revision/tls-cert: {{ .Release.Revision | quote }}
+{{- end }}
```
pooneh-m: Actually can you explain how this annotation is causing restart?
Should annotations be inside the if statement?

markmandel: So this puts an annotation on each pod of the deployment - with the Helm revision number.

When you update helm with a new install, the revision number increases. Since there is now a new annotation on each underlying pod in the Deployment, kubernetes will do a rolling update to the new deployment, thus restarting everything.

`annotations` could also be in the if statement. Should work either way (it'll be an empty annotation block otherwise). I don't mind to be honest - if you feel like it should include the `annotations` block, I'll change it :+1:

pooneh-m: Thanks for the explanation.

markmandel:
@googlebot rescan

Are you alive my bot friend?

Issue #15, opened 2017-12-08
Game Server health checking


markmandel:
# Design

## Health Configuration
```yaml
apiVersion: "stable.agon.io/v1alpha1"
kind: GameServer
metadata:
  name: "simple-udp"
spec:
  portPolicy: "static"
  containerPort: 7654
  hostPort: 7777
  # new health section
  health:
    # defaults to false, but can be set to true
    disabled: false
    # If the `Health()` function doesn't get called at least once every timeout seconds, then
    # the game server is not healthy. Defaults to "5"
    periodSeconds: 3
    # Minimum consecutive failures for the probe to be considered failed after having succeeded. Defaults to 3. Minimum value is 1.
    failureThreshold: 3
    # Number of seconds after the container has started before health check is initiated. Defaults to 5 seconds
    initialDelaySeconds: 5
  template:
    spec:
      containers:
      - name: simple-udp
        image: gcr.io/agon-images/udp-server:0.1
```

## SDK API

`SDK.Health()`

The `Health()` function on the SDK object ill need to be called regularly below the `timeout` threshold time to be considered healthy.

## Failure

1. If any of the backing Pod containers fails for any reason before the GameServer moves to `Ready` then, it should restart as per the `restartPolicy` (which defaults to "Always")
1. If the GameServer Pod fails (for any reason) after the `Ready` state, then it doesn't restart, but moves the GameServer to an `Unhealthy` state - and then it's up to the managing code to determine what to do at that point.
1. If the SDK sidecar fails, then it should always be restarted.

## Outside scope
- The GameServer Pod fails to be scheduled for any reason, such as lack of resources (cpu, memory) or a being allocated an unavailable port - this will be managed in a separate ticket
 
# Implementation
## gRPC
Health is a unidirectional stream from the gameserver client -> the sidecar. The sidecar will update the State to `UnHealthy` if it doesn't receive a healthcheck event within the allotted time.

## Kubernetes
(Theory - need investigation) 
1. The sidecar will proxy the liveness check for the GameServer container through a `gshealthz` url endpoint. It will track `Health()` messages and if they drop below the set threshhold, return a `500`. 
1. Once the GameServer is `Ready`, then this will always return `200` - which (in theory) should mean that Kubernetes will never restart the GameServer container.

PR #996, opened 2019-08-13
Speed up CI Build

markmandel:
- Run more things in parallel.
- Cache C++ build output, as that is the longer path.

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -137,7 +137,6 @@ run-sdk-conformance-no-build: ensure-build-sdk-image
 run-sdk-conformance-test:
 	$(MAKE) run-sdk-command COMMAND=build-sdk-test
 	$(MAKE) run-sdk-conformance-no-build
-	$(MAKE) run-sdk-command COMMAND=clean
 
 # Run a conformance test for all SDKs supported
 run-sdk-conformance-tests:
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -29,6 +29,13 @@ steps:
     - '--key=$_HTMLTEST_CACHE_KEY'
   waitFor: ['-']
 
+- name: gcr.io/$PROJECT_ID/restore_cache
+  id: cpp-sdk-build-restore-cache
+  args:
+    - '--bucket=gs://$_CACHE_BUCKET'
+    - '--key=$_CPP_SDK_BUILD_CACHE_KEY'
+  waitFor: ['-']
+
 #
 # Creates the initial make + docker build platform
 #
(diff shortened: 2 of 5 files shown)


cloudbuild.yaml
```diff
@@ -119,19 +119,30 @@ steps:
...
   dir: "build"
-  args: [ "-j", "4", "build"]
+  args: [ "-j", "4", "build-images"]
 - name: "make-docker"
   id: push-images
   waitFor:
-    - build
+    - build-images
   dir: "build"
   args: [ "-j", "3", "push"]
```
roberthbailey: We build 4 images now, so this can be `-j 4`.

cloudbuild.yaml
```diff
@@ -119,19 +119,30 @@ steps:
...
+# Build all the SDKs
+#
+
+- name: "make-docker"
+  id: build-sdks
+  waitFor:
+    - lint
+    - ensure-build-sdk-image-base
+  dir: "build"
+  args: [ "-j", "4", "--output-sync=target", "build-sdks"]
```
roberthbailey: It might be worth adding a note to https://github.com/googleforgames/agones/tree/master/build/build-sdk-images#adding-support-for-a-new-language to change the parallelism argument here when adding support for a new language.

markmandel: Admittedly, only C++ has a build command, so -j 4 is currently overkill anyway.

markmandel:
First pass seemed to scrape off 3 minutes. Not as much as I would like.

markmandel:
@googlebot rescan

markmandel:
That's better - a 14 minute build.

aLekSer:
Cool gain in execution time. However as I remember there is a change that steps `tests` and `run-sdk-conformance-tests` would overlap when latter would remove `./node_modules` and `make tests` (
`Step #15 - "tests": /usr/bin/make run-sdk-command COMMAND=test SDK_FOLDER=node` in particular)
https://github.com/googleforgames/agones/blob/c5fa6ef21470589641223973fbdcf9aa89c4ba42/build/includes/sdk.mk#L140

You could see from logs that Step 15 `tests` and Step 21 `SDK conformance` runs simultaneously:
https://console.cloud.google.com/cloud-build/builds/8662c7b1-7c90-464e-8cb2-c6a20a4f91fa?project=agones-images

markmandel:
Why does it need to be removed? Could we not gitignore it instead?

aLekSer:
we could, I need to recap pros and cons as well as reason behind adding clean step for Node JS.

markmandel:
Looking at the code - it's already .gitignored. So I can remove the clean step. Ran it locally, and git did not complain about anything not being added :+1:

markmandel:
One thing I don't understand. I upped the e2e test parrallel count to 64 (from 32), no increase in performance. Seems odd.

markmandel:
Just removed 4 minutes from the e2e test!

test/e2e/fleet_test.go
```diff
@@ -56,7 +56,10 @@ func TestFleetScaleUpEditAndScaleDown(t *testing.T) {
 	fixtures := []bool{true, false}
 
 	for _, usePatch := range fixtures {
+		usePatch := usePatch
```
markmandel: I figure someone will ask about this:
https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

build/includes/sdk.mk
```diff
@@ -137,7 +137,6 @@ run-sdk-conformance-no-build: ensure-build-sdk-image
 run-sdk-conformance-test:
 	$(MAKE) run-sdk-command COMMAND=build-sdk-test
 	$(MAKE) run-sdk-conformance-no-build
-	$(MAKE) run-sdk-command COMMAND=clean
```
aLekSer: Note that there is also a clean step in Rust. Will check if Rust package files are gitignored.
I suggest adding new make target for clean.

aLekSer: Rust target dir have appropriate .gitignore. https://github.com/googleforgames/agones/blob/master/test/sdk/rust/.gitignore
So the only request for adding new target for clean left.

Could be:
```
clean-sdk-conformance-tests:
	$(MAKE) run-all-sdk-command COMMAND=clean
```

aLekSer: Created a PR linked to this.

cloudbuild.yaml
markmandel: I should be clearer on this - there is currently only 1 sdk that is building, but I set it to 4 to (a) be future proof and (b) since we have multiple concurrent steps, I wanted to max things out at half the cores on the machine.

So I don't _think_ we need a comment on this, at least at this time.

Does that sound reasonable?

roberthbailey: yep.

PR #881, opened 2019-07-09
Promote allocation to v1

roberthbailey:
This is part of #703.

/assign @markmandel
/label kind/breaking

PR #1001, opened 2019-08-15
Revert the change to promote the service for multicluster allocation to v1

roberthbailey:
It is still in active development.

Partial revert of #881 where I was a bit overzealous promoting to stable. 

/assign @pooneh-m


roberthbailey:
I think these were the only two places that needed to be changed back, but I may have missed one. We'll see if the presubmits pass. :)

PR #899, opened 2019-07-11
Upgrading Terraform to 0.12.3

aLekSer:
Fixed configuration tf files according to a new format.
`terraform fmt` all files.
Also updated providers versions.

As part of #657

PR #1006, opened 2019-08-20
Fix the required version of terraform.

roberthbailey:
This was missed in #899 (and should be live now so it isn't version guarded).

### DIFF
--- a/site/content/en/docs/Installation/terraform.md
+++ b/site/content/en/docs/Installation/terraform.md
@@ -9,7 +9,7 @@ description: >
 
 ## Prerequisites
 
-- Terraform v0.11.13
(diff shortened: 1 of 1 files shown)


markmandel:
So just to be 100% clear - the older version of terraform no longer works?

roberthbailey:
> So just to be 100% clear - the older version of terraform no longer works?

Correct. I tried using the older version and terraform apply gives an error (I don't recall which one) because we've started using syntax that was introduced in 0.12. 

Also note that the terraform module linked from the guide pulls the agones configs from the **master branch**, so this has been broken since #889 was submitted. 

/cc @aLekSer

PR #1007, opened 2019-08-21
Add Sidecar log level parameter to GS specification

### LINKED ISSUE #971: Configurable log level for agones sidecar

Currently the agones sidecar logs all received Health Pings and GameServer events with the log level INFO

A way to control the log level to be sent to stdout, logrus already has such a feature: SetLevel. This would allow us to set the log level to WARN or ERROR and avoid seeing these logs, which currently just make it more harder to spot relevant errors.

**Additional context**

I did find the issue related to logs in Stackdriver (https://github.com/googleforgames/agones/issues/335), but it doesn't apply directly to our usecase, as we forward all stdout to our logging system.

aLekSer:
Add ability to specify `logrus` log level in yaml configuration and apply this log level on GameServer creation step.

Closes  #971 .

### DIFF
--- a/examples/gameserver.yaml
+++ b/examples/gameserver.yaml
@@ -35,7 +35,7 @@ spec:
   ports:
     # name is a descriptive name for the port
   - name: default
-    # portPolicy has two options:
+    # portPolicy has three options:
     # - "Dynamic" (default) the system allocates a free hostPort for the gameserver, for game clients to connect to
     # - "Static", user defines the hostPort that the game client will connect to. Then onus is on the user to ensure that the
     # - "Passthrough" dynamically sets the `containerPort` to the same value as the dynamically selected hostPort.
@@ -60,6 +60,13 @@ spec:
     # Minimum consecutive failures for the health probe to be considered failed after having succeeded.
     # Defaults to 3. Minimum value is 1
     failureThreshold: 3
+  # logging parameters for game server sidecar
+  logging:
+    # sdkServer logging parameter has three options:
+    #  - "Info" (default) The SDK server will output all messages except for debug messages
+    #  - "Debug" The SDK server will output all messages including debug messages
+    #  - "Error" The SDK server will only output error messages
+    sdkServer: Info
   # Pod template configuration
   # https://v1-12.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#podtemplate-v1-core
   template:
--- a/install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
@@ -92,6 +92,21 @@ properties:
               type: integer
               minimum: 1
               maximum: 65535
+      logging:
+        type: object
+        title: Define logging levels for agones system components
+        properties:
+          sdkServer:
+            type: string 
+            description: |
+              sdkServer logging parameter has three options:
+              - "Info" (default) The SDK server will output all messages except for debug messages
+              - "Debug" The SDK server will output all messages including debug messages
+              - "Error" The SDK server will only output error messages
+            enum:
+            - Error
+            - Info
+            - Debug
       scheduling:
         type: string
         enum:
--- a/install/yaml/install.yaml
+++ b/install/yaml/install.yaml
@@ -339,6 +339,21 @@ spec:
                             type: integer
                             minimum: 1
                             maximum: 65535
+                    logging:
+                      type: object
+                      title: Define logging levels for agones system components
+                      properties:
+                        sdkServer:
+                          type: string 
+                          description: |
+                            sdkServer logging parameter has three options:
+                            - "Info" (default) The SDK server will output all messages except for debug messages
+                            - "Debug" The SDK server will output all messages including debug messages
+                            - "Error" The SDK server will only output error messages
+                          enum:
+                          - Error
+                          - Info
+                          - Debug
                     scheduling:
                       type: string
                       enum:
@@ -594,6 +609,21 @@ spec:
                     type: integer
                     minimum: 1
                     maximum: 65535
+            logging:
+              type: object
+              title: Define logging levels for agones system components
+              properties:
+                sdkServer:
+                  type: string 
+                  description: |
+                    sdkServer logging parameter has three options:
+                    - "Info" (default) The SDK server will output all messages except for debug messages
+                    - "Debug" The SDK server will output all messages including debug messages
+                    - "Error" The SDK server will only output error messages
+                  enum:
+                  - Error
+                  - Info
+                  - Debug
             scheduling:
               type: string
               enum:
@@ -860,6 +890,21 @@ spec:
                             type: integer
                             minimum: 1
                             maximum: 65535
+                    logging:
+                      type: object
+                      title: Define logging levels for agones system components
+                      properties:
+                        sdkServer:
+                          type: string 
+                          description: |
+                            sdkServer logging parameter has three options:
+                            - "Info" (default) The SDK server will output all messages except for debug messages
+                            - "Debug" The SDK server will output all messages including debug messages
+                            - "Error" The SDK server will only output error messages
+                          enum:
+                          - Error
+                          - Info
+                          - Debug
                     scheduling:
                       type: string
                       enum:
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -133,10 +133,18 @@ type GameServerSpec struct {
 	Ports []GameServerPort `json:"ports"`
 	// Health configures health checking
 	Health Health `json:"health,omitempty"`
-	// Scheduling strategy. Defaults to "Packed".
+	// Scheduling strategy. Defaults to "Packed"
 	Scheduling apis.SchedulingStrategy `json:"scheduling,omitempty"`
+	// Logging specifies log levels for Agones system containers
+	Logging Logging `json:"logging,omitempty"`
 	// Template describes the Pod that will be created for the GameServer
 	Template corev1.PodTemplateSpec `json:"template"`
+}
+
+// Logging specifies log levels for Agones system containers
+type Logging struct {
+	// SdkServer is a log level for SDK server (sidecar) logs. Defaults to "Info"
+	SdkServer string `json:"sdkServer,omitempty"`
 }
 
 // GameServerState is the state for the GameServer
@@ -213,9 +221,17 @@ func (gss *GameServerSpec) ApplyDefaults() {
 	gss.applyPortDefaults()
 	gss.applyHealthDefaults()
 	gss.applySchedulingDefaults()
+	gss.applyLogLevelDefaults()
 }
 
-// applyContainerDefaults applues the container defaults
+// applyLogLevelDefaults applies the log level default - "Info"
+func (gss *GameServerSpec) applyLogLevelDefaults() {
+	if gss.Logging.SdkServer == "" {
+		gss.Logging.SdkServer = "Info"
+	}
+}
+
+// applyContainerDefaults applies the container defaults
 func (gss *GameServerSpec) applyContainerDefaults() {
 	if len(gss.Template.Spec.Containers) == 1 {
 		gss.Container = gss.Template.Spec.Containers[0].Name
--- a/pkg/apis/agones/v1/gameserver_test.go
+++ b/pkg/apis/agones/v1/gameserver_test.go
@@ -61,11 +61,12 @@ func TestGameServerApplyDefaults(t *testing.T) {
 	t.Parallel()
 
 	type expected struct {
-		protocol   corev1.Protocol
-		state      GameServerState
-		policy     PortPolicy
-		health     Health
-		scheduling apis.SchedulingStrategy
+		protocol          corev1.Protocol
+		state             GameServerState
+		policy            PortPolicy
+		health            Health
+		scheduling        apis.SchedulingStrategy
+		sdkServerLogLevel string
 	}
 	data := map[string]struct {
 		gameServer GameServer
@@ -93,6 +94,7 @@ func TestGameServerApplyDefaults(t *testing.T) {
 					InitialDelaySeconds: 5,
 					PeriodSeconds:       5,
 				},
+				sdkServerLogLevel: "Info",
 			},
 		},
 		"defaults on passthrough": {
@@ -116,6 +118,7 @@ func TestGameServerApplyDefaults(t *testing.T) {
 					InitialDelaySeconds: 5,
 					PeriodSeconds:       5,
 				},
+				sdkServerLogLevel: "Info",
 			},
 		},
 		"defaults are already set": {
@@ -138,6 +141,7 @@ func TestGameServerApplyDefaults(t *testing.T) {
 								{Name: "testing", Image: "testing/image"},
 								{Name: "testing2", Image: "testing/image2"}}},
 					},
+					Logging: Logging{SdkServer: "Info"},
 				},
 				Status: GameServerStatus{State: "TestState"}},
 			container: "testing2",
@@ -152,6 +156,7 @@ func TestGameServerApplyDefaults(t *testing.T) {
 					InitialDelaySeconds: 11,
 					PeriodSeconds:       12,
 				},
+				sdkServerLogLevel: "Info",
 			},
 		},
 		"set basic defaults on static gameserver": {
@@ -173,6 +178,7 @@ func TestGameServerApplyDefaults(t *testing.T) {
 					InitialDelaySeconds: 5,
 					PeriodSeconds:       5,
 				},
+				sdkServerLogLevel: "Info",
 			},
 		},
 		"health is disabled": {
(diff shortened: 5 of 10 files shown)


aLekSer:
What is left: documentation update and UT.

aLekSer:
Build failed because I should have performed `make gen-api-docs`
```
Step #15 - "tests": includes/website.mk:96: recipe for target 'test-gen-api-docs' failed
```

pkg/apis/agones/v1/gameserver.go
```diff
@@ -135,6 +135,8 @@ type GameServerSpec struct {
 	Health Health `json:"health,omitempty"`
 	// Scheduling strategy. Defaults to "Packed".
 	Scheduling apis.SchedulingStrategy `json:"scheduling,omitempty"`
+	// LogLevel specifies the log level for gameserver sidecar.
+	LogLevel string `json:"loglevel,omitempty"`
```
markmandel: Random thought: Is "loglevel" too generic? Considering it's just for the sdk sidecar - should we pick a name that's more specific? `sdk-server-loglevel` (seems long), not sure what the right answer is here,  figured I would ask the question first.

Just wondering if the name might be confusing.

aLekSer: yes, you are right, that would suit better.

aLekSer:
Flaky E2E test:
```
Step #20: --- FAIL: TestFleetScaleUpEditAndScaleDown (0.01s)
Step #20: --- FAIL: TestFleetScaleUpEditAndScaleDown/Use_fleet_Patch_false (156.24s)
Step #20: fleet_test.go:120: 
Step #20: Error Trace:	fleet_test.go:120
Step #20: Error: Expected nil, but got: &errors.errorString{s:"timed out waiting for the condition"}
Step #20: Test: TestFleetScaleUpEditAndScaleDown/Use_fleet_Patch_false
```

site/content/en/docs/Reference/gameserver.md
```diff
@@ -49,6 +49,8 @@ spec:
     # Minimum consecutive failures for the health probe to be considered failed after having succeeded.
     # Defaults to 3. Minimum value is 1
     failureThreshold: 3
+  # SdkServer Log Level configuration. Could be any of panic, fatal, error, warn, warning, info, debug and trace. Defaults to info
```
markmandel: we should `feature` wrap these items.

install/yaml/install.yaml
```diff
@@ -860,6 +882,17 @@ spec:
                             type: integer
                             minimum: 1
                             maximum: 65535
+                    loglevel:
```
markmandel: Pretty sure this validation also needs to be changed, with the rename?

aLekSer: yes, that's true

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -92,6 +92,17 @@ properties:
               type: integer
               minimum: 1
               maximum: 65535
+      sdk-server-loglevel:
+        type: string
+        enum:
```
roberthbailey: I don't think we should expose so many log levels. This seems to be externalizing the fact that we are using logrus and would make it nearly impossible to change our logging library in the future. I feel like there are really three categories that people want:

1. Default (what we have today, which in basically the info level)
2. Errors and above
3. Debugging - everything

What if we abstracted the log levels from logrus and just offered something smaller like those three? Alternatively, we could give people a way to specifically disable the ping logs since those are the ones that cause the most noise in the log file.

site/content/en/docs/Reference/gameserver.md
aLekSer: Done

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
aLekSer: Yes, less level would be more concise. I will do the change cause that would actually give more space for the changes in the future.

markmandel:
Looks like the CRD docs need to be manually updated, before it will pass tests.

markmandel:
Ah, might actually be:
```
go: git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /go/pkg/mod/cache/vcs/83dba939f95a790e497d565fc4418400145a1a514f955fa052f662d56e920c3e: exit status 128:
	fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection timed out
```

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -92,6 +92,17 @@ properties:
               type: integer
               minimum: 1
               maximum: 65535
+      sdkServerLogLevel:
+        type: string 
+        description: |
+                  sdkServerLogLevel has three options:
+                  - "Info" (default) output SDK server messages of all log levels except debug messages
```
roberthbailey: Suggestions for making this a bit more consistent:

```
gameserverSidecarLogLevel has three options:
- "Info" (default) The game server sidecar will output all messages except for debug messages
- "Debug" The game server sidecar will output all messages including debug messages
- "Error" The game server sidecar will only output error messages

File: install/helm/agones/templates/crds/_gameserverspecvalidation.yaml [hunk 6]
```diff
@@ -92,6 +92,17 @@ properties:
               type: integer
               minimum: 1
               maximum: 65535
+      sdkServerLogLevel:
```
roberthbailey: You use sdk server in some places and game server sidecar in others. I'm not sure which is better, but we should be consistent one way or the other. I'm a bit partial to game server sidecar, since users of agones won't realize that the sidecar has to do with the SDK and because when you are looking at the logs in a k8s cluster those are the words used to name the container we inject into the pod with the game server container. 

I'd suggest making this field be `gameserverSidecarLogLevel`.

markmandel: So this took me back - I didn't realise that was what we called the sdk server sidecar container!
`"agones-gameserver-sidecar"` - in retrospect, possibly not the best name! But agree, consistency is key.

I guess my only question is - do we ever thing we'd add another sidecar of some sort? In which case, should we be specific about which sidecar's log level we are manipulating?

roberthbailey: > I guess my only question is - do we ever thing we'd add another sidecar of some sort? 

Are you suggesting making it `sidecarLogLevel` for now (shorter) or are you saying that we might have another sidecar for gameservers and that `gameserverSidecarLogLevel` isn't specific enough?

roberthbailey: Maybe the gameserver prefix is unnecessary?

markmandel: My concern was that "sidecar" could be too generic - that's why I was thinking sdk-server.

My other concern would be that "sidecar" is a well known K8s pattern, but I'm not sure if its knowledge we can assume across all our users?

roberthbailey: I'd suggest either `agonesGameserverSidecarLogLevel` (if we want to exactly match the name of the sidecar container) or `gameserverSidecarLogLevel` (if we want a shorter field name). I'm inclined to use the longer name if we aren't going to shorten it all the way to "sidecar".

markmandel: I think `agones` is redundant, since we're on an Agones GameServer. Ditto for `gameserverSidecarLogLevel` - we are in a gameServer already.

Maybe I wasn't clear - I prefer `sdkServerLogLevel` because it's specific to which thing it is being set to. "sidecar" could essentially be anything, and is also confusing to those who don't know Kubernetes patterns. What if the user creates their own sidecars, are we somehow overwriting their log level? If we ever add another sidecar for Agones GameServers (not likely, but possible), now we can't configure log levels for each one individually, because we don't have a name for each one.

Does that make more sense?

aLekSer: I also like sdkServerLogLevel more.
However I can change parameter to gameserverSidecarLogLevel if it would be more natural to the users.

aLekSer:
Updated this PR as per latest Mark comment, updated the docs also.
If we need to change the ENV variables like the following it would be another PR.
```
        - name: SIDECAR_IMAGE # overwrite the GameServer sidecar image that is used
```

examples/gameserver.yaml
```diff
@@ -35,7 +35,7 @@ spec:
   ports:
     # name is a descriptive name for the port
   - name: default
-    # portPolicy has two options:
+    # portPolicy has three options:
     # - "Dynamic" (default) the system allocates a free hostPort for the gameserver, for game clients to connect to
     # - "Static", user defines the hostPort that the game client will connect to. Then onus is on the user to ensure that the
     # - "Passthrough" dynamically sets the `containerPort` to the same value as the dynamically selected hostPort.
```
markmandel: Does this need an example in it, like we have in the fleet.yaml?

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
markmandel: I had an idea this morning in the shower :shower: 

I think we should have a logging section. Looking at commentary in #1036 , it's possible we will have more than 1 sidecar. In which case, we likely don't want multiple top spec level logging config params.

So howabout this:

```yaml
spec:
  # ... other fields ... 
  logging:
    sdkServer: "warning"
```

Then if we have to add more logging parameters, they aren't peppered through the config, they can be scoped under this `logging` element.

WDYT?

markmandel: Or maybe that should be `loggingLevel` ? (or `level` should be a sub-element to `logging` ?) Not 100% sure. But I think this is on a better path.

examples/gameserver.yaml
markmandel: Should we also have a log level example in this file?

aLekSer: Yes, definitely. I think we can omit updating fleet.yaml? Or we should update both?

aLekSer:
Updated PR and tested with different gameserver and fleet configurations.

examples/gameserver.yaml
aLekSer: Added new parameter to `example/gameserver.yaml`

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -92,6 +92,21 @@ properties:
               type: integer
               minimum: 1
               maximum: 65535
+      logging:
+        type: object
+        title: Define logging level for game server sidecars
```
roberthbailey: nit: Maybe say "logging levels for agones system components"

pkg/apis/agones/v1/gameserver.go
```diff
@@ -133,12 +133,20 @@ type GameServerSpec struct {
...
-	// Scheduling strategy. Defaults to "Packed".
+	// Scheduling strategy. Defaults to "Packed"
 	Scheduling apis.SchedulingStrategy `json:"scheduling,omitempty"`
+	// Logging specifies the log level for gameserver sidecars
+	Logging Logging `json:"logging,omitempty"`
 	// Template describes the Pod that will be created for the GameServer
 	Template corev1.PodTemplateSpec `json:"template"`
 }
 
+// Logging contains log levels for Gameserver Pod containers
```
roberthbailey: nit: Suggest replacing "Gameserver Pod containers" with "Agones system containers"

pkg/sdkserver/sdkserver.go
```diff
@@ -192,8 +193,21 @@ func (s *SDKServer) Run(stop <-chan struct{}) error {
...
+		s.logLevel = "info"
+	} else {
+		s.logLevel = strings.ToLower(gs.Spec.Logging.SdkServer)
+	}
+	s.logger.WithField("logLevel", s.logLevel).Info("Setting LogLevel configuration")
+	level, err := logrus.ParseLevel(s.logLevel)
+	if err == nil {
+		s.logger.Logger.SetLevel(level)
+	} else {
+		s.logger.WithError(err).Info("Specified wrong Logging.SdkServer")
```
roberthbailey: what log level is actually being used by logrus in this case? I assume it's info, in which case we should also set s.logLevel to "info" so that they are consistent.

roberthbailey: on second thought, is there a reason that we store loglevel in the struct? why not make it a local variable?

aLekSer: yes, you are right it should be simplified

aLekSer:
Applied these comments. @roberthbailey you can review one more time.

pkg/sdkserver/sdkserver.go
aLekSer: doing it as a member of struct was intended to support changing the logLevel on the go, by configuring GS Spec.

aLekSer:
I have merged master to the branch.

aLekSer:
Updated publishVersion to new one:
{{% feature publishVersion="1.1.0" %}}

Also I have found an issue with `make gen-api-docs` that if PR moves through the release, then it doesn't see new changes and doesn't update versions accordingly. 
So it was needed to perform an extra step before the command:
`git checkout master ../site/content/en/docs/Reference/agones_crd_api_reference.html`

aLekSer:
Build failed for some strange reason - no status comment was generated here on `github` PR:
```
Status	
 Build failed
Status detail	
failed unmarshalling build config cloudbuild.yaml: proto: google.devtools.cloudbuild.v1.Build.timeout|"timeout": google.protobuf.Duration: invalid duration value "1h"
Build id	
0867e0f5-6a9b-48b8-92bb-f9abb0d96c16
Image	
—
Trigger	
Push to github_pr_1007 branch (agones)
```

pkg/apis/agones/v1/gameserver.go
```diff
@@ -133,12 +133,20 @@ type GameServerSpec struct {
 	Ports []GameServerPort `json:"ports"`
 	// Health configures health checking
 	Health Health `json:"health,omitempty"`
-	// Scheduling strategy. Defaults to "Packed".
+	// Scheduling strategy. Defaults to "Packed"
 	Scheduling apis.SchedulingStrategy `json:"scheduling,omitempty"`
+	// Logging specifies the log level for gameserver sidecars
```
roberthbailey: Can you make this match the comment below, e.g. "Logging specifies the log levels for Agones system containers"?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -133,12 +133,20 @@ type GameServerSpec struct {
...
+	Logging Logging `json:"logging,omitempty"`
 	// Template describes the Pod that will be created for the GameServer
 	Template corev1.PodTemplateSpec `json:"template"`
 }
 
+// Logging contains log levels for Agones system containers
+type Logging struct {
+	// SdkServer is a log level for SDK server (sidecar) logs
```
roberthbailey: Please add `Defaults to "Info"` at the end of this comment.

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -152,12 +155,14 @@ func TestGameServerApplyDefaults(t *testing.T) {
 					InitialDelaySeconds: 11,
 					PeriodSeconds:       12,
 				},
+				sdkServerLogLevel: "Info",
```
roberthbailey: this test is "defaults are already set" so the log level should be set in the gameserver as well as in the expected result.

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -205,17 +212,19 @@ func TestGameServerApplyDefaults(t *testing.T) {
 							Protocol:      corev1.ProtocolTCP,
 						},
 					},
-					Health: Health{Disabled: true},
+					Health:  Health{Disabled: true},
+					Logging: Logging{SdkServer: "Debug"},
```
roberthbailey: rather than piggyback on an existing test case ("convert from legacy single port to multiple" to test applying a custom log level, please add a new test case that specifically tests setting the log level.

site/content/en/docs/Reference/fleet.md
```diff
@@ -57,6 +57,11 @@ spec:
       health:
         initialDelaySeconds: 30
         periodSeconds: 60
+{{% feature publishVersion="1.0.0" %}}
```
roberthbailey: This should now be 1.1.0

pkg/apis/agones/v1/gameserver_test.go
aLekSer: OK, adding a separate test

aLekSer:
Applied comments and added `agones_crd_api_reference.html`

pkg/apis/agones/v1/gameserver.go
aLekSer: Will update in a moment to:
```
 Logging specifies log levels for Agones system containers
```

pkg/apis/agones/v1/gameserver_test.go
aLekSer: Done

Issue #398, opened 2018-10-24
Moving cluster to a new node pool doesn't recreate all fleets


KamiMay:
I've noticed something weird today. I needed to swap node pool in GKE so I created new node pool and deleted old one. I expected all instances in the old node pool to recover in the new one after some time. However in my particular case I could only see 1 of the 3 servers in `workloads` page on GCloud. So I checked fleets to see if it has min availability which was 1 of each kind = 3. And `kubectl describe fleets`  indicated that 3 servers were online and available, however when I tried to connect to one that was listed but not in `workloads` it failed to connect to it, I was able to connect to the one appearing in the `workloads`, but not others. I had to delete fleets and recreate them for them to appear and work correctly again.

markmandel:
I have a strong feeling this is because if a pod get's deleted, the backing GameServer is left in a zombie state (i.e. not deleted along with it).

We should implement functionality to say that if Pod get's removed, the owning GameServer should be deleted too. This should solve this issue.

markmandel:
Actually - I'm not sure what this is - I tested deleting the backing Pod from a GameServer, and the GameServer gets deleted. More investigation required!

KamiShepard:
I think best way to recreate it is to follow what I have done. Because it has recreated one of the fleets but not the other two. So it might be a random factor in there. Would suggest trying it with a few different fleets. It seems to be random. Sometimes happens sometimes doesn't.

KamiShepard:
I will try reproduce this issue today and provide all the relevant details along the way

KamiShepard:
Before migrating to a new node pool:

- I have 3 instances running. All of thaem are 1vCPU 3.75GB RAM
- I have 3 fleets running, each of them have 1 replica set + agones-controller and matchmaker.
- It looks like this on gcloud:
![image](https://user-images.githubusercontent.com/2967343/48638466-a2de2a80-e9c8-11e8-8366-1e486fc068c9.png)
- At this point when i run `kubectl describe fleets` I get following:

```
Name:         deathmatch-server
Namespace:    default
Labels:       <none>
Annotations:  gameMode=deathmatch
              kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"stable.agones.dev/v1alpha1","kind":"Fleet","metadata":{"annotations":{"gameMode":"deathmatch"},"name":"deathmatch-server","namespace":"d...
API Version:  stable.agones.dev/v1alpha1
Kind:         Fleet
Metadata:
[...]

PR #1008, opened 2019-08-22
Documentation on how to upgrade Agones and/or Kubernetes.

### LINKED ISSUE #555: Document how to do upgrades of Agones
Right now the upgrade experience is likely quite tricky -- and risky.

Best bet right now may be to advocate spinning up a new cluster, and moving to a new cluster at upgrade time, and if so, we should document this.

RBAC rule changes particularly frighten me for upgrades.

We should test out some scenarios and see where things fall over, and document them (likely also find some issues a long the way).

markmandel:
Documentation is deliberately the safest upgrade paths.

Closes #555

### DIFF
--- a/site/content/en/docs/Installation/upgrading.md
+++ b/site/content/en/docs/Installation/upgrading.md
@@ -1,0 +1,116 @@
+---
+title: "Upgrading Agones and Kubernetes"
+linkTitle: "Upgrading"
+weight: 100
+date: 2019-08-16T00:19:19Z
+description: >
+  Strategies and techniques for managing Agones and Kubernetes upgrades in a safe manner
+---
+
+> Note: Whichever approach you take to upgrading Agones, make sure to test it in your development environment 
+> before applying it to production 
+
+## Upgrading Agones
+
+The following are strategies for safely upgrading Agones from one version to another. They may require adjustment to 
+your particular game architecture but should provide a solid foundation for updating Agones safely.
+
+The recommended approach is to use [multiple clusters](#multiple-clusters), such that the upgrade can be tested
+gradually with production load and easily rolled back if need arises.
+
+### Multiple Clusters
+
+Essentially, we want to do a [blue/green deployment](https://martinfowler.com/bliki/BlueGreenDeployment.html) to the 
+new version of Agones.  This means that we can slowly migrate our game servers from the older version's cluster (blue)
+to the new version's cluster (green), and ensure nothing surprising happens during this process.
+
+This also allows easy rollback to the previous infrastrucutre that we already know to be working in production, with
+minimal interruptions to player experience.
+
+The following are steps to implement this:
+
+1. Create a new cluster of the same size or smaller as the current cluster.
+2. Install the new version of Agones on the new cluster.
+3. Deploy the same set of Fleets and/or GameServers from the old cluster into the new cluster.
+4. With your matchmaker, start sending a small percentage of your matched players' game sessions to the new cluster.
+5. Assuming everything is working successfully on the new cluster, slowly increase the percentage of matched sessions to the new cluster, until you reach 100%.
+6. Once you are comfortable with the stability of the new cluster with the new Agones version, shut down the old cluster.
+7. Congratulations - you have now upgraded to a new version of Agones! 👍
+
+### Single Cluster
+
+If you are upgrading a single cluster, we recommend creating a maintenance window, in which your game goes offline
+for the period of your upgrade, as there will be a short period in which Agones will be non-responsive during the upgrade.
+
+### Installation with install.yaml
+
+If you installed [Agones with install.yaml]({{< relref "_index.md#install-with-yaml" >}}), then you will need to delete
+the previous installation of Agones before upgrading to the new version, as we need to remove all of Agones before installing
+the new version.
+
+1. Start your maintenance window.
+1. Delete the current set of Fleets and/or GameServers in your cluster.
+1. Make sure to delete the same version of Agones that was previously installed, for example:
+   `kubectl delete -f https://raw.githubusercontent.com/googleforgames/agones/<old-release-version>/install/yaml/install.yaml`
+1. Install Agones [with install.yaml]({{< relref "_index.md#install-with-yaml" >}}).
+1. Deploy the same set of Fleets and/or GameServers back into the cluster.
+1. Run any other tests to ensure the Agones installation is working as expected.
+1. Close your maintenance window.
+7. Congratulations - you have now upgraded to a new version of Agones! 👍
+
+### Installation with Helm
+
+Helm features capabilities for upgrading to newer versions without having to delete the older version. For details on
+how to use Helm for upgrades, see the [helm upgrade](https://helm.sh/docs/helm/#helm-upgrade) documentation.
+
+Given the above, the steps for upgrade are simpler:
+
+1. Start your maintenance window.
+1. Run `helm upgrade` with the appropriate arguments, such a `--version`, for your specific upgrade
+1. Run any other tests to ensure the Agones installation is working as expected.
+1. Close your maintenance window.
+7. Congratulations - you have now upgraded to a new version of Agones! 👍
+
+## Upgrading Kubernetes
+
+The following are strategies for safely upgrading the underlying Kubernetes cluster from one version to another.
+They may require adjustment to your particular game architecture but should provide a solid foundation for updating your cluster safely.
+
+The recommended approach is to use [multiple clusters](#multiple-clusters-1), such that the upgrade can be tested
+gradually with production load and easily rolled back if need arises.
+
+### Multiple Clusters
+
+This process is very similar to the [Agones: Multiple Cluster](#multiple-clusters) approach above.
+
+Essentially, we want to do a [blue/green deployment](https://martinfowler.com/bliki/BlueGreenDeployment.html) to the new 
+version of Kubernetes.  This means that we can slowly migrate our game servers from the older version's cluster (blue) 
+to the new version's cluster (green), and ensure nothing surprising happens during this process.
+
+This also allows easy rollback to the previous infrastructure that we already know to be working in production,
+with minimal interruptions to player experience.
+
+The following are steps to implement this:
+
+1. Create a new cluster of the same size or smaller as the current cluster, with the new version of Kubernetes
+2. Install the same version of Agones on the new cluster, as you have on the previous cluster.
+3. Deploy the same set of Fleets and/or GameServers from the old cluster into the new cluster.
+4. With your matchmaker, start sending a small percentage of your matched players' game sessions to the new cluster.
+5. Assuming everything is working successfully on the new cluster, slowly increase the percentage of matched sessions to the new cluster, until you reach 100%.
+6. Once you are comfortable with the stability of the new cluster with the new Kubernetes version, shut down the old cluster.
+7. Congratulations - you have now upgraded to a new version of Kubernetes! 👍
+
+### Single Cluster
+
+If you are upgrading a single cluster, we recommend creating a maintenance window, in which your game goes offline
+for the period of your upgrade, as there will be a short period in which Agones will be non-responsive during the node
+upgrades.
+
+1. Start your maintenance window.
+1. Scale your Fleets down to 0 and/or delete your GameServers.
+1. Start and complete your node upgrades.
+1. Start and complete you master upgrade.
+1. Scale your Fleets back up and/or recreate your GameServers. 
+1. Run any other tests to ensure the Agones installation is still working as expected.
+1. Close your maintenance window.
+7. Congratulations - you have now upgraded to a new version of Kubernetes! 👍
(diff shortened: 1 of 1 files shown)


markmandel:
Preview of the page: https://7226a9c-dot-preview-dot-agones-images.appspot.com/site/docs/installation/upgrading/

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+version of Kubernetes.  This means that we can slowly migrate our game servers from the older version's cluster (blue) 
+to the new version's cluster (green), and ensure nothing surprising happens during this process.
+
+This also allows easy rollback to the previous infrastructure that we already know to be working in production,
+with minimal interruptions to player experience.
+
+The following are steps to implement this:
+
+1. Create a new cluster of the same size or smaller as the current cluster, with the new version of Kubernetes
+2. Install the same version of Agones on the new cluster, as you have on the previous cluster.
```
roberthbailey: This is useful if you want to separate the upgrade of k8s vs agones. But I imagine most folks will actually want to do them together (for minor upgrades) since agones will only be stress tested against a single patch version of k8s at a time.

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+7. Congratulations - you have now upgraded to a new version of Kubernetes! 👍
+
+### Single Cluster
+
+If you are upgrading a single cluster, we recommend creating a maintenance window, in which your game goes offline
+for the period of your upgrade, as there will be a short period in which Agones will be non-responsive during the node
+upgrades.
+
+1. Start your maintenance window.
+1. Scale your Fleets down to 0 and/or delete your GameServers.
```
roberthbailey: can you explain why? will node upgrades kill allocated game servers? is it possible to set PDBs and/or graceful termination periods on them automatically to make node upgrades work without disruption?

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+1. Start and complete your node upgrades.
```
roberthbailey: The k8s docs recommend doing master upgrades before node upgrades.

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+1. Start and complete you master upgrade.
```
roberthbailey: upgrade(s) -- since folks, especially on prem or when not using a provider, will likely have a multi-master setup.

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+
+The following are strategies for safely upgrading Agones from one version to another. They may require adjustment to 
+your particular game architecture but should provide a solid foundation for updating Agones safely.
+
+The recommended approach is to use [multiple clusters](#multiple-clusters), such that the upgrade can be tested
+gradually with production load and easily rolled back if need arises.
+
+### Multiple Clusters
+
+Essentially, we want to do a [blue/green deployment](https://martinfowler.com/bliki/BlueGreenDeployment.html) to the
```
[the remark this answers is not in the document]
markmandel: Lemme see if I can reword it. The process is essentially the same, but slower (:man_shrugging:) but if it doesn't work, maybe we should cut it.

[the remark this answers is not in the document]
markmandel: Actually, changed my mind. Cutting it entirely, and rewording.

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+5. Assuming everything is working successfully on the new cluster, slowly increase the percentage of matched sessions to the new cluster, until you reach 100%.
+6. Once you are comfortable with the stability of the new cluster with the new Agones version, shut down the old cluster.
+7. Congratulations - you have now upgraded to a new version of Agones! 👍
+
+### Single Cluster
+
+If you are upgrading a single cluster, we recommend creating a maintenance window, in which your game goes offline
+for the period of your upgrade, as there will be a short period in which Agones will be non-responsive during the upgrade.
+
+### Installation with install.yaml
```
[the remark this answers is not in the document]
markmandel: Whoops. Good catch.

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+1. Run any other tests to ensure the Agones installation is working as expected.
+1. Close your maintenance window.
+7. Congratulations - you have now upgraded to a new version of Agones! 👍
+
+### Installation with Helm
+
+Helm features capabilities for upgrading to newer versions without having to delete the older version. For details on
+how to use Helm for upgrades, see the [helm upgrade](https://helm.sh/docs/helm/#helm-upgrade) documentation.
+
+Given the above, the steps for upgrade are simpler:
```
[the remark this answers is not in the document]
markmandel: I'd _really_ like to break apart and reorganise the installation docs as they currently stand -- they are one big mung of things, that diverge only sometimes for random topics.

I started to dig into what you have above, because I think we should do it, but it just made me want to pull this whole thing apart more - as then i want a section on "why choose which installation method", but then I was just unravelling more and more string as I pulled on the thread.

TL;DR - can we do this later?

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,116 @@
...
+
+1. Start your maintenance window.
+1. Run `helm upgrade` with the appropriate arguments, such a `--version`, for your specific upgrade
+1. Run any other tests to ensure the Agones installation is working as expected.
+1. Close your maintenance window.
+7. Congratulations - you have now upgraded to a new version of Agones! 👍
+
+## Upgrading Kubernetes
```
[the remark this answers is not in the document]
markmandel: So I'm not sure what the right text here is. I like what you have here re: risk on minor vs patch upgrades, but I wonder if the text about doing them "in place" is the best advice. My concern is that if you do it in place, you have to have a maintenance window, because you can't do with live gameservers running. Even for patch upgrades, doing it with two clusters means you can do it while your game is active, without downtime.

So I'm not sure what the right thing to say here is :man_shrugging: Maybe just the section on differences in risk between versions? Especially if you have custom code that also interacts with the K8s api?

site/content/en/docs/Installation/upgrading.md
markmandel: Added a note to this effect.  PTAL. I don't _think_ we need a combined K8s + Agones upgrade at the same time doc? Or do you think we do?

site/content/en/docs/Installation/upgrading.md
markmandel: Ah good catch, switching.

I went looking for a URL that called this out in the K8s docs, but couldn't fine one. Do you have one at hand? Would be good to link to it.

site/content/en/docs/Installation/upgrading.md
markmandel: Added some words to the docs - but the TL;DR is basically to prevent #398 .

We can mitigate this with separate nodepools for the controller, but for this guide, I figured it was best to write the simplest version I could, with strategies I knew would work for all circumstances. I figured we could then iterate out to more sophisticated upgrade paths as needed. 

The other option is to split this into a "separate controller node group" section, and a "everything on the same node group" section, with seperate upgrade strategies, but wasn't sure if that was too much for this instance.

I err'd towards a path that I knew would work for all use cases, since I figured stable was best for 1.0.

markmandel: > is it possible to set PDBs and/or graceful termination periods on them automatically to make node upgrades work without disruption

On this note - it's _possible_ (probably more graceful termination, we don't have control over which Pods get terminated with PDBs AFAIK?), but I don't think it's been thoroughly tested as a strategy, so I didn't want to document it at this stage. Just trying to be safe, essentially.

site/content/en/docs/Installation/upgrading.md
roberthbailey: I agree that we don't need another section (or doc) on doing them together. Just a note that as long as you are upgrading agones and either switching clusters or having maintenance that you might as well make sure that you are running the right k8s version underneath it.

site/content/en/docs/Installation/upgrading.md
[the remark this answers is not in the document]
roberthbailey: This one is tricky. I'm not sure if it's reasonable to force new cluster creation and workload migration each time there is a security patch (which seem relatively frequent). But I see your point about maintenance windows if you do in-place.... Leaving it as-is seems reasonable for now, and we can work on some of the trickier edge cases in a follow up.

site/content/en/docs/Installation/upgrading.md
roberthbailey: That sounds reasonable. I agree that we can expand on the instructions over time.

site/content/en/docs/Installation/upgrading.md
roberthbailey: It took me a while looking, and the only place I could find it was buried in the version skew documentation: https://kubernetes.io/docs/setup/release/version-skew-policy/#kubelet

I know that having the master be newer than the nodes is enforced in GKE and kubeadm (and likely elsewhere) fwiw.

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,130 @@
...
+
+We essentially want to transition our GameServer allocations from a cluster with the old version of Kubernetes,
+to a cluster with the upgraded version of Kubernetes while ensuring nothing surprising 
+happens during this process.
+
+This also allows easy rollback to the previous infrastructure that we already know to be working in production, with
+minimal interruptions to player experience.
+
+
+Essentially, we want to do a [blue/green deployment](https://martinfowler.com/bliki/BlueGreenDeployment.html) to the new
```
roberthbailey: Did you mean to leave this part in here, after removing it from the above section?

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,130 @@
...
+The following are strategies for safely upgrading the underlying Kubernetes cluster from one version to another.
+They may require adjustment to your particular game architecture but should provide a solid foundation for updating your cluster safely.
+
+The recommended approach is to use [multiple clusters](#multiple-clusters-1), such that the upgrade can be tested
+gradually with production load and easily rolled back if need arises.
+
+Agones currently has only a [single supported Kubernetes version]({{< relref "_index.md#usage-requirements" >}}) for each version,
+so it is recommended to do supported minor (1.12.1 ➡ 1.13.2) Kubernetes version upgrades at the same time as matching Agones upgrades.
+
+Patch upgrades (e.g. 1.12.1 ➡ 1.12.3) within the same major version of Kubernetes can be done at any time.
```
roberthbailey: s/major/minor

site/content/en/docs/Installation/upgrading.md
```diff
@@ -0,0 +1,130 @@
...
+## Upgrading Kubernetes
+
+The following are strategies for safely upgrading the underlying Kubernetes cluster from one version to another.
+They may require adjustment to your particular game architecture but should provide a solid foundation for updating your cluster safely.
+
+The recommended approach is to use [multiple clusters](#multiple-clusters-1), such that the upgrade can be tested
+gradually with production load and easily rolled back if need arises.
+
+Agones currently has only a [single supported Kubernetes version]({{< relref "_index.md#usage-requirements" >}}) for each version,
+so it is recommended to do supported minor (1.12.1 ➡ 1.13.2) Kubernetes version upgrades at the same time as matching Agones upgrades.
```
roberthbailey: i'm not sure how i feel about this arrow... it's a bit distracting in the rendered paragraph. Feel free to leave it in if you like the way it looks. :)

site/content/en/docs/Installation/upgrading.md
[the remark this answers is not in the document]
markmandel: Agreed on tricky. Also, depending on your Kubernetes control layer (cloud, kubeadm, etc), there are likely different techniques you can deploy. e.g. nodepools are GKE specific, but maybe with a combo of regional cluster + multple nodepools, you can do a complete upgrade in cluster, within minimal/practically no downtime?

site/content/en/docs/Installation/upgrading.md
markmandel: Cool. I'll not worry about a link. We can have the same order in our docs too :+1:

site/content/en/docs/Installation/upgrading.md
markmandel: Doh. Thanks.

site/content/en/docs/Installation/upgrading.md
markmandel: -> feels so boring compared to :arrow_right: 

:smile:

site/content/en/docs/Installation/upgrading.md
markmandel: Done!

> Agones currently has only a [single supported Kubernetes version]({{< relref "_index.md#usage-requirements" >}}) for each version,
so it is recommended to do supported minor (e.g. 1.12.1 ➡ 1.13.2) Kubernetes version upgrades at the same time as matching Agones upgrades.

PR #1009, opened 2019-08-23
Fix outdated links in comments

aLekSer:
Remove some redundant links to ObjectMeta API reference. So we don't
need to update all of them on Kubernetes version changes.


site/content/en/docs/Reference/gameserver.md
```diff
@@ -13,7 +13,7 @@ A full GameServer specification is available below and in the {{< ghlink href="e
 apiVersion: "agones.dev/v1"
 kind: GameServer
 # GameServer Metadata
-# https://v1-8.docs.kubernetes.io/docs/api-reference/v1.8/#objectmeta-v1-meta
+# https://v1-12.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#objectmeta-v1-meta
```
roberthbailey: In the other cases you removed the comments linking to the k8s docs but in this case you updated the link. I'm curious why you didn't update the link in the other places as well.

aLekSer: The idea was to keep this link in one place, but than realize that this could be helpful to someone if we have for any CRD. Will update this PR

aLekSer:
@roberthbailey  I have updated all links, now files are consistent.

PR #1023, opened 2019-08-29
add vendor grpc third_party to be referenced by protos

pooneh-m:
Following these steps: https://github.com/googleforgames/agones/blob/master/build/README.md#dependencies
Third_party directory does not get added with go mod vendor as described.

### DIFF
--- a/vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis/LICENSE
+++ b/vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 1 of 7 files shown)


markmandel:
Just so I'm 100% clear, did you add this in manually?

Happy to approve this going through,  I'm just concerned it might get accidentally deleted on a `go mod` operation that updates the vendor directory -- such as updating the grpc-gateway library itself, will that delete this third_party directory? or will we need to manually ensure it always is there?

markmandel:
Sounds like this is a fun(?) common problem: https://github.com/golang/go/issues/26366

pooneh-m:
Exactly, running go mod vendor recreates the vendor directory therefore removes the third_party directory.

I followed [this](https://github.com/googleforgames/agones/blob/master/build/README.md#dependencies) to add the third_party directory, which is exactly what I needed:
AGONES_PATH=/wherever/your/agones/path/is
cp -R $GOPATH/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.5.1/third_party $AGONES_PATH/vendor/github.com/grpc-ecosystem/grpc-gateway/

Ideally we should remove the dependency to vendor and only use go.mod?

markmandel:
Got it - so in a previous life, someone already encountered this :+1: 

Maybe in a future PR we could do something like `make go-deps` and it would do all the appropriate things...? But for the future.

I'm assuming in a future PR we are going to remove the protos in /protos and update the tooling to reference this `third_party` directory instead?

pooneh-m:
Good idea.

Issue #485, opened 2019-01-14
FleetAllocation creation error lacks of information


Kuqd:
If your current fleet is full and you try to allocate a new GameServer via FleetAllocation as such

```golang
fa, err := client.Agones().StableV1alpha1().FleetAllocations("").Create(&newfa)
```

the error contains

```golang
&k8serrors.StatusError{
    ErrStatus: v1.Status{
        Code:    500,
        Reason:  v1.StatusReasonInternalError,
        Status:  v1.StatusFailure,
        Message: "Internal error occurred: admission webhook \"mutations.stable.agones.dev\" denied the request: NotFound ",
    },
}
```

I think we should improve this so that we can handle gracefully this kind of error currently the only way to differentiate from `InternalServerError` is to use a function like this

```golang
func isUnallocatable(err error) bool {
	switch t := err.(type) {
	case k8serrors.APIStatus:
		return t.Status().Reason == v1.StatusReasonInternalError && t.Status().Status == v1.StatusFailure &&
			strings.Contains(t.Status().Message, "admission webhook \"mutations.stable.agones.dev\" denied the request: NotFound")
	}
	return false
}
```

This is may be obsolete as FleetAllocation will deprecated ?

markmandel:
Yes - this exact problem is solved by `GameServerAllocation` - in that it only returns an error if something goes horribly wrong.

If an allocation fails, it returns with a status of `UnAllocated` for exactly this scenario.

Kuqd:
Ok let’s close this then

PR #1025, opened 2019-08-30
Define the proto definition for the allocator service

pooneh-m:
Define the v1alpha1 gRPC API protobuf definition for the allocator service according to the [GameServerAllocation](https://github.com/googleforgames/agones/blob/master/pkg/apis/allocation/v1/gameserverallocation.go) Extension API server.

### DIFF
--- a/cmd/allocator/v1alpha1/allocation.proto
+++ b/cmd/allocator/v1alpha1/allocation.proto
@@ -1,0 +1,96 @@
+// Copyright 2019 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package v1alpha1;
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "google/api/annotations.proto";
+
+service AllocationService {
+ rpc PostAllocate(AllocationRequest) returns (AllocationResponse) {
+   option (google.api.http) = {
+     post: "/v1alpha1/gameserverallocation"
+     body: "*"
+   };
+ }
+}
+
+message AllocationRequest {
+  // The k8s namespace that is hosting the targetted fleet of gameservers to be allocated
+  string namespace = 1;
+
+  // If specified, multi-cluster policies are applied. Otherwise, allocation will happen locally.
+  MultiClusterSetting multiClusterSetting = 2;
+
+  // The required allocation. Defaults to all GameServers.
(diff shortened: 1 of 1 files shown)


pooneh-m:
Upgrading the protobugf package to fix the build error for proto3 dependency: https://github.com/googleforgames/agones/pull/1027

cmd/allocator/v1alpha1/allocation.pb.go
```diff
@@ -0,0 +1,539 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
markmandel: This will need a licence header as well. We have examples in the sdk gen tooling on how we did this.

pooneh-m: > Looks like this issue is the linter is failing on the generated code. I think you will need to have a line "This code was autogenerated. Do not edit directly." - which is included in https://github.com/googleforgames/agones/blob/master/build/boilerplate.go.txt
> 
> May be easier to get all the make commands working in this PR?

Thanks, yes I am working on the make commands. Will update this PR when done.

markmandel:
Just checking this is good for review - happy to do so, but noticed it's been pared back to just the proto. Wasn't sure if more was coming.

pooneh-m:
To keep the reviews focused, I broke down the PR to two, one for the proto and one for the generated code and the scripts for code generation. I'll send the latter PR by the end of today.

cmd/allocator/v1alpha1/allocation.proto
```diff
@@ -0,0 +1,97 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package v1alpha1;
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
```
markmandel: I had no idea this existed! Nice find!

cmd/allocator/v1alpha1/allocation.proto
```diff
@@ -0,0 +1,97 @@
...
+  // MetaPatch is optional custom metadata that is added to the game server at
+  // allocation You can use this to tell the server necessary session data
+  MetaPatch metaPatch = 6;
+}
+
+message AllocationResponse {
+  GameServerAllocationState state = 1;
+
+  // The allocation state
+  // TODO: instead of using state return 500 for unallocated and 409 for contention
```
markmandel: We should talk about this TODO more - you can see some background here, about why we made this decision:
https://github.com/googleforgames/agones/issues/485
https://github.com/googleforgames/agones/issues/436

We very deliberately didn't have an unfound game server be an error. I'm not sure it's appropriate here?

Also, I think it will be confusing to have 2 different approaches for inside cluster vs outside at this stage.

pooneh-m: Thanks for the explanation. For a RESTful API, the http status represents the status of the result and to me receiving 200 for an unsuccessful response is contradicting. In kubernetes world, status is a good representation of the state of a CRD, but this does not apply to http response standards. 

I think the allocation REST API should return a meaningful http status in case of failure instead of 200. For now, I removed the TODO and opened an issue to discuss this further.

markmandel: That's makes sense :+1: 

Will add some notes to the issue.

PR #1027, opened 2019-08-30
Upgrade protobuf go package dependency to v3

pooneh-m:
Upgrade go package dependency to 1.3.2 for golang/protobuf in go.mod and ran go mod vendor. 
The additional changes to pborman/uuid are due to outdated vendor.


markmandel:
The previous go protoc package (1.2.0) supported proto 3, so I'm not sure why we are upgrading? Is there a strong reason to do so?

Given that were aiming for 1.0 in the next release, I'm thinking we shouldn't do this to maintain a stable, known system, unless there is a very good reason to do this?

pooneh-m:
Makes sense. I am introducing an allocator API proto definition and getting dependency to the latest proto package, but it is not necessary.

PR #1029, opened 2019-09-03
Cpp SDK. Fixed regex for version detection. Fixed mingw build.

dsazonoff:
Fixed Agones version detection for CMake scripts (based on git tag).
Fixed mingw (windows) build of zlib and gRPC dependency.
Added more informative error for mingw build, if OpenSSL was not found (rare case); because building of OpenSSL with mingw in a cross-platform way is tricky (require CMake->MSYS->OpenSSL chain). I'm not sure that we should add an instruction how to build OpenSSL to our documentation. Prebuilt OpenSSL for windows may be found [there](https://slproweb.com/products/Win32OpenSSL.html). OpenSSL is required only for successful gRPC build.

### DIFF
--- a/sdks/cpp/CMakeLists.txt
+++ b/sdks/cpp/CMakeLists.txt
@@ -50,7 +50,7 @@ if (Git_FOUND)
         OUTPUT_VARIABLE output
     )
     if (${result} EQUAL 0)
-        string(REGEX MATCH "\\d+.\\d+.\\d+" AGONES_VERSION ${output})
+        string(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]+" AGONES_VERSION ${output})
     endif()
 else()
     message(WARNING "Git was not found. Current Agones version is set to ${AGONES_VERSION}")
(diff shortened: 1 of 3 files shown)


dsazonoff:
I'm not sure, why tests are failed. I didn't touch any documentation.
/assign markmandel

aLekSer:
I performed `make test-gen-api-docs` locally:
```bash
> git fetch https://github.com/GoogleCloudPlatform/agones.git pull/1029/head:pr_1029 && git checkout pr_1029        
> make test-gen-api-docs
{{% feature publishVersion="0.12.0" %}}
sort /Users/alex/go/src/agones.dev/agones/site/content/en/docs/Reference/agones_crd_api_reference.html > /tmp/result.sorted                                                                
diff -bB /tmp/result.sorted /tmp/generated.html.sorted
```
All is fine no errors. On CI build it generates an error, which is strange.

markmandel:
I would suggest rebasing master into your branch -- sounds like you are working off some super old code.

markmandel:
Ah, there was an issue which I just fixed #1031 - waiting for it to be merged. That will actually solve the build issue

roberthbailey:
#1031 has merged.

markmandel:
Calling @devjgm :wave:  - Please take a look

devjgm:
My cmake-fu is not very strong. So, I'll say this LGTM, but I am not a cmake expert and I'm the furthest thing from a windows/mingw expert.

PR #1031, opened 2019-09-03
Fix for git.apache.org being down.

markmandel:


markmandel:
Everything comes back to gen-crd-api-reference-docs - it cant generate docs because it _has_ to pull down a fresh clone of git.apache.org/thrift.git into the go-mod cache -- because our old version of opencensus has it as a dependency, even though it doesn't use it.

I've tried every hacky workaround I can think of... but nothing seems to work.

Anyone know a good way to mess around with the gomod cache?

markmandel:
I think I have a solution: https://twitter.com/paulfantom/status/1168500874718732288

Running tests now!

go.mod
```diff
@@ -55,3 +55,5 @@ require (
 )
 
 replace k8s.io/apimachinery => ./vendor_fixes/k8s.io/apimachinery
+
+replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
```
roberthbailey: I'm not sure I understand why we need both this _and_ the ENV variable for the go modules proxy. Shouldn't the env variable alone be sufficient (and a smaller change)? And then we can fix the dependencies when updating opencensus?

markmandel:
That's a good point. Lemme test that out and update 👍

markmandel:
Looks good - got rid of the changes to dependencies 👍

Issue #956, opened 2019-07-29
Game server container crash before Ready, should restart, not move to Unhealthy


markmandel:
**What happened**:

If the game server container crashes at any stage, the GameServer moved to Unhealthy straight away.

**What you expected to happen**:

According to: https://agones.dev/site/docs/guides/health-checking/

> If any of the GameServer container fails health checking, or exits before the GameServer moves to Ready then, it is restart as per the restartPolicy (which defaults to “Always”)

**How to reproduce it (as minimally and precisely as possible)**:

We will need a e2e test that has a CRASH command on udp-simple that does a `sdk.Exit(1)` and then test what happens when a crash occurs before `Ready` and after `Ready`.

**Anything else we need to know?**:

Logic can be found here:
https://github.com/googleforgames/agones/blob/master/pkg/gameservers/health.go#L87

**Environment**:
- Agones version: development
- Kubernetes version (use `kubectl version`): 1.12
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): helm
- Troubleshooting guide log(s): N/A
- Others:

Colstuwjx:
I also met this issue, version table see below:
```
agones: 0.11.0 
xonotic version: 0.5
```

BTW, it seems that our gameserver\fleet crd not maintains POD replica count, if I found the pod is unhealthy, and maunally deleted it, things would be broken, and gameserver/fleet would never try to recreate a new POD...

Thanks.

markmandel:
> BTW, it seems that our gameserver\fleet crd not maintains POD replica count, if I found the pod is unhealthy, and maunally deleted it, things would be broken, and gameserver/fleet would never try to recreate a new POD...

We have an e2e test for this, so _in theory_ this shouldn't happen, but if you are finding this an issue - please file a new bug with detailed replication instructions so that we can reproduce and fix. Thank you! 👍

aLekSer:
I was trying to develop such E2E test. It seems that we need a separate `simple-udp` server which would part of the time fail to start say randomly one time out of 2 (os.Exit() before Ready).
I was not able to send something to UDP server before Ready because it does not have allocated ports.
Added such check and test fail on if statement:
```
func SendGameServerUDP(gs *agonesv1.GameServer, msg string) (string, error) {
	if len(gs.Status.Ports) < 1 {
		return "", errors.New("Port was not assigned to the GameServer")
	}
        ..
}
```

markmandel:
[...]

PR #1038, opened 2019-09-05
E2E test for Unhealthy GameServer on process crash

markmandel:
More preparation for #956

### DIFF
--- a/pkg/gameservers/health.go
+++ b/pkg/gameservers/health.go
@@ -131,8 +131,9 @@ func (hc *HealthController) evictedPod(pod *corev1.Pod) bool {
 func (hc *HealthController) failedContainer(pod *corev1.Pod) bool {
 	container := pod.Annotations[agonesv1.GameServerContainerAnnotation]
 	for _, cs := range pod.Status.ContainerStatuses {
-		if cs.Name == container && cs.State.Terminated != nil {
-			return true
+		if cs.Name == container {
+			// sometimes on a restart, the cs.State can be running and the last state will be merged
+			return cs.State.Terminated != nil || cs.LastTerminationState.Terminated != nil
 		}
 	}
 	return false
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"fmt"
+	"net"
 	"testing"
 	"time"
 
@@ -206,6 +207,27 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
 	assert.True(t, metav1.IsControlledBy(pod, readyGs))
 
 	err = podClient.Delete(pod.ObjectMeta.Name, nil)
+	assert.NoError(t, err)
+
+	_, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateUnhealthy, 3*time.Minute)
+	assert.NoError(t, err)
+}
+
+func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
+	t.Parallel()
+	gs := defaultGameServer()
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(defaultNs, gs)
+	if err != nil {
+		t.Fatalf("Could not get a GameServer ready: %v", err)
+	}
+
+	gsClient := framework.AgonesClient.AgonesV1().GameServers(defaultNs)
+	defer gsClient.Delete(readyGs.ObjectMeta.Name, nil) // nolint: errcheck
+
+	conn, err := net.Dial("udp", fmt.Sprintf("%s:%d", readyGs.Status.Address, readyGs.Status.Ports[0].Port))
+	assert.NoError(t, err)
+	defer conn.Close() // nolint: errcheck
+	_, err = conn.Write([]byte("CRASH"))
 	assert.NoError(t, err)
 
 	_, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateUnhealthy, 3*time.Minute)
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -215,20 +215,27 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
 
 func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
 	t.Parallel()
+
+	l := logrus.WithField("test", "TestGameServerUnhealthyAfterReadyCrash")
+
 	gs := defaultGameServer()
 	readyGs, err := framework.CreateGameServerAndWaitUntilReady(defaultNs, gs)
 	if err != nil {
 		t.Fatalf("Could not get a GameServer ready: %v", err)
 	}
 
+	l.WithField("gs", readyGs.ObjectMeta.Name).Info("GameServer created")
+
 	gsClient := framework.AgonesClient.AgonesV1().GameServers(defaultNs)
 	defer gsClient.Delete(readyGs.ObjectMeta.Name, nil) // nolint: errcheck
 
-	conn, err := net.Dial("udp", fmt.Sprintf("%s:%d", readyGs.Status.Address, readyGs.Status.Ports[0].Port))
+	address := fmt.Sprintf("%s:%d", readyGs.Status.Address, readyGs.Status.Ports[0].Port)
+	conn, err := net.Dial("udp", address)
 	assert.NoError(t, err)
 	defer conn.Close() // nolint: errcheck
 	_, err = conn.Write([]byte("CRASH"))
 	assert.NoError(t, err)
+	l.WithField("address", address).Info("sent UDP packet")
 
 	_, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateUnhealthy, 3*time.Minute)
 	assert.NoError(t, err)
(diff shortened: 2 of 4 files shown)


test/e2e/gameserver_test.go
```diff
@@ -211,6 +211,27 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Could not get a GameServer ready: %v", err)
+	}
+
+	gsClient := framework.AgonesClient.AgonesV1().GameServers(defaultNs)
+	defer gsClient.Delete(readyGs.ObjectMeta.Name, nil) // nolint: errcheck
+
+	// there is no response, so don't wait for the timeout
+	go func() {
+		// ignore the error, we don't care about the response
```
roberthbailey: Is there any chance that this call fails? If so this test will become flaky. If you look at the response and retry on errors (until the server acks the crash) then the test will be more reliable.

markmandel: Admittedly it's a bit hacky. What I'm doing is ignoring the error that will come back when we don't get a response, because I didn't feel like writing a separate method to just send, and not wait for an ACK response.

While it's super unlikely there will fail (we use this all through e2e with no retries), I will admit to cutting a couple of corners here :confused: 

I'll update the test. I can't look at myself in the mirror anymore out of embarrassment :smile:

markmandel: Of course, now I've updated it, it won't pass... how very strange.

markmandel:
I feel like this might be a race condition of some kind. Maybe the health.go file is not receiving the pod shutdown event / the event is being merged into the restart event - which may be possible.

Attaching zips of the udp-simple logs (json & csv format) - can see the container receiving the CRASH message, so that is definitely happening.
[container_cluster_name_e2e-test-cluster_namespace_id_default__logs__2019-09-05T03-50.zip](https://github.com/googleforgames/agones/files/3581643/container_cluster_name_e2e-test-cluster_namespace_id_default__logs__2019-09-05T03-50.zip)

Then we see the container restart, and the binary attempt to re-register as Ready. Hard to see on the e2e cluster if it actually gets to ready or not.

For this test, we could look at the `status.container.lastState` as well as the current state. I'm not sure how that will help solve #956 as that could cause Unhealthy statuses to occur when we don't want them to. But maybe one step at a time.

I'll go hunt down the relevant docs in the game server controller as well.

At least this will help lock down the behaviour when things go wrong in GameServers.

markmandel:
I can reproduce this locally, if I run all of `gameserver_test.go`. 

This is the logs from the health controller for the gameserver in question - as you can see, there is no moving to Unhealthy.

```
{"gsKey":"default/udp-server8qh7b","message":"Enqueuing","queue":"agones.dev.HealthController","severity":"info","source":"*gameservers.HealthController","time":"2019-09-06T00:17:15.904024795Z"}
{"gsKey":"default/udp-server8qh7b","message":"Processing","queue":"agones.dev.HealthController","severity":"info","source":"*gameservers.HealthController","time":"2019-09-06T00:17:15.909286119Z"}
{"gsKey":"default/udp-server8qh7b","message":"Synchronising","severity":"info","source":"*gameservers.HealthController","time":"2019-09-06T00:17:15.909381042Z"}
{"gsKey":"default/udp-server8qh7b","message":"Enqueuing","queue":"agones.dev.HealthController","severity":"info","source":"*gameservers.HealthController","time":"2019-09-06T00:17:17.719181804Z"}
{"gsKey":"default/udp-server8qh7b","message":"Enqueuing","queue":"agones.dev.HealthController","severity":"info","source":"*gameservers.HealthController","time":"2019-09-06T00:17:17.72305824Z"}
{"gsKey":"default/udp-server8qh7b","message":"Processing","queue":"agones.dev.HealthController","severity":"info","source":"*gameservers.HealthController","time":"2019-09-06T00:17:17.724500147Z"}
{"gsKey":"default/udp-server8qh7b","message":"Synchronising","severity":"info","source":"*gameservers.HealthController","time":"2019-09-06T00:17:17.724567851Z"}
```

The gameserver ends up in Ready state, and never moves. So yeah, race condition of some kind!

markmandel:
Okay, test this new solution multiple times - can't get it to fail anymore.

test/e2e/gameserver_test.go
```diff
@@ -211,6 +213,34 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
 
+func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
+	t.Parallel()
+
+	l := logrus.WithField("test", "TestGameServerUnhealthyAfterReadyCrash")
+
+	gs := defaultGameServer()
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(defaultNs, gs)
+	if err != nil {
+		t.Fatalf("Could not get a GameServer ready: %v", err)
```
roberthbailey: any particular reason this one uses `t.Fatalf(...)` while below you use `assert.NoError(...)`

pkg/gameservers/health.go
```diff
@@ -125,8 +125,11 @@ func (hc *HealthController) unschedulableWithNoFreePorts(pod *corev1.Pod) bool {
 func (hc *HealthController) failedContainer(pod *corev1.Pod) bool {
 	container := pod.Annotations[agonesv1.GameServerContainerAnnotation]
 	for _, cs := range pod.Status.ContainerStatuses {
-		if cs.Name == container && cs.State.Terminated != nil {
-			return true
+		if cs.Name == container {
+			// sometimes on a restart, these cs.State can be running and the last state will be merged
+			if cs.State.Terminated != nil || cs.LastTerminationState.Terminated != nil {
```
roberthbailey: nit: can probably be `return cs.State.Terminated != nil || cs.LastTerminationState.Terminated != nil` to break the loop once the desired container is found.

pkg/gameservers/health.go
```diff
@@ -125,8 +125,11 @@ func (hc *HealthController) unschedulableWithNoFreePorts(pod *corev1.Pod) bool {
 func (hc *HealthController) failedContainer(pod *corev1.Pod) bool {
 	container := pod.Annotations[agonesv1.GameServerContainerAnnotation]
 	for _, cs := range pod.Status.ContainerStatuses {
-		if cs.Name == container && cs.State.Terminated != nil {
-			return true
+		if cs.Name == container {
+			// sometimes on a restart, these cs.State can be running and the last state will be merged
```
roberthbailey: nit: s/these/the/

roberthbailey: question - was this a bug before where there was a race between checking the health and the container restarting?
question2 - it seems like this will always return true if the container has *ever* failed, which is a change from before (where it was transient but could flip back to false after a restart). Is that the correct behavior?

test/e2e/gameserver_test.go
markmandel: Mostly - because copy paste :smile: it's what we have everywhere..

Lower down, it would be super rare to fail (how could you not write to a buffer that doesn't track if it gets received or not?)

pkg/gameservers/health.go
markmandel: Yes! Change made.

pkg/gameservers/health.go
markmandel: These are good questions:

question1 - So the bug was a couple of things:
1) The issue outlined in #956 , but also, that could potentially not happen (i.e. a race condition could mean on restart, it would remain in a Ready state) -- so it was inconsistent. At least now, it consistently moves to an unhealthy state on crash.

2) This is essentially true. If the pod has ever failed (in the last 2 iterations, because that is all that is tracked), and the pod gets updated (say the sidecar crashes), we will move to unhealthy - but at least consistently so.

The next PR I'm going to do, is actually solve #956. My current thinking is - when a GameServer goes to Ready, we put an annotation on the pod with the current `Running` gameserver container ID. Then we always know what not to look at before - i.e. if there was a status.State of Running, and it matched our annotation, we would know not to look at the LastTerminatedState, since that came before.

If that annotation is not there, we don't ever move to Unhealthy, and allow restarts to occur as per normal K8s.

There might be some other edge cases in there -- but I think that gets us to what is the currently documented health checking strategy.

TL;DR - this is an interim step.

markmandel: I just realised that 1.0 RC comes out Tuesday. Happy for this to be merged post 1.0 stable release. It means we have #956 outstanding for 1.0, but I don't think it's a hard blocker for 1.0.

I wouldn't want this to be included in 1.0 without the follow up PR to fix #956 - I've started work on it, but probably safer to wait.

roberthbailey:
When I rebased to head, this PR lost the changes moving the simple udp container version from 0.14 to 0.15 since that version bump was done as part of https://github.com/googleforgames/agones/pull/1059. So we need to move from 0.15 to 0.16 in this PR now.

/hold

markmandel:
Did a local rebase - I think this should be good to go now @roberthbailey

roberthbailey:
Do we not need to bump the simple-udp container version any longer?

markmandel:
There are no changes to the simple-udp implementation, so I'm not sure why we need to bump it?

markmandel:
Oh yeah, I wrote in the description for this PR that there was a change in udp-simple. But i don't see any. Weird. I forget when happened? Whomever merges/squashes this, please feel free to remove that line from the commit.

roberthbailey:
Is there a way to modify commit messages when merging a PR?

markmandel:
Yeah, when you click "merge" it squashes it and gives you the option to edit the commit message on GitHub.

roberthbailey:
Neat. I guess I'd always clicked through that part quickly without reading it before.

PR #1051, opened 2019-09-10
Update golangci-lint to 1.18, add bodyclose check

aLekSer:
Fix some tests which leak connections (`bodyclose` linter).

This version of `golangci-lint` would be needed for upgrading to Golang 1.13.

Fixed also:
```
pkg/util/apiserver/apiserver.go:128:22: appendAssign: append result not assigned to the same slice (gocritic)                                                                                
        list.APIResources = append(as.resourceList[groupVersion].APIResources, resource)
```

### DIFF
--- a/test/e2e/allocator_test.go
+++ b/test/e2e/allocator_test.go
@@ -73,7 +73,6 @@ func TestAllocator(t *testing.T) {
 	}
 
 	// wait for the allocation system to come online
-	var response *http.Response
 	err = wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
 		// create the rest client each time, as we may end up looking at an old cert
 		var client *http.Client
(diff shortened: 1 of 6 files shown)


test/e2e/allocator_test.go
```diff
@@ -82,7 +82,7 @@ func TestAllocator(t *testing.T) {
 			return true, err
 		}
 
-		response, err = client.Post(requestURL, "application/json", bytes.NewBuffer(body))
+		response, err = client.Post(requestURL, "application/json", bytes.NewBuffer(body)) // nolint: bodyclose
```
roberthbailey: it seems like this one should be fixed to actually close the body.

pkg/util/apiserver/apiserver.go
```diff
@@ -108,13 +108,12 @@ func NewAPIServer(mux *http.ServeMux) *APIServer {
 // e.g. http://localhost:8001/apis/scheduling.k8s.io/v1beta1
 // as well as registering a CRDHandler that all http requests for the given APIResource are routed to
 func (as *APIServer) AddAPIResource(groupVersion string, resource metav1.APIResource, handler CRDHandler) {
-	list, ok := as.resourceList[groupVersion]
+	_, ok := as.resourceList[groupVersion]
```
roberthbailey: I'm not sure I understand the changes in this file. Should this be included? It doesn't have anything to do with fixing the lint errors.

roberthbailey:
Now that 1.0 has been cut, can you address comments so that we can get this merged? Thanks!

pkg/util/apiserver/apiserver.go
aLekSer: Removing this change from PR.

aLekSer:
There was a need to rewrite `test/e2e/allocator_test.go`. Added defer and moved code reading from body inside of `wait.PollImmediate`

Issue #980, opened 2019-08-05
Allocation policy needs to expose namespace of the targeted cluster


pooneh-m:
[Allocation policy CRD](https://github.com/googleforgames/agones/blob/master/pkg/apis/multicluster/v1alpha1/gameserverallocationpolicy.go) is currently only exposing the LoadBalancer service endpoint for allocator service and not the namespace. This works fine if there is an allocator service for each namespace that game servers are deployed to https://github.com/googleforgames/agones/issues/854.

However, currently, there is only one allocator service deployed to agones-system namespace and it is assumed that all clusters in multi-cluster scenario are using the same namespace name for deploying game servers.

The solution is to expose namespace in allocation policy.
[Bonus: This also enables A/B testing on the same cluster by deploying two version of game servers and setting weight on allocation policy to redirect a percentage of traffic to a namespace in the cluster.]

roberthbailey:
Since this is still a v1alpha1 API group (indicating active development) I'm OK changing this between 0.12 and 1.0.

markmandel:
Can this now be closed?

https://help.github.com/en/articles/closing-issues-using-keywords might be useful going forward as well.

pooneh-m:
Thanks. Didn't know about the keyword. Next time :)

PR #1052, opened 2019-09-10
Adding namespace for multi-cluster allocation policy

pooneh-m:
Addressing: https://github.com/googleforgames/agones/issues/980


pkg/apis/multicluster/v1alpha1/gameserverallocationpolicy.go
```diff
@@ -32,9 +32,14 @@ type GameServerAllocationPolicySpec struct {
 
 // ClusterConnectionInfo defines the connection information for a cluster
 type ClusterConnectionInfo struct {
-	ClusterName         string   `json:"clusterName"`
+	// Optional: the name of the targeted cluster
+	ClusterName string `json:"clusterName"`
+	// The endpoint for the allocator service in the targeted cluster
```
roberthbailey: nit: the comment should be "endpoints" since this is an array. 

As a side note, what is the use case for having more than one endpoint here? Since this is connection info for a single cluster, do we expect to have multiple ways to connect to the allocation endpoint in the cluster? It would be worth updating the comment to explain what we expect in this field and what clients should do if there is more than one entry (and maybe also if it's valid to have an empty array).

pkg/apis/multicluster/v1alpha1/gameserverallocationpolicy.go
```diff
@@ -32,9 +32,14 @@ type GameServerAllocationPolicySpec struct {
...
 	AllocationEndpoints []string `json:"allocationEndpoints"`
-	SecretName          string   `json:"secretName"`
+	// The name of the secret that contains TLS client certificates to connect the allocator server in the targeted cluster
+	SecretName string `json:"secretName"`
+	// the cluster namespace to allocate gameserver from
```
roberthbailey: nit: start comment with a capital.

roberthbailey: another nit: I think it's a bit more proper to phrase it as "The cluster namespace from which to allocate gameservers" since ending sentences with 'from' or 'to' leaves the reading thinking you forgot to type the rest of it.

pkg/apis/multicluster/v1alpha1/gameserverallocationpolicy.go
roberthbailey: It looks like from the code / comment below, we currently just walk the list in order (rather than picking a random one each time).

pooneh-m: Thanks. Updated the comment. 
It is not expected to have multiple endpoints as of now, but it is possible in the future and it follows the [kubernetes ingress model](https://github.com/nginxinc/kubernetes-ingress/blob/master/vendor/k8s.io/api/core/v1/types.go#L3481), which stores the LB IPs as a list.

pkg/apis/multicluster/v1alpha1/gameserverallocationpolicy.go
pooneh-m: Done. Thanks for the suggestion.

pkg/apis/multicluster/v1alpha1/gameserverallocationpolicy.go
roberthbailey: Since this api is still alpha I think this is reasonable, but we should check with the networking folks on k8s whether they think this was a good idea or if still exists for backwards compatibility (e.g. they are stuck with it) before we promote the api to being stable.

PR #1056, opened 2019-09-12
Marking Gameservers with Evicted backing Pods as Unhealthy

### LINKED ISSUE #1028: GameServer status does not account for Evicted Pods
**What happened**:
When the pod running your game server gets evicted due to exceeding a hard limit you have set in your pod/gameserver yaml, the gameserver entry remains unchanged and will still have its previous status of Allocated/Ready. This is incorrect as the underlying pod is no longer running and K8s shows it as Evicted.

The game server status should ideally change to a value that represents the pod is no longer running. So could be one of the existing status' of unhealthy or shutdown or potentially a new status of evicted.

Apply this slightly modified GameServer yaml taken from the Quickstart:

```
apiVersion: "agones.dev/v1"
kind: GameServer
metadata:
  name: "eph-test"
spec:
  ports:
  - name: default
    portPolicy: Dynamic
    containerPort: 7654
  template:
    spec:
      containers:
      - name: eph-test
        image: gcr.io/agones-images/udp-server:0.14
        stdin: true 
        tty: true 
        resources:
          limits:
            memory: "32Mi"
            cpu: "20m"
            ephemeral-storage: "50Mi"
```
`kubectl create -f <gs.yaml>`

Then run `kubectl exec -it eph-test sh` to get a shell into the running pod.
Now we simulate exceeding the ephemeral storage limit by creating a file larger than the limit.
Run `fallocate -l 100M /home/server/test` to create a 100MB file. Now wait a minute or two until K8s does its resource check and decides to Evict your pod. The shell should terminate automatically as the pod has been killed. 
Running `kubectl describe gs eph-test` will show your GameServer with status Ready, while running `kubectl describe pod eph-test` will show your pod with status Evicted (and citing exceeding ephemeral storage in the list of events)

I have only tested this with ephemeral storage limits. I dont think its possible to exceed the other limits as you just get throttled by the system.

**Environment**:
- Agones version: 
12
Client v1.14.3
Server v1.12.9-gke.13
GKE
- Install method (yaml/helm):
yaml
- Others:

aLekSer:
Add additional health check to the `Gameservers` Controller.
E2E test added which set Ephemeral Storage limit to 0 MB and wait for Unhealthy state.

Closes #1028 .


aLekSer:
I would add an E2E test for this feature and the mark it as Ready for Review.

PR #1061, opened 2019-09-20
Use non-ephemeral ports for side car HTTP and gRPC.

### LINKED ISSUE #851: Sidecar occasionally fails to start up
**What happened**:
Out of 10 gameservers in a new fleet, 3 never left the 'Scheduled' state. Further investigation revealed that the game server pod successfully started, however the agones sidecar logged an error.

> Truncated Logs:
> {"message":"Starting SDKServer grpc service...","severity":"info"}
> {"message":"Starting SDKServer grpc-gateway...","severity":"info"}
> {"error":"listen tcp 127.0.0.1:59358: bind: address already in use","message":"Could not serve http server","severity":"fatal"}

After deleting the affected gameservers, new gameservers and sidecars do start up without any issue.

If the sidecar of a gameserver fails to initialize, either recycle the gameserver or repair the sidecar.

No consistent repro steps. Seems to be most replicable when rapidly scaling up a new gameserverset, but is not consistent in its frequency.

Fleet Yaml:

```
apiVersion: stable.agones.dev/v1alpha1
kind: Fleet
metadata:
  name: battleserver
spec:
  replicas: 10
  template:
    spec:
      health:
        disabled: true
        initialDelaySeconds: 60
      ports:
      - containerPort: 443
        name: default
        protocol: TCP
      template:
        spec:
          containers:
            image: <REMOVED>
            name: battleserver
```

**Environment**:
- Agones version: 
0.10.0
Client: "v1.14.2"
Server: "v1.13.6-gke.13"
- Others:

drichardson:
Replaced 59357 with 9357 and 59358 with 9358.

Fixes #851.

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -118,7 +118,7 @@ ensure-build-sdk-image:
 run-sdk-conformance-local: TIMEOUT ?= 30
 run-sdk-conformance-local: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch
 run-sdk-conformance-local: ensure-agones-sdk-image
-	docker run -e "ADDRESS=" -p 59357:59357 \
+	docker run -e "ADDRESS=" -p 9357:9357 \
 	 -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" $(sidecar_tag)
 
 # Run SDK conformance test, previously built, for a specific SDK_FOLDER
@@ -130,7 +130,7 @@ run-sdk-conformance-no-build: TESTS ?= ready,allocate,setlabel,setannotation,gam
 run-sdk-conformance-no-build: ensure-agones-sdk-image
 run-sdk-conformance-no-build: ensure-build-sdk-image
 	DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
-	sleep $(DELAY) && docker run -p 59357:59357 -e "ADDRESS=" \
+	sleep $(DELAY) && docker run -p 9357:9357 -e "ADDRESS=" \
 	 -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" --net=host  $(sidecar_tag)
 
 # Run SDK conformance test for a specific SDK_FOLDER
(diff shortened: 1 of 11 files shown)


roberthbailey:
This is a breaking change, because sidecars launched by the next release of Agones will only listen on the lower port while game servers compiled with older SDKs (including those from the recent 1.0 release) will attempt to connect to the higher, ephemeral ports.

roberthbailey:
I've been trying to think of a way to do dynamic port binding in the sdkserver and have the client discover it, but I don't think there is a reasonable way to do it. 

Instead, I think we need to change the sdkserver to listen on both the old and the new ports, and mark the ephemeral ports as deprecated. We can then start a discussion about how long to support them before we drop them entirely. 

For the SDKs, if we change the port they try to connect to in the 1.1 release, then any gameservers compiled with the 1.1 SDKs won't work with agones 1.0 sdkservers. I don't think we've talked much about SDK compatibility with older agones releases (we've mostly said we want to ensure forward compatibility so that you don't have to recompile your game to _upgrade_ agones) but we should consider whether we care about also making it easy to downgrade if your game is compiled with a newer SDK. If we want to be extra careful, we can change the sdk port in 1.2 to ensure +/- 1 version compatibility between the SDKs and the sdkservers, or we can say that this bug is serious enough that we are ok just keeping SDK forward compatibility.

drichardson:
@roberthbailey Listening on both ports makes sense for transition. In that case, I'm going to delete this pull request because I'm not going to do that right now.

roberthbailey:
I'll start working on a PR to listen on both ports.

roberthbailey:
(and thanks for identifying the root cause of the bug!)

drichardson:
> 
> 
> I'll start working on a PR to listen on both ports.

BTW, make sure a bind failure on the ephemeral port does not cause the side car to fail to start up (otherwise, the fix would be pointless ;)

roberthbailey:
Good point. That should be quick and easy to test.

drichardson:
@roberthbailey one other thing. I got a clang-format failure on this but I don't see any information about the failure other than the file it occurred in (sdks/cpp/src/agones/sdk.cc). I've run the build on my own machine and see the same failure, but still don't know what exactly is failure (not even the line number).

roberthbailey:
You can use `clang-format` to automatically format the cpp files (like gofmt on the golang files). See https://github.com/googleforgames/agones/blob/3f9834690f13092d5246e8f426b53f797830b8c0/CONTRIBUTING.md#formatting

PR #1063, opened 2019-09-23
Add cleanup for GameServerSets, add E2E test

aLekSer:
Add cleanup for GameServerSets, sometimes them are not deleted with Fleets. Also add E2E test which creates GameServerSet without a Fleet.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -270,6 +270,11 @@ func (f *Framework) CleanUp(ns string) error {
 		return err
 	}
 
+	err = agonesv1.GameServerSets(ns).DeleteCollection(deleteOptions, listOptions)
+	if err != nil {
+		return err
+	}
+
 	return agonesv1.GameServers(ns).
 		DeleteCollection(deleteOptions, listOptions)
 }
--- a/test/e2e/fleet_test.go
+++ b/test/e2e/fleet_test.go
@@ -562,6 +562,7 @@ func TestFleetGSSpecValidation(t *testing.T) {
 
 // TestFleetNameValidation is built to test Fleet Name length validation,
 // Fleet Name should have at most 63 chars.
+// nolint:dupl
 func TestFleetNameValidation(t *testing.T) {
 	t.Parallel()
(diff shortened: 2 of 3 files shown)


test/e2e/framework/framework.go
```diff
@@ -270,6 +270,11 @@ func (f *Framework) CleanUp(ns string) error {
 		return err
 	}
 
+	err = agonesv1.GameServerSets(ns).DeleteCollection(deleteOptions, listOptions)
```
markmandel: So I don't know if this is actually a good idea.

GameServerSets should be deleted with Fleets - if they are not, that's a sign of a bigger issue.

So I worry this is going to bandaid over some other kind of issue.

markmandel:
@googlebot take another look.

test/e2e/framework/framework.go
aLekSer: Got the point, I think there could be some timeout issue with cleanup phase during E2E or there could be some other interruption and resource does not released in a timely manner in that case.
The solution could be useful if we met this situation again.
By the way we could also add some atomicity to fleet delete with E2E to cover this situation when resource was not released if it possible to achieve.

markmandel: > By the way we could also add some atomicity to fleet delete with E2E to cover this situation when resource was not released if it possible to achieve.

Do a foreground delete rather than a background one? Wonder if it's worth the time to wait?

This still feels like a thing that's preventing the symptoms rather than letting us find a cure. So I don't think I'm in favour of this PR. I'd rather see it fail, and then let us try and diagnose what the condition was.

aLekSer: Agree, should I close this PR?

markmandel: Yeah I think so :disappointed: 

As another PR, maybe if we had better logging/diagnostic information in the cleanup events, so we could see causes a little better? I don't think we ever expected cleanup to fail.

aLekSer:
Close this PR according to inline comments. It is better to have E2E tests failing to diagnose the root cause, than mitigate the results.

PR #1066, opened 2019-09-24
Changed remote cluster to validate PEM certificate instead of DER

pooneh-m:
Currently allocator service is expecting TLS certificate of [PEM format](https://github.com/googleforgames/agones/blob/master/cmd/allocator/main.go#L117) and the client is expecting server TLS signing certificate of [DER format](https://github.com/googleforgames/agones/blob/master/pkg/gameserverallocations/allocator.go#L358). This discrepancy was not detected because e2e tests are [skipping cert validation](https://github.com/googleforgames/agones/blob/master/test/e2e/allocator_test.go#L134) and we don't have a good test coverage for multi-cluster allocation scenarios. 

This only impact the multi-cluster allocation feature which is in v1alpha1. The issue was detected while writing e2e tests for multi-cluster allocation.


markmandel:
So this looks fine, but would love some more info in the commit about why this change is in place / what it actually does.

I've no idea the difference between PEM /  DER is for SSL certs?

roberthbailey:
It would also be nice to mention (confirm?) that this code is only in the path of the v1alpha1 APIs for allocation, since otherwise this would be considered a breaking change.

pooneh-m:
I updated the description with more info. The change is impacting v1alpha1 feature. PEM and DER are different formats for certificates.

PR #1068, opened 2019-09-24
Fix Access Api Example

ramonberrutti:
In the example, it is missing the logger definition.


ramonberrutti:
@googlebot I signed it!

PR #1069, opened 2019-09-25
GameServer container restart before Ready, move to Unhealthy state After

### LINKED ISSUE #956: Game server container crash before Ready, should restart, not move to Unhealthy

markmandel:
This brings our implementation inline with what our health checking documentation states that we do.

To solve this, we store the currently running GameServer containerID (which is unique to that running instance), as an annotation on the GameServer Pod.

When the annotation is not there, we know the Pod is not yet Ready, so we can ignore it when our Unhealthy check occurs, and restarts of the GameServer container can happen as per usual.

When the annotation is there, we know to check for failure, but to avoid ContainerStatus.LastTerminationState polluting the result since that crash/failure may have happened before the GameServer is ready, we can compare the current ContainerID to the one stored in the
annotation -- which when equal means we can skip looking at the LastTerminationState, as it was before the GameServer was marked as Ready.

Lots of unit and e2e test updates to go with this to test it out.

Closes #956

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -761,6 +761,27 @@ func (c *Controller) syncGameServerRequestReadyState(gs *agonesv1.GameServer) (*
 	c.loggerForGameServer(gs).Info("Syncing RequestReady State")
 
 	gsCopy := gs.DeepCopy()
+	pod, err := c.gameServerPod(gs)
+	// The most likely reason this fails is if the master has gone down, as it's very unlikely
+	// the Pod got deleted between sending `RequestReady` and now,
+	// so go into backoff until whatever the error is resolves itself.
+	if err != nil {
+		return nil, err
+	}
+
+	// track the ready gameserver container, so we can determine that after this point, we should move to Unhealthy if
+	// there is a container crash/restart after we move to Ready
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == gs.Spec.Container {
+			if _, ok := pod.ObjectMeta.Annotations[agonesv1.GameServerReadyContainerIDAnnotation]; !ok {
+				pod.ObjectMeta.Annotations[agonesv1.GameServerReadyContainerIDAnnotation] = cs.ContainerID
+				if _, err = c.podGetter.Pods(gs.ObjectMeta.Namespace).Update(pod); err != nil {
+					return nil, errors.Wrapf(err, "error adding ready containerid to pod %s", pod.ObjectMeta.Name)
+				}
+			}
+			break
+		}
+	}
 
 	// if the address hasn't been populated, and the Ready request comes
 	// before the controller has had a chance to do it, then
@@ -768,12 +789,6 @@ func (c *Controller) syncGameServerRequestReadyState(gs *agonesv1.GameServer) (*
 	addressPopulated := false
 	if gs.Status.NodeName == "" {
 		addressPopulated = true
-		pod, err := c.gameServerPod(gs)
-		// NotFound should never happen, and if it does -- something bad happened,
-		// so go into workerqueue backoff.
-		if err != nil {
-			return nil, err
-		}
 		gsCopy, err = c.applyGameServerAddressAndPort(gsCopy, pod)
 		if err != nil {
 			return gs, err
@@ -781,7 +796,7 @@ func (c *Controller) syncGameServerRequestReadyState(gs *agonesv1.GameServer) (*
 	}
 
 	gsCopy.Status.State = agonesv1.GameServerStateReady
-	gs, err := c.gameServerGetter.GameServers(gs.ObjectMeta.Namespace).Update(gsCopy)
+	gs, err = c.gameServerGetter.GameServers(gs.ObjectMeta.Namespace).Update(gsCopy)
 	if err != nil {
 		return gs, errors.Wrapf(err, "error setting Ready, Port and address on GameServer %s Status", gs.ObjectMeta.Name)
 	}
--- a/pkg/gameservers/health.go
+++ b/pkg/gameservers/health.go
@@ -104,7 +104,7 @@ func NewHealthController(health healthcheck.Handler,
 // isUnhealthy returns if the Pod event is going
 // to cause the GameServer to become Unhealthy
 func (hc *HealthController) isUnhealthy(pod *corev1.Pod) bool {
-	return hc.evictedPod(pod) || hc.unschedulableWithNoFreePorts(pod) || hc.failedContainer(pod)
+	return hc.evictedPod(pod) || hc.unschedulableWithNoFreePorts(pod) || hc.failedContainerAfterReady(pod)
 }
 
 // unschedulableWithNoFreePorts checks if the reason the Pod couldn't be scheduled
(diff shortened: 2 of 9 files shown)


test/e2e/gameserver_test.go
```diff
@@ -213,6 +213,96 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
+	defer conn.Close() // nolint: errcheck
+
+	logger.Info("crashing, and waiting to see restart")
+	err = wait.Poll(time.Second, 3*time.Minute, func() (bool, error) {
+		// not sure when the pod process will actually start, so lets send a crash every time, just in case
+		// process may not be there, so we don't care about the error
+		conn.Write([]byte("CRASH")) // nolint: errcheck
+		pod, err := framework.KubeClient.CoreV1().Pods(defaultNs).Get(newGs.ObjectMeta.Name, metav1.GetOptions{})
+		if err != nil {
+			return true, err
```
roberthbailey: should this return false to retry the loop?

pkg/gameservers/controller.go
```diff
@@ -710,27 +710,39 @@ func (c *Controller) syncGameServerRequestReadyState(gs *agonesv1.GameServer) (*
 	c.loggerForGameServer(gs).Info("Syncing RequestReady State")
 
 	gsCopy := gs.DeepCopy()
+	pod, err := c.gameServerPod(gs)
+	// NotFound should never happen, and if it does -- something bad happened,
```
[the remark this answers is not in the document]
markmandel: "never happen" is a little of an exaggeration -- if we've got a ReadyRequest, then the pod has spun up, and sent the request.

For this to fail, a Pod would need to spin up, send the `RequestReady` command, and then be deleted before the controller noticed.

So "Never" is not actually true, but it should be remarkably rare.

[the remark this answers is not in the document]
markmandel: I updated the comment - you'll see it on the next push

pkg/gameservers/controller.go
```diff
@@ -710,27 +710,39 @@ func (c *Controller) syncGameServerRequestReadyState(gs *agonesv1.GameServer) (*
...
+	}
+
+	// track the ready gameserver container, so we can determine that after this point, we should move to Unhealthy if
+	// there is a container crash/restart after we move to Ready
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == gs.Spec.Container {
+			pod.ObjectMeta.Annotations[agonesv1.GameServerReadyContainerIDAnnotation] = cs.ContainerID
+		}
+	}
+	_, err = c.podGetter.Pods(gs.ObjectMeta.Namespace).Update(pod)
```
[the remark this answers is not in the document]
markmandel: Yeah - that's an excellent point. I'll add a check that if it's there, we don't need to do it again.

test/e2e/gameserver_test.go
markmandel: So we created the GameServer previously, and waited for it to be `Scheduled`. If the Pod isn't there, then something has gone horribly wrong, since `Scheduled` state indicates that a Pod exists in some form or another.

markmandel:
Weird this is failing. It's working on my machine... Will look into it tomorrow.

pkg/gameservers/health.go
```diff
@@ -126,12 +126,26 @@ func (hc *HealthController) evictedPod(pod *corev1.Pod) bool {
...
+	if !ok {
+		return false
+	}
+
 	container := pod.Annotations[agonesv1.GameServerContainerAnnotation]
 	for _, cs := range pod.Status.ContainerStatuses {
 		if cs.Name == container {
+			// if the current container is the Ready container instance, and is running, then the LastTerminationState
+			// was from before the GameServer was ready, so we can ignore it.
+			if cs.State.Terminated == nil && cs.ContainerID == containerID && cs.LastTerminationState.Terminated != nil {
```
roberthbailey: Maybe a silly question, but the comment mentions that we can ignore the last termination state but it's still part of the conditional check. If `cs.State.Terminated == nil && cs.ContainerID == containerID` returns true and `cs.LastTerminationState.Terminated != nil` returns false (because this is the first instantiation of the container), then we fail this check. I think that we still correctly return false in the following check, but why not just do it here? Is there an edge cases that this is checking? If so, let's make sure there is a test for it.

test/e2e/gameserver_test.go
roberthbailey: Got it. So the idea is to break the loop and return an error to let the error bubble up.

pkg/gameservers/controller.go
```diff
@@ -761,27 +761,42 @@ func (c *Controller) syncGameServerRequestReadyState(gs *agonesv1.GameServer) (*
...
+	}
+
+	// track the ready gameserver container, so we can determine that after this point, we should move to Unhealthy if
+	// there is a container crash/restart after we move to Ready
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == gs.Spec.Container {
+			if _, ok := pod.ObjectMeta.Annotations[agonesv1.GameServerReadyContainerIDAnnotation]; !ok {
+				pod.ObjectMeta.Annotations[agonesv1.GameServerReadyContainerIDAnnotation] = cs.ContainerID
+				if _, err = c.podGetter.Pods(gs.ObjectMeta.Namespace).Update(pod); err != nil {
+					return nil, errors.Wrapf(err, "error adding ready containerid to pod %s", pod.ObjectMeta.Name)
```
roberthbailey: qq: Why apply the annotation to the pod instead of to the gameserver? If we applied it to the game server then you could combine this with the update call below, which would make the state transition to ready (and this container identification) happen in a single API call. Right now there is a slight chance that this succeeds and the following call fails (or never happens due to a crash) and then we've marked the container id but never moved the game server to ready.

markmandel: That's a really good question - so in [health.go](https://github.com/googleforgames/agones/pull/1069/files#diff-c3108a3059e5decdd651ad0d0b47479eR107) we've only ever looked up the Pod.

In _theory_ we could also lookup the GameServer and check it's status. Maybe that's actually a better option. Since it's likely in the Informer cache anyway. There are only a few states (maybe just one? Scheduled) that come before being Ready, so there aren't a lot of edge cases to check -- and then we don't need to even check containerids.

The more I think about it, this seems like it might be a better approach. How does that sound to you?

markmandel: Digging into this, this would need to change the strategy on the controller - as there could be an error getting the GameServer, so we may have to push this into a queue (whereas we don't have to do it right now)

Context:
https://github.com/googleforgames/agones/blob/2c8cc9bb18bce4b8bafa3d61246b0980e058c9e6/pkg/gameservers/health.go#L87

So there are pros and cons both ways.

markmandel:
Let's hold off on continuing to review this - I believe I have a much simpler approach.

You can see the [work in progress here](https://github.com/googleforgames/agones/compare/master...markmandel:bug/restart-before-ready-v2?expand=1)

markmandel:
Closing this, as #1099 is a much better approach.

PR #1071, opened 2019-09-25
Add a simple-tcp game server to use for testing.

roberthbailey:


roberthbailey:
> can we also add a link here: https://agones.dev/site/docs/examples/

Done.

PR #1072, opened 2019-09-25
Bind the sdkserver to non-ephemeral ports

roberthbailey:
This is in addition to the (now legacy) ephemeral ports.

### DIFF
--- a/cmd/sdk-server/main.go
+++ b/cmd/sdk-server/main.go
@@ -41,8 +41,12 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


markmandel:
Just as a note - waiting on a final decision on #851 before reviewing this.

roberthbailey:
/close

I'm going to abandon this and implement this feature as [described here](https://github.com/googleforgames/agones/issues/851#issuecomment-535260745).

PR #1073, opened 2019-09-25
Update of Docsy and Hugo to latest versions.

markmandel:
Mainly done to get access to some new [shortcodes](https://www.docsy.dev/docs/adding-content/shortcodes/), especially the `alert` and `swaggerui` shortcodes.

Also, lots of bug fixes that I had in the queue for both projects are now integrated, so I can remove the workarounds.


roberthbailey:
So this PR has one commit... and 10k files changes. I assume a bunch of the changes were automatically generated but from the PR description it sounds like you did a few things by hand as well. It would be really useful to separate those things into separate commits so that we can easily review the changes made by hand (and easily reproduce the automatic changes if necessary). As it is, I'm not really sure how to properly review this.

markmandel:
_nods_ makes perfect sense, I'll get on it!

site/content/en/docs/Advanced/_index.md
```diff
@@ -4,5 +4,3 @@ date: 2019-01-03T05:44:55Z
 description: "Advanced Guides, Techniques and walk-throughs"
 weight: 300
 ---
-
```
markmandel: Just to be clear - this is removed, as that pagelist is provided automatically in the latest version of Hugo + Docsy, so it's no longer needed.

markmandel:
Gentle bump on re-review. The split in commits was done.

PR #1078, opened 2019-09-26
Sdkserver port configuration

roberthbailey:
First step of fixing https://github.com/googleforgames/agones/issues/851. 

I'm marking this as draft as I still need to do some testing.

### DIFF
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -70,6 +70,13 @@ const (
 	// This will mean that users will need to lookup what port has been opened through the server side SDK.
 	Passthrough PortPolicy = "Passthrough"
 
+	// SdkServerLogLevelInfo will cause the SDK server to output all messages except for debug messages.
+	SdkServerLogLevelInfo SdkServerLogLevel = "Info"
+	// SdkServerLogLevelDebug will cause the SDK server to output all messages including debug messages.
+	SdkServerLogLevelDebug SdkServerLogLevel = "Debug"
+	// SdkServerLogLevelError will cause the SDK server to only output error messages.
+	SdkServerLogLevelError SdkServerLogLevel = "Error"
+
 	// RoleLabel is the label in which the Agones role is specified.
 	// Pods from a GameServer will have the value "gameserver"
 	RoleLabel = agones.GroupName + "/role"
@@ -135,16 +142,10 @@ type GameServerSpec struct {
 	Health Health `json:"health,omitempty"`
 	// Scheduling strategy. Defaults to "Packed"
 	Scheduling apis.SchedulingStrategy `json:"scheduling,omitempty"`
-	// Logging specifies log levels for Agones system containers
-	Logging Logging `json:"logging,omitempty"`
+	// SdkServer specifies parameters for the Agones SDK Server sidecar container
+	SdkServer SdkServer `json:"sdkServer,omitempty"`
 	// Template describes the Pod that will be created for the GameServer
 	Template corev1.PodTemplateSpec `json:"template"`
-}
-
-// Logging specifies log levels for Agones system containers
-type Logging struct {
-	// SdkServer is a log level for SDK server (sidecar) logs. Defaults to "Info"
-	SdkServer string `json:"sdkServer,omitempty"`
 }
 
 // GameServerState is the state for the GameServer
@@ -184,6 +185,19 @@ type GameServerPort struct {
 	Protocol corev1.Protocol `json:"protocol,omitempty"`
 }
 
+// SdkServerLogLevel is the log level for SDK server (sidecar) logs
+type SdkServerLogLevel string
+
+// SdkServer specifies parameters for the Agones SDK Server sidecar container
+type SdkServer struct {
+	// LogLevel for SDK server (sidecar) logs. Defaults to "Info"
+	LogLevel SdkServerLogLevel `json:"logLevel,omitempty"`
+	// GRPCPort is the port on which the SDK Server binds the gRPC server to accept incoming connections
+	GRPCPort int32 `json:"grpcPort,omitempty"`
+	// HTTPPort is the port on which the SDK Server binds the HTTP gRPC gateway server to accept incoming connections
+	HTTPPort int32 `json:"httpPort,omitempty"`
+}
+
 // GameServerStatus is the status for a GameServer resource
 type GameServerStatus struct {
 	// GameServerState is the current state of a GameServer, e.g. Creating, Starting, Ready, etc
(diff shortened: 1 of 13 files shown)


roberthbailey:
Ok, I've done some testing and this is now ready for review. Note that this only makes the ports configurable in the sdkserver, so currently you need to recompile the gameserver with a new SDK with a different port number to actually take advantage of this feature. I'll add SDK support in separate PRs (one per SDK to make them easier to review in parallel). 

Testing cases:

1. Not setting port, existing SDK: works as before
2. Setting port, existing SDK: SDK fails to connect (expected) and gs in unhealthy
3. Setting port, recompiled SDK with new port: works as expected
4. Not setting port, recompiled SDK with new port: SDK fails to connect (expected) and gs in unhealthy

pkg/apis/agones/v1/gameserver.go
```diff
@@ -221,13 +225,13 @@ func (gss *GameServerSpec) ApplyDefaults() {
...
 	gss.applySchedulingDefaults()
-	gss.applyLogLevelDefaults()
+	gss.applySdkServerDefaults()
 }
 
-// applyLogLevelDefaults applies the log level default - "Info"
-func (gss *GameServerSpec) applyLogLevelDefaults() {
-	if gss.Logging.SdkServer == "" {
-		gss.Logging.SdkServer = "Info"
+// applySdkServerDefaults applies the default log level ("Info") for the sidecar
```
roberthbailey: So @aLekSer added this function in https://github.com/googleforgames/agones/pull/1007 and I modified it, but I'm wondering if we should just remove it. The only reason to explicitly set this field is if we are worried about compatibility and implicit values. There is code in `pkg/sdkserver/sdkserver.go` that defaults the log level to info so doing it here feels redundant and it also means that when you describe a game server you **always** see this field, even if you didn't explicitly set it. 

I think it would be better to just leave the struct empty unless it's explicitly set by a user. WDYT?

aLekSer: Yes, I added some redundancy at the early development stage and forgot to remove.
This could be removed now since SDK server is setting default INFO loglevel.

markmandel: So I would actually argue to keep it - basically because when you do a `kubectl get gameserver foo -o yaml` or a `kubectl describe gs foo` you can see what the default value is set to in those commands.

Same as what happens with Pods - you get a bunch more stuff back than what you originally set, because it's the default that has been set by the system.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -135,16 +135,20 @@ type GameServerSpec struct {
...
 	Template corev1.PodTemplateSpec `json:"template"`
 }
 
-// Logging specifies log levels for Agones system containers
-type Logging struct {
-	// SdkServer is a log level for SDK server (sidecar) logs. Defaults to "Info"
-	SdkServer string `json:"sdkServer,omitempty"`
+// SdkServer specifies parameters for the Agones SDK Server sidecar container
+type SdkServer struct {
+	// Logging is a log level for SDK server (sidecar) logs. Defaults to "Info"
```
markmandel:
```suggestion
	// Logging is a log level for SDK server (sidecar) logs. Defaults to "Info". Can also be "Error" or "Warning"
```

pkg/apis/agones/v1/gameserver.go
```diff
@@ -135,16 +135,20 @@ type GameServerSpec struct {
...
+	Logging string `json:"logging,omitempty"`
```
markmandel: Probably not for this PR, but should this be its own type, with predefined values as variables - LoggingError, LoggingWarning, and LoggingInfo ?

site/content/en/docs/Reference/fleet.md
```diff
@@ -58,9 +58,11 @@ spec:
         initialDelaySeconds: 30
         periodSeconds: 60
 {{% feature publishVersion="1.1.0" %}}
```
markmandel: This is going to make an ugly gap in the final rendering - have a look at #1081 - if you review and merge that first, you can take advantage of the split I've already got in there.

site/content/en/docs/Reference/gameserver.md
```diff
@@ -50,13 +50,20 @@ spec:
     # Defaults to 3. Minimum value is 1
     failureThreshold: 3
 {{% feature publishVersion="1.1.0" %}}
-  # logging parameters for game server sidecar
-  logging:
+  # Parameters for game server sidecar
```
markmandel: Same gap comment as above. (because of the feature shortcode)

roberthbailey:
> Just to confirm - this doesn't cover the changes to the SDK(s) so it can connect to the specified ports. I assume that is coming in later PRs? (No problem if it is, just wanted to be sure)

Yes, that's what I was trying to explain [here](https://github.com/googleforgames/agones/pull/1078#issuecomment-535774711):

> Note that this only makes the ports configurable in the sdkserver, so currently you need to recompile the gameserver with a new SDK with a different port number to actually take advantage of this feature. I'll add SDK support in separate PRs (one per SDK to make them easier to review in parallel).

pkg/apis/agones/v1/gameserver.go
roberthbailey: Maybe? Are there examples of other enums being done this way instead of just using a string with documentation?

markmandel: Yeah, lots, e.g.:
https://github.com/googleforgames/agones/blob/master/pkg/apis/agones/v1/gameserver.go#L36-L71

Pattern taken from Kubernetes e.g.:
https://github.com/googleforgames/agones/blob/master//vendor/k8s.io/api/core/v1/types.go#L345-L357 (and lots more in that file)

/cc @aLekSer for visibility. I should have picked this up on initial review, apologies.

pkg/apis/agones/v1/gameserver.go
roberthbailey: In that case, should we also default in the two port values? The downside of defaulting is that it makes it impossible to tell later if the value was set by the user or defaulted by the system. 

For port values, I was planning to do something along the lines of adding a flag/env variable to the game server container that the SDK can find and set the port to connect to. If we don't default the port values, we can just set this if the user requested a specific port. If we default the values, we will need to always set it. The same applies on the sdkserver sidecar container. Right now, in this PR I only set the flags to configure the two ports if a user specified a non-default port in the gameserver config. If we default the port, then the flags will always be specified, because the values will always be non-nil.

site/content/en/docs/Reference/fleet.md
roberthbailey: I'll merge your PR and rebase.

pkg/apis/agones/v1/gameserver.go
roberthbailey: I've gone ahead and done this in the same PR. No need to create another one for this change.

pkg/apis/agones/v1/gameserver.go
roberthbailey: I actually think that with a type defined for this we don't need to enumerate all possible values here (but saying the default is still useful).

pkg/apis/agones/v1/gameserver.go
roberthbailey: As long as I'm mucking with this, would LogLevel make more sense instead of Logging? We picked Logging because it could apply to multiple things, but it really represents the (singular) log level for the sdkserver.

roberthbailey:
I've rebased, picking up the fix for the spacing issues in the website. 

I think there are two outstanding questions:

1. Defaulting for the sdk server parameters: do we default logging, all, or none?
1. Name for the log level / logging parameter: do we change it from Logging to LogLevel?

pkg/apis/agones/v1/gameserver.go
markmandel: Yeah, I agree - I think are right - `LogLevel` is better.

markmandel:
> Defaulting for the sdk server parameters: do we default logging, all, or none?

I would default all. AFAIK, it seems to be the standard Kubernetes action, so that's what we've followed elsewhere. Also makes it easier to introspect what the values actually are from looking at that final `GameServer` configuration.

> Name for the log level / logging parameter: do we change it from Logging to LogLevel?

I agree - LogLevel is better. :+1:

roberthbailey:
This is ready for another review pass.

roberthbailey:
> Looks like you got caught by a linting fail, 

I still haven't figured out how to get GoLand to run gofmt on save.

PR #1079, opened 2019-09-26
Add SDK server HTTP API test

aLekSer:
Add conformance test for Rest API SDK server port 59358.
Does not contain swagger pregenerated files. Switched Go file to other
file extension, because it is not possible to exclude one file from
GolangCI linter run.

For #927

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -61,7 +61,7 @@ gen-sdk-grpc: COMMAND := gen
 gen-sdk-grpc: run-sdk-command
 
 # Runs a command on all supported languages, use COMMAND variable to select which command.
-run-all-sdk-command: run-sdk-command-go run-sdk-command-rust run-sdk-command-cpp run-sdk-command-node
+run-all-sdk-command: run-sdk-command-go run-sdk-command-rust run-sdk-command-cpp run-sdk-command-node run-sdk-command-restapi
 
 run-sdk-command-node:
 	$(MAKE) run-sdk-command COMMAND=$(COMMAND) SDK_FOLDER=node
@@ -74,6 +74,10 @@ run-sdk-command-rust:
 
 run-sdk-command-go:
 	$(MAKE) run-sdk-command COMMAND=$(COMMAND) SDK_FOLDER=go
+
+run-sdk-command-restapi:
+	$(MAKE) run-sdk-command COMMAND=$(COMMAND) SDK_FOLDER=restapi
+
 
 # Runs a command for a specific SDK if it exists.
 run-sdk-command:
@@ -117,9 +121,9 @@ ensure-build-sdk-image:
 # Run SDK conformance Sidecar server in docker in order to run
 # SDK client test against it. Useful for test development
 run-sdk-conformance-local: TIMEOUT ?= 30
-run-sdk-conformance-local: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch
+run-sdk-conformance-local: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch,reserve
 run-sdk-conformance-local: ensure-agones-sdk-image
-	docker run -e "ADDRESS=" -p 59357:59357 \
+	docker run -e "ADDRESS=" -p 59357:59357 -p 59358:59358 \
 	 -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" $(sidecar_tag)
 
 # Run SDK conformance test, previously built, for a specific SDK_FOLDER
@@ -132,8 +136,8 @@ run-sdk-conformance-no-build: GRPC_PORT ?= 59357
 run-sdk-conformance-no-build: HTTP_PORT ?= 59358
 run-sdk-conformance-no-build: ensure-agones-sdk-image
 run-sdk-conformance-no-build: ensure-build-sdk-image
-	DOCKER_RUN_ARGS="--net host -e AGONES_SDK_GRPC_PORT=$(GRPC_PORT) $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
-	docker run -p $(GRPC_PORT):$(GRPC_PORT) -e "ADDRESS=" -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" -e "DELAY=$(DELAY)" \
+	DOCKER_RUN_ARGS="--net host -e AGONES_SDK_GRPC_PORT=$(GRPC_PORT) -e AGONES_SDK_HTTP_PORT=$(HTTP_PORT) $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
+	docker run -p $(GRPC_PORT):$(GRPC_PORT) -p $(HTTP_PORT):$(HTTP_PORT) -e "ADDRESS=" -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" -e "DELAY=$(DELAY)" \
 	--net=host $(sidecar_tag) --grpc-port $(GRPC_PORT) --http-port $(HTTP_PORT)
 
 # Run SDK conformance test for a specific SDK_FOLDER
@@ -151,8 +155,12 @@ run-sdk-conformance-test-go:
 run-sdk-conformance-test-rust:
 	$(MAKE) run-sdk-conformance-test SDK_FOLDER=rust
 
+run-sdk-conformance-test-rest:
+	$(MAKE) run-sdk-conformance-test SDK_FOLDER=restapi HTTP_PORT=9050
+	$(MAKE) run-sdk-command COMMAND=clean SDK_FOLDER=restapi
+
 # Run a conformance test for all SDKs supported
-run-sdk-conformance-tests: run-sdk-conformance-test-node run-sdk-conformance-test-go run-sdk-conformance-test-rust
+run-sdk-conformance-tests: run-sdk-conformance-test-node run-sdk-conformance-test-go run-sdk-conformance-test-rust run-sdk-conformance-test-rest
 
 # Clean package directories and binary files left
 # after building conformance tests for all SDKs supported
--- a/test/sdk/restapi/http-api-test.go.nolint
+++ b/test/sdk/restapi/http-api-test.go.nolint
@@ -1,0 +1,113 @@
+// Copyright 2019 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"log"
+	"os"
+	"time"
+
+	"agones.dev/agones/test/sdk/restapi/swagger"
+	"golang.org/x/net/context"
+)
+
+func main() {
+	log.Println("Client is starting")
+	conf := swagger.NewConfiguration()
+	portStr := os.Getenv("AGONES_SDK_HTTP_PORT")
+	conf.BasePath = "http://localhost:" + portStr
+	cli := swagger.NewAPIClient(conf)
+
+	ctx := context.Background()
+
+	// Wait for SDK server to start the test (30 seconds)
+	for c := 0; c < 5; c++ {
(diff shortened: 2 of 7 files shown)


aLekSer:
I am trying to avoid autogenerated code from including to this PR. So I generate `swagger` folder inside docker. And clean all files using `clean` command.
There is an issue in `golangci-lint` that it is not possible not to include all dependent files to the PR:
https://github.com/golangci/golangci-lint/issues/301
That's why I have added `.go.nolint` extension and copy file write before an execution using `go run`.
Other ways to resolve an issue with `golangci-lint` limitations would be:

1. Use Node JS client code, which would run all queries 
1. Add Swagger auto generated code to Agones repo
1. Use bash script with curl

aLekSer:
/assign @roberthbailey 
This is an HTTP endpoint test.

roberthbailey:
Can you explain why `http-api-test.go.nolint ` can't pass the linter?

build/includes/sdk.mk
```diff
@@ -129,11 +133,12 @@ run-sdk-conformance-no-build: RANDOM := $(shell bash -c 'echo $$RANDOM')
 run-sdk-conformance-no-build: DELAY ?= $(shell bash -c "echo $$[ ($(RANDOM) % 5 ) + 1 ]")
 run-sdk-conformance-no-build: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch,reserve
 run-sdk-conformance-no-build: PORT ?= 59357
```
roberthbailey: If the other one is HTTP_PORT then this one should be GRPC_PORT.

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,96 @@
...
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+	go func() {
+		gs, _, err := cli.SDKApi.WatchGameServer(ctx)
+		if err != nil {
+			log.Fatalf("Error in WatchGameServer: %v\n", err)
+		}
+		uid = gs.ObjectMeta.Uid
```
roberthbailey: https://github.com/googleforgames/agones/blob/master/test/sdk/go/sdk-client-test.go sets the uid through a channel, which seems more thread safe.

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,96 @@
...
+	}()
+
+	_, _, err := cli.SDKApi.Health(ctx, swagger.SdkEmpty{})
+	if err != nil {
+		log.Fatalf("Could not GetGameserver: %v\n", err)
```
roberthbailey: send health check

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,96 @@
...
+		} else {
+			log.Printf("Could not send Ready: %v\n", err)
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+	go func() {
+		gs, _, err := cli.SDKApi.WatchGameServer(ctx)
+		if err != nil {
+			log.Fatalf("Error in WatchGameServer: %v\n", err)
```
roberthbailey: I think log.Fatal should only happen in the main goroutine. Can this pass the error back to the main goroutine instead of a uid to signal that it failed?

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	creationTS := gs.ObjectMeta.CreationTimestamp
+
+	_, _, err = cli.SDKApi.SetLabel(ctx, swagger.SdkKeyValue{"creationTimestamp", creationTS})
+	if err != nil {
+		log.Fatalf("Could not SetLabel: %v\n", err)
+	}
+
+	time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
roberthbailey: Maybe this wouldn't be necessary if you used a channel to pass back the uid (which would block).

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,96 @@
...
+	uid = gs.ObjectMeta.Uid
+	_, _, err = cli.SDKApi.SetAnnotation(ctx, swagger.SdkKeyValue{"UID", uid})
+	if err != nil {
+		log.Fatalf("Could not SetAnnotation: %v\n", err)
+	}
+	time.Sleep(1 * time.Second)
```
roberthbailey: why a sleep here?

build/includes/sdk.mk
```diff
@@ -143,9 +148,11 @@ run-sdk-conformance-test: ensure-build-sdk-image
 
 # Run a conformance test for all SDKs supported
 run-sdk-conformance-tests:
+	$(MAKE) run-sdk-conformance-test SDK_FOLDER=restapi HTTP_PORT=9050
 	$(MAKE) run-sdk-conformance-test SDK_FOLDER=node
 	$(MAKE) run-sdk-conformance-test SDK_FOLDER=go   PORT=9001
 	$(MAKE) run-sdk-conformance-test SDK_FOLDER=rust
+	$(MAKE) run-sdk-command COMMAND=clean SDK_FOLDER=restapi
```
roberthbailey: The node sdk folder has a clean.sh script but it isn't called after running the tests. 

Should it be? Or this one removed? Either way, the pattern for testing the sdks should be consistent.

aLekSer:
> Can you explain why `http-api-test.go.nolint ` can't pass the linter?

First of all if we add this Go file we should also add multiple files generated by Swagger. This full import:
	`agones.dev/agones/test/sdk/restapi/swagger`
Second is that Swagger resulted auto-generated code does not pass the linter itself. It does not have comments prior to every exported function.

build/includes/sdk.mk
aLekSer: I think the reason with Node js was that we gitignored the results of clean. Will check it out soon.

aLekSer:
Example of linting errors can be found here:
```
test/sdk/restapi/swagger/client.go:111:6: `typeCheckParameter` is unused (deadcode)
test/sdk/restapi/swagger/client.go:125:6: `parameterToString` is unused (deadcode)
test/sdk/restapi/swagger/client.go:335:6: `reportError` is unused (deadcode)
test/sdk/restapi/swagger/client.go:440:6: `strlen` is unused (deadcode)
test/sdk/restapi/swagger/api_sdk.go:209: 209-467 lines are duplicate of `test/sdk/restapi/swagger/api_sdk.go:298-556` (dupl)
test/sdk/restapi/swagger/api_sdk.go:298: 298-556 lines are duplicate of `test/sdk/restapi/swagger/api_sdk.go:387-645` (dupl)
test/sdk/restapi/swagger/api_sdk.go:387: 387-645 lines are duplicate of `test/sdk/restapi/swagger/api_sdk.go:476-734` (dupl)
test/sdk/restapi/swagger/api_sdk.go:476: 476-734 lines are duplicate of `test/sdk/restapi/swagger/api_sdk.go:209-467` (dupl)
test/sdk/restapi/swagger/api_sdk.go:34: 34-200 lines are duplicate of `test/sdk/restapi/swagger/api_sdk.go:654-820` (dupl)
test/sdk/restapi/swagger/api_sdk.go:654: 654-820 lines are duplicate of `test/sdk/restapi/swagger/api_sdk.go:34-200` (dupl)
test/sdk/restapi/swagger/api_sdk.go:565: 565-645 lines are duplicate of `test/sdk/restapi/swagger/api_sdk.go:654-734` (dupl)
test/sdk/restapi/swagger/client.go:425:4: ineffectual assignment to `expires` (ineffassign)
```

test/sdk/restapi/http-api-test.go.nolint
aLekSer: It is redundant, removed

aLekSer:
There is one more issue with Swagger or `sdk.swagger.json` definition file, it generates wrong function for WatchGameServer which waits forever until the end of test, however it should receive and somehow output granular `GameServer` results, not wait till EOF.

aLekSer:
One more difference in that we receive `"result"` key while in swagger we expect to receive nested GS fields straight away :
```
 {"result":{"object_meta":{"name":"local","namespace":"default","uid":"3260771541929705927","resource_version":"v1","generation":"1","creation_timestamp":"1570544129","annotations":{"annotation":"true"},"labels":{"agones.dev/sdk-creationTimestamp":"1570544129","islocal":"true"}},"status":{"state":"Allocated","address":"127.0.0.1","ports":[{"name":"default","port":7777}]}}}
```

aLekSer:
Find a similar issue in `grpc-gateway`:
https://github.com/grpc-ecosystem/grpc-gateway/issues/579

roberthbailey:
> I think we should create new ticket to track above and use this version of HTTP test for all methods except WatchGameServer for now.

Agreed. It's better to test all methods but one than no methods at all.

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,96 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"log"
+	"os"
+	"time"
+
+	"agones.dev/agones/test/sdk/restapi/swagger"
```
roberthbailey: I see... it's this import here that causes the linter to fail (unless we check in the swagger files, but they don't pass the lister either).

roberthbailey:
If we leave the go file name as-is and skip the test for `WatchGameserver` it sounds like we get the rest of this PR cleaned up and merged while filing some follow-up issues. WDYT?

aLekSer:
I have updated this PR and include a TODO which could be removed when Swagger json would be updated.
I was not able to get rid of clean step for this test.
Other than that this PR is good to go.

test/sdk/restapi/http-api-test.go.nolint
aLekSer: Used channel as in Go SDK test.

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,111 @@
...
+			log.Printf("Could not send Ready: %v\n", err)
+		}
+		time.Sleep(1*time.Second)
+	}
+
+	c := make(chan string)
+
+	once := true
+	go func() {
+		gs, _, err := cli.SDKApi.WatchGameServer(ctx)
```
roberthbailey: This looks like it only fires once. I see two ramifications:

1. If the first time fails for some reason the test will fail. Should it be in a loop?
1. The error flow shouldn't be indented

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,111 @@
...
+		log.Printf("Watch response: %+v", gs)
+		if err != nil {
+			log.Printf("Error in WatchGameServer: %v\n", err)
```
roberthbailey: return (or continue if in a loop) here to keep the rest of the control flow further left.

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	creationTS := gs.ObjectMeta.CreationTimestamp
+
+	_, _, err = cli.SDKApi.SetLabel(ctx, swagger.SdkKeyValue{"creationTimestamp", creationTS})
+	if err != nil {
+		log.Fatalf("Could not SetLabel: %v\n", err)
+	}
+
+	// TODO: fix WatchGameServer() HTTP API Swagger definition and remove the following lines
```
roberthbailey: The go routine above writes to `c` as does this one. Can we delete this TODO and the below go routine? Or does the watch never fire?

test/sdk/restapi/http-api-test.go.nolint
aLekSer: I will look closely at your suggestions tomorrow. 
`WatchGameServer()` would not succeed currently, as it wait for closing the connection not line termination symbol (tracked [here #1106](https://github.com/googleforgames/agones/issues/1106)  ). So I left this lines as is and they would actually work, when Swagger spec and generator would be fixed.

test/sdk/restapi/http-api-test.go.nolint
aLekSer: The watch function does not return, mentioned this in above comment, it is waiting for EOF, and  "result". Opened a ticket for that.

build/includes/sdk.mk
aLekSer: Yes, this step is needed, because it also delete a copy of `test/sdk/restapi/http-api-test.go.nolint` which becomes `http-api-test.go` and would result in errors on `make lint` - about 50 linting errors would be displayed, all from swagger generated code.

test/sdk/restapi/http-api-test.go.nolint
```diff
@@ -0,0 +1,111 @@
...
+	"os"
+	"time"
+
+	"agones.dev/agones/test/sdk/restapi/swagger"
+	"golang.org/x/net/context"
+)
+
+func main() {
+	log.Println("Client is starting")
+	time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
aLekSer: Fixed

roberthbailey:
This PR now has a conflicting file.

Issue #1043, opened 2019-09-06
SDK package should be versioned


pooneh-m:
Currently the sdk package is not versioned. They should be versioned in case there are breaking changes in the future.
https://github.com/googleforgames/agones/blob/master/sdks/go/sdk.go#L16
https://github.com/googleforgames/agones/blob/master/sdk.proto

markmandel:
How do you think we should be versioning?

Ideally, these sdks should never, ever break.

pooneh-m:
It is not expected that the SDK would have a breaking change, but that's the assumption that almost everyone make when building a product and history proven that it will happen at one point. Using a package name that reflects the version and putting sdks in the version directory is one approach.

roberthbailey:
@pooneh-m - do you expect that someone would need to import multiple versions of the SDK into a single build (so that they would need different package names)? Or that if we made a breaking change the sidecar would export both the old and new grpc endpoints simultaneously (maybe even on different ports)?

I wouldn't mind having the gRPC API between the sdk and the sidecar have a version in the name (e.g. v1) but now that we have released 1.0 of Agones I'm not sure how we actually change the SDK proto in a backwards compatible way to achieve this. It seems like adding versioning to the sdk package would itself be a breaking change....

What if we left the current proto unversioned and if/when we need a breaking change we create a v2 of the interface (e.g. `sdk2.proto`) and add a concrete version at that point?

pooneh-m:
> @pooneh-m - do you expect that someone would need to import multiple versions of the SDK into a single build (so that they would need different package names)? Or that if we made a breaking change the sidecar would export both the old and new grpc endpoints simultaneously (maybe even on different ports)?
> 

No, we don't expect multiple SDK versions in a single build, but if a breaking change is introduced to SDK, the old stable version should still work and users should be able to choose either versions.

> I wouldn't mind having the gRPC API between the sdk and the sidecar have a version in the name (e.g. v1) but now that we have released 1.0 of Agones I'm not sure how we actually change the SDK proto in a backwards compatible way to achieve this. It seems like adding versioning to the sdk package would itself be a breaking change....
> 
> What if we left the current proto unversioned and if/when we need a breaking change we create a v2 of the interface (e.g. `sdk2.proto`) and add a concrete version at that point?

Makes sense to introduce package versioning for sdk after v2 to avoid breaking change to v1.

markmandel:
[...]

PR #1080, opened 2019-09-26
Feature Stages Documentation

markmandel:
Feedback on below very much warranted and appreciated - from where it sits in the documentation, to the contents itself.

---

Definitions of feature stages, how they are enabled/disabled, the support to be expected at each stage, and how each stage can be identified.

Updates to the contributing guide as well, as this needs to be kept in mind for new features and design work.

Credit / Reference:
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages
https://istio.io/about/feature-stages/#feature-phase-definitions


site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
...
+Any `alpha` or `beta` Game Server SDK functionality will have a comment block before each function declaring it either
+`alpha` or `beta`, as well as documentation indicating it as such.
+
+Any other SDK functionality not marked as `alpha` or `beta` can be assumed to be `stable`.
+
+### REST & gRPC APIs 
+
+REST and gRPC API will have versioned paths were appropriate to indicate their feature stage.
+
+For example, a REST API with a prefix of `v1alpha1` is an `alpha` stage feature.
```
pooneh-m: Can you please clarify?

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
...
+### Custom Resource Definitions (CRDs)
+
+This refers to Kubernetes resource for Agones, such as `GameServer`, `Fleet` and `GameServerAllocation`.
+
+#### New CRDs
+
+For new resources, the stage of the resource will be indicated by the `apiVersion` of the resource.
+
+For example: `apiVersion: "agones.dev/v1"` is a `stable` resource, `apiVersion: "agones.dev/v1beta1"` is a `beta`
+ stage resource, and `apiVersion: "agones.dev/v1alpha1"` is a `alpha` stage resource.
```
pooneh-m: an alpha?

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
...
+      - name: simple-udp
+        image: gcr.io/agones-images/udp-server:0.15
+``` 
+
+As resource attributes progress through their stages, there may be breaking changes, as backward conversion between
+ `alpha`, `beta` and `stable` positioning in the resource configuration may not be guaranteed.
+
+### Agones Game Server SDK
+
+Any `alpha` or `beta` Game Server SDK functionality will have a comment block before each function declaring it either
```
pooneh-m: SDK path and its package should be reflective of its version. WDYT?

CONTRIBUTING.md
```diff
@@ -48,6 +48,16 @@ When submitting pull requests, make sure to do the following:
 - Remove trailing whitespace. Many editors will do this automatically.
 - Ensure any new files have [a trailing newline](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file)
 
+## Feature Stages
+
+Often, new features will need to go through experimental stages so that we can gather feedback and adjust as necessary.
+
+You can see this project's [feature stage documentation](https://agones.dev/docs/guides/feature-stages/) on the Agones
```
roberthbailey: Note - this will be a dead link until the website page is published. Probably ok, but the github changes go live immediately.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
+---
+title: "Feature Stages"
+linkTitle: "Feature Stages"
+date: 2019-09-26T01:20:41Z
```
roberthbailey: This doesn't have a publish date but I'm not seeing it in the website preview: https://a96d2ff-dot-preview-dot-agones-images.appspot.com/site/docs/

Any idea why it isn't showing up?

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
...
+  This page provides a description of the various stages that Agones features can be in, and the
+  relative maturity and support level expected for each level.
+---
+
+A feature within Agones can be in `Alpha`, `Beta` or `Stable` stage.
+
+## Feature Gates
+
+`Alpha` and `Beta` features can be enabled or disabled through the `agones.featureGate` configuration option 
+that can be found in the [Helm configuration]({{< ref "/docs/Installation/helm.md#configuration" >}}) documentation.
```
roberthbailey: This implies that if you install via other mechanisms (e.g. install.yaml) that you cannot toggle feature gates. This pushes us a bit further down the "helm is the right way" path, which is probably ok but we should be cognizant that we are doing it.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
...
+Documentation on configuring each non-`stable` feature gate can be found on the documentation for the specific 
+feature.
+
+## Description of Stages
+
+### Alpha
+
+An `Alpha` feature means:
+
+* Disabled by default.
```
roberthbailey: Does this also mean that we need to turn multicluster allocation off by default?

site/content/en/docs/Guides/feature-stages.md
aLekSer: On recent master after adding this file and running locally with `make site-server` the file got displayed as it should, refer to the screen:

Screenshot 2019-10-01 at 17 09 43

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
...
+> through the `Beta` period.
+
+### Beta
+
+A `Beta` feature means:
+
+* Enabled by default, but able to be disabled through a feature gate.
+* The feature is well tested. Enabling the feature is considered safe.
+* Support for the overall feature will not be dropped, though details may change.
+* The schema and/or semantics of objects may change in incompatible ways in a subsequent beta or stable release. When
```
aLekSer:
```suggestion
* The schema and/or semantics of objects may change in incompatible ways in a subsequent beta or stable releases. When
```

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Description of Stages
+
+### Alpha
+
+An `Alpha` feature means:
+
+* Disabled by default.
+* Might be buggy. Enabling the feature may expose bugs.
+* Support for feature may be dropped at any time without notice.
```
aLekSer:
```suggestion
* Support for this feature may be dropped at any time without notice.
```

site/content/en/docs/Guides/feature-stages.md
markmandel: It's possible it got deleted, as we only keep the past 30 versions.

site/content/en/docs/Guides/feature-stages.md
markmandel: Should probably have examples?

Basically I'm saying if the REST Apis's path is something like: `/v1alpha1/dothing/` then it's alpha. Does that make sense?

site/content/en/docs/Guides/feature-stages.md
markmandel: Yeah - I'm not sure what the right answer here is. I honestly put this approach in here to hope someone would disagree with me :smile: 

Say we go by package -- maybe we do a subpackage? e.g. `sdk.alpha` ? I think it's trickier to do multiple top level packages across different languages -- but subpackages would solve the problems here with the current approach not being very obvious about which status of the SDK you are working against.

The only concern I can think of is that I don't think we can do multiple packages in our proto definition -- so while the user won't see this, we'll need to be clear in our comments on the proto definitions themselves.

How does that sound?

site/content/en/docs/Guides/feature-stages.md
markmandel: We don't have a feature gate for it at the moment, so yes, probably?

We would need to build one in for beta, so people can turn it off as well.

Also, in CI - what do we turn on or off? :smile: I just thought of that too. What do we do there?

site/content/en/docs/Guides/feature-stages.md
markmandel: Yeah that is a good point - but you really can't configure much with the install.yaml -- we likely should be clear that install.yaml is good for a "getting started", and helm is good for "production" (another reason to rejig the install pages)

CONTRIBUTING.md
markmandel: True - it usually goes live within minutes of the commit going to master - so not something I'm deeply concerned about.

roberthbailey:
I'm happy removing the gates (and their documentation) once they hit stable. I think having a list of things you can try _right now_ makes sense and you should be able to tell which feature gates apply to your version of Agones using the drop down on the website that takes you to the version specific docs.

site/content/en/docs/Guides/feature-stages.md
roberthbailey: The preview was there, just not the new page (it's easier to review site changes looking at the actual site rather than the raw md). I see it now though -- not sure what happened before.

site/content/en/docs/Guides/feature-stages.md
roberthbailey: sgtm.

site/content/en/docs/Guides/feature-stages.md
roberthbailey: Our existing CI should run with the default settings (to match the beta / stable behavior). For each feature gate, we should decide if we need to run a separate job with it enabled and any feature specific tests. Then when feature graduate to beta and are on by default we can move the tests to the main suite and delete the feature specific tests. 

Of course, this would require some rework to our CI setup, so for now we probably just say that alpha stuff is tested manually? Maybe an exception makes sense for multicluster allocation since it's already in the test suite and it may not be worth breaking out at this point.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,124 @@
...
+## Feature Gates
+
+`Alpha` and `Beta` features can be enabled or disabled through the `agones.featureGate` configuration option 
+that can be found in the [Helm configuration]({{< ref "/docs/Installation/helm.md#configuration" >}}) documentation.
+
+The current set of `alpha` and `beta` feature gates are:
+
+| Feature | Default | Stage | Since |
+|---------|---------|-------|-------|
+| None currently |
```
roberthbailey: Please add multicluster allocation to this table. Even if we put an asterisk and say it was grandfathered in before the process was put into place, it's clearly alpha stage (by the api definition) and we should call that out here.

site/content/en/docs/Guides/feature-stages.md
pooneh-m: Yes, it would be useful to make an example or provide a URI versioning template e.g. <api-version>/<noun>. https://restfulapi.net/versioning/
Also, we should be more specific about proto versioning. Proto package and directory should reflect the version.

site/content/en/docs/Guides/feature-stages.md
pooneh-m: > The only concern I can think of is that I don't think we can do multiple packages in our proto definition -- so while the user won't see this, we'll need to be clear in our comments on the proto definitions themselves.
> 
> How does that sound?

[This](https://cloud.google.com/apis/design/versioning) has some recommendations:

- "For all versions of the API other than v1, the major version number must be encoded as the last component of the proto package name. For example, google.pubsub.v2."

- "The minor and patch numbers should be reflected in the API configuration and documentation. They must not be encoded in the proto package name."

- "A stable API version must only depend on the latest stable version of other APIs."

site/content/en/docs/Guides/feature-stages.md
markmandel: Done!

site/content/en/docs/Guides/feature-stages.md
markmandel: Waiting on resolution on #1043 before coming back to this section.

site/content/en/docs/Guides/feature-stages.md
markmandel: Added a REST example, WDYT - is that enough?

CONTRIBUTING.md
roberthbailey: I guess it's only relevant if we hide the feature stages until the next release (using a publish date in the future). If we are just going to push it live then it's not something to worry about.

site/content/en/docs/Guides/feature-stages.md
roberthbailey: I think we should just grandfather in multicluster allocation at this point. It might be worth adding a way to disable it (if you really want to) but we should leave it on by default, leave it on in CI, and tackle this with the next new feature we introduce that is behind a gate.

markmandel:
@roberthbailey @pooneh-m I think this is pretty close - I updated the gRPC section with a resolution from the research in #1043. PTAL, let me know what you think.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,131 @@
...
+Any other SDK functionality not marked as `alpha` or `beta` can be assumed to be `stable`.
+
+### REST & gRPC APIs 
+
+REST and gRPC API will have versioned paths were appropriate to indicate their feature stage.
+
+For example, a REST API with a prefix of `v1alpha1` is an `alpha` stage feature: 
+`http://api.example.com/v1alpha1/exampleaction`.
+
+Similar to the SDK, Any `alpha` or `beta` gRPC functionality will be a subpackage of the main API package.
```
pooneh-m: How about if there is a breaking change to the v1 API?

pooneh-m: s/Any/any

pooneh-m: We should perhaps mention that alpha/beta features are a super-set of the main package?

CONTRIBUTING.md
```diff
@@ -48,6 +48,16 @@ When submitting pull requests, make sure to do the following:
 - Remove trailing whitespace. Many editors will do this automatically.
 - Ensure any new files have [a trailing newline](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file)
 
+## Feature Stages
```
pooneh-m: How about if there is a breaking change in the controller? How Agones is going to be versioned? v1.x.x -> v2.x.x?

markmandel: While I don't think it _likely_ there will be breaking change in the controller, we should probably mention we are following semver in the Agones releases. That sounds good?

site/content/en/docs/Guides/feature-stages.md
markmandel: So you are saying all stable features should be available in the alpha/beta sdk subpackage?

So I should be able to do `sdk.alphav1.Ready()`? 

If so, I'm not sure what the benefit would be?

markmandel: Ta!

CONTRIBUTING.md
pooneh-m: Should be good enough. This is a good example: https://kubernetes.io/docs/setup/release/version-skew-policy/

site/content/en/docs/Guides/feature-stages.md
markmandel: This is a good question. Ideally, we never want this to happen - since the SDK is so important -- but I may be being optimistic that it won't ever happen - or at least to the degree that we want a whole new implementation of the SDK.

I'm almost tempted to say "handle it if/when we need to?, and hopefully we won't have to?"

markmandel: Oh wait, wrong section - this isn't the SDK section, this is the API section.

So my only concern here, is that if we do a v1/v2 type API surface, it's completely different from the SDK. Maybe that is fine though? I'd like to have consistency however.

WDYT?

pooneh-m: > So you are saying all stable features should be available in the alpha/beta sdk subpackage?
> 
I think now I understand why you wanted to run two SDK packages side by side because one is supporting a none stable API and the other supports a stable API. If this is what you are recommending, please make is clear. Otherwise, I assumed alpha API would be a super set of stable API and there is only one SDK running. So if sdk is now supporting an alpha feature SelfNominate(), I thought I create a new snapshot of the SDK and version it as sdk.v1alpha1 which is a super set of sdk[.v1] package and expose it for experimentation.

pooneh-m: Optimistically, I also expect major versioning would not happen However, it wouldn't hurt to mention that if there is any required incompatible change to the stable version there should be an extensive communication and planning.

markmandel: This sounds like a great addition as well :+1:

markmandel: Aaah! Yes! That is the disconnect. Yes, that is exactly the way I figured it would work, but I wasn't clear. Let me write words to that effect (it sounds like it will be similar to what we have in CRDs as well).

We can also do something like keep alpha/beta SDKs in place for a period after they move to stable, so that people have time to move over as well :+1:

markmandel: How strongly do you feed about having a v1/v2 interface/package for the gRPC/REST endpoints.

I'm thinking I might be backtracking on my position of not doing this - just because it's going to be really hard to go back and slot this in if we need it (SDK comes to mind).

I'm not a huge fan of packages called `v1` (hello K8s), but I can't think of a better way.

pooneh-m: Version in the package name for gRPC for v1 is optional. 
For REST, I have always seen APIs to be versioned, but couldn't find a recommendation specific to v1 versioning. I would recommend designing REST API with version in the path to avoid URL changes between versions.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,145 @@
...
+
+As resource attributes progress through their stages, there may be breaking changes, as backward conversion between
+ `alpha`, `beta` and `stable` positioning in the resource configuration may not be guaranteed.
+
+### Agones Game Server SDK
+
+Any `alpha` or `beta` Game Server SDK functionality will be a subpackage of the `sdk` package. For example
+, functionality found in a `sdk.alphav1` package should be considered at the `alpha` feature stage.
+
+Only experimental functionality will be found in any `alpha` and `beta` SDK packages, and as such may change as
```
roberthbailey: This implies that you would import both the alpha and stable SDKs to start using the alpha sdk functions. I'm not actually sure how that would work in practice (do you have to call `Connect` twice?). Does the alpha package piggyback on the stable one somehow?

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,145 @@
...
+
+As SDK features move to through feature stages towards `stable`, the previous version of the SDK API
+will remain for at least one release to enable easy migration to the more stable feature stage (i.e. from `alpha
+` -> `beta`, `beta` -> `alpha`)
+
+Any other SDK functionality not marked as `alpha` or `beta` can be assumed to be `stable`.
+
+### REST & gRPC APIs 
+
+REST and gRPC API will have versioned paths were appropriate to indicate their feature stage.
```
roberthbailey: s/were/where

CONTRIBUTING.md
pooneh-m: Any change in the behavior that a user can get dependency on could be a breaking change, such as not setting finalizers on game servers.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -0,0 +1,145 @@
...
+
+An `Alpha` feature means:
+
+* Disabled by default.
+* Might be buggy. Enabling the feature may expose bugs.
+* Support for this feature may be dropped at any time without notice.
+* The API may change in incompatible ways in a later software release without notice.
+* Recommended for use only in short-lived testing clusters, due to increased risk of bugs and lack of long-term support.
+
+> Note: Please do try `Alpha` features and give feedback on them. This is important to ensure less breaking changes
```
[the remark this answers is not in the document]
markmandel: Nice. Good catch.

site/content/en/docs/Guides/feature-stages.md
markmandel: Yes, you would need to explicitly import the alpha package to use it. But I feel like from an implementation detail we would hide the gRPC details from the user.

From a gRPC technical details, I posted some details on how to do this technically on #1043

> Aha, so I found the right stackoverflow post:
https://stackoverflow.com/questions/52634217/access-multiple-grpc-services-over-the-same-connection
> 
> This shows we can both register multiple services with the same endpoint, and also configure a singular connection with multiple clients - at least with Go.
>
> Even if we end up having 2 connections to the same local endpoint, I don't think it's a drastic issue?

If we had `alpha/`beta` be a superset of `stable`, would we demark a section of sdk.proto for alpha/beta functionality, and add/remove from there as needed? (and basically hide this from the user via our published SDK? This would have some complications for anyone generating a SDK, I would assume?) 

Not to say I'm 100% against a superset, it was just my original viewpoint. I'm not completely married to the idea - maybe a superset is easier for us/the end user.

markmandel:
@pooneh-m @roberthbailey I'm almost wondering if we want to coordinate a video call and resolve this PR?

pooneh-m:
> @pooneh-m @roberthbailey I'm almost wondering if we want to coordinate a video call and resolve this PR?

It seems like the outstanding issue is the gRPC versioning. Should we consult with gRPC experts then to help making the decision? Other than that, LGTM.

markmandel:
We have an internal gRPC office hours, I'm going to go tomorrow morning, and will get a resolution.

markmandel:
Had a chat with the gRPC team:

1) Confirmed: On the server side, yes, you can register multiple proto defined services through a single gRPC ip and port (multiple registrations on a single server)
2) Confirmed: Client side, you can clients from multiple protos reuse the connection to that above single server from the same multiple protos. Some languages will share "channels", in others, if the client works out you are writing code to connect to the exact same place that the client already has a connection to - it will automatically reuse the TCP connection.

So I think we are good to go on the currently documented approach?

markmandel:
Anything outstanding on this feature stage doc?

markmandel:
Oooh! Now we have to actually build the feature gate functionality :smile:

markmandel:
```
--- FAIL: TestMultiClusterAllocationOnLocalCluster (0.44s)
    --- FAIL: TestMultiClusterAllocationOnLocalCluster/Packed (44.51s)
        gameserverallocation_test.go:80: Namespace gsa-multicluster-local-8a7d0c9d-1d3d-11ea-b2b9-0242c0a80a03 is created
        gameserverallocation_test.go:80: ServiceAccount gsa-multicluster-local-8a7d0c9d-1d3d-11ea-b2b9-0242c0a80a03/agones-sdk is created
        gameserverallocation_test.go:80: RoleBinding gsa-multicluster-local-8a7d0c9d-1d3d-11ea-b2b9-0242c0a80a03/agones-sdk-access is created
        gameserverallocation_test.go:182: 
            	Error Trace:	gameserverallocation_test.go:182
            	Error:      	Expected nil, but got: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:""}, Status:"Failure", Message:"an error on the server (\"no multi-cluster allocation policy is specified\") has prevented the request from succeeding (post gameserverallocations.allocation.agones.dev)", Reason:"InternalError", Details:(*v1.StatusDetails)(0xc0008ac1e0), Code:500}}
            	Test:       	TestMultiClusterAllocationOnLocalCluster/Packed
```

That's a new one. @pooneh-m in case you know what's up here.

pooneh-m:
> ```
> --- FAIL: TestMultiClusterAllocationOnLocalCluster (0.44s)
>     --- FAIL: TestMultiClusterAllocationOnLocalCluster/Packed (44.51s)
>         gameserverallocation_test.go:80: Namespace gsa-multicluster-local-8a7d0c9d-1d3d-11ea-b2b9-0242c0a80a03 is created
>         gameserverallocation_test.go:80: ServiceAccount gsa-multicluster-local-8a7d0c9d-1d3d-11ea-b2b9-0242c0a80a03/agones-sdk is created
>         gameserverallocation_test.go:80: RoleBinding gsa-multicluster-local-8a7d0c9d-1d3d-11ea-b2b9-0242c0a80a03/agones-sdk-access is created
>         gameserverallocation_test.go:182: 
>             	Error Trace:	gameserverallocation_test.go:182
>             	Error:      	Expected nil, but got: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:""}, Status:"Failure", Message:"an error on the server (\"no multi-cluster allocation policy is specified\") has prevented the request from succeeding (post gameserverallocations.allocation.agones.dev)", Reason:"InternalError", Details:(*v1.StatusDetails)(0xc0008ac1e0), Code:500}}
>             	Test:       	TestMultiClusterAllocationOnLocalCluster/Packed
> ```
> 
> That's a new one. @pooneh-m in case you know what's up here.

It's a flake. I opened https://github.com/googleforgames/agones/issues/1114 while ago, but it didn't happen again. I'll reopen it to fix it. I think a retry will fix it.

PR #1086, opened 2019-09-28
Use environment variables to dynamically set the grpc port in the golang sdk.

roberthbailey:
I wrote some code to show what option 1 from https://github.com/googleforgames/agones/issues/851#issuecomment-536127066 would look like, since it seemed like the best option to me. 

I'll add tests and do some more checking if we decide to go this route, but I've already verified:

1. The sdk / sdkserver can be set to different ports and the gameserver becomes ready
1. If you specify the env vars they get overridden
1. If you specify the env vars in a configmap using EnvFromSource they get overridden

so it seems pretty robust.

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strconv"
 	"sync"
 	"time"
 
@@ -53,6 +54,12 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
+)
+
+const (
+	sdkserverSidecarName = "agones-gameserver-sidecar"
+	grpcPortEnvVar       = "AGONES_SDK_GRPC_PORT"
+	httpPortEnvVar       = "AGONES_SDK_HTTP_PORT"
 )
 
 // Controller is a the main GameServer crd controller
@@ -520,11 +527,9 @@ func (c *Controller) syncDevelopmentGameServer(gs *agonesv1.GameServer) (*agones
 // createGameServerPod creates the backing Pod for a given GameServer
 func (c *Controller) createGameServerPod(gs *agonesv1.GameServer) (*agonesv1.GameServer, error) {
 	sidecar := c.sidecar(gs)
-	var pod *corev1.Pod
 	pod, err := gs.Pod(sidecar)
-
-	// this shouldn't happen, but if it does.
 	if err != nil {
+		// this shouldn't happen, but if it does.
 		c.loggerForGameServer(gs).WithError(err).Error("error creating pod from Game Server")
 		gs, err = c.moveToErrorState(gs, err.Error())
 		return gs, err
@@ -539,6 +544,7 @@ func (c *Controller) createGameServerPod(gs *agonesv1.GameServer) (*agonesv1.Gam
 	}
 
 	c.addGameServerHealthCheck(gs, pod)
+	c.addSDKServerEnvVars(gs, pod)
 
 	c.loggerForGameServer(gs).WithField("pod", pod).Info("creating Pod for GameServer")
 	pod, err = c.podGetter.Pods(gs.ObjectMeta.Namespace).Create(pod)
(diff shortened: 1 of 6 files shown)


pkg/gameservers/controller.go
```diff
@@ -639,6 +639,50 @@ func (c *Controller) addGameServerHealthCheck(gs *agonesv1.GameServer, pod *core
 	})
 }
 
+func (c *Controller) addSDKEnvironmentVarsToGameServer(gs *agonesv1.GameServer, pod *corev1.Pod) {
```
markmandel: I think we need to set the environment variables to all containers, not just the gameserver one.

Reason being, lots of people are running sidecars with the sdk embedded there, rather than doing the direct integration into the gameserver binary itself.

So we need to ensure the SDK will work in any container that is running in the GameServer Pod

pkg/gameservers/controller.go
```diff
@@ -639,6 +639,50 @@ func (c *Controller) addGameServerHealthCheck(gs *agonesv1.GameServer, pod *core
...
+	var env []corev1.EnvVar
+	if gs.Spec.SdkServer.GRPCPort != 0 {
+		env = append(env, corev1.EnvVar{
+			Name:  "AGONES_SDK_GRPC_PORT",
+			Value: strconv.Itoa(int(gs.Spec.SdkServer.GRPCPort)),
+		})
+	}
+	if gs.Spec.SdkServer.HTTPPort != 0 {
+		env = append(env, corev1.EnvVar{
+			Name:  "AGONES_SDK_GRPC_PROXY_PORT",
```
markmandel:
```suggestion
			Name:  "AGONES_SDK_HTTP_PORT",
```
?

Or REST_PORT ?

Not a strong opinion, just a thought.

pkg/gameservers/controller.go
roberthbailey: Makes sense. I've reversed the logic to apply to all containers except for the sdksidecar instead of just the indicated game server container.

pkg/gameservers/controller.go
roberthbailey: Changed the variable and made it into a constant.

roberthbailey:
> We should also include some docs on these environment variables

I was waiting for a +1 on the approach before finishing the PR. I've added unit tests and documentation.

pkg/gameservers/controller.go
```diff
@@ -639,6 +645,52 @@ func (c *Controller) addGameServerHealthCheck(gs *agonesv1.GameServer, pod *core
...
+				// don't change the container spec.
+				continue
+			}
+
+			// Filter out environment variables that have reserved names.
+			// From https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
+			env := c.Env[:0]
+			for _, e := range c.Env {
+				if !reservedEnvironmentVariableName(e.Name) {
+					fmt.Printf("adding %s to final env\n", e.Name)
```
markmandel: Should this be removed? If not, it should probably go through logrus.

sdks/go/sdk.go
```diff
@@ -39,11 +41,29 @@ type SDK struct {
...
 
+func port() int {
+	portStr := os.Getenv("AGONES_SDK_GRPC_PORT")
+	if portStr == "" {
+		// Environment variable is not set; use the default port.
+		return defaultPort
+	}
+	p, err := strconv.Atoi(portStr)
+	if err != nil {
+		// Environment variable cannot be parsed; use the default port.
```
markmandel: Should we log a warning here?

sdks/go/sdk.go
```diff
@@ -39,11 +41,29 @@ type SDK struct {
...
+		return defaultPort
+	}
+	if p < 1 || p > 65535 {
+		// Environment variable is not a valid port; use the default port.
```
markmandel: Maybe log a warning here?

pkg/gameservers/controller.go
roberthbailey: Yes, left over debugging. As an aside, I couldn't figure out why modifying c in place had no effect and I had to reassign the modified c back to the containers array (hence the debugging).

sdks/go/sdk.go
roberthbailey: Do you mean printing to stdout (like in the `WatchGameServer` function? There isn't a logger defined in this package.

roberthbailey: I added some print statements to stderr since that seems like a better place to put them than stdout.

roberthbailey:
I found a few more places on the website to update with the new environment variable.

I also noticed that https://agones.dev/site/docs/guides/client-sdks/unreal/#settings is rendering the list content incorrectly. It looks correct in my markdown preview window, so I'm not sure what docsy/hugo is doing to it.

sdks/go/sdk.go
markmandel: Ah yes, we don't have a logger here. This makes sense :+1:

PR #1092, opened 2019-10-02
Set the port to connect to the sdkserver based on the AGONES_SDK_GRPC_PORT environment variable

roberthbailey:
@devjgm -- looking for some feedback on the way that I'm getting the environment variable and converting it to an integer. I'm not sure whether either is proper C++ (it felt like I was writing C so I'm guessing there's a better way to do it). 

Part of https://github.com/googleforgames/agones/issues/851.

### DIFF
--- a/sdks/cpp/src/agones/sdk.cc
+++ b/sdks/cpp/src/agones/sdk.cc
@@ -20,6 +20,7 @@
 namespace agones {
 
 struct SDK::SDKImpl {
+  std::string host_;
   std::shared_ptr<grpc::Channel> channel_;
   std::unique_ptr<agones::dev::sdk::SDK::Stub> stub_;
   std::unique_ptr<grpc::ClientWriter<agones::dev::sdk::Empty>> health_;
@@ -27,8 +28,10 @@ struct SDK::SDKImpl {
 };
 
 SDK::SDK() : pimpl_{std::make_unique<SDKImpl>()} {
-  pimpl_->channel_ = grpc::CreateChannel("localhost:59357",
-                                         grpc::InsecureChannelCredentials());
+  const char* port = std::getenv("AGONES_SDK_GRPC_PORT");
+  pimpl_->host_ = std::string("localhost:") + (port ? port : "59357");
+  pimpl_->channel_ =
+      grpc::CreateChannel(pimpl_->host_, grpc::InsecureChannelCredentials());
 }
 
 SDK::~SDK() {}
@@ -37,6 +40,8 @@ bool SDK::Connect() {
   if (!pimpl_->channel_->WaitForConnected(
           gpr_time_add(gpr_now(GPR_CLOCK_REALTIME),
                        gpr_time_from_seconds(30, GPR_TIMESPAN)))) {
+    std::cerr << "Could not connect to the sidecar at " << pimpl_->host_
+              << ".\n";
     return false;
   }
 
--- a/examples/cpp-simple/server.cc
+++ b/examples/cpp-simple/server.cc
@@ -50,7 +50,7 @@ int main() {
 
   std::cout << "Attempting to connect...\n" << std::flush;
   if (!sdk->Connect()) {
-    std::cerr << "Could not connect to the sidecar. Exiting!\n";
+    std::cerr << "Exiting!\n";
     return -1;
   }
   std::cout << "...handshake complete.\n" << std::flush;


sdks/cpp/src/agones/sdk.cc
```diff
@@ -26,8 +28,33 @@ struct SDK::SDKImpl {
   std::unique_ptr<grpc::ClientContext> health_context_;
 };
 
+std::string port() {
```
devjgm: I think the use of `std::getenv()` here is good. For parsing strings to numbers you generally want to use something like `std::strtoll` rather than `sscanf`, unless you're parsing a complicated string.

However, in this case, I slightly question the need to parse and validate the port number, then just to turn it back into a string again. My suggest would be to always use the value in the env var if present, otherwise use the default. Something like:

```cc
namespace {
std::string GetPort() {
  const char* port = std::getenv("AGONES_SDK_GRPC_PORT");
  return port ? port : "5937";
}
}  // namespace
```

What do you think?

roberthbailey: The advantage of the current code is that if you pass an invalid value for the port (not a number or a number that isn't a valid port number to bind to) then you get both an error message and a client that attempts to connect to the default port:

```
$ docker run -e AGONES_SDK_GRPC_PORT=sathoeu gcr.io/robertbailey-agones/cpp-simple-server:0.10
C++ Game Server has started!
Getting the instance of the SDK.
Attempting to connect...
Unable to parse 'sathoeu' defined in AGONES_SDK_GRPC_PORT into an integer
```

With your suggestion, I get:

```
$ docker run -e AGONES_SDK_GRPC_PORT=stahou gcr.io/robertbailey-agones/cpp-simple-server:0.10
C++ Game Server has started!
Getting the instance of the SDK.
Attempting to connect...
Could not connect to the sidecar. Exiting!
```

roberthbailey: I've replaced `sscanf` with `std::strtoll` but I think that having the extra parsing (and error messages), while much more verbose than your suggestion (which also works) is a better experience for anyone trying to debug connection issues between their game and the sdkserver sidecar container. In practice, this shouldn't actually happen, so maybe it's just me overtraining on my attempts to debug this while testing modifying port numbers, and it's more code to maintain, so if you feel strongly then I'm happy to acquiesce.

sdks/cpp/src/agones/sdk.cc
```diff
@@ -26,8 +28,33 @@ struct SDK::SDKImpl {
...
+  const std::string default_grpc_port = "59357";
+  const char* env_port = std::getenv("AGONES_SDK_GRPC_PORT");
+  if (env_port == 0) {
```
devjgm: Points should be compared to `nullptr`.

sdks/cpp/src/agones/sdk.cc
```diff
@@ -26,8 +28,33 @@ struct SDK::SDKImpl {
...
+    std::cerr << "Environment variable AGONES_SDK_GRPC_PORT not defined, using "
```
devjgm: I don't think this is an error case. I suggest removing this error log.

sdks/cpp/src/agones/sdk.cc
```diff
@@ -26,8 +28,33 @@ struct SDK::SDKImpl {
...
+                 "default port "
+              << default_grpc_port << "\n";
+    return default_grpc_port;
+  }
+  char* end;
+  long port = std::strtoll(env_port, &end, 10);
+  if (env_port == end) {
```
devjgm: I think this also needs to look at `errno` to do proper error checking.

sdks/cpp/src/agones/sdk.cc
devjgm: OK. I've made some comments below in case you want to keep this port validation code. Of course, I'll approve with this.

But I will try to clarify why I think this is a mistake.

1. This function isn't using the port, so I claim this is the wrong place to do any verification of the port. Does this function know *how* the port is going to be used? Is this function *sure* that the port must be a number and not, say, an alphanumeric service name (like from `/etc/services`), or something else? This function is enforcing some policy about what makes a valid port number without any knowledge of what that policy should or needs to be.

2. Validating the port number here is non-trivial. A 20-line function is more likely to have bugs than is a 2-line function.

The code that *should* validate the port number is the code that actually uses it. If an invalid port number is given, does `grpc::CreateChannels` return an error (or nullptr)? If so, that's where and how we should report the error.

In your example above that prints "Could not connect to the sidecar", can *that* error message simply include the grpc target it was trying to connect to? For example, if it said "Could not connect to the sidecar at 127.0.0.1:stahou" that would make it clear to the user what the problem is.

devjgm: The C++ Style guide says that this function name should start with a capital letter and likely have some verb in it. So I'd suggest `GetPort` or `LookupPort` or something like that.

roberthbailey: Maybe it isn't the best SDK design, but the `Connect` function just returns a boolean (success or failure) and doesn't log any sort of error message (the client can if they want to, which is what we see here). The address that is being connected to is opaque to the client, and the client has no way to know what address to add to the log message. 

The reason I added the output with your suggestion is to show that (a) `grpc::CreateChannels` successfully creates a "channel" to `127.0.0.1:stahou` and (b) `channel_->WaitForConnected` attempts to connect and just times out. 

I added the extra error message because we happen to know that in our case the thing we are connecting to is a port on localhost. Since we know more information about the endpoint than the libraries, we can add validation and error messages that the libraries can't (they probably do work with an alphanumeric service name since they are designed to be more flexible). 

In practice, this should only be an issue for people trying to run things manually outside of agones, so I think adding an error message from inside the `Connect` function might be the simplest thing to do (rather than adding error messages inside of the constructor). 

Here's what the output (from the simple server) looks like after this change:

```
$ docker run --net host -e AGONES_SDK_GRPC_PORT=3344aaa  gcr.io/robertbailey-agones/cpp-simple-server:0.10
C++ Game Server has started!
Getting the instance of the SDK.
Attempting to connect...
Could not connect to the sidecar at localhost:3344aaa.
Exiting!
```

sdks/cpp/src/agones/sdk.cc
```diff
@@ -20,15 +20,18 @@
...
   std::unique_ptr<agones::dev::sdk::SDK::Stub> stub_;
   std::unique_ptr<grpc::ClientWriter<agones::dev::sdk::Empty>> health_;
   std::unique_ptr<grpc::ClientContext> health_context_;
 };
 
 SDK::SDK() : pimpl_{std::make_unique<SDKImpl>()} {
-  pimpl_->channel_ = grpc::CreateChannel("localhost:59357",
-                                         grpc::InsecureChannelCredentials());
+  const char* port = std::getenv("AGONES_SDK_GRPC_PORT");
+  pimpl_->host_ = "localhost:" + (port ? std::string(port) : "5937");
```
devjgm: Nit: It might be nice to pull the `std::string` construction outside the ternary expression. Consider this 

```cc
piple_->host_ = std::string("localhost:") + (port ? port : "59357");
```

devjgm: "5937" may be a typo?

roberthbailey: Yes. Fixed.

roberthbailey: done.

roberthbailey:
/assign @pooneh-m 

(for approval, since @devjgm already lgtm'd it)

PR #1093, opened 2019-10-02
Use an environment variable to dynamically set the grpc port in the nodejs sdk

roberthbailey:
/assign @steven-supersolid

Part of #851.

### DIFF
--- a/sdks/nodejs/src/agonesSDK.js
+++ b/sdks/nodejs/src/agonesSDK.js
@@ -19,9 +19,28 @@ const services = require('../lib/sdk_grpc_pb');
 
 class AgonesSDK {
 	constructor() {
-		this.client = new services.SDKClient('localhost:59357', grpc.credentials.createInsecure());
+		this.client = new services.SDKClient('localhost:'+this.port, grpc.credentials.createInsecure());
 		this.healthStream = undefined;
 		this.emitters = [];
+	}
+
+	get port() {
+		const defaultPort = '59357';
+		const port = process.env.AGONES_SDK_GRPC_PORT;
+		if (port === undefined) {
+			console.log(`Environment variable AGONES_SDK_GRPC_PORT not defined, using default port ${defaultPort}`);
+			return defaultPort;
+		}
+		const portNum = parseInt(port, 10);
+		if (isNaN(portNum)) {
+			console.log(`Unable to parse '${port}' defined in AGONES_SDK_GRPC_PORT into an integer`);
+			return defaultPort;
+		}
+		if (portNum < 1 || portNum > 65535) {
+			console.log(`Invalid port ${portNum} defined in AGONES_SDK_GRPC_PORT. It must be between 1 and 65535`);
+			return defaultPort;
+		}
+		return port;
 	}
 
 	async connect() {
@@ -32,7 +51,7 @@ class AgonesSDK {
 				} else {
 					resolve();
 				}
-			})
+			});
 		});
 	}
 
--- a/sdks/nodejs/spec/agonesSDK.spec.js
+++ b/sdks/nodejs/spec/agonesSDK.spec.js
@@ -26,6 +26,37 @@ describe('agones', () => {
 		agonesSDK = new AgonesSDK();
 	});
 
+	describe('port', () => {
+		it('returns the default port if $AGONES_SDK_GRPC_PORT is not defined', async () => {
+			let port = agonesSDK.port;
+			expect(port).toEqual('59357');
+		});
+
+		it('returns the default port if $AGONES_SDK_GRPC_PORT is not an integer', async () => {
+			process.env.AGONES_SDK_GRPC_PORT = "random string";
+			let port = agonesSDK.port;
+			expect(port).toEqual('59357');
+		});
+
+		it('returns the default port if $AGONES_SDK_GRPC_PORT is to large of an integer', async () => {
+			process.env.AGONES_SDK_GRPC_PORT = '4455667788';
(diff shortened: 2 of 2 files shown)


sdks/nodejs/src/agonesSDK.js
```diff
@@ -19,7 +19,26 @@ const services = require('../lib/sdk_grpc_pb');
 
 class AgonesSDK {
 	constructor() {
-		this.client = new services.SDKClient('localhost:59357', grpc.credentials.createInsecure());
+		function port() {
```
steven-supersolid: I would consider moving this to a method to make it easier to test.
Either way we should add a test to maintain code coverage

sdks/nodejs/src/agonesSDK.js
```diff
@@ -19,7 +19,26 @@ const services = require('../lib/sdk_grpc_pb');
...
+			const defaultPort = 59357;
+			const port = process.env.AGONES_SDK_GRPC_PORT;
+			if (port === undefined) {
+				console.log("Environment variable AGONES_SDK_GRPC_PORT not defined, using default port ${defaultPort}");
```
steven-supersolid: I think the linter should fix the quotes here to avoid having mixed quotes. If you run `npm run lint` that should do the trick.
It would be worth thinking about running lint and tests/coverage automatically

sdks/nodejs/src/agonesSDK.js
```diff
@@ -19,7 +19,26 @@ const services = require('../lib/sdk_grpc_pb');
...
+				return defaultPort;
+			}
+			const portNum = parseInt(port, 10);
+			if (isNaN(portNum)) {
+				console.log("Unable to parse ${port} defined in AGONES_SDK_GRPC_PORT into an integer\n");
```
steven-supersolid: Nit: No \n required in console.log as it already includes a newline

sdks/nodejs/src/agonesSDK.js
steven-supersolid: Actually I think you may have intended to use a backquote here in order to use variable substitution

sdks/nodejs/src/agonesSDK.js
roberthbailey: I've moved it to a separate method. I thought doing it inline was cute since it reduced scope, but maintaining test coverage is definitely more important.

sdks/nodejs/src/agonesSDK.js
roberthbailey: I've switched to backticks and also added single quotes around the string to make it stand out more in the console output.

sdks/nodejs/src/agonesSDK.js
roberthbailey: Thanks for catching this (I'd fixed it in one of the lines but not the other two).

roberthbailey:
The build failed fetching kubectl:

```
The command '/bin/sh -c curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VER}/bin/linux/amd64/kubectl &&     chmod go+rx ./kubectl &&     mv ./kubectl /usr/local/bin/kubectl' returned a non-zero code: 56
```

Restarting it.

roberthbailey:
@steven-supersolid - not sure when you are going to be back, but I'm hoping we can get this one in prior to the 1.1 release.

sdks/nodejs/src/agonesSDK.js
```diff
@@ -19,11 +19,30 @@ const services = require('../lib/sdk_grpc_pb');
 
 class AgonesSDK {
 	constructor() {
-		this.client = new services.SDKClient('localhost:59357', grpc.credentials.createInsecure());
+		this.client = new services.SDKClient('localhost:'+this.port(), grpc.credentials.createInsecure());
 		this.healthStream = undefined;
 		this.emitters = [];
 	}
 
+	port() {
```
steven-supersolid: Minor but I tend to use a name like `getPort()` for this or you could convert the method to a getter `get port()` which would then be used as `this.port`. The current name implies to me that an action called `port()` is being performed.

steven-supersolid:
Minor comment and perhaps down to personal preference but LGTM apart from that!

sdks/nodejs/src/agonesSDK.js
roberthbailey: Thanks for the advice. I've changed it to a getter and updated the tests as well.

roberthbailey:
/assign @pooneh-m 

Can you take a look as an approver given the LGTM from @steven-supersolid above? Thanks!

sdks/nodejs/src/agonesSDK.js
```diff
@@ -19,11 +19,30 @@ const services = require('../lib/sdk_grpc_pb');
...
 class AgonesSDK {
 	constructor() {
-		this.client = new services.SDKClient('localhost:59357', grpc.credentials.createInsecure());
+		this.client = new services.SDKClient('localhost:'+this.port, grpc.credentials.createInsecure());
 		this.healthStream = undefined;
 		this.emitters = [];
 	}
 
+	get port() {
+		const defaultPort = '59357';
```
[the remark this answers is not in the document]
roberthbailey: The default port is left in for backwards compatibility. If you use a newer SDK (1.1+) with an older version of Agones (1.0), then the agones controller will not set the environment variable and it would be nice if the sidecar didn't crash. 

We can go back through at some point and remove the defaults, but I think for now we need to leave them in.

PR #1095, opened 2019-10-03
Use non-ephemeral port for Go SDK conformance test

aLekSer:
Add `PORT` parameter for SDK conformance tests.
Local SDK server could use different ports as well as SDK client.
First step to run SDK tests in parallel.


aLekSer:
/assign @roberthbailey 
This is an initial version which uses ENV variable to switch to SDK client to another port. Will use different ports for Node JS and Rust once they would be merged.

build/includes/sdk.mk
```diff
@@ -128,21 +128,23 @@ run-sdk-conformance-no-build: TIMEOUT ?= 30
 run-sdk-conformance-no-build: RANDOM := $(shell bash -c 'echo $$RANDOM')
 run-sdk-conformance-no-build: DELAY ?= $(shell bash -c "echo $$[ ($(RANDOM) % 5 ) + 1 ]")
 run-sdk-conformance-no-build: TESTS ?= ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch,reserve
+run-sdk-conformance-no-build: PORT ?= 59357
 run-sdk-conformance-no-build: ensure-agones-sdk-image
 run-sdk-conformance-no-build: ensure-build-sdk-image
-	DOCKER_RUN_ARGS="--network=host $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
-	docker run -p 59357:59357 -e "ADDRESS=" -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" -e "DELAY=$(DELAY)" \
-	 --net=host  $(sidecar_tag)
+	DOCKER_RUN_ARGS="--net host -e AGONES_SDK_GRPC_PORT=$(PORT) $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
```
roberthbailey: Right now all of the sdks with conformance tests use the grpc port, but the unity sdk (at least) uses the http port. I think @markmandel mentioned that you might have been working on some rest conformance tests as well, which would use the http port. Is there a way you are thinking that we could extend this to work with both the grpc and http ports?

build/includes/sdk.mk
```diff
@@ -128,21 +128,23 @@ run-sdk-conformance-no-build: TIMEOUT ?= 30
...
+	docker run -p $(PORT):59357 -e "ADDRESS=" -e "TEST=$(TESTS)" -e "TIMEOUT=$(TIMEOUT)" -e "DELAY=$(DELAY)" \
```
roberthbailey: So this maps the configured port on the machine to the default port in the container. I think it would be a bit better to do `$(PORT):$(PORT)` and then pass `--grpc-port=$(PORT)` to the sidecar container so that we are testing having the sdkserver bind to the configured port as well.

aLekSer: Yes, that's true, performing this change

build/includes/sdk.mk
aLekSer: I have added a separate PR for Rest HTTP port testing:
 https://github.com/googleforgames/agones/pull/1079

cmd/sdk-server/Dockerfile
```diff
@@ -21,4 +21,5 @@ COPY --chown=agones:root ./bin/sdk-server.linux.amd64 /home/agones/sdk-server
 COPY --chown=agones:root ./bin/LICENSES ./bin/dependencies-src.tgz /home/agones/
 
 USER agones
-ENTRYPOINT ["/home/agones/sdk-server"]
\ No newline at end of file
+ENTRYPOINT ["/home/agones/sdk-server"]
+CMD [ "--grpc-port", "59357" ]
```
roberthbailey: Does it work without this change? The binary already defaults this flag (in code) so it doesn't seem like we should also need to default it in the dockerfile.

aLekSer: Let me check your assumption.

aLekSer: No need for this `CMD` block, removed

aLekSer:
I need to update commit text accordingly

PR #1099, opened 2019-10-04
GameServer container restart before Ready, move to Unhealthy state After (v2)

### LINKED ISSUE #956: Game server container crash before Ready, should restart, not move to Unhealthy

markmandel:
This brings our implementation inline with what our health checking documentation states that we do.

This is done by implementing extra checks in the HealthController to determine if it's appropriate to move to Unhealthy rather than allow a restart to occur.

Replaced PR #1069

Closes #956

### DIFF
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -91,6 +91,10 @@ const (
 	// DevAddressAnnotation is an annotation to indicate that a GameServer hosted outside of Agones.
 	// A locally hosted GameServer is not managed by Agones it is just simply registered.
 	DevAddressAnnotation = "agones.dev/dev-address"
+	// GameServerReadyContainerIDAnnotation is an annotation that is set on the GameServer
+	// becomes ready, so we can track when restarts should occur and when a GameServer
+	// should be moved to Unhealthy.
+	GameServerReadyContainerIDAnnotation = agones.GroupName + "/ready-container-id"
 )
 
 var (
@@ -423,6 +427,25 @@ func (gs *GameServer) IsDeletable() bool {
 // IsBeingDeleted returns true if the server is in the process of being deleted.
 func (gs *GameServer) IsBeingDeleted() bool {
 	return !gs.ObjectMeta.DeletionTimestamp.IsZero() || gs.Status.State == GameServerStateShutdown
+}
+
+// IsBeforeReady returns true if the GameServer Status has yet to move to or past the Ready
+// state in its lifecycle, such as Allocated or Reserved, or any of the Error/Unhealthy states
+func (gs *GameServer) IsBeforeReady() bool {
+	switch gs.Status.State {
+	case GameServerStatePortAllocation:
+		return true
+	case GameServerStateCreating:
+		return true
+	case GameServerStateStarting:
+		return true
+	case GameServerStateScheduled:
+		return true
+	case GameServerStateRequestReady:
+		return true
+	}
+
+	return false
 }
 
 // FindGameServerContainer returns the container that is specified in
--- a/pkg/gameservers/health.go
+++ b/pkg/gameservers/health.go
@@ -190,6 +190,10 @@ func (hc *HealthController) syncGameServer(key string) error {
 		return nil
 	}
 
+	if skip, err := hc.skipUnhealthy(gs); err != nil || skip {
+		return err
+	}
+
 	hc.loggerForGameServer(gs).Info("Issue with GameServer pod, marking as GameServerStateUnhealthy")
 	gsCopy := gs.DeepCopy()
 	gsCopy.Status.State = agonesv1.GameServerStateUnhealthy
@@ -201,4 +205,48 @@ func (hc *HealthController) syncGameServer(key string) error {
 	hc.recorder.Event(gs, corev1.EventTypeWarning, string(gsCopy.Status.State), "Issue with Gameserver pod")
 
 	return nil
+}
+
+// skipUnhealthy determines if it's appropriate to not move to Unhealthy when a Pod's
+// gameserver container has crashed, or let it restart as per usual K8s operations.
+// It does this by checking a combination of the current GameServer state and annotation data that stores
+// which container instance was live if the GameServer has been marked as Ready.
+// The logic is as follows:
+//   - If the GameServer is not yet Ready, allow to restart (return true)
+//   - If the GameServer is in a state past Ready, move to Unhealthy
+func (hc *HealthController) skipUnhealthy(gs *agonesv1.GameServer) (bool, error) {
+	pod, err := hc.podLister.Pods(gs.ObjectMeta.Namespace).Get(gs.ObjectMeta.Name)
+	if err != nil {
+		// Pod doesn't exist, so the GameServer is definitely not healthy
+		if k8serrors.IsNotFound(err) {
+			return false, nil
+		}
+		// if it's something else, go back into the queue
+		return false, errors.Wrapf(err, "error retrieving Pod %s for GameServer to check status", gs.ObjectMeta.Name)
+	}
+	if !metav1.IsControlledBy(pod, gs) {
+		// This is not the Pod we are looking for 🤖
+		return false, nil
+	}
+	if gs.IsBeforeReady() {
+		return hc.failedContainer(pod), nil
+	}
+
+	// finally, we need to check if there a failed container happened after the gameserver was ready or before.
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == gs.Spec.Container {
+			if cs.State.Terminated != nil {
+				return false, nil
+			}
+			if cs.LastTerminationState.Terminated != nil {
+				// if the current container is running, and is the ready container, then we know this is some
+				// other pod update, and we previously had a restart before we got to being Ready, and therefore
+				// shouldn't move to Unhealthy.
+				return cs.ContainerID == gs.Annotations[agonesv1.GameServerReadyContainerIDAnnotation], nil
+			}
+			break
+		}
+	}
+
+	return false, nil
 }
--- a/pkg/gameservers/health_test.go
+++ b/pkg/gameservers/health_test.go
@@ -79,6 +79,112 @@ func TestHealthUnschedulableWithNoFreePorts(t *testing.T) {
 	assert.False(t, hc.unschedulableWithNoFreePorts(pod))
 }
 
+func TestHealthControllerSkipUnhealthy(t *testing.T) {
+	t.Parallel()
+
+	fixtures := map[string]struct {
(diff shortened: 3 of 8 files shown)


markmandel:
/hold

e2e test found a bug in my implementation. Need to fix!

roberthbailey:
I'm waiting for @markmandel to remove the hold before I review.

markmandel:
/hold cancel 

While I think this should stay in feature freeze, as I would like to go through a development cycle for extra confidence, it is ready for review again now.

markmandel:
OMG! We can finally get this reviewed an in a release! :dancer: I feel like I've been working on this bug forever :smile:

roberthbailey:
Now that the release is out this is the next review up on my list.

test/e2e/gameserver_test.go
```diff
@@ -213,6 +213,113 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
+		// create a connection each time, as weird stuff happens if the receiver isn't up and running.
+		conn, err := net.Dial("udp", address)
+		assert.NoError(t, err)
+		defer conn.Close() // nolint: errcheck
+		// doing this last, so that there is a short delay between the message being sent, and the check.
+		_, err = conn.Write([]byte("READY"))
+		logger.WithError(err).WithField("gs", newGs.ObjectMeta.Name).WithField("state", newGs.Status.State).Info("no ready, trying again")
+		return false, nil
+	})
+	if !assert.NoError(t, err) {
```
roberthbailey: elsewhere I just see `assert.NoError()` but this one wraps that and calls `FailNow`. Is that to short circuit running the remainder of the test? If so, would it be simpler to just write

```go
if err != nil {
    assert.FailNow(t, ...)
}
```

I find the statements that check the boolean result of the assert to be hard to read, and they are basically just checking if err is or isn't nil and calling another assert, which can be done more simply.

test/e2e/gameserver_test.go
```diff
@@ -213,6 +213,113 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
+
+		// create a connection each time, as weird stuff happens if the receiver isn't up and running.
+		conn, err := net.Dial("udp", address)
+		assert.NoError(t, err)
+		defer conn.Close() // nolint: errcheck
+		// doing this last, so that there is a short delay between the message being sent, and the check.
+		// also gives the process a chance to come back up
+		_, err = conn.Write([]byte("CRASH"))
+		logger.WithError(err).Info("no crash, trying again")
+		return false, nil
```
roberthbailey: should this return `false, err` instead of nil? Or will returning an error short circuit the polling function? It seems weird that the error writing the udp packet is logged but otherwise ignored.

test/e2e/gameserver_test.go
```diff
@@ -213,6 +213,113 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
+	err = wait.Poll(time.Second, 3*time.Minute, func() (bool, error) {
+		conn, err := net.Dial("udp", address)
+		assert.NoError(t, err)
+		defer conn.Close() // nolint: errcheck
+		_, err = conn.Write([]byte("CRASH"))
+		if err != nil {
+			logger.WithError(err).Warn("could not send CRASH udp packet")
+			return false, nil
+		}
+		return true, nil
```
roberthbailey: Why is successfully sending the packet sufficient to break the loop? Shouldn't we verify that the pod actually exited (e.g. the game server received the packet)?

test/e2e/gameserver_test.go
```diff
@@ -213,6 +213,113 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
...
+		logger.WithError(err).WithField("gs", newGs.ObjectMeta.Name).WithField("state", newGs.Status.State).Info("no ready, trying again")
+		return false, nil
+	})
+	if !assert.NoError(t, err) {
+		assert.FailNow(t, "Could not make GameServer Ready")
+	}
+	// now crash, should be unhealthy, since it's after being Ready
+	logger.Info("crashing again, should be unhealthy")
+	// retry on crash, as with the restarts, sometimes Go takes a moment to send this through.
+	err = wait.Poll(time.Second, 3*time.Minute, func() (bool, error) {
```
roberthbailey: All three of these udp sending loops are very similar in structure (write a packet, check for state, try again). Have you considered moving them into a helper function?

pkg/gameservers/health.go
```diff
@@ -190,15 +190,61 @@ func (hc *HealthController) syncGameServer(key string) error {
...
-
-	if _, err := hc.gameServerGetter.GameServers(gs.ObjectMeta.Namespace).Update(gsCopy); err != nil {
-		return errors.Wrapf(err, "error updating GameServer %s to unhealthy", gs.ObjectMeta.Name)
+	skip, err := hc.skipUnhealthy(gs)
+	if err != nil {
+		return err
 	}
 
-	hc.recorder.Event(gs, corev1.EventTypeWarning, string(gsCopy.Status.State), "Issue with Gameserver pod")
+	if !skip {
```
roberthbailey: This should be inverted. 

```
if skip {
  // The game server hasn't been marked ready, so it doesn't transition to unhealthy. 
  return nil
}
```

To leave the other code with one less level of indentation.

pkg/gameservers/health.go
```diff
@@ -190,15 +190,61 @@ func (hc *HealthController) syncGameServer(key string) error {
...
+			return errors.Wrapf(err, "error updating GameServer %s to unhealthy", gs.ObjectMeta.Name)
+		}
+
+		hc.recorder.Event(gs, corev1.EventTypeWarning, string(gsCopy.Status.State), "Issue with Gameserver pod")
+	}
 
 	return nil
 }
+
+// skipUnhealthy checks the GameServer is Ready or after Ready, and what the container crash state
```
roberthbailey: This comment is really hard to read. I tried taking a stab at rewriting it but didn't have any particularly great ways to phrase it. If you can think of a way to say it more clearly that'd be awesome.

pkg/gameservers/health_test.go
```diff
@@ -79,15 +79,114 @@ func TestHealthUnschedulableWithNoFreePorts(t *testing.T) {
...
+			hc := NewHealthController(healthcheck.NewHandler(), m.KubeClient, m.AgonesClient, m.KubeInformerFactory, m.AgonesInformerFactory)
+			gs := &agonesv1.GameServer{ObjectMeta: metav1.ObjectMeta{Name: "test", Namespace: defaultNs}, Spec: newSingleContainerSpec()}
+			gs.ApplyDefaults()
+			pod, err := gs.Pod()
+			assert.NoError(t, err)
+
+			v.setup(gs, pod)
+
+			m.KubeClient.AddReactor("list", "pods", func(action k8stesting.Action) (bool, runtime.Object, error) {
+				return true, &corev1.PodList{Items: []corev1.Pod{*pod}}, nil
```
roberthbailey: It looks like this doesn't cover the branches where pod doesn't exist or k8s returns an error to put the request back into the queue.

pkg/gameservers/health_test.go
```diff
@@ -79,15 +79,114 @@ func TestHealthUnschedulableWithNoFreePorts(t *testing.T) {
...
+		expected bool
+	}{
+		"scheduled and terminated container": {
+			setup: func(gs *agonesv1.GameServer, pod *corev1.Pod) {
+				gs.Status.State = agonesv1.GameServerStateScheduled
+				pod.Status.ContainerStatuses = []corev1.ContainerStatus{{
+					Name:  gs.Spec.Container,
+					State: corev1.ContainerState{Terminated: &corev1.ContainerStateTerminated{}},
+				}}
+			}, expected: true},
```
roberthbailey: please put expected on a new line

pkg/gameservers/health_test.go
```diff
@@ -79,15 +79,114 @@ func TestHealthUnschedulableWithNoFreePorts(t *testing.T) {
...
+				pod.Status.ContainerStatuses = []corev1.ContainerStatus{{
+					Name:  gs.Spec.Container,
+					State: corev1.ContainerState{Terminated: &corev1.ContainerStateTerminated{}},
+				}}
+			},
+			expected: false,
+		},
+		"before ready, with no terminated container": {
+			setup: func(gs *agonesv1.GameServer, pod *corev1.Pod) {
+				gs.Status.State = agonesv1.GameServerStateScheduled
```
roberthbailey: the before ready states are all the same, given the other test coverage it may not matter but you could vary which state is set to one of the other 4 states that are before ready.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -91,6 +91,10 @@ const (
 	// DevAddressAnnotation is an annotation to indicate that a GameServer hosted outside of Agones.
 	// A locally hosted GameServer is not managed by Agones it is just simply registered.
 	DevAddressAnnotation = "agones.dev/dev-address"
```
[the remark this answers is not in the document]
markmandel: Good catch, we should fix this.

test/e2e/gameserver_test.go
markmandel: You are indeed correct - returning the error will short circuit the poll. It may return an error as the receiver may not be up and running yet, so it's useful for debugging purposes, but otherwise it is expected.

test/e2e/gameserver_test.go
markmandel: Nice. Done!

test/e2e/gameserver_test.go
markmandel: Fixed, now we have a helper method.

pkg/gameservers/health.go
markmandel: Ah excellent suggestion!

pkg/gameservers/health.go
markmandel: Took a stab. PTAL when I push my changes up. I think it's better though.

markmandel:
@roberthbailey this should be good to review again :+1:

markmandel:
dancing!

PR #1110, opened 2019-10-14
Move allocation proto to root level proto

pooneh-m:
This is to address a comment in https://github.com/googleforgames/agones/pull/1041


roberthbailey:
Do we also need to update any makefiles to reference the new location? Or are all of the references in https://github.com/googleforgames/agones/pull/1041?

pooneh-m:
> Do we also need to update any makefiles to reference the new location? Or are all of the references in #1041?

All the references are in #1041.

Issue #1112, opened 2019-10-14
Please create a root OWNERS file for agones


fejta:
Can you create OWNERS files? At least one in the root dir.

/assign @roberthbailey 

```yaml
{
 insertId: "x0pnb4f8jk0fp"  
 jsonPayload: {
  author: "pooneh-m"   
  component: "hook"   
  event-GUID: "7a7a5af0-eeae-11e9-92e0-9607e4f41e57"   
  event-type: "pull_request"   
  file: "prow/plugins/approve/approvers/owners.go:164"   
  func: "k8s.io/test-infra/prow/plugins/approve/approvers.Owners.GetSuggestedApprovers"   
  level: "warning"   
  msg: "Couldn't find/suggest approvers for each files. Unapproved: [""]"   
  org: "googleforgames"   
  plugin: "approve"   
  pr: 1110   
  repo: "agones"   
  url: "https://github.com/googleforgames/agones/pull/1110"   
 }
 labels: {…}  
 logName: "projects/oss-prow/logs/hook"  
 receiveTimestamp: "2019-10-14T18:14:48.747545067Z"  
 resource: {…}  
 severity: "ERROR"  
 timestamp: "2019-10-14T18:14:41Z"  
}
```

roberthbailey:
I could have sworn that I did this.... I looked through my PR history and found https://github.com/googleforgames/agones/pull/1016, where I apparently created an owners file in the build directory (even though the PR title says "at the root of the repository"). Oops.

PR #1113, opened 2019-10-14
Move the owners file to the root of the repository.

roberthbailey:
Fixes https://github.com/googleforgames/agones/issues/1112


OWNERS
```diff
@@ -4,10 +4,11 @@ approvers:
...
   - dzlier-gcp
   - EricFortin
-  - ilkercelikyilmaz
-  - jkowalski
-  - joeholley
   - markmandel
   - pooneh-m
   - roberthbailey
 
+emeritus_approvers:
```
pooneh-m: Does this mean this group cannot approve anymore or will not be suggested as reviewer?

roberthbailey: Yes (to both). See https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#emeritus.

pooneh-m: Thanks. Does it make sense that the same set of people who their role is being changed to emeritus_approvers, approve the change or if not responding then someone else approve? 

The same [link](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#emeritus) says:

>  **They** are encouraged to add **themselves** as an "emeritus" approver under the emeritus_approvers key.

pooneh-m:
> Also move Googlers to emeritus_approvers who aren't actively doing code reviews.

Can we separate the two PRs for emeritus_approvers and moving the file to root directory? I am fine with the file move, but prefer the impacted approvers to approve being moved to emeritus_approvers.

roberthbailey:
I removed the second commit and updated the PR text.

roberthbailey:
The second commit is now a separate PR (https://github.com/googleforgames/agones/pull/1116).

PR #1116, opened 2019-10-15
Move Googlers no longer actively reviewing code to emeritus owners

roberthbailey:
Split out of https://github.com/googleforgames/agones/pull/1113.

/assign @jkowalski @joeholley @ilkercelikyilmaz

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -4,10 +4,11 @@ approvers:
   - cyriltovena
   - dzlier-gcp
   - EricFortin
-  - ilkercelikyilmaz
-  - jkowalski
-  - joeholley
   - markmandel
   - pooneh-m
   - roberthbailey
 
+emeritus_approvers:
+  - joeholley
+  - jkowalski
+  - ilkercelikyilmaz


roberthbailey:
> What will be the process to become an active reviewer? Modify this file?
> I assume we can't approve any change while we are emeritus_approver, right?

Right now, the owners file is really only used by prow to auto-assign PRs to reviewers. Being moved out of the approvers section means that you won't get PRs assigned to you. 

If/when we switch to using prow to automatically merge PRs (rather than click the button) then not being an approver means you can't approve PRs for prow to merge them. 

In the current state, you will still be an owner of the repo, still be able to approve PRs, and still be able to manually merge them. You won't lose any permissions (yet). 

But you bring up a good question which is that we have a process for promoting folks to be project approvers while we don't have a written policy about the process for becoming emeritus and/or regaining approver powers. 

https://github.com/googleforgames/agones/blob/master/docs/governance/community_membership.md#responsibilities-and-privileges says that we expect approvers to "be responsive to review requests in a timely manner". If folks aren't actively reviewing PRs, then I think it's reasonable to say that they are not acting an a project approver any longer.

roberthbailey:
@joeholley / @jkowalski - any thoughts?

PR #1123, opened 2019-10-17
Change the API for allocator to allocation.proto

pooneh-m:


roberthbailey:
I'm marking this as a breaking change, since it's changing the API surface for the allocator. Since this is still an alpha feature that's ok, but I think it'll help craft the release notes for 1.1 (so that anyone using this feature will be made aware).

PR #1124, opened 2019-10-17
Drop the Extension API Server reference from agones-allocator

pooneh-m:
agones-allocator is referencing allocation extension API and this change is removing the dependency by referencing the allocator lib directly. 

Notes:
- ClusterRole for agones-allocator is now expended to access secrets, nodes and game servers
- The node game server counter is not synced with the counter used by game server controller and therefore the packed scheduling is not as accurate. (At this point packed or redistributed scheduling is not applicable for multi-cluster allocation as the destination cluster may not need the same scheduling setting as the origin cluster.)

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -24,13 +24,19 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"agones.dev/agones/pkg"
 	"agones.dev/agones/pkg/allocation/converters"
 	pb "agones.dev/agones/pkg/allocation/go/v1alpha1"
+	allocationv1 "agones.dev/agones/pkg/apis/allocation/v1"
 	"agones.dev/agones/pkg/client/clientset/versioned"
+	"agones.dev/agones/pkg/client/informers/externalversions"
+	"agones.dev/agones/pkg/gameserverallocations"
+	"agones.dev/agones/pkg/gameservers"
 	"agones.dev/agones/pkg/metrics"
 	"agones.dev/agones/pkg/util/runtime"
+	"agones.dev/agones/pkg/util/signals"
 	"github.com/heptiolabs/healthcheck"
 	prom "github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


cmd/allocator/main.go
```diff
@@ -193,18 +237,32 @@ func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
 	logger.WithField("request", request).Infof("allocation request received")
 
 	gsa := converters.ConvertAllocationRequestV1Alpha1ToGSAV1(&request)
-	allocation := h.agonesClient.AllocationV1().GameServerAllocations(gsa.ObjectMeta.Namespace)
-	allocatedGsa, err := allocation.Create(gsa)
+	resultObj, err := h.allocationCallback(gsa, h.stop)
```
roberthbailey: Why is there a callback defined instead of just calling `h.allocator.Allocate(gsa, stop)` here?

cmd/allocator/main.go
```diff
@@ -193,18 +237,32 @@ func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		http.Error(w, err.Error(), httpCode(err))
 		logger.WithField("gsa", gsa).WithError(err).Info("calling allocation extension API failed")
```
roberthbailey: This error message should be updated.

cmd/allocator/main.go
```diff
@@ -193,18 +237,32 @@ func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
+	w.Header().Set("Content-Type", "application/json")
+	if status, ok := resultObj.(*metav1.Status); ok {
+		w.WriteHeader(int(status.Code))
+		err = json.NewEncoder(w).Encode(status)
+		if err != nil {
+			http.Error(w, "internal server error", http.StatusInternalServerError)
+			logger.Error(err)
```
roberthbailey: Maybe add some context to the log, e.g. `logger.WithError(err).Error("Unable to encode status in json")`

cmd/allocator/main.go
roberthbailey: It looks like it's necessary for testing.

cmd/allocator/main.go
```diff
@@ -180,7 +222,9 @@ func (h *httpHandler) postOnly(in handler) handler {
 }
 
 type httpHandler struct {
-	agonesClient versioned.Interface
+	allocator          *gameserverallocations.Allocator
```
roberthbailey: I don't see that the allocator or stop fields are used anywhere. Why are they part of the struct instead of being defined locally in the `newServiceHandler` function?

PR #1122, opened 2019-10-16
Faster subsequent Rust SDK conformance builds

aLekSer:
Add save-cache target for Rust. Fixed Rust SDK test subsequent call was downloading and rebuilding all dependencies.

PR #1125, opened 2019-10-21
Added dynamic configuration of http port in the rust sdk

roberthbailey:
/assign @markmandel 

Part of #851.

### DIFF
--- a/sdks/rust/src/sdk.rs
+++ b/sdks/rust/src/sdk.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


roberthbailey:
@aLekSer - it just occurred to me that because of https://github.com/googleforgames/agones/pull/1122 we might need to bump a number to get this into the cache for e2e tests....

aLekSer:
Looking into this issue. As you can see the first failed test run used the current version of Rust code, I have seen this in my PR also. So bumping the number is not always required. It only is needed if we are updating `Config.toml` with new dependencies.

PR #1128, opened 2019-10-21
Simplify the selection of the dynamic port in the nodejs sdk.

roberthbailey:
This brings the NodeJS SDK in line with the C++ / Unity / Rust SDKs.

Part of #851.

### DIFF
--- a/sdks/nodejs/src/agonesSDK.js
+++ b/sdks/nodejs/src/agonesSDK.js
@@ -25,22 +25,8 @@ class AgonesSDK {
 	}
 
 	get port() {
-		const defaultPort = '59357';
 		const port = process.env.AGONES_SDK_GRPC_PORT;
-		if (port === undefined) {
-			console.log(`Environment variable AGONES_SDK_GRPC_PORT not defined, using default port ${defaultPort}`);
-			return defaultPort;
-		}
-		const portNum = parseInt(port, 10);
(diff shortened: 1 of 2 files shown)


sdks/nodejs/src/agonesSDK.js
```diff
@@ -25,22 +25,8 @@ class AgonesSDK {
 	}
 
 	get port() {
-		const defaultPort = '59357';
 		const port = process.env.AGONES_SDK_GRPC_PORT;
```
steven-supersolid: In JavaScript it's common to use `||` e.g. do `return process.env.AGONES_SDK_GRPC_PORT || '59357'`
Without examining grpc code too much it seems port should be a number so you could cast using `return Number(process.env.AGONES_SDK_GRPC_PORT) || 59357`

steven-supersolid: doh too slow :)

roberthbailey: I'm happy to fix it in a follow-up PR. :)

roberthbailey: https://github.com/googleforgames/agones/pull/1129

> it seems port should be a number

The value is returned and concatenated with 'localhost' so I think returning a string is ok.

PR #1131, opened 2019-10-22
Add dynamic configuration of the sidecar http port to the unreal sdk.

roberthbailey:
So I've been banging my head against UE4 today. The new code certainly does the right thing when the environment variable isn't set. But I can't convince my system to set the environment so that UE4 will actually see any values when the game is run. 

It turns out that cross compilation only works on Windows, so I can't compile a game to Linux to try and run it in a docker container (which is how I've tested setting environment variables for the other SDKs). And no matter what tricks I try, it always returns empty string when the example game I'm using is run on my mac. 

So... this _might_ work. Based on the documentation, it _should_ work, but without being able to actually test it I can't be sure if it does. 

Part of #851.

### DIFF
--- a/site/content/en/docs/Guides/Client SDKs/unreal.md
+++ b/site/content/en/docs/Guides/Client SDKs/unreal.md
@@ -34,9 +34,6 @@ Available settings:
 {{% feature expiryVersion="1.1.0" %}}
 - Agones Sidecar IP. (default: `http://localhost:59358`)
 {{% /feature %}}
-{{% feature publishVersion="1.1.0" %}}
-- Agones Sidecar IP. (default: `http://localhost:${AGONES_SDK_HTTP_PORT}`)
-{{% /feature %}}
 - Health Ping Enabled. Whether the server sends a health ping to the Agones sidecar. (default: `true`)
 - Health Ping Seconds. Interval of the server sending a health ping to the Agones sidecar. (default: `5`)
 - Debug Logging Enabled. Debug logging for development of this Plugin. (default: `false`)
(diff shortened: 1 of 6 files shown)


site/content/en/docs/Guides/Client SDKs/unreal.md
```diff
@@ -34,9 +34,6 @@ Available settings:
 {{% feature expiryVersion="1.1.0" %}}
 - Agones Sidecar IP. (default: `http://localhost:59358`)
 {{% /feature %}}
-{{% feature publishVersion="1.1.0" %}}
-- Agones Sidecar IP. (default: `http://localhost:${AGONES_SDK_HTTP_PORT}`)
```
pooneh-m: Why is this removed?

pooneh-m:
Please also get a second approval from someone familiar with unreal SDK. The change for setting default port looks good to me.

roberthbailey:
I figured out why I couldn't get this to work yesterday.

`FGenericPlatformMisc::GetEnvironmentVariable(...)` always returns an empty string (`return FString();`) rather than reading from the environment. Changing it to `FPlatformMisc::GetEnvironmentVariable(...)` actually reads the environment and the plugin now does the right thing.

I've been able to test locally that I connect to different ports by setting different environment variables in a sample game so I'm much more confident about this change now.

site/content/en/docs/Guides/Client SDKs/unreal.md
roberthbailey: Because I've removed this as a setting in the unreal editor. The code no longer uses the setting (and the setting no longer exists) so it should not be documented. 

I've previously changed it to reference the environment variable (even though I hadn't updated the code) but as I looked at the code I realized it didn't make sense as a setting.

PR #1132, opened 2019-10-22
improve stackdriver metric type

cyriltovena:
As per the documentation https://godoc.org/contrib.go.opencensus.io/exporter/stackdriver

This will improve the metric typing and should now appears as a gke container in stackdriver interface.

If someone has time to test it and let me know if everything is correct, that would be awesome.

### DIFF
--- a/pkg/metrics/exporter.go
+++ b/pkg/metrics/exporter.go
@@ -21,13 +21,12 @@ import (
 
 	"cloud.google.com/go/compute/metadata"
 	"contrib.go.opencensus.io/exporter/stackdriver"
+	"github.com/pkg/errors"
 	prom "github.com/prometheus/client_golang/prometheus"
 	"go.opencensus.io/exporter/prometheus"
 	"go.opencensus.io/stats/view"
 	"google.golang.org/genproto/googleapis/api/monitoredres"
 )
-
-const unknown = "unknown"
 
 // RegisterPrometheusExporter register a prometheus exporter to OpenCensus with a given prometheus metric registry.
 // It will automatically add go runtime and process metrics using default prometheus collectors.
@@ -54,22 +53,10 @@ func RegisterPrometheusExporter(registry *prom.Registry) (http.Handler, error) {
 // RegisterStackdriverExporter register a Stackdriver exporter to OpenCensus.
 // It will add Agones metrics into Stackdriver on Google Cloud.
 func RegisterStackdriverExporter(projectID string, defaultLabels string) (*stackdriver.Exporter, error) {
-	instanceID, err := metadata.InstanceID()
+	monitoredRes, err := getMonitoredResource(projectID)
 	if err != nil {
-		logger.WithError(err).Warn("error getting instance ID")
-		instanceID = unknown
+		logger.WithError(err).Warn("error discovering monitored resource")
 	}
-	zone, err := metadata.Zone()
-	if err != nil {
-		logger.WithError(err).Warn("error getting zone")
-		zone = unknown
-	}
-	clusterName, err := metadata.InstanceAttributeValue("cluster-name")
-	if err != nil {
-		logger.WithError(err).Warn("error getting cluster-name")
-		clusterName = unknown
-	}
-
 	labels, err := parseLabels(defaultLabels)
 	if err != nil {
 		return nil, err
@@ -78,21 +65,8 @@ func RegisterStackdriverExporter(projectID string, defaultLabels string) (*stack
 	sd, err := stackdriver.NewExporter(stackdriver.Options{
 		ProjectID: projectID,
 		// MetricPrefix helps uniquely identify your metrics.
-		MetricPrefix: "agones",
-		Resource: &monitoredres.MonitoredResource{
-			Type: "k8s_container",
-			Labels: map[string]string{
-				"project_id":   projectID,
-				"instance_id":  instanceID,
-				"zone":         zone,
-				"cluster_name": clusterName,
-
-				// See: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/
-				"namespace_id":   os.Getenv("POD_NAMESPACE"),
-				"pod_id":         os.Getenv("POD_NAME"),
-				"container_name": os.Getenv("CONTAINER_NAME"),
-			},
-		},
+		MetricPrefix:            "agones",
+		Resource:                monitoredRes,
 		DefaultMonitoringLabels: labels,
 	})
 	if err != nil {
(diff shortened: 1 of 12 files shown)


cyriltovena:
Why the tag approved?

markmandel:
I think prow approves since you are an approver? @roberthbailey ?

roberthbailey:
> Why the tag approved?

Yes, that is from prow. If you weren't an approver, the tag could be added by an approver typing `/approve` in a PR comment or by approving via the github code review flow. 

Right now it doesn't mean anything, but if we decide to switch to tide for automatic merges then it'll be required for a PR to be merged.

pkg/metrics/exporter.go
```diff
@@ -49,19 +54,48 @@ func RegisterPrometheusExporter(registry *prom.Registry) (http.Handler, error) {
 // RegisterStackdriverExporter register a Stackdriver exporter to OpenCensus.
 // It will add Agones metrics into Stackdriver on Google Cloud.
 func RegisterStackdriverExporter(projectID string) (sd *stackdriver.Exporter, err error) {
-	// Default project will be used
+	instanceID, err := metadata.InstanceID()
```
roberthbailey: Does this code only run on GCP?

If not, I wonder if we should allow a user to customize the instance / zone / cluster-name values via environment variables for aggregation purposes.

pkg/metrics/exporter.go
```diff
@@ -49,19 +54,48 @@ func RegisterPrometheusExporter(registry *prom.Registry) (http.Handler, error) {
...
+		},
 	})
 	if err != nil {
 		return
 	}
 
 	// Register it as a metrics exporter
 	view.RegisterExporter(sd)
-	return
+	return sd, nil
```
roberthbailey: It's confusing to have some return paths use the blank return and have others pass the return values. I think the function is long enough that you should always return the explicit values and remove the named return values from the function declaration but that's just my personal preference.

pkg/metrics/exporter.go
cyriltovena: No technically you can run this outside of GCP. However if you do you won't be able to correlate with existing GCP workload because `instance_id`,`zone` and `cluster_name` won't be valid which makes sense.

I think we should allow users to be able to set `DefaultMonitoringLabels *Labels` via configuration, so they can tag their custom workload, but it wasn't there already so that can be done in another PR.

pkg/metrics/exporter.go
cyriltovena: yes I agree, I wasn't also a fan of that naked return and hesitant to change it, now I'm convinced.

pkg/metrics/exporter.go
cyriltovena: ok let's do this.

pkg/metrics/exporter.go
```diff
@@ -49,19 +54,48 @@ func RegisterPrometheusExporter(registry *prom.Registry) (http.Handler, error) {
...
+		logger.WithError(err).Warn("error getting cluster-name")
+		clusterName = unknown
+	}
+
 	sd, err = stackdriver.NewExporter(stackdriver.Options{
 		ProjectID: projectID,
 		// MetricPrefix helps uniquely identify your metrics.
 		MetricPrefix: "agones",
+		Resource: &monitoredres.MonitoredResource{
+			Type: "container",
```
ano12ak: According to the Stackdriver document, I think "Type" should be "gke_container" or "k8s_container".
https://cloud.google.com/monitoring/custom-metrics/creating-metrics#monitored_resources_for_custom_metrics

(I tried to run this build then I got following error.
2019/11/06 12:11:41 Failed to export to Stackdriver: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Unrecognized resource type: container.: timeSeries[0-17])

cyriltovena: I used the driver documentation. But yeah I think you're right.

roberthbailey:
@cyriltovena - is this ready for another review pass?

pkg/metrics/exporter.go
cyriltovena: changed, let me know.

cyriltovena:
Yes !

pkg/metrics/util_test.go
```diff
@@ -291,3 +293,57 @@ agones_gameservers_node_count_count 3
 agones_nodes_count{empty="false"} 2
 agones_nodes_count{empty="true"} 1
 `
+
+func Test_parseLabels(t *testing.T) {
+	tests := []struct {
```
roberthbailey: Another way you could write this (to more easily explore edge cases that should create errors) is to have two sets of tests, one that should produce errors and another that should work. I think the set of tests that should work is relatively small (you have 3 right now, which mostly covers it). But the set of inputs that should cause an error should be made much longer. This is easier if it's more compact, e.g.

```
errorInputs = []string{
  "a=b,",
  "a=b,c",
  ...
}
```

since the want is always nil and the wantErr is always true for these you can collapse them into a shorter list form, and then expand the list.

What about testing inputs with spaces added in various places? Or extra equals signs? Should non-ascii characters be accepted? Does stack driver impose any constraints on what characters can be in a key? And so on.

pkg/metrics/exporter.go
```diff
@@ -48,20 +53,55 @@ func RegisterPrometheusExporter(registry *prom.Registry) (http.Handler, error) {
...
+
+	sd, err := stackdriver.NewExporter(stackdriver.Options{
 		ProjectID: projectID,
 		// MetricPrefix helps uniquely identify your metrics.
 		MetricPrefix: "agones",
+		Resource: &monitoredres.MonitoredResource{
+			Type: "k8s_container",
+			Labels: map[string]string{
+				"project_id":   projectID,
+				"instance_id":  instanceID,
```
roberthbailey: Do these have to be set? Or can they be left blank if we can't read from the metadata server?

markmandel:
```
site/docs/installation/upgrading/index.html
  Non-OK status: 404 --- site/docs/installation/upgrading/index.html --> https://helm.sh/docs/helm/#helm-upgrade
```

Rebase / Upgrade to master - should solve that problem.

markmandel:
@roberthbailey anything left on this before we can merge?

site/content/en/docs/Installation/helm.md
```diff
@@ -36,8 +36,12 @@ _We recommend to install Agones in its own namespaces (like `agones-system` as s
 you can use the helm `--namespace` parameter to specify a different namespace._
 
 When running in production, Agones should be scheduled on a dedicated pool of nodes, distinct from where Game Servers are scheduled for better isolation and resiliency. By default Agones prefers to be scheduled on nodes labeled with `agones.dev/agones-system=true` and tolerates node taint `agones.dev/agones-system=true:NoExecute`. If no dedicated nodes are available, Agones will
+<<<<<<< HEAD
```
roberthbailey: Need to resolve this merge conflict.

cyriltovena: Should be resolved.

markmandel: @roberthbailey is this good to go now?

(I'm on a push to close out PRs before EOY :smiley:)

pkg/metrics/exporter.go
```diff
@@ -48,20 +52,30 @@ func RegisterPrometheusExporter(registry *prom.Registry) (http.Handler, error) {
 
 // RegisterStackdriverExporter register a Stackdriver exporter to OpenCensus.
 // It will add Agones metrics into Stackdriver on Google Cloud.
-func RegisterStackdriverExporter(projectID string) (sd *stackdriver.Exporter, err error) {
-	// Default project will be used
-	sd, err = stackdriver.NewExporter(stackdriver.Options{
+func RegisterStackdriverExporter(projectID string, defaultLabels string) (*stackdriver.Exporter, error) {
+	monitoredRes, err := getMonitoredResource(projectID)
+	if err != nil {
+		logger.WithError(err).Warn("error discovering monitored resource")
```
roberthbailey: Is it ok if monitoredRes is nil later on or does this need to be followed by a `return nil, err` line?

cyriltovena: no it's ok this is a best effort.

cyriltovena: by default it is nil.

cyriltovena:
> New changes are detected. LGTM label has been removed.

just rebased buddy ! :)

PR #1138, opened 2019-10-23
Remove the svg version of the old logo.

roberthbailey:
/assign @markmandel


markmandel:
Nice catch - totally missed this file was still there!

/approve

Issue #1137, opened 2019-10-23
GKE: Update documentation + Automation to disable node automatic updates for gameserver node pools


markmandel:
```
Starting November 8, 2019, we will enable node autoupgrades by default for new clusters and node pools created with the Google Cloud Platform Console, gcloud command line (CLI), and the Google Kubernetes Engine (GKE) API. We are making this change to ensure your clusters have the most recent default Kubernetes version, bug fixes, and security patches.
```

Since this could break in-game play, we should have our examples and automation disable it by default, for at least the gameserver nodepools (maybe agones-system as well?)

This might be a good opportunity to update the Terraform scripts for GKE + Agones, and remove/replace the deployment manager scripts in the development tooling while we are at it.

roberthbailey:
November 8th is ~1 week after we cut the 1.1 release, so it would make sense to update the install instructions before the 1.1 release to address this.

roberthbailey:
> This might be a good opportunity to update the Terraform scripts for GKE + Agones, and remove/replace the deployment manager scripts in the development tooling while we are at it.

I'm not going to have time to tackle this part right now. I think it's ok if development clusters created via deployment manager have auto upgrades enabled until we can fix it, since I expect that they get recycled more frequently.

roberthbailey:
@aLekSer - do you have interest in replacing deployment manager with terraform?

aLekSer:
Yes, I can do that. However I think version with gcloud tool should stay/renamed.

aLekSer:
Currently producing a version of `make gcloud-test-cluster` and `make clean-gcloud-test-cluster` targets with Terraform (to get rid of deployment-manager).

markmandel:
This has been done!

PR #1141, opened 2019-10-23
Update the instructions for installing a GKE cluster with terraform to disable automatic node upgrades

roberthbailey:
This is the second part of #1137.

/assign @markmandel

### DIFF
--- a/build/cluster.tf
+++ b/build/cluster.tf
@@ -64,6 +64,10 @@ resource "google_container_cluster" "primary" {
     name       = "default"
     node_count = "${var.cluster["initialNodeCount"]}"
 
+    management {
+      auto_upgrade = false
+    }
+
     node_config {
       machine_type = "${var.cluster["machineType"]}"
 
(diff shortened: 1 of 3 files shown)


markmandel:
This looks good to me - @aLekSer can you check the terraform bits real quick, just because you know TF way better than I do.

aLekSer:
Now terraform config looks much better.
Approving these changes also.

markmandel:
@roberthbailey I'm leaning towards merging this after 1.1.0 stable, just because it changes actual code -- but what do you think? I could also be persuaded to merge this now?

roberthbailey:
> what do you think?

I'd lean towards doing it now, before the release cut. 

The terraform example always pulls the cluster.tf file from the master branch (`source = "git::https://github.com/googleforgames/agones.git//build/?ref=master"`) so even if you merge later, running the example will pick up the changes to disable node auto updates. 

Then what's missing is the documentation updates and the updates to the example. The documentation updates could be guarded behind the 1.2 release and the example changes would be visible in the master branch but not on the release branch.

PR #1142, opened 2019-10-24
Fix CPP SDK archive path - created separate one

### LINKED ISSUE #1136: CI: Build cache cpp-sdk-build contains unnecessary archives
Currently the following files are added to build cache for every new build:
```
agonessdk-1.0.0-b8432d1-linux-arch_64.tar.gz
```
**What happened**:

The moment it happened is here:
https://github.com/googleforgames/agones/blob/88f1d4eebf96584322efa2dfb981a672590a4bc0/sdks/cpp/Makefile#L38
I expect that  `package_path` (`.install`) would be used to store all the archives. And `.build` would contain only files needed for the build.

`du -sh ./sdks/cpp/.build`
`make archive`
`du -sh ./sdks/cpp/.build` - becomes significantly bigger but should not.

**Environment**:
- Agones version: 
- Install method (yaml/helm):
- Others:

aLekSer:
It was a problem with using cache and putting archive to the same folder
as cached path.

Closes #1136.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -17,6 +17,7 @@
 /build
 /cmd
 /sdks/cpp/.build
+/sdks/cpp/.archives
 .idea
 *.zip
 /release
--- a/build/includes/release.mk
+++ b/build/includes/release.mk
@@ -55,7 +55,7 @@ do-release:
 
 	$(MAKE) -j 4 build VERSION=$(RELEASE_VERSION) REGISTRY=$(release_registry) FULL_BUILD=1
 	cp $(agones_path)/cmd/sdk-server/bin/agonessdk-server-$(RELEASE_VERSION).zip $(agones_path)/release
-	cp $(agones_path)/sdks/cpp/.build/agonessdk-$(RELEASE_VERSION)-linux-arch_64.tar.gz $(agones_path)/release
+	cp $(agones_path)/sdks/cpp/.archives/agonessdk-$(RELEASE_VERSION)-linux-arch_64.tar.gz $(agones_path)/release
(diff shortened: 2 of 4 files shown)


markmandel:
Do you need review on this? I (and I think others) tend to ignore draft PRs unless specifically pinged, so wasn't sure if you did / didn't.

markmandel:
A few more places that need updating (although you did release.mk):
image

aLekSer:
Thank you Mark, I added one more directory `.packages`, with intention that `.build` would stay there for the same purpose. So not all of these occurrences should be removed. I will check when I got better internet. Cloudbuild.yaml could be left as here.

markmandel:
Sorry, I don't see .packages anywhere?

I think we should definitely add .archives to Dockerignore
Also cloudbuild.yaml will definitely need to be updated, as otherwise it won't be able to store the archive for each build

-- unless I'm missing something?

aLekSer:
Applied your requests, waiting for Cloud Build logs to see if they push archives per each build to GS bucket.

aLekSer:
Now archive is copied as it should:
```
Already have image (with digest): gcr.io/cloud-builders/gsutil
Copying file://agonessdk-1.1.0-4034ad3-linux-arch_64.tar.gz [Content-Type=application/x-tar]...
/ [0/1 files][ 0.0 B/ 79.8 MiB] 0% Done - - [0/1 files][ 72.2 MiB/ 79.8 MiB] 90% Done - [1/1 files][ 79.8 MiB/ 79.8 MiB] 100% Done \
Operation completed over 1 objects/79.8 MiB.
```

PR #1145, opened 2019-10-24
Refresh client CA certificate if changed

pooneh-m:
If a client CA certificate is changed for agones-allocator, it requires the service pod restart (delete and recreate). To allow picking up new secrets when updated, this change watches the secret mounted volume and if changed reload the certificates in the TLS config.

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"agones.dev/agones/pkg"
@@ -43,6 +44,7 @@ import (
 	"github.com/spf13/viper"
 	"go.opencensus.io/plugin/ochttp"
 	"go.opencensus.io/stats/view"
+	"gopkg.in/fsnotify.v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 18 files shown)


cmd/allocator/main.go
```diff
@@ -202,7 +244,8 @@ func getCACertPool(path string) (*x509.CertPool, error) {
 				return nil, fmt.Errorf("ca cert is not readable or missing: %s", err.Error())
 			}
 			if !caCertPool.AppendCertsFromPEM(caCert) {
-				return nil, fmt.Errorf("client cert %s cannot be installed", certFile)
+				logger.Errorf("client cert %s cannot be installed", certFile)
```
roberthbailey: why isn't this an error any longer?

cmd/allocator/main.go
```diff
@@ -104,15 +105,47 @@ func main() {
...
+	if err := watcher.Add(certDir); err != nil {
+		logger.WithError(err).Fatalf("cannot watch folder %s for secret changes", certDir)
+	}
+
+	tlsCer, err := tls.LoadX509KeyPair(tlsDir+"tls.crt", tlsDir+"tls.key")
 	if err != nil {
-		logger.WithError(err).Fatal("could not get CA certs")
+		logger.WithError(err).Fatal("server TLS could not be loaded")
 	}
+	caCertPool := loadCACertPool()
```
roberthbailey: I haven't looked at this closely so maybe it's ok, but I'm wondering if there is a race condition setting this variable in the go routine and reading it in the tls config. Is this code thread safe?

cmd/allocator/main.go
pooneh-m: Because if a customer mistakenly add a bad client certificate to the authorized list, their service should handle it gracefully instead of crashing.

pooneh-m:
> /cc @cyriltovena is this a better approach than what we've done previously?

Can you please give a context? What are you comparing this with?

markmandel:
> > /cc @cyriltovena is this a better approach than what we've done previously?
> 
> Can you please give a context? What are you comparing this with?

Good question - so we did stuff in the install path so that if the helm generated cert was updated, it would recreate the backing pods for webhooks, etc -- basically restarting the service to reload the file.

I guess the reasoning here is more for being able to cycle certs on a more regular basis without having to restart the whole service? (which does sound better!)

cmd/allocator/main.go
pooneh-m: It should be fine. Go routine updates the variable from a file on a single thread and tls config reads it on the main thread. Do you see a potential thread safety issue?

roberthbailey: If the to two routines are running on different cores and nothing tells the compiler / runtime that the variable is shared, couldn't the update be cached in the first core and never become visible to the other core? 

I feel like this is why https://godoc.org/sync/atomic#Value exists.

roberthbailey:
I think this change doesn't preclude you from using helm to rotate the certs via container restarts (as was done before) but it enables you to also rotate them by updating secrets without restarting the containers. The two approaches aren't mutually exclusive.

pooneh-m:
> I guess the reasoning here is more for being able to cycle certs on a more regular basis without having to restart the whole service? (which does sound better!)

I didn't notice there is question for me. Exactly, this removes the requirement for restarting pods to pick up the new certificates.

cmd/allocator/main.go
pooneh-m: I ran a simplified experiment to change a pointer to a storage location inside a loop on the go routine and read on the main thread, which is similar to what we do here and it works fine. If instead of storing storage location `a := &A{}`, I store the value, e.g. `a := A{}` then the value won't change on the main thread.

```
package main

import (
        "fmt"
)

func main() {
        type A struct {
                I int
        }

        a := &A{}
        counter := 0

        go func() {
                for {
                        counter++
                        a = &A{
                                I: counter,
                        }
                }
        }()

        for {
                fmt.Println(a)
        }
}
```

markmandel: If you want to be super safe, wrap the relevant pieces in a RWMutex?

roberthbailey:
I've lost track of where we are on this PR -- is it back in my court for review?

pooneh-m:
> I've lost track of where we are on this PR -- is it back in my court for review?

There is a feedback to address the thread safety concern by using RWMutex. I'll apply the change and will let you know when it is ready to review.

cmd/allocator/main.go
pooneh-m: Done

cmd/allocator/main.go
```diff
@@ -135,6 +174,14 @@ func main() {
 	logger.WithError(err).Fatal("allocation service crashed")
 }
 
+func loadCACertPool() *x509.CertPool {
+	caCertPool, err := getCACertPool(certDir)
+	if err != nil {
+		logger.WithError(err).Fatal("could not get CA certs")
+	}
+	return caCertPool
```
roberthbailey: If there is an error will this just return nil?

Would it be better to leave caCertPool unmodified if there is an error parsing changes to the cert directory? Or have the process crash?

Right now this error gets logged and then has a potentially weird side effect.

cmd/allocator/main.go
roberthbailey: Then should the return above (line 249) also log but not return an error?

cmd/allocator/main.go
pooneh-m: logger.WithError(err).**Fatal**(..) takes care of crashing.

cmd/allocator/main.go
pooneh-m: Good catch. Thanks!

cmd/allocator/main.go
roberthbailey: Yup, sorry I missed that.

PR #1158, opened 2019-10-28
Broke Allocator Status in Grafana Dashboard

markmandel:
When I renamed the Allocator services, I broke the dashboard components that were referencing things by their old name.

Version set grafana in Makefile because of https://github.com/helm/charts/issues/18215

### DIFF
--- a/build/grafana/dashboard-status.yaml
+++ b/build/grafana/dashboard-status.yaml
@@ -40,8 +40,7 @@ data:
       "editable": true,
       "gnetId": null,
       "graphTooltip": 0,
-      "id": 10,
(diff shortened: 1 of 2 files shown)


build/grafana/dashboard-status.yaml
```diff
@@ -1309,111 +1321,13 @@ data:
           ],
           "repeat": null,
           "repeatDirection": "h",
-          "repeatIteration": 1563127715245,
```
markmandel: @cyriltovena everything seems fine, but any idea why this gets deleted?

I _think_ it might be repeated in the previous version?

PR #1169, opened 2019-11-04
Implementation of GameServer() for Unity

markmandel:
- Refactored `SendRequestAsync` to return a `AsyncResult` which contains   both a success bool, and the returned body text from the http call.
- Used the swagger.json to generate the basic structure of the   `GameServer` object, and then hand modified as needed.
- Had to include third_party/MiniJSON because Unity doesn't natively support `Dictionary`, which means we can't get Label or Annotation data. Then had to manually import the MiniJSON data into GameServer and related model objects.
- Used a set of symlinks to share the Agones Unity SDK code with the unity-simple example. Couldn't think of a better way.
- Minor version update on the example project as that is the earliest version I could get in 2018.4.x
- Added `GameServer` command to the simple-unity example as well.

Work on #927

### DIFF
--- a/examples/unity-simple/Assets/Scripts/UdpEchoServer.cs
+++ b/examples/unity-simple/Assets/Scripts/UdpEchoServer.cs
@@ -73,6 +73,14 @@ namespace AgonesExample
 
                         echoBytes = Encoding.UTF8.GetBytes($"Allocate {ok}");
                         break;
+                    
+                    case "GameServer":
+                        var gameserver = await agones.GameServer();
+                        Debug.Log($"Server - GameServer {gameserver}");
+
+                        ok = gameserver != null;
+                        echoBytes = Encoding.UTF8.GetBytes(ok ? $"GameServer() Name: {gameserver.ObjectMeta.Name} {ok}" : $"GameServer(): {ok}");
+                        break;
 
                     case "Label":
                         if (recvTexts.Length == 3)
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -14,10 +14,14 @@
 // limitations under the License.
 
 using System;
+using System.Collections.Generic;
+using System.Net;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Agones.Model;
+using MiniJSON;
 using UnityEngine;
 using UnityEngine.Networking;
 
@@ -31,8 +35,7 @@ namespace Agones
         /// <summary>
         /// Interval of the server sending a health ping to the Agones sidecar.
         /// </summary>
-        [Range(0.01f, 5)]
-        public float healthIntervalSecond = 5.0f;
+        [Range(0.01f, 5)] public float healthIntervalSecond = 5.0f;
 
         /// <summary>
         /// Whether the server sends a health ping to the Agones sidecar.
@@ -78,7 +81,23 @@ namespace Agones
         /// </returns>
         public async Task<bool> Ready()
         {
-            return await SendRequestAsync("/ready", "{}");
+            return await SendRequestAsync("/ready", "{}").ContinueWith(task => task.Result.ok);
+        }
+
+        /// <summary>
+        /// Retrieve the GameServer details
+        /// </summary>
+        /// <returns>The current GameServer configuration</returns>
+        public async Task<GameServer> GameServer()
+        {
+            var result = await SendRequestAsync("/gameserver", "{}", UnityWebRequest.kHttpVerbGET);
+            if (!result.ok)
+            {
+                return null;
+            }
+
+            var data = Json.Deserialize(result.json) as Dictionary<string, object>;
+            return new GameServer(data);
         }
 
         /// <summary>
@@ -89,7 +108,7 @@ namespace Agones
         /// </returns>
         public async Task<bool> Shutdown()
         {
-            return await SendRequestAsync("/shutdown", "{}");
+            return await SendRequestAsync("/shutdown", "{}").ContinueWith(task => task.Result.ok);
         }
 
         /// <summary>
@@ -100,7 +119,7 @@ namespace Agones
         /// </returns>
         public async Task<bool> Allocate()
         {
-            return await SendRequestAsync("/allocate", "{}");
+            return await SendRequestAsync("/allocate", "{}").ContinueWith(task => task.Result.ok);
         }
 
         /// <summary>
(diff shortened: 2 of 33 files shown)


roberthbailey:
/assign @pooneh-m 

(since you have a lot more c# experience than I do)

roberthbailey:
/assign @dzlier-gcp 

Also Dane...

markmandel:
Weird this flaked - looking at the logs, it completed:

```
time="2019-11-04 05:17:12.715" level=info msg="checking final crash state" gs=udp-serverv658t state=Unhealthy test=TestGameServerRestartBeforeReadyCrash
time="2019-11-04 05:17:12.715" level=info msg="Unhealthy! We are done!" test=TestGameServerRestartBeforeReadyCrash
--- FAIL: TestGameServerRestartBeforeReadyCrash (269.62s)
    gameserver_test.go:303: 
        	Error Trace:	gameserver_test.go:303
        	Error:      	Could not make GameServer Ready: %v
        	Test:       	TestGameServerRestartBeforeReadyCrash
        	Messages:   	timed out waiting for the condition
FAIL
```

Restarting the e2e test.

sdks/unity/model/GameServer.cs
```diff
@@ -0,0 +1,124 @@
...
+{
+    /// <summary>
+    /// A GameServer Custom Resource Definition object We will only export those resources that make the most sense. Can always expand to more as needed.
+    /// </summary>
+    public class GameServer :  IEquatable<GameServer>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GameServer" /> class.
+        /// </summary>
+        public GameServer(IReadOnlyDictionary<string, object> data)
```
pooneh-m: null check?

sdks/unity/model/GameServer.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            this.ObjectMeta = new GameServerObjectMeta((Dictionary<string,object>) data["object_meta"]);
+            this.Spec = new GameServerSpec((Dictionary<string,object>) data["spec"]);
```
pooneh-m: Why this requires force conversion (throw exception if not converted) while status requires soft conversion with `as` (null value if not converted)? If intended can you add a comment to document?

sdks/unity/model/SpecHealth.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// </summary>
+    public class SpecHealth : IEquatable<SpecHealth>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SpecHealth" /> class.
+        /// </summary>
+        public SpecHealth(IReadOnlyDictionary<string, object> data)
+        {
+            this.Disabled = data.TryGetValue("disabled", out var disabled) && bool.Parse((string) disabled);
+            this.PeriodSeconds = (Int64) data["period_seconds"];
```
pooneh-m: Can we use Int64.Parse for these as well?

sdks/unity/AgonesSdk.cs
```diff
@@ -78,7 +81,23 @@ private void OnApplicationQuit()
...
+        /// <returns>The current GameServer configuration</returns>
+        public async Task<GameServer> GameServer()
+        {
+            var result = await SendRequestAsync("/gameserver", "{}", UnityWebRequest.kHttpVerbGET);
+            if (!result.ok)
+            {
+                return null;
+            }
+
+            var data = Json.Deserialize(result.json) as Dictionary<string, object>;
```
pooneh-m: Can we use a JSON deserializer, such as newtonsoft that accept an your predefined objects with json names, instead of converting them to a dictionary and find the values?

examples/unity-simple/Assets/Scripts/UdpEchoServer.cs
```diff
@@ -73,6 +73,14 @@ async void Update()
 
                         echoBytes = Encoding.UTF8.GetBytes($"Allocate {ok}");
                         break;
+                    
+                    case "GameServer":
+                        var gameserver = await agones.GameServer();
```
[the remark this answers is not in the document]
markmandel: That would be a good thing. But we already have the rest of the SDK being async, and it wasn't written that way in the first place :disappointed: we probably should at least be consistent?

sdks/unity/AgonesSdk.cs
```diff
@@ -78,7 +81,23 @@ private void OnApplicationQuit()
         /// </returns>
         public async Task<bool> Ready()
         {
-            return await SendRequestAsync("/ready", "{}");
+            return await SendRequestAsync("/ready", "{}").ContinueWith(task => task.Result.ok);
```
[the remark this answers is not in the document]
markmandel: We are now no longer returning a `Task<bool>` but a `Task<AsyncResult>` from `SendRequestAsync`, since we needed more information from it.

sdks/unity/AgonesSdk.cs
markmandel: I wish. We can't use NuGet, so we can't pull in newtonsoft. There is a [Unity specific version](https://assetstore.unity.com/packages/tools/input-management/json-net-for-unity-11347) of it, but it's huge, and seems like a massive dependency. And we can't force Users to pull things down from the asset store.

MiniJSON is lightweight, and does what we need, even though we have to do more work.

pooneh-m: > We can't use NuGet
Why not?

markmandel: It's not something Unity supports. Also they have older / kinda hacked version of Mono running in there. It's a whole thing.

sdks/unity/model/SpecHealth.cs
markmandel: This is an actual `Int64` stored in the map, so there is no string to parse.

site/content/en/docs/Guides/Client SDKs/unity.md
```diff
@@ -40,6 +40,15 @@ async void SomeMethod()
 }
 ```
 
+To get the details on the [backing `GameServer`]({{< relref "_index.md#gameserver" >}}) call `GameServer()`.
+
+Will return `null` if something there is an error in retrieving the `GameServer` record.
```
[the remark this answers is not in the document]
markmandel: :man_facepalming: thank you!

markmandel:
Should be ready for re-review now.

markmandel:
Thanks for the approval @pooneh-m - would love to get @dzlier-gcp to also look, just because of their Unity experience as well.

Issue #1044, opened 2019-09-06
Upgrade to Kubernetes 1.13


roberthbailey:
1.14 is now generally available on the three major cloud providers, so following our guidance for supported versions, it's time to upgrade to 1.13. 

Given that 1.0 is shipping next week, we should wait until after that to make the change.

heartrobotninja:
That is exciting because I can probably get rid of the special apiamachinery directory.

roberthbailey:
/assign

I'm planning to start working on this once the 1.1 release is finished.

heartrobotninja:
Let me know if you want any help. It might be worth having me handle the go lib portion since I should be able to get rid the modified apimachinery package.

roberthbailey:
List of items to do for upgrading to 1.13 (this is an expanded version of the [1.12 issue](https://github.com/googleforgames/agones/issues/717#issuecomment-506432956)):

- [ ] Update e2e cluster to run against 1.13
  - [ ] Update deployment manage script
  - [ ] Recreate cluster with new scripts
- [ ] Update prow cluster to use 1.13 (even though we aren't using it yet, we should keep it in sync)
  - [ ] Update deployment script cluster
  - [ ] Recreate cluster with new scripts
- [ ] Update the dev tooling to create 1.13 clusters
  - [ ] GKE
  - [ ] Minikube
  - [ ] Kind
  - [ ] Update kubectl
- [ ] Update terraform submodules
  - [ ] GKE
  - [ ] Azure
- [ ] Update documentation for creating clusters to 1.13
  - [ ] Usage requirements
  - [ ] GKE
  - [ ] Minikube
  - [ ] EKS
  - [ ] AKS
  - [ ] Helm documentation (currently still mentions 1.11 so we missed this one last time)
- [ ] Update links to k8s documentation
  - [ ] examples/gameserver.yaml
  - [ ] site/content/en/docs/Reference/gameserver.md
  - [ ] examples/fleet.yaml
  - [ ] site/content/en/docs/Reference/fleet.md
  - [ ] site/content/en/docs/Reference/fleetautoscaler.md
- [ ] Update to client-go 10.0 (based on [compatibility matrix](https://github.com/kubernetes/client-go#compatibility-matrix))

markmandel:
Sounds like a good time to make a config param and/or shortcode in the Hugo site that specifies the supported version, so we only have to update one place.

roberthbailey:
[...]

PR #1178, opened 2019-11-07
Update documentation for the 1.2 release to use Kubernetes 1.13.

roberthbailey:
Also update links from examples to Kubernetes 1.13 docs.

Closes https://github.com/googleforgames/agones/issues/1044

/hold

Adding a hold until we can test Kubernetes 1.13 on EKS & AKS.

### DIFF
--- a/site/content/en/docs/Installation/_index.md
+++ b/site/content/en/docs/Installation/_index.md
@@ -17,6 +17,7 @@ If no dedicated nodes are available, Agones will run on regular nodes.
 
 ## Usage Requirements
 
+{{% feature expiryVersion="1.2.0" %}}
 - Kubernetes cluster version 1.12
     - [Minikube](https://github.com/kubernetes/minikube), [Kind](https://github.com/kubernetes-sigs/kind), [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/),
       [Azure Kubernetes Service](https://azure.microsoft.com/en-us/services/kubernetes-service/) and [Amazon EKS](https://aws.amazon.com/eks/) have been tested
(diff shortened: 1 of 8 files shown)


markmandel:
I see your point from before - we can't have a central place, as we couldn't feature block the new version of Kubernetes.

Sounds like find and replace it is!

roberthbailey:
@aLekSer - are you able to test these instructions of EKS / AKS?

aLekSer:
Hello @roberthbailey I will test and let you know soon.

site/content/en/docs/Installation/_index.md
```diff
@@ -273,6 +338,28 @@ sudo az aks install-cli
...
+AKS_RESOURCE_GROUP=akstestrg     # Name of the resource group your AKS cluster will be created in
+AKS_NAME=akstest     # Name of your AKS cluster
+AKS_LOCATION=westeurope     # Azure region in which you'll deploy your AKS cluster
+
+# Create the Resource Group where your AKS resource will be installed
+az group create --name $AKS_RESOURCE_GROUP --location $AKS_LOCATION
+
+# Create the AKS cluster - this might take some time. Type 'az aks create -h' to see all available options
+# The following command will create a single Node AKS cluster. Node size is Standard A1 v1 and Kubernetes version is 1.13.12. Plus, SSH keys will be generated for you, use --ssh-key-value to provide your values
+az aks create --resource-group $AKS_RESOURCE_GROUP --name $AKS_NAME --node-count 1 --generate-ssh-keys --node-vm-size Standard_A4_v2 --kubernetes-version 1.13.12
```
aLekSer: This command has executed fine for me, I was able to install agones afterwards using install.yaml. The only issue I met is that I need to run `kubectl create namespace agones-system`.
At the end I have tested with creating simple-udp fleet.
Running E2E tests right now, I expect that we should increase node-count to 4 as in GKE cluster.

aLekSer: E2E tests are failing for instance here:
```
FAIL: TestPingHTTP (30.11s)
    ping_test.go:50:
                Error Trace:    ping_test.go:50
                Error:          Expected nil, but got: &url.Error{Op:"Get", URL:"http://51.105.176.201:80", Err:(*net.OpError)(0xc00040a000)}
```

roberthbailey: I'm happy to increase the node count to 4. 

What is the IP address for that can't be fetched? Can you tell if it's because the cluster is too small and there are pending pods?

aLekSer: I need to test one more time, could be connectivity issue between PC running E2E tests and the cluster.

aLekSer: I followed all instructions for AKS cluster, issues does not related to this change.
Something is missing in description on how to create UDP firewall for port range. 
Therefore I think I would send a separate PR for the fix of docs to make E2E tests passing on both AKS and EKS.

roberthbailey: Thanks for testing this out @aLekSer!

roberthbailey: I've adjusted the AKS instructions to create a 4 node cluster instead of a 1 node cluster.

aLekSer:
Tested this PR on AWS EKS, it works fine.

roberthbailey:
@markmandel - @aLekSer tested this on EKS/AKS and I know I tested on GKE. It's been long enough that I can't remember if I tested on minikube / kind but I _think_ I did. Might be worth one more pass and then we should get this merged.

roberthbailey:
I just went through the minikube instructions and verified that they work (and KIND isn't part of this PR -- it's in the developer docs but not part of the user facing website docs). 

I made one small fix -- the GKE and minikube sections somehow lost their link pointing users to continue to the installing agones section at the bottom of the page, so I've added that in. 

/hold cancel

PR #1181, opened 2019-11-07
Implementation of Unity SDK.Connect()

markmandel:
Useful async function to await when the sidecar sdk server comes up, so as to not need to retry other SDK functionality.

Part of #927

### DIFF
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -73,6 +73,38 @@ namespace Agones
         #endregion
 
         #region AgonesRestClient Public Methods
+
+        /// <summary>
+        /// Async method that waits to connect to the SDK Server. Will timeout
+        /// and return false after 30 seconds.
+        /// </summary>
+        /// <returns>A task that indicated whether it was successful or not</returns>
+        public async Task<bool> Connect()
+        {
+            for (var i = 0; i < 30; i++)
+            {
+                Log($"Attempting to connect...{i + 1}");
+                try
+                {
+                    var gameServer = await GameServer();
+                    if (gameServer != null)
+                    {
+                        Log("Connected!");
+                        return true;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Log($"Connection exception: {ex.Message}");    
+                }
+                
+                Log("Connection failed, retrying.");
+                await Task.Delay(1000);
+            }
+
+            return false;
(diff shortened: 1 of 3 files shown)


sdks/unity/AgonesSdk.cs
```diff
@@ -73,6 +73,31 @@ private void OnApplicationQuit()
...
+        /// Async method that waits to connect to the SDK Server. Will timeout
+        /// and return false after 30 seconds.
+        /// </summary>
+        /// <returns>A task that indicated whether it was successful or not</returns>
+        public async Task<bool> Connect()
+        {
+            for (var i = 0; i < 30; i++)
+            {
+                Log($"Attempting to connect...{i + 1}");
+                var gameServer = await GameServer();
```
pooneh-m: try-catch for when GameServer throws exception to retry on the exception as well?

sdks/unity/AgonesSdk.cs
```diff
@@ -73,6 +73,31 @@ private void OnApplicationQuit()
...
+                if (gameServer != null)
+                {
+                    Log("Connected!");
+                    return true;
+                }
+                
+                Log("Connection failed, retrying.");
```
pooneh-m: Instead of delaying for a second, have you considered retrying in a loop with the condition that currentTime - startTime < 30 second?

sdks/unity/AgonesSdk.cs
markmandel: Ooh good idea. I assume I can catch the generic [Exception](https://docs.microsoft.com/en-us/dotnet/api/system.exception?view=netframework-4.8) to catch everything?

sdks/unity/AgonesSdk.cs
markmandel: Interesting! Would that be a big CPU churn when the SDK hasn't spun up yet?

sdks/unity/AgonesSdk.cs
pooneh-m: yes.

sdks/unity/AgonesSdk.cs
pooneh-m: You're right.

sdks/unity/AgonesSdk.cs
markmandel: Done!

PR #1188, opened 2019-11-18
Fix infinite creation of GameServerSets when 1000m CPU limit was used

### LINKED ISSUE #1184: Agones fails to start the pod after update cpu limits to 1000m
**What happened**:

After update deployment we saw the pods not be able to start, the update was change it from 1500m to 1000m

nothing crashes

deploy a fleet with gameserver cpu limits to 1000m, it will work without any issue, update the fleet with other image tag, label or other change and `kubectl apply -f` the manifest again, you will see all the pods not being able to start.

Looks like when the fleet is created does the conversion from 1000m to 1 in the cpu limit, but not in the update of the fleet.

**Environment**:
- Agones version: 1.1.0
`Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.10", GitCommit:"e3c134023df5dea457638b614ee17ef234dc34a6", GitTreeState:"clean", BuildDate:"2019-07-08T03:40:54Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}`
- Cloud provider or hardware configuration: AWS/kops

aLekSer:
Add Semantic.DeepEqual() to compare equal values autoupdated after
creation, for instance 1000m -> 1 CPU limit.
More accurate comparison of GSSet vs Fleet is used.

Closes #1184.

### DIFF
--- a/pkg/fleets/controller.go
+++ b/pkg/fleets/controller.go
@@ -17,7 +17,6 @@ package fleets
 import (
 	"encoding/json"
 	"fmt"
-	"reflect"
 	"time"
 
 	"agones.dev/agones/pkg/apis/agones"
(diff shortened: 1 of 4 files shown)


pkg/fleets/controller.go
```diff
@@ -534,6 +534,42 @@ func (c *Controller) updateFleetStatus(fleet *agonesv1.Fleet) error {
 	return errors.Wrapf(err, "error updating status of fleet %s", fCopy.ObjectMeta.Name)
 }
 
+// equalGameServerTemplates compares template with different Requests and Limits scales
+func equalGameServerTemplates(a, b agonesv1.GameServerTemplateSpec) bool {
```
markmandel: Sounds like we should be using apiequality.Semantic.DeepEqual, like we do [here](https://github.com/googleforgames/agones/blob/master/pkg/fleetautoscalers/controller.go#L264)

(We should probably find other places we use reflect.DeepEquals() and replace those as well.

PR #1192, opened 2019-11-20
Flaky: TestGameServerUnhealthyAfterReadyCrash

markmandel:
This should fix the flakiness with this test.

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -335,13 +335,23 @@ func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
 	defer gsClient.Delete(readyGs.ObjectMeta.Name, nil) // nolint: errcheck
 
 	address := fmt.Sprintf("%s:%d", readyGs.Status.Address, readyGs.Status.Ports[0].Port)
-	conn, err := net.Dial("udp", address)
-	assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


test/e2e/gameserver_test.go
```diff
@@ -335,13 +335,23 @@ func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
...
 
 	address := fmt.Sprintf("%s:%d", readyGs.Status.Address, readyGs.Status.Ports[0].Port)
-	conn, err := net.Dial("udp", address)
-	assert.NoError(t, err)
-	defer conn.Close() // nolint: errcheck
-	_, err = conn.Write([]byte("CRASH"))
-	assert.NoError(t, err)
-	l.WithField("address", address).Info("sent UDP packet")
 
+	// keep crashing, until we move to Unhealthy. Solves potential issues with controller Informaer cache
```
roberthbailey: Maybe I'm missing something but how does this loop exit when the game server moves to unhealthy? It only terminates when there is an error writing a packet to the udp destination, which seems unrelated to the game server becoming unhealthy.

markmandel: Yeah, I just wanted to give it an eventual exit condition. So it will stop when there is nothing left to send things to (i.e. it errors out)

PR #1193, opened 2019-11-22
Implement Reserve for Unity

markmandel:
Implementation of Reserve for the Unity Game Server SDK.

Work on #927

### DIFF
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -95,16 +95,16 @@ namespace Agones
                 }
                 catch (Exception ex)
                 {
-                    Log($"Connection exception: {ex.Message}");    
+                    Log($"Connection exception: {ex.Message}");
                 }
-                
+
                 Log("Connection failed, retrying.");
                 await Task.Delay(1000);
             }
 
             return false;
         }
-        
+
         /// <summary>
         /// Marks this Game Server as ready to receive connections.
         /// </summary>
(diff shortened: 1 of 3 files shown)


markmandel:
Doh! Forgot the @!!!

sdks/unity/AgonesSdk.cs
```diff
@@ -183,9 +183,34 @@ public async Task<bool> SetAnnotation(string key, string value)
...
+        /// Move the GameServer into the Reserved state for the specified Timespan (0 seconds is forever)
+        /// Smallest unit is seconds.
+        /// </summary>
+        /// <param name="duration">The time span to reserve for</param>
+        /// <returns>
+        /// A task that represents the asynchronous operation and returns true if the request was successful
+        /// </returns>
+        public async Task<bool> Reserve(TimeSpan duration)
+        {
+            string json = JsonUtility.ToJson(new Duration(seconds: duration.Seconds));
```
pooneh-m: duration.Duration() instead of new Duration()?
https://docs.microsoft.com/en-us/dotnet/api/system.timespan.duration?view=netframework-4.8

dzlier-gcp: This is a `Duration` struct as defined above, I assume to match the expected JSON format of the request.

site/content/en/docs/Guides/Client SDKs/unity.md
```diff
@@ -60,6 +60,14 @@ Will return `null` if there is an error in retrieving the `GameServer` record.
 ```csharp
 var gameserver = await agones.GameServer();
 ```
+
+To mark the GameServer as [Reserved]({{< relref "_index.md#reserve-seconds" >}}) for a duration call `Reserve(duration
+ TimeSpan)`.
```
dzlier-gcp: `Reserve(TimeSpan duration)`

sdks/unity/AgonesSdk.cs
markmandel: Yes, that is indeed the reason.

Unless there is a good reason to grab the absolute value of the timespan?

PR #1209, opened 2019-12-05
Setting Unreal plugin version to 3

domgreen:
Think the `FileVersion` on the UE Pluigin was bumped by accident.

See docs:

```
"FileVersion" gives the version of the Plugin descriptor file, and should usually set to the highest version that is allowed by the Engine (currently, this is "3")
``` 

Link https://docs.unrealengine.com/en-US/Programming/Plugins/index.html

Signed-off-by: Dominic Green <dominicgreen1@gmail.com>

### DIFF
--- a/sdks/unreal/Agones/Agones.uplugin
+++ b/sdks/unreal/Agones/Agones.uplugin
@@ -1,5 +1,5 @@
 {
-  "FileVersion": 4,
+  "FileVersion": 3,
   "Version": 1,
   "VersionName": "0.2",
   "FriendlyName": "Agones",


markmandel:
Would we consider this a bug fix?

Just checking, since we're in feature freeze atm, but if it's a bug fix, it can come through.

domgreen:
> Would we consider this a bug fix?
> 
> Just checking, since we're in feature freeze atm, but if it's a bug fix, it can come through.

Yep, I would consider this a bug fix ... without this Unreal was unable to load the Agones plugin due to an invalid `FileVersion`.

domgreen:
Thanks. 

For completenes here is a screenshot of the bug:
image

markmandel:
Thanks for pushing this through! We definitely need the Unreal expertise!

PR #1144, opened 2019-10-24
Updates GRPC to v1.20.1

cyriltovena:
This version is the one used by the latest release of OpenCensus which is why I made this PR.

I've also update protobuf to the same version as OpenCensus and grpc-gateway.

PR #1215, opened 2019-12-07
Update GRPC to v1.20.1

### LINKED ISSUE #1214: Updates GRPC to v1.20.1
Taking over work for https://github.com/googleforgames/agones/pull/1144

I'm working over here: https://github.com/markmandel/agones/tree/pull/1144

Tracking the work here, and issues that are being hit.

This version is the one used by the latest release of OpenCensus.

markmandel:
Taking over work started in #1144, so we can upgrade Opencensus

- Updates gRPC in all places, which updates proto and grpc-gateway
- Regenerates the code for all gRPC SDKs as well
- Need to workaround a issue with the /watch Swagger codegen, as the swagger changed between versions.
  This likely means the JSON structure from grpc-gateway for streaming operations also changed.
  This is a breaking change, but we have no control over it (not sure if anyone is using this
  specific REST api endpoint, so it's probably a non issue).
- Use the C++ library requisites file in the cloud build cache key, so it gets updated
  as dependencies get updated.

Closes #1214

Co-authored-by: Cyril Tovena <cyril.tovena@gmail.com>


markmandel:
@cyriltovena because I put you as an additional author, you need to say ` @googlebot I consent.` in a comment, and then it's good to go when we get out of feature freeze next week.

cyriltovena:
@googlebot I consent

PR #1220, opened 2019-12-09
Configurable Log Level for Agones Controller

### LINKED ISSUE #1218: Configurable Log Level for Agones controllers
There is an additional billing cost caused by excessive Agones controller logging.

We need to prioritise the logs, some should be moved to debug level.
Add Helm variables to configure log level.

**Additional context**
Similar work we have done for Agones SDK server:
https://github.com/googleforgames/agones/issues/971

aLekSer:
Add log level parameter to Helm.

Part of #1218.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -68,6 +68,7 @@ const (
 	apiServerSustainedQPSFlag    = "api-server-qps"
 	apiServerBurstQPSFlag        = "api-server-qps-burst"
 	logDirFlag                   = "log-dir"
+	logLevelFlag                 = "log-level"
 	logSizeLimitMBFlag           = "log-size-limit-mb"
 	kubeconfigFlag               = "kubeconfig"
 	defaultResync                = 30 * time.Second
@@ -101,6 +102,15 @@ func main() {
 
 	if ctlConf.LogDir != "" {
 		setupLogging(ctlConf.LogDir, ctlConf.LogSizeLimitMB)
+	}
+
+	logger.WithField("logLevel", ctlConf.LogLevel).Info("Setting LogLevel configuration")
+	level, err := logrus.ParseLevel(strings.ToLower(ctlConf.LogLevel))
+	if err == nil {
+		runtime.SetLevel(level)
+	} else {
+		logger.WithError(err).Info("Specified wrong Logging.SdkServer. Setting default loglevel - Info")
+		runtime.SetLevel(logrus.InfoLevel)
 	}
 
 	logger.WithField("version", pkg.Version).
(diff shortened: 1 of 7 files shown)


cmd/controller/main.go
```diff
@@ -103,6 +104,16 @@ func main() {
 		setupLogging(ctlConf.LogDir, ctlConf.LogSizeLimitMB)
 	}
 
+	logger.WithField("logLevel", ctlConf.LogLevel).Info("Setting LogLevel configuration")
```
markmandel: This all looks really good :+1:

pkg/fleetautoscalers/controller.go
```diff
@@ -115,7 +115,7 @@ func (c *Controller) Run(workers int, stop <-chan struct{}) error {
 		return err
 	}
 
-	c.baseLogger.Info("Wait for cache sync")
+	c.baseLogger.Debug("Wait for cache sync")
```
markmandel: I'm wondering if we want to move the info=>debug logging items to a separate PR?

I'm just wondering if we want to come up with some rules of thumb on what should be `info and what should be `debug`? @pooneh-m @roberthbailey what do you think? I've no idea if there is prior art here?

Just thinking long term, and also for consistency?

roberthbailey: Rules of thumb sound good, but we should discuss them in an issue (so they aren't lost in this PR) and document them somewhere. 

I've created https://github.com/googleforgames/agones/issues/1223.

cmd/controller/main.go
aLekSer: Thanks, we can split that up into two parts. I just scanned through the logging, when nothing happens, no events but some `syncing`, `enqueuing` appears always.
Need to add log example to the ticket.

pkg/fleetautoscalers/controller.go
pooneh-m: https://reflectoring.io/logging-levels/ has some suggestions. :) I don't see an easy way of describing logging level other that use your judgement and get a code review for a second opinion. My judgement is if I delete a log would it reduce my confidence that the system is doing what is expected, if so make it Info, otherwise Debug.

markmandel:
If we want to push this PR through, happy to take the logLevel configuration options (and fix the linting issues).

If we remove the logging level changes, I can re-review, and likely approve.

markmandel:
Couple of other thoughts:

1. In this PR, should we edit the `make install` commands to install at the debug/trace logging level? Seems like we would want this as we are developing?
2. If you set the logging level for the controller, should that also be the default log level for all gameservers as well?

aLekSer:
1. In this PR, should we edit the make install commands to install at the debug/trace logging level? Seems like we would want this as we are developing?
Yes, I will add this into `install` make target.
2. If you set the logging level for the controller, should that also be the default log level for all gameservers as well?
This is a bit more tricky one, I see at as a different PR.

aLekSer:
Moving out log level changes into a separate branch.

markmandel:
Retrying. Looks like a flaky Multi-allocator e2e test.

Issue #1216, opened 2019-12-09
`agones.allocator.http.expose` and `agones.allocator.http.response` are documented but not consumed by Helm


TBBle:
**What happened**:

We attempted to disable the Agones Allocator service during an install of Agones using Helm, as documented at https://agones.dev/site/docs/installation/helm/#configuration and https://github.com/googleforgames/agones/blob/master/install/helm/agones/values.yaml

Since this is minikube, and the default serviceType (`LoadBalancer`) doesn't work out-of-the-box, the installation timed out waiting for an external IP address to be allocated to the LoadBalancer.

**What you expected to happen**:

Agones Allocator service would not attempt to expose itself via http.

**How to reproduce it (as minimally and precisely as possible)**:

Install Agones on Minikube using helm:

```
`helm upgrade --install --wait --set agones.ping.http.expose=false --set agones.ping.udp.serviceType=NodePort 
--set agones.allocator.http.expose=false agones agones/agones --version 1.1.0`
```

The installation will time out.

**Anything else we need to know?**:

It looks like these two options were accidentally cut-and-paste into the docs from the http ping endpoint:
* `agones.allocator.http.expose`
* `agones.allocator.http.response`

as they are not referenced in the [Allocation Service template](https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/service/allocation.yaml), and the latter doesn't make any sense here.

The `response` config option should probably just be removed. The `expose` is slightly trickier, my initial thought is to fix it to behave the way the ping services do (use `agones.allocator.install` for the Deployment and `agones.allocator.http.expose` for the Service) but I'm not sure that makes sense, since unlike the ping services, having this Service exist but only be reachable from inside the cluster is a reasonable use-case.

So maybe we can just remove `agones.allocator.http.expose` and instead rely on users setting `agones.allocator.http.serviceType` to `ClusterIP` for in-cluster-only support.

**Environment**:
- Agones version: 1.1.0
- Kubernetes version (use `kubectl version`): 1.15
- Cloud provider or hardware configuration: Minikube over Hyper-V
- Install method (yaml/helm): helm
- Troubleshooting guide log(s): N/A
- Others:

pooneh-m:
Thanks @TBBle for reporting the issue. I'll look into that.

PR #1221, opened 2019-12-09
Fix documentation for allocator helm args

pooneh-m:
closes: https://github.com/googleforgames/agones/issues/1216


aLekSer:
It seems that I commented, but forgot to start the review yesterday.

PR #1224, opened 2019-12-10
Docs: allocator service should have save-config

aLekSer:
Removed unused response field.
Fix the following warning:             
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply

Left-over of #1221 .

### DIFF
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -117,7 +117,6 @@ agones:
                 operator: Exists
     replicas: 3
     http:
-      response: ok
       port: 443
       serviceType: LoadBalancer
     generateTLS: true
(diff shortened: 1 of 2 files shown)


aLekSer:
@pooneh-m  I think parameter `response` should also be removed.
Response does not contain ok word. It has either this in case of success:
```
{"state":1,"gameServerName":"simple-udp-cc4xp-j29hn","ports":[{"name":"default","port":7262}],"address":"35.230.107.168","nodeName":"gke-test-cluster-default-01eb262a-94gs"}     
```
or State 2:
```
{"state":2}
```

aLekSer:
Additionally to run openssl on Mac one of next command could help:
```
 echo 'export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"' >> ~/.bash_profile
#or 
brew link --force openssl
```

pooneh-m:
Thanks @aLekSer for fixing it.

PR #1168, opened 2019-11-01
Allow fleet with empty Replicas

aLekSer:
Need this to allow FleetAutoScaler with Fleet RollingUpdate strategy.
Verified on GKE cluster.

E2E test added which is reproducing #1156 ticket.

Closes #1156 .

PR #1226, opened 2019-12-10
Revert: Make it possible to create a Fleet with 0 replicas

markmandel:
This reverts some of the changes in #1168, which introduced a breaking change that defaulted Fleets to have 1 replica.

Due to the nature of Go, this also meant that if you attempted to create a Fleet with 0 replicas, it would change it to being 1 replica.

Implemented this fix to ensure there is no breaking change in the release and its also useful to be able to create Fleets with 0 replicas in them.

Also including an extra e2e test to make sure creating a Fleet with 0 replicas continues to function into the future.

### DIFF
--- a/examples/fleet.yaml
+++ b/examples/fleet.yaml
@@ -25,7 +25,7 @@ kind: Fleet
 metadata:
   name: fleet-example
 spec:
-  # the number of GameServers to keep Ready or Allocated in this Fleet. Defaults to 1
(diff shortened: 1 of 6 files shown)


markmandel:
@aLekSer do you see any issue with this change?

markmandel:
Thanks for the quick approval.

Going to wait on @aLekSer 's feedback before merging, and restarting the release process.

aLekSer:
I think it would be useful.

PR #1232, opened 2019-12-13
Move conflict messaged to Debug Logging

markmandel:
Conflicts when attempting to update Kubernetes resources is expected and will happen frequently in Agones.

Currently they are reported as errors, and (a) produce a lot of noise, and (b) are reported as errors, which can often be seen as potential issues by end users, as well as red-herring reasons for actual issues.

We may want to wait until #1220 is complete before merging.

Will help with issues also raised in #1218.

### DIFF
--- a/pkg/util/workerqueue/workerqueue.go
+++ b/pkg/util/workerqueue/workerqueue.go
@@ -26,6 +26,7 @@ import (
 	"agones.dev/agones/pkg/util/runtime"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


pkg/util/workerqueue/workerqueue.go
```diff
@@ -150,8 +151,14 @@ func (wq *WorkerQueue) processNextWorkItem() bool {
 	}
 
 	if err := wq.SyncHandler(key); err != nil {
+		// Conflicts are expected, so only show them in debug operations.
+		if k8serror.IsConflict(err) {
+			wq.logger.Debug(err)
```
roberthbailey: Should this also use the WithField to include the keyName in the debug log?

markmandel: That's a good idea. I'll make that change 👍

PR #1240, opened 2019-12-18
Fix cleanup Agones resources script

aLekSer:
Remove old stable.agones.dev pods getter which was hiding new one.


install/helm/agones/scripts/delete_agones_resources.sh
```diff
@@ -17,14 +17,11 @@
...
 
 for ns in $namespaces; do
-
   # Building the list of pods we need to ensure are deleted.
   gs=$(kubectl -n $ns get gs -o jsonpath='{.items[*].metadata.name}')
 
   for g in $gs; do
     pod=$(kubectl -n $ns get po -l agones.dev/gameserver=$g -o jsonpath='{.items[*].metadata.name}')
-    # TODO(roberthbailey): Remove this after we cut Agones 0.12.0
-    pod=$(kubectl -n $ns get po -l stable.agones.dev/gameserver=$g -o jsonpath='{.items[*].metadata.name}')
```
aLekSer: This variable hide the pod list retrieved by a previous line. This could be the reason why previously `make uninstall` timeout on gameservers crds which not able to be deleted.

aLekSer:
Hello @roberthbailey , please take a look. Either we can merge this or we need to add one more `    pods="$pods $pod"` in between.

PR #1253, opened 2019-12-30
Update Alpine image to 3.11

markmandel:
I think it's been about a year since we updated minor version.

This should also resolve a security issue in musl.


roberthbailey:
There are some dockerfiles in the examples directory that could use this update as well. Do you want to include those in this change or send a separate PR?

markmandel:
I was thinking separate PR. The Alpine images for the examples also need doing, so we could do a full example image update set of PRs to update all the things on images.

Also, examples being updated felt like a much lower priority IMHO than our deployed infrastructure.

Should I create a "Update example images" ticket, with things like language choice, alpine image, SDK, etc?

roberthbailey:
> Should I create a "Update example images" ticket, with things like language choice, alpine image, SDK, etc?

SGTM.

PR #1256, opened 2020-01-03
Feature Gate implementation

markmandel:
Implementation of feature gates based on documentation written in https://agones.dev/site/docs/guides/feature-stages/#feature-gates

This feature was going to be necessary before any new alpha features development could go forward, so that it could be hidden behind flags.

### DIFF
--- a/cmd/sdk-server/main.go
+++ b/cmd/sdk-server/main.go
@@ -65,7 +65,7 @@ var (
 
 func main() {
 	ctlConf := parseEnvFlags()
-	logger.WithField("version", pkg.Version).
+	logger.WithField("version", pkg.Version).WithField("featureGates", runtime.EncodeFeatures()).
 		WithField("ctlConf", ctlConf).Info("Starting sdk sidecar")
 
 	if ctlConf.Delay > 0 {
@@ -253,6 +253,7 @@ func parseEnvFlags() config {
 	pflag.Int(delayFlag, viper.GetInt(delayFlag), "Time to delay (in seconds) before starting to execute main. Useful for tests")
 	pflag.Int(timeoutFlag, viper.GetInt(timeoutFlag), "Time of execution (in seconds) before close. Useful for tests")
 	pflag.String(testFlag, viper.GetString(testFlag), "List functions which shoud be called during the SDK Conformance test run.")
+	runtime.FeaturesBindFlags()
 	pflag.Parse()
 
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
@@ -266,6 +267,9 @@ func parseEnvFlags() config {
 	runtime.Must(viper.BindEnv(grpcPortFlag))
 	runtime.Must(viper.BindEnv(httpPortFlag))
 	runtime.Must(viper.BindPFlags(pflag.CommandLine))
+	runtime.Must(runtime.FeaturesBindEnv())
+
+	runtime.Must(runtime.ParseFeaturesFromEnv())
 
 	return config{
 		IsLocal:   viper.GetBool(localFlag),
--- a/pkg/util/runtime/features.go
+++ b/pkg/util/runtime/features.go
@@ -1,0 +1,110 @@
+// Copyright 2020 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runtime
+
+import (
+	"net/url"
+	"strconv"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+const (
+	featureGateFlag = "feature-gates"
+
+	// FeatureExample is an example feature gate flag, used for testing and demonstrative purposes
+	FeatureExample Feature = "Example"
+)
+
+var (
+	// featureDefaults is a map of all Feature Gates that are
+	// operational in Agones, and what their default configuration is.
+	featureDefaults = map[Feature]bool{
+		FeatureExample: true,
+	}
+
+	// featureGates is the storage of what features are enabled
+	// or disabled.
+	featureGates map[Feature]bool
+)
+
+// Feature is a type for defining feature gates.
+type Feature string
+
+// FeaturesBindFlags does the Viper arguments configuration. Call before running pflag.Parse()
+func FeaturesBindFlags() {
+	viper.SetDefault(featureGateFlag, "")
+	pflag.String(featureGateFlag, viper.GetString(featureGateFlag), "Flag to pass in the url query list of feature flags to enable or disable")
+}
+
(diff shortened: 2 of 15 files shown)


pkg/util/runtime/features.go
```diff
@@ -0,0 +1,98 @@
...
+// ParseFeatures parses the url encoded query string of features and stores the value
+// for later retrieval
+func ParseFeatures(queryString string) error {
+	features := map[Feature]bool{}
+	values, err := url.ParseQuery(queryString)
+	if err != nil {
+		return errors.Wrap(err, "error parsing query string for feature gates")
+	}
+
+	for k := range values {
```
roberthbailey: Should this do any error checking on the value of the feature name? A simple check would be length (e.g. not more than 64 chars) or if we wanted to be more pedantic you could verify that the feature name was something hard-coded into this file (in a list of features that we know about).

pkg/util/runtime/features.go
```diff
@@ -0,0 +1,98 @@
...
+
+	featureGates = features
+	return nil
+}
+
+// FeatureEnabled returns if a Feature is enabled or not
+func FeatureEnabled(feature Feature) bool {
+	enabled, ok := featureGates[feature]
+	if !ok {
+		return featureDefaults[feature]
```
roberthbailey: This will default to false if we forget to add something to the default feature set, but I wonder if it would make more sense to also check for existence here and log an error if it doesn't exist (before returning false).

pkg/util/runtime/features_test.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func TestFeatures(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Empty query string", func(t *testing.T) {
+		err := ParseFeatures("")
+		assert.NoError(t, err)
+		assert.False(t, FeatureEnabled(FeatureExample))
+		assert.Empty(t, EncodeFeatures())
```
roberthbailey: This is in the same package as the code so you could check if `featureGates` is nil. Is this a proxy for that check?

cmd/sdk-server/main.go
```diff
@@ -65,7 +65,7 @@ var (
 
 func main() {
 	ctlConf := parseEnvFlags()
-	logger.WithField("version", pkg.Version).
+	logger.WithField("version", pkg.Version).WithField("flags", runtime.EncodeFeatures()).
```
roberthbailey: Why log the feature gates as "flags" instead of "featureGates" (or "feature-gates")?

(same thing in the other cmd/.../main.go files).

markmandel: Yeah, that's a completely valid point. Making the change now.

`featureGates` seems most consistent with the other names.

pkg/util/runtime/features_test.go
markmandel: This was also just to make sure we got the same thing out as we put in, and that it didn't fall over if there was no actual features.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -19,7 +19,7 @@ A feature within Agones can be in `Alpha`, `Beta` or `Stable` stage.
 
 ## Feature Gates
 
-`Alpha` and `Beta` features can be enabled or disabled through the `agones.featureGate` configuration option 
+`Alpha` and `Beta` features can be enabled or disabled through the `agones.featureGates` configuration option 
 that can be found in the [Helm configuration]({{< ref "/docs/Installation/Install Agones/helm.md#configuration" >}}) documentation.
 
 The current set of `alpha` and `beta` feature gates are:
```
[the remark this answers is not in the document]
markmandel: I like it. Done!

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -174,6 +174,7 @@ The following tables lists the configurable parameters of the Agones chart and t
 | --------------------------------------------------- | ----------------------------------------------------------------------------------------------- | ---------------------- |
 
 | `agones.controller.logLevel`                        | Agones Controller Log level. Log only entries with that severity and above                      | `info`                |
+| `agones.featureGates`                               | A URL query encoded string of Flags to enable/disable e.g. "Example=true&OtherThing=false"      | ``                    |
```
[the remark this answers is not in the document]
markmandel: I like it. Done.

pkg/util/runtime/features.go
markmandel: This actually made me realise that the default value is not properly tested, because of this. Updating the tests.

Given that the default for `alpha` (i.e. getting started) with a feature should be false by default - it actually means you only need to put an entry in at the `beta` stage. So I'm wondering if it's actually good that it is defaulted to false, because it's one less step to add when creating a new feature?

WDYT?

pkg/util/runtime/features.go
markmandel: I like the idea of throwing an error if the feature gate is not a valid one in the list -- will help with user error. Adding this in now.

pkg/util/runtime/features.go
```diff
@@ -0,0 +1,120 @@
...
+	features := map[Feature]bool{}
+	values, err := url.ParseQuery(queryString)
+	if err != nil {
+		return errors.Wrap(err, "error parsing query string for feature gates")
+	}
+
+	for k := range values {
+		f := Feature(k)
+		if !containsFeatureGate(f) {
+			return errors.Errorf("Feature Gate '%s' is not a valid Feature Gate", f)
```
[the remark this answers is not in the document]
markmandel: Oooh! TIL!

pkg/util/runtime/features.go
```diff
@@ -0,0 +1,120 @@
...
+	for k, v := range featureGates {
+		values.Add(string(k), strconv.FormatBool(v))
+	}
+	return values.Encode()
+}
+
+// containsFeatureGate returns true if the passed in Feature
+// is in the list of current FeatureGates that are being
+// operated in this application
+func containsFeatureGate(f Feature) bool {
```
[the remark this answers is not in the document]
markmandel: yyyyeeeesssssss. :fire:

markmandel:
Weird flake. Haven't seen this one before.

```
--- FAIL: TestFleetUpdates (0.00s)
    --- FAIL: TestFleetUpdates/rolling (383.06s)
        fleet_test.go:409: 
            	Error Trace:	fleet_test.go:409
            	Error:      	Expected nil, but got: &errors.errorString{s:"timed out waiting for the condition"}
            	Test:       	TestFleetUpdates/rolling
        fleet_test.go:414: 
            	Error Trace:	fleet_test.go:414
            	Error:      	Expected nil, but got: &errors.errorString{s:"timed out waiting for the condition"}
            	Test:       	TestFleetUpdates/rolling
FAIL
```

pkg/util/runtime/features.go
roberthbailey: Now that we validate that feature names must be in the defaults, this comment no longer applies.

pkg/util/runtime/features.go
```diff
@@ -0,0 +1,114 @@
...
+// configured by FeaturesBindFlags() and FeaturesBindEnv()
+func ParseFeaturesFromEnv() error {
+	return ParseFeatures(viper.GetString(featureGateFlag))
+}
+
+// ParseFeatures parses the url encoded query string of features and stores the value
+// for later retrieval
+func ParseFeatures(queryString string) error {
+	features := map[Feature]bool{}
+	// copy the defaults into this map
```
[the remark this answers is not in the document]
markmandel: Actually, I should remove the fallback, because there is no reason for it to be there.

I did this because I wanted to replace all the features in one switch, rather than one at a time.

Also makes testing easier, because you can run ParseFeatures() multiple times, and it cleans up after itself.

Issue #679, opened 2019-03-28
Create npm package for Node.js sdk


msannikov:

rorygarand:
I'll put in a PR for this. I'd really like to use npm for my project.

rorygarand:
Actually it looks like all that's really needed is access to Google's npm account. I'd love to help out if I can. Ping me if that's the case.

markmandel:
Could we / should we use https://github.blog/2019-05-10-introducing-github-package-registry/ ?

Or should we have our own npm package? (I've no idea about this ecosystem)

rorygarand:
Interesting. That would definitely work, but in my mind it's more a question of how Google handles (or wants to handle) packages they maintain.

I'm not sure if it matters to anyone at any level whether developers install agones via:

```sh
$ npm install @github/agones
```
vs.

```sh
$ npm install @google/agones
```

markmandel:
I think we could have an agones specific account - we don't necessarily need to have it tied to google itself.

rorygarand:
I'm agnostic about the approach. The github package registry is pretty neat but it is invite-only at the moment.

If adding to the npm registry is as simple as someone creating an account, I don't mind managing that.

Getting the project into either registry would make my life easier so let me know what I can do to help.

markmandel:
@thisisnotapril what's the common practice for this kind of thing? (storing account credentials for external repositories, etc?)

rorygarand:
@markmandel @thisisnotapril any updates on this? i'm currently forking the repo so i can maintain my own registry through github with this as a package. 

I wouldn't mind taking ownership of pushing to the googleforgames registry whenever there are version updates, since I'm already doing that anyways for myself, but I think I would need some access credentials to the repo that maybe you don't want to give to a rando off the internet.

I can also walk someone else through the process, or write something up etc. It's not terribly difficult but would make dev a little easier for my team. Currently we have to manually ensure we're all on the same version of agones locally before deploying.

markmandel:
Hey! We should totally do this. Let's push it to [this repository](https://github.com/googleforgames/agones/packages).

What would be ideal for me :smile: is to have a Makefile target I can add to our release process.
Somewhere in here would likely be ideal:
https://github.com/googleforgames/agones/blob/master/build/includes/release.mk

Then I don't even need to know how NPM works! :smile: 

[...]

PR #1264, opened 2020-01-08
Node packaging

rorygarand:
Adds a publish script to send the node sdk to the github package registry.

Related to: https://github.com/googleforgames/agones/issues/679.


build/build-sdk-images/node/publish.sh
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+AGONES_PATH=$1
+RELEASE_VERSION=$2
+
+npm login --registry=https://npm.pkg.github.com
+
+cd $AGONES_PATH/sdks/nodejs
+
+npm version $RELEASE_VERSION
```
steven-supersolid: How/when does this script get run?

I think updating the version number should be a careful change and follow semver, so potentially even a manual step. Or are these variables automatically populated and the aim is to keep the package version the same as the release version? If that is the case then what will happen when there are no SDK changes and how do we still comply with semver?

steven-supersolid: OK I've read #679 . Perhaps best to first focus on the publishing part and ensure it works and can look at again if we want to do version sync?

markmandel: I'm happy to remove the version update. i was going to handle the release part - sounds like a check in the release checklist that we update the version (we can do this by hand?) would work.

So for this PR, we could drop the `npm version $RELEASE_VERSION` command?

markmandel:
Ah, this did actually break a thing!

We have a SDK test here:
https://github.com/markmandel/agones/blob/master/test/sdk/nodejs/testSDKClient.js#L15

So we'll need to change the name there in this PR too for it to pass.

We'll also need to update the nodejs example:

So it's worth noting this is a breaking change, just because the package name has changed.

build/build-sdk-images/node/publish.sh
rorygarand: Correct. I only added that as a potential way to address an earlier comment from you @markmandel.

@steven-supersolid I believe the idea is to run this from `release.mk` but I was going to leave that part to Mark.

rorygarand:
I was planning on updating the docs, is this PR the best place to do it?

Also, I'll add a fix to the test and remove the `npm version` command.

markmandel:
Yep this PR will be the perfect place - we love having it all in one spot.

build/build-sdk-images/node/publish.sh
steven-supersolid: Looks good and it seems we keep the SDK version the same as the Agones version?

markmandel:
```
> @ testSDK /go/src/agones.dev/agones/test/sdk/nodejs
> node ./testSDKClient.js

internal/modules/cjs/loader.js:670
    throw err;
    ^

Error: Cannot find module '@googleforgames/agones'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:668:15)
    at Function.Module._load (internal/modules/cjs/loader.js:591:27)
    at Module.require (internal/modules/cjs/loader.js:723:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at Object.<anonymous> (/go/src/agones.dev/agones/test/sdk/nodejs/testSDKClient.js:15:19)
    at Module._compile (internal/modules/cjs/loader.js:816:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:827:10)
    at Module.load (internal/modules/cjs/loader.js:685:32)
    at Function.Module._load (internal/modules/cjs/loader.js:620:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:877:12)
```

Not sure how to solve these problems, but figured I'd find the relevant error message for you.

Let me know if you need help running this test locally.

steven-supersolid:
Oh it won't work until published but we shouldn't publish until the tests all pass :) There is however another way for local testing using npm link.
From the test sdk folder run `npm link <relative path to sdk>`  - this could be added to the script. Please note it will change `package-lock.json` but any changes can be discarded
https://docs.npmjs.com/cli/link.html

build/build-sdk-images/node/publish.sh
rorygarand: That was the intent... I took out the `npm version` command for now though, we can always add it later if we want.

markmandel:
Passing now! This LGTM - @steven-supersolid any issues on your end, or can I approve and merge?

steven-supersolid:
Oh I thought I already approved. LGTM

PR #1267, opened 2020-01-10
Update golangci-lint, add more linters

aLekSer:
Fix minor issues, but plenty of them. Add new linters.
Add pointers when it is necessary for the sake of performance (avoid coping structures over 512 B in size).
Most fixes come from gocritic linter.
https://go-critic.github.io/overview.html

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -46,6 +46,7 @@ output:
 
 linters:
   enable:
+    - deadcode
     - megacheck
     - govet
     - golint
@@ -53,16 +54,39 @@ linters:
     - unconvert
     - dupl
     - goconst
+    - gocritic
     - gocyclo
     - goimports
+    - interfacer
     - maligned
     - megacheck
+    - scopelint
+    - staticcheck
+    - structcheck
     - misspell
     - unparam
     - nakedret
     - bodyclose
 
+linters-settings:
+  gocritic:
+    enabled-tags:
+      - performance
+      - opinionated
+      - diagnostic
+    disabled-checks:
+      - paramTypeCombine
+      - unnamedResult
+    settings: # settings passed to gocritic
+      hugeParam:
+        sizeThreshold: 512
+      rangeValCopy:
+        sizeThreshold: 512
+        skipTestFuncs: true
+
 issues:
   # This turns off the default excludes - which was causing the linter
   # to miss things like erroneous comments
   exclude-use-default: false
+  exclude:
+    - Using the variable on range scope .* in function literal
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -243,19 +243,20 @@ func getCACertPool(path string) (*x509.CertPool, error) {
 	}
 
 	for _, file := range filesInfo {
-		if strings.HasSuffix(file.Name(), ".crt") || strings.HasSuffix(file.Name(), ".pem") {
-			certFile := filepath.Join(path, file.Name())
-			caCert, err := ioutil.ReadFile(certFile)
-			if err != nil {
-				logger.Errorf("ca cert is not readable or missing: %s", err.Error())
-				continue
-			}
-			if !caCertPool.AppendCertsFromPEM(caCert) {
-				logger.Errorf("client cert %s cannot be installed", certFile)
-				continue
-			}
-			logger.Infof("client cert %s is installed", certFile)
+		if !strings.HasSuffix(file.Name(), ".crt") && !strings.HasSuffix(file.Name(), ".pem") {
+			continue
 		}
+		certFile := filepath.Join(path, file.Name())
+		caCert, err := ioutil.ReadFile(certFile)
+		if err != nil {
+			logger.Errorf("ca cert is not readable or missing: %s", err.Error())
+			continue
+		}
+		if !caCertPool.AppendCertsFromPEM(caCert) {
+			logger.Errorf("client cert %s cannot be installed", certFile)
+			continue
+		}
+		logger.Infof("client cert %s is installed", certFile)
 	}
 
 	return caCertPool, nil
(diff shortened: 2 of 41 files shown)


cmd/allocator/main.go
```diff
@@ -212,21 +212,21 @@ func newServiceHandler(kubeClient kubernetes.Interface, agonesClient versioned.I
 }
 
 // Set up our client which we will use to call the API
-func getClients() (*kubernetes.Clientset, *versioned.Clientset, error) {
+func getClients() (kubeClient *kubernetes.Clientset, agonesClient *versioned.Clientset, err error) {
```
roberthbailey: I don't understand the suggested changes to this function; the old way looks more correct to me stylistically. Do you know which check triggered this?

.golangci.yml
```diff
@@ -46,23 +46,50 @@ output:
...
     - nakedret
     - bodyclose
 
+linters-settings:
+  gocritic:
+    enabled-tags:
+      - performance
+      - opinionated
+      - diagnostic
+    disabled-checks:
```
roberthbailey: Did these checks produce too many diffs or did they produce diffs that you didn't agree with?

cmd/allocator/main.go
markmandel: Yeah, I'm not a fan of named return variables either.

aLekSer: I think the idea behind this check was to make function signature more readable from the first line, especially when it returns same 3 types for instance, which can be misleading without names.
But as we have 3 votes to not use them I will remove this check today.

.golangci.yml
aLekSer: I don't like paramTypeCombine especially in this case:
```
            ^
pkg/gameservers/controller.go:91:1: paramTypeCombine: func(
        wh *webhooks.WebHook,
        health healthcheck.Handler,
        minPort, maxPort int32,
        sidecarImage string,
        alwaysPullSidecarImage bool,
        sidecarCPURequest resource.Quantity,
        sidecarCPULimit resource.Quantity,
        sdkServiceAccount string,
        agonesInformerFactory externalversions.SharedInformerFactory) *Controller could be replaced with func(wh *webhooks.WebHook,
        health healthcheck.Handler,
        minPort, maxPort int32,
        sidecarImage string,
        alwaysPullSidecarImage bool,
        sidecarCPURequest,
        sidecarCPULimit resource.Quantity,

        sdkServiceAccount string,
        agonesInformerFactory externalversions.SharedInformerFactory) *Controller (gocritic)
```
Seems it just combined `sidecarCPURequest, sidecarCPULimit resource.Quantity,`, but I can add this check also if it is needed.

aLekSer: `sloppyReassign` example:
```
pkg/gameservers/controller.go:389:5: sloppyReassign: re-assignment to `err` can be replaced with `err := c.syncGameServerShutdownState(gs)` (gocritic)
        if err = c.syncGameServerShutdownState(gs); err != nil {
```

aLekSer:
Some error on TestAllocatorCrossNamespace E2E:
https://gist.github.com/aLekSer/87c8f03ce7af07063866f994f30bad2f

.golangci.yml
roberthbailey: For the paramTypeCombine check, the common suggestion I've heard is that when parameter lists get that long you should create a struct and pass the struct instead of a list of individual parameters. The reason being that it adds better type checking, and when the parameter list gets long it becomes easy to pass things in the wrong order, especially if parameters have the same type. This check seems to make that worse, so I agree that we shouldn't enable it. 

The change in sloppyReassign looks correct to me.

cmd/allocator/main.go
```diff
@@ -323,6 +324,7 @@ func (h *httpHandler) allocateHandler(w http.ResponseWriter, r *http.Request) {
 
 func httpCode(err error) int {
 	code := http.StatusInternalServerError
+	// nolint: gocritic
```
roberthbailey: Does it want to turn this into an if statement? A switch with one case seems strange.

test/e2e/allocator_test.go
```diff
@@ -223,7 +223,7 @@ func createAllocationPolicy(t *testing.T, p *multiclusterv1alpha1.GameServerAllo
 	t.Logf("created allocation policy %v", policy)
 }
 
-func getAllocatorEndpoint(t *testing.T) (string, int32) {
+func getAllocatorEndpoint(t *testing.T) (ip string, intPort int32) {
```
roberthbailey: Is this left over from the old check?

test/e2e/allocator_test.go
```diff
@@ -369,7 +370,7 @@ func refreshAllocatorTLSCerts(t *testing.T, host string) []byte {
 	return pub
 }
 
-func generateTLSCertPair(t *testing.T, host string) ([]byte, []byte) {
+func generateTLSCertPair(t *testing.T, host string) (pemPub []byte, pemPriv []byte) {
```
roberthbailey: This one too...

cmd/allocator/main.go
aLekSer: Yes, but it is a type switch, so it seems there is no way to switch this structure into an if statement.
 https://github.com/golang/go/wiki/Switch#type-switch

aLekSer: The only way could be this syntax:
```
if t, ok := err.(k8serror.APIStatus) ; ok {
 ...
}
```

aLekSer: Both options will be equivalent:
https://play.golang.org/p/LkcbMXHC6gU

aLekSer: By the way, same one case switch could be found here:
https://github.com/kubernetes/apimachinery/blob/51168acb8295af7876446c9e6a5e77aee0a838a7/pkg/api/errors/errors.go#L577

test/e2e/allocator_test.go
aLekSer: yes, did not notice these 2 named results. Fixed now

cmd/allocator/main.go
roberthbailey: Clearly they aren't running gocritic. :)

I'm guessing that the tool prefers that you use the equivalent syntax (explicitly checking for the desired type) makes sense when you are checking for a single type, because then you have a simple if statement instead of a switch statement with a single case.

Issue #1098, opened 2019-10-03
Add Fuzz Tests


roberthbailey:
We should consider adding fuzz testing to Agones. 

Kubernetes uses https://github.com/google/gofuzz (originally was in the Kubernetes repo but was split out). 

https://github.com/google/oss-fuzz is Google’s free fuzzing-as-a-service platform and has recently added support for Go fuzzers. 

There are probably other options as well that I'm not aware of.

aLekSer:
Question here what parts of Agones should be covered with Fuzz testing?
Found a recent thread in Kubernetes slack about fuzz testing CRDs
https://kubernetes.slack.com/archives/CAR30FCJZ/p1573755986055900
Tried to make a test using this tool:
https://github.com/munnerz/crd-schema-fuzz
But we are not using fresh dependencies so I was not able build a simple test.
Also I could not find https://github.com/kubernetes/apimachinery/tree/master/pkg/api/apitesting this functionality in the vendored apimachinery.

roberthbailey:
> Question here: What parts of Agones should be covered with Fuzz testing?

I would imagine that our api surface would make the most sense. I think looking at how k8s does fuzz testing and doing something similar (which it looks like you were trying to do) would be a great starting point.

aLekSer:
Thanks for your answer.
Two more questions which is worth to consider:
https://github.com/fuzzitdev/fuzzit
https://github.com/dvyukov/go-fuzz

Fuzzit is free for open source, prometheus is using it, it adds a new `fuzzit: pass` label on repo main page, and could be exucuted on every pull request. 

Currently investigating if we can add fuzz tests also for `applyBufferPolicy()` func of a `FleetAutoScaler`.

aLekSer:
We should somehow use `fuzzer` to test API:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#update-the-fuzzer

aLekSer:
Followed an example from the book "Programming Kubernetes" https://github.com/programming-kubernetes/pizza-apiserver/tree/master/pkg/apis/restaurant/fuzzer
Was able to run a simple roundtrip fuzz test for `agones.dev/v1`. 
It seems that we need to have additional `addKnownTypes()` call for the `__internal` GroupVersion:
```
var SchemeGroupVersion = schema.GroupVersion{Group: agones.GroupName, Version: k8sruntime.APIVersionInternal}
//with next line commented out
[...]

PR #1269, opened 2020-01-13
Fuzz Roundtrip test for v1 Agones schemas

aLekSer:
Add Roundtrip tests for agones.dev external types.
Skip multicluster v1alpha1 API schema for now.
For #1098 .


aLekSer:
I have removed Multicluster v1alpha1 IP which causes test fail "agones.dev/agones/pkg/apis/multicluster/v1alpha1" 
We can add this API into a separate test with `t.skip()` initially.

PR #1278, opened 2020-01-15
Need to change to https url on swagger-codegen download

markmandel:
This broke CI, so fixing it!


markmandel:
More context as a historical artefact:
https://support.sonatype.com/hc/en-us/articles/360041287334

PR #1281, opened 2020-01-16
Site: Add video & podcasts to third party section

markmandel:
New podcast section under "Third Party", and added last year's Cloud Next event.

### DIFF
--- a/site/content/en/docs/Third Party Content/podcasts.md
+++ b/site/content/en/docs/Third Party Content/podcasts.md
@@ -1,0 +1,19 @@
+---
+title: "Third Party Podcasts"
+linkTitle: "Podcasts"
+date: 2020-01-15
+description: "Community contributed podcasts on Agones."
+weight: 20
+---
+
+## 2019
+
+- Nov 6 - [Google Cloud Platform Podcast: Supersolid with Kami May]
+    (https://gcppodcast.com/post/episode-202-supersolid-with-kami-may/)
+
+## 2018
+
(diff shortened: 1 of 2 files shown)


site/content/en/docs/Third Party Content/podcasts.md
```diff
@@ -0,0 +1,11 @@
+---
+title: "Third Party Podcasts"
+linkTitle: "Podcasts"
+date: 2020-01-15
+description: "Community contributed podcasts on Agones."
+weight: 20
+---
+
+- [Google Cloud Platform Podcast - #202 Supersolid with Kami May](https://gcppodcast.com/post/episode-202-supersolid-with-kami-may/)
+- [Google Cloud Platform Podcast - #142 Agones with Mark Mandel and Cyril Tovena](https://gcppodcast.com/post/episode-142-agones-with-mark-mandel-and-cyril-tovena/)
```
roberthbailey: I think that we should keep these in reverse chronological order.

It's really hard to tell what the publish date of the gcp podcast episodes are, but it seems that gcp#142 came before k8s#26 and so this line should be at the bottom of the list.

markmandel: I like it! Should we add the date to the listing as well? I was debating doing that, but wasn't sure.

roberthbailey: I think that would be a good idea.

markmandel:
Now in reverse chronological and broken out by year.

PR #1283, opened 2020-01-17
CPP SDK example code: join threads

aLekSer:
Fix core dump on example code.

The threads goes out of main() scope without joining or detaching, which leads to an error:
```
...marked for Shutdown
terminate called without an active exception
Aborted
```


aLekSer:
`Detach()` could be changed to `Join()` and a global variable or switch from std::thread to some scoped threads, but this version is enough to avoid termination errors.

examples/cpp-simple/server.cc
```diff
@@ -114,6 +114,8 @@ int main() {
       std::cout << "...marked for Shutdown\n" << std::flush;
     }
   }
+  health.detach();
```
devjgm: It's preferable to join rather than detach, when possible. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cp26-dont-detach-a-thread

Can the thread bodies detect that the SDK was shutdown and then exit gracefully? That way main can join the threads here instead of detach them.

aLekSer: Yes, I will include that shutdown trigger for threads into example code.

markmandel:
/assign @devjgm 

@devjgm is this good to go now? I defer to you on all things C++ :smile:

examples/cpp-simple/server.cc
```diff
@@ -115,5 +120,9 @@ int main() {
     }
   }
 
+  stop_threads.store(1);
+  health.join();
```
devjgm: There are other "return -1" statements that will return from main without joining these threads. Will that also exhibit the problem this PR is fixing?

examples/cpp-simple/server.cc
```diff
@@ -115,5 +120,9 @@ int main() {
...
+  watch.join();
```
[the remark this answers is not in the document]
markmandel: WatchGameServer() will stop when the sidecar shuts down, as the connection will close.

markmandel:
Gentle bump on this - looks like there are still some outstanding questions.

examples/cpp-simple/server.cc
markmandel: @aLekSer did you have an answer to this question?

aLekSer: Yes, that could lead to an issue, so need one more update.

examples/cpp-simple/server.cc
```diff
@@ -41,7 +46,11 @@ void WatchUpdates(std::shared_ptr<agones::SDK> sdk) {
               << std::flush;
   });
 }
-
+void join_threads(std::thread health, std::thread watch) {
```
devjgm: Instead of a function, I think we should use an RAII object to join the threads. This should work with the current code and all of its return statements, and it will be more future proof because it will continue to work even if we add new return statements in the future.

How about something like this (untested):

```cc
class ThreadJoiner {
 public:
  explicit ThreadJoiner(std::thread t)
    : t_(std::move(t)) {}

  ~ThreadJoiner() {
    t_.join();
  }
 private:
  std::thread t_;
};
```
then in main() use it like

```cc
int main() {
  ...
  ThreadJoiner health(std::thread(DoHealth, sd));
  ThreadJoiner watch(std::thread(WatchUpdates, sdk));
```

markmandel: Fancy!

aLekSer: I also was thinking about doing so, but some issues with raising exceptions from destructor this could cause. However let stick to what you suggest.

examples/cpp-simple/server.cc
```diff
@@ -109,6 +133,7 @@ int main() {
       if (!status.ok()) {
         std::cerr << "Could not run Shutdown():" << status.error_message()
                   << ". Exiting!\n";
+        join_threads(std::move(health), std::move(watch));
```
devjgm: I think this function call (and the function itself) were meant to be removed?

Otherwise, I think this change looks good. Thanks for the fix and the updates!

PR #1285, opened 2020-01-19
Stubs for SDK alpha/beta/stable functionality

markmandel:
This gets the requisite pieces in place, such that alpha and beta features can be written for the GameServer SDKs.

This includes:
- Moving the sdk.proto into the proto directory
- Updating all the gen scripts for clients for the new sdk.proto location
- creating new alpha.proto and beta.proto definitions, with an empty service stub
- Integrated the new alpha and beta service stubs into the sdkserver local and sidecar implementations.
- Moved all the swagger generated files under ./sdks/swagger (felt like  an appropriate place)

I've deliberately not update all the clients as (a) I felt it would make this PR too big and (b) each language's client may have slightly different integration requirements. I figured it could be handled on a case by case basis, as required.


pkg/sdkserver/alpha/localsdk.go
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package alpha
+
+import "agones.dev/agones/pkg/sdk/alpha"
+
+var (
```
roberthbailey: nit: This can be on a single line since there is only one variable.

pkg/sdkserver/alpha/sdkserver.go
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package alpha
+
+import "agones.dev/agones/pkg/sdk/alpha"
+
+var (
```
roberthbailey: nit: This can be on a single line since there is only one variable.

pkg/sdkserver/beta/localsdk.go
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package beta
+
+import (
+	"agones.dev/agones/pkg/sdk/beta"
+)
+
+var (
```
roberthbailey: nit: This can be on a single line since there is only one variable.

pkg/sdkserver/beta/localsdk.go
```diff
@@ -0,0 +1,32 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package beta
+
+import (
```
roberthbailey: nit: This can be on a single line since there is only one import.

pkg/sdkserver/beta/sdkserver.go
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package beta
+
+import "agones.dev/agones/pkg/sdk/beta"
+
+var (
```
roberthbailey: nit: This can be on a single line since there is only one variable.

cmd/sdk-server/main.go
```diff
@@ -164,39 +170,57 @@ func main() {
...
-	sdk.RegisterSDKServer(grpcServer, localSDK)
-	return
+	alphaSDK := serveralpha.NewLocalSDKServer()
+	betaSDK := serverbeta.NewLocalSDKServer()
+
+	sdk.RegisterSDKServer(grpcServer, stableSDK)
+	sdkalpha.RegisterSDKServer(grpcServer, alphaSDK)
+	sdkbeta.RegisterSDKServer(grpcServer, betaSDK)
+	return func() {
+		stableSDK.Close()
```
roberthbailey: Should we also close the alpha and beta sdks?  The function comment makes it sound like all SDKs will be cleaned up.

cmd/sdk-server/main.go
```diff
@@ -164,39 +170,57 @@ func main() {
...
-	localSDK.SetExpectedSequence(expectedFuncs)
-	sdk.RegisterSDKServer(grpcServer, localSDK)
-	return
+	stableSDK.SetExpectedSequence(expectedFuncs)
+
+	sdk.RegisterSDKServer(grpcServer, stableSDK)
+	sdkalpha.RegisterSDKServer(grpcServer, alphaSDK)
+	sdkbeta.RegisterSDKServer(grpcServer, betaSDK)
+	return func() {
+		stableSDK.Close()
```
roberthbailey: same here.

pkg/sdkserver/beta/sdkserver.go
markmandel: Updated all of them :+1:

cmd/sdk-server/main.go
markmandel: That's an interesting question. Neither the alpha or beta SDKs have/need a Close() operation. 

Should we add empty ones, just for consistency?

roberthbailey: Since alpha / beta are currently just empty stubs, I can't tell if they will grow to need a close operation (in which case we should add it now so we don't forget to add it here later) or whether we don't expect them to ever need a close operation. In the latter case, I think it's fine to update the function comments. Something like

"... SDK servers, and returns a function that cleans up the SDK implementations when they are finished being used."

Maybe that isn't any better than what you had. :)

markmandel: Since I don't think we can guarantee it either way - I've added a stub function in both alpha and beta SDK, and called it in both returned functions.

At least then if someone _does_ need it, it's easy to see where to place tear down logic.

Issue #1276, opened 2020-01-14
Flaky: TestGameServerReserve


markmandel:
Haven't had a chance to work out why this is flaky, but it is. Seen it a few times across some PRs.

```
--- FAIL: TestGameServerReserve (120.77s)
    gameserver_test.go:494: 
        	Error Trace:	gameserver_test.go:494
        	Error:      	Received unexpected error:
        	            	timed out waiting for the condition
        	            	waiting for GameServer to be Reserved default/udp-serverttdbq
        	            	agones.dev/agones/test/e2e/framework.(*Framework).WaitForGameServerState
        	            		/go/src/agones.dev/agones/test/e2e/framework/framework.go:131
        	            	agones.dev/agones/test/e2e.TestGameServerReserve
        	            		/go/src/agones.dev/agones/test/e2e/gameserver_test.go:493
        	            	testing.tRunner
        	            		/usr/local/go/src/testing/testing.go:909
        	            	runtime.goexit
        	            		/usr/local/go/src/runtime/asm_amd64.s:1357
        	Test:       	TestGameServerReserve
    gameserver_test.go:495: 
        	Error Trace:	gameserver_test.go:495
        	Error:      	Not equal: 
        	            	expected: "Reserved"
        	            	actual  : "Ready"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(v1.GameServerState) (len=8) "Reserved"
        	            	+(v1.GameServerState) (len=5) "Ready"
        	            	 
        	Test:       	TestGameServerReserve
```

markmandel:
Still happening - I wonder what is going on :thinking: 

```
time="2020-01-22 01:14:44.524" level=info msg="Scaling fleet" fleet=simple-fleet-mkb2n patch="[{ \"op\": \"replace\", \"path\": \"/spec/replicas\", \"value\": 1 }]" scale=1
--- FAIL: TestGameServerReserve (123.91s)
    gameserver_test.go:494: 
        	Error Trace:	gameserver_test.go:494
        	Error:      	Received unexpected error:
        	            	timed out waiting for the condition
        	            	waiting for GameServer to be Reserved default/udp-server4f5xf
        	            	agones.dev/agones/test/e2e/framework.(*Framework).WaitForGameServerState
        	            		/go/src/agones.dev/agones/test/e2e/framework/framework.go:131
[...]

PR #1292, opened 2020-01-22
Show GameServer name in TestGameServerReserve flakiness fail

markmandel:
Extra debugging, so we can dig into what is going wrong in this test.

Work on #1276

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -491,8 +491,8 @@ func TestGameServerReserve(t *testing.T) {
 	assert.Equal(t, "ACK: RESERVE\n", reply)
 
 	gs, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateReserved, time.Minute)
-	assert.NoError(t, err)
-	assert.Equal(t, agonesv1.GameServerStateReserved, gs.Status.State)
+	assert.NoError(t, err, fmt.Sprintf("GameServer Name: %s", readyGs.ObjectMeta.Name))
+	assert.Equal(t, agonesv1.GameServerStateReserved, gs.Status.State, fmt.Sprintf("GameServer Name: %s", readyGs.ObjectMeta.Name))
 
 	// it should go back after 10 seconds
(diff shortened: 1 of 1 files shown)


test/e2e/gameserver_test.go
```diff
@@ -491,13 +491,16 @@ func TestGameServerReserve(t *testing.T) {
...
+	assert.NoError(t, err, readyGs.ObjectMeta.Name)
+	assert.Equal(t, agonesv1.GameServerStateReserved, gs.Status.State, readyGs.ObjectMeta.Name)
 
 	// it should go back after 10 seconds
 	gs, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateReady, 15*time.Second)
 	assert.NoError(t, err)
 	assert.Equal(t, agonesv1.GameServerStateReady, gs.Status.State)
+
+	logrus.Info("SLEEPING")
+	time.Sleep(time.Minute)
```
roberthbailey: Was the sleep for local development? I don't think we want to commit that.

test/e2e/gameserver_test.go
```diff
@@ -491,13 +491,16 @@ func TestGameServerReserve(t *testing.T) {
 	assert.Equal(t, "ACK: RESERVE\n", reply)
 
 	gs, err = framework.WaitForGameServerState(readyGs, agonesv1.GameServerStateReserved, time.Minute)
-	assert.NoError(t, err)
-	assert.Equal(t, agonesv1.GameServerStateReserved, gs.Status.State)
+	assert.NoError(t, err, readyGs.ObjectMeta.Name)
```
roberthbailey: I'm not exactly sure how this prints, but something like `fmt.Sprintf("ready game server name = %s", readyGs.ObjectMeta.Name)` might make this easier to understand in the failure output.

test/e2e/gameserver_test.go
markmandel: :man_facepalming:

test/e2e/gameserver_test.go
markmandel: Done on both!

Issue #1287, opened 2020-01-21
Gameserver failed to start in the namespace with runAsNonRoot pod security context.


Feng-Yin:
**What happened**:
Gameserver failed to start in the namespace with runAsNonRoot pod security context.

**What you expected to happen**:
Gameserver can be started normally.

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy a Gameserver into the namespace with runAsNonRoot pod security context.
2. ```kubectl describe pod <gameserver [pod>```](url) will give an error message: 
```Error: container has runAsNonRoot and image has non-numeric user (agones), cannot verify user is non-root```

**Anything else we need to know?**:
1. The link used for starting minikube with restrict setting: https://minikube.sigs.k8s.io/docs/tutorials/using_psp/
2. need to enable hostNetwork in restricted.PodSecurityPolicy
  ```
  hostNetwork: true
  hostPorts:
  - min: 7000
    max: 8000
  ```

**Environment**:
- Agones version: 1.1.0
- Kubernetes version (use `kubectl version`): 1.16.2
- Cloud provider or hardware configuration: local minikube v1.5.2
- Install method (yaml/helm): helm
- Troubleshooting guide log(s):
- Others:

**Comment by markmandel** [comment]
Thanks for the bug report!

Kubernetes 1.13 is the supported version for Agones at this time.
https://agones.dev/site/docs/installation/#usage-requirements

Does this error still occur on this version?

Also, can you please provide step by step instructions on how to reproduce?

The exact pod security policy you used, and the an example GameServer we can use to reproduce would be ideal.

Thanks!

**Comment by Feng-Yin** [comment]
The psp.yaml I'm using:
```
---
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: privileged
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: "*"
  labels:
    addonmanager.kubernetes.io/mode: EnsureExists
spec:
  privileged: true
  allowPrivilegeEscalation: true
  allowedCapabilities:
  - "*"
  volumes:
  - "*"
  hostNetwork: true
  hostPorts:
  - min: 0
    max: 65535
  hostIPC: true
  hostPID: true
  runAsUser:
    rule: 'RunAsAny'
  seLinux:
    rule: 'RunAsAny'
  supplementalGroups:
    rule: 'RunAsAny'
  fsGroup:
    rule: 'RunAsAny'
---
[...]

PR #1293, opened 2020-01-22
Use a numeric User ID for the "agones" user in the SDK sidecar

TBBle:
This fixes failure of the sdk sidecar container to start when run on a system with a `PodSecurityPolicy` specifying [`RunAsUser`](https://v1-13.docs.kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups) as "MustRunAsNonRoot".

Kubernetes cannot determine if a non-numeric User in the container metadata is 'root' or not.

Fixes: #1287

### DIFF
--- a/cmd/allocator/Dockerfile
+++ b/cmd/allocator/Dockerfile
@@ -15,10 +15,10 @@
 FROM alpine:3.11
 
 RUN apk --update add ca-certificates && \
-    adduser -D agones
+    adduser -D -u 1000 agones
 
 COPY --chown=agones:root ./bin/allocator /home/agones/allocator
 COPY --chown=agones:root ./bin/LICENSES ./bin/dependencies-src.tgz /home/agones/
 
-USER agones
+USER 1000
 ENTRYPOINT ["/home/agones/allocator"]
--- a/cmd/allocator/Dockerfile
+++ b/cmd/allocator/Dockerfile
@@ -17,8 +17,8 @@ FROM alpine:3.11
 RUN apk --update add ca-certificates && \
     adduser -D -u 1000 agones
 
-COPY --chown=agones:root ./bin/allocator /home/agones/allocator
-COPY --chown=agones:root ./bin/LICENSES ./bin/dependencies-src.tgz /home/agones/
+COPY --chown=agones:agones ./bin/allocator /home/agones/allocator
+COPY --chown=agones:agones ./bin/LICENSES ./bin/dependencies-src.tgz /home/agones/
 
 USER 1000
 ENTRYPOINT ["/home/agones/allocator"]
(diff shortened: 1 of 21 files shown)


TBBle:
@TBBle: When you come back to this, push out the same change to the other Docker files, thanks to @Feng-Yin for testing them all.

TBBle:
Note: Following up with my employer regarding CLA, which we have signed, but I am apparently not added to the authorised contributors list.

markmandel:
@TBBle this is awesome! Thanks for taking this on :+1: 

Are you going to the same for the all the Agones components (and maybe the examples?), either in this PR or others?

Just wondering if I should hold off on reviewing/merging this, in case you were going to expand it further.

TBBle:
I'll do the same thing for the other services, since my colleague has already tested that the same change applies to them and functions. I'll have a look at the examples, and do them if I'm confident. I'm currently waiting for the CLA to be sorted out at my end, so I'm hopeful for 'tomorrow' but not confident.

TBBle:
I've now updated the other Agones components (tested locally) and the examples (not tested locally). So this is all ready to go except the CLA, which is pending an action on our end.

markmandel:
Everything looks great - I want to run a quick check on the changes to the examples, but once I've done that, happy to approve!

markmandel:
Quick thought on this - if you want to have the example images be updated, please go through each example's Makefile and increment the version number :+1: 

If we're happy leaving the example images alone, then we can continue as is. (Probably better to increment though)

markmandel:
Smoke tested the change in the simple-udp example -- looks good!

TBBle:
I updated the versions in all the Makefiles. I also noticed I'd missed the unity-simple example, as it was running as root, and I'd been looking for `adduser` or `useradd` calls. So apart from that this is right to go pending the CLA (which I haven't heard back from at my end yet). If I have more time before the CLA list is updated, I'll take a look at the Unity sample, but I don't think it should otherwise block this PR.

markmandel:
Sounds good to me! I can take care of updating the images, and pushing them up. :+1: they are probably due for a review to make sure they all still work.

TBBle:
@googlebot I signed it!

markmandel:
Just having a look @TBBle - seems like your account is still not added [as a contributor](https://cla.developers.google.com/about#add-contributors) :disappointed:  it might take some time though to propagate, I'll take another look tomorrow.

TBBle:
It's a corporate CLA, apparently we're waiting to hear back from someone at Google. I'll keep chasing at my end.

TBBle:
@googlebot I signed it!

This time I *can* see it on https://cla.developers.google.com/clas

PR #1303, opened 2020-01-29
Extend Agones Unreal SDK

WVerlaek:
- Implements additional SDK methods: `SetLabel`, `SetAnnotation`, and `GetGameServer`.
- Makes the Agones Hook publicly accessible, allowing the SDK methods to be called from game logic.
- Adds retrying of HTTP requests (except health). Number of retries is configurable in the Agones plugin settings.

Tested by:
- Running `TESTS=ready,health,gameserver,setlabel,setannotation make run-sdk-conformance-local`
- Calling the SDK methods in AgonesHook from game logic inside the UnrealEditor and verifying the results
- Creating a Debug build of the game server and deploying it in a fleet, verifying Ready, Health and SetLabel/SetAnnotation are called correctly

Example `SetLabel` requests based off data received by `GetGameServer` (output from SDK conformance test):
image


markmandel:
1. This looks awesome. Thank you so much for doing this work! :man_cartwheeling: 
2. Glad to see you got the test harness working. Doesn't have to be this PR, but any thoughts on building a Make target so we can all run it?
3. You say you built a Debug test server? Any chance you want to contribute that to ./examples? (also can be a subsequent PR)

I can take a look through, but @devjgm you are our usual C++ person (I believe Unreal is different???) do you have time to do a review?

markmandel:
I had a quick look - barely means anything to me :smile: 

@padhansell - since it looks like you are doing a review (thanks!) please let us know when you think this is good to go, and I'll approve. :+1:

sdks/unreal/Agones/Source/Agones/AgonesHook.cpp
```diff
@@ -34,15 +40,24 @@ FAgonesHook::FAgonesHook()
...
+	HttpRetryManager = MakeShared<FHttpRetrySystem::FManager>(
+		FHttpRetrySystem::FRetryLimitCountSetting(RetryLimitCount),
+		FHttpRetrySystem::FRetryTimeoutRelativeSecondsSetting());
+
+	UE_LOG(LogAgonesHook, Log, TEXT("Initialized Agones Hook, Sidecar address: %s, Health Enabled: %s, Health Ping: %f, Debug: %s, Request Retry Limit: %d")
 		, *SidecarAddress
 		, (Settings->bHealthPingEnabled ? TEXT("True") : TEXT("False"))
 		, Settings->HealthPingSeconds
-		, (Settings->bDebugLogEnabled ? TEXT("True") : TEXT("False")));
+		, (Settings->bDebugLogEnabled ? TEXT("True") : TEXT("False"))
```
padhansell: @WVerlaek we discussed this yesterday but adding here for posterity!

In a subsequent PR we should probably ditch `bDebugLogEnabled` and instead just make use of the log levels in Unreal's logging system, which would:
a) Allow developers to change the logging behaviour through .ini configuration files, and on-the-fly through console commands. As an example, through the in-game console (or remotely using the "messaging" system), you could run the command `log LogAgonesHook error` to suppress any logs with a severity less than `Error`
b) Make the plugin more consistent with the rest of the engine
c) Remove some code

WVerlaek:
> Glad to see you got the test harness working. Doesn't have to be this PR, but any thoughts on building a Make target so we can all run it?

> You say you built a Debug test server? Any chance you want to contribute that to ./examples? (also can be a subsequent PR)

Thanks for taking a look! Not sure how easy it would be to create a make target or example for the unreal plugin, might potentially be quite a large undertaking. I've been testing the plugin using an existing unreal engine project, a new sample project would need to be created as well (which could be quite a lot to check in to the repo?)

sdks/unreal/Agones/Source/Agones/AgonesHook.cpp
domgreen: Agree, standardize on the engine and how UE4 enginers' work would be fantastic in a follow-up. 

Mentioned in the past that blueprint support would be fantastic.

WVerlaek:
Have checked that compiling for Editor and `Shipping` builds work

markmandel:
> I've been testing the plugin using an existing unreal engine project, a new sample project would need to be created as well (which could be quite a lot to check in to the repo?)

If there is a way to maybe do a skeleton project? I.e. something like "create a new project, and then copy these files into this place" ? 

That being said, the Unity sample is 62M, which is fine - how big are we talking?

Issue #1156, opened 2019-10-28
RollingUpdate shuts down all servers at once


KamiMay:
**What happened**: I adjusted RollingUpdate settings to replace each Ready GameServer pod one by one using following configuration
```
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
```
But every time I deployed Fleet configuration with this setting it would instantly shut down all Ready servers and replace them with new version.

1) Deployed my yaml and ran `kubectl get gs` got following 
```
NAME                             STATE       ADDRESS      PORT   NODE                       AGE
lobby-server-odd-2vmld-2m6h9        Shutdown    34.89.86.    7442   gke-qa-default-pool-3943   39s
lobby-server-odd-2vmld-6mlms        Shutdown    35.246.21.   7812   gke-qa-default-pool-rkw7   39s
lobby-server-odd-2vmld-9n8z9        Shutdown    35.235.55.   7382   gke-qa-default-pool-hjgr   1m
lobby-server-odd-2vmld-c6wzw        Shutdown    35.235.55.   7104   gke-qa-default-pool-hjgr   39s
lobby-server-odd-2vmld-pjttg        Shutdown    35.235.55.   7294   gke-qa-default-pool-hjgr   39s
lobby-server-odd-2vmld-pxqqs        Shutdown    35.235.53.   7139   gke-qa-default-pool-1wwf   39s
lobby-server-odd-2vmld-rb755        Shutdown    34.89.86.    7912   gke-qa-default-pool-3943   39s
lobby-server-odd-2vmld-xqbdl        Shutdown    35.242.175   7098   gke-qa-default-pool-bhwl   39s
lobby-server-odd-2vmld-zrl4n        Shutdown    35.246.21.   7367   gke-qa-default-pool-rkw7   39s
lobby-server-odd-2vmld-zw448        Ready       35.246.21.   7027   gke-qa-default-pool-rkw7   39s
```

2) Refreshing
```
NAME                             STATE       ADDRESS      PORT   NODE                       AGE
lobby-server-odd-2vmld-zw448        Ready       35.246.21.   7027   gke-qa-default-pool-rkw7   57s
```

3) Refreshing
```
NAME                             STATE       ADDRESS      PORT   NODE                       AGE
lobby-server-odd-2vmld-666zz        Scheduled   34.89.86.    7053   gke-qa-default-pool-3943   4s
[...]

PR #1305, opened 2020-01-30
Do not remove replica field if set to zero

pooneh-m:
Starting from Agones v1.2 the replica field was made optional. The API is also tagged with omitempty, meaning for JSON serialization the field is omitted if set to zero. omitempty tag makes the API backward incompatible with Agones v1.1.


markmandel:
If we make this change, how do we update Replicas, without changing the replica count?

This is particularly important for Fleets that are controlled by Autoscalers.

pooneh-m:
> this change, how do we update Replicas, without changing the replica count?
> 
> This is particularly important for Fleets that are controlled by Autoscalers.

Oh, thanks for reminding me. I forgot the use-case. I'll close the PR then.

markmandel:
I'd like to reopen this for discussion, because I was thinking about this more, and I _think_ adding omitemtpy might have been a mistake:

- If you want to make a change through the Go API without the replica field, you have PATCH (which is really what it is for)
- In the case of the autoscaler, if it changes the replicas field before you update, the K8s api will refuse to update the GameServer resource until you are on the same generation anyway
- Having omitempty makes it super hard to use the go library to scale a fleet to 0 without using a Patch command.

So I'm going to reopen this, mainly as well to see if all the e2e tests pass, and to hear if anyone has additional thoughts and opinions.

Sorry it took me so long to come around on this. Sometimes it takes me a while. :disappointed:

markmandel:
It might be worth noting that there is a [UpdateScale](https://github.com/googleforgames/agones/blob/master/pkg/client/clientset/versioned/typed/agones/v1/fleet.go#L49) interface as well for Fleets, so you _could_ use that to scale to 0.

@aLekSer do you have thoughts on this?

aLekSer:
I will test this steps again [RollingUpdate shuts down all servers at once](https://github.com/googleforgames/agones/issues/1156)
I have added a E2E test for this TestFleetAutoScalerRollingUpdate at https://github.com/googleforgames/agones/pull/1168/files

markmandel:
> I will test this steps again on this PR changes [RollingUpdate shuts down all servers at once](https://github.com/googleforgames/agones/issues/1156)
> A E2E test was added for this TestFleetAutoScalerRollingUpdate at https://github.com/googleforgames/agones/pull/1168/files

Thanks! Looks like e2e tests have passed.

This shouldn't affect the kubectl test you did previously, as it doesn't route through our go code - it talks directly to the Kubernetes API. That being said, let's still double check to make sure.

If the kubectl test passes for you @aLekSer , I recommend we merge this PR.

aLekSer:
RollingUpdate should be working, I will follow the docs tomorrow to see that there are now regressions, there are couple of options how to change a Fleet.
As for now I have tested on this change with this Fleet yaml:
```kind: Fleet                                                                                                                                                                                      [26/4748]
metadata:
  name: fleet-example
spec:
  scheduling: Packed
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 1
BY-IT00060:agones alexander.apalikov$ q
BY-IT00060:agones alexander.apalikov$
BY-IT00060:agones alexander.apalikov$ cat ./fl.yaml
apiVersion: "agones.dev/v1"
kind: Fleet
metadata:
  name: fleet-example
spec:
  scheduling: Packed
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
  template:
    metadata:
      labels:
        foo: ba2r
    spec:
      ports:
      - name: default
        portPolicy: Dynamic
        containerPort: 26000
      health:
        initialDelaySeconds: 30
        periodSeconds: 60
      template:
        spec:
          containers:
          - name: simple-udp
            image: gcr.io/agones-images/udp-server:0.16
---
apiVersion: "autoscaling.agones.dev/v1"
kind: FleetAutoscaler
metadata:
  name: server-autoscaler
spec:
  fleetName: fleet-example
  policy:
    type: Buffer
    buffer:
      bufferSize: 5
      minReplicas: 5
      maxReplicas: 10
```
If I change the label and do `kubectl apply -f ./f.yaml` again, rollingUpdate keeps working.

markmandel:
@aLekSer gentle bump - have you had a chance to test this out yet?

aLekSer:
Hello, partly yes, will update ASAP.

aLekSer:
Regarding the docs, I have tested with this PR `kubectl edit `:
```
When Fleet update contains only changes to the replicas parameter, then new GameServers will be created/deleted straight away, which means in that case maxSurge and maxUnavailable parameters for a RollingUpdate will not be used.
```
And the opposite if we have replicas and other field change than RollingUpdate is still working.

However when I use fleet with fleetautoscaler than we break the doc statement:
```
1. With kubectl apply: you should omit replicas parameter in a Fleet Spec before re-applying the Fleet configuration.
```
I did created a fleet with maxSurge one and `replicas: 5`, FAS with buffer of 5. Allocated one GS and then update the label scaling from 1 to 5 was in one second time, it should be step by step: remove one - add one:
```
5m41s       Normal    ScalingGameServerSet    Fleet             Scaling active GameServerSet flat-example-2rpmh from 5 to 6                                                                              
2m29s       Normal    ScalingGameServerSet    Fleet             Scaling inactive GameServerSet flat-example-2rpmh from 6 to 5                                                                            
2m26s       Normal    ScalingGameServerSet    Fleet             Scaling inactive GameServerSet flat-example-2rpmh from 5 to 4                                                                            
2m20s       Normal    ScalingGameServerSet    Fleet             Scaling inactive GameServerSet flat-example-2rpmh from 4 to 3                                                                            
2m11s       Normal    CreatingGameServerSet   Fleet             Created GameServerSet flat-example-t7bv5                                                                                                 
2m10s       Normal    ScalingGameServerSet    Fleet             Scaling inactive GameServerSet flat-example-2rpmh from 3 to 2                                                                            
2m10s       Normal    ScalingGameServerSet    Fleet             Scaling active GameServerSet flat-example-t7bv5 from 1 to 2                                                                              
2m10s       Normal    ScalingGameServerSet    Fleet             Scaling active GameServerSet flat-example-t7bv5 from 2 to 3                                                                              
2m10s       Normal    ScalingGameServerSet    Fleet             Scaling active GameServerSet flat-example-t7bv5 from 3 to 4                                                                              
2m9s        Normal    ScalingGameServerSet    Fleet             Scaling active GameServerSet flat-example-t7bv5 from 4 to 5     
```

aLekSer:
Tested this on version 1.3.0 - same issue applies to it, so docs was inaccurate.
https://gist.github.com/aLekSer/f0da2fb8a5a43d170d0dd711bfbd6f33
Just as a side note. `Replicaset` does contain omitempty:
```
// ReplicaSetSpec is the specification of a ReplicaSet.
type ReplicaSetSpec struct {
	// Replicas is the number of desired replicas.
	// This is a pointer to distinguish between explicit zero and unspecified.
	// Defaults to 1.
	// More info: https://kubernetes.io/docs/concepts/workloads/controllers/replicationcontroller/#what-is-a-replicationcontroller
	// +optional
	Replicas *int32 `json:"replicas,omitempty" protobuf:"varint,1,opt,name=replicas"`
```
As well as `DeploymentSpec`
https://github.com/kubernetes/kubernetes/blob/7e8f31b31318104f0288350652eaca0b1c84d4fc/staging/src/k8s.io/api/apps/v1/types.go#L274

markmandel:
> I did created a fleet with maxSurge one and replicas: 5, FAS with buffer of 5. Allocated one GS and then update the label scaling from 1 to 5 was in one second time, it should be step by step: remove one - add one:

Sounds like we should create a bug to track this separately.

> Mark @markmandel do we need to change Replicas to pointer at some point?

Using a pointer seems like an elegant solution, in that while it does change the API, it won't actually break functionality.

That being said, we need to work out what the root cause of the issue is here. `kubectl apply` doesn't go through our go code, so there's something in our controller(s) that is causing this, and see if that change will solve the problem - otherwise there is probably no point in making the change at this stage.

PR #1312, opened 2020-02-01
Player Tracking Proto and Go stubs

markmandel:
Implementation of the alpha functions for player tracking in the proto definition.

This includes a interface and non-working stub for the Go gRPC clients as well.

Copyright notices where updated to 2020 on the other languages as they were regenerated during the process.

Work on #1033


pkg/sdkserver/alpha/localsdk.go
```diff
@@ -14,7 +14,10 @@
 
 package alpha
 
-import "agones.dev/agones/pkg/sdk/alpha"
+import (
+	"agones.dev/agones/pkg/sdk/alpha"
+	"golang.org/x/net/context"
```
roberthbailey: Is there a reason you can't just import "context" (https://golang.org/pkg/context/)?

pkg/sdkserver/alpha/sdkserver.go
```diff
@@ -26,3 +29,28 @@ type SDKServer struct{}
 func NewSDKServer() *SDKServer {
 	return &SDKServer{}
 }
+
+// PlayerConnect should be called when a player connects
+func (s SDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerId) (*alpha.Empty, error) {
+	panic("implement me")
+}
+
+//PlayerDisconnect should be called when a player disconnects
```
roberthbailey: nit: Add a space between // and the start of the comment

roberthbailey: Also a super nit, but comments should be formatted as sentences ending with a period.

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

proto/sdk/alpha/alpha.proto
```diff
@@ -20,4 +20,56 @@ option go_package = "alpha";
...
+message Empty {
+}
+
+// Store a count variable
+message Count {
+    int64 count = 1;
+}
+
+// The unique identifier for a given player
+message PlayerId {
```
roberthbailey: I'm wondering if a single string is sufficient for this. What if you have multiple types of "players"? With a single string field the programmer will end up doing some funny string concatenation / parsing logic to shove it all into one field. If we had the option for some generic metadata (labels?), then you could use that instead to namespace different types of players.

I'm not necessarily saying we should change it right now (so we don't fall into the trap of over-designing), but I'm wondering if there is a way for us to get some feedback from folks that would actually use this. Maybe the best way is to get it in as alpha so folks can start playing around with it?

pkg/sdkserver/alpha/localsdk.go
markmandel: You know what - my IDE always brought in x/net/context and I never questioned it - I assumed there was a dependency there in gRPC... and I was totally wrong.

Excellent spotting. Fixed!

proto/sdk/alpha/alpha.proto
markmandel: So this is the design that was outlined [in the accompanying ticket ](#1033), and nobody disagreed with it. (Also, in 1:1 discussions, everyone seemed okay with the design).

_puts opinionated hat on_ :tophat: I don't think it's Agone's job to be storing/transferring data about players - that's why we only use a playerId. Any other metadata about players a game system wants to store, they should store that separately, and attach it in that system to this unique id that is passed between Agones and other systems for identification purposes.

roberthbailey: Seems reasonable. It sounds like you've thought about this and gotten some feedback.

markmandel:
Gentle bump on review please

markmandel:
Can I please get review on this, and #1324 , because I'm currently stuck on player tracking until these get merged.

markmandel:
Thanks @devjgm but I need approval from someone with write access - so @pooneh-m or @roberthbailey at this point.

roberthbailey:
Looking now.

pkg/sdkserver/alpha/sdkserver.go
```diff
@@ -14,15 +14,43 @@
 
 package alpha
 
-import "agones.dev/agones/pkg/sdk/alpha"
+import (
+	"agones.dev/agones/pkg/sdk/alpha"
+	"golang.org/x/net/context"
```
roberthbailey: just "context" as in the previous file.

pkg/sdkserver/alpha/localsdk.go
```diff
@@ -14,20 +14,49 @@
...
 	return &LocalSDKServer{}
 }
 
-// Close tears down all the things
+// Close tears down all the things.
 func (l *LocalSDKServer) Close() {
 	// placeholder in case things need to be shut down
 }
+
+// PlayerConnect should be called when a player connects
```
roberthbailey: nit: end comment sentence with a period.

(I was going to let this slide until I found other things to comment on. As long as you are fixing those, this should be an easy addition)

pkg/sdkserver/alpha/sdkserver.go
markmandel: Thank you!

PR #1317, opened 2020-02-04
Removing Interpolation-only expressions

chrisst:
Terraform 0.12.14 will start adding deprecation warnings for any strings that
only contain an interpolation as they can be replaced with straight references.

More at https://discuss.hashicorp.com/t/terraform-0-12-14-released/

This syntax has been available since Terraform 0.12 launched and it looks like 0.12 is the preferred version. Just in case I added a version constraint for


pooneh-m:
Thanks for the change. It is good to have a professional terraformer on the project. :)

PR #1322, opened 2020-02-05
Fix for `go mod vendor` command

aLekSer:
Some modules are not needed as well as some test files in a `vendor` directory.

See https://github.com/golang/go/issues/33353 for the reference


aLekSer:
This could be applied independently of PR #1321. Checked that result is the same.

PR #1324, opened 2020-02-05
CRD implementation of alpha player tracking

markmandel:
Update to CRD implementation, and pkg/api/gameserver.go and default value implementation.

Work on #1033


install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -148,4 +148,30 @@ properties:
...
+        properties:
+          players:
+            type: object
+            title: Configuration of player capacity
+            properties:
+              initialCapacity:
+                type: integer
+                title: The intial player capacity that this Game Server has
+                minimum: 0
+              webhook:
```
aLekSer: Do we need to add a CABundle as we have for FAS Webhook Policy?
https://agones.dev/site/docs/reference/agones_crd_api_reference/#WebhookPolicy

markmandel: That's an excellent question. I copied what we have in https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/crds/fleetautoscaler.yaml - and there is no reference to a caBundle. I assume there should be?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -277,15 +303,16 @@ func (gss *GameServerSpec) applyHealthDefaults() {
...
+func (gs *GameServer) applyStatusDefaults() {
 	if gs.Status.State == "" {
 		gs.Status.State = GameServerStateCreating
-		// applyStateDefaults() should be called after applyPortDefaults()
+		// applyStatusDefaults() should be called after applyPortDefaults()
 		if gs.HasPortPolicy(Dynamic) || gs.HasPortPolicy(Passthrough) {
 			gs.Status.State = GameServerStatePortAllocation
 		}
 	}
+	gs.Status.Alpha.Players.Capacity = gs.Spec.Alpha.Players.InitialCapacity
```
roberthbailey: Should this be wrapped in an if statement? Since this is an alpha feature, presumably support for it will initially be off by default. But this code will always fill in a status field (with a value of 0 since spec won't be possible to set), right?

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -333,6 +336,7 @@ func TestGameServerApplyDefaults(t *testing.T) {
 			assert.Equal(t, test.expected.scheduling, test.gameServer.Spec.Scheduling)
 			assert.Equal(t, test.expected.health, test.gameServer.Spec.Health)
 			assert.Equal(t, test.expected.sdkServer, test.gameServer.Spec.SdkServer)
+			assert.Equal(t, test.expected.alphaPlayerCapacity, test.gameServer.Status.Alpha.Players.Capacity)
```
roberthbailey: To my comment above, would it make more sense to verify that the status is not there at all unless we explicitly enable a feature flag?

pkg/apis/agones/v1/gameserver.go
markmandel: You know, I completed forgot that feature flags where globals :man_facepalming: 

I will make this change :+1:

markmandel:
Added the feature flag testing -- I had to add some locking to the feature flag implementation because of this. PTAL.

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
aLekSer: Thanks for noticing this, definitely it should be.

markmandel: I've implemented it here (awaiting to push up), I think caBundle should be added to fleetautoscaler validation in a separate PR, just to keep things separate. Sound good?

aLekSer: Thanks, now I see it.

markmandel:
This should be good for review again :+1:

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -148,4 +148,32 @@ properties:
...
+        type: object
+        title: Alpha properties for the GameServer
+        properties:
+          players:
+            type: object
+            title: Configuration of player capacity
+            properties:
+              initialCapacity:
+                type: integer
+                title: The intial player capacity that this Game Server has
```
roberthbailey: nit: I was always told not to end sentences with 'has'. Maybe rephrase as 'The initial player capacity of this Game Server'

pkg/apis/agones/v1/gameserver.go
```diff
@@ -150,6 +151,20 @@ type GameServerSpec struct {
 	SdkServer SdkServer `json:"sdkServer,omitempty"`
 	// Template describes the Pod that will be created for the GameServer
 	Template corev1.PodTemplateSpec `json:"template"`
+	// AlphaSpec describes the alpha properties for the GameServer
```
roberthbailey: nit: The variable is actually "Alpha". AlphaSpec is the type.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -150,6 +151,20 @@ type GameServerSpec struct {
...
+	Alpha AlphaSpec `json:"alpha,omitempty"`
+}
+
+// AlphaSpec is the alpha properties of the GameServer
```
roberthbailey: nit: end with a period.

Also, maybe s/is/contains ?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -150,6 +151,20 @@ type GameServerSpec struct {
...
+type AlphaSpec struct {
+	Players PlayersSpec `json:"players"`
+}
+
+// PlayersSpec tracks the initial player capacity, and what webhooks to send events to when there are
+// connection/disconnection events.
+type PlayersSpec struct {
+	InitialCapacity int64                          `json:"initialCapacity,omitempty"`
```
roberthbailey: I recall a discussion elsewhere about a concrete type vs a pointer in terms of compatibility. I want to double check that this is good as an int64 and shouldn't be a *int64.

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -319,8 +326,15 @@ func TestGameServerApplyDefaults(t *testing.T) {
 		},
 	}
 
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
+
+	// otherwise the race condition detector is not happy.
```
roberthbailey: Otherwise refers to what here?

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -319,8 +326,15 @@ func TestGameServerApplyDefaults(t *testing.T) {
 		},
 	}
 
+	runtime.FeatureTestMutex.Lock()
```
roberthbailey: This locks it from changing, but the test doesn't set it to an explicit value. Should the test

1. Lock the mutex
2. Capture the current value
3. Set a specific value for the test
4. Defer resetting the current value and unlocking the mutex

?

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
markmandel: Words good! :+1:

pkg/apis/agones/v1/gameserver.go
markmandel: Nice catch!

pkg/apis/agones/v1/gameserver.go
markmandel: "contains" is good. Done.

pkg/apis/agones/v1/gameserver.go
markmandel: This is an interesting question. I believe you are referring to #1325.

So I thought about this - and my thought where the following on making it a pointer:
- This is only the initial value to populate the status with, it will never get updated. So at no point will you want to update the GameServer with a different value (or if you did, it shouldn't do anything).
- If you aren't using player tracking, it will default to 0, which will default the status value to 0 which won't impact anything because you aren't using player tracking (and would be what we would likely default metric values to, even if we do make it a pointer (i.e. nil == 0).

So I left it as a concrete type that would default to 0.

WDYT?

pkg/apis/agones/v1/gameserver_test.go
markmandel: That should actually go away. That was leftover from a previous locking strategy attempt.

pkg/apis/agones/v1/gameserver_test.go
markmandel: The value for the Flags gets reset and set the appropriate value for the test here:
https://github.com/googleforgames/agones/pull/1324/files/641016f2ef479806519854856adb95150ee04007#diff-71e2ed803fe1fca8402cc9afb9117956R335

`ParseFeatures` will reset the entire feature flag set on each pass through, so on any test that requires an explicit Feature Flag setup, then `ParseFeatures` should be all that's required.

I'm also okay with the flags remaining post test - if a test fails because of a different Feature Flag than expected - then it should be explicitly setting that value, and seems like a valid test failure. Do you feel like it should return to a default state at the end of a test?

Does that cover the above?

pkg/apis/agones/v1/gameserver.go
roberthbailey: > This is only the initial value to populate the status with, it will never get updated.

Is that going to be enforced? e.g. I wouldn't be able to `kubectl edit gs my-gameserver` and change this part of the spec from 10 -> 20? Would that be rejected by validation or just ignored? If I need a different player capacity is the only way to get it to create a new game server / fleet?

> If you aren't using player tracking, it will default to 0, which will default the status value to 0

That sgtm.

markmandel: > Is that going to be enforced? e.g. I wouldn't be able to kubectl edit gs my-gameserver and change this part of the spec from 10 -> 20? Would that be rejected by validation or just ignored?

I was probably going to ignore it, but we could throw a validation error. I don't have strong opinions. Happy to add to `GameServer.ValidateUpdate` a validation check if you feel it should be there.

> If I need a different player capacity is the only way to get it to create a new game server / fleet?

The current capacity it stored in the GameServer.Status. In the [design](https://github.com/googleforgames/agones/issues/1033#issuecomment-571433206), `SDK.Alpha.SetPlayerCapacity(int capacity)` to update the capacity from within the GameServer. But you could also do it from the K8's API to edit the status value.

roberthbailey: I think a validation error would make sense. Otherwise we ignore the new value but don't let the user know that we are ignoring it.

pkg/apis/agones/v1/gameserver_test.go
roberthbailey: Yes, thanks for clarifying.

pkg/apis/agones/v1/gameserver.go
markmandel: SGTM. Implementing!

markmandel: Erk. Just realised - we actually don't do _any_ validation on GameServers on Update (we only validate on Create). 

See: https://github.com/googleforgames/agones/blob/master/pkg/gameservers/controller.go#L147

We have an issue for this #530

But never got deep into it, as there were a few edge cases for it.

Do we want to defer this work if/when we return to #530 ?

roberthbailey: That's reasonable. We should add a note to #530 and think about re-prioritizing it if this becomes a point of friction for users.

markmandel: Note added :+1:

PR #1328, opened 2020-02-08
Flakiness: TestGameServerAllocationDeletionOnUnAllocate

### LINKED ISSUE #1326: Flaky: TestGameServerAllocationDeletionOnUnAllocate
This one seems odd to me, and only happens sometimes.

```
--- FAIL: TestGameServerAllocationDeletionOnUnAllocate (0.18s)
    gameserverallocation_test.go:388: 
        	Error Trace:	gameserverallocation_test.go:388
        	Error:      	Expected nil, but got: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:""}, Status:"Failure", Message:"the server is currently unable to handle the request (post gameserverallocations.allocation.agones.dev)", Reason:"ServiceUnavailable", Details:(*v1.StatusDetails)(0xc0000e41e0), Code:503}}
        	Test:       	TestGameServerAllocationDeletionOnUnAllocate
```

More context: 
https://console.cloud.google.com/gcr/builds/1c60176f-02b8-421d-8218-451a3e566cd6?project=258182270954

markmandel:
Realised that Go test will compile and run tests in parallel per file,
which means we could shutdown the controller while an allocation is in
process - which means, it will fail completely, as the Kubernetes API
cannot reach the controller's http endpoint for allocation.

Split the controller tests into a separate package which could be run
with -parallel=1 so that there can be no race condition.

Also did some grouping of e2e test functionality to reduce code
repetition.

Closes #1326

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -239,11 +239,16 @@ test-go:
 # Runs end-to-end tests on the current configured cluster
 # For minikube user the minikube-test-e2e targets
 test-e2e: $(ensure-build-image)
-	echo "Starting e2e test runner!"
-	$(GO_TEST) $(agones_package)/test/e2e $(ARGS) $(GO_E2E_TEST_ARGS) \
+	echo "Starting e2e integration test!"
+	$(GO_TEST) $(ARGS) $(agones_package)/test/e2e $(GO_E2E_TEST_ARGS) \
 		--gameserver-image=$(GS_TEST_IMAGE) \
 		--pullsecret=$(IMAGE_PULL_SECRET)
-	echo "Finishing e2e test runner!"
+	echo "Finishing e2e integration test!"
+	echo "starting e2e controller failure test"
+	$(GO_TEST) -parallel=1 $(agones_package)/test/e2e/controller $(GO_E2E_TEST_ARGS) \
+    		--gameserver-image=$(GS_TEST_IMAGE) \
+    		--pullsecret=$(IMAGE_PULL_SECRET)
+	echo "Finishing e2e controller failure test!"
 
 # Runs end-to-end stress tests on the current configured cluster
 # For minikube user the minikube-stress-test-e2e targets
--- a/test/e2e/controller/crash_test.go
+++ b/test/e2e/controller/crash_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package e2e
+package controller
 
 import (
 	"testing"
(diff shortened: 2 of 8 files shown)


markmandel:
Got bit by different flaky bug #1276 - running again.

markmandel:
Running a bunch of tests to make sure it's working, and if that passes, will cleanup and submit.

test/e2e/controller/missing_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package e2e
+package controller
```
pooneh-m: Can we rename the file name to something else? missing_test.go sounds like it includes the tests that were missing from test coverage :D

test/e2e/controller/missing_test.go
```diff
@@ -75,8 +75,67 @@ func deleteAgonesControllerPods() error {
...
+	})
+}
+
 // getAgonesControllerPods returns all the Agones controller pods
 func getAgonesControllerPods() (*corev1.PodList, error) {
 	opts := metav1.ListOptions{LabelSelector: labels.Set{"agones.dev/role": "controller"}.String()}
 	return framework.KubeClient.CoreV1().Pods("agones-system").List(opts)
 }
+
+func defaultGameServer(namespace string) *agonesv1.GameServer {
```
pooneh-m: Can we refactor [this](https://github.com/googleforgames/agones/blob/master/test/e2e/gameserver_test.go#L577) instead?

markmandel: Yeah I thought about centralising that too - moving it into framework maybe. Wasn't sure if 2 duplication was enough to warrant it. If you think so, I'm happy to do it.

I assume that is what you mean here?

test/e2e/controller/missing_test.go
markmandel: `crash_test.go` ? That better?

Tests where the controller crashes for whatever reason? That sound better?

pooneh-m: Thanks, much better!

test/e2e/controller/missing_test.go
pooneh-m: Correct, thanks for the clarification. framework.go should be a good place or you can add a new file such as testhelper and centralize the common helper methods there?

test/e2e/controller/missing_test.go
markmandel: SGTM! Will make the change.

test/e2e/controller/missing_test.go
markmandel: Sounds good - I like the idea of testhelper.go or similar (I think we have a previous pattern for this somewhere, maybe, I can copy).

PR #1334, opened 2020-02-11
Extra Debugging for TestGameServerReserve

markmandel:
Adds new WaitForGameServerStateWithLogger() function that makes it easier to track debug output for state checking for specific tests.

Also dumped out event data for the pod and gameserver on this test to see what else we can see.


markmandel:
Marking this as Ready for review (a) to try and get more data on why this is failing and (b) there are some useful tools in here that might be valuable to others.

Issue #1330, opened 2020-02-10
Metrics: Export to Stackdriver is not working


aLekSer:
There are no stackdriver metrics due to error in Labels:
```
Failed to export to Stackdriver: rpc error: code = InvalidArgument
 desc = One or more TimeSeries could not be written:
 Unrecognized resource label: instance_id: timeSeries[2,3,12];
 Unrecognized resource label: namespace_id: timeSeries[1,11]; 
Unrecognized resource label: pod_id: timeSeries[0,5,7]; 
Unrecognized resource label: zone: timeSeries[4,6,8-10]
```
**What happened**:

No stackdriver metrics on the dashboard, which was working several month ago.
New errors on Agones Controller logs.

**What you expected to happen**:
Stackdriver metrics are correctly visualised.

**How to reproduce it (as minimally and precisely as possible)**:
https://agones.dev/site/docs/guides/metrics/#stackdriver-installation

**What should be done to fix an issue**
- [x] Update documentation to use not `Global` but `Kubernetes Container` resource for each of gamesever, and upload new screenshot from Stackdriver.
- [x] Fix labels as in 
https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/issues/92
- [ ] Fix `view.Distribution` with zero range value.
- [ ] Optionally: update Stackdriver exporter library.

**Anything else we need to know?**:
There are two Pull requests which solved mentioned above ticket
They contain fixes for:
```
func getMonitoredResource(projectID string) (*monitoredres.MonitoredResource, error) {
...
}
```

**Environment**:
- Agones version: 
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Install method (yaml/helm):
- Troubleshooting guide log(s):
- Others:

aLekSer:
We can switch from getMonitoredResource() function to `monitoredresource.Autodetect()` after updating to version 0.22 of opencensus and  `contrib.go.opencensus.io/exporter/stackdriver v0.12.0` as it is done in latest example here:
https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/6ee7f9652d2a9e707fea22c56d06235db6289426/examples/stats/main.go#L51

markmandel:
@bbf Have you seen this on recent releases?

bbf:
While I have not tested any recent releases, I can imagine why a few things stopped working. I'm very interested in overhauling the Stackdriver integration of Agones, so if possible give me some time to look into it.

[...]

PR #1335, opened 2020-02-11
Stackdriver - fix getMonitoredResource

aLekSer:
Labels were changed for k8s_container monitored resource. Fix error message, update docs.
Tested on Stackdriver dashboard. Old dashboard's charts are not available and should be readded with `Kubernetes Container` resource instead of `Global`

Part of #1330.

### DIFF
--- a/pkg/metrics/exporter.go
+++ b/pkg/metrics/exporter.go
@@ -95,10 +95,6 @@ func SetReportingPeriod(forPrometheus, forStackdriver bool) {
 }
 
 func getMonitoredResource(projectID string) (*monitoredres.MonitoredResource, error) {
-	instanceID, err := metadata.InstanceID()
-	if err != nil {
-		return nil, errors.Wrap(err, "error getting instance ID")
-	}
 	zone, err := metadata.Zone()
 	if err != nil {
 		return nil, errors.Wrap(err, "error getting zone")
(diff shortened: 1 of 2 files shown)


aLekSer:
Finally CI passes flaky `make test` step and TestGameServerReserve. Will make additional test of Stackdriver dashboard.

aLekSer:
Tested on `agones/gameserver_count` metric, chart below:
GS_Count_by_State
ChartSettings

markmandel:
/assign @bbf 

Can you take a look? I'm totally not into metrics, but if you say this is good to go, happy to approve and push through.

cyriltovena:
Come here to say where is the doc of this ? I had hard time trying to find what should go inside this map and I used the doc IIRC.

aLekSer:
@cyriltovena There is a definition of these 2 types of Monitored Resource:
[Kubernetes Container - k8s_container](https://cloud.google.com/monitoring/api/resources#tag_k8s_container) and  [GKE Container](https://cloud.google.com/monitoring/api/resources#tag_gke_container) later is not working now in GKE cluster I am creating with `make gcloud-test-cluster`.

aLekSer:
I would add `metadata.system_labels.node_name` instead of `instance_id` here and test one more time

cyriltovena:
@aLekSer thanks for the clarification!

aLekSer:
This version with switching from `instance_id` key to a new one did not work:
```
2020/02/15 08:50:47 Failed to export to Stackdriver: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Field timeSeries[0].resource.labels had an invalid value of "metadata.system_labels.node_name": Label key contains invalid characters.: timeSeries[0];
```
I think we can skip this label.

bbf:
In the past I had not deeply inspected Agones' monitoring code, and now that I did, I fear that the Stackdriver integration will only work on very simple scenarios. Stackdriver has a different set of limitations to achieve its level of scalability, that Prometheus doesn't have. So I'm not sure if we should merge this fix, which probably will allow users to temporarily use Stackdriver, considering that we might have to introduce several breaking changes to really fix the integration. And because we are leveraging OpenCensus, the required breaking changes might impact both integrations.

@markmandel WTDY? I should be able to spend some time in early March to propose and implement the changes, or we can just commit this and break everything again soon.

markmandel:
What does "break everything again" actually mean? What actually breaks?

markmandel:
I'll bring up another point -- we have a RC coming out next week.

I'd hate for Stackdriver integration to just not work on the next release. That seems like the ultimate worst scenario.

What do you all think?

bbf:
> What does "break everything again" actually mean? What actually breaks?

Most likely changing the monitored resource being changed, as well as many of the metric definitions.

> I'll bring up another point -- we have a RC coming out next week.
> 
> I'd hate for Stackdriver integration to just not work on the next release. That seems like the ultimate worst scenario.

Given that the breakage will be unavoidable, and this is an incremental improvement towards a working solution, I think we can merge this.

LGTM

PR #1345, opened 2020-02-18
Unreal SDK add Allocate + Reserve and changes to the plugin settings

WVerlaek:
Adds the `Allocate` and `Reserve` methods to the Unreal plugin, and an option to configure whether `Ready` should be sent automatically at startup.

Also removed the debug logging option in the settings following up from https://github.com/googleforgames/agones/pull/1303#discussion_r372858159 to be more consistent with UnrealEngine logging behaviour (and changed the default verbosity to Verbose)

### DIFF
--- a/sdks/unreal/Agones/Source/Agones/AgonesHook.cpp
+++ b/sdks/unreal/Agones/Source/Agones/AgonesHook.cpp
@@ -123,28 +123,26 @@ void FAgonesHook::SetLabel(const FString& Key, const FString& Value)
 {
 	FKeyValuePair Label = { Key, Value };
 	FString Json;
-	if (FJsonObjectConverter::UStructToJsonObjectString(Label, Json))
+	if (!FJsonObjectConverter::UStructToJsonObjectString(Label, Json))
 	{
-		SendRequest(SidecarAddress + SetLabelSuffix, Json, FHttpVerb::PUT, true);
+		UE_LOG(LogAgonesHook, Error, TEXT("Failed to set label, error serializing key-value pair (%s: %s)"), *Key, *Value);
+		return;
 	}
-	else
-	{
-		UE_LOG(LogAgonesHook, Error, TEXT("Failed to send request, error serializing key-value pair (%s: %s)"), *Key, *Value);
-	}
+
+	SendRequest(SidecarAddress + SetLabelSuffix, Json, FHttpVerb::PUT, true);
 }
 
 void FAgonesHook::SetAnnotation(const FString& Key, const FString& Value)
 {
 	FKeyValuePair Annotation = { Key, Value };
 	FString Json;
-	if (FJsonObjectConverter::UStructToJsonObjectString(Annotation, Json))
+	if (!FJsonObjectConverter::UStructToJsonObjectString(Annotation, Json))
 	{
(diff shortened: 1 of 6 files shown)


site/content/en/docs/Guides/Client SDKs/unreal.md
```diff
@@ -39,6 +39,6 @@ Available settings:
 
 - Health Ping Enabled. Whether the server sends a health ping to the Agones sidecar. (default: `true`)
 - Health Ping Seconds. Interval of the server sending a health ping to the Agones sidecar. (default: `5`)
-- Debug Logging Enabled. Debug logging for development of this Plugin. (default: `false`)
 - Request Retry Limit. Maximum number of times a failed request to the Agones sidecar is retried. Health requests are not retried. (default: `30`)
+- Send Ready at Startup. Automatically send a Ready request when the server starts. Disable this to manually control when the game server should be marked as ready. (default: `true`)
```
markmandel: Does there need to be more docs around how to use the the supplied methods, such as `Allocate`, `Ready`, `SetLabel` etc?

markmandel:
@domgreen, @padhansell - would either of you like to a gander and see if this is good to go please?

markmandel:
(hit rebuild on the CI. Failing on Flaky test)

site/content/en/docs/Guides/Client SDKs/unreal.md
domgreen: I tend to agree, it might be worth adding a paragraph on how these mehtods / structs are available to be set from within UE4Editor?

domgreen: Might also be worth adding a paragraph around using this locally ... eg. how we can turn off the healthcheck/ready checks when just launching via UE4Editor that makes it easier for most GameDevs to iterate fast before pushing into our clusters.

sdks/unreal/Agones/Source/Agones/AgonesHook.cpp
```diff
@@ -167,6 +175,25 @@ void FAgonesHook::GetGameServer(const FGameServerRequestCompleteDelegate& Delega
...
+	FDuration Duration = { Seconds };
+	FString Json;
+	if (FJsonObjectConverter::UStructToJsonObjectString(Duration, Json))
+	{
+		SendRequest(SidecarAddress + ReserveSuffix, Json, FHttpVerb::POST, true);
+	}
+	else
+	{
+		UE_LOG(LogAgonesHook, Error, TEXT("Failed to send reserve request, error serializing duration (%d)"), Seconds);
+	}
```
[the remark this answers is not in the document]
WVerlaek: don't believe unreal has a specific guideline for this but agree it looks better, will change

site/content/en/docs/Guides/Client SDKs/unreal.md
WVerlaek: Have added some docs on how to use the sdk methods

WVerlaek: For local dev (or for running outside an agones environment) I think the plugin could use some more work to support disabling executing the sdk methods / auto health checks. Was thinking about an additional config value `bAgonesEnabled` which can be set to `default`, `forceEnabled`, or `forceDisabled`, where `default` will disable calling out to the agones sidecar when running locally inside the unreal editor

PR #1346, opened 2020-02-18
Add more details on Allocate() behaviour

aLekSer:
There was lack of details what is the result of SDK.Allocate() function
call.

### DIFF
--- a/site/content/en/docs/Guides/Client SDKs/_index.md
+++ b/site/content/en/docs/Guides/Client SDKs/_index.md
@@ -56,6 +56,17 @@ functions that implement the core responsibilities of the SDK.
 For language specific documentation, have a look at the respective source (linked above), 
 and the {{< ghlink href="examples" >}}examples{{< /ghlink >}}.
 
+Calling any of state changing functions mentioned below does not guarantee that GameServer Custom Resource object would actually change its state right after the call. For instance, it could be moved to the `Shutdown` state elsewhere (for example, when a fleet scales down), which leads to no changes in `GameServer` object. You can verify the result of this call by waiting for the desired state in a callback to WatchGameServer() function.
+
+Functions which changes GameServer state or settings are:
+
+1. Ready()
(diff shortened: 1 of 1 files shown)


markmandel:
This is actually true for all setting functions in the SDK - should we make this a higher level comment in the docs (i.e. cover all mutation methods) - rather than make it specific to Allocate?

aLekSer:
Well, I also thought about that, I think we should move it higher. And for Allocate just point, so that it is not skipped. Because Allocate "pin" GameServer not to be deleted more than other states.

KamiMay:
This explains why we had issues using this SDK to allocate servers. We ended up reverting back to our original way of allocation.

site/content/en/docs/Guides/Client SDKs/_index.md
```diff
@@ -153,7 +153,9 @@ and the {{< ghlink href="examples" >}}examples{{< /ghlink >}}.
 ### Allocate()
 
 With some matchmakers and game matching strategies, it can be important for game servers to mark themselves as `Allocated`.
-For those scenarios, this SDK functionality exists. 
+For those scenarios, this SDK functionality exists.
+
+Calling `Allocate()` function does not guarantee that GameServer becomes `Allocated` after this call. For instance, it could be moved to the `Shutdown` state elsewhere (for example when fleet scales down). You can verify the result of this call by waiting for the `Allocated` Status in GameServer() function output.
```
markmandel: I would actually advocate using `Watch()` to react to when the Allocation actually happens, rather than using `GameServer()` (polling the GameServer doesn't hurt, since it's cached locally, but Watch would be better)

markmandel:
> Well, I also thought about that, I think we should move it higher. And for Allocate just point, so that it is not skipped. Because Allocate "pins" GameServer not to be deleted more than other states.

Yeah that makes sense - so basically, do both? I'm down for both.

PR #1348, opened 2020-02-19
Implement Alpha field best practices

### LINKED ISSUE #1347: Alpha field in the stable API should follow best practices
An [Alpha](https://github.com/googleforgames/agones/blob/master/pkg/apis/agones/v1/gameserver.go#L155) field is added but not following best practices of introducing an [unstable version to a stable API](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-unstable-features-to-stable-versions):

* The field should be a pointer
* Use // +optional tag
* omitempty for GameServerStatus.Alpha
* Protected by feature gate for the API, storage and validation.

markmandel:
Best practices outlined in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-unstable-features-to-stable-versions

Closes #1347


pkg/apis/agones/v1/gameserver.go
```diff
@@ -152,7 +152,8 @@ type GameServerSpec struct {
 	// Template describes the Pod that will be created for the GameServer
 	Template corev1.PodTemplateSpec `json:"template"`
 	// Alpha describes the alpha properties for the GameServer.
-	Alpha AlphaSpec `json:"alpha,omitempty"`
+	// +optional
+	Alpha *AlphaSpec `json:"alpha,omitempty"`
```
pooneh-m: Can you change the name to a meaningful name that will be used when the feature is stable?

markmandel: So as per our [Feature Stages](https://agones.dev/site/docs/guides/feature-stages/#new-crd-attributes):

> alpha and beta attributes will have a corresponding alpha or beta parent element in their configuration to delineate their feature stage.
>
> If no such parent exists, this attribute is a stable feature.

So when this becomes stable, it will move to the parent element and/or we can deprecate over stages (i.e. populate beta with alpha, stable with beta, etc)

pooneh-m: Then should the feature stage guideline be updated? I am in favor of using longer term and meaningful name for a feature than alpha, beta. Are you thinking of renaming it to stable when feature is stable?

markmandel: So the plan was to take the contents of AlphaSpec and move it into a similar BetaSpec when moving these items to beta, and then moving from a BetaSpec up to GameServerSpec when these items are stable.

So, I would advocate leaving the Feature Stage guide as is.

The idea being that we want it to be very, very clear on what is Alpha and Beta within a stable CRD , with the top level prefix - as otherwise I feel like it could easily be confusing which features are alpha/beta at first glance at CRD reference / documentation.

pooneh-m: Thanks for explaining. Feature gate should prevent the feature to be enabled by default. Even if someone don't bother reading doc to learn a field is mark as alpha and just set the field, that field will not work without enabling the feature. Based on k8s guide, it seems how k8s is operating.

markmandel: > Even if someone don't bother reading doc to learn a field is mark as alpha and just set the field, that field will not have an impact without enabling the feature.

Yeah, which I think is a very confusing user experience. I'd rather be overtly explicit in configuration, than have a silent failure, with no error message, that a user has to go work out what went wrong.

People miss documentation all the time, for a variety of reasons. If we have a choice to make something obvious, I'd always like to fall on that side.

I'm not seeing any downside to our current approach (with the top level alpha/beta elements) - especially with the changes you suggested that created this PR. Do you see potential downsides?

markmandel: @roberthbailey you also helped shape the feature gate documentation - do you have thoughts on the above?

markmandel: Actually, in case it's not clear. The feature name is now "Alpha", alpha is the top level element.

So in reality, the abbreviated configuration is (from the design in #1033):

```yaml
spec:
  # new configuration
  alpha:
    players:
      initialCapacity: 10 # sets the initial player capacity. Defaults to 0.
```

So it's still clear what is being configured.

pooneh-m: > > Even if someone don't bother reading doc to learn a field is mark as alpha and just set the field, that field will not have an impact without enabling the feature.
> 
> Yeah, which I think is a very confusing user experience. I'd rather be overtly explicit in configuration, than have a silent failure, with no error message, that a user has to go work out what went wrong.

Validation webhook?

> 
> People miss documentation all the time, for a variety of reasons. If we have a choice to make something obvious, I'd always like to fall on that side.
> 
> I'm not seeing any downside to our current approach (with the top level alpha/beta elements) - especially with the changes you suggested that created this PR. Do you see potential downsides?

The downside that I see is we move a feature to beta and without deprecation move it to stable release. That changes schema between beta and stable. The way it should work is beta is the candidate and your configuration should work without any change between last iteration on beta and stable versions.

markmandel: > Validation webhook? 

That would work for me :+1:  (most objects have a `Validate()` on create at least, which will pick up 98% of issues here) - but we would need to turn off what we have now, which is set the alpha property pointer to nil when the flag is not enabled, such as to catch it on the validation webhook side.

> The downside that I see is we move a feature to beta and without deprecation move it to stable release. That changes schema between beta and stable. The way it should work is beta is the candidate and your configuration should work without any change between last iteration on beta and stable versions.

So with your advocated approach we are optimising for users not having to have breaking changes as we move between feature stages, which I can get behind. And the Validation error approach provides visibility into when things go wrong. So that makes me happy too.

The only thing I get mildly stuck on, is that for everything else (crd versions, SDK versions, api endpoints), the user will need to change their CRD group name or endpoint path with each version change (admittedly that is a smaller change, I think? :man_shrugging: ) , and that the feature stage of the item in question is part of the item (in the group name/url path/etc). Whereas here we deviate from that path. Everything else would be a breaking change wherein a user has to specifically choose a version they are working against, but on a CRD that doesn't happen.

It feels like we're mixing approaches here a bit, which maybe is fine, but I just want to make sure we are aware of it if we choose to move forward with your proposal.

pooneh-m: @jkowalski, please chime in if you have an opinion on this.
Examples in [this](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-unstable-features-to-stable-versions) also use the long-term naming instead of Alpha.

>   // width indicates how wide the object is.
  // This field is alpha-level and is only honored by servers that enable the Frobber2D feature.
  // +optional
  Width  *int32 `json:"width,omitempty"`

pooneh-m: @jkowalski and @Yingxin-Jiang also confirmed that the name of an unstable field in a stable API should not be changing between the versions **by design** (e.g. Alpha -> Beta -> final naming).

markmandel: Sounds like we've reach consensus! I will jump on these changes, and changes to the Feature Guide as well, so that they are in place before 1.4 is released!

Thanks everyone, and thanks for bringing this up @pooneh-m !

markmandel:
This should be good to review again.

pooneh-m:
Very nice! Thanks for the change.

PR #1349, opened 2020-02-19
Update the kind make targets to work with kind 0.6.0 and later.

roberthbailey:
As described in the breaking changes section for the [0.6.0 release of kind](https://github.com/kubernetes-sigs/kind/releases/tag/v0.6.0):

> You do not need to and should not do export KUBECONFIG="$(kind get kubeconfig-path)". This command is deprecated and will be removed in a future release.

I recently re-installed kind following the current installation instructions and got version 0.7.0 which indeed does remove this command (and our existing make targets were thus rendered broken).


roberthbailey:
[testing our prow webhook, which had been failing]

PR #1358, opened 2020-02-24
Fix UE4 plugin compilation error in AgonesHook.h.

drichardson:
Moved AgonesHook.h and Model directory under Public, since two files, Model/GameServer.h
and Model/GameServerObjectMeta.h, are required to be public due to their
use in AgonesHook.h.

Arrange all C++ files under Private or Public directories to reduce
chance of this kind of problem happening again and to follow UE4
standard module layout.

## Error observed before fix

```
  Building 4 actions with 6 processes...
    [1/4] Module.Agones.cpp
    In file included from E:/UnrealEngine/botg/Intermediate/Build/Linux/B4D820EA/ShooterGameServer/Development/Agones/Module.Agones.cpp:2:
    In file included from E:/UnrealEngine/botg/Plugins/Agones/Source/Agones/Agones.cpp:15:
    E:\UnrealEngine\botg\Plugins\Agones\Source\Agones\Public\Agones.h(19,10): fatal error: 'AgonesHook.h' file not found
    #include "AgonesHook.h"
             ^~~~~~~~~~~~~~
    1 error generated.
```

## Testing Done

1. Create default UE4 4.24.2 C++ project.
2. Add Agones plugin
3. Compile with DefaultBuildSettings = BuildSettingsVersion.V2 and DefaultBuildSettings = BuildSettingsVersion.V1 in Target.cs.

Compilation succeeds.

No functional tests performed.


domgreen:
@drichardson would it be worth adding a README about the expected structure of UE4 plugins and a link to some docs ... when looking at this we found it hard to understand how to structure plugins.

drichardson:
> 
> 
> @drichardson would it be worth adding a README about the expected structure of UE4 plugins and a link to some docs ... when looking at this we found it hard to understand how to structure plugins.

Yes, I think it would be worth it so that at least all UE4 plugin contributors are on the same page and that documentation, where available, can be directly referenced.

I'll create a separate PR for this.

roberthbailey:
> would it be worth adding a README about the expected structure of UE4 plugins and a link to some docs

From the change it looks like the general idea is to put all header files into the public directory and all cpp files into the private directory. Is the guidance more complex than that?

roberthbailey:
The sdk conformance tests failed:
```
make[2]: *** [run-sdk-conformance-no-build] Error 1
includes/sdk.mk:139: recipe for target 'run-sdk-conformance-no-build' failed
make[1]: *** [run-sdk-conformance-test] Error 2
includes/sdk.mk:146: recipe for target 'run-sdk-conformance-test' failed
includes/sdk.mk:156: recipe for target 'run-sdk-conformance-test-go' failed
make: *** [run-sdk-conformance-test-go] Error 2
```
but since this SDK isn't part of the conformance tests, it's just a coincidence. 

I'm going to re-start the tests to see if it's a flake.

drichardson:
> 
> 
> > would it be worth adding a README about the expected structure of UE4 plugins and a link to some docs
> 
> From the change it looks like the general idea is to put all header files into the public directory and all cpp files into the private directory. Is the guidance more complex than that?

Yes, that's all the guidance that is needed for this change. The Public/Private separation is widely used in the engine's own modules, though the Private part is in fact optional, since anything not in Public (or one of the other explicitly listed in the build files *PublicIncludePaths* setting) will be private.

If you think the directory structure makes more documentation unnecessary, then fine. Otherwise let me know and I will create a README that explicitly talks about this.

Additionally, is there any way we can get a Windows machine to setup a build bot for the UE4 plugin so that we can at least catch compilation errors?

roberthbailey:
> If you think the directory structure makes more documentation unnecessary, then fine. Otherwise let me know and I will create a README that explicitly talks about this.

I think a short README would be really helpful, and it gives us a place to put other guidance we want to add in the future so that it's easy for new contributors to find.

> Additionally, is there any way we can get a Windows machine to setup a build bot for the UE4 plugin so that we can at least catch compilation errors?

This might be tricky. @markmandel and I have been talking about what it would take (licensing wise) to get some CI in place for UE and I don't think we have a satisfactory answer yet.

drichardson:
> 
> This might be tricky. @markmandel and I have been talking about what it would take (licensing wise) to get some CI in place for UE and I don't think we have a satisfactory answer yet.

In that case, can we at least add a pre-release checklist item to compile the plugin on\
- the minimum supported version of UE4
- the latest released version of UE4 (currently 4.24.2)

Otherwise, we will continue to have released versions of Agones whose UE4 SDKs don't even compile (e.g. https://github.com/googleforgames/agones/pull/1320).

domgreen:
Unsure why we didn't catch this locally ☹️ will investigate.

Would be great to have some pre-release checklists. Unreal licencing can be difficult to navigate.

Ludea:
I asked Azure to add Unreal on their platform [#367](https://github.com/microsoft/azure-pipelines-image-generation/issues/367)

Issue #1223, opened 2019-12-10
Create and document rules of thumb for log levels in agones code


roberthbailey:
As per https://github.com/googleforgames/agones/pull/1220 we should come up with some general thoughts on what should be info and what should be debug for log lines in the agones code. 

As mentioned during the last community meeting, too much logging can become expensive.

aLekSer:
Some statistics of the log messages (total 300 rows were recorded), messages with count > 5:

| Message  | Count |
| --------- | ----- |
| Processing | 48 |
| Enqueuing | 40 |
| Enqueuing immediately | 38 |
| syncGameServerSet | 27 |
| Reconciling GameServerSet | 27 |
| syncGameServerSet finished | 27 |
| Synchronising | 22 |
| Update Fleet Status | 20 |
| Added webhook handler | 9 |
| Starting workers... | 8 |
| Wait for cache sync | 6 |

https://docs.google.com/spreadsheets/d/1UYy6XELta2AU6X1P223xIlGsr2-aJXTpJEIdAZVVH4o/edit?usp=sharing

aLekSer:
More real statistics from @KamiMay 
https://github.com/googleforgames/agones/issues/1218#issuecomment-564064461

markmandel:
I really like the https://reflectoring.io/logging-levels/ article that @pooneh-m suggested in the other thread! I'm quite tempted to link to that in the development guide and practically call it a day.

One rule I think should be add though - full dumps of an object (such as a `GameServer`), or JSON packet into logrus should only be done on `Debug` or `Trace` levels.

We do this a bunch! (or I did this a bunch I should say :blush: ).

markmandel:
Maybe this is a tad heavy handed, but he vague rule of thumb I've been working on for the Agones controller at least is - if it has to do with the controller itself - starting, stopping, config etc - that's an Info level event.

If it has to do with a transient resource (fleet, gameserver etc) then that ends up being a Debug level event.

aLekSer:
@akremsa, can you please look is there some places which should be logged with level down or up? And probably make a change to `contributing.md` to document all of those.

akremsa:
Ok, I will take a look.

markmandel:
@akremsa the sdkserver should be a great place to start. I feel like there is a large chunk of `Info` logging that could be moved to `debug` :+1:

markmandel:
This looks like it's completed! If nobody objects, I'll close this next week.

PR #1359, opened 2020-02-24
Updated log levels in pkg

akremsa:
This PR can be considered as a part of #1223 

Updated log levels in pkg package.

### DIFF
--- a/pkg/sdkserver/localsdk.go
+++ b/pkg/sdkserver/localsdk.go
@@ -125,7 +125,7 @@ func NewLocalSDKServer(filePath string) (*LocalSDKServer, error) {
 
 	go func() {
 		for value := range l.update {
-			logrus.Debug("gameserver update received")
+			logrus.Debug("Gameserver update received")
 			l.updateObservers.Range(func(observer, _ interface{}) bool {
 				observer.(chan struct{}) <- value
 				return true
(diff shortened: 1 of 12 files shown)


pkg/sdkserver/localsdk.go
```diff
@@ -125,7 +125,7 @@ func NewLocalSDKServer(filePath string) (*LocalSDKServer, error) {
 
 	go func() {
 		for value := range l.update {
-			logrus.Info("gameserver update received")
+			logrus.Debug("gameserver update received")
```
aLekSer: Nit: Not connected to the logLevel, but for consistency with log messages above.
```suggestion
			logrus.Debug("GameServer update received")
```

CONTRIBUTING.md
```diff
@@ -101,3 +101,9 @@ repository, we have a [community membership guide](./docs/governance/community_m
 
 - [How to write a good Git Commit message](https://chris.beams.io/posts/git-commit/) -
   Great way to make sure your Pull Requests get accepted.
+- **Log levels usage:**
```
markmandel: Please move this entire section it's own PR. Where possible, smaller PRs are better, especially PRs that touch on multiple aspects.

Here we have a PR that is about log levels in SDK Server, but also touches on contributing guidelines. It's very hard to approve / reject both.

akremsa:
@markmandel  thanks for the review. I've removed changes from the md file and added other log levels updates to this PR because all of them are related to the same package - pkg, and I think it's better to keep everything in a single PR.

aLekSer:
@markmandel I think we could merge this PR before the release. It would significantly reduce logging in SDK server with Info level used by default.

PR #1371, opened 2020-02-27
Separate e2e tests in build Makefile

drichardson:
As part of investigating #1276, it was convenient to have separate make rules for the e2e integration test and e2e controller failure test.

Also a few minor updates to the build README.


build/Makefile
```diff
@@ -238,18 +238,20 @@ test-go:
...
+test-e2e: test-e2e-integration
 	echo "starting e2e controller failure test"
 	$(GO_TEST) -parallel=1 $(agones_package)/test/e2e/controller $(GO_E2E_TEST_ARGS) \
     		--gameserver-image=$(GS_TEST_IMAGE) \
     		--pullsecret=$(IMAGE_PULL_SECRET)
 	echo "Finishing e2e controller failure test!"
 
+test-e2e-integration: $(ensure-build-image)
+	echo "Starting e2e integration test!"
+	$(GO_TEST) $(ARGS) $(agones_package)/test/e2e $(GO_E2E_TEST_ARGS) \
```
aLekSer: Could we leave both:
test-e2e-integration and test-controller-failure make targets.
I mean we can then write:
```
test-e2e: test-e2e-integration
  $(MAKE) test-e2e-controller-failure
```

drichardson: That would evaluate ensure-build-image twice, which appears to be okay but I haven't tested it. Is ensure-build-image idempotent?

aLekSer: I assume it is not. Thanks for this notice.

aLekSer: So we can add this into the documents:
`make test-e2e ARGS='-run NOTHING'`

markmandel: I agree with @aLekSer - we could do:

```Makefile
test-e2e: 
  $(MAKE) test-e2e-integration
  $(MAKE) test-e2e-failure # this need to be written
```
And everything should work as we would like -- and if you only want to run one or the other, you can - but they can't be run in parrallel.

Since $(ensure-build-image) is idempotent (it gets called a lot), there's no real downside I can see.

WDYT?

drichardson: I just haven't want to re-test this because I spent $20 messing around with this test last week on a cluster and I don't want to do that again. I'll make the changes and just let the cloud build system try it out.

drichardson:
@aLekSer I assume this was a mistake, but you credited yourself as the author of my work.

PR #1373, opened 2020-02-28
Fix for critical GKE/GCP Terraform Bugs

### LINKED ISSUE #1372: Terraform: clean up duplicate tf config files.
There are multiple `helm.tf` files in the repo with the similar purpose:

1. Create a development cluster with current .tf file, all local disk changes applies.
1. Intended to be used as a submodule in that case terraform will clone the whole repo and master version would be used. 

**What happened**:
Redundant files. Developers and users could work on a different clusters.

I think we should have one file of each type with proper variables to specify the behaviour.

**Environment**:
- Agones version: 
- Install method (yaml/helm):
- Others:

markmandel:
This fixes several critical issues, and also cleans up other aspect along the way. Primarily that the firewall rule would never allow traffic to the cluster, since they were erroneously configured and on a
completely different network to the cluster.

Bug Fixes:
1. Remove the creation of a separate network for each cluster
1. Make sure the firewall rules don't filter on tags, but target on network tags.
1. Be able to pass in a `network` argument (defaulting to "default"), such that if you want to run a cluster on a different network it is easy to configure - this means that the firewall rule is always on the same network as the cluster.
1. Controller deployment was set to "Always" image pull policy on both the controller and the sidecar. This should never be configured this way for production systems, and is therefore an unsafe default.
1. Switched provider away from using `google-beta` since it is not the stable provider.

Cleanup:
1. Remove duplicate helm.tf and cluster.tf in root ./install/terraform path, as they were diverging, and served no value. Both ./gke and ./gke-local installation scripts have also been updated, as well as the documentation.
1. Needed to upgrade Terraform in the build shell, since all modules used a later version.
1. Cleaned up deprecation comments
1. Removed the broken symlinks in ./build

I would like to also make it easier to have only one GCP firewall rule for n number of clusters, but we can tackle that in a later PR.

Closes #1372


install/terraform/modules/gke/cluster.tf
```diff
@@ -23,24 +23,31 @@ data "google_client_config" "default" {}
 # Run `terraform taint null_resource.test-setting-variables` before second execution
 resource "null_resource" "test-setting-variables" {
   provisioner "local-exec" {
-    command = "${format("echo Current variables set as following - name: %s, project: %s, machineType: %s, initialNodeCount: %s, zone: %s",
+    command = <<EOT
```
markmandel: Main reason for this cleanup was because the previous formatting totally broke the Terraform plugin for IntelliJ. 

Also, I think it reads nicer?

site/content/en/docs/Installation/Terraform/gke.md
```diff
@@ -87,6 +87,17 @@ Configurable parameters:
 - agones_version - the version of agones to install (default is the latest version from the [Helm repository](https://agones.dev/chart/stable))
 - machine_type - machine type for hosting game servers (default is "n1-standard-4")
 - node_count - count of game server nodes for the default node pool (default is "4")
+- network - the name of the network you want your cluster and firewall rules to reside (default is "default")
+
+{{% alert title="Warning" color="warning"%}}
+On the lines that read `source = "git::https://github.com/googleforgames/agones.git//install/terraform/modules/gke/?ref=master"`
```
roberthbailey: I think @aLekSer had left this pointed at the master branch (rather than a release branch) on purpose.

site/content/en/docs/Installation/Terraform/gke.md
```diff
@@ -87,6 +87,17 @@ Configurable parameters:
...
+make sure to change `?ref=master` to match your targeted Agones release, as Terraform modules can change between
+releases.
+
+For example, if you are targeting release {{< release-branch >}}, then you will want to have 
+`source = "git::https://github.com/googleforgames/agones.git//install/terraform/modules/gke/?ref=release-{{< release-branch >}}"`
```
roberthbailey: If this is the right approach, can we can do this automatically instead of asking users to do it?

examples/terraform-submodules/gke-local/module.tf
```diff
@@ -14,7 +14,7 @@
 
 
 // Run:
-//  terraform apply [-var agones_version="1.3.0"]
+//  terraform apply -var project="<YOUR_GCP_ProjectID>" [-var agones_version="1.1.0"]
```
roberthbailey: Why change the agones version to something older? If anything, what about using 1.4.0?

site/content/en/docs/Installation/Terraform/gke.md
```diff
@@ -87,6 +87,17 @@ Configurable parameters:
 - agones_version - the version of agones to install (default is the latest version from the [Helm repository](https://agones.dev/chart/stable))
 - machine_type - machine type for hosting game servers (default is "n1-standard-4")
 - node_count - count of game server nodes for the default node pool (default is "4")
+- network - the name of the network you want your cluster and firewall rules to reside (default is "default")
```
roberthbailey: This isn't customizable in the gke-local or gke-module. They both hard code the network to default.

site/content/en/docs/Installation/Terraform/gke.md
aLekSer: Working on that, will provide more details soon.

examples/terraform-submodules/gke-local/module.tf
markmandel: I copy/pasted the previous command, but you are right, should update it to latest. Will do 👍

site/content/en/docs/Installation/Terraform/gke.md
markmandel: Oh that's a good point! I was thinking of the terraform/gke module. Will remove.

site/content/en/docs/Installation/Terraform/gke.md
markmandel: This seems like a brittle approach to me? How do we ensure that we don't accidentally break production systems with development changes?

site/content/en/docs/Installation/Terraform/gke.md
markmandel: So stack overflow says this can't be dynamic by design (since terraform has to know the value is the same at `init` and `apply` time), but I didn't actually try it.

I'll try it and find out.

site/content/en/docs/Installation/Terraform/gke.md
aLekSer: Well we can use both options. The master branch was used so that anyone can play and test `terraform apply` with those modules before the release.

markmandel:
> make gcloud-terraform-cluster
Ah good catch. I will fix this.

I figure long term, this make target will go away, as it will get merged into our `make gcloud-test-cluster` command, and there will be no difference between the two.

site/content/en/docs/Installation/Terraform/gke.md
markmandel: Seems like we keep optimising for our development, rather than our users -- we should really be optimising for our users.

We can build extra tools for ourselves to handle our development scenarios, but we shouldn't be foisting that onto our users, in my humble opinion.

examples/terraform-submodules/gke/module.tf
```diff
@@ -47,17 +47,44 @@ variable "node_count" {
...
-  source = "git::https://github.com/googleforgames/agones.git//install/terraform/?ref=master"
+module "gke_cluster" {
+  // ***************************************************************************************************
+  // Update ?ref= to the agones release you are installing. For example, ?ref=release-1.3.0 corresponds
+  // to Agones version 1.3.0
+  // ***************************************************************************************************
+  source = "git::https://github.com/googleforgames/agones.git//install/terraform/modules/gke/?ref=master"
 
   cluster = {
     "zone"             = "us-west1-c"
```
aLekSer: We probably should make a var out of zone also.

examples/terraform-submodules/gke-local/module.tf
```diff
@@ -24,30 +24,43 @@ provider "google-beta" {
...
+    "project"          = var.project
+    "network"          = "default"
   }
 }
 
 module "helm_agones" {
-
   source = "../../../install/terraform/modules/helm"
 
   agones_version         = var.agones_version
```
aLekSer: Next line:
```
values_file            = ""
```
Should be changed to 
```
values_file            = var.values_file
``` 
and please create a new variable values_file
It should contain 
`-var values_file="../helm/agones/values.yaml"`
when we install from current branch, not release master.

install/terraform/modules/helm/helm.tf
```diff
@@ -85,8 +85,8 @@ locals {
 resource "helm_release" "agones" {
   name         = "agones"
   force_update = "true"
-  repository   = "${data.helm_repository.agones.metadata.0.name}"
-  chart        = "${var.chart}"
+  repository   = data.helm_repository.agones.metadata.0.name
+  chart        = var.chart
   timeout      = 420
 
   # Use terraform of the latest >=0.12 version
```
aLekSer: "${length(var.values_file) == 0 ? "" : file("${var.values_file}")}",
This is where `values_file` would be used

install/terraform/modules/gke/cluster.tf
```diff
@@ -23,24 +23,31 @@ data "google_client_config" "default" {}
...
+    EOT
   }
 }
 resource "google_container_cluster" "primary" {
   name     = var.cluster["name"]
   location = var.cluster["zone"]
   project  = var.cluster["project"]
-  provider = google-beta
+  network  = var.cluster["network"]
+  provider = google
```
chrisst: There's no need to specify this. The provider is effectively determined by string parsing the resource name so this resource will default to `google` if not explicitly overridden. 

Also FWIW we don't consider the beta provider as an unstable provider. Instead we just provide it as a way for people to call the beta GCP APIs if they are interested in using features/fields that are only available in the `beta version of the product. There are a few more [details about the provider versions here](https://www.terraform.io/docs/providers/google/guides/provider_versions.html). That being said if you aren't using beta features then there is no need to use the beta provider.

examples/terraform-submodules/gke/module.tf
```diff
@@ -47,17 +47,44 @@ variable "node_count" {
...
     "name"             = var.name
     "machineType"      = var.machine_type
     "initialNodeCount" = var.node_count
     "project"          = var.project
+    "network"          = "default"
```
chrisst: Were you planning on changing this to a var and setting the default in the variable instead?

markmandel: Good call. Done.

examples/terraform-submodules/gke/module.tf
markmandel: :+1: done.

install/terraform/modules/gke/cluster.tf
markmandel: Thanks for the cleanup and explanation :+1: removing the reference.

My concern with the google beta api surface isn't stable, was that if we happen to inadvertently use a beta feature. And since we aren't, there doesn't seem to be a good reason to take the risk.

site/content/en/docs/Installation/Terraform/gke.md
markmandel: From @chrisst 's comments, actually added them to the module definitions - so leaving this in, and adding zone.

examples/terraform-submodules/gke-local/module.tf
markmandel: I'm not sure I understand why we want to do this?

Do we want to be sending users to look into our helm files? Seems like something they shouldn't need or want to do?

markmandel:
@aLekSer looks like you are handling all the fixes for terraform.mk n the developer tooling?

I figure it's fine to leave that broken for this PR and for stable release, since that's developer tooling, and it doesn't matter if we fix it post 1.4.0 - then post stable release we can look at fixing that after stable release. How do you feel about that?

examples/terraform-submodules/gke-local/module.tf
aLekSer: This is needed for developer tooling, we can fix this after 1.4, but what this is optional:
1. there could be values from the helm chart;
2. For development we should have an option to change something to values file and it would be reflected in the deployment

aLekSer:
LGTM

PR #1375, opened 2020-02-28
Terraform make targets: Switch from the plane structure to module terraform config

aLekSer:
make targets for development

The main difference between `make gcloud-test-cluster` and `make gcloud-terraform-cluster` is that one use helm repo later one use local chart.

 There is an option to limit the number of resources created by using `terraform apply -target=resource` 
https://www.terraform.io/docs/commands/apply.html#target-resource

In the case when we want similar functionality to `make gcloud-test-cluster` and `make install` we should use `-target=module.gke_cluster` and `-target=module.helm_agones`

Note that `values_file` and `chart` variables are there to have one `Helm.tf` module performing two functions:

1. we can install local version, local values_file is used.
1. we can install release which is achieved by
```
tag_name = "${var.agones_version != "" ? "agones.image.tag" : "skip"}"
```
For #657 and #1373

### DIFF
--- a/build/includes/terraform.mk
+++ b/build/includes/terraform.mk
@@ -70,4 +70,4 @@ endif
 
 gcloud-terraform-destroy-cluster:
 	$(DOCKER_RUN) bash -c 'cd $(mount_path)/build/terraform/gke && \
-	 terraform destroy -auto-approve'
+	 terraform destroy -target module.helm_agones.helm_release.agones -auto-approve && sleep 60 && terraform destroy -auto-approve'
--- a/build/includes/terraform.mk
+++ b/build/includes/terraform.mk
@@ -16,11 +16,11 @@
 terraform-init:
 terraform-init: $(ensure-build-image)
 	docker run --rm -it $(common_mounts) $(DOCKER_RUN_ARGS) $(build_tag) bash -c '\
-	cd $(mount_path)/examples/terraform-submodules/gke-local && terraform init && gcloud auth application-default login'
+	cd $(mount_path)/build/terraform/gke && terraform init && gcloud auth application-default login'
 
 terraform-clean:
-	rm -r ../examples/terraform-submodules/gke-local/.terraform
-	rm ../examples/terraform-submodules/gke-local/.tfstate*
+	rm -r ../build/terraform/gke/.terraform || true
+	rm ../build/terraform/gke/terraform.tfstate* || true
 
 # Creates a cluster and install release version of Agones controller
 # Version could be specified by AGONES_VERSION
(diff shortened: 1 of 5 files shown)


aLekSer:
`make gcloud-terraform-cluster` is working. Next steps before Ready toReview:

1. Add back parameters of the target.
1. Fix `gcloud-terraform-install`

Fixed next error which was due to the wrong `ProjectID`:
```                                                                                                                                                                                                    
Error: googleapi: Error 500: Internal error encountered., backendError                                                                                                                                   
                                                                                                                                                                                                         
  on ../../../install/terraform/modules/gke/cluster.tf line 35, in resource "google_container_cluster" "primary":   
```

markmandel:
Curious - why can't be re-apply the full terraform config each and every time?

I.e. have a single `make install`  (or maybe we come up with a better name? `make apply` ?)

If the cluster already exists, terraform is going to ignore it, and then the helm install will get updated as expected.  If the cluster doesn't exist, it will create it anyway. We actually can accumulate several commands into one? WDYT?

aLekSer:
>Curious - why can't be re-apply the full terraform config each and every time?
I.e. have a single make install (or maybe we come up with a better name? make apply ?)

Currently we have only one step to create cluster and install:
`make gcloud-terraform-install`

>If the cluster already exists, terraform is going to ignore it, and then the helm install will get updated as expected. If the cluster doesn't exist, it will create it anyway. We actually can accumulate several commands into one? WDYT?

After it you should do `make gcloud-terraform-destroy` if something is changed in helm only, say version. Because helm provider is not that stable. It timeouts if applied second time with different version.

aLekSer:
Reworded the description.
Details is covered here also:
https://github.com/googleforgames/agones/tree/master/build#make-gcloud-terraform-cluster

build/includes/terraform.mk
```diff
@@ -16,54 +16,59 @@
...
 		-var image_pull_secret="$(IMAGE_PULL_SECRET)" \
 		-var ping_service_type="$(PING_SERVICE_TYPE)" \
 		-var crd_cleanup="$(CRD_CLEANUP)" \
-		-var "cluster={name=\"$(GCP_CLUSTER_NAME)\", machineType=\"$(GCP_CLUSTER_NODEPOOL_MACHINETYPE)\", \
-		 zone=\"$(GCP_CLUSTER_ZONE)\", project=\"$(GCP_PROJECT)\", \
-		 initialNodeCount=\"$(GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT)\"}"'
-	$(MAKE) gcloud-auth-cluster
+		-var chart="../../../install/helm/agones/" \
+		-var name=$(GCP_CLUSTER_NAME) -var machine_type="$(GCP_CLUSTER_NODEPOOL_MACHINETYPE)" \
+		-var values_file="../../../install/helm/agones/values.yaml" \
```
markmandel: Curious - why are we passing in the value file here? It's defaulted as part of the helm config.

build/includes/terraform.mk
```diff
@@ -16,54 +16,59 @@
...
+		 terraform apply -auto-approve  -var agones_version="$(AGONES_VERSION)" \
+		-var name=$(GCP_CLUSTER_NAME) -var machine_type="$(GCP_CLUSTER_NODEPOOL_MACHINETYPE)" \
+		-var values_file="" \
+		-var zone="$(GCP_CLUSTER_ZONE)" -var project="$(GCP_PROJECT)" \
+		-var node_count=$(GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT)'
+	GCP_CLUSTER_NAME=$(GCP_CLUSTER_NAME) $(MAKE) gcloud-auth-cluster
 
 # Creates a cluster and install current version of Agones controller
 # Set all necessary variables as `make install` does
+# Unifies previous `make gcloud-test-cluster` and `make install` targets
```
markmandel: Should we delete at least `make gcloud-test-cluster` and/or `make install` ? Or rename all the terraform scripts into those targets?

Moving all the developer tooling over to terraform has some really nice implications for e2e testing as well - since we could update Our kubernetes version with a change to a terraform script, and CI would apply the change, which is awesome.

(oooh, that could also be bad, don't want competing CI builds adding and removing versions. Maybe we have to name clusters based on version, or something).

Doesn't have to be in this PR, but figured I would ask.

build/includes/terraform.mk
aLekSer: I also thought to remove, just to make it more obvious and to avoid extra digging to the configs.

build/includes/terraform.mk
aLekSer: That question need more investigation, by the way Terraform providers looks less stable to me, than what we have now for development tooling

markmandel: Sounds like we will have two ways of developing - one with the terraform script, and then also this usual way, and we can test the terraform way until we feel comfortable with it being switched over?

Works for me.

aLekSer: Yes, and also currently there is no make target for `make uninstall && make install`, which was pretty fast and eliminate creating cluster every time to aplly some fix

build/includes/terraform.mk
aLekSer: Agree with you, it is better to remove. Updated PR

aLekSer:
Tested this with switching some `values.yaml` parameter for example MIN_PORT for Gameservers. And I was able to apply this on the running cluster with Agones. Just `helm.tf` module was used in this case, and controller started with appropriate controller ENV parameter.

examples/terraform-submodules/gke-local/module.tf
```diff
@@ -41,11 +37,30 @@ variable "machine_type" {
   default = "n1-standard-4"
 }
 
+variable "name" {
+  default = "agones-terraform-example"
+}
+
+variable "values_file" {
+  default = "../../../install/helm/agones/values.yaml"
```
markmandel: I'm still trying to understand why we are doing this?

Helm already defaults to it's default values, so why reset it all over again?

Or is this meant to be an example?

aLekSer: This gke-local is intended to be used by developers, who have the full repo installed.
Imagine that developer wants to add some new one, or change its default, for example starting line:
```
agones:
  featureGates: "Example"
```
That way by changing local `values.yaml` he is able to use terraform with local file.
By the way it seems that I forgot about the option to `--set ` this parameters in Terraform for instance `--set gameservers.minPort=1000`

aLekSer: By the way all charts which were published contains `values.yaml` inside:
`https://agones.dev/chart/stable`
For example this one: https://storage.googleapis.com/agones-chart/agones-1.4.0.tgz

markmandel: If there is a developer only terraform config - I think this should be put in put in ./build, not in ./examples.

I'm sure users would also find value in having local only versions as well, so it's likely they may use this as reference and reuse.

aLekSer: Thanks for an update, that makes a lot of sense. I will move files according to their purpose.

build/terraform/gke/module.tf
```diff
@@ -41,11 +37,30 @@ variable "machine_type" {
   default = "n1-standard-4"
 }
 
+variable "name" {
+  default = "agones-terraform-example"
```
markmandel: Since this is now in our build scripts, should this be something like `agones-tf-cluster` ? So it's similar to the other cluster name, which is `test-cluster`. This is no longer an example, seems like a strange name.

What do you think? That was the best name I could come up with.

aLekSer:
Resolved cluster name

build/includes/terraform.mk
```diff
@@ -16,54 +16,58 @@
...
 gcloud-terraform-install: PING_SERVICE_TYPE := "LoadBalancer"
 gcloud-terraform-install: CRD_CLEANUP := true
+gcloud-terraform-install: GCP_CLUSTER_NAME ?= test-cluster
 gcloud-terraform-install:
 ifndef GCP_PROJECT
 	$(eval GCP_PROJECT=$(shell sh -c "gcloud config get-value project 2> /dev/null"))
 endif
 	$(DOCKER_RUN) bash -c ' \
-		cd $(mount_path)/install/terraform && terraform apply -auto-approve -var agones_version="$(VERSION)" -var image_registry="$(REGISTRY)" \
+	cd $(mount_path)/build/terraform/gke && terraform apply -auto-approve -var agones_version="$(VERSION)" -var image_registry="$(REGISTRY)" \
```
markmandel: Should this target install the latest compiled version, like [Make install](https://github.com/googleforgames/agones/blob/master/build/Makefile#L294) ?

aLekSer: It got set using 
`agones_version="$(VERSION)"` above
```
GCP_CLUSTER_NAME=test-cl2 make gcloud-terraform-install
docker run --rm -v /Users/alekser/go/src/github.com/agones/agones/build//.config/gcloud:/root/.config/gcloud -v ~/.kube/:/root/.kube -v ~/.helm:/root/.helm -v /Users/alekser/go/src
/github.com/agones/agones:/go/src/agones.dev/agones -v /Users/alekser/go/src/github.com/agones/agones/build//.gomod:/go/pkg/mod -v /Users/alekser/go/src/github.com/agones/agones/bu
ild//.gocache:/root/.cache/go-build -e "KUBECONFIG=/root/.kube/config" -e "GO111MODULE=on" -w /go/src/agones.dev/agones  agones-build:8f473e7ff2 bash -c ' \                                             
        cd /go/src/agones.dev/agones/build/terraform/gke && terraform apply -auto-approve -var agones_version="1.5.0-39911ef" -var image_registry="gcr.io/agones-images" \                               
                -var pull_policy=""Always"" \
                -var always_pull_sidecar="true" 
```

aLekSer: After that we are doing the same as in `make install`
https://github.com/googleforgames/agones/blob/5413b28aa7336603550c0de68d76461448f550f6/install/terraform/modules/helm/helm.tf#L91-L105
Setting helm parameter "agones.image.tag" to be an up to date `$(VERSION)`

aLekSer: Tested this changes and the resulted controller pod uses correct image:
```
Controlled By:        ReplicaSet/agones-controller-55d9bdd8f7                                                                                                                                     [50/706]
Containers:
  agones-controller:
    Container ID:   docker://0a986a93103382805c304f837c68449719d47ac4390058effd1052e47719d466
    Image:          gcr.io/agones-images/agones-controller:1.5.0-39911ef
```
Registry could be changed by a `REGISTRY` make variable.

aLekSer:
Added 
```
GCP_TF_CLUSTER_NAME ?= agones-tf-cluster
```
Will add Feature Gates variable into helm in the upcoming Pull Request.

build/includes/terraform.mk
markmandel: Awesome - thanks for the explanation!

PR #1384, opened 2020-03-04
Switch godoc.org links to pkg.go.dev

markmandel:
pkg.go.dev is nicer - you can choose specific versions of Go APIs to browse, and the latest tag is the default - not master (you can't even see master!)

PR #1394, opened 2020-03-05
Switch godoc.org links to pkg.go.dev

markmandel:
pkg.go.dev is nicer - you can choose specific versions of Go APIs to browse, and the latest tag is the default - not master (you can't even see master!)


markmandel:
@aLekSer this should include your feedback from #1384

PR #1396, opened 2020-03-06
Allow ports to be added to any container in a GS pod

benclive:
We have a use case for mapping ports to sidecars for auth purposes. Our pod contains a sidecar container for auth as well as a main game server container.

We require clients to be able to access the auth sidecar as well as the main game container, but Agones only maps ports to the main game container right now. This PR adds support for specifying a "ContainerName" in the GameServerPort object. It will default to the main game container if omitted so should be backwards compatible with existing configs.

**Example**
Using a modified simple-udp `fleet.yaml` gives a gameserver with ports allocated to both containers. (I ran `nc` inside alpine to avoid docker networking issues with KinD)
```console
$ kubectl get gs
NAME                     STATE   ADDRESS      PORT   NODE                   AGE
simple-udp-trwmb-rflmk   Ready   172.17.0.5   7460   agones-control-plane   22m
$ docker run -i alpine nc -u 172.17.0.5 7460
Hello
ACK: Hello
^Cpunt!
$ docker run -i alpine nc -u 172.17.0.5 7602
Hello, 5000!
ACK: Hello, 5000!
^Cpunt!
```

Game Server Description:
```
  Ports:
    Container Name:  simple-udp
    Container Port:  7654
    Host Port:       7460
    Name:            default
    Port Policy:     Dynamic
    Protocol:        UDP
    Container Name:  simple-udp-5000
    Container Port:  5000
    Host Port:       7602
    Name:            other
    Port Policy:     Dynamic
    Protocol:        UDP
```

fleet.yaml
```yaml
apiVersion: "agones.dev/v1"
kind: Fleet
metadata:
  name: simple-udp
spec:
  replicas: 1
  template:
    spec:
      container: simple-udp
      ports:
      - name: default
        containerPort: 7654
      - name: other
        containerPort: 5000
        containerName: simple-udp-5000
      template:
        spec:
          containers:
          - name: simple-udp
            image: gcr.io/agones-images/udp-server:0.18
            resources:
              requests:
                memory: "64Mi"
                cpu: "20m"
              limits:
                memory: "64Mi"
                cpu: "20m"
          - name: simple-udp-5000
            image: gcr.io/agones-images/udp-server:0.18
            args:
              - -port
              - "5000"
            resources:
              requests:
                memory: "64Mi"
                cpu: "20m"
              limits:
                memory: "64Mi"
                cpu: "20m"
```


aLekSer:
Thanks for this PR, it seems that this feature would be really useful.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -532,24 +554,26 @@ func (gs *GameServer) Pod(sidecars ...corev1.Container) (*corev1.Pod, error) {
...
 
 	for _, p := range gs.Spec.Ports {
 		cp := corev1.ContainerPort{
 			ContainerPort: p.ContainerPort,
 			HostPort:      p.HostPort,
 			Protocol:      p.Protocol,
 		}
-		gsContainer.Ports = append(gsContainer.Ports, cp)
+		portApplied := false
+		for j, container := range pod.Spec.Containers {
```
domgreen: nit - `j` is this just index?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -417,6 +423,22 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 					Message: ErrHostPortDynamic,
 				})
 			}
+
+			if p.ContainerName != ""  && gss.Container != "" {
```
domgreen: is there tests around one or both of these being an empty string?

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -375,6 +375,12 @@ func TestGameServerValidate(t *testing.T) {
 				Name:       "main",
 				HostPort:   5001,
 				PortPolicy: Dynamic,
+			}, {
+				Name:       "sidecar",
+				HostPort:   5002,
+				PortPolicy: Static,
+				ContainerPort: 5002,
+				ContainerName: "anothertest",
```
domgreen: is this the only place it needs to be tested?

aLekSer:
`make test-gen-api-docs` have failed.
Please run `make gen-api-docs` and add resulted html file here into the PR.

markmandel:
Hi! Thanks so much for doing this work! I had a TODO from a previous Slack conversation to write a ticket for this for ages, and never got around to it! This is going to be really useful.

A few high level things, and then I'll drop some comments on more specific stuff:
Since this is a CRD and functionality change, we'll need it to go through [Feature Stages](https://agones.dev/site/docs/guides/feature-stages/#new-crd-attributes) as an Alpha feature, and be implemented as such.

Which means it will also need a feature gate - you can add one [here](https://github.com/googleforgames/agones/blob/master/pkg/util/runtime/features.go).

We should also add an [e2e test](https://github.com/googleforgames/agones/tree/master/test/e2e) for this - I wonder if we can run 2 simpe-udp containers side by side and open them both up.

Right now the [e2e test installation](https://github.com/googleforgames/agones/blob/master/build/Makefile#L283-L300) doesn't enable any feature flags, but I'll enable that today so it's ready for you as well.

Please let us know if you have any questions on the above.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -192,7 +192,9 @@ type GameServerPort struct {
 	// When `Static` portPolicy is specified, `HostPort` is required, to specify the port that game clients will
 	// connect to
 	PortPolicy PortPolicy `json:"portPolicy,omitempty"`
-	// ContainerPort is the port that is being opened on the game server process
+	// ContainerName is the container on which to open the port. Defaults to the game server container.
+	ContainerName string `json:"containerName,omitempty"`
```
markmandel: Should this be `Container string` to match what we have at the top level for specifying the "game server container"? Seems like it should be more consistent.

Also, when you do the Alpha feature flagging - this should be made into a pointer, so it's easily omitted.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -417,6 +423,22 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 					Message: ErrHostPortDynamic,
 				})
 			}
+
+			if p.ContainerName != "" && gss.Container != "" {
+				containerFound := false
+				for _, container := range gss.Template.Spec.Containers {
```
markmandel: Would it make more sense to update `FindGameServerContainer` to something that can `FindContainer` which takes a container name passed in?

Might make this code a little nicer, and likely easier to test.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -532,24 +554,26 @@ func (gs *GameServer) Pod(sidecars ...corev1.Container) (*corev1.Pod, error) {
...
 	for _, p := range gs.Spec.Ports {
 		cp := corev1.ContainerPort{
 			ContainerPort: p.ContainerPort,
 			HostPort:      p.HostPort,
 			Protocol:      p.Protocol,
 		}
-		gsContainer.Ports = append(gsContainer.Ports, cp)
+		portApplied := false
+		for j, container := range pod.Spec.Containers {
+			if container.Name == p.ContainerName {
```
markmandel: Probably also makes sense to update and rename `ApplyToPodGameServerContainer` to use with this logic, since it's almost the same, except you would want to pass in a name, rather than use the gs.Spec.Container by default.

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -543,18 +549,29 @@ func TestGameServerPod(t *testing.T) {
 	assert.True(t, metav1.IsControlledBy(pod, fixture))
 	assert.Equal(t, fixture.Spec.Ports[0].HostPort, pod.Spec.Containers[0].Ports[0].HostPort)
 	assert.Equal(t, fixture.Spec.Ports[0].ContainerPort, pod.Spec.Containers[0].Ports[0].ContainerPort)
+	assert.Equal(t, fixture.Spec.Ports[0].ContainerName, pod.Spec.Containers[0].Name)
 	assert.Equal(t, corev1.Protocol("UDP"), pod.Spec.Containers[0].Ports[0].Protocol)
 	assert.True(t, metav1.IsControlledBy(pod, fixture))
 
 	sidecar := corev1.Container{Name: "sidecar", Image: "container/sidecar"}
 	fixture.Spec.Template.Spec.ServiceAccountName = "other-agones-sdk"
+	fixture.Spec.Ports = append(fixture.Spec.Ports, GameServerPort{
```
markmandel: Do we want to split this into two tests? Once with the default port setup, and one with this new functionality - just to get coverage on both scenarios?

pkg/apis/agones/v1/gameserver_test.go
markmandel: Other than the comment below on splitting things up, I don't see any other unit tests that need updating at this point.

But like i mentioned in the main comment - would definitely want a e2e test.

benclive:
I've iterated on all your suggestions: thanks for the feedback! There are a couple of instances where errors are newly returned but I opted to ignore them due to the game server name definitely being set at that point.
Do you want me to propagate them through the call stack to handle them correctly, or is there a logger to track those without re-writing a lot of method signatures?

I've also run into a problem with the docs after `make site-server`. When I add the hugo feature block I get some odd formatting issues. Any ideas there?
```
{{% feature publishVersion="1.5.0" %}}
    # the name of the container to open the port on. Defaults to the game server container if omitted or empty.
    container: simple-udp
{{% /feature %}}
```
results in large white space
image

aLekSer:
@benclive thanks for noticing this. As a workaround you can put the whole "```yaml" block into the shortcode:
```
{{% feature publishVersion="1.5.0" %}}
{{% / feature %}}
```

benclive:
@markmandel The E2E tests are now failing on my change; I'm waiting for https://github.com/googleforgames/agones/pull/1397 to be merged to make any more progress as I need to enable the feature flag in the E2E tests.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -530,15 +531,23 @@ func (gs *GameServer) FindGameServerContainer() (int, corev1.Container, error) {
 
 // ApplyToPodContainer applies func(v1.Container) to the specified container in the pod.
```
aLekSer: Nit
```suggestion
// ApplyToPodContainer applies func(corev1.Container) to the specified container in the pod.
```

pkg/apis/agones/v1/gameserver.go
```diff
@@ -328,6 +333,10 @@ func (gss *GameServerSpec) applyPortDefaults() {
 		if p.Protocol == "" {
 			gss.Ports[i].Protocol = "UDP"
 		}
+
+		if p.Container == nil || *p.Container == "" {
```
markmandel: This should probably be wrapped in FeatureContainerPortAllocation, such that this value is not set when the flag is enabled. Since if people aren't using the alpha feature, it probably shouldn't show up as a value in the CRD.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -533,21 +560,21 @@ func (gs *GameServer) Pod(sidecars ...corev1.Container) (*corev1.Pod, error) {
...
-	}
-
 	for _, p := range gs.Spec.Ports {
 		cp := corev1.ContainerPort{
 			ContainerPort: p.ContainerPort,
 			HostPort:      p.HostPort,
 			Protocol:      p.Protocol,
 		}
-		gsContainer.Ports = append(gsContainer.Ports, cp)
+		err := gs.ApplyToPodContainer(pod, *p.Container, func(c corev1.Container) corev1.Container {
```
markmandel: This should also be wrapped in a FeatureContainerPortAllocation - such that the previous logic stays in place for when it's disabled, and the new logic is in place for the new feature.

Then we'll need to split out a test for the enabled and disabled unit tests. ([Here's an example](https://github.com/googleforgames/agones/pull/1399/files#diff-a2a0019b8308830fa9857437be51f29dR910-R914))

pkg/apis/agones/v1/gameserver.go
```diff
@@ -666,6 +693,12 @@ func (gs *GameServer) CountPorts(f func(policy PortPolicy) bool) int {
 func (gs *GameServer) Patch(delta *GameServer) ([]byte, error) {
 	var result []byte
 
+	if !runtime.FeatureEnabled(runtime.FeatureContainerPortAllocation) {
```
markmandel: I think this might be the wrong spot for this - I think this should be in the sections I've highlighted above.

aLekSer:
Sorry, my fault that CLA has broken. Thinking that signing CLA for both emails - primary and secondary.

aLekSer:
I signed it

pkg/apis/agones/v1/gameserver.go
benclive: I left the old test in place, and enabled the flag for the new test. They should both pass now 🤞

site/content/en/docs/Reference/gameserver.md
```diff
@@ -77,6 +80,79 @@ spec:
...
+    # name is a descriptive name for the port
+  - name: default
+    # portPolicy has three options:
+    # - "Dynamic" (default) the system allocates a free hostPort for the gameserver, for game clients to connect to
+    # - "Static", user defines the hostPort that the game client will connect to. Then onus is on the user to ensure that the
+    # port is available. When static is the policy specified, `hostPort` is required to be populated
+    # - "Passthrough" dynamically sets the `containerPort` to the same value as the dynamically selected hostPort.
+    #      This will mean that users will need to lookup what port has been opened through the server side SDK.
+    portPolicy: Static
+    # the name of the container to open the port on. Defaults to the game server container if omitted or empty.
```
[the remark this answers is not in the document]
benclive: I think its useful but I'm not sure how to keep it in sync once the feature moves out of alpha. Is searching for "(Alpha)" or similar on each release sufficient to keep this up to date?

Either way, I've added it for now!

pkg/apis/agones/v1/gameserver.go
benclive: I wasn't sure where to actually use the feature flag - thanks for the suggestions!

markmandel:
Looks like API Docs are still failing.

Please run `make gen-api-docs` to solve the issue :+1:

aLekSer:
This E2E test is related to this PR:
```

time="2020-03-17 11:36:46.817" level=info msg="waiting for fleet condition" fleet=simple-fleet-zrtj7
--- FAIL: TestFleetGSSpecValidation (2.22s)
    fleet_test.go:589: 
        	Error Trace:	fleet_test.go:589
        	Error:      	"[{FieldValueInvalid Could not find a container named testing container}]" should have 2 item(s), but has 1
        	Test:       	TestFleetGSSpecValidation
panic: runtime error: index out of range [1] with length 1 [recovered]
	panic: runtime error: index out of range [1] with length 1

goroutine 223 [running]:
testing.tRunner.func1(0xc000197a00)
	/usr/local/go/src/testing/testing.go:874 +0x69f
panic(0x1b29080, 0xc0000419a0)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
agones.dev/agones/test/e2e.TestFleetGSSpecValidation(0xc000197a00)
	/go/src/agones.dev/agones/test/e2e/fleet_test.go:592 +0x1a24
testing.tRunner(0xc000197a00, 0x1c7add0)
	/usr/local/go/src/testing/testing.go:909 +0x19a
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:960 +0x652
FAIL	agones.dev/agones/test/e2e	31.066s
FAIL
make[1]: *** [test-e2e-integration] Error 1
Makefile:246: recipe for target 'test-e2e-integration' failed
```

benclive:
I can't tell if the current failure is related to my changes or not: Running `test/e2e/gameserver_test.go` locally on `kind` seems to work for me but I haven't been able to run the whole suite consistently due to my machine not being powerful enough. 
I will attempt to get GKE working. In the meantime, is there any other way to debug these tests?

markmandel:
Ah, TestGameServerReserve is flaky - #1276 - I'll restart it.

site/content/en/docs/Reference/gameserver.md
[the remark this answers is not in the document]
markmandel: Yeah, I think that might be the answer. I can't think of a better way.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -28,6 +28,7 @@ The current set of `alpha` and `beta` feature gates are:
 |--------------|---------|---------|-------|-------|
 | Multicluster Allocation<sup>*</sup> | N/A | Enabled | `Alpha` | 0.11.0 |
 | Example Gate (not in use) | `Example` | Disabled | None | 0.13.0 |
+| [Port Allocations to Multiple Containers]({{< ref "/docs/Reference/gameserver.md" >}}) | `ContainerPortAllocation` | Disabled | `Alpha` | 1.5.0 |
```
markmandel: You'll hate me, but this need to be wrapped in a `feature` shortcode like you did previously, so it doesn't show up with this release. :man_facepalming:  I should have mentioned that in my comments previously.

(Might need to make a copy of the table, not sure, depends how formatting works out)

I swear this is the last thing :+1:

benclive: Good point! I had to wrap the table as you mentioned but all done now :)

PR #1397, opened 2020-03-06
Add feature gate block to Make install

markmandel:
So we can test feature gated features, adding a config option to set feature flags, and setting it to enable the currently available feature flags.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -287,6 +287,7 @@ install: PING_SERVICE_TYPE := "LoadBalancer"
 install: ALLOCATOR_SERVICE_TYPE := "LoadBalancer"
 install: CRD_CLEANUP := true
 install: LOG_LEVEL := "debug"
+install: FEATURE_GATES := "PlayerTracking=true"
 install: $(ensure-build-image) install-custom-pull-secret
 	$(DOCKER_RUN) \
 		helm upgrade --install --wait --namespace=agones-system\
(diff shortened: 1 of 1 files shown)


build/Makefile
```diff
@@ -287,6 +287,7 @@ install: PING_SERVICE_TYPE := "LoadBalancer"
 install: ALLOCATOR_SERVICE_TYPE := "LoadBalancer"
 install: CRD_CLEANUP := true
 install: LOG_LEVEL := "debug"
+install: FEATURE_GATES := "PlayerTracking=true"
```
roberthbailey: Is the idea that we turn on all feature gates for develpers? Does that make it so that the developer testing experience diverges too much from the production experience?

markmandel: For e2e tests, all features will need to be enabled so we can run all the tests. (unless we want to split things? This starts to get complicated, and I'm not sure what the right answer here is)

So I figured it made sense to turn then on for developers as well (especially since they would also likely need to be enabling them). FEATURE_GATES is overwrite-able, so one could turn off features as needed.

Counter argument - We should probably be the ones testing Alpha state the most, as I expect others won't necessarily be.

But definitely open to suggestions - if there is a better way?

roberthbailey: I think the other option is to run multiple test suites. I'm not sure if it's worth the trouble though.

If we want to try to run tests w/o features to better simulate what our default configuration will be for the next release, one way to do it without too much pain would be to change the post-submit tests (that run against github master in cloud build) to run without feature gates and skip the e2e tests for alpha features. That would a relatively small change to our testing configuration, but we'd need to have a better way to keep an eye on those test results. I don't think it would be a problem yet, but if we started introducing changes to say the helm install that were behind feature gates, since we share the same e2e cluster for pre-submit and post-submit tests I could see that causing breakage down the line....

TL;DR: Probably not worth it right now; but maybe file a feature request so that we can discuss what we'd want and store our ideas for how we could accomplish it.

markmandel: Issue filed: https://github.com/googleforgames/agones/issues/1411

PR #1399, opened 2020-03-07
Alpha SDK.SetPlayerCapacity & GetPlayerCapacity

markmandel:
This PR doesn't complete this functionality, but was enough to warrant review.

It does two things:
1. The refactoring to move alpha and beta SDK Server implementations into the same file as the main SDK Server, as being able to reuse the current logic made sense on review.
2. Initial Implementation of SDK GetPlayerCapacity & SetPlayerCapacity methods with unit tests. E2E tests can be implemented once #1397 is resolved.

Next steps will be to implement the local SDK functionality, and conformance tests for Go for these two functions.

Work on #1033


markmandel:
Would this be easier to review if I attempted to split it into some smaller PRs?

I was thinking about it, and I could likely split it into 2 at least, if people would prefer?

roberthbailey:
> Would this be easier to review if I attempted to split it into some smaller PRs?

Always. ;)

pkg/sdkserver/localsdk.go
```diff
@@ -425,3 +429,30 @@ func (l *LocalSDKServer) setGameServerFromFilePath(filePath string) error {
 	l.gs = convert(&gs)
 	return nil
 }
+
+/* Alpha Functionality */
```
roberthbailey: Can you also add an empty block / area for beta functionality?

Looking at this, it's going to be easy to miss graduating the blocks here from alpha -> beta -> ga when features move. Is there part of the feature documentation that needs to cover moving the code in this file?

pkg/sdkserver/sdkserver.go
```diff
@@ -587,3 +604,68 @@ func (s *SDKServer) healthy() bool {
...
+
+// GetPlayerCapacity returns the current player capacity.
+func (s *SDKServer) GetPlayerCapacity(ctx context.Context, _ *alpha.Empty) (*alpha.Count, error) {
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return nil, errors.New(string(runtime.FeaturePlayerTracking) + " not enabled")
+	}
+	s.gsUpdateMutex.RLock()
+	defer s.gsUpdateMutex.RUnlock()
+
+	count := &alpha.Count{
```
roberthbailey: nit: It seems like this could just be inlined:

`return &alpha.Count{Count: s.gsPlayerCapacity}, nil`

pkg/sdkserver/sdkserver.go
```diff
@@ -587,3 +604,68 @@ func (s *SDKServer) healthy() bool {
...
+		Count: s.gsPlayerCapacity,
+	}
+	return count, nil
+}
+
+// updatePlayerCapacity updates that status player capacity
```
roberthbailey: nit: Not quite a sentence..

proto/sdk/alpha/alpha.proto
```diff
@@ -37,15 +37,16 @@ service SDK {
...
     rpc SetPlayerCapacity (Count) returns (Empty) {
         option (google.api.http) = {
             post: "/alpha/player/capacity"
             body: "*"
         };
     }
 
-    // get the current player capacity
+    // Get the last player capacity that was set through the SDK.
+    // if the player capacity is set from outside the SDK, use SDK.GameServer() instead.
```
roberthbailey: Nit: Start the new sentence with a capital letter.

sdks/go/alpha.go
```diff
@@ -0,0 +1,50 @@
...
+// Alpha is the struct for Alpha SDK functionality
+type Alpha struct {
+	client alpha.SDKClient
+	ctx    context.Context
+}
+
+// newAlpha creates a new Alpha SDK with the passed in connection.
+func newAlpha(conn *grpc.ClientConn) *Alpha {
+	return &Alpha{
+		ctx:    context.Background(),
```
roberthbailey: Maybe this is how all of the existing code works, but it's more typical to pass context in with each rpc rather than store it in a struct and reuse it. I know from experience that it can be quite a pain to retrofit code to pass context down the stack later...

sdks/go/alpha.go
```diff
@@ -0,0 +1,50 @@
...
+		client: alpha.NewSDKClient(conn),
+	}
+}
+
+// GetPlayerCapacity gets the last player capacity that was set through the SDK.
+// if the player capacity is set from outside the SDK, use SDK.Alpha.GameServer() instead.
```
roberthbailey: Nit: Start new sentence with a capital.

sdks/go/alpha_test.go
```diff
@@ -0,0 +1,65 @@
...
+// limitations under the License.
+
+package sdk
+
+import (
+	"testing"
+
+	"agones.dev/agones/pkg/sdk/alpha"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
```
roberthbailey: Import "context" instead.

pkg/sdkserver/sdkserver_test.go
```diff
@@ -901,6 +903,83 @@ func TestSDKServerReserveTimeout(t *testing.T) {
 	wg.Wait()
 }
 
+/* Alpha Functionality */
+
+func TestSDKServerPlayerCapacity(t *testing.T) {
```
roberthbailey: There are a couple of comments that "if the player capacity is set from outside the SDK, use SDK.GameServer() instead." It would be nice to have a test that proves the comment true.

pkg/sdkserver/localsdk.go
markmandel: > Looking at this, it's going to be easy to miss graduating the blocks here from alpha -> beta -> ga when features move. Is there part of the feature documentation that needs to cover moving the code in this file?

This is a good point. Rather than having a block like this, does it make sense to make sure that "Alpha" is somehow in the comment block for the function, which means (a) we don't have to move it and (b) it should be much easier to pick up on? (or will it? :thinking:)

So rather than

```go
/* Alpha Functionality */

// PlayerConnect should be called when a player connects.
func (l *LocalSDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerId) (*alpha.Empty, error) {
	panic("implement me")
}
```

We would have

```go
// PlayerConnect should be called when a player connects.
// (Alpha)
func (l *LocalSDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerId) (*alpha.Empty, error) {
	panic("implement me")
}

// PlayerConnect should be called when a player connects.
// (Alpha)
func (l *LocalSDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerId) (*alpha.Empty, error) {
	panic("implement me")
}
```

etc? WDYT?

Might still require a change to the Feature Stage docs too though.

markmandel: Or maybe we should have some kind of Alpha + FeatureFlag combo?

```go
// PlayerConnect should be called when a player connects.
// (Alpha:PlayerTesting)
func (l *LocalSDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerId) (*alpha.Empty, error) {
	panic("implement me")
}
```

Other options:
- `[Alpha:PlayerTesting]`
- `Alpha:PlayerTesting`

Make it easy to search and replace for specific Alpha functionality. I specifically didn't put a space in, just because it's easy to get inconsistent with spacing.

sdks/go/alpha.go
markmandel: Hmnn, that's an interesting point. I'm also not sure why we're storing the background context, since it's already stored inside the Context package.

Any issue to shifting the alpha implementations to look like:

```go
// GetPlayerCapacity gets the last player capacity that was set through the SDK.
// if the player capacity is set from outside the SDK, use SDK.Alpha.GameServer() instead.
func (a *Alpha) GetPlayerCapacity() (int64, error) {
	c, err := a.client.GetPlayerCapacity(context.Background(), &alpha.Empty{})
	return c.Count, errors.Wrap(err, "could not get player capacity")
}
```

Since we're never attempting to cancel them?

I figure the others could be cleaned up in a later PR as needed?

sdks/go/alpha_test.go
markmandel: Sod. Thank you.

pkg/sdkserver/sdkserver_test.go
markmandel: Thanks for bringing this up - I wasn't sure the best approach here.

So I deliberately haven't implemented the Alpha properties on SDK.GameServer() because I'm not sure what the right approach here is (do we have a separate SDK.Alpha().GameServer()? can we integrate it into the stable version? Each has trade offs), so I wrote a comment of what I figured might make sense, with intent to change it later on once we hashed out the details in a separate PR/discussion.

I'm wondering if there should be a `TODO:` of some kind here for this PR, so it's clear this isn't implemented functionality that needs to be updated later on? Would that work?

pkg/sdkserver/localsdk.go
roberthbailey: Putting it into each function makes sense. Then we can also put the functions into the file where we want them to live permanently (assuming they graduate out of alpha) instead of ending up with everything just being tacked onto the end and/or shuffling the blocks of code around during promotion (which can lead to missed changes during code review).

For the comment annotation, it's more verbose but I think it might be nice to do something like

```
// [Stage:Alpha]
// [FeatureFlag:PlayerTesting]
```

Since it will be more apparent to new developers how to emulate it without needing to look for documentation outside of the code.

pkg/sdkserver/sdkserver.go
roberthbailey: You still left this on two lines. Was there a reason to keep the local `count` variable that I'm missing? It seems unnecessary.

proto/sdk/alpha/alpha.proto
```diff
@@ -37,15 +37,16 @@ service SDK {
         };
     }
 
-    // change the player capacity to a new value
+    // changes the player capacity to a new value
```
roberthbailey: Nit: Start the new sentence with a capital letter.

sdks/go/alpha.go
roberthbailey: Context isn't just for canceling though; it can also be used to pass things like authn/authz tokens, trace ids (to track call dependencies), etc. 

I think the suggested change is better than what you currently have, but is there a reason that (for golang clients) we don't set the function signature to include the context? e.g.

```
// GetPlayerCapacity gets the last player capacity that was set through the SDK.
// if the player capacity is set from outside the SDK, use SDK.Alpha.GameServer() instead.
func (a *Alpha) GetPlayerCapacity(ctx context.Context) (int64, error) {
	c, err := a.client.GetPlayerCapacity(ctx, &alpha.Empty{})
	return c.Count, errors.Wrap(err, "could not get player capacity")
}
```

pkg/sdkserver/sdkserver_test.go
roberthbailey: A TODO would be fine since this is alpha and there are some unanswered questions.

pkg/sdkserver/localsdk.go
markmandel: I like that a lot. Let's run with this :smile:

pkg/sdkserver/sdkserver.go
markmandel: Ah, I missed a bit. Fixed!

sdks/go/alpha.go
markmandel: Probably that it was just how it was done originally, nothing more than that, and to keep it inline with the other SDKs.

I'll make the change to move these parts to use the background Context.

Issue #1370, opened 2020-02-27
GameServer remains "STATE:Creating" if not create serviceaccount


suecideTech:
**What happened**:
I created a GameServer in a namespace that does not create a serviceaccount for agones-sdk.(mistake operation)
As a result, GameServer remains `STATE: Creating` and I did not understand the cause even when I saw the `kubectl describe`.

$ kubectl describe gameserver ...
```
Status:
  Address:
  Node Name:
  Ports:           <nil>
  Reserved Until:  <nil>
  State:           Creating
Events:
  Type    Reason          Age   From                   Message
  ----    ------          ----  ----                   -------
  Normal  PortAllocation  25s   gameserver-controller  Port allocated
```

**What you expected to happen**:
I want a message that "error looking up serviceaccount for agones-sdk" displayed in the event of `kubectl describe` and the `STATE` becomes `STATE: error`

**How to reproduce it (as minimally and precisely as possible)**:
Create a GameServer without creating an serviceaccount for agones-sdk.

**Anything else we need to know?**:
I couldn't find the cause without looking at the agones-controller log.
```
{"error":"error creating Pod for GameServer simple-udp-mrlql: pods \"simple-udp-mrlql\" is forbidden: error looking up service account test/agones-sdk: serviceaccount \"agones-sdk\" not found","gsKey":"test/simple-udp-mrlql","message":"","queue":"agones.dev.GameServerControllerCreation","severity":"error","source":"*gameservers.Controller","subqueue":"creation","time":"2020-02-27T14:25:43.406369349Z"}
```

**Environment**:
- Agones version: 1.3.0
- Kubernetes version (use `kubectl version`): v1.13.12-gke.30 (v1.13.5)
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): yaml
- Troubleshooting guide log(s): See above
- Others: N/A

I would be grateful if you could fixed this.

markmandel:
This is documented here:
https://agones.dev/site/docs/installation/install-agones/helm/#namespaces

So this is working as intended.

suecideTech:
I think it's better to fix below.

```
$ git diff
diff --git a/pkg/gameservers/controller.go b/pkg/gameservers/controller.go
index 0a185116..ee1a7cd4 100644
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
[...]

PR #1400, opened 2020-03-07
Add ErrorHandling for failed to create pods because of forbidden

suecideTech:
When failed to create GameServer Pods becouse of  "forbidden", GameServer's STATE remained "Creating".
eg. agones-sdk serviceaccount does not exist

Fixed so that put the reason of error to Event and STATE change to "Error".

#### Test results
- Add unit test case and passed.
- Passed e2e test. (Create GameServer without making agones-sdk serviceaccount)

#1370


suecideTech:
@googlebot I signed it!

Issue #1299, opened 2020-01-28
Gameservers using nodejs sdk die with GOAWAY ENHANCE_YOUR_CALM too_many_pings


msannikov:
**What happened**:
All my gameservers die after 30-90 minutes when I started using nodejs sdk and upgraded agones from 1.0.0 to 1.3.0:
Received a GOAWAY with error code ENHANCE_YOUR_CALM and debug data equal to "too_many_pings"

**Anything else we need to know?**:
This seems to be related to grpc/grpc-node#138. I've tried to upgrade agones's grpc to the latest v1.24.2, but that did not help.

gameserver container logs:
```
I 2020-01-23T18:24:14.871311349Z NODE_ENV production
...
E 2020-01-23T19:00:05.907467004Z   Received a GOAWAY with error code ENHANCE_YOUR_CALM and debug data equal to "too_many_pings" 
```

agones-gameserver-sidecar logs:
```
I 2020-01-23T18:24:15.216010496Z Starting sdk sidecar 
...
I 2020-01-23T18:24:17.885301089Z Health Ping Received 
...
I 2020-01-23T18:25:05.185913691Z Received Ready request, adding to queue 
...
I 2020-01-23T18:25:05.902915961Z Health Ping Received 
...
I 2020-01-23T19:00:10.401435848Z GameServer Health Check failed 
I 2020-01-23T19:00:15.264434202Z Sending GameServer Event to connectedStreams 
I 2020-01-23T19:00:15.401710405Z GameServer Health Check failed 
I 2020-01-23T19:00:20.401923432Z GameServer Health Check failed 
I 2020-01-23T19:00:20.402013117Z GameServer has failed health check 
...
```

**Environment**:
- Agones version: 1.3.0
- Kubernetes version (use `kubectl version`): 1.13.12-gke.17
- Cloud provider or hardware configuration: GCP
- Install method (yaml/helm): N/A
- Troubleshooting guide log(s): N/A
- Others: N/A

markmandel:
@KamiMay @steven-supersolid - have you seen this?

I'm also assuming you are using the most up to date nodejs client as well?

steven-supersolid:
We're using the REST API but may be related to grpc/grpc-node#717

There is a comment to enable further debug https://github.com/grpc/grpc-node/issues/717#issuecomment-457408937

To troubleshoot @msannikov is it possible to run against Agones 1.2.0 and see if the error still occurs - this will make comparisons between versions easier?

[...]

PR #1413, opened 2020-03-14
Add shutdown duration option to Node.js simple 

steven-supersolid:
The example code now reads the first argument if specified
If --help is passed then will exit with a help message
If a number is passed then that will be used for the shutdown duration

Converted setTimeout to use `await` to make code clearer
Updated watch to give more relevant information set during the test
Clear intervals correctly so these don't prevent the app from exiting
Found the app still won't exit cleanly so potentially a bug in `sdk.close()` (fix as separate PR)

Updated Makefile and dockerfile to pass in a parameter
Updated documentation


examples/nodejs-simple/Makefile
```diff
@@ -43,7 +43,7 @@ build:
 	cd $(root_path) && docker build -f $(project_path)/Dockerfile --tag=$(server_tag) .
 
 run:
-	docker run --network=host $(server_tag)
+	docker run --network=host $(server_tag) $(args)
```
markmandel: Nice stuff - will be useful for testing #1299 :+1: 

We should also increment the tag version in this file, so we can rebuild and push it up for people to use.

examples/nodejs-simple/README.md
```diff
@@ -2,11 +2,14 @@
 
 This is a very simple "server" that doesn't do much other than show how the SDK works in node.js.
 
-It will
-- Setup the Agones SDK
-- Call `sdk.ready()` to register that it is ready with Agones.
+It will:
```
markmandel: Do we want to document what the command is (or the Docker run command?) should be to run this for longer than 60 seconds?

examples/nodejs-simple/Makefile
steven-supersolid: Would that be server-tag and should it be 1.5 now?

examples/nodejs-simple/README.md
steven-supersolid: That is documented around line 120 after example debug output but perhaps clearer if I reorder?

examples/nodejs-simple/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@google-cloud/agones-sdk": "1.3.0"
+    "@google-cloud/agones-sdk": "1.4.0"
```
steven-supersolid: Perhaps should update as part of the release @markmandel ?

markmandel: I don't think this is necessary, unless there is a code change in either the SDK or the example.

Unless you can see a good reason to?

examples/nodejs-simple/Makefile
markmandel: `server_tag = $(REPOSITORY)/nodejs-simple-server:0.3` :arrow_right: `server_tag = $(REPOSITORY)/nodejs-simple-server:0.4`

examples/nodejs-simple/README.md
```diff
@@ -39,71 +42,98 @@ docker run --network=host gcr.io/agones-images/nodejs-simple-server:0.1
...
+You can optionally specify how long the server will stay up once the basic tests are complete.
+To do this pass arguments through, e.g. to increase the shutdown duration to 120 seconds:
+```
+$ make args="120" run
+```
+
+Please note that there is a max sleep time in Node.js of 2,147,483 seconds (~24 days). Values will be capped at this and negative values or zero will also be set to this.
+So to keep the server alive for a long time you could use the following:
+```
+$ make args="0" run
```
markmandel: I would make these (no pun intended :smile: ) the full Docker run command.

That way people don't need to pull down the full example source code -- and can also replicated within their Agones install as well.

examples/nodejs-simple/README.md
markmandel: I would also like the explicit command line argument documented.

Or is it just "120" as seconds ? Should we use a key-value pair? Like --timeout=120 ?

Would be much more explicit, and also expandable down the line. WDYT?

examples/nodejs-simple/package.json
steven-supersolid: The main reason would be to keep the code sample up to date and consistent with the release. We could also change to a non-pinned dependency

examples/nodejs-simple/README.md
steven-supersolid: Ah sure. I was just basing this on the previous instructions. I usually run with `npm start` anyway.
If there's a need for both should we have both ways of running?
@roberthbailey you wrote the original README - any thoughts on this?

examples/nodejs-simple/package.json
markmandel: But if there is no public API code change between versions, then it is up to date with the latest release :smile:

examples/nodejs-simple/README.md
markmandel: Ah, if it's been this way for a while, I'll be less opinionated on changing to Docker run... but it still would be nice I think?

examples/nodejs-simple/README.md
```diff
@@ -32,78 +36,44 @@ $ make build
...
+```
+$ make args="--help" run
+$ docker run --network=host gcr.io/agones-images/nodejs-simple-server:0.4 --help
+$ npm start -- --help
+```
+
+You can optionally specify how long the server will stay up once the basic tests are complete with the `--timeout` option.
+To do this pass arguments through, e.g. to increase the shutdown duration to 120 seconds:
+```
+$ make args="--timeout=120" run
```
markmandel: Love this! :smile:

markmandel:
Weird, I can't update this branch to master. There is no button? 
image

steven-supersolid:
Wonder if it's because I merged without the button last time? But I had the button just now, so pressed it :)

markmandel:
I'll wait for this to run through, and then it gets to be first in the queue :smile:

PR #1414, opened 2020-03-15
Flaky: TestGameServerReserve

### LINKED ISSUE #1276: Flaky: TestGameServerReserve

markmandel:
After testing, it seems that the K8s API servers can get blocked for an indeterminate period during tests - a period that can be longer than the 10 seconds in which our GameServer goes from Ready->Reserverd->Ready - which caused the previous version of the test to fail.

Updated the test to only look at the event stream to ensure the e2e test state is verified. While it would be ideal to look at the state itself, we will assume that unit tests cover that the Event stream is accurate to the GameServer state being changed.

Fixes #1276


roberthbailey:
Thinking about this a bit -- it seems like the old way didn't work because it was looking for the "edges" of state transitions (which could be missed) and the new way is relying on the event stream having correct data persisted (which isn't guaranteed). 

Would it be better to step back and think of a different way for the program under test to communicate it's status to the test runner? Maybe setting an annotation on itself or having the test running read the container logs?

markmandel:
> Thinking about this a bit -- it seems like the old way didn't work because it was looking for the "edges" of state transitions (which could be missed)

100% agreed :+1:  that's exactly what happens, it gets missed.

> and the new way is relying on the event stream having correct data persisted (which isn't guaranteed).

Not sure I understand this point -- why isn't the event stream not guaranteed to have the correct data? -- assuming it is covered appropriately by unit tests ([which I believe it is](https://github.com/googleforgames/agones/blob/master/pkg/sdkserver/sdkserver_test.go#L38)), it should have appropriate guarantees, no?

roberthbailey:
I was trying to find something in the Kubernetes docs to substantiate this, but alas I didn't see anything. 

My understanding is that Kubernetes events are not guaranteed. They can be dropped if the control plane is overloaded. They can be coalesced if duplicates are seen. I've always heard that you should never write a controller that relies on events for it's control logic, since you can't rely on events showing up, staying around, or having the correct cardinality. So I'm equally leery of writing test logic that relies on them.

markmandel:
Gotcha! I've yet to see the event stream not be consistent -- but it definitely does go away after a while, so I can see some concern there. I wonder how long events are held into for.

That being said -- I'm struggling to work out a better way to test this. This is the best idea I've got so far.

Does it make sense to run with this for now, but put a big comment on it re:potential flakiness down the road, and if we hit that, then we re-review to see if we can come up with a better answer?

markmandel:
I've just had 4 almost consecutive of these failures on https://github.com/googleforgames/agones/pull/1409 - I'm at the point where I may just disable this test if this isn't a valid solution, it's giving me that much frustration.

markmandel:
Oops, hit the wrong button! :relaxed:

roberthbailey:
> Does it make sense to run with this for now, but put a big comment on it re:potential flakiness down the road, and if we hit that, then we re-review to see if we can come up with a better answer?

I'm ok with that, since I've sufficiently communicated the potential for flakes to you.

markmandel:
I went to see how long events get persisted for - the only thing I could find was this:
https://github.com/kubernetes/kubernetes/issues/52521

Which seems to indicate that events are persisted for an hour (at least by default). Seems like it _should_ be long enough :man_shrugging:

roberthbailey:
1 hour is the default TTL, but Kubernetes service providers are free to change the default (and without doing some sort of benchmark you wouldn't know if or when they did). But I agree that it should be good enough for now.

Issue #1377, opened 2020-02-28
Terraform: No Agones helm installation when cluster parameter (e.g. Name) has changed (and it being recreated)


aLekSer:
Agones helm resource is not recreated if the cluster module was changed.

`terraform apply` twice on the next file (with changing the name of a cluster, or a zone), and helm module would not be updated, and thus there is a success apply without Agones on second time: https://github.com/GoogleCloudPlatform/agones/blob/5319b13d87bd9427fe36bf53a12c13c6773b061e/examples/terraform-submodules/gke/module.tf#L63

**What happened**:
There is an option to create a cluster with Agones, then update a cluster, and no Agones would be set up on second apply:
```
make gcloud-terraform-cluster
```
or using 
```
terraform apply -var project="<YOUR_GCP_ProjectID>" [-var agones_version="1.0.0"]`
```
described [here - Install with Terraform](https://agones.dev/site/docs/installation/terraform/gke/)

**What you expected to happen**:
Helm and Agones would be installed every time with `terraform apply`
**How to reproduce it (as minimally and precisely as possible)**:

```
make gcloud-terraform-cluster
vi ./examples/terraform-submodules/gke-local/main.tf // update something
make gcloud-terraform-cluster
```

**Anything else we need to know?**:
Logs:
```
//First `terraform apply` logs:                   
Apply complete! Resources: 6 added, 0 changed, 1 destroyed.                      
//Second apply, cluster was changed  (say name or node_count parameter)                                                                                                   
Apply complete! Resources: 3 added, 0 changed, 3 destroyed.     
$ kubectl get crds | grep agones | wc -l
0
```
I expect that both helm and cluster would be recreated.
**Environment**:
- Agones version: 1.4-rc
- Kubernetes version (use `kubectl version`): 1.14
- Cloud provider or hardware configuration: 
- Install method (yaml/helm): Terraform 0.12.21
- Troubleshooting guide log(s):
- Others:

aLekSer:
There is an option to taint some resources, but probably we can mark this resources as linked - create one after another.

markmandel:
@aLekSer is this fixed with the move to Helm 3?

aLekSer:
Hello @markmandel  give me 30 minutes to provide more details.

aLekSer:
The results are similar:
```
//First `terraform apply` logs:                   
Apply complete! Resources: 4 added, 0 changed, 0 destroyed.               
[...]

PR #1418, opened 2020-03-17
Fix wrong condition check for Memory limit

### LINKED ISSUE #1419: No validation when helm parameter `agones.image.sdk.cpuRequest` set less than `cpuLimit`
Similar to what we have in #1298 if we set cpuRequest < cpuLimits then no pods for GameServers could be created.

**What happened**:
Helm allows to set `agones.image.sdk` parameters which does not correlate to each other.

There should be an error on `helm install` step.

```
$ helm install ./install/helm/agones --namespace agones-system --name agones --set agones.image.tag=1.5.0-7791336 --set agones.image.sdk.cpuRequest=64Mi --set $ agones.image.sdk.cpuLimit=62Mi
$ kubectl apply -f ./examples/simple-udp/fleet.yaml
$ kubectl get gs
NAME                     STATE            ADDRESS   PORT   NODE   AGE
simple-udp-2b5l4-48lvp   Creating                                 0s
simple-udp-2b5l4-6pwzs   PortAllocation                           0s
simple-udp-2b5l4-ltjll   Shutdown                                 0s
simple-udp-2b5l4-trdtz   Error                                    0s
simple-udp-2b5l4-wxp2w   Shutdown                                 0s 
$ kubectl describe gss
...
                                       
  Normal  SuccessfulCreate  36s (x5 over 36s)  gameserverset-controller  (combined from similar events): Created gameserver: simple-udp-2b5l4-ccpkx                                                      
  Normal  SuccessfulDelete  36s                gameserverset-controller  Deleted gameserver in state Unhealthy: simple-udp-2b5l4-hcmfw                                                                   
  Normal  SuccessfulDelete  36s                gameserverset-controller  Deleted gameserver in state Unhealthy: simple-udp-2b5l4-pqj7r                                                                   
  Normal  SuccessfulDelete  36s (x2 over 36s)  gameserverset-controller  (combined from similar events): Deleted gameserver in state Unhealthy: simple-udp-2b5l4-rj4hs    
```

Similar issue with memoryRequest and memoryLimit parameters .

**Environment**:
- Agones version: 1.4
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.10-gke.27", GitCommit:"145f9e21a4515947d6fb10819e5a336aff1b6959", GitTreeState:"clean", BuildDate:"2020-02-21T18:01:40Z", GoVersion:"go1.12.12b4", Compiler:"gc", Platform:"linux/amd64"}
```
- Others:

aLekSer:
This is needed to start writing validation for these new limits as well as CPU ones.

Closes #1419.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"agones.dev/agones/pkg"
+	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	"agones.dev/agones/pkg/client/clientset/versioned"
 	"agones.dev/agones/pkg/client/informers/externalversions"
 	"agones.dev/agones/pkg/fleetautoscalers"
@@ -44,6 +45,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"gopkg.in/natefinch/lumberjack.v2"
+	corev1 "k8s.io/api/core/v1"
 	extclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/client-go/informers"
@@ -119,8 +121,11 @@ func main() {
 	logger.WithField("version", pkg.Version).WithField("featureGates", runtime.EncodeFeatures()).
 		WithField("ctlConf", ctlConf).Info("starting gameServer operator...")
 
-	if err := ctlConf.validate(); err != nil {
-		logger.WithError(err).Fatal("Could not create controller from environment or flags")
+	if errs := ctlConf.validate(); len(errs) != 0 {
+		for _, err := range errs {
+			logger.WithError(err)
+		}
+		logger.Fatal("Could not create controller from environment or flags")
 	}
 
 	// if the kubeconfig fails BuildConfigFromFlags will try in cluster config
@@ -382,14 +387,19 @@ type config struct {
 }
 
 // validate ensures the ctlConfig data is valid.
-func (c *config) validate() error {
+func (c *config) validate() []error {
+	validationErrors := make([]error, 0)
 	if c.MinPort <= 0 || c.MaxPort <= 0 {
-		return errors.New("min Port and Max Port values are required")
+		validationErrors = append(validationErrors, errors.New("min Port and Max Port values are required"))
 	}
 	if c.MaxPort < c.MinPort {
-		return errors.New("max Port cannot be set less that the Min Port")
+		validationErrors = append(validationErrors, errors.New("max Port cannot be set less that the Min Port"))
 	}
-	return nil
+	resourceErrors := agonesv1.ValidateResource(c.SidecarCPURequest, c.SidecarCPULimit, corev1.ResourceCPU)
+	validationErrors = append(validationErrors, resourceErrors...)
+	resourceErrors = agonesv1.ValidateResource(c.SidecarMemoryRequest, c.SidecarMemoryLimit, corev1.ResourceMemory)
+	validationErrors = append(validationErrors, resourceErrors...)
+	return validationErrors
 }
 
 type runner interface {
--- a/cmd/controller/main_test.go
+++ b/cmd/controller/main_test.go
@@ -55,6 +55,7 @@ func errorsContainString(t *testing.T, errs []error, expected string) {
 	for _, v := range errs {
 		if expected == v.Error() {
 			found = true
+			break
 		}
 	}
 	assert.True(t, found, "Was not able to find '%s'", expected)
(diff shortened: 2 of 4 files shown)


pkg/gameservers/controller.go
```diff
@@ -646,7 +646,7 @@ func (c *Controller) sidecar(gs *agonesv1.GameServer) corev1.Container {
 	if !c.sidecarCPULimit.IsZero() {
 		limits[corev1.ResourceCPU] = c.sidecarCPULimit
 	}
-	if !c.sidecarCPULimit.IsZero() {
+	if !c.sidecarMemoryLimit.IsZero() {
 		limits[corev1.ResourceMemory] = c.sidecarMemoryLimit
```
markmandel: Since we're here, do we want to put in some `logrus.Warnf()` log messages for both of these -- otherwise they fail silently.

WDYT?

aLekSer: Will update this

aLekSer: What Warn message should I add?
Similar to this one:
QoS would not be assigned to "Guaranteed" unless CPU and Memory are the same in Limit and Request for the sidecar?

markmandel: Something like:
```
	if !c.sidecarMemoryLimit.IsZero() {
		limits[corev1.ResourceMemory] = c.sidecarMemoryLimit
       } else {
             c.logger.Warnf("Sidecar memory limit of %s is less than or equal to zero. Skipping.", c.sidecarMemoryLimit)
       }
```

I was thinking something more like that? 

I feel like if people are changing memory and cpu limits, then they know what they are doing, and it's up to them to control it as they see fit. But a 0 or negative value isn't going to work, so we should warn people that we are skipping setting the value.

Sound good?

aLekSer: Mark thanks for this request, I have added validation on the place where we set those parameters, not where we use them.
So we will close #1419 issue.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -456,6 +457,24 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 	return causes, len(causes) == 0
 }
 
+// ValidateResource validates limit or Memory CPU resources used for containers in pods
+// If a GameServer is invalid there will be > 0 values in
+// the returned array
+func ValidateResource(request resource.Quantity, limit resource.Quantity, resourceName corev1.ResourceName) []error {
```
aLekSer: I have placed ValidateResource() here so we will check PodTemplate in a Fleet and GameServer using exact same function.

cmd/controller/main_test.go
```diff
@@ -0,0 +1,59 @@
...
+	assert.Len(t, errs, 2)
+	errorsContainString(t, errs, "Resource cpu request value must be non negative")
+}
+
+func errorsContainString(t *testing.T, errs []error, expected string) {
+	s := ""
+	for _, v := range errs {
+		s += v.Error()
+	}
+	assert.Contains(t, s, expected)
```
akremsa: Is it possible to put assert.Contains inside a for loop and exit once string match found?

aLekSer:
Flaky #1377 .

cmd/controller/main.go
```diff
@@ -382,14 +387,19 @@ type config struct {
 }
 
 // validate ensures the ctlConfig data is valid.
-func (c *config) validate() error {
+func (c *config) validate() []error {
+	validationErrors := make([]error, 0)
 	if c.MinPort <= 0 || c.MaxPort <= 0 {
-		return errors.New("min Port and Max Port values are required")
+		validationErrors = append(validationErrors, errors.New("min Port and Max Port values are required"))
```
markmandel: Love this. :+1:

cmd/controller/main_test.go
aLekSer: Done

cmd/controller/main_test.go
```diff
@@ -0,0 +1,61 @@
...
+	errs = c.validate()
+	assert.Len(t, errs, 2)
+	errorsContainString(t, errs, "Resource cpu request value must be non negative")
+}
+
+func errorsContainString(t *testing.T, errs []error, expected string) {
+	found := false
+	for _, v := range errs {
+		if expected == v.Error() {
+			found = true
```
akremsa: We can safely break the loop here because there is no point to check other elements of the slice

aLekSer: Added break

PR #1428, opened 2020-03-25
Gke module fixes

chrisst:
* Allow GKE version to be specified and default to the latest. So that version doesn't need to keep being bumped.

### DIFF
--- a/install/terraform/modules/gke/variables.tf
+++ b/install/terraform/modules/gke/variables.tf
@@ -25,11 +25,13 @@ variable "cluster" {
   type        = map
 
   default = {
-    "zone"             = "us-west1-c"
-    "name"             = "test-cluster"
-    "machineType"      = "n1-standard-4"
-    "initialNodeCount" = "4"
-    "project"          = "agones"
-    "network"          = "default"
+    "zone"                = "us-west1-c"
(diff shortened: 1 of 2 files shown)


install/terraform/modules/gke/variables.tf
```diff
@@ -25,11 +25,13 @@ variable "cluster" {
...
-    "initialNodeCount" = "4"
-    "project"          = "agones"
-    "network"          = "default"
+    "zone"                = "us-west1-c"
+    "name"                = "test-cluster"
+    "machineType"         = "n1-standard-4"
+    "initialNodeCount"    = "4"
+    "project"             = "agones"
+    "network"             = "default"
+    "min_cluster_version" = "latest"
```
markmandel: This should definitely not be latest. We only support a specific version of Kubernetes at any given point and time.

You can see this here: https://agones.dev/site/docs/installation/#usage-requirements

So I'm not sure if we should allow people to customise the Kubernetes version?

(Also, looks like we ran into some rate limiting by github, which is why the tests where failing)

chrisst: Ah sorry didn't pay attention to this, sounds like opinionated versioning in the module is the right call :)

PR #1440, opened 2020-04-01
Alpha GameServer attributes added to SDK

markmandel:
Rather than recreating SDK.GameServer() and SDK.WatchGameServer() in the alpha SDK, I thought it would be easiest both in the short and long term to mimic what we have done in the GameServer CRD, and populate the Alpha attributes from these SDK methods when the feature flag for Player Tracking is enabled.

This is done by adding logic to the conversion function that converts the K8s GameServer definition into the SDK GameServer definition.

This includes regenerated gRPC code for the supported clients.

I also had to fix the line endings in the C# gen.sh, so including that in this PR.

Work on #1033


pkg/sdkserver/sdk.go
```diff
@@ -66,5 +67,13 @@ func convert(gs *agonesv1.GameServer) *sdk.GameServer {
 		result.Status.Ports = append(result.Status.Ports, grpcPort)
 	}
 
+	if runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		result.Status.Players = &sdk.GameServer_Status_PlayerStatus{}
```
roberthbailey, line 71: Is the idea to always include an empty Players block even if gs.Status.Players is nil? Why not do

```
if gs.Status.Players != nil {
  result.Status.Players = &sdk.GameServer_Status_PlayerStatus{
    Count: gs.Status.Players.Count,
    Capacity: gs.Status.Players.Capacity,
  }
}
```

pkg/sdkserver/sdk_test.go
```diff
@@ -75,13 +79,33 @@ func TestConvert(t *testing.T) {
...
+		gs := fixture.DeepCopy()
+
+		sdkGs := convert(gs)
+		eq(t, fixture, sdkGs)
+		assert.Zero(t, sdkGs.ObjectMeta.DeletionTimestamp)
+		assert.Nil(t, sdkGs.Status.Players)
+
+		now := metav1.Now()
+		gs.DeletionTimestamp = &now
+		sdkGs = convert(gs)
```
roberthbailey, line 94: In this test you convert twice and in the test with the feature enabled you only do it once. Why the difference?

pkg/sdkserver/sdk.go
markmandel: That's a good question. My original thought was to have less opportunities for panics on nil, but I think your way is better. I'll make the change :+1:

pkg/sdkserver/sdk_test.go
markmandel: The second convert was to test if the DeletionTimestamp was converted correctly. Once we had had tested that once, there didn't seem to be any reason to repeat the test.

The second block could be moved to it's own `t.Run()` block. That would probably read a little nicer too. WDYT?

roberthbailey: I think that is much clearer. Thanks.

PR #1446, opened 2020-04-02
Update to OpenCensus version 0.22.3

aLekSer:
This should fix Prometheus error messages on exporting with empty tags
list . Allocator Metrics.

Takeover this PR  #893
Part of #1330 

Fixing this error in logs (happened periodically):
```
textPayload: "2020/02/07 15:14:14 Failed to export to Prometheus: inconsistent label cardinality: expected 1 label values but got 0 in []string(nil)
```


aLekSer:
Tested this change with Grafana and Prometheus.
Error with `inconsistent label cardinality` now does not happen in the logs.
https://github.com/googleforgames/agones/issues/1330#issuecomment-584737095

markmandel:
@roberthbailey @pooneh-m should we aim to merge this in this release, or hold it until next release to give it time to bake? (Since RC is next Tuesday)

I'm inclined to push this to the next release, just to give it time to be tested over time. WDYT?

aLekSer:
Setting it to on-hold. Test on Stackdriver failed:
```
2020/04/03 16:11:56 Failed to export to Stackdriver: rpc error: code = InvalidArgument desc = Field timeSeries[17].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.
```

aLekSer:
Strange enough, but I was fixing similar thing a while ago:
https://github.com/googleforgames/agones/pull/554

roberthbailey:
@aLekSer - can you split the vendor changes into a separate commit from the changes to pkg/ to make it easier to see how much impact there is on our code?

@markmandel - at first glance this seems to be making non-trivial code changes to our metric exporting which makes me tend to agree to push it to the next release to give it time to bake. On the other hand, how much extra testing does that actually get us? how much testing of this code path happens during our e2es? and would anyone test a pre-release build to verify the functionality on top of what the author and reviewer do? If not and it isn't destabilizing, then that would argue to get it in now since we would only get feedback on how well it works once it's released.

aLekSer:
Similar error is mentioned here also:
https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/issues/234

markmandel:
Just taking a look - at first glance LGTM, @cyriltovena do you have time to double check this, since it was originally your code?

cyriltovena:
I'll run through it quickly now

pkg/gameserverallocations/metrics.go
```diff
@@ -54,12 +54,12 @@ func init() {
 
 // default set of tags for latency metric
 var latencyTags = []tag.Mutator{
-	tag.Insert(keyMultiCluster, "none"),
-	tag.Insert(keyClusterName, "none"),
-	tag.Insert(keySchedulingStrategy, "none"),
-	tag.Insert(keyFleetName, "none"),
-	tag.Insert(keyNodeName, "none"),
-	tag.Insert(keyStatus, "none"),
+	tag.Insert(keyMultiCluster, ""),
```
cyriltovena, line 57: I'm guessing now this is allowed to be empty ?

aLekSer: Insert() should add this tag https://github.com/census-instrumentation/opencensus-go/blob/master/tag/map.go#L107:6
I will check tomorrow on how this applies to this metrics. Probably add a test for this.

cyriltovena:
image
image
image
image
image
image
image
image
image
image

cyriltovena:
The only one that does not work is the go-client cache. But I assume I did not need cache for this try out ?

aLekSer:
Thanks @cyriltovena so much for providing screenshots, I need to fix stackdriver error message which is left for this PR.
```
2020/04/03 16:11:56 Failed to export to Stackdriver: rpc error: code = InvalidArgument desc = Field timeSeries[17].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.
```

pkg/gameserverallocations/metrics.go
aLekSer: @cyriltovena This change break Stackdriver exporter, so I readded `"none"` for now. I think this change could be done separately.

aLekSer:
With "none" in tag values export to Stackdriver is working.
Stackdriver output

aLekSer:
Debugged opencensus stackdriver exporter code a bit, added more logging:
```
E 2020-04-13T07:28:37.571709353Z 2020/04/13 07:28:37 Failed to export this to Stackdriver 0: gameservers_count
E 2020-04-13T07:28:37.571744133Z 2020/04/13 07:28:37 Failed to export this to Stackdriver 1: k8s_client_workqueue_unfinished_work_seconds
E 2020-04-13T07:28:37.571749105Z 2020/04/13 07:28:37 Failed to export this to Stackdriver 2: fleets_replicas_count
E 2020-04-13T07:28:37.571752888Z 2020/04/13 07:28:37 Failed to export this to Stackdriver 3: k8s_client_workqueue_work_duration_seconds
E 2020-04-13T07:28:37.571796974Z 2020/04/13 07:28:37 Failed to export to Stackdriver: rpc error: code = InvalidArgument desc = Field timeSeries[8].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.
```
Stackdriver export is not working on the most recent stackdriver exporter also.

The error with Stackdriver metrics is gone only when I comment out whole file: https://github.com/googleforgames/agones/blob/master/pkg/metrics/kubernetes_client.go
Need to understand which metric or group of them lead to this situation, because error on exporter like `timeSeries[8]` does not help to find erroneous metric.

aLekSer:
Narrowing down - `k8s.io/client-go/util/workqueue` metrics are failing with Stackdriver exporter `workqueue.SetProvider(c)`.

aLekSer:
Was able to fix Stackdriver exporter uploading if parameter is changed from `view.Distribution(0)` or `view.Distribution()` (both cause an error like above) to something like:
```
	runtime.Must(view.Register(&view.View{
		Name:        "k8s_client_workqueue_latency_seconds",
		Measure:     workQueueLatencyStats,
		Description: "How long an item stays in the work queue.",
		Aggregation: view.Distribution(0.0001),
		TagKeys:     []tag.Key{keyQueueName},
	}))
```

markmandel:
How is this PR looking? I think it's good to merge, but wanted to check first.

aLekSer:
This PR will fix errors in Prometheus exporter, and I will produce a separate PR for fixing Stackdriver exporter `view.Distribution(..)` - with multiple parameters. (Even adding 0.0001 instead of 0 is fixing Stackdriver)

PR #1447, opened 2020-04-02
SDK: GetPlayerCount & PlayerDis/Connect

markmandel:
Implementation of GetPlayerCount, PlayerConnect and PlayerDisconnect on the SDK Server.

Work on #1033

### DIFF
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -53,10 +53,12 @@ import (
 type Operation string
 
 const (
-	updateState          Operation = "updateState"
-	updateLabel          Operation = "updateLabel"
-	updateAnnotation     Operation = "updateAnnotation"
-	updatePlayerCapacity Operation = "updatePlayerCapacity"
+	updateState             Operation     = "updateState"
+	updateLabel             Operation     = "updateLabel"
+	updateAnnotation        Operation     = "updateAnnotation"
+	updatePlayerCapacity    Operation     = "updatePlayerCapacity"
+	updatePlayerCount       Operation     = "updatePlayerCount"
+	playerCountUpdatePeriod time.Duration = time.Second
 )
 
 var (
@@ -96,6 +98,8 @@ type SDKServer struct {
 	reserveTimer       *time.Timer
 	gsReserveDuration  *time.Duration
 	gsPlayerCapacity   int64
+	gsPlayerCount      int64
+	connectedPlayers   map[string]bool
 }
 
 // NewSDKServer creates a SDKServer that sets up an
@@ -129,6 +133,7 @@ func NewSDKServer(gameServerName, namespace string, kubeClient kubernetes.Interf
 		gsAnnotations:      map[string]string{},
 		gsUpdateMutex:      sync.RWMutex{},
 		gsWaitForSync:      sync.WaitGroup{},
+		connectedPlayers:   map[string]bool{},
 	}
 
 	s.informerFactory = factory
@@ -230,11 +235,12 @@ func (s *SDKServer) Run(stop <-chan struct{}) error {
 		go wait.Until(s.runHealth, s.healthTimeout, stop)
 	}
 
-	// populate player capacity value
+	// populate player tracking values
 	if runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
 		s.gsUpdateMutex.Lock()
 		if gs.Status.Players != nil {
 			s.gsPlayerCapacity = gs.Status.Players.Capacity
+			s.gsPlayerCount = gs.Status.Players.Count
 		}
 		s.gsUpdateMutex.Unlock()
 	}
@@ -272,6 +278,8 @@ func (s *SDKServer) syncGameServer(key string) error {
 		return s.updateAnnotations()
 	case updatePlayerCapacity:
 		return s.updatePlayerCapacity()
+	case updatePlayerCount:
+		return s.updatePlayerCount()
 	}
 
 	return errors.Errorf("could not sync game server key: %s", key)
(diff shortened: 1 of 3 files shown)


pkg/sdkserver/sdkserver.go
```diff
@@ -547,21 +555,68 @@ func (s *SDKServer) stopReserveTimer() {
 // [Stage:Alpha]
 // [FeatureFlag:PlayerTesting]
 func (s *SDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerId) (*alpha.Empty, error) {
-	panic("implement me")
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return nil, errors.New(string(runtime.FeaturePlayerTracking) + " not enabled")
+	}
+	s.logger.WithField("PlayerId", id).Debug("Player Connected")
+
+	s.gsUpdateMutex.RLock()
```
roberthbailey, line 563: This looks like a potential race condition. You grab a read lock, drop it, and then conditionally grab a write lock depending on the result of the first read. I doubt there will be so much lock contention that just grabbing the write lock and doing the entire check and update inside of it will cause performance issues, but it will guarantee safety of mutations.

pkg/sdkserver/sdkserver.go
```diff
@@ -547,21 +555,68 @@ func (s *SDKServer) stopReserveTimer() {
...
 // [Stage:Alpha]
 // [FeatureFlag:PlayerTesting]
 func (s *SDKServer) PlayerDisconnect(ctx context.Context, id *alpha.PlayerId) (*alpha.Empty, error) {
-	panic("implement me")
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return nil, errors.New(string(runtime.FeaturePlayerTracking) + " not enabled")
+	}
+	s.logger.WithField("PlayerId", id).Debug("Player Disconnected")
+
+	s.gsUpdateMutex.RLock()
```
roberthbailey, line 591: Same here.

pkg/sdkserver/sdkserver.go
roberthbailey: I'm also wondering if using the standard lock and then defer unlock would be any different than explicitly unlocking. Does the enqueue call have the potential to block?

pkg/sdkserver/sdkserver.go
```diff
@@ -547,21 +555,68 @@ func (s *SDKServer) stopReserveTimer() {
...
+	s.logger.WithField("PlayerId", id).Debug("Player Connected")
+
+	s.gsUpdateMutex.RLock()
+	// check if the player id is in the player map
+	_, ok := s.connectedPlayers[id.PlayerId]
+	s.gsUpdateMutex.RUnlock()
+
+	// this player has already connected, so exit
+	if ok {
+		return &alpha.Empty{}, nil
```
roberthbailey, line 570: Should this case return an error ("player already registered")? Right now there are no error returns unless the feature flag is disabled.

pkg/sdkserver/sdkserver.go
```diff
@@ -547,21 +555,68 @@ func (s *SDKServer) stopReserveTimer() {
...
 // [Stage:Alpha]
 // [FeatureFlag:PlayerTesting]
 func (s *SDKServer) GetPlayerCount(ctx context.Context, _ *alpha.Empty) (*alpha.Count, error) {
-	panic("implement me")
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return nil, errors.New(string(runtime.FeaturePlayerTracking) + " not enabled")
+	}
+	s.gsUpdateMutex.RLock()
+	defer s.gsUpdateMutex.RUnlock()
+	return &alpha.Count{Count: s.gsPlayerCount}, nil
```
roberthbailey, line 619: I'm also curious why you keep an extra variable and don't just take the size of the map here. The way you have it means you have to ensure that your code guarantees that the size of the map equals gsPlayerCount.

pkg/sdkserver/sdkserver.go
markmandel: Since this going back up to gRPC, rather than throwing an error (which we could pick up if this was just Go), should PlayerConnect and PlayerDisconnect return a bool value, as to whether or not they were successful?

pkg/sdkserver/sdkserver.go
markmandel: That's a really good point. Let me switch this out!

pkg/sdkserver/sdkserver.go
markmandel: > This looks like a potential race condition.

Well spotted. I agree, I will make this change here and below.

> I'm also wondering if using the standard lock and then defer unlock would be any different than explicitly unlocking. Does the enqueue call have the potential to block?

The enqueue could possibly result in a deadlock, since the sync function will try and lock the same mutex. So to be safe, we Unlock before we enqueue.

pkg/sdkserver/sdkserver.go
markmandel: So looking at this some more - I realised it's because of this:
https://github.com/googleforgames/agones/blob/ae51639fcdab374871fb832d09c3649eb17bc0b1/pkg/sdkserver/sdkserver.go#L243

Wherein we set the value of gsPlayerCount from the current Count from status.

In theory, this was just in case the sidecar crashes and restarts -- but if that happens, we lose all the connected player map data anyway, so all kind of things are going to go wrong.

So.... I'm not quite sure what the right answer is? At least with this split between count and the map, the number would be mostly correct (and not potentially revert to 0 on crash).

What are you thoughts?

markmandel: So I lay in bed thinking about this last night -- and I kept coming back to: I think we should store the connected player tokens on the CRD.

Reasons being:
1. If the sidecar crashes, we no longer lose all this connection data (although we do lose any in-memory transition data, but that is true of all SDK methods)
1. We don't gain API hits, as we are already updating the GameServer resource with a delay, so we could do it at the same time (so also eventually consistent model). 
1. We could implement a `SDK.IsPlayerConnected(token): bool` method now
1. We could also drop the webhook requirements, as the a Kubernetes GameServer update would have all the information about a player connecting/disconnecting, since it has both the before and after states (and would be a nice batch operation).

This does mean we'd be storing more data on the CRD, but let's do some back of napkin calculations to see if is anything to worry about.

Let's assume 100 players per GameServer, and each has a Guid - so max of  100*16 bytes = 
1600 bytes per GameServer.

Let's say 500 nodes with 30 gameserver per node, so 15000 GaneServers in a cluster. 

Total extra data: 15000 × 1600 = 24MB -- which seems doable.

If you think my points hold water, I'll update the design ticket for a more general consensus - but WDYT?

cyriltovena: I think storing player's token in the CRD (GameServer I guess) is fine. 

Later you could add an upper limit that could be globally configured. I don't think there's will be game with more than 300 player on the same gameserver. 

Another idea or one that could be combine with:

If the side car crashes and not the gameserver it will come back  with the same filesystem ? You could use some sort of write ahead log if so. But honestly I'm not worried about this too much.

roberthbailey: I think storing the players makes sense, because having them only in memory in the sidecar seems like a recipe for some un-reproducible bugs down the road. 

In addition to the amount of data being stored, how does this impact our update rates on the game server crd?

markmandel: > In addition to the amount of data being stored, how does this impact our update rates on the game server crd?

I would suggest only updating the list at the same time we are updating the player counts (with a delay, so we can batch updates), so there is no additional increase in API requests above what we are doing already for player tracking.

markmandel:
Closing PR, and will come back around to it when we finalise things on https://github.com/googleforgames/agones/issues/1033#issuecomment-613179001

PR #1451, opened 2020-04-06
Flaky: TestLocalSDKServerPlayerCapacity

markmandel:
Go routine could fire after we run the method that fires
the watched event, so let's wait until we know watching is
actively working.


markmandel:
SDK Conformance flakiness :disappointed:

PR #1458, opened 2020-04-09
Fixed flaky TestGameServerWithPortsMappedToMultipleContainers

### LINKED ISSUE #1450: Flaky: TestGameServerWithPortsMappedToMultipleContainers
```
--- FAIL: TestGameServerWithPortsMappedToMultipleContainers (41.07s)
    gameserver_test.go:527: Could not message GameServer: read udp 192.168.10.2:45098->35.233.192.59:7273: read: connection refused
```

Seen this crop up a few times.

akremsa:
Closes #1450 
I've run this test more than 100 times and found that both pings can fail. 

Fix proof:
Game port №1:
image

Game port №2:
image

Suggestion: Is there any opportunity to stop tests immediately by pressing ctr+c/ctr+z if we run them inside a container? It can take some time for some goroutines to finish their work.

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	"fmt"
 	"net"
+	"strings"
 	"testing"
 	"time"
 
@@ -480,17 +481,19 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
 	t.Parallel()
 	firstContainerName := "udp-server"
 	secondContainerName := "second-udp-server"
+	firstPort := "gameport"
+	secondPort := "second-gameport"
 	gs := &agonesv1.GameServer{ObjectMeta: metav1.ObjectMeta{GenerateName: "udp-server", Namespace: defaultNs},
 		Spec: agonesv1.GameServerSpec{
 			Container: firstContainerName,
 			Ports: []agonesv1.GameServerPort{{
 				ContainerPort: 7654,
-				Name:          "gameport",
+				Name:          firstPort,
 				PortPolicy:    agonesv1.Dynamic,
 				Protocol:      corev1.ProtocolUDP,
 			}, {
 				ContainerPort: 5000,
-				Name:          "second-gameport",
+				Name:          secondPort,
 				PortPolicy:    agonesv1.Dynamic,
 				Protocol:      corev1.ProtocolUDP,
 				Container:     &secondContainerName,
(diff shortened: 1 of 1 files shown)


test/e2e/gameserver_test.go
```diff
@@ -522,17 +525,39 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
...
+			}
+			return err == nil && strings.Contains(res, "Ping 2"), nil
+		})
+
+		reply <- errPoll
+	}(pollCh2)
+
+	err1 := <-pollCh1
+	err2 := <-pollCh2
+	assert.Nil(t, err1, fmt.Sprintf("expected no errors after polling a port: %s", firstPort))
```
aLekSer, line 559: No need to add fmt here as we have msgAndArgs parameter in assert.Nil():
```suggestion
	assert.Nil(t, err1, "expected no errors after polling a port: %s", firstPort)
```

test/e2e/gameserver_test.go
```diff
@@ -522,17 +525,39 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
...
+			if err != nil {
+				t.Logf("Could not message GameServer on %s: %v. Will try again...", firstPort, err)
+			}
+			return err == nil && strings.Contains(res, "Ping 1"), nil
+		})
+		reply <- errPoll
+	}(pollCh1)
+
+	go func(reply chan error) {
+		errPoll := wait.PollImmediate(interval, timeOut, func() (done bool, err error) {
```
aLekSer, line 546: Strange enough that wait.PollImmediate() parameters are different for similar goroutine functions. Is there an option to declare one function and pass only `firstPort` string and `pollCh1`. Thus we can make code easier to read.

akremsa: I prefer to keep it as is because otherwise, we will have too much function argument

test/e2e/gameserver_test.go
```diff
@@ -522,17 +525,40 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("Could not message GameServer: %v", err)
-	}
-	assert.Contains(t, secondContainerReply, "Ping 2")
+	pollCh1 := make(chan error)
+	pollCh2 := make(chan error)
+
+	go func() {
+		expectedMsg := "Ping 1"
+		errPoll := wait.PollImmediate(interval, timeOut, func() (done bool, err error) {
```
markmandel, line 536: Love the polling - seems like a good solution!

test/e2e/gameserver_test.go
```diff
@@ -522,17 +525,40 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
...
 
-	secondContainerReply, err := e2eframework.SendGameServerUDPToPort(readyGs, "second-gameport", "Ping 2")
-	if err != nil {
-		t.Fatalf("Could not message GameServer: %v", err)
-	}
-	assert.Contains(t, secondContainerReply, "Ping 2")
+	pollCh1 := make(chan error)
+	pollCh2 := make(chan error)
+
+	go func() {
```
markmandel, line 534: Is there a reason this couldn't be 2 polling statements, one after another?

Then it would make this test much simpler, and remove the need for channels. I don't think we have to test for both UDP endpoints at the same time.

WDYT?

akremsa: The main idea why I've introduced 2 goroutines is that in a worst-case scenario if we ping ports one by one, this **single** test will run 2 mins at least - there are 2 timeouts, 60 sec for each ping. According to my tests, there is a little chance of this scenario, but I decided to take it into account. I've tried to make this code as simple as possible. 
Do you think we need to call polling statements one after another anyway?

markmandel: Since failure should be an exception not the regular path, I would have them one after the other, for sake of simplicity of the testing code. This would make it easier to maintain going forward.

One optimisation you could make that would cut things down time in failure scenarios, is that if the first one fails, do a `assert.FailNow()`; to fast fail the test and stop the second one from even starting.

akremsa: Ok, good suggestion!

akremsa:
@markmandel  Applied your review notes.

test/e2e/gameserver_test.go
```diff
@@ -522,17 +525,31 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
...
+	expectedMsg2 := "Ping 2"
+	errPoll = wait.PollImmediate(interval, timeOut, func() (done bool, err error) {
+		res, err := e2eframework.SendGameServerUDPToPort(readyGs, secondPort, expectedMsg2)
+		if err != nil {
+			t.Logf("Could not message GameServer on %s: %v. Will try again...", secondPort, err)
+		}
+		return err == nil && strings.Contains(res, expectedMsg2), nil
+	})
+
+	assert.Nil(t, errPoll, "expected no errors after polling a port: %s", secondPort)
```
markmandel, line 552:
```suggestion
	assert.NoError(t, errPoll, "expected no errors after polling a port: %s", secondPort)
```

Just displays the details of the passed in error message. Other than that, this looks good to go :+1:

akremsa: Done

Issue #1013, opened 2019-08-24
Add GameServer state duration metric


cyriltovena:
**Is your feature request related to a problem? Please describe.**

Being able to know how long a GameServer stays in a specific state can be very useful when troubleshooting Agones Performance. This could allow users to know how long a GameServer stays unallocated or even how long it takes to get it ready. We could also figure if  game sessions are running correctly based on the average session time.

**Describe the solution you'd like**
We could watch for state changes via the Kubernetes Event API and record for each how long the last state was set for. 

**Describe alternatives you've considered**
We could also simply adds a new timestamp property to the GameServer CRD  and record duration when the state change. 

>**This might rely on the watch API of the GameServer CRD and Events. We should verify how using the watch api affect the duration.**

**Additional context**
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#-strong-read-operations-strong--289

aLekSer:
Relates to (duplicates) https://github.com/googleforgames/agones/issues/831

aLekSer:
I have a working solution for this with injecting the code only on State change. So we can keep track of all `GameServers` last state change timestamps in a map or similar structure (could add retention period or LRU cache) and record previous state duration metric at that moment.
See the branch with a solution:
https://github.com/aLekSer/agones/tree/metrics/state-duration
Need to add proper Graphana Chart into Dashboard configs.

aLekSer:
@markmandel Do you think we could close this issue as well?

markmandel:
I think so! If there are new requirements, let's open a new issue. :raised_hands:

PR #1468, opened 2020-04-13
New metric - state duration

### LINKED ISSUE #831: Expose GameServer state change metrics
I would like to know how long GameServers spend in a given state - for example how long does a game session last i.e. how long does it take to go from Allocated to Shutdown or how long does it take for a GameServer to become Ready. This is very helpful during deployments where a dramatic change might be a cause for concern.

I discussed this with @markmandel and we have two possible approaches:
1. We could add code to watch events. We might be able to use ListOptions to filter it down to events about GameServers. Then the code maintains an internal cache of state -> timestamp for each game server. The cache might evict based on LRU. LRU would mostly work if we expect game servers to roughly run for the same time.
2. Modify the GameServer CRD by adding a list of <state, timestamp> pairs to it. State updates will append to this list.

aLekSer:
Add `GameServer` state duration metric reusing current function `recordGameServerStatusChanges()`

For #1013
Closes #831

### DIFF
--- a/pkg/metrics/controller.go
+++ b/pkg/metrics/controller.go
@@ -43,7 +43,19 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-const noneValue = "none"
+const (
+	noneValue = "none"
+
+	// GameServersStateCount is the size of LRU cache and should contain all gameservers state changes
+	// Upper bound could be estimated as 10_000 of gameservers in total each moment, 10 state changes per each gameserver
+	// and about 10 minutes for a game session, and 6 gameservers per hour.
+	// For one hour 600k capacity would be enough, even if no records would be deleted.
+	// And calcDuration algorithm is removing those records, which already has been changed (old statuses).
+	// Key is Namespace, fleetName, GameServerName, State and float64 as value.
+	// Roughly 256 + 63 + 63 + 16 + 4 = 400 bytes per every record.
+	// In total we would have 229 MiB of space required to store GameServer State durations.
+	GameServersStateCount = 600_000
+)
 
 var (
 	// MetricResyncPeriod is the interval to re-synchronize metrics based on indexed cache.
@@ -90,8 +102,8 @@ func NewController(
 
 	// GameServerStateLastChange Contains the time when the GameServer
 	// changed its state last time
-	// on delete remove GameServerName key
-	lruCache, err := lru.New(1 << 12)
+	// on delete and state change remove GameServerName key
+	lruCache, err := lru.New(GameServersStateCount)
 	if err != nil {
 		logger.WithError(err).Fatal("Unable to create LRU cache")
 	}
@@ -279,17 +291,11 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
 		}
 		recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(newGs.Status.State)),
 			tag.Upsert(keyFleetName, fleetName), tag.Upsert(keyNamespace, newGs.GetNamespace())}, gameServerTotalStats.M(1))
-		if newGs.Status.State == agonesv1.GameServerStateReady {
-			diff := time.Now().Sub(newGs.ObjectMeta.CreationTimestamp.Local()).Seconds()
-			c.logger.Info("Time taken to become ready", diff)
-			recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(newGs.Status.State)),
-				tag.Upsert(keyFleetName, fleetName)}, gsReadyDuration.M(diff*1000.))
-			tag.Upsert(keyFleetName, fleetName)}, gameServerTotalStats.M(1))
 
 		// Calculate the duration of the current state
 		duration, err := c.calcDuration(oldGs, newGs)
 		if err != nil {
-			c.logger.Debug(err.Error())
+			c.logger.Warn(err.Error())
 		} else {
 			recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(oldGs.Status.State)),
 				tag.Upsert(keyFleetName, fleetName)}, gsStateDurationSec.M(duration))
@@ -300,7 +306,8 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
 // calcDuration - calculate duration between state changes
 // store current time from creationTimestamp for each update received
 // Assumptions: there is a possibility that one of the previous state change timestamps would be evicted,
-// this measure would be skipped
+// this measure would be skipped. This is a trade off between accuracy of distribution calculation and the performance.
+// Presumably occasional miss would not change the statistics too much.
 func (c *Controller) calcDuration(oldGs, newGs *agonesv1.GameServer) (duration float64, err error) {
 	// currentTime - GameServer time from its start
 	currentTime := c.now().UTC().Sub(newGs.ObjectMeta.CreationTimestamp.Local().UTC()).Seconds()
--- a/pkg/metrics/controller.go
+++ b/pkg/metrics/controller.go
@@ -55,17 +55,18 @@ func init() {
 
 // Controller is a metrics controller collecting Agones state metrics
 type Controller struct {
-	logger           *logrus.Entry
-	gameServerLister listerv1.GameServerLister
-	nodeLister       v1.NodeLister
-	gameServerSynced cache.InformerSynced
-	fleetSynced      cache.InformerSynced
-	fasSynced        cache.InformerSynced
-	nodeSynced       cache.InformerSynced
-	lock             sync.Mutex
-	stateLock        sync.Mutex
-	gsCount          GameServerCount
-	faCount          map[string]int64
+	logger                    *logrus.Entry
+	gameServerLister          listerv1.GameServerLister
+	nodeLister                v1.NodeLister
+	gameServerSynced          cache.InformerSynced
+	fleetSynced               cache.InformerSynced
+	fasSynced                 cache.InformerSynced
+	nodeSynced                cache.InformerSynced
+	lock                      sync.Mutex
+	stateLock                 sync.Mutex
+	gsCount                   GameServerCount
+	faCount                   map[string]int64
+	gameServerStateLastChange *lru.Cache
 }
 
 // NewController returns a new metrics controller
@@ -84,21 +85,25 @@ func NewController(
 	fasInformer := fas.Informer()
 	node := kubeInformerFactory.Core().V1().Nodes()
 	nodeInformer := node.Informer()
-	lruCache, err := lru.New(1 << 24)
+
+	// GameServerStateLastChange Contains the time when the GameServer
+	// changed its state last time
+	// on delete remove GameServerName key
+	lruCache, err := lru.New(1 << 12)
 	if err != nil {
 		logger.Error("Could not create LRU cache ", err)
 	}
-	GameServerStateLastChange = lruCache
 
 	c := &Controller{
-		gameServerLister: gameServer.Lister(),
-		nodeLister:       node.Lister(),
-		gameServerSynced: gsInformer.HasSynced,
-		fleetSynced:      fInformer.HasSynced,
-		fasSynced:        fasInformer.HasSynced,
-		nodeSynced:       nodeInformer.HasSynced,
-		gsCount:          GameServerCount{},
-		faCount:          map[string]int64{},
+		gameServerLister:          gameServer.Lister(),
+		nodeLister:                node.Lister(),
+		gameServerSynced:          gsInformer.HasSynced,
+		fleetSynced:               fInformer.HasSynced,
+		fasSynced:                 fasInformer.HasSynced,
+		nodeSynced:                nodeInformer.HasSynced,
+		gsCount:                   GameServerCount{},
+		faCount:                   map[string]int64{},
+		gameServerStateLastChange: lruCache,
 	}
 
 	c.logger = runtime.NewLoggerWithType(c)
@@ -279,37 +284,45 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
 			tag.Upsert(keyFleetName, fleetName)}, gameServerTotalStats.M(1))
 
 		// Calculate the duration from the start of the GameServer
-		c.stateLock.Lock()
-		defer c.stateLock.Unlock()
-		err := calcDuration(newGs, oldGs)
+		err := c.calcDuration(newGs, oldGs)
 		if err != nil {
 			c.logger.Info(err.Error())
 		}
 	}
 }
 
-func calcDuration(newGs, oldGs *agonesv1.GameServer) error {
+func (c *Controller) calcDuration(newGs, oldGs *agonesv1.GameServer) error {
+	c.stateLock.Lock()
+	defer c.stateLock.Unlock()
 	fleetName := newGs.Labels[agonesv1.FleetNameLabel]
-	diff := time.Now().Sub(newGs.ObjectMeta.CreationTimestamp.Local()).Seconds()
+	if fleetName == "" {
+		fleetName = defaultFleetTag
+	}
+	diff := time.Now().UTC().Sub(newGs.ObjectMeta.CreationTimestamp.Local().UTC()).Seconds()
+	//TODO: add namespace here, remove status to the value
 	key := newGs.ObjectMeta.Name + "/" + string(newGs.Status.State)
 	oldKey := oldGs.ObjectMeta.Name + "/" + string(oldGs.Status.State)
-	if !GameServerStateLastChange.Contains(key) {
-		GameServerStateLastChange.Add(key, diff)
+	//TODO: here we should check if we loose part of messages and old state was load_or_store
+	//Switch from structure in key to a structure
+	if !c.gameServerStateLastChange.Contains(key) {
+		c.gameServerStateLastChange.Add(key, diff)
+		c.logger.Infof("Adding new key %s, %f", key, diff)
 		recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(newGs.Status.State)),
-			tag.Upsert(keyFleetName, fleetName)}, gsStateDuration.M(diff*1000.))
+			tag.Upsert(keyFleetName, fleetName)}, gsStateDuration.M(diff))
 	} else {
-		val, ok := GameServerStateLastChange.Get(key)
+		val, ok := c.gameServerStateLastChange.Get(key)
 		if !ok {
 			return errors.New("Could not find expected key")
 		}
 		duration := diff - val.(float64)
 		recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(oldGs.Status.State)),
-			tag.Upsert(keyFleetName, fleetName)}, gsStateDuration.M(duration*1000.))
-		GameServerStateLastChange.Add(key, diff)
-		GameServerStateLastChange.Remove(oldKey)
+			tag.Upsert(keyFleetName, fleetName)}, gsStateDuration.M(duration))
+		c.logger.Infof("Recording new metric %s, %f", key, duration)
+		c.gameServerStateLastChange.Add(key, diff)
+		c.gameServerStateLastChange.Remove(oldKey)
 	}
 	if newGs.Status.State == agonesv1.GameServerStateShutdown {
-		GameServerStateLastChange.Remove(oldKey)
+		c.gameServerStateLastChange.Remove(oldKey)
 	}
 	return nil
 }
--- a/pkg/metrics/controller.go
+++ b/pkg/metrics/controller.go
@@ -82,6 +82,7 @@ func NewController(
 	fasInformer := fas.Informer()
 	node := kubeInformerFactory.Core().V1().Nodes()
 	nodeInformer := node.Informer()
+	//events := kubeClient.CoreV1().Events(defaultNs)
 
 	c := &Controller{
 		gameServerLister: gameServer.Lister(),
@@ -264,6 +265,12 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
 		}
 		recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(newGs.Status.State)),
 			tag.Upsert(keyFleetName, fleetName), tag.Upsert(keyNamespace, newGs.GetNamespace())}, gameServerTotalStats.M(1))
+		if newGs.Status.State == agonesv1.GameServerStateReady {
+			diff := time.Now().Sub(newGs.ObjectMeta.CreationTimestamp.Local()).Seconds()
+			c.logger.Info("Time taken to become ready", diff)
+			recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(newGs.Status.State)),
+				tag.Upsert(keyFleetName, fleetName)}, gsReadyDuration.M(diff*1000.))
+		}
 	}
 }
 
--- a/pkg/metrics/controller.go
+++ b/pkg/metrics/controller.go
@@ -68,6 +68,7 @@ type Controller struct {
 	gsCount                   GameServerCount
 	faCount                   map[string]int64
 	gameServerStateLastChange *lru.Cache
+	now                       func() time.Time
 }
 
 // NewController returns a new metrics controller
@@ -105,6 +106,9 @@ func NewController(
 		gsCount:                   GameServerCount{},
 		faCount:                   map[string]int64{},
 		gameServerStateLastChange: lruCache,
+		now: func() time.Time {
+			return time.Now()
+		},
 	}
 
 	c.logger = runtime.NewLoggerWithType(c)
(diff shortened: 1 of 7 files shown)


aLekSer:
This approach uses current subscription to `gsInformer`:
```
	gsInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
		UpdateFunc: c.recordGameServerStatusChanges,
	}, 0)
```
Need to perform more tests on this approach.
LRU cache was used instead of map because it could grow under certain circumstances indefinitely.

pkg/metrics/controller.go
```diff
@@ -262,9 +269,43 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+		if err != nil {
+			c.logger.Info(err.Error())
+		}
 	}
 }
 
+func calcDuration(newGs, oldGs *agonesv1.GameServer) error {
+	fleetName := newGs.Labels[agonesv1.FleetNameLabel]
+	diff := time.Now().Sub(newGs.ObjectMeta.CreationTimestamp.Local()).Seconds()
+	key := newGs.ObjectMeta.Name + "/" + string(newGs.Status.State)
```
akremsa, line 286: fmt.Sprintf("%s/%s", v1, v2)

pkg/metrics/controller.go
```diff
@@ -262,9 +269,43 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+		err := calcDuration(newGs, oldGs)
+		if err != nil {
+			c.logger.Info(err.Error())
+		}
 	}
 }
 
+func calcDuration(newGs, oldGs *agonesv1.GameServer) error {
+	fleetName := newGs.Labels[agonesv1.FleetNameLabel]
+	diff := time.Now().Sub(newGs.ObjectMeta.CreationTimestamp.Local()).Seconds()
```
akremsa, line 285: time.Now().UTC

pkg/metrics/controller_metrics.go
```diff
@@ -32,6 +32,7 @@ var (
 	gameServerTotalStats      = stats.Int64("gameservers/total", "The total of gameservers", "1")
 	nodesCountStats           = stats.Int64("nodes/count", "The count of nodes in the cluster", "1")
 	gsPerNodesCountStats      = stats.Int64("gameservers_node/count", "The count of gameservers per node in the cluster", "1")
+	gsStateDuration           = stats.Float64("gameservers_state/duration", "The duration of gameservers to be in a particular state", stats.UnitMilliseconds)
```
akremsa, line 35: gsStateDurationSec ?

pkg/metrics/controller_metrics.go
```diff
@@ -110,6 +111,13 @@ var (
 			Description: "The count of gameservers per node in the cluster",
 			Aggregation: view.Distribution(0.00001, 1.00001, 2.00001, 3.00001, 4.00001, 5.00001, 6.00001, 7.00001, 8.00001, 9.00001, 10.00001, 11.00001, 12.00001, 13.00001, 14.00001, 15.00001, 16.00001, 32.00001, 40.00001, 50.00001, 60.00001, 70.00001, 80.00001, 90.00001, 100.00001, 110.00001, 120.00001),
 		},
+		&view.View{
+			Name:        "gameserver_state_duration",
+			Measure:     gsStateDuration,
+			Description: "The time gameserver exist in the current state in seconds",
```
akremsa, line 117: Gameserver state duration\lasting (seconds)

pkg/metrics/controller_test.go
```diff
@@ -176,3 +178,22 @@ func TestControllerGameServersNodeState(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestCalcDuration(t *testing.T) {
+	lruCache, err := lru.New(16)
+	GameServerStateLastChange = lruCache
+	assert.NoError(t, err, "Not able to create a cache")
```
akremsa, line 185: Nit: unable to...

pkg/metrics/controller.go
```diff
@@ -80,6 +82,11 @@ func NewController(
 	fasInformer := fas.Informer()
 	node := kubeInformerFactory.Core().V1().Nodes()
 	nodeInformer := node.Informer()
+	lruCache, err := lru.New(1 << 24)
+	if err != nil {
+		logger.Error("Could not create LRU cache ", err)
```
akremsa, line 87: missing return

akremsa: WithError()...

pkg/metrics/controller.go
```diff
@@ -262,9 +269,43 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
 
+func calcDuration(newGs, oldGs *agonesv1.GameServer) error {
+	fleetName := newGs.Labels[agonesv1.FleetNameLabel]
+	diff := time.Now().Sub(newGs.ObjectMeta.CreationTimestamp.Local()).Seconds()
+	key := newGs.ObjectMeta.Name + "/" + string(newGs.Status.State)
+	oldKey := oldGs.ObjectMeta.Name + "/" + string(oldGs.Status.State)
+	if !GameServerStateLastChange.Contains(key) {
+		GameServerStateLastChange.Add(key, diff)
+		recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(newGs.Status.State)),
+			tag.Upsert(keyFleetName, fleetName)}, gsStateDuration.M(diff*1000.))
```
akremsa, line 291: Should it be "*1000"? Check what .Seconds() returns, seems no need to multiply diff here

pkg/metrics/controller.go
```diff
@@ -262,9 +269,43 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+	} else {
+		val, ok := GameServerStateLastChange.Get(key)
+		if !ok {
+			return errors.New("Could not find expected key")
+		}
+		duration := diff - val.(float64)
```
akremsa, line 297: Probably it is worth to validate duration after this action

aLekSer:
This is how new metric could look like in Grafana, need to add `by (type)` to get graph for each state, currently all are presented in one.
Screenshot 2020-04-14 at 17 07 40

Also need to skip some incoming timestamps from evicted keys. Could achieve this by switching to `{GameState, time}` structure for values and `namespace, name` for keys in the cache.

aLekSer:
There is an issue with Shutdown state, which seems to be inaccurate.
Two examples of charts for different state durations;
Ready State
Creating

markmandel:
Does this need review of some kind? I generally don't look at PR's in Draft unless specifically asked to, but saw this one seemed to have been sitting here for some time.

aLekSer:
Will update this and my other PRs today. I switched to another task related to Terraform yesterday.

aLekSer:
TestGameServerReserve E2E failed:
```
time="2020-04-29 21:30:32.971" level=info msg="GameServer Ready" gs=udp-server5mkt5
--- FAIL: TestGameServerReserve (162.30s)
    gameserver_test.go:632: 
        	Error Trace:	gameserver_test.go:632
        	Test:       	TestGameServerReserve
```

aLekSer:
I am going to add Grafana in a separate review.

aLekSer:
https://pkg.go.dev/k8s.io/client-go@v1.5.1/1.5/tools/cache?tab=doc#ResourceEventHandler
There could be a loose of some transient states, I assume that them are very small and we can skip some of them without loosing big picture.
```
 OnUpdate is called when an object is modified. Note that oldObj is the
    last known state of the object-- it is possible that several changes
    were combined together, so you can't use this to see every single
    change. OnUpdate is also called when a re-list happens, and it will
    get called even if nothing changed. This is useful for periodically
    evaluating or syncing something.
```

aLekSer:
Error in `test-go`:
```
--- FAIL: TestGameServerPod_WithSidecar_NoErrros (0.00s)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1500ff8]

goroutine 35 [running]:
testing.tRunner.func1(0xc0002ec100)
	/usr/local/go/src/testing/testing.go:874 +0x69f
panic(0x1616120, 0x2179750)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
agones.dev/agones/pkg/apis/agones/v1.(*GameServer).Pod(0xc000442000, 0xc00041bde8, 0x1, 0x1, 0x4778c5, 0x1ad7350, 0x64a81e)
	/go/src/agones.dev/agones/pkg/apis/agones/v1/gameserver.go:612 +0x448
agones.dev/agones/pkg/apis/agones/v1.TestGameServerPod_WithSidecar_NoErrros(0xc0002ec100)
	/go/src/agones.dev/agones/pkg/apis/agones/v1/gameserver_test.go:1263 +0x41c
testing.tRunner(0xc0002ec100, 0x17afde0)
	/usr/local/go/src/testing/testing.go:909 +0x19a
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:960 +0x652
FAIL	agones.dev/agones/pkg/apis/agones/v1	0.103s
```

markmandel:
@cyriltovena you good to review this? I defer this type of stuff to you :smile:

pkg/metrics/controller.go
```diff
@@ -258,11 +271,60 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+		if err != nil {
+			c.logger.Debug(err.Error())
+		} else {
+			recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(oldGs.Status.State)),
+				tag.Upsert(keyFleetName, fleetName)}, gsStateDurationSec.M(duration))
+		}
+	}
+}
+
+// calcDuration - calculate duration between state changes
```
cyriltovena, line 290:
```suggestion
// calcDuration - calculates the duration between state changes
```

pkg/metrics/controller.go
```diff
@@ -258,11 +271,60 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+// store current time from creationTimestamp for each update received
+// Assumptions: there is a possibility that one of the previous state change timestamps would be evicted,
+// this measure would be skipped
+func (c *Controller) calcDuration(oldGs, newGs *agonesv1.GameServer) (duration float64, err error) {
+	duration = 0.
```
cyriltovena, line 295:
```suggestion
	var duration float64
```

pkg/metrics/controller.go
```diff
@@ -258,11 +271,60 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+	defer c.stateLock.Unlock()
+	switch {
+	case newGs.Status.State == agonesv1.GameServerStateCreating || newGs.Status.State == agonesv1.GameServerStatePortAllocation:
+		duration = currentTime
+	case !c.gameServerStateLastChange.Contains(oldGSKey):
+		err = errors.New(fmt.Sprintf("Unable to calculate '%s' state duration of '%s' GameServer", oldGs.Status.State, oldGs.ObjectMeta.Name))
+	default:
+		val, ok := c.gameServerStateLastChange.Get(oldGSKey)
+		if !ok {
+			err = errors.New(fmt.Sprintf("Could not find expected key %s", oldGSKey))
```
cyriltovena, line 312: > Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.

```suggestion
			err = errors.New(fmt.Sprintf("could not find expected key %s", oldGSKey))
```

pkg/metrics/controller.go
```diff
@@ -258,11 +271,60 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
-			fleetName = "none"
+			fleetName = defaultFleetTag
 		}
 		recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(newGs.Status.State)),
 			tag.Upsert(keyFleetName, fleetName)}, gameServerTotalStats.M(1))
+
+		// Calculate the duration of the current state
+		duration, err := c.calcDuration(oldGs, newGs)
+		if err != nil {
+			c.logger.Debug(err.Error())
```
cyriltovena, line 282: I would think warning is better suited for this. I would also wrap the message to add more meat about the context.

pkg/metrics/controller.go
akremsa: I think this line is redundant at all because there is "duration" variable declared as a return parameter.

cyriltovena: good point !

pkg/metrics/controller.go
```diff
@@ -258,11 +271,60 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+}
+
+// calcDuration - calculate duration between state changes
+// store current time from creationTimestamp for each update received
+// Assumptions: there is a possibility that one of the previous state change timestamps would be evicted,
+// this measure would be skipped
+func (c *Controller) calcDuration(oldGs, newGs *agonesv1.GameServer) (duration float64, err error) {
+	duration = 0.
+	// currentTime - GameServer time from its start
+	currentTime := time.Now().UTC().Sub(newGs.ObjectMeta.CreationTimestamp.Local().UTC()).Seconds()
```
akremsa, line 297: Please specify a measure of time-related variables. It will make code much easier to read and to understand. _currentTimeSec_

pkg/metrics/controller_metrics.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 var (
+	distributionSeconds       = []float64{0, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2, 3}
```
cyriltovena, line 24: Not sure about those buckets. What are your expectations from a range of min to max duration for each state ?

I have a feeling this could go way beyond 3 seconds, since this is a histogram you're going to have an average and quantile very wrong.

Might even needs to be configurable by the users with a sane default. Let's start with a just a sane default for now. This is a game of precisions vs memory usage, I would have more than 20buckets for instance.

pkg/metrics/controller.go
```diff
@@ -258,11 +271,60 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+
+	newGSKey := fmt.Sprintf("%s/%s/%s", newGs.ObjectMeta.Namespace, newGs.ObjectMeta.Name, string(newGs.Status.State))
+	oldGSKey := fmt.Sprintf("%s/%s/%s", oldGs.ObjectMeta.Namespace, oldGs.ObjectMeta.Name, string(oldGs.Status.State))
```
akremsa, line 300: Can we omit extra string casting here?

pkg/metrics/controller_test.go
```diff
@@ -284,3 +285,36 @@ func TestControllerGameServersNodeState(t *testing.T) {
...
+	gs3.ObjectMeta.CreationTimestamp = gs1.ObjectMeta.CreationTimestamp
+	duration, err = c.calcDuration(gs2, gs3)
+	assert.NoError(t, err, "Unable to caculate duration of a particular state")
+	assert.True(t, duration > 0., "Time diff should be calculated properly")
+
+	// gs1 state should already be deleted, error should be generated
+	// emulation of evicted key for gs1
+	duration, err = c.calcDuration(gs1, gs3)
+	assert.Error(t, err, "We should receive an error, metric should not be measured")
+	assert.True(t, duration == 0., "Time diff should be calculated properly")
```
cyriltovena, line 313: You could inject time.Now in the Controller.

now func() time.Time so that you can fake now and have a better expectations than >= 0

pkg/metrics/controller_test.go
```diff
@@ -284,3 +285,36 @@ func TestControllerGameServersNodeState(t *testing.T) {
 		{labels: []string{"false"}, val: int64(2)},
 	})
 }
+
+func TestCalcDuration(t *testing.T) {
```
akremsa, line 289: Please, make this test to be a table-driven one and add a description for each test case like "what you test - what you expect". It will make this test much more readable, easy to understand, and easy to extend.
Example: https://github.com/googleforgames/agones/blob/master/pkg/apis/agones/v1/gameserver_test.go#L48

pkg/metrics/controller_metrics.go
aLekSer: You are right, buckets boundaries are unusual for state duration.

aLekSer:
Thank you for the review. Will add a documentation that this is an estimation of what State durations are. With a chance that we loose one extra small duration or one extra long, say > 1 day. But this should not change 99% too much. But this is a trade-off made not to introduce additional burden of event stream handling.

aLekSer:
Returning back to this PR, sorry for the delay.

aLekSer:
Tested new metrics with `RESERVE` call on `simple-udp`. The resulted 50% quantile of the metrics was 12 seconds, but originally Reserve was set for 10 seconds:
State Duration Metric

pkg/metrics/controller.go
```diff
@@ -258,11 +273,66 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+	newGSKey := fmt.Sprintf("%s/%s/%s/%s", newGs.ObjectMeta.Namespace, fleetName, newGs.ObjectMeta.Name, newGs.Status.State)
+	oldGSKey := fmt.Sprintf("%s/%s/%s/%s", oldGs.ObjectMeta.Namespace, fleetName, oldGs.ObjectMeta.Name, oldGs.Status.State)
+
+	c.stateLock.Lock()
+	defer c.stateLock.Unlock()
+	switch {
+	case newGs.Status.State == agonesv1.GameServerStateCreating || newGs.Status.State == agonesv1.GameServerStatePortAllocation:
+		duration = currentTime
+	case !c.gameServerStateLastChange.Contains(oldGSKey):
+		err = errors.New(fmt.Sprintf("unable to calculate '%s' state duration of '%s' GameServer", oldGs.Status.State, oldGs.ObjectMeta.Name))
```
akremsa, line 314: **fmt.Errorf()** can be used as a substitute for this statement

pkg/metrics/controller.go
```diff
@@ -258,11 +273,66 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+		return 0, err
+	default:
+		val, ok := c.gameServerStateLastChange.Get(oldGSKey)
+		if !ok {
+			err = errors.New(fmt.Sprintf("could not find expected key %s", oldGSKey))
```
akremsa, line 319: the same

pkg/metrics/controller.go
```diff
@@ -258,11 +273,66 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+	}
+
+	// Assuming that no State changes would occur after Shutdown
+	if newGs.Status.State != agonesv1.GameServerStateShutdown {
+		c.gameServerStateLastChange.Add(newGSKey, currentTime)
+		c.logger.Debugf("Adding new key %s, relative time: %f", newGSKey, currentTime)
+	}
+	if duration < 0. {
+		duration = 0
+		err = errors.New(fmt.Sprintf("negative duration for '%s' state of '%s' GameServer", oldGs.Status.State, oldGs.ObjectMeta.Name))
```
akremsa, line 333: and here

pkg/metrics/controller_test.go
```diff
@@ -284,3 +287,113 @@ func TestControllerGameServersNodeState(t *testing.T) {
...
+	gsName2 := "exampleGameServer2"
+	currentTime := creationTimestamp.Local()
+	// Add one second each time Duration is calculated
+	c.now = func() time.Time {
+		currentTime = currentTime.Add(1 * time.Second)
+		return currentTime
+	}
+	type result struct {
+		duration float64
+		err      error
```
akremsa, line 310: Why do you use error type here? You use _assert.EqualError_ below and take a string representation of an error.
It's not critical, just wondering.

pkg/metrics/controller_test.go
```diff
@@ -284,3 +287,113 @@ func TestControllerGameServersNodeState(t *testing.T) {
...
+			if tc.expected.err != nil {
+				assert.EqualError(t, err, tc.expected.err.Error(), "We should receive an error, metric should not be measured")
+			} else {
+				assert.NoError(t, err, "Unable to caculate duration of a particular state")
+			}
+			assert.Equal(t, tc.expected.duration, duration, "Time diff should be calculated properly")
+		})
+	}
+	assert.Len(t, c.gameServerStateLastChange.Keys(), 0, "We should not have any keys after the test")
+}
```
akremsa, line 399: Good job!

pkg/metrics/controller_test.go
aLekSer: To pass a `nil` and compare err with nil looks more natural.

markmandel:
@cyriltovena just a gentle a reminder that RC is next Tuesday, so if want this in the next release, would love it if you could see if this is good to go :smile:

aLekSer:
At least three times today `GnuTLS recv error (-54)`:
```
Step 10/21 : RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VER}/bin/linux/amd64/kubectl &&     chmod go+rx ./kubectl &&     mv ./kubectl /usr/local/bin/kubectl
 ---> Running in d824b8d25f3f
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

curl: (56) GnuTLS recv error (-54): Error in the pull function.
```
Possibly we can switch to wget.
https://github.com/curl/curl/issues/1707

markmandel:
That's needs updating as well, so we could do this as part of #1478

pkg/metrics/controller.go
cyriltovena: WDYT ?

pkg/metrics/controller.go
```diff
@@ -81,15 +86,25 @@ func NewController(
 	node := kubeInformerFactory.Core().V1().Nodes()
 	nodeInformer := node.Informer()
 
+	// GameServerStateLastChange Contains the time when the GameServer
+	// changed its state last time
+	// on delete remove GameServerName key
+	lruCache, err := lru.New(1 << 12)
```
markmandel, line 92: What does `1 << 12` mean? (also how did we come up with this number?)

pkg/metrics/controller.go
markmandel: I would agree, Warn() would be better, unless we are specifically trying to ignore any error here?

pkg/metrics/controller.go
```diff
@@ -258,11 +273,66 @@ func (c *Controller) recordGameServerStatusChanges(old, next interface{}) {
...
+			recordWithTags(context.Background(), []tag.Mutator{tag.Upsert(keyType, string(oldGs.Status.State)),
+				tag.Upsert(keyFleetName, fleetName)}, gsStateDurationSec.M(duration))
+		}
+	}
+}
+
+// calcDuration - calculate duration between state changes
+// store current time from creationTimestamp for each update received
+// Assumptions: there is a possibility that one of the previous state change timestamps would be evicted,
+// this measure would be skipped
```
markmandel, line 295: Should we write something about why we don't care if the item is skipped?

Is it because we're assuming that we'll get enough data across enough GameServers that the occasional miss won't happen?

site/content/en/docs/Guides/metrics.md
```diff
@@ -60,6 +60,7 @@ Follow the [Stackdriver Installation steps](#stackdriver-installation) to see yo
 | agones_fleet_autoscalers_limited                | The fleet autoscaler is capped (1)                                  | gauge     |
 | agones_gameservers_node_count                   | The distribution of gameservers per node                            | histogram |
 | agones_nodes_count                              | The count of nodes empty and with gameservers                       | gauge     |
+| agones_gameservers_state_duration  | The distribution of gameserver state duration in seconds. Note: this metric could have some missing samples by design. Do not use the `_total` counter the real value for state changes.     | histogram     |
```
markmandel, line 63: This should probably be wrapped in a feature shortcode, yeah?

aLekSer: Yes, will update this.

aLekSer:
Updating today.

pkg/metrics/controller.go
```diff
@@ -41,7 +43,19 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-const noneValue = "none"
+const (
+	noneValue = "none"
+
+	// GameServersStateCount is the size of LRU cache and should contain all gameservers state changes
```
markmandel, line 49: Looks good!

aLekSer:
Minor changes to comments.

aLekSer:
Example of what values would we get per fleet (`simple-udp` and `none`):
`agones_gameserver_state_duration_sum/agones_gameserver_state_duration_count`
<a href="https://ibb.co/gRSwm0p">Screenshot-2020-06-03-at-23-06-14</a><br /><a target='_blank' href='https://imgbb.com/'>Prometheus Metrics</a><br />

aLekSer:
@cyriltovena Cyril, I think I have applied all your comments. Can you please re-review?

markmandel:
Just checking in on this. See how it's doing?

aLekSer:
This version was finalised. @cyriltovena one more bump on this.

pkg/metrics/controller_metrics.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 var (
+	distributionSeconds       = []float64{0, 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384}
```
cyriltovena, line 24: nit: I would use a different name or not use a variable.

I think this boundaries only fits your metrics.

aLekSer: Thanks for the update I would rename it to `stateDurationSeconds`.

PR #1476, opened 2020-04-16
GameServer CRD with Updated Player Tracking

markmandel:
This updates the GameServer CRD based on the [updated design](https://github.com/googleforgames/agones/issues/1033#issuecomment-613184537) for Player Tracking.


pkg/apis/agones/v1/gameserver.go
```diff
@@ -241,8 +239,9 @@ type GameServerStatusPort struct {
 
 // PlayerStatus stores the current player capacity values
 type PlayerStatus struct {
-	Count    int64 `json:"count"`
-	Capacity int64 `json:"capacity"`
+	Count    int64    `json:"count"`
+	Capacity int64    `json:"capacity"`
+	IDs      []string `json:"IDs"`
```
roberthbailey, line 244: I'm wondering if we should have a type here, e.g. `type PlayerID string` and say this as `IDs []PlayerID` instead. Or maybe we only use type aliases when the values are an enum?

markmandel: I guess my question would be, what does a type alias here give us, when it's always going to be asking for a string user provided value to populate this value?

roberthbailey: Good question. I'm not sure it gives us much. The only thing I think might be valuable is that it would be an obvious place to add a comment about _why_ a player id is a string and maybe some suggestions / best practices (if they emerge) about what people might want to store as the player id value.

As a side note, are we going to enforce an upper limit on how long this slice can be?

markmandel: > Good question. I'm not sure it gives us much.

Thinking about how this will be exposed to the user -- it would be as a CRD, so it will fall back to a String, or via the gRPC api, in which case it will also end up being a string. So I'm not sure there is value here since it will get translated at the higher levels.

> As a side note, are we going to enforce an upper limit on how long this slice can be?

I wasn't -- I had concerns that might be premature optimisation (and might change performance based on cloud provider/k8s setup?) . Figured it might be better to see what happens long term. But if we have a good thought on what's a "good" number, maybe it's better to go the other way and implement a limit, and then see if people complain?? Not sure.

roberthbailey: Leaving it without a limit lets people use it for use cases we haven't thought of (a plus), but it could also lead to situations where it stores too much (arbitrary) data and leads to performance issues in the apiserver. I'm not sure what a reasonable upper bound is, and maybe if someone is using it to store tens of thousands of strings they can deal with the consequences?

markmandel: So far we have made an assumption that if you abuse something, then you should run your own load tests to determine if it's going to be performant or not.

For example: You could set 10,000 custom labels on a GameServer with the SDK, and try and change all of them every millisecond. That would likely be a bad idea, but we don't stop you.

markmandel: I would probably advocate leaving this as is (i.e. no limit). I'm philosophically more inclined to enable more flexibility for users, unless we specifically know there is a bottleneck/performance issue of some kind.

PR #1479, opened 2020-04-17
Monitoring: fix error on Stackdriver exporter

### LINKED ISSUE #1330: Metrics: Export to Stackdriver is not working

aLekSer:
Zero value cannot be used as the only value for view.Distribution().

Closes #1330


aLekSer:
Also found some useful examples. First uses new version of `client-go`, so we will need to update metrics with update to 1.15:
https://github.com/knative/pkg/blob/3c828cf99f67f8066e956490f634ebe90abd6892/metrics/reflector_test.go
Example of how distribution could be created:
https://github.com/knative/pkg/blob/4e57475bc87c1aba3e7fcc0207b593dc7186eb83/metrics/workqueue.go#L78

aLekSer:
On Grafana Workerqueue metrics, I have changed looks the same as before.
Screenshot 2020-04-22 at 23 04 39
However I got a strange issue on `GameServers` dashboard, that could be caused by Grafana update:
```
Dashboard init (failed)
Failed create dashboard model
undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

markmandel:
/assign @cyriltovena 

Can you review? I've no idea what any of this means :smile:

pkg/metrics/kubernetes_client.go
```diff
@@ -81,14 +82,14 @@ func init() {
...
-		Aggregation: view.Distribution(0, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2, 3),
+		Aggregation: view.Distribution(distributionSeconds...),
 	}))
 
 	runtime.Must(view.Register(&view.View{
 		Name:        "k8s_client_cache_list_items",
 		Measure:     cacheListItemCountStats,
 		Description: "Count of items in a list from the Kubernetes API.",
-		Aggregation: view.Distribution(0),
+		Aggregation: view.Distribution(0.00001),
```
cyriltovena, line 92: I wonder why past me didn't put any buckets here ?

I think we should have 0,10,50,100,150,250,300.

pkg/metrics/kubernetes_client.go
```diff
@@ -109,14 +110,14 @@ func init() {
...
-		Aggregation: view.Distribution(0),
+		Aggregation: view.Distribution(distributionSeconds...),
 	}))
 
 	runtime.Must(view.Register(&view.View{
 		Name:        "k8s_client_cache_watch_events",
 		Measure:     cacheItemsInWatchesCountStats,
 		Description: "Number of items in watches on the Kubernetes API.",
-		Aggregation: view.Distribution(0),
+		Aggregation: view.Distribution(0.00001),
```
cyriltovena, line 120: Same here. you should use the same kind of buckets. It work for our dashboard to not have bucket cause I do average mostly but if you want 99th % you can't without proper buckets.

cyriltovena:
Looks good just add bucket I forgot when I created this.

PR #1481, opened 2020-04-18
Changes in proto for Player Tracking design update

markmandel:
**What type of PR is this?**
/kind feature

This includes changes to make PlayerID consistent in casing.

Work on #1033

Code generation will come in the next PR to make things easier to
review.

### DIFF
--- a/proto/sdk/alpha/alpha.proto
+++ b/proto/sdk/alpha/alpha.proto
@@ -21,42 +21,94 @@ import "google/api/annotations.proto";
 
 // SDK service to be used in the GameServer SDK to the Pod Sidecar.
 service SDK {
-    // Call when a player has connected.
-    rpc PlayerConnect (PlayerId) returns (Empty) {
+    // PlayerConnect increases the SDK’s stored player count by one, and appends this playerID to GameServer.Status.Players.IDs.
+    //
+    // GameServer.Status.Players.Count and GameServer.Status.Players.IDs are then set to update the player count and id list a second from now,
+    // unless there is already an update pending, in which case the update joins that batch operation.
+    //
+    // PlayerConnect returns true and adds the playerID to the list of playerIDs if this playerID was not already in the
+    // list of connected playerIDs.
+    //
+    // If the playerID exists within the list of connected playerIDs, PlayerConnect will return false, and the list of
+    // connected playerIDs will be left unchanged.
+    //
+    // An error will be returned if the playerID was not already in the list of connected playerIDs but the player capacity for
+    // the server has been reached. The playerID will not be added to the list of playerIDs.
+    //
+    // Warning: Do not use this method if you are manually managing GameServer.Status.Players.IDs and GameServer.Status.Players.Count
+    // through the Kubernetes API, as indeterminate results will occur.
+    rpc PlayerConnect (PlayerID) returns (Bool) {
         option (google.api.http) = {
             post: "/alpha/player/connect"
             body: "*"
         };
     }
 
-    // Call when a player has disconnected.
-    rpc PlayerDisconnect (PlayerId) returns (Empty) {
+    // Decreases the SDK’s stored player count by one, and removes the playerID from GameServer.Status.Players.IDs.
+    //
+    // GameServer.Status.Players.Count and GameServer.Status.Players.IDs are then set to update the player count and id list a second from now,
+    // unless there is already an update pending, in which case the update joins that batch operation.
+    //
+    // PlayerDisconnect will return true and remove the supplied playerID from the list of connected playerIDs if the
+    // playerID value exists within the list.
+    //
+    // If the playerID was not in the list of connected playerIDs, the call will return false, and the connected playerID list
+    // will be left unchanged.
+    //
+    // Warning: Do not use this method if you are manually managing GameServer.status.players.IDs and GameServer.status.players.Count
+    // through the Kubernetes API, as indeterminate results will occur.
+    rpc PlayerDisconnect (PlayerID) returns (Bool) {
         option (google.api.http) = {
             post: "/alpha/player/disconnect"
             body: "*"
         };
     }
 
-    // Change the player capacity to a new value.
+    // Update the GameServer.Status.Players.Capacity value with a new capacity.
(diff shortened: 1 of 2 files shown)


proto/sdk/alpha/alpha.proto
```diff
@@ -21,44 +21,76 @@ import "google/api/annotations.proto";
...
+    // Update the status.capacity value with a new capacity.
     rpc SetPlayerCapacity (Count) returns (Empty) {
         option (google.api.http) = {
             post: "/alpha/player/capacity"
             body: "*"
         };
     }
 
-    // Get the last player capacity that was set through the SDK.
+    // Retrieves the current capacity. This is always accurate, even if the value hasn’t been updated to the GameServer status yet.
```
roberthbailey, line 66: Nit: The comment about it possibly being set outside of the SDK is true for the following methods as well. And if it's set outside the SDK is the result still always accurate?

roberthbailey: Actually, thinking about this more, what happens if the player count or ids are set from outside the game server? Will the game server overwrite the changes on the next status update? Will the ids be merged (possibly making the count no longer match the length of ids and/or making the list have duplicate values)?

markmandel: > Nit: The comment about it possibly being set outside of the SDK is true for the following methods as well. And if it's set outside the SDK is the result still always accurate?

Ah, that is a very good point. It it's set outside the SDK, it will no longer be accurate. In which case, yes -- one should use sdk.GetGameServer / WatchGameServer. I'll add a note here too.

> Actually, thinking about this more, what happens if the player count or ids are set from outside the game server? Will the game server overwrite the changes on the next status update? Will the ids be merged (possibly making the count no longer match the length of ids and/or making the list have duplicate values)?

Oooh, that is also an excellent point. I've been working on this, and currently it would overwrite -- but since we're have to retrieve the GameSever before applying the SDK updates, we can definitely merge the two. it's a bit more computationally, but someone will definitely try and do this, so I think it's worth doing it.

markmandel: If someone changes the count manually, and then uses the SDK to update the connected/disconnected players, then the count will update -- which I think is fine. That seems like an expected scenario. If you self manage these values, then it's on you to manage them yourself.

markmandel: We can probably keep things close to accurate as well on retrieval by getting the latest GameServer from the Informer Cache and merging any additions/deletions pending. It will still have some race conditions if values are set from outside, but it's as close as we can get through retrieval.

markmandel: Okay, I'm going to stop with the stream of consciousness 😃 I'm now leaning towards some strong documentation that you only get to choose one option: SDK or Manual management. (After digging into some code options) -- at least until such time that someone comes along with a use case that requires this additional complexity.

proto/sdk/alpha/alpha.proto
```diff
@@ -21,44 +21,76 @@ import "google/api/annotations.proto";
 
 // SDK service to be used in the GameServer SDK to the Pod Sidecar.
 service SDK {
-    // Call when a player has connected.
-    rpc PlayerConnect (PlayerId) returns (Empty) {
+    // PlayerConnect increases the SDK’s stored player count by one, and appends this playerID to status.players.id.
+    // status.players.count and status.players.ids is then set to update to the player count and id list a second from now, unless there is already an update pending, in which case the update joins that batch operation.
```
aLekSer, line 25: Nit: this comment is extra-long, should be split.

proto/sdk/alpha/alpha.proto
```diff
@@ -21,44 +21,76 @@ import "google/api/annotations.proto";
...
             post: "/alpha/player/connect"
             body: "*"
         };
     }
 
-    // Call when a player has disconnected.
-    rpc PlayerDisconnect (PlayerId) returns (Empty) {
+    // Decreases the SDK’s stored player count by one, and removes the playerID from status.players.id
+    //
+    // status.players.count and status.players.ids is then set to update to the player count and id list a second from now, unless there is already an update pending, in which case the update joins that batch operation.
```
aLekSer, line 44: Same as above.

proto/sdk/alpha/alpha.proto
```diff
@@ -21,44 +21,76 @@ import "google/api/annotations.proto";
 
 // SDK service to be used in the GameServer SDK to the Pod Sidecar.
 service SDK {
-    // Call when a player has connected.
-    rpc PlayerConnect (PlayerId) returns (Empty) {
+    // PlayerConnect increases the SDK’s stored player count by one, and appends this playerID to status.players.id.
```
aLekSer, line 24:
```suggestion
    // PlayerConnect increases the SDK’s stored player count by one, and appends this playerID to status.players.ids.
```
As per design v2:
```
status.players.ids - The set of connected player ids at this given moment.
```

proto/sdk/alpha/alpha.proto
```diff
@@ -21,44 +21,76 @@ import "google/api/annotations.proto";
...
+    // status.players.count and status.players.ids is then set to update to the player count and id list a second from now, unless there is already an update pending, in which case the update joins that batch operation.
+    //
+    // PlayerConnect returns true and adds the playerID to the list of playerIDs if the playerIDs was not already in the
```
aLekSer, line 27:
```suggestion
    // PlayerConnect returns true and adds the playerID to the list of playerIDs if this playerID was not already in the
```

markmandel:
Pushed up a stack of updated documentation - mostly around when one should use the SDK commands or not. PTAL.

proto/sdk/alpha/alpha.proto
roberthbailey: I agree that trying to do merges seems like a path that will lead to lots of surprises and edge cases. 

+1 to documenting clearly that it's an antipattern.

markmandel:
> lgtm, but please address @aLekSer's comments before merging.

I _think_ that I have. @aLekSer would you mind confirming in case I missed something?

PR #1482, opened 2020-04-20
Terraform: update GKE cluster version, use locals and `lookup` to set default values

aLekSer:
Add new variable to select Kubernetes major and minor version.
`google_container_engine_versions` data source added, which allows to select proper node and master versions which applicable for every zone in the region if it is used with a regional cluster.

**What type of PR is this?**
 /kind feature

For #1478 .

It would allow proper version selection by specifying Kubernetes version prefix and cluster location. I.e. for regional clusters, `us-west2` could be used, for zonal clusters `us-west2-c`

Useful documents:
https://www.terraform.io/docs/providers/google/d/google_container_engine_versions.html#latest_master_version
https://www.terraform.io/docs/providers/google/r/container_cluster.html#min_master_version
https://www.terraform.io/docs/providers/google/r/container_cluster.html#node_version

### DIFF
--- a/build/terraform/gke/module.tf
+++ b/build/terraform/gke/module.tf
@@ -97,17 +97,21 @@ variable "feature_gates" {
   default = ""
 }
 
+variable "kubernetes_version" {
+  default = "1.15"
+}
 
 module "gke_cluster" {
   source = "../../../install/terraform/modules/gke"
 
   cluster = {
-    "name"             = var.name
-    "zone"             = var.zone
-    "machineType"      = var.machine_type
-    "initialNodeCount" = var.node_count
-    "project"          = var.project
-    "network"          = var.network
+    "name"              = var.name
+    "zone"              = var.zone
+    "machineType"       = var.machine_type
+    "initialNodeCount"  = var.node_count
+    "project"           = var.project
+    "network"           = var.network
+    "kubernetesVersion" = var.kubernetes_version
   }
 }
 
--- a/examples/terraform-submodules/gke/module.tf
+++ b/examples/terraform-submodules/gke/module.tf
@@ -46,7 +46,6 @@ variable "machine_type" {
 variable "node_count" {
   default = "4"
 }
-
 variable "zone" {
   default     = "us-west1-c"
   description = "The GCP zone to create the cluster in"
@@ -65,6 +64,10 @@ variable "feature_gates" {
   default = ""
 }
 
+variable "kubernetes_version" {
+  default = "1.15"
+}
+
 module "gke_cluster" {
   // ***************************************************************************************************
   // Update ?ref= to the agones release you are installing. For example, ?ref=release-1.3.0 corresponds
@@ -73,12 +76,13 @@ module "gke_cluster" {
   source = "git::https://github.com/googleforgames/agones.git//install/terraform/modules/gke/?ref=master"
 
   cluster = {
-    "name"             = var.name
-    "zone"             = var.zone
-    "machineType"      = var.machine_type
-    "initialNodeCount" = var.node_count
-    "project"          = var.project
-    "network"          = var.network
+    "name"              = var.name
+    "zone"              = var.zone
+    "machineType"       = var.machine_type
+    "initialNodeCount"  = var.node_count
+    "project"           = var.project
+    "network"           = var.network
+    "kubernetesVersion" = var.kubernetes_version
   }
 }
 
--- a/examples/terraform-submodules/gke/module.tf
+++ b/examples/terraform-submodules/gke/module.tf
@@ -64,10 +64,6 @@ variable "feature_gates" {
   default = ""
 }
 
-variable "kubernetes_version" {
-  default = "1.15"
-}
-
 module "gke_cluster" {
   // ***************************************************************************************************
   // Update ?ref= to the agones release you are installing. For example, ?ref=release-1.3.0 corresponds
@@ -76,13 +72,12 @@ module "gke_cluster" {
   source = "git::https://github.com/googleforgames/agones.git//install/terraform/modules/gke/?ref=master"
 
   cluster = {
-    "name"              = var.name
-    "zone"              = var.zone
-    "machineType"       = var.machine_type
-    "initialNodeCount"  = var.node_count
-    "project"           = var.project
-    "network"           = var.network
-    "kubernetesVersion" = var.kubernetes_version
+    "name"             = var.name
+    "zone"             = var.zone
+    "machineType"      = var.machine_type
+    "initialNodeCount" = var.node_count
+    "project"          = var.project
+    "network"          = var.network
   }
 }
 
--- a/install/terraform/modules/gke/cluster.tf
+++ b/install/terraform/modules/gke/cluster.tf
@@ -31,12 +31,6 @@ locals {
   kubernetesVersion = lookup(var.cluster, "kubernetesVersion", "1.15")
 }
 
-data "google_container_engine_versions" "kubernetes" {
-  project        = local.project
-  location       = local.zone
-  version_prefix = local.kubernetesVersion
-}
-
 # echo command used for debugging purpose
 # Run `terraform taint null_resource.test-setting-variables` before second execution
 resource "null_resource" "test-setting-variables" {
@@ -56,9 +50,7 @@ resource "google_container_cluster" "primary" {
   project  = local.project
   network  = local.network
 
-  min_master_version = data.google_container_engine_versions.kubernetes.latest_master_version
-
-  node_version = data.google_container_engine_versions.kubernetes.latest_node_version
+  min_master_version = local.kubernetesVersion
 
   node_pool {
     name       = "default"
--- a/install/terraform/modules/gke/cluster.tf
+++ b/install/terraform/modules/gke/cluster.tf
@@ -19,10 +19,22 @@ terraform {
 
 data "google_client_config" "default" {}
 
+# A list of all parameters used in interpolation var.cluster
+# Set values to default if not key was not set in original map
+locals {
+  project           = lookup(var.cluster, "project", "agones")
+  zone              = lookup(var.cluster, "zone", "us-west1-c")
+  name              = lookup(var.cluster, "name", "test-cluster")
+  machineType       = lookup(var.cluster, "machineType", "n1-standard-4")
+  initialNodeCount  = lookup(var.cluster, "initialNodeCount", "4")
+  network           = lookup(var.cluster, "network", "default")
+  kubernetesVersion = lookup(var.cluster, "kubernetesVersion", "1.15")
+}
+
 data "google_container_engine_versions" "kubernetes" {
-  project        = var.cluster["project"]
-  location       = var.cluster["zone"]
-  version_prefix = var.cluster["kubernetesVersion"]
+  project        = local.project
+  location       = local.zone
+  version_prefix = local.kubernetesVersion
 }
 
 # echo command used for debugging purpose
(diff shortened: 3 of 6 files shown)


aLekSer:
An error in C# SDK build step:
```
includes/sdk.mk:107: recipe for target 'build-build-sdk-image' failed
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
E: Failed to fetch https://packages.microsoft.com/ubuntu/18.04/prod/pool/main/d/dotnet-sdk-2.2/dotnet-sdk-2.2.402-x64.deb  GnuTLS recv error (-54): Error in the pull function.
Fetched 44.1 MB in 3s (12.2 MB/s)
  GnuTLS recv error (-54): Error in the pull function.
Err:9 https://packages.microsoft.com/ubuntu/18.04/prod bionic/main amd64 dotnet-sdk-2.2 amd64 2.2.402-1
```

install/terraform/modules/gke/variables.tf
```diff
@@ -25,11 +25,12 @@ variable "cluster" {
...
-    "initialNodeCount" = "4"
-    "project"          = "agones"
-    "network"          = "default"
+    "zone"              = "us-west1-c"
+    "name"              = "test-cluster"
+    "machineType"       = "n1-standard-4"
+    "initialNodeCount"  = "4"
+    "project"           = "agones"
+    "network"           = "default"
+    "kubernetesVersion" = "1.15"
```
markmandel, line 34: Question: Since this is a map, doesn't that mean users have to set all these values, they can't choose individual ones?

In which case the default value isn't actually useful.

It may be worth doing a slow transition away from a map here? @chrisst I think you had thoughts here.

examples/terraform-submodules/gke/module.tf
```diff
@@ -65,6 +64,10 @@ variable "feature_gates" {
   default = ""
 }
 
+variable "kubernetes_version" {
```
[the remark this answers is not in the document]
roberthbailey: Having an override is useful (and we may end up wanting to use it for skew testing in the future) but the docs are pretty clear that we only test/support one version. So we should make sure that any places where we expose this to users have the same warning as we have on https://agones.dev/site/docs/installation/.

site/content/en/docs/Installation/Terraform/gke.md
```diff
@@ -92,6 +92,7 @@ Configurable parameters:
 - network - the name of the VPC network you want your cluster and firewall rules to be connected to (default is "default")
 - log_level - possible values: Fatal, Error, Warn, Info, Debug (default is "info")
 - feature_gates - a list of alpha and beta version features to enable. For example, "PlayerTracking=true&ContainerPortAllocation=true"
+- kubernetes_version - a Kubernetes version of the GKE cluster
```
roberthbailey, line 95: This is where we need a big warning if we add this as a variable.

build/terraform/gke/module.tf
```diff
@@ -97,17 +97,21 @@ variable "feature_gates" {
   default = ""
 }
 
+variable "kubernetes_version" {
```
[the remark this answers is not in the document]
aLekSer: Variable was used here so that we can make changes to this file more rare. And imagine the situation when between Agones releases Kubernetes version would be deprecated but with Terraform we are cloning the Agones repo (as in docs) only at release points. Another thing is if we want to compare Terraform configs of Agones at version `1.6.0` and `1.7.0` to perform regression testing for example, then we can select the same Kubernetes version to compare at the same environment.

install/terraform/modules/gke/variables.tf
chrisst: The issue is that maps get blindly replaced so if you don't provide `network` but override other variables then the line `var.cluster["network"]` will fail as there is no network supplied.

The simplest change is to flatten the object to individual fields but that would be a breaking change for consumers. You can convert to using locals internally and set the defaults in those based on the content of the `cluster` variable but it will require a bit of magic internally. https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/ has some good examples of setting locals based on input variables and merging in defaults.

aLekSer: @chrisst Thank you for an explanation on how we can fix this replacement in maps.

examples/terraform-submodules/gke/module.tf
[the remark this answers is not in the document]
markmandel: I agree having an override is useful, I feel like we shouldn't be showing it in the example, the example should be best practice - in which case, the best practice will be to default to the kubernetes version that is applied as the default in the module.

WDYT?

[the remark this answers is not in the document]
aLekSer: Got it, I will not use this param in examples.

site/content/en/docs/Installation/Terraform/gke.md
aLekSer: Will remove this, we don't need this for examples configuration.

install/terraform/modules/gke/variables.tf
aLekSer: @chrisst Applied your comment, used lookup with locals.

aLekSer:
Flaky SDK test in C# `make tests` step:
```
/usr/share/dotnet/sdk/2.2.402/NuGet.targets(123,5): error : The file '/go/src/agones.dev/agones/sdks/csharp/sdk/obj/project.assets.json' already exists. [/go/src/agones.dev/agones/sdks/csharp/test/csharp-sdk-test.csproj]
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/transport/timeout_encoding.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/transport/transport.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/transport/transport_op_string.o
includes/sdk.mk:89: recipe for target 'run-sdk-command' failed
make[1]: *** [run-sdk-command] Error 1
includes/sdk.mk:85: recipe for target 'run-sdk-command-csharp' failed
make: *** [run-sdk-command-csharp] Error 2
```

aLekSer:
@markmandel This Pull Request is ready for review. Note that there is an additional requirement to use var `terraform destroy -var project=<GCP_ProjectID>`, without it we got the following:
```
google_container_engine_versions.kubernetes: Refreshing state...                                                                                                                   
                                                                                                                                                                                                           
Error: project: required field is not set                                                                                                                                                                  
                                                                                                                                                                                                           
  on ../../../install/terraform/modules/gke/cluster.tf line 34, in data "google_container_engine_versions" "kubernetes":                                                                                   
  34: data "google_container_engine_versions" "kubernetes" {   
```
Optionally we can add next workaround in this place:
```
data "google_container_engine_versions" "kubernetes" {
  # if there is no project specified, for example in destroy
  project        = local.project == "" ? "agones" : local.project 
```

install/terraform/modules/gke/variables.tf
markmandel: @chrisst - mind taking a look, make sure this is lining up with your proposed solution?

chrisst: Looks much more flexible! The `defaults` section also now just serves mostly as documentation unless a user completely omits the variable. Since the locals section serves as the source of truth in a partially defined block you'll want to keep these two in sync.

aLekSer:
Well, it seems that I could also test all GKE Terraform related changes using this https://github.com/googleforgames/cloud-game-servers-examples/tree/master/terraform-4-cluster configs for backward compatibility.

install/terraform/modules/gke/cluster.tf
```diff
@@ -19,37 +19,57 @@ terraform {
...
-  name     = var.cluster["name"]
-  location = var.cluster["zone"]
-  project  = var.cluster["project"]
-  network  = var.cluster["network"]
+  name     = local.name
+  location = local.zone
+  project  = local.project
+  network  = local.network
+
+  min_master_version = data.google_container_engine_versions.kubernetes.latest_master_version
```
markmandel, line 59: So I'm looking this over, and I'm wondering why we are doing this? Why not lean on the Google Cloud's defaulting of the minor version of the major version for Kubernetes?

It might be better to stick to what is the current minimum version default, rather than what is latest, since that will be the most stable. If people want the absolute latest, then they can opt in themselves.

@roberthbailey do you have thoughts on this?

install/terraform/modules/gke/variables.tf
markmandel: Awesome! This makes me happy.

install/terraform/modules/gke/cluster.tf
aLekSer: In this design user could use a specific version by setting `var.kubernetesVersion=1.15.2`, for instance.
There are also two other parameters in this data source which could be used to set another version:
https://www.terraform.io/docs/providers/google/d/google_container_engine_versions.html#default_cluster_version
and `valid_master_versions`. We can use smallest out of the list.

aLekSer: Similar issue I found, you can see all parameters and list of versions returned:
https://github.com/hashicorp/terraform/issues/20761

markmandel: My point is, if I provide "1.15", GKE is already going to give us the default version for 1.15.x -- why are we being opinionated and forcing it to the latest 1.15.x version, when it may not be the default.

Do we have evidence that that is the best version? Most supported? most stable? Seems like the safer bet to stick with whatever GKE decides is the default, since they do the most testing.

Unless I'm missing what is happening here?

aLekSer: Well the reason behind this is that `default_cluster_version` is `1.14.10-gke.27` and does not correlate to prefix which was selected:
https://gist.github.com/aLekSer/a7ac5c7728c848f3329de4caacc7c712

aLekSer: For valid node versions for `us-west1-c` there exist next versions:
```
gcloud container get-server-config --zone us-west1-c
validNodeVersions:                                                                                                                                                                                         
- 1.16.8-gke.12
- 1.16.8-gke.9                                                                                                                                                                                             
- 1.15.11-gke.12
...
```
However `prefix` in `data.google_container_engine_versions.kubernetes.latest_master_version` would select only one (latest) following the prefix rule and which would be enabled for each zone in a region (if location is a region). (Currently could not see a difference in versions which are enabled on different regions). Which would be `1.15.11-gke.12`. 
I can roll back using this new `data` source here and move it into a separate PR which might be easier to review later.

aLekSer: Also an elegant on how we can achieve what was before and what introduced by `data.google_container_engine_versions`:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/d4e7dc6a585770474ea4cdee3452cf98b404c6e2/modules/beta-private-cluster/main.tf#L41
```
 master_version_regional = var.kubernetes_version != "latest" ? var.kubernetes_version : data.google_container_engine_versions.region.latest_master_version
```

aLekSer: Rolled back this change.

aLekSer:
Updating PR, removed data source. Agree with you it was too many changes for one PR. Now this PR looks much more safe.

aLekSer:
By the way current version without data source lead to the same version as with it - `1.15.11-gke.12` .

install/terraform/modules/gke/cluster.tf
```diff
@@ -19,37 +19,49 @@ terraform {
 
 data "google_client_config" "default" {}
 
+# A list of all parameters used in interpolation var.cluster
+# Set values to default if not key was not set in original map
+locals {
```
markmandel, line 24: @chrisst you are our resident TF expert. This look good to you? If so, I'm happy to approve this PR :+1:

aLekSer: Note that I have checked that removing `kubernetesVersion` from the `clluster` map
does not affect the result, as it was before.
https://github.com/aLekSer/agones/blob/terraform%2Fupdate-gke-version/build/main.tf#L26-L34

chrisst: LGTM!

PR #1485, opened 2020-04-22
Improved gameserver unit tests

akremsa:
**What type of PR is this?**
> /kind cleanup

Added missing unit tests, refactored some old ones. During working on tests I've found a missing check in Validate method, which is related to #1396

Before:
Screenshot 2020-04-22 at 15 46 07
Screenshot 2020-04-22 at 15 45 39

After:
Screenshot 2020-04-22 at 15 47 12
Screenshot 2020-04-22 at 15 47 37

### DIFF
--- a/pkg/apis/agones/v1/common.go
+++ b/pkg/apis/agones/v1/common.go
@@ -28,7 +28,7 @@ import (
 // Block of const Error messages
 const (
 	ErrContainerRequired        = "Container is required when using multiple containers in the pod template"
-	ErrHostPortDynamic          = "HostPort cannot be specified with a Dynamic PortPolicy"
+	ErrHostPortDynamic          = "HostPort cannot be specified with a Dynamic or Passthrough PortPolicy"
 	ErrPortPolicyStatic         = "PortPolicy must be Static"
 	ErrContainerPortRequired    = "ContainerPort must be defined for Dynamic and Static PortPolicies"
 	ErrContainerPortPassthrough = "ContainerPort cannot be specified with Passthrough PortPolicy"
--- a/pkg/apis/agones/v1/common.go
+++ b/pkg/apis/agones/v1/common.go
@@ -28,7 +28,7 @@ import (
 // Block of const Error messages
 const (
 	ErrContainerRequired        = "Container is required when using multiple containers in the pod template"
-	ErrHostPortDynamic          = "HostPort cannot be specified with a Dynamic or Passthrough PortPolicy"
+	ErrHostPort                 = "HostPort cannot be specified with a Dynamic or Passthrough PortPolicy"
 	ErrPortPolicyStatic         = "PortPolicy must be Static"
 	ErrContainerPortRequired    = "ContainerPort must be defined for Dynamic and Static PortPolicies"
 	ErrContainerPortPassthrough = "ContainerPort cannot be specified with Passthrough PortPolicy"
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -360,8 +360,20 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 			causes = append(causes, metav1.StatusCause{
 				Type:    metav1.CauseTypeFieldValueNotSupported,
 				Field:   "players",
-				Message: fmt.Sprintf("Value cannot be set unless feature flag %s is enabled,", runtime.FeaturePlayerTracking),
+				Message: fmt.Sprintf("Value cannot be set unless feature flag %s is enabled", runtime.FeaturePlayerTracking),
 			})
+		}
+	}
+
+	if !runtime.FeatureEnabled(runtime.FeatureContainerPortAllocation) {
+		for _, p := range gss.Ports {
+			if p.Container != nil {
+				causes = append(causes, metav1.StatusCause{
+					Type:    metav1.CauseTypeFieldValueNotSupported,
+					Field:   fmt.Sprintf("%s.container", p.Name),
+					Message: fmt.Sprintf("Value cannot be set unless feature flag %s is enabled", runtime.FeatureContainerPortAllocation),
+				})
+			}
 		}
 	}
 
@@ -439,7 +451,7 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 				})
 			}
 
-			if p.Container != nil && gss.Container != "" {
+			if p.Container != nil && gss.Container != "" && runtime.FeatureEnabled(runtime.FeatureContainerPortAllocation) {
 				_, _, err := gss.FindContainer(*p.Container)
 				if err != nil {
 					causes = append(causes, metav1.StatusCause{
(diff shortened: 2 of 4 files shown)


pkg/apis/agones/v1/gameserver.go
```diff
@@ -572,22 +584,13 @@ func (gs *GameServer) FindGameServerContainer() (int, corev1.Container, error) {
 // ApplyToPodContainer applies func(v1.Container) to the specified container in the pod.
 // Returns an error if the container is not found.
 func (gs *GameServer) ApplyToPodContainer(pod *corev1.Pod, containerName string, f func(corev1.Container) corev1.Container) error {
-	var container corev1.Container
```
akremsa, line 575: Just decided to make this code a bit easier. In the previous implementation, containerIndex was the **last** match of a containereNames. But in general, having 2 containers with the same name is not valid, so it's ok to return the first occurrence.

aLekSer: There is a `Pod` validation which prohibit having two containers in one `Pod` with the same name.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -360,11 +360,23 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 			causes = append(causes, metav1.StatusCause{
 				Type:    metav1.CauseTypeFieldValueNotSupported,
 				Field:   "players",
-				Message: fmt.Sprintf("Value cannot be set unless feature flag %s is enabled,", runtime.FeaturePlayerTracking),
+				Message: fmt.Sprintf("Value cannot be set unless feature flag %s is enabled", runtime.FeaturePlayerTracking),
 			})
 		}
 	}
 
+	if !runtime.FeatureEnabled(runtime.FeatureContainerPortAllocation) {
```
aLekSer, line 368: That's a great way to separate this Feature logic.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -439,7 +451,7 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 				})
 			}
 
-			if p.Container != nil && gss.Container != "" {
+			if p.Container != nil && gss.Container != "" && runtime.FeatureEnabled(runtime.FeatureContainerPortAllocation) {
```
aLekSer, line 454: 👍

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -505,8 +577,365 @@ func TestGameServerValidate(t *testing.T) {
...
+	}
+	assert.Contains(t, fields, "container")
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
+
+	t.Run("validation of "+string(runtime.FeatureContainerPortAllocation), func(t *testing.T) {
+		err := runtime.ParseFeatures(string(runtime.FeatureContainerPortAllocation) + "=true")
+		assert.NoError(t, err)
+
+		// // container was not found
```
aLekSer, line 889: Nit: double comment sign.
```suggestion
		// container was not found
```

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -358,17 +427,22 @@ func TestGameServerApplyDefaults(t *testing.T) {
 }
 
 func TestGameServerValidate(t *testing.T) {
-	gs := GameServer{
+	t.Parallel()
+	var fields []string
```
markmandel, line 431: Should this be turned into a table based test, since it's grown so large?

akremsa: I thought about it... but wasn't sure if it is worth spending extra time on that. But once you mentioned it - let's do that.

pkg/apis/agones/v1/common.go
```diff
@@ -28,7 +28,7 @@ import (
 // Block of const Error messages
 const (
 	ErrContainerRequired        = "Container is required when using multiple containers in the pod template"
-	ErrHostPortDynamic          = "HostPort cannot be specified with a Dynamic PortPolicy"
+	ErrHostPortDynamic          = "HostPort cannot be specified with a Dynamic or Passthrough PortPolicy"
```
akremsa, line 31: There is a check for both policies so I decided to update this description and cover it in tests
https://github.com/googleforgames/agones/pull/1485/files#diff-5d7500992f221aa57cd7cee64628157bR446

akremsa:
@markmandel I've refactored Validate method and extracted tests which test features functionality in a separate function. Please check.

pkg/apis/agones/v1/common.go
markmandel: That's a good call :+1: wondering if we should rename this variable. Maybe `ErrHostPort` ? WDYT?

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -357,186 +426,759 @@ func TestGameServerApplyDefaults(t *testing.T) {
...
-		}
-		gs.ApplyDefaults()
-		causes, ok = gs.Validate()
-		assert.True(t, ok)
-		assert.Empty(t, causes)
-	})
 }
 
-func TestGameServerPod(t *testing.T) {
+func TestGameServerPod_NoErrors(t *testing.T) {
```
markmandel, line 1180: Nit: 
```suggestion
func TestGameServerPodNoErrors(t *testing.T) {
```

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -357,186 +426,759 @@ func TestGameServerApplyDefaults(t *testing.T) {
...
+}
 
-	gs = GameServer{
-		Spec: GameServerSpec{
-			Container: "my_image",
-			Template: corev1.PodTemplateSpec{
-				Spec: corev1.PodSpec{
-					Containers: []corev1.Container{
-						{Name: "my_image", Image: "foo/my_image"},
+func TestGameServerValidate_Features(t *testing.T) {
```
markmandel, line 1039: Nit: 
```suggestion
func TestGameServerValidateFeatures(t *testing.T) {
```

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -794,34 +1499,66 @@ func TestGameServerIsBeforeReady(t *testing.T) {
...
-			Container: name,
-			Template: corev1.PodTemplateSpec{
-				Spec: corev1.PodSpec{
-					Containers: []corev1.Container{
-						{Name: name, Image: "foo/mycontainer"},
-						{Name: "notmycontainer", Image: "foo/notmycontainer"},
+	var testCases = []struct {
+		description string
+		gs          *GameServer
+		errExpected string
```
markmandel, line 1505: Something to consider: Could pull the expected items into their own struct definition. I find it's a bit easier to read when there are > 1 expected items.

Don't have to do this, but thought I'd suggest it and see if it resonated for you.

pkg/apis/agones/v1/common.go
akremsa: Yes, it makes sense

pkg/apis/agones/v1/gameserver_test.go
akremsa: I like this idea - nice way to separate expected values. Applied all your suggestions, thanks for the review.

PR #1465, opened 2020-04-11
Update Grafana to the  6.7 release.

cyriltovena:
I also had to update helm to 2.12 for this to works. I've run e2e and verify that all dashboards where working correctly.
The gameserver dashboard is now using the new gauge graph panel, see screenshot below.

Not sure if there is anything to do with the build image ?

Signed-off-by: Cyril Tovena <cyril.tovena@gmail.com>

Screen Shot 2020-04-11 at 6 17 11 PM

PR #1487, opened 2020-04-23
Docs: add Grafana version explicitly

aLekSer:
Set Grafana version to the docs.
Add new make target `helm-repo-update`.

**What type of PR is this?**
/kind documentation

User can install previous version of the Grafana 6.2 which does not support new `GameServers` dashboard. This could lead to an error, which is hard to debug (but this error caused by missing `bargauge`):
```
undefined is not iterable (cannot read property Symbol(Symbol.iterator)) in grafana
```

Leftover of #1465.
Met in #1479.

BarGauge

### DIFF
--- a/site/content/en/docs/Guides/metrics.md
+++ b/site/content/en/docs/Guides/metrics.md
@@ -179,7 +179,7 @@ Now let's install some Grafana dashboards.
 
 ### Grafana installation
 
-Grafana is a open source time series analytics platform which supports Prometheus data source. We can also install easily import pre-built dashboards.
+Grafana is a open source time series analytics platform which supports Prometheus data source. We can also easily import pre-built dashboards.
 
 First we will install [Agones dashboard](#grafana-dashboards) as [config maps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) in our cluster.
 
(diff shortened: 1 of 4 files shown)


site/content/en/docs/Guides/metrics.md
```diff
@@ -190,8 +190,8 @@ kubectl apply -f ./build/grafana/
 Now we can install [grafana chart](https://github.com/helm/charts/tree/master/stable/grafana) from stable repository. (Replace `<your-admin-password>` with the admin password of your choice)
 
 ```bash
-helm install --wait --name grafana stable/grafana --namespace metrics \
-  --set adminPassword=<your-admin-password> -f ../build/grafana.yaml
+helm install --wait --name grafana stable/grafana --version=5.0.11 --namespace metrics \
```
markmandel, line 193: @cyriltovena any issue with specifying a specific version?

cyriltovena: No you can even make it latest. We release 6.7.3. Always good to be up to date.

cyriltovena: Which is 5.0.13 btw

aLekSer: 5.0.13 chart is still appVersion 6.7.1:
https://github.com/helm/charts/blob/866bf2f655ede9aa1bb6f4145db5862035cf49ad/stable/grafana/Chart.yaml#L4

markmandel: 5.0.13 is the latest helm chart too. :+1:

PR #1498, opened 2020-04-25
Add PlayerTracking IDs to SDK Convert function

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

When converting from CRD->gRPC model for SDK.GetGameServer(), add the Player Tracking IDs

Work on #1033

None

### DIFF
--- a/pkg/sdkserver/sdk_test.go
+++ b/pkg/sdkserver/sdk_test.go
@@ -94,13 +94,14 @@ func TestConvert(t *testing.T) {
 		assert.NoError(t, runtime.ParseFeatures(string(runtime.FeaturePlayerTracking)+"=true"))
 
 		gs := fixture.DeepCopy()
-		gs.Status.Players = &agonesv1.PlayerStatus{Capacity: 10, Count: 5}
+		gs.Status.Players = &agonesv1.PlayerStatus{Capacity: 10, Count: 5, IDs: []string{"one", "two"}}
 
 		sdkGs := convert(gs)
 		eq(t, fixture, sdkGs)
 		assert.Zero(t, sdkGs.ObjectMeta.DeletionTimestamp)
 		assert.Equal(t, gs.Status.Players.Capacity, sdkGs.Status.Players.Capacity)
 		assert.Equal(t, gs.Status.Players.Count, sdkGs.Status.Players.Count)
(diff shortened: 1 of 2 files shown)


pkg/sdkserver/sdk_test.go
```diff
@@ -94,13 +94,14 @@ func TestConvert(t *testing.T) {
...
 		gs := fixture.DeepCopy()
-		gs.Status.Players = &agonesv1.PlayerStatus{Capacity: 10, Count: 5}
+		gs.Status.Players = &agonesv1.PlayerStatus{Capacity: 10, Count: 5, IDs: []string{"one", "two"}}
 
 		sdkGs := convert(gs)
 		eq(t, fixture, sdkGs)
 		assert.Zero(t, sdkGs.ObjectMeta.DeletionTimestamp)
 		assert.Equal(t, gs.Status.Players.Capacity, sdkGs.Status.Players.Capacity)
 		assert.Equal(t, gs.Status.Players.Count, sdkGs.Status.Players.Count)
+		assert.Equal(t, gs.Status.Players.IDs, sdkGs.Status.Players.IDs)
```
aLekSer, line 104: I think in the future there is a need to disable (reset) `featureGates` after this and similar test.

markmandel: This is actually by design. This let's us randomly test out various areas of the code with Feature flags both on and off, and find errors -- it's already found one 😃

When setting feature flags at the beginning of a Test, it overwrites all feature flags, so you know you have the desired state at that point.

aLekSer: Thanks for an explanation, just was curious what was expected here.

markmandel: This is also why we have the `runtime.FeatureTestMutex.Lock()` in place, so that Feature Flag'd tests can't interfere with each other.

I found in my testing of this, you actually can't rely on removing t.Parrallel, as without it, go test runs the tests _in that file_ in serial, but each file becomes its own executable, and they are run in parallel, and everything becomes hard at that point.

aLekSer: Well this might be also reasonable to test some combinations of featureFlags.

PR #1507, opened 2020-05-01
SDKServer Player Tracking implementation

markmandel:
**What type of PR is this?**
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Implementation of the SDK gRPC Server, with accompanying unit tests.

Work on #1033

None.

### DIFF
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -38,6 +38,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/clock"
@@ -53,10 +54,12 @@ import (
 type Operation string
 
 const (
-	updateState          Operation = "updateState"
-	updateLabel          Operation = "updateLabel"
-	updateAnnotation     Operation = "updateAnnotation"
-	updatePlayerCapacity Operation = "updatePlayerCapacity"
+	updateState             Operation     = "updateState"
+	updateLabel             Operation     = "updateLabel"
+	updateAnnotation        Operation     = "updateAnnotation"
+	updatePlayerCapacity    Operation     = "updatePlayerCapacity"
+	updateConnectedPlayers  Operation     = "updateConnectedPlayers"
+	playerCountUpdatePeriod time.Duration = time.Second
 )
 
 var (
@@ -96,6 +99,7 @@ type SDKServer struct {
 	reserveTimer       *time.Timer
 	gsReserveDuration  *time.Duration
 	gsPlayerCapacity   int64
+	gsConnectedPlayers []string
 }
 
 // NewSDKServer creates a SDKServer that sets up an
@@ -129,6 +133,7 @@ func NewSDKServer(gameServerName, namespace string, kubeClient kubernetes.Interf
 		gsAnnotations:      map[string]string{},
 		gsUpdateMutex:      sync.RWMutex{},
 		gsWaitForSync:      sync.WaitGroup{},
+		gsConnectedPlayers: []string{},
 	}
 
 	s.informerFactory = factory
@@ -230,11 +235,12 @@ func (s *SDKServer) Run(stop <-chan struct{}) error {
 		go wait.Until(s.runHealth, s.healthTimeout, stop)
 	}
 
-	// populate player capacity value
+	// populate player tracking values
 	if runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
 		s.gsUpdateMutex.Lock()
 		if gs.Status.Players != nil {
 			s.gsPlayerCapacity = gs.Status.Players.Capacity
+			s.gsConnectedPlayers = gs.Status.Players.IDs
 		}
 		s.gsUpdateMutex.Unlock()
 	}
@@ -272,6 +278,8 @@ func (s *SDKServer) syncGameServer(key string) error {
 		return s.updateAnnotations()
 	case updatePlayerCapacity:
 		return s.updatePlayerCapacity()
+	case updateConnectedPlayers:
+		return s.updateConnectedPlayers()
 	}
 
 	return errors.Errorf("could not sync game server key: %s", key)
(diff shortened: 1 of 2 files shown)


pkg/sdkserver/sdkserver_test.go
```diff
@@ -985,6 +986,186 @@ func TestSDKServerPlayerCapacity(t *testing.T) {
...
+	list, err := sc.GetConnectedPlayers(context.Background(), e)
+	assert.NoError(t, err)
+	assert.Empty(t, list.List)
+
+	ok, err := sc.IsPlayerConnected(context.Background(), &alpha.PlayerID{PlayerID: "1"})
+	assert.NoError(t, err)
+	assert.False(t, ok.Bool, "no player connected yet")
+
+	// sdk value should always be correct, even if we send more than one update per second.
+	for i := 0; i < 3; i++ {
```
aLekSer, line 1060: Probably it would be easier to understand why `3` (if I got it right):
```suggestion
	for i := 0; i < capacity; i++ {
```

pkg/sdkserver/sdkserver_test.go
```diff
@@ -985,6 +986,186 @@ func TestSDKServerPlayerCapacity(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.True(t, ok.Bool, "Player "+token+" should not yet be connected")
+
+		ok, err = sc.IsPlayerConnected(context.Background(), id)
+		assert.NoError(t, err)
+		assert.True(t, ok.Bool, "Player "+token+" should be connected")
+	}
+	count, err = sc.GetPlayerCount(context.Background(), e)
+	assert.NoError(t, err)
+	assert.Equal(t, int64(3), count.Count)
```
aLekSer, line 1073: Same as above?
```suggestion
	assert.Equal(t, capacity, count.Count)
```

pkg/sdkserver/sdkserver_test.go
```diff
@@ -985,6 +986,186 @@ func TestSDKServerPlayerCapacity(t *testing.T) {
...
+
+	list, err = sc.GetConnectedPlayers(context.Background(), e)
+	assert.NoError(t, err)
+	assert.Equal(t, []string{"0", "1", "2"}, list.List)
+
+	// on an update, confirm that the update hits the K8s api, only once
+	select {
+	case value := <-updated:
+		assert.Equal(t, int64(3), value.Count)
```
aLekSer, line 1082: here also
```suggestion
		assert.Equal(t, capacity, value.Count)
```

pkg/sdkserver/sdkserver_test.go
```diff
@@ -985,6 +986,186 @@ func TestSDKServerPlayerCapacity(t *testing.T) {
...
+	}
+
+	// should return an error if we try and add another, since we're at capacity
+	nopePlayer := &alpha.PlayerID{PlayerID: "nope"}
+	_, err = sc.PlayerConnect(context.Background(), nopePlayer)
+	assert.EqualError(t, err, "Players are already at capacity")
+
+	// sdk value should always be correct, even if we send more than one update per second.
+	// let's leave one player behind
+	for i := 0; i < 2; i++ {
```
aLekSer, line 1103:
```suggestion
	for i := 0; i < capacity-1; i++ {
```

pkg/sdkserver/sdkserver_test.go
```diff
@@ -985,6 +986,186 @@ func TestSDKServerPlayerCapacity(t *testing.T) {
...
+		token := strconv.Itoa(i)
+		id := &alpha.PlayerID{PlayerID: token}
+		ok, err := sc.PlayerDisconnect(context.Background(), id)
+		assert.NoError(t, err)
+		assert.True(t, ok.Bool, "Player "+token+" should be disconnected")
```
aLekSer, line 1108: Might be a bit easier to read
```suggestion
		assert.True(t, ok.Bool, "Player %s should be disconnected", token)
```

pkg/sdkserver/sdkserver.go
```diff
@@ -547,35 +555,108 @@ func (s *SDKServer) stopReserveTimer() {
 // [Stage:Alpha]
 // [FeatureFlag:PlayerTracking]
 func (s *SDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerID) (*alpha.Bool, error) {
-	panic("implement me")
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return nil, errors.New(string(runtime.FeaturePlayerTracking) + " not enabled")
```
aLekSer, line 559: Could be something like:
```suggestion
		return nil, errors.New(fmt.Sprintf("feature gate '%s' is not enabled", runtime.FeaturePlayerTracking))
```

pkg/sdkserver/sdkserver.go
```diff
@@ -547,35 +555,108 @@ func (s *SDKServer) stopReserveTimer() {
...
+
+	// the player is already connected, return false.
+	for _, playerID := range s.gsConnectedPlayers {
+		if playerID == id.PlayerID {
+			return &alpha.Bool{Bool: false}, nil
+		}
+	}
+
+	if int64(len(s.gsConnectedPlayers)) >= s.gsPlayerCapacity {
+		return &alpha.Bool{}, errors.New("Players are already at capacity")
```
aLekSer, line 574: lower case for errors
https://github.com/golang/go/wiki/CodeReviewComments#error-strings
```suggestion
		return &alpha.Bool{}, errors.New("players are already at capacity")
```

pkg/sdkserver/sdkserver.go
```diff
@@ -547,35 +555,108 @@ func (s *SDKServer) stopReserveTimer() {
...
+		if playerID == id.PlayerID {
+			found = i
+		}
+	}
+	if found == -1 {
+		return &alpha.Bool{Bool: false}, nil
+	}
+
+	// let's retain the original order, as it should be a smaller patch on data change
+	s.gsConnectedPlayers = append(s.gsConnectedPlayers[:found], s.gsConnectedPlayers[found+1:]...)
```
aLekSer, line 607: Would it work if found is of len(found) ?

pkg/sdkserver/sdkserver.go
```diff
@@ -547,35 +555,108 @@ func (s *SDKServer) stopReserveTimer() {
...
+
+	return &alpha.Bool{Bool: true}, nil
 }
 
-// IsPlayerConnected returns if the player ID is connected or not
+// IsPlayerConnected returns if the playerID is currently connected to the GameServer.
+// This is always accurate, even if the value hasn’t been updated to the GameServer status yet.
 // [Stage:Alpha]
 // [FeatureFlag:PlayerTracking]
 func (s *SDKServer) IsPlayerConnected(ctx context.Context, id *alpha.PlayerID) (*alpha.Bool, error) {
```
aLekSer, line 617: let `IsPlayerConnected(ctx, nil)` return `false, errors.New("Nil parameter")` and not lead to panic.

pkg/sdkserver/sdkserver.go
```diff
@@ -547,35 +555,108 @@ func (s *SDKServer) stopReserveTimer() {
...
+	s.gsConnectedPlayers = append(s.gsConnectedPlayers, id.PlayerID)
+	s.workerqueue.EnqueueAfter(cache.ExplicitKey(string(updateConnectedPlayers)), playerCountUpdatePeriod)
+
+	return &alpha.Bool{Bool: true}, nil
 }
 
 // PlayerDisconnect should be called when a player disconnects.
 // [Stage:Alpha]
 // [FeatureFlag:PlayerTracking]
 func (s *SDKServer) PlayerDisconnect(ctx context.Context, id *alpha.PlayerID) (*alpha.Bool, error) {
```
aLekSer, line 587: Same here I think we should check if id is nil at the beginning of the function.

pkg/sdkserver/sdkserver.go
```diff
@@ -547,35 +555,108 @@ func (s *SDKServer) stopReserveTimer() {
...
-	panic("implement me")
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return nil, errors.New(string(runtime.FeaturePlayerTracking) + " not enabled")
+	}
+	s.logger.WithField("PlayerId", id).Debug("Player Disconnected")
```
aLekSer, line 591: Not sure if previous version parse pointers in a correct way.
```suggestion
	s.logger.WithField("PlayerId", id. PlayerID).Debug("Player Disconnected")
```

pkg/sdkserver/sdkserver.go
```diff
@@ -547,35 +555,108 @@ func (s *SDKServer) stopReserveTimer() {
...
+
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+
+	found := -1
+	for i, playerID := range s.gsConnectedPlayers {
+		if playerID == id.PlayerID {
+			found = i
```
aLekSer, line 599: break here?

pkg/sdkserver/sdkserver.go
```diff
@@ -694,3 +775,36 @@ func (s *SDKServer) updatePlayerCapacity() error {
 	s.recorder.Event(gs, corev1.EventTypeNormal, "PlayerCapacity", fmt.Sprintf("Set to %d", gs.Status.Players.Capacity))
 	return nil
 }
+
+// updateConnectedPlayers updates the Player IDs and Count fields in the GameServer's Status.
+func (s *SDKServer) updateConnectedPlayers() error {
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return errors.New(string(runtime.FeaturePlayerTracking) + " not enabled")
+	}
+	s.logger.WithField("playerIDs", s.gsConnectedPlayers).Debug("updating connected players")
```
aLekSer, line 784: Might go also after lock.
```suggestion
        s.gsUpdateMutex.RLock()
	s.logger.WithField("playerIDs", s.gsConnectedPlayers).Debug("updating connected players")
```

pkg/sdkserver/sdkserver_test.go
markmandel: Oh I like that. Nice.

pkg/sdkserver/sdkserver.go
markmandel: We've currently got this code all through localsdk.go as well. Should I change both?

markmandel: In fact, this should `errors.Errorf(...)` - which is much better. I'll make the change here, and come back around to localsdk.go in a later PR.

pkg/sdkserver/sdkserver.go
markmandel: I can't see this ever actually happening, and if it does, a panic should occur, so we fix it. I think I'll leave this as is.

pkg/sdkserver/sdkserver.go
markmandel: Ah, excellent point.

pkg/sdkserver/sdkserver.go
markmandel: Also excellent suggestion.

pkg/sdkserver/sdkserver.go
markmandel: Good catch!

pkg/sdkserver/sdkserver.go
```diff
@@ -547,43 +555,117 @@ func (s *SDKServer) stopReserveTimer() {
 // [Stage:Alpha]
 // [FeatureFlag:PlayerTracking]
 func (s *SDKServer) PlayerConnect(ctx context.Context, id *alpha.PlayerID) (*alpha.Bool, error) {
-	panic("implement me")
+	if !runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		return nil, errors.Errorf("%s not enabled", runtime.FeaturePlayerTracking)
+	}
+	s.logger.WithField("PlayerId", id).Debug("Player Connected")
```
aLekSer, line 561: Sorry Mark, one more
```suggestion
	s.logger.WithField("PlayerId", id.PlayerID).Debug("Player Connected")
```

markmandel: Good catch. Applying now!

PR #1510, opened 2020-05-02
Unity SDK: Fix SpecHealth parsing

cadfoot:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Fix type cast error in Unity SDK when parsing SpecHealth

None

The issue occurs only when health check is disabled.
`disabled` value is already type of `bool` (boxed), thus casting it to `string` causes `InvalidCastException`


cadfoot:
@googlebot I signed it!

cadfoot:
just found out I messed up with the last commit and brought bad casting back, fixing this

PR #1519, opened 2020-05-06
PlayerConnect/Disconnect & related Go SDK functions

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Includes implementations and unit tests for PlayerConnect, PlayerDisconnect, GetPlayerCount, IsPlayerConnected and GetConnectedPlayers.

Conformance tests will come in the next PR.

Work on #1507


sdks/go/alpha.go
```diff
@@ -46,3 +46,40 @@ func (a *Alpha) SetPlayerCapacity(capacity int64) error {
...
+// PlayerConnect increases the SDK’s stored player count by one, and appends this playerID to status.players.id.
+// Returns true and adds the playerID to the list of playerIDs if the playerIDs was not already in the
+// list of connected playerIDs.
+func (a *Alpha) PlayerConnect(id string) (bool, error) {
+	ok, err := a.client.PlayerConnect(context.Background(), &alpha.PlayerID{PlayerID: id})
+	return ok.Bool, errors.Wrap(err, "could not register connected player")
+}
+
+// PlayerDisconnect Decreases the SDK’s stored player count by one, and removes the playerID from status.players.id
+//
```
aLekSer, line 59: Extra line could be removed
```suggestion
```

sdks/go/alpha.go
```diff
@@ -46,3 +46,40 @@ func (a *Alpha) SetPlayerCapacity(capacity int64) error {
...
+func (a *Alpha) GetPlayerCount() (int64, error) {
+	count, err := a.client.GetPlayerCount(context.Background(), &alpha.Empty{})
+	return count.Count, errors.Wrap(err, "could not get player count")
+}
+
+// IsPlayerConnected returns if the playerID is currently connected to the GameServer.
+// This is always accurate, even if the value hasn’t been updated to the GameServer status yet.
+func (a *Alpha) IsPlayerConnected(id string) (bool, error) {
+	ok, err := a.client.IsPlayerConnected(context.Background(), &alpha.PlayerID{PlayerID: id})
+	return ok.Bool, errors.Wrap(err, "could not get if played is connection")
```
aLekSer, line 77:
```suggestion
	return ok.Bool, errors.Wrap(err, "could not get if player is connected")
```

sdks/go/alpha_test.go
```diff
@@ -36,26 +36,66 @@ func TestAlphaGetAndSetPlayerCapacity(t *testing.T) {
 	capacity, err := a.GetPlayerCapacity()
```
aLekSer, line 36: I think there should be `FeaturePlayerTracking` enabled for this test:
```
err := runtime.ParseFeatures(string(runtime.FeaturePlayerTracking) + "=true")
				assert.NoError(t, err)
```
WDYT?

markmandel: I'm going to disagree. This is testing against a mock, of which there is only very light logic, that's also not controlled by a feature flag.

There is no need for feature flags at the SDK level either, since this is all controlled by the backing SDK Server.

So I don't see any reason to have this here? Unless I'm missing something.

sdks/go/alpha.go
markmandel: Nice catch.

sdks/go/alpha.go
markmandel: Ah! much better!

sdks/go/alpha_test.go
aLekSer: Thanks for the explanation. Agree with you.

PR #1521, opened 2020-05-07
Add tolerations for delete hook script

### LINKED ISSUE #1504: Helm delete doesn't support tolerations/affinities
**What happened**:
Helm delete will fail due to the job created by hook script `pre_delete_hook.yaml` if there are tolerations/affinities.

Successful release deletion when running helm delete.

Add tolerations/affinities to cluster, run helm delete.

**Environment**:
- Agones version: 1.4.0
- Kubernetes version (use `kubectl version`): 1.14
- Others:

aLekSer:
Tolerations for delete Agones resources same as for the controller.

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature

/kind hotfix

Currently Job for deleting resources created on default node pool. If we add taints to all nodes in default node pool, than this Job could not be scheduled.
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

Closes #1504

### DIFF
--- a/install/helm/agones/templates/hooks/pre_delete_hook.yaml
+++ b/install/helm/agones/templates/hooks/pre_delete_hook.yaml
@@ -47,6 +47,10 @@ spec:
         volumeMounts:
         - name: script
           mountPath: /scripts/
+      {{- if .Values.agones.controller.tolerations }}
+      tolerations:
+{{ toYaml .Values.agones.controller.tolerations | indent 8 }}
+      {{- end }}
       volumes:
       - name: script
         configMap:


install/helm/agones/templates/hooks/pre_delete_hook.yaml
```diff
@@ -47,6 +47,10 @@ spec:
         volumeMounts:
         - name: script
           mountPath: /scripts/
+      {{- if .Values.agones.controller.tolerations }}
```
markmandel, line 50: I like this approach. It's the easiest implementation. If people need to have their own custom toleration, we could provide an extra configuration for delete_job toleration, and then default to the controller toleration :+1:

PR #1527, opened 2020-05-07
Player Tracking Go SDK Conformance Tests

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Conformance tests for the Go SDK.

Work on #1507

None.

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -34,8 +34,7 @@ SDK_FOLDER ?= go
 COMMAND ?= gen
 SDK_IMAGE_TAG=$(build_sdk_prefix)$(SDK_FOLDER):$(build_sdk_version)
 DEFAULT_CONFORMANCE_TESTS = ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch,reserve
-ALPHA_CONFORMANCE_TESTS = getplayercapacity,setplayercapacity
-
+ALPHA_CONFORMANCE_TESTS = getplayercapacity,setplayercapacity,playerconnect,playerdisconnect,getplayercount,isplayerconnected,getconnectedplayers
 
 .PHONY: test-sdks test-sdk build-sdks build-sdk gen-all-sdk-grpc gen-sdk-grpc run-all-sdk-command run-sdk-command build-example
 
--- a/test/sdk/go/sdk-client-test.go
+++ b/test/sdk/go/sdk-client-test.go
@@ -106,6 +106,37 @@ func main() {
 		if c != capacity {
 			log.Fatalf("Player Capacity should be %d, but is %d", capacity, c)
 		}
+
+		playerID := "1234"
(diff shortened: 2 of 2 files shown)


build/includes/sdk.mk
```diff
@@ -34,8 +34,7 @@ SDK_FOLDER ?= go
 COMMAND ?= gen
 SDK_IMAGE_TAG=$(build_sdk_prefix)$(SDK_FOLDER):$(build_sdk_version)
 DEFAULT_CONFORMANCE_TESTS = ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch,reserve
-ALPHA_CONFORMANCE_TESTS = getplayercapacity,setplayercapacity
-
+ALPHA_CONFORMANCE_TESTS = getplayercapacity,setplayercapacity,playerconnect,playerdisconnect,getplayercount,isplayerconnected,getconnectedplayers
```
roberthbailey, line 37: it looks like these don't have to be in the same order as the code. do we just check that they were all called?

test/sdk/go/sdk-client-test.go
```diff
@@ -106,6 +106,37 @@ func main() {
...
+			log.Fatalf("No connected players returned")
+		}
+
+		if ok, err := sdk.Alpha().PlayerDisconnect(playerID); err != nil {
+			log.Fatalf("Error registering player as disconnected: %s", err)
+		} else if !ok {
+			log.Fatalf("PlayerDisconnect returned false")
+		}
+
+		if c, err = sdk.Alpha().GetPlayerCount(); err != nil {
```
roberthbailey, line 135: I don't know if it would work, but it would be nice to call this once with the player connected (and verify that result is 1) and then again here to check that it is back to 0. 

Similarly, it would be nice to verify that IsPlayerConnected returns false for a non-existent player and also that it returns false for "1234" after the player disconnects.

I'm sure this is all covered by unit tests, but it would be nice to have it covered by the conformance tests as well.

build/includes/sdk.mk
markmandel: You are right, it doesn't matter the order they are called in - it checks that SDK methods are called with the expected arguments.

test/sdk/go/sdk-client-test.go
markmandel: This is definitely covered by the unit tests, but I can't see how we could do this with the current structure of the conformance tests. The conformance tests check that a method has been called once with a specific value, so there is no concept for ordering.

@aLekSer unless you can think of anything I'm not thinking of?

roberthbailey: That's what I thought. I don't want to block this change, but let's file an issue so we can discuss if it would be worth modifying the conformance test infra.

PR #1531, opened 2020-05-08
Improved fleetautoscalers - TestApplyWebhookPolicy unit tests + applyWebhookPolicy refactoring

akremsa:
**What type of PR is this?**
> /kind cleanup

Added missing unit tests to _TestApplyWebhookPolicy_ and refactored _applyWebhookPolicy_ method. 
Added an absent error check after `b, err := json.Marshal(faReq)` which could be a potential bug.

Before:
Screenshot 2020-05-08 at 16 47 18

After:
image

By the way, I suggest using _url.ParseRequestURI_ instead of _url.Parse_.


pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -53,82 +54,109 @@ func computeDesiredFleetSize(fas *autoscalingv1.FleetAutoscaler, f *agonesv1.Fle
...
+	if f == nil {
+		return 0, false, errors.New("nil Fleet passed")
 	}
-	var faResp autoscalingv1.FleetAutoscaleReview
-	servicePath := ""
-	if w.Service != nil {
-		if w.Service.Path != nil {
-			servicePath = *w.Service.Path
+
+	if w.URL != nil && w.Service != nil {
```
aLekSer, line 65: 👍 additional check would be great. Validated on Update but still something wrong could happen.

akremsa: Agree, taking into account the fact that another validation happens in a different package.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -31,6 +31,7 @@ import (
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	autoscalingv1 "agones.dev/agones/pkg/apis/autoscaling/v1"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
```
aLekSer, line 34: I think there is no need to have an alias for logrus:
```suggestion
	"github.com/sirupsen/logrus"
```

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,100 +42,131 @@ var client = http.Client{
...
 		strings.NewReader(string(b)),
 	)
 	if err != nil {
-		return f.Status.Replicas, false, err
+		return 0, false, err
 	}
-	defer res.Body.Close() // nolint: errcheck
+	defer func() {
+		if cerr := res.Body.Close(); cerr != nil {
+			log.Error(cerr)
```
aLekSer, line 138: I think we can use `c.baseLogger` here, we could add receiver to the function `(c *Controller) applyWebhookPolicy()`:
https://github.com/googleforgames/agones/blob/cf3d21dbd4d2200760f0edf0198dfd4de2806477/pkg/fleetautoscalers/controller.go#L87

akremsa: Decided to handle it in a different way, please check.

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -152,55 +79,535 @@ func (t testServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			} else {
+				assert.Equal(t, tc.expected.replicas, replicas)
+				assert.Equal(t, tc.expected.limited, limited)
+				assert.Nil(t, err)
+			}
+		})
+	}
+}
+
+func TestApplyWebhookPolicy_NilFleet(t *testing.T) {
```
markmandel, line 545:
```suggestion
func TestApplyWebhookPolicyNilFleet(t *testing.T) {
```

Nit

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,100 +41,134 @@ var client = http.Client{
...
-
 	switch fas.Spec.Policy.Type {
 	case autoscalingv1.BufferPolicyType:
 		return applyBufferPolicy(fas.Spec.Policy.Buffer, f)
 	case autoscalingv1.WebhookPolicyType:
 		return applyWebhookPolicy(fas.Spec.Policy.Webhook, f)
 	}
 
-	return f.Status.Replicas, false, errors.New("wrong policy type, should be one of: Buffer, Webhook")
+	return 0, false, errors.New("wrong policy type, should be one of: Buffer, Webhook")
```
markmandel, line 51: Curious why the change to 0?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,100 +41,134 @@ var client = http.Client{
...
+			Status:    f.Status,
+		},
+		Response: nil,
 	}
 
-	u, err := url.Parse(urlStr)
+	b, err := json.Marshal(faReq)
 	if err != nil {
-		return f.Status.Replicas, false, err
+		return 0, false, err
```
markmandel, line 123: Again, why the change to 0? We have a consistent pattern across the code base of not doing this, so I'm curious why we're deviating from it at this stage?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,100 +41,134 @@ var client = http.Client{
...
 
 		if w.Service.Namespace == "" {
 			w.Service.Namespace = "default"
 		}
-		scheme := "http://"
+
+		scheme := "http"
 		if w.CABundle != nil {
-			scheme = "https://"
+			scheme = "https"
```
markmandel, line 93: This looks like some seriously comprehensive validation! Nice stuff!

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: I didn't notice this pattern that is why I decided to make these changes.
I believe this is a common practice to return either a successful result and nil error or a default value and non-nil value. When I return 0 I explicitly show that it is not expected to process this value anymore, only an error should be handled. When I see code where both f.Status.Replicas(for example) and error are returned I am not sure whether these replicas are supposed to be handled somehow outside of this function or not. It brings some sort of ambiguity to the code, in my opinion. That is why I think it is better to return default values explicitly. According to this function, there are no places where both int and err values are used.
Why do you think we should return both values as is?

markmandel: Essentially, if you are returning an error, the value that accompany's it has no value -- since there is an error, something bad has happened, so there is no extra value in being specific and allocating a new value when a semantic one is already present.

I'm trying to find a convention wherein it should go one way or another, but no success so far. @roberthbailey you've always got thoughts on Go readability and conventions -- care to weight in and tiebreak here?

akremsa: If we are talking about allocations, we return basic types in this function - (int32, bool, error). There is no heap allocation here.

pkg/fleetautoscalers/fleetautoscalers.go
roberthbailey: If an error is returned, shouldn't this value be ignored by the caller?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,100 +41,134 @@ var client = http.Client{
...
-	servicePath := ""
-	if w.Service != nil {
-		if w.Service.Path != nil {
-			servicePath = *w.Service.Path
+
+	if w.URL != nil && w.Service != nil {
+		return 0, false, errors.New("service and url cannot be used simultaneously")
+	}
+
+	var u *url.URL
```
roberthbailey, line 67: The section to set the url is complicated enough that it might be nice to split into a separate helper method (which could also make the control flow a bit simpler and reduce the number of levels of indentation).

pkg/fleetautoscalers/fleetautoscalers.go
roberthbailey: I've usually seen what @akremsa said - you return a default value for the type (e.g. nil for a pointer, 0 for integer, "" for a string, etc) since you expect the caller to ignore the value. 

I've looked through https://golang.org/doc/effective_go.html#errors, https://blog.golang.org/error-handling-and-go, and https://github.com/golang/go/wiki/CodeReviewComments and don't see any strong guidance one way or the other. 

From those links I checked out the golang source for file.go and it uses the pattern of returning `"", err` or `0, err` depending on the function signature (e.g. https://golang.org/src/os/file.go?s=8470:8507#L151). 

So if you want me to tie-break I'd go with returning the default value for the type.

markmandel: > So if you want me to tie-break I'd go with returning the default value for the type.

Okily! In which case, I remove my objections, and resolve this conversation.

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: @roberthbailey is it better?

roberthbailey: Part of the reason I suggested making it a separate function is so that you could reduce the indentation and simplify the code flows (since you can return right away on errors).

```
func buildURLFromWebhookPolicy(w *autoscalingv1.WebhookPolicy) (*url.URL, error) {
  if w.URL != nil {
    if *w.URL == "" {
      return nil, errors.New("URL was not provided")
    }
    return url.ParseRequestURI(*w.URL)
  }

  // notice how the next block of code can now be at the top level since the previous logic has returned in all branches
  if w.Service.Name == "" {
    return nil, errors.New("service name was not provided")
  }
  var servicePath string
  if w.Service.Path != nil {
    servicePath = *w.Service.Path
  }
  if w.Service.Namespace == "" {
    w.Service.Namespace = "default"
  }
  scheme := "http"
  if w.CABundle != nil {
    scheme = "https"
    if err := setCABundle(w.CABundle); err != nil {
      return nil, err
    }
  }
  return &url.URL{
    Scheme: scheme,
    Host:   fmt.Sprintf("%s.%s.svc:8000", w.Service.Name, w.Service.Namespace),
    Path:   servicePath,
  }, nil
}
```

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,94 +41,147 @@ var client = http.Client{
...
 		if err != nil {
-			return f.Status.Replicas, false, err
+			return nil, err
+		}
+	} else {
+		if w.Service.Name == "" {
+			return nil, errors.New("service name was not provided")
+		}
+
+		var servicePath string
```
roberthbailey, line 72: I'm not clear on why service Path is stored as a local variable but service namespace and the ca bundle are set on the webhook policy.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,94 +41,147 @@ var client = http.Client{
...
+			scheme = "https"
+
+			err = setCABundle(w.CABundle)
+			if err != nil {
+				return nil, err
+			}
 		}
-		urlStr = fmt.Sprintf("%s%s.%s.svc:8000/%s", scheme, w.Service.Name, w.Service.Namespace, servicePath)
+
+		u = createURL(scheme, w.Service.Name, w.Service.Namespace, servicePath)
```
roberthbailey, line 91: createURL is only used once and just sets fields on a struct, so I don't see why it isn't inlined.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,94 +41,147 @@ var client = http.Client{
...
+	}
+	client.Transport = &http.Transport{
+		TLSClientConfig: &tls.Config{
+			RootCAs: rootCAs,
+		},
+	}
+	return nil
+}
+
+func applyWebhookPolicy(w *autoscalingv1.WebhookPolicy, f *agonesv1.Fleet) (replicas int32, limited bool, err error) {
```
roberthbailey, line 120: why does this function use named return parameters?

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: Ok, now I get your point. Updated.

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: changed this logic

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: I've created a separate method in order to write several simple unit tests to check that URL corresponds to the following pattern - _name.namespace.svc:8000_
https://github.com/googleforgames/agones/pull/1531/files#diff-81c1d605645c546c9896631e7db55addR582

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: I did that in order to catch and wrap an error in a defer() statement in case of an error after res.Body.Close()

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,18 +41,94 @@ var client = http.Client{
...
+
+	if w.Service.Namespace == "" {
+		w.Service.Namespace = "default"
+	}
+
+	scheme := "http"
+	if w.CABundle != nil {
+		scheme = "https"
+
+		err = setCABundle(w.CABundle)
```
roberthbailey, line 82: nit:

`if err := setCABundle(w.CABundle); err != nil { ... }`

pkg/fleetautoscalers/fleetautoscalers.go
roberthbailey: That makes sense. Can you add a comment to the createURL function so that this is clear to future maintainers?

akremsa: done

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: done

akremsa:
seems some failing tests should be updated due to new Golang version 1.14.4

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,18 +41,92 @@ var client = http.Client{
...
-
 	switch fas.Spec.Policy.Type {
 	case autoscalingv1.BufferPolicyType:
 		return applyBufferPolicy(fas.Spec.Policy.Buffer, f)
 	case autoscalingv1.WebhookPolicyType:
 		return applyWebhookPolicy(fas.Spec.Policy.Webhook, f)
 	}
 
-	return f.Status.Replicas, false, errors.New("wrong policy type, should be one of: Buffer, Webhook")
+	return 0, false, errors.New("wrong policy type, should be one of: Buffer, Webhook")
```
aLekSer, line 51: Agree with this change. This function is used only in `syncFleetAutoscaler()` and `if err != nil` it will return without updating the fleet.

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -152,55 +79,550 @@ func (t testServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			statusAllocatedReplicas: -20,
+			statusReadyReplicas:     40,
+			expected: expected{
+				replicas: 50,
+				limited:  false,
+				err:      "invalid character 'i' looking for beginning of value",
+			},
+		},
+		{
+			description: "Nil URL, empty namespace - default one should be used, no such host err is returned",
```
aLekSer, line 524: I see that url is not nil here, cause it is `&url`

akremsa: URL as a WebhookClientConfig field is nil. It can be either URL or Service provided. Check this case, for example :https://github.com/googleforgames/agones/pull/1531/files/93d0ff777d37bc3a44357f02f56c7c3f40423ea4#diff-81c1d605645c546c9896631e7db55addR408

akremsa:
@markmandel @roberthbailey
PR is updated and ready to be merged

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -152,55 +79,592 @@ func (t testServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				assert.Nil(t, err)
+			}
+		})
+	}
+}
+
+func TestApplyWebhookPolicyNilFleet(t *testing.T) {
+	t.Parallel()
+
+	url := "/scale"
```
aLekSer, line 548:
```suggestion
	url := "scale"
```
Can you please check if we can omit redundant slash sign.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,18 +41,92 @@ var client = http.Client{
...
+	if w == nil {
+		return 0, false, errors.New("nil WebhookPolicy passed")
+	}
+
+	if f == nil {
+		return 0, false, errors.New("nil Fleet passed")
+	}
+
+	if w.URL != nil && w.Service != nil {
+		return 0, false, errors.New("service and url cannot be used simultaneously")
```
aLekSer, line 122:
```suggestion
		return 0, false, errors.New("service and URL cannot be used simultaneously")
```

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,18 +41,92 @@ var client = http.Client{
...
+	return nil
+}
+
+func applyWebhookPolicy(w *autoscalingv1.WebhookPolicy, f *agonesv1.Fleet) (replicas int32, limited bool, err error) {
+	if w == nil {
+		return 0, false, errors.New("nil WebhookPolicy passed")
+	}
+
+	if f == nil {
+		return 0, false, errors.New("nil Fleet passed")
```
aLekSer, line 118: I would prefer error message in this form:
```suggestion
		return 0, false, errors.New("fleet parameter must not be nil")
```

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,18 +41,92 @@ var client = http.Client{
...
+		TLSClientConfig: &tls.Config{
+			RootCAs: rootCAs,
+		},
+	}
+	return nil
+}
+
+func applyWebhookPolicy(w *autoscalingv1.WebhookPolicy, f *agonesv1.Fleet) (replicas int32, limited bool, err error) {
+	if w == nil {
+		return 0, false, errors.New("nil WebhookPolicy passed")
```
aLekSer, line 114: Same as below:
```suggestion
		return 0, false, errors.New("WebhookPolicy parameter must not be nil")
```

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -152,55 +79,592 @@ func (t testServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+			if assert.NotNil(t, res) {
+				assert.Equal(t, tc.expected, res.String())
+			}
+		})
+	}
+}
+
+func TestBuildURLFromWebhookPolicyNoNamespace(t *testing.T) {
+	url := "/testurl"
```
aLekSer, line 616:
```suggestion
	url := "testurl"
```

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -41,18 +41,92 @@ var client = http.Client{
...
+func buildURLFromWebhookPolicy(w *autoscalingv1.WebhookPolicy) (*url.URL, error) {
+	if w.URL != nil {
+		if *w.URL == "" {
+			return nil, errors.New("URL was not provided")
+		}
+
+		return url.ParseRequestURI(*w.URL)
+	}
+
+	if w.Service.Name == "" {
```
aLekSer, line 63: Major change, please add one more nil check here:
```suggestion
	if w.Service == nil {
		return nil, errors.New("WebhookPolicy should contain URL or Service")
	}
	if w.Service.Name == "" {
```

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -152,55 +79,592 @@ func (t testServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		description             string
+		webhookPolicy           *autoscalingv1.WebhookPolicy
+		specReplicas            int32
+		statusReplicas          int32
+		statusAllocatedReplicas int32
+		statusReadyReplicas     int32
+		expected                expected
+	}{
+		{
+			description: "Allocated replicas per cent < 70%, no scaling",
```
aLekSer, line 351: Please add one more test case here, when both Service and URL would be nil.

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -152,55 +79,592 @@ func (t testServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			},
+			expected: expected{
+				url: "http://service1.default.svc:8000/testurl",
+				err: "",
+			},
+		},
+		{
+			description: "No url provided, empty string should be used",
+			webhookPolicy: &autoscalingv1.WebhookPolicy{
+				Service: &admregv1b.ServiceReference{
```
aLekSer, line 645: Same as above: Please add one more test case here, when both Service and URL would be nil

pkg/fleetautoscalers/fleetautoscalers_test.go
akremsa: Yes, we can, it is mentioned in docs. I will remove this slash.

pkg/fleetautoscalers/fleetautoscalers.go
akremsa: Very good suggestion! Thanks for the review.

pkg/fleetautoscalers/fleetautoscalers_test.go
akremsa: Actually it's enough to add this test in one test suit.

akremsa:
@aLekSer done

Issue #1534, opened 2020-05-08
Change the multi-cluster allocation API version to stable


pooneh-m:
**Is your feature request related to a problem? Please describe.**
Currently the API version for multi-cluster allocation is v1alpha. With more feature adaptation, the upgrade story for the cluster mesh using the multi-cluster allocation feature will be more complicated.

**Describe the solution you'd like**
Change the multi-cluster allocation API to a stable version. Any new change after that should be implemented in a backward compatible manner.

PR #1540, opened 2020-05-13
Change the multi-cluster allocation API version to stable

pooneh-m:
**What type of PR is this?**
/kind breaking

Changes the API version for multi-cluster allocation to stable.

Closes https://github.com/googleforgames/agones/issues/1534


steven-supersolid:
I'm not sure why but the Node.js SDK `package-lock.json` file was changed so this should probably be removed from the review

pooneh-m:
> I'm not sure why but the Node.js SDK `package-lock.json` file was changed so this should probably be removed from the review

Thanks. I removed the package-lock.json.

PR #1552, opened 2020-05-14
Player Tracking: Json "IDs" => "ids"

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Fix on a small capitalisation issue on JSON output of the GameServer CRD
changes.

Work on #1033

None


markmandel:
> Can you please update the swagger file as well: https://github.com/googleforgames/agones/blob/master/sdks/swagger/sdk.swagger.json#L348

Ah, excellent catch. Will do. We should probably have a test for this.

markmandel:
So I ran the generation grpc & swagger, and it looks like this value doesn't touch the SDK, so there are no changes to the swagger.json -- it's only a CRD storage/retrieval change.

pooneh-m:
Should the IDs in the proto be updated as well to conform with the [naming conventions](https://developers.google.com/protocol-buffers/docs/style)?
https://github.com/googleforgames/agones/blob/master/proto/sdk/sdk.proto#L151

markmandel:
> Should the IDs in the proto be updated as well to conform with the [naming conventions](https://developers.google.com/protocol-buffers/docs/style)?
> https://github.com/googleforgames/agones/blob/master/proto/sdk/sdk.proto#L151

Ooh! Yeah, I think they should. I'll do a separate PR with this change.

PR #1561, opened 2020-05-15
Fleet Aggregate Player Tracking Logic

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Implementation of aggregation of Player Tracking values at the Fleet and GameServerSet levels.

Work on #1033

Includes both unit and e2e tests.

This completes the logic for player tracking :raised_hands: 

Docs are the final piece left :muscle:


pkg/gameserversets/controller.go
```diff
@@ -606,5 +606,21 @@ func computeStatus(list []*agonesv1.GameServer) agonesv1.GameServerSetStatus {
 		}
 	}
 
+	if runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		// to make this code simpler, we will loop around the gs list twice.
+		status.Players = &agonesv1.AggregatedPlayerStatus{}
+		for _, gs := range list {
+			if gs.ObjectMeta.DeletionTimestamp.IsZero() &&
+				(gs.Status.State == agonesv1.GameServerStateReady ||
```
pooneh-m, line 614: If a GS is ready or reserved, does it still have any player?

markmandel: Good question :thinking: 

In theory it _might_, say leftover from a previous session, and the GameServer has gone back into a pool, and the players are still connected.

I figured it made sense to err on the side of more flexible.

pkg/fleets/controller.go
```diff
@@ -531,6 +531,17 @@ func (c *Controller) updateFleetStatus(fleet *agonesv1.Fleet) error {
 		fCopy.Status.ReservedReplicas += gsSet.Status.ReservedReplicas
 		fCopy.Status.AllocatedReplicas += gsSet.Status.AllocatedReplicas
 	}
+	if runtime.FeatureEnabled(runtime.FeaturePlayerTracking) {
+		// to make this code simpler, we will loop around the gsSet list twice.
```
roberthbailey, line 535: simpler while the feature gate is in place? Do you want to add a TODO to consolidate this with the above loop when the feature graduates to ga?

markmandel: > simpler while the feature gate is in place? 

Yeah exactly. Let me be explicit about this. Excellent idea on the TODO, doing that now.

PR #1568, opened 2020-05-18
Site: Fix publish issue with date update

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The publishDate will default to the date, so when the date is in the future, the entire page won't render, which then breaks links and the website cannot publish from master.

None.

None.


site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: "Allocator Service"
 date: 2020-05-19T05:45:05Z
+publishDate: 2019-10-25T05:45:05Z
```
markmandel, line 4: @pooneh-m you'll find this interesting :smile:

markmandel:
> It is hard to get that `publishDate` is actually needed here, probably we can add this into the docs describing website contributions.

Also have a TODO to run the `site-static` target during `site-test` to confirm that the production site build is tested during CI.

PR #1571, opened 2020-05-19
Portallocator - DeAllocate method refactoring

akremsa:
**What type of PR is this?**
> /kind cleanup

I would like to discuss this change. Not sure why an anonymous function is stored in a _found_ variable and there are 2 locks - 1st inside this function and 2nd before _gs.Spec.Ports_ loop.
Was it made on purpose?

What if we use a single lock statement at the beginning of the function and remove the anonymous one? I've written a couple of benchmarks to make sure that this change doesn't make things worse. 
Example of a benchmark (didn't add it to this commit):
```
func BenchmarkDeallocate(b *testing.B) {
	for x := 0; x < b.N; x++ {
		pa.Allocate(gs)
		pa.DeAllocate(gs)
	}
}
```

Before:
cpu profile (GS was preallocated):
image

After:
cpu profile (GS was preallocated):
image

Yes, it's not a significant improvement but at the end of the day, code looks a bit easier to understand. What do you think?

### DIFF
--- a/pkg/gameservers/portallocator.go
+++ b/pkg/gameservers/portallocator.go
@@ -178,25 +178,16 @@ func (pa *PortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.GameServer 
 
 // DeAllocate marks the given port as no longer allocated
 func (pa *PortAllocator) DeAllocate(gs *agonesv1.GameServer) {
+	pa.mutex.Lock()
+	defer pa.mutex.Unlock()
+
 	// skip if it wasn't previously allocated
-
-	found := func() bool {
-		pa.mutex.RLock()
-		defer pa.mutex.RUnlock()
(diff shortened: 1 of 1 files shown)


markmandel:
So the reason for this was such that DeAllocate could first grab a Read only Lock (which multiple processes could still use) while it was checking to see if the port was previously allocated, and then if found, it could THEN get the Full lock, which would lock the port allocation structure from being accessed during the operation.

It's probably a micro-optimisation, but it was aimed at allowing more concurrent throughput, than single operation optimisation.

So... your change looks interesting, but the benchmark may not be valid for this use case, since it's only looking at the performance of this one function, not all of them working in concert concurrently.

akremsa:
Ok, I get the idea, thanks for the explanation. In a current implementation, RLock is used to wrap a single quick operation - get data by a key from a map ("if" statement can be avoided actually) that is why I decided to discuss this change. 
I will close this PR.

PR #1576, opened 2020-05-20
Add support for internal game server ports

### LINKED ISSUE #749: Allow no ports for GameServer
We would like to leverage company's API gateway to provide https termination, authentication, rate limit, ip filter and more in front of game server which uses websocket (probably quic later) for communication. Our K8s cluster is private GKE, and has not public ip for nodes, and all traffic will route from API gateway.

Currently, ports expose game server using node ip and ports. The K8s cluster works well but as it does not use node's port for traffic, ports configuration is a little annoying. It would be better if game server can config without ports (no hostport too).

theminecoder:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This PR adds support for specifying internal only ports, which don't get assigned a hostPort by the controller. This is useful for games that really on a custom proxy solution that sits in front of the game servers (minecraft servers are a big case for this kind of setup) and having the container port expose could open a security hole (eg: in network scenarios, minecraft authentication is happen on the proxy level, not the individual game server).

Closes #749 

As stated in my last pr, I'm new to go in general so if I need to change anything to fit better with go semantics, let me know :)

### DIFF
--- a/install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
@@ -64,17 +64,20 @@ properties:
             portPolicy:
               title: the port policy that will be applied to the game server
               description: |
-                  portPolicy has three options:
+                  portPolicy has four options:
                   - "Dynamic" (default) the system allocates a random free hostPort for the gameserver, for game clients to connect to
                   - "Static", user defines the hostPort that the game client will connect to. Then onus is on the user to ensure that the
                   port is available. When static is the policy specified, `hostPort` is required to be populated
                   - "Passthrough" dynamically sets the `containerPort` to the same value as the dynamically selected hostPort.
                   This will mean that users will need to lookup what port has been opened through the server side SDK.
+                  - "Internal" means that the system won't allocate a hostPort for this port. The onus is on the user to find a way
+                  to connect players to that port, whether it be through a custom proxy or otherwise.
               type: string
               enum:
               - Dynamic
               - Static
               - Passthrough
+              - Internal
             protocol:
               title: Protocol being used. Defaults to UDP. TCP is the only other option
               type: string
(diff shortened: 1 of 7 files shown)


theminecoder:
> cyclomatic complexity 31 of func `(*GameServerSpec).Validate` is high (> 30)

Not sure the best way to handle this. I assume its related to me adding the internal check option to the containerPort check option, which I can't think of a better way to do without duplicating the entire check.

aLekSer:
The function is too long:
`cyclomatic complexity 31`, this is not your fault. You can use `//nolint:` to supress this error message.
First of all there is a helm template here:
https://github.com/googleforgames/agones/blob/bcde5c59febcf2eefef9d6c6969f25b1e4d1b1a5/install/helm/agones/templates/crds/_gameserverspecvalidation.yaml#L64
```
              - Dynamic
              - Static
              - Passthrough
```
You miss changes in this file.

markmandel:
Again, this would have been good as a design ticket first, as per our [contributing guidelines](https://github.com/googleforgames/agones/blob/master/CONTRIBUTING.md#submitting-code-via-pull-requests).

I have a few questions about the overall design of this:

- Would being able to define no ports have been a simpler solution? What would would be the downsides to this approach?
- What's the value add of having a port defined that's not exposed?
- If we do add Internal, is the extra complexity worth it, vs providing the ability to have no ports exposed?
- "Internal" seems like an odd name choice to me (personally). Is there something better? "None"?

theminecoder:
Will convert to a design ticket once I get a chance to play with the 0 port spec idea, however I do fancy the idea to still be able to spec my ports on the object for completeness sake. I have a feeling also that the fix in https://github.com/googleforgames/agones/pull/1576/files#diff-310ad1666561c0a99ea0effb347aea2bR124 may still be required however, as currently without it the controller locks up in the tests when trying to assign 0 non internal ports.

markmandel:
Checking in here, to see if this work is still ongoing? I may end up closing this PR just to keep things clean, unless there are objections?

theminecoder:
No objections here.

markmandel:
Thanks for taking a stab at thing though! Would definitely still love your future input!

Issue #1557, opened 2020-05-15
Fleet and GameServerSet: exponential backoff could be used or GameServers should be moved to Error and stop creating new, when PodTemplateSpec is invalid


aLekSer:
**What happened**:
Apply a fleet with invalid PodTemplateSpec. 
Example config and `kubectl describe gss` is here:
https://gist.github.com/aLekSer/7ac97a66cfc7369b771a5f47b4f58f1c
Fleet with 3 Replicas, note the last line:
```
  Normal  SuccessfulCreate  2m6s (x1269763 over 25h)  gameserverset-controller  (combined from similar events): Created gameserver: simple-fleet-7bbw7-xzchx-rt88q
```
This Fleet with wrong config generates about `1269763/ 25 /3600 ≈ 14.1` or about 7 GameServer create events (if counted with delete events).

Consequences:

1. Next commands hang:
```
kubectl get events
kubectl describe componentstatus 
```
2. New `GameServers` could not be created, as I can see from logs 30 seconds is not enough to start listening on GRPC endpoint - thus `NewSDK()` function crashes:
```
2020/05/15 13:05:21 Creating SDK instance
2020-05-15T13:05:21 Created GameServer sidecar "timestamp"
2020/05/15 13:05:51 Could not connect to sdk: could not connect to localhost:9357: context deadline exceeded\n",
2020-05-15T13:06:21 "Starting SDKServer grpc service...", 
```
More details on similar GameServer Pod:
https://gist.github.com/aLekSer/e9103a44ef59b8b9c260f7eb651ee83a

3. Could not use Monitoring - `kubectl port-forward` returns
```
error: error upgrading connection: error dialing backend: ssh: rejected: connect failed (Connection timed out)            
```
`make setup-grafana` hangs 

**What you expected to happen**:
There should be a way to make sure wrong Fleet configurations do not decrease the service quality for good Fleet configurations.
The rate of creating `GameServers` with wrong Spec should be decreased over time by means of exponential backoff.
There is an option to use approach of `ReplicaSet` or HPA.

**How to reproduce it (as minimally and precisely as possible)**:
`kubectl apply -f` next config:
https://gist.github.com/aLekSer/7ac97a66cfc7369b771a5f47b4f58f1c
Wait some time, see that Kubernetes API becomes slower, for example `kubectl get events` is considerably slower.
Also I will gather some `PodTemplateSpec` misconfigurations here https://github.com/googleforgames/agones/issues/1298#issuecomment-601743662

[...]

PR #1578, opened 2020-05-20
Validate GameServer PodTemplateSpec

### LINKED ISSUE #1298: GameServerSpec: Add PodTemplateSpec validation
There were 2 validation of PodTemplateSpec added recently - for Labels and Annotations. This approach could be generalised for all fields in a PodTemplateSpec.
This could run system into a failure state without informing user that he choose wrong pod template.
For instance fleet mentioned below with resources requests.memory > limits.memory is a mistake and no pod could be created from such template. 
Fleet would start to fire up new GameServers infinitely as in aforementioned bugs with Labels.

Use a standard Validation logic as could be found here for a ReplicaSet:
https://github.com/kubernetes/kubernetes/blob/52d7614a8ca5b8aebc45333b6dc8fbf86a5e7ddf/pkg/apis/apps/validation/validation.go#L674

**Additional context**
Fleet which does not signalise of a failure config:
```
apiVersion: "agones.dev/v1"
kind: Fleet
metadata:
  name: simple-udp-failure
spec:
  replicas: 2
  template:
    spec:
      ports:
      - name: default
        containerPort: 7654
      template:
        spec:
          containers:
          - name: simple-udp
            image: gcr.io/agones-images/udp-server:0.17
            resources:
              requests:
                memory: "66Mi"
                cpu: "20m"
              limits:
                memory: "64Mi"
                cpu: "20m"
```

danieloliveira079:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The text below tries to summarise what this PR is delivering.
As per suggested by Alex on https://github.com/googleforgames/agones/issues/1298 there is the possibility of using Kubernetes validation packages. 

I have been exploring this approach and below one can find my findings and conclusions. I would like to know if that is what is expected in terms of implementation and possible solution for the issue reported on https://github.com/googleforgames/agones/issues/1557.

Unfortunately, adding the Kubernetes dependencies required a few changes that were out of the scope of this PR. 

### Added Kubernetes dependencies:
```go
// pkg/apis/agones/v1/gameserver.go
import (
	k8s_api_v1 "k8s.io/kubernetes/pkg/apis/core/v1"
	k8s_validation "k8s.io/kubernetes/pkg/apis/core/validation"
)
```

  - Convert_v1_PodTemplateSpec_To_core_PodTemplateSpec: The conversion is required and the resulting object is passed to the `ValidPodTemplateSpec` method.

  - ValidatePodTemplateSpec: Validates the PodTemplateSpec using Kubernetes validation package.

- Required change `make test-gen-api-docs` script: 
  - Generator: https://github.com/ahmetb/gen-crd-api-reference-docs
  - Dependency: https://github.com/kubernetes/gengo
  - Reason: https://github.com/kubernetes/gengo/blob/master/parser/parse.go#L581
  - Without those changes the following error is thrown:
        ```bash
        go: k8s.io/kubernetes@v1.15.11 requires
            k8s.io/api@v0.0.0: reading k8s.io/api/go.mod at revision v0.0.0: unknown revision v0.0.0
        ```

### Findings: 

If the decision about using Kubernetes Valdiation is moving forward the list below should be taken into consideration

1) Validation and required fields
  - When using `ValidatePodTemplateSpec` there is a minimal list of fields that must be present in order to satisfy the validation:
    - spec.containers:
      - TerminationMessagePolicy: FieldValueRequired
      - ImagePullPolicy: FieldValueRequired
    - spec:
      - RestartPolicy: FieldValueRequired
      - DNSPolicy: FieldValueRequired
        - None: Requires DNSConfig

Those fields must be either part of the manifest or added somewhere during the admission lifecycle.

This PR adds a [applySpecDefaults](https://github.com/danieloliveira079/agones/blob/bug/validate_pod_template/pkg/apis/agones/v1/gameserver.go#L276) method and extends the [applyContainerDefaults](https://github.com/danieloliveira079/agones/blob/bug/validate_pod_template/pkg/apis/agones/v1/gameserver.go#L304) to satisfy those requirements

2) Validating from the GameServer Controller (creationMutationHandler) 
- Calls `gs.ApplyDefaults()` but at this point the Fleet has been pushed already. Calling the validation here may not be effective. 
- https://github.com/googleforgames/agones/blob/master/pkg/gameservers/controller.go#L233

3) Fleets Controller (creationValidationHandler) 
- calls `fleet.Validate()` but does not fulfil GameServer PodTemplateSpec with Defaults.
- https://github.com/googleforgames/agones/blob/master/pkg/fleets/controller.go#L180
 - If all the minimal required fields are not present the validation will fail. Therefore, it must be informed by the user (manifest) since it is not clear that those fields are required
- Example from GitHub does not include those https://raw.githubusercontent.com/googleforgames/agones/release-1.5.0/examples/simple-udp/fleet.yaml
- Consider calling `gs.ApplyDefaults()` from the `creationValidationHandler` method.

4) GameServerSet 
- It calls `validateGSSpec` and consequently apply defaults
- https://github.com/googleforgames/agones/blob/master/pkg/apis/agones/v1/gameserverset.go#L105
- https://github.com/googleforgames/agones/blob/master/pkg/apis/agones/v1/common.go#L76

Conclusion

- Using the Kubernetes Validation package requires some changes in the logic of the Fleet Controller
- A refactoring is required for the `gs.Validate()` since some of the validations will be handled by the Kubernetes package
- https://github.com/googleforgames/agones/blob/master/pkg/apis/agones/v1/gameserver.go#L354

## References

This is the list of validations handled by the `ValidatePodTemplateSpec` provided by Kubernetes:

```
# List of validations

- Labels
- Annotations
- Pod Specific Annotation
  - Tolerations
  - Seccomp
  - AppArmor
- PodSpec
  - Volumes
  - Containers
  - InitContainers
  - RestartPolicy
  - DNSPolicy
  - NodeSelector Labels
  - Pod Security Context
  - ImagePullSecrets
  - Affinity
  - Pod DNS Config
  - Readness Gates
  - Service Account Name
  - Node Name
  - Active Deadline
  - Tolerations
  - Host Name and Subdomain
  - Host Aliases
  - Priority Class Name
  - Runtime Class Name
  - Preemption Policy
```

Closes #1298


aLekSer:
Linter has failed. Please perform:
```
goimports -w pkg/apis/agones/v1/gameserver_test.go
```

go.mod
```diff
@@ -44,9 +43,33 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0-20170531160350-a96e63847dc3
 	k8s.io/api v0.15.11
-	k8s.io/apiextensions-apiserver v0.0.0-20200318010201-8546efc3bc75 // kubernetes-1.15.11
+	k8s.io/apiextensions-apiserver v0.0.0 // kubernetes-1.15.11
```
aLekSer, line 46: This one version is non existent. And could lead to a problem.

danieloliveira079: good catch

pkg/apis/agones/v1/gameserver.go
```diff
@@ -461,25 +495,44 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
...
-			for _, err := range validationErrors {
-				causes = append(causes, metav1.StatusCause{
-					Type:    metav1.CauseTypeFieldValueInvalid,
-					Field:   "container",
-					Message: err.Error(),
-				})
-			}
-		}
 	}
-	objMetaCauses := validateObjectMeta(&gss.Template.ObjectMeta)
```
danieloliveira079, line 475: The validation of the ObjectMeta field is handled by `k8s_validation.ValidatePodTemplateSpec` method.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -498,16 +551,6 @@ func ValidateResource(request resource.Quantity, limit resource.Quantity, resour
 	return validationErrors
 }
 
-// validateResources validate CPU and Memory resources
-func validateResources(container corev1.Container) []error {
```
danieloliveira079, line 502: The validation of resources will be handled by `k8s_validation.ValidatePodTemplateSpec` method.

go.mod
```diff
@@ -44,9 +43,33 @@ require (
...
-	k8s.io/apiextensions-apiserver v0.0.0-20200318010201-8546efc3bc75 // kubernetes-1.15.11
+	k8s.io/apiextensions-apiserver v0.15.11
 	k8s.io/apimachinery v0.15.11
 	k8s.io/client-go v0.15.11
 	k8s.io/kube-openapi v0.0.0-20190709113604-33be087ad058 // indirect
+	k8s.io/kubernetes v1.15.11
 	k8s.io/utils v0.0.0-20200124190032-861946025e34 // indirect
 )
+
+replace (
```
danieloliveira079, line 54: This is required due to the fact that `k8s.io/kubernetes v1.15.11` will require `v0.0.0` of those packages. The require block will enforce the right version.

danieloliveira079:
Using the template provided by @aLekSer https://github.com/googleforgames/agones/issues/1298
```yaml
apiVersion: "agones.dev/v1"
kind: Fleet
metadata:
  name: simple-udp-failure
spec:
  replicas: 2
  template:
    spec:
      ports:
      - name: default
        containerPort: 7654
      template:
        spec:
          containers:
          - name: simple-udp
            image: gcr.io/agones-images/udp-server:0.17
            resources:
              requests:
                memory: "66Mi"
                cpu: "20m"
              limits:
                memory: "64Mi"
                cpu: "20m"
```
The output is:
```bash
 kubectl create -f fleet.yaml 
The Fleet "" is invalid: spec.spec.containers[0].resources.requests: spec.spec.containers[0].resources.requests: Invalid value: "66Mi": must be less than or equal to memory limit
```

pkg/apis/agones/v1/gameserver.go
```diff
@@ -259,13 +264,29 @@ func (gs *GameServer) ApplyDefaults() {
...
+	if gss.Template.Spec.RestartPolicy == "" {
+		// TODO: Check if that is what the Agones team is expecting.
+		// Options: Always, OnFailure or Never https://pkg.go.dev/k8s.io/api/core/v1@v0.15.11?tab=doc#RestartPolicy
+		gss.Template.Spec.RestartPolicy = corev1.RestartPolicyOnFailure
+	}
+
+	if gss.Template.Spec.DNSPolicy == "" {
+		// TODO: Check if that is what the Agones team is expecting.
+		// Options: ClusterFirstWithHostNet, ClusterFirst, Default or None https://pkg.go.dev/k8s.io/api/core/v1@v0.15.11?tab=doc#DNSPolicy
+		gss.Template.Spec.DNSPolicy = corev1.DNSDefault
```
danieloliveira079, line 286: Which DNSPolicy is expected to be used by default?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -284,6 +305,18 @@ func (gss *GameServerSpec) applyContainerDefaults() {
...
+
+	for idx, c := range gss.Template.Spec.Containers {
+		if c.TerminationMessagePolicy == "" {
+			gss.Template.Spec.Containers[idx].TerminationMessagePolicy = corev1.TerminationMessageFallbackToLogsOnError
+		}
+
+		if c.ImagePullPolicy == "" {
+			// TODO: Check if that is what the Agones team is expecting.
+			// Options: Always or IfNotPresent https://pkg.go.dev/k8s.io/api/core/v1@v0.15.11?tab=doc#PullPolicy
+			gss.Template.Spec.Containers[idx].ImagePullPolicy = corev1.PullIfNotPresent
```
danieloliveira079, line 317: Which ImagePullPolicy is expected to be used by default?

markmandel: Whatever the default is for Kubernetes 👍

pkg/apis/agones/v1/gameserver.go
markmandel: Whatever the default is for Kubernetes +1

pkg/apis/agones/v1/gameserver.go
```diff
@@ -259,13 +264,29 @@ func (gs *GameServer) ApplyDefaults() {
...
 	gss.applySchedulingDefaults()
 	gss.applySdkServerDefaults()
 }
 
+// applySpecDefaults applies the PodTemplateSpec defaults
+func (gss *GameServerSpec) applySpecDefaults() {
+	if gss.Template.Spec.RestartPolicy == "" {
+		// TODO: Check if that is what the Agones team is expecting.
+		// Options: Always, OnFailure or Never https://pkg.go.dev/k8s.io/api/core/v1@v0.15.11?tab=doc#RestartPolicy
+		gss.Template.Spec.RestartPolicy = corev1.RestartPolicyOnFailure
```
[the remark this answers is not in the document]
markmandel: Whatever the default is for Kubernetes :+1:  (I think It's Always?)

go.mod
markmandel: I might take this for a spin, and tweak it, and see what happens :smile: 

I wonder if we can shrink the number of dependencies and still make it work.

markmandel:
Can we rebase this down to 2 commits? 1 for the new code, and one for everything in vendor?

Will make this much easier to go through :smile:

pkg/apis/agones/v1/gameserver.go
```diff
@@ -461,25 +495,44 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
...
+	if len(podTemplateSpecCauses) > 0 {
+		causes = append(causes, podTemplateSpecCauses...)
 	}
 
 	return causes, len(causes) == 0
 }
 
+// ValidatePodTemplateSpec validates the PodTemplateSpec of a GameServer
+func ValidatePodTemplateSpec(spec *corev1.PodTemplateSpec) []metav1.StatusCause {
+	causes := []metav1.StatusCause{}
```
markmandel, line 510: We would likely want to put a feature flag here, but we can get to that later.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -461,25 +495,44 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
...
+
+	// Before calling the validation, the pod template must be converted to the Core version
+	outPodTemplateSpec := core.PodTemplateSpec{}
+	if err := k8s_api_v1.Convert_v1_PodTemplateSpec_To_core_PodTemplateSpec(spec, &outPodTemplateSpec, nil); err != nil {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeUnexpectedServerResponse, // TODO: Confirm the type to be returned
```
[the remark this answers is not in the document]
markmandel: Maybe CauseTypeFieldValueInvalid? Since if it can't convert, it's likely an invalid pod spec?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -461,25 +495,44 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
...
+			Message: err.Error(),
+		})
+
+		return causes
+	}
+
+	errList := k8s_validation.ValidatePodTemplateSpec(&outPodTemplateSpec, field.NewPath("spec"))
+	for _, err := range errList {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseType(err.Type),
```
markmandel, line 527: This should probably be an official type, I would probably use CauseTypeFieldValueInvalid here as well.

pkg/apis/agones/v1/gameserver.go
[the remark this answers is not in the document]
roberthbailey: It is: https://godoc.org/k8s.io/api/core/v1#RestartPolicy

I was hoping there was a constant like "DefaultRestartPolicy" but it looks like they only define the policies.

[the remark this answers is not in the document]
roberthbailey: Alternatively, if we just leave these blank can we rely on k8s to fill in the defaults (rather than filling them in ourselves)?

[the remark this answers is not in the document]
danieloliveira079: Unfortunately, this is not possible because the validation is happening before pushing the spec to K8S.

pkg/apis/agones/v1/gameserver.go
danieloliveira079: The problem is that it may pass back the wrong information. There will be situation where the err.Type will not be about invalid values. Instead it will be related to required fields. In this case we are justing sending back to the caller what K8S is returning from the validation process. Does that make sense?

markmandel: Aah, I see - I didn't pick up on the fact it was bringing through K8s types. Then yes, this makes total sense :+1:

go.mod
markmandel: Took it for a spin. Could not improve on this in any way :frowning:

markmandel:
This all looks good to me. Any objections - @roberthbailey , @pooneh-m , @aLekSer ?

I can only see this as a net positive for the users. 

Assuming we get consensus, I would recommend the following:

2. Let's get a feature flag for this, and have it be an Alpha feature, just in case it breaks people -- also, having a feature flag allows you to incrementally do work behind the feature flag without impacting the wider code base.
1. Seems like there is some refactoring to do, etc -- let's split up this work into smaller PR's. I think one massive PR is going to be difficult to review. I.e. breaking up each refactor into a separate PR, the vendoring of the new deps into its own, etc. 

How does that sound?

roberthbailey:
No objections from me.

pooneh-m:
This change looks good. However, even with the feature flag we need to know if it breaks any existing scenario by extensive testing. If it does, we need to either bump the Agones version to support the feature by default or make it backward compatible. I vote for the latter. If it is backward compatible as is, we don't need a feature flag.

aLekSer:
@pooneh-m Additional thought, this feature would just predict a future failure one step earlier (on `kubectl apply` step, not on the `CreatePod` by the `GameServer` controller step). So in terms of ability to create a `Ready` pod scenarios with and without this feature is equivalent.
 Because if the pod could not be validated it would lead into an `Error` state - thus this kind of `GameServers` with `PodTemplateSpec` which is invalid could not become Ready without this Feature also. Pod would not be created for these `GameServers`.

>However, even with the feature flag we need to know if it breaks any existing scenario by extensive testing.

pooneh-m:
> @pooneh-m Additional thought, this feature would just predict a future failure one step earlier (on `kubectl apply` step, not on the `CreatePod` by the `GameServer` controller step). So in terms of ability to create a `Ready` pod scenarios with and without this feature is equivalent.
> Because if the pod could not be validated it would lead into an `Error` state - thus this kind of `GameServers` with `PodTemplateSpec` which is invalid could not become Ready without this Feature also. Pod would not be created for these `GameServers`.

Thanks that's what I expected as well. This change is improving the prevention for an already failing state. Therefore, it should not be considered as a breaking change, unless the defaults that are set here are different than the k8s defaults. We should test all the default scenarios to make sure there is no change in the behavior.

markmandel:
> Using the Kubernetes validation package to validate Specs (feature flags or backward compatibility):
3.1 Option 1: Keep the validation logic within the same project's package as it is now. That means just changing the logic/code inline.

I think keeping it inline is fine, unless there is reason not to. That being said, we should still wrap it in a feature flag.

> Thanks that's what I expected as well. This change is improving the prevention for an already failing state. Therefore, it should not be considered as a breaking change, unless the defaults that are set here are different than the k8s defaults. We should test all the default scenarios to make sure there is no change in the behavior.

100% agree. There _should_ be no change in the behaviour, but I figure better to wrap it in a feature flag just in case there's an edge case we haven't considered.

danieloliveira079:
Tks folks for all the feedback. I will check how the feature flag can be integrated into the logic and possibly create new PR. Is that ok if I keep this one open or should I make it a draft again?

aLekSer:
It is ok to keep it open. You can add `FeatureFlag` inside this PR if you want to.

markmandel:
I'm probably inclined to say to close this one for historical reference, and then start the several smaller PRs based on this work.

That way there is less noise amongst the signal, and it will make it easier to us to review as well.

markmandel:
Checking in here, to see if this work is still ongoing? I may end up closing this PR just to keep things clean, unless there are objections?

danieloliveira079:
> Checking in here, to see if this work is still ongoing? I may end up closing this PR just to keep things clean, unless there are objections?

Agreed @markmandel. Future work can use this PR just for reference. Feel free to close it.

aLekSer:
I like this PR, without FeatureGate it is ready enough to be merged. I assume that this one would soon be handled.

markmandel:
Cool - closing this PR.

Looking forward to the future work :+1:

PR #1581, opened 2020-05-21
Fix the Unreal Plugin's GetGameServer

dotcom:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The fix is extremely simple.

**first commit :**
I think the point where `OnProcessRequestComplete().BindLambda()` after `SendRequest()` is done needs to be fixed.(In `GetGameServer()`)
https://github.com/googleforgames/agones/compare/master...dotcom:master?diff=split#diff-80dbe36c10c363357ae5c22ced334ea2L151

For implementers who want to add more processing to the Request, I have split the processing.

**second commit:**
We should check whether the request is acceptable or not with `EHttpResponseCodes::IsOk`

> Once review has occurred, please rebase your PR down to a single commit. This will ensure a nice clean Git history.

I'm separating the commits for two different fixes, but squash them as needed.

I'd like further comments from people who are familiar with UE4.

Thank you.

### DIFF
--- a/sdks/unreal/Agones/Source/Agones/Private/AgonesHook.cpp
+++ b/sdks/unreal/Agones/Source/Agones/Private/AgonesHook.cpp
@@ -105,18 +105,21 @@ bool FAgonesHook::IsTickableWhenPaused() const
 
 void FAgonesHook::Ready()
 {
-	SendRequest(SidecarAddress + ReadySuffix, TEXT("{}"), FHttpVerb::POST, true);
+	TSharedRef<IHttpRequest> Req = MakeRequest(SidecarAddress + ReadySuffix, TEXT("{}"), FHttpVerb::POST, true);
+	SendRequest(Req);
 }
 
 void FAgonesHook::Health()
 {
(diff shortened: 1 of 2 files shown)


WVerlaek:
LGTM, interesting that the order of BindLambda needed to change, I don't think we noticed any issues there. But the changes look sensible, especially checking the response code for errors

dotcom:
@WVerlaek  
Thank you for checking.
BindLambda after SendRequest seems to register the callback after execution, but I think it's safe when HttpThread works as a FSingleThreadRunnable. (This would be handled when Tick in the HttpThread is called.) However, in the case of FRunnable, I think this is not safe. (Please correct me if I'm wrong.)
But I'm not sure where it actually has a branch to treat HttpThread as FSingleThreadRunnable or FRunnable.

roberthbailey:
@markmandel - how do you feel about squeezing in this bug fix right before the release cut?

PR #1585, opened 2020-05-26
Metrics: add namespace to distinguish fleets with the same name

### LINKED ISSUE #1501: Metrics: add namespace to distinguish same name fleets
Metrics summed up for fleets in a different namespaces, should be splitted.

**What happened**:
Created two namesake fleets with name "fleet-example".
All count metrics got summed up for the same names:

Grafana

Have a selector, or use key as "namespace/fleet" for all metrics which are using `fleetKey`

```
$ kubectl create namespace xbox
$ helm install --set "gameservers.namespaces={default,xbox}" --namespace agones-system --name my-release agones/agones
$ kubectl apply -f ./examples/fleet.yaml
$ kubectl apply -f ./examples/fleet.yaml -n xbox
```
Now we have 2 fleets with 2 GSs each.
But Grafana show one fleet with 4 GS.

**Environment**:
- Agones version: 1.5
- Kubernetes version (use `kubectl version`): 1.15
- Others:

akremsa:
**What type of PR is this?**
> /kind bug

Fleet metrics from different namespaces are split. 

Closes #1501

Screenshot 2020-05-26 at 18 10 11

gameservers_count:
image
Prometheus:
image


aLekSer:
Seems that github breaks website check again:
```
 Non-OK status: 429 --- site/docs/installation/creating-cluster/aks/index.html --> https://github.com/dgkanatsios/AksNodePublicIPController
site/docs/installation/creating-cluster/minikube/index.html
  Non-OK status: 429 --- site/docs/installation/creating-cluster/minikube/index.html --> https://github.com/kubernetes/minikube#requirements
✘✘✘ failed in 45.759002885s
133 errors in 109 documents
```

aLekSer:
@akremsa Please include `metrics/controller_test.go` which is failing:
```

    controller_test.go:71: 
        	Error Trace:	controller_test.go:71
        	            				controller_test.go:250
        	Error:      	Should be true
        	Test:       	TestControllerFleetAutoScalerState
        	Messages:   	no TimeSeries found with labels: [first-fleet name-switch default]
```

markmandel:
> Html Test: Attempt {1..5}

It's so weird. Locally, it will do 5 retries (shell sees `$i` as 1, then 2 then 3, etc). On Cloud Build, it sees $i as {1..5} - I'm not sure why.

pkg/metrics/gameservers_count.go
```diff
@@ -52,18 +52,24 @@ func (c GameServerCount) record(gameservers []*agonesv1.GameServer) error {
...
-	for _, g := range gameservers {
-		c.increment(g.Labels[agonesv1.FleetNameLabel], g.Status.State)
+
+	var namespace string
+	if len(gameservers) != 0 {
+		// counts gameserver per state and fleet
+		for _, g := range gameservers {
+			c.increment(g.Labels[agonesv1.FleetNameLabel], g.Status.State)
+		}
+		namespace = gameservers[0].GetNamespace()
```
aLekSer, line 62: We should get namespace here per gameserver

akremsa:
@aLekSer is right. I was checking gameservers_count metric and found that gs amount is different for each namespace. Fixed in the latest commit. @markmandel  please check.

pkg/metrics/controller_test.go
```diff
@@ -69,7 +70,8 @@ func assertMetricData(t *testing.T, exporter *metricExporter, metricName string,
 		}
 		e, ok := expectedValuesAsMap[serialize(actualLabelValues)]
 		assert.True(t, ok, "no TimeSeries found with labels: %v", actualLabelValues)
-
+		fmt.Println("ACT", actualLabelValues)
```
aLekSer, line 73: There is no need to print this or we could use `t.Log`

pkg/metrics/gameservers_count.go
```diff
@@ -24,16 +24,23 @@ import (
...
+
+type fleetKey struct {
+	name      string
+	namespace string
+}
 
 // increment adds the count of gameservers for a given fleetName and state
-func (c GameServerCount) increment(fleetName string, state agonesv1.GameServerState) {
+func (c GameServerCount) increment(fleetName, fleetNamespace string, state agonesv1.GameServerState) {
+	key := fleetKey{name: fleetName, namespace: fleetNamespace}
```
aLekSer, line 36: Cool to use a struct as a key. 👍

pkg/metrics/controller_test.go
akremsa: It was used for a debug purpose, I will remove it.

pkg/metrics/gameservers_count.go
akremsa: Let's cover a whole gameserver_countfile with tests in a separate PR.

PR #1591, opened 2020-05-28
Remove redundant "helm test" pod from "install.yaml"

aLekSer:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Helm test template would create an additional pod in an error state. 
Add new Helm variable to avoid creating new pod.

Thanks to Davit Mkrtchyan for pointing on this issue.

### DIFF
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -146,10 +146,11 @@ agones:
       name: agones-allocator
       pullPolicy: IfNotPresent
 
-
 gameservers:
   namespaces:
   - default
   minPort: 7000
   maxPort: 8000
 
+helm:
+  installTests: false
--- a/install/helm/agones/templates/tests/test-runner.yaml
+++ b/install/helm/agones/templates/tests/test-runner.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.helm.installTests }}
 # Copyright 2020 Google LLC All Rights Reserved.
 #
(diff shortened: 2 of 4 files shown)


install/helm/agones/values.yaml
```diff
@@ -146,10 +146,11 @@ agones:
...
       pullPolicy: IfNotPresent
 
-
 gameservers:
   namespaces:
   - default
   minPort: 7000
   maxPort: 8000
 
+helm:
```
markmandel, line 155: Question - should this be on by default? Or should it be off?

I would have thought "off" would be better? Or no?

aLekSer: Generally this PR lacks of changes to `helm.md`. There is a statement about use `helm test agones --cleanup` as a simple test of your installation.

markmandel: Right, but should we be installing this test pod by default on production deployments by default?

This seems to me like this is something people should opt into, rather than opt out. I didn't realise that the helm test would be installing a test pod on each cluster -- or maybe I'm missing something?

roberthbailey: +1 for making it false by default, to match what this is changing in install.yaml.

aLekSer: Agree, I have updated this PR and subj.

PR #1596, opened 2020-05-29
C# SDK Cleanup & Nuget Package

### LINKED ISSUE #1583: HealthCheckLoop Never invoked in C# SDK
**What happened**:
Agones Game Server becomes Unhealthy because Health is never signaled by the SDK.
HealthCheckLoop function is never invoked in Agones Game Server SDK for C#

HealthCheckLoop function should be invoked by the SDK in [AgonesSDK.cs](https://github.com/googleforgames/agones/blob/release-1.5.0/sdks/csharp/sdk/AgonesSDK.cs) and since it being private there's no way to call it externally.

```
		private async Task HealthCheckLoop()
		{
			while(!cts.IsCancellationRequested)
			{
				if(HealthEnabled)
					await HealthAsync();
				await Task.Delay(HealthInterval * 1000);
			}
		}
```

Use the AgonesSDK for C#

**Environment**:
- Agones version: SDK 1.0.2
- Install method (yaml/helm):
- Others:

rcreasey:
**What type of PR is this?**
/kind bug
/kind cleanup
/kind documentation
/kind feature

Updates the C# SDK with the following:
- Adds stubs for Nuget package publishing.
- Updates the gRPC requirement to `2.28` due to timeout issues.
- Adds an Interface to easily implement mock testing.
- Removes the "baked-in" health checks and requires explicit `Health` pings on instanced clients.
- Changes console output to use Microsoft logging classes.
- Ran the code through a linter.

Closes #1583

This will still require additional work on the Google side to provide necessary credentials to get the nuget package pushed to official repositories.


rcreasey:
@googlebot I signed it!

sdks/csharp/sdk/AgonesSDK.nuspec
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package >
+  <metadata>
+    <id>AgonesSDK</id>
+    <version>1.0.4</version>
```
rcreasey, line 5: This was arbitrarily picked, feel free to pick a starting version.

rcreasey:
To build this as a nuget package the command is straight forward:
```
dotnet pack csharp-sdk.csproj --configuration Release --include-source --output ./pkgs /p:Version=1.0.4
dotnet nuget push /Users/rcreasey/projects/public/agones/sdks/csharp/sdk/pkgs/AgonesSDK.1.0.4.symbols.nupkg <nuget repository credentials>
```

I've tested this internally on our private Nuget repositories and have been running sdk integrated game servers for the last two weeks.

sdks/csharp/sdk/AgonesSDK.nuspec
markmandel: Should align with next Agones version - so 1.7.0

markmandel:
Looking at the logs, it looks like `make run-sdk-command COMMAND=test SDK_FOLDER=csharp` is failing (this might take a while to build, just letting you know)

Something in the SDK tests isn't working.

markmandel:
A couple of quick questions:

Just to confirm, is this a breaking change in the SDK? I.e will old integrations still work with this when they upgrade, or will people need to change the code?

> Updates the gRPC requirement to 2.28 due to timeout issues.

Can you expand more on this? Maybe this is something we should upgrade across the board.

markmandel:
> Updates the gRPC requirement to 2.28 due to timeout issues.

Are there any issues with the generated code output? Doesn't look like there are, as the generated code wasn't touched, but just wanted to be sure.

rcreasey:
There should be no breaking changes to the gRPC update; there was an issue with our unit tests starting before the agones api endpoint was available (we're packing the [dev server](https://agones.dev/site/docs/guides/client-sdks/local/) as a fixture and delayed the startup to test) and the stream connection wasn't attempting a retry properly.  Instead, it was failing with:

```
System.Net.Internals.SocketExceptionFactory+ExtendedSocketException (111): Connection refused 127.0.0.1:9357
```

Updating it to the `2.28` grpc-dotnet release seemed to fix the issue.  

No agones code gen required.

sdks/csharp/sdk/AgonesSDK.nuspec
```diff
@@ -0,0 +1,23 @@
...
+    <projectUrl>https://agones.dev/</projectUrl>
+    <description>Game Server SDK for Agones, the open source project for hosting dedicated game servers on Kubernetes</description>
+    <tags>agones, googleforgames, kubernetes</tags>
+    <repository url="https://github.com/googleforgames/agones" />
+    <dependencies>
+      <group targetFramework=".NETStandard2.0">
+        <dependency id="Google.Api.CommonProtos" version="1.7.0" exclude="Build,Analyzers" />
+        <dependency id="Google.Protobuf" version="3.11.2" exclude="Build,Analyzers" />
+        <dependency id="Grpc" version="2.26.0" exclude="Build,Analyzers" />
+        <dependency id="Grpc.Core" version="2.26.0" exclude="Build,Analyzers" />
```
sbeca, line 19: Versions don't match updating dependencies in `csproj`

rcreasey: 👍

markmandel:
@pooneh-m since you are our resident C# expert, would you mind taking a look at these changes?

sdks/csharp/sdk/AgonesSDK.cs
```diff
@@ -11,60 +11,56 @@
...
 {
-	public sealed class AgonesSDK : IDisposable
+	public sealed class AgonesSDK : IAgonesSDK
 	{
 		public string Host { get; } = Environment.GetEnvironmentVariable("AGONES_SDK_GRPC_HOST") ?? "localhost";
-		public int Port { get; } = Convert.ToInt32((Environment.GetEnvironmentVariable("AGONES_SDK_GRPC_PORT") ?? "9357"));
+		public int Port { get; } = Convert.ToInt32(Environment.GetEnvironmentVariable("AGONES_SDK_GRPC_PORT") ?? "9357");
 
 		/// <summary>
 		/// The timeout for gRPC calls.
```
pooneh-m, line 30: While we are here, can the name and comment be updated to show that this is duration in seconds?

sdks/csharp/sdk/AgonesSDK.cs
```diff
@@ -128,34 +127,35 @@ public async Task<Status> ReserveAsync(long seconds)
...
-				return await client.GetGameServerAsync(new Empty(), deadline: DateTime.UtcNow.AddSeconds(RequestTimeout), cancellationToken: ctoken);
+				return await client.GetGameServerAsync(new Empty(),
+					deadline: DateTime.UtcNow.AddSeconds(RequestTimeout),
+					cancellationToken: ctoken);
 			}
 			catch (RpcException ex)
 			{
-				Console.Error.WriteLine(ex.Message);
-				//Should I rethrow the exception?
-				return null;
```
pooneh-m, line 158: This is change in the behavior. I agree that this is how it should have been handled originally, but may break a customer use-case that is already depending on this behavior.
@markmandel would it be ok to suddenly throw exception GetGameServerAsync()? How other SDKs are handling this?

pooneh-m:
Thanks @rcreasey for the changes and for the fix.

sdks/csharp/sdk/AgonesSDK.cs
markmandel: Yeah, I think it's reasonable to throw an exception here. I'm also assuming that in C#, exception throwing is the appropriate way to deal with errorrs?

For example, [Go returns an error](https://github.com/googleforgames/agones/blob/master/sdks/go/sdk.go#L74-L77)

The only maybe interesting part, is - do we want to throw the GRPC exception, or do we want to wrap it in our own exception type?

rcreasey: Ah, I forgot about this.  Yeah, I figured surfacing the exception would be better than just sinking it like it was previously doing.  Typically, libraries shouldn't do this and should surface it to the consuming callsite.

As for whether or not to make a custom AgonesException to wrap it in, I'll leave it up to you.  I'm fine with handling a gRPC exception in my code.

sdks/csharp/sdk/AgonesSDK.cs
```diff
@@ -11,60 +11,56 @@
...
+	public sealed class AgonesSDK : IAgonesSDK
 	{
 		public string Host { get; } = Environment.GetEnvironmentVariable("AGONES_SDK_GRPC_HOST") ?? "localhost";
-		public int Port { get; } = Convert.ToInt32((Environment.GetEnvironmentVariable("AGONES_SDK_GRPC_PORT") ?? "9357"));
+		public int Port { get; } = Convert.ToInt32(Environment.GetEnvironmentVariable("AGONES_SDK_GRPC_PORT") ?? "9357");
 
 		/// <summary>
 		/// The timeout for gRPC calls.
 		/// </summary>
 		public double RequestTimeout { get; set; }
```
rcreasey, line 32: Sure thing.  Is this sufficient?

```suggestion
		public double RequestTimeoutSec { get; set; }
```

sdks/csharp/sdk/AgonesSDK.cs
rcreasey: See below comment

sdks/csharp/sdk/AgonesSDK.cs
pooneh-m: Looks great! Thanks!

sdks/csharp/sdk/AgonesSDK.cs
pooneh-m: If there is no dependency to this SDK that this change would break them, it looks good to me. I don't think we need to wrap the exception.

PR #1598, opened 2020-05-30
Unity SDK: Added ability to Start / Stop healthchecks.

Davidnovarro:
There was no public method to start/stop health checks manually.

Now we have both options to start/stop health checks manually or automatically.

>/kind hotfix

### DIFF
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -38,14 +38,19 @@ namespace Agones
         [Range(0.01f, 5)] public float healthIntervalSecond = 5.0f;
 
         /// <summary>
-        /// Whether the server sends a health ping to the Agones sidecar.
+        /// Whether to start sending health ping to the Agones sidecar.
         /// </summary>
-        public bool healthEnabled = true;
+        [SerializeField] private bool startHealthChecks = true;
 
         /// <summary>
         /// Debug Logging Enabled. Debug logging for development of this Plugin.
         /// </summary>
         public bool logEnabled = false;
+
+        /// <summary>
+        /// Indicates whether health checks are currently started.
+        /// </summary>
+        public bool HealthIsStarted { get; private set; }
 
         private string sidecarAddress;
         private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
@@ -54,20 +59,26 @@ namespace Agones
         {
             public string key;
             public string value;
-            public KeyValueMessage(string k, string v) => (key, value) = (k, v);
+
+            public KeyValueMessage(string key, string value)
+            {
+                this.key = key;
+                this.value = value;
+            }
         }
 
         #region Unity Methods
         // Use this for initialization.
         private void Awake()
         {
-            String port = Environment.GetEnvironmentVariable("AGONES_SDK_HTTP_PORT");
+            string port = Environment.GetEnvironmentVariable("AGONES_SDK_HTTP_PORT");
             sidecarAddress = "http://localhost:" + (port ?? "9358");
         }
 
         private void Start()
         {
-            HealthCheckAsync();
+            if (startHealthChecks)
+                StartHealthChecks();
         }
 
         private void OnApplicationQuit()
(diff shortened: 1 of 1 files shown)


aLekSer:
Thanks for the PR, 

There is a ticket https://github.com/googleforgames/agones/issues/1583 about `HealthCheckLoop()` could this issue be fixed along with this PR?

sdks/unity/AgonesSdk.cs
```diff
@@ -38,36 +38,47 @@ public class AgonesSdk : MonoBehaviour
         [Range(0.01f, 5)] public float healthIntervalSecond = 5.0f;
 
         /// <summary>
-        /// Whether the server sends a health ping to the Agones sidecar.
+        /// Whether to start sending health ping to the Agones sidecar.
         /// </summary>
-        public bool healthEnabled = true;
+        [SerializeField] private bool startHealthChecks = true;
```
markmandel, line 43: Is there a good reason to change this name?

This would be a breaking change for the SDK, so we would like to avoid that where possible.

sdks/unity/AgonesSdk.cs
```diff
@@ -212,6 +223,26 @@ public async Task<bool> Reserve(TimeSpan duration)
             return await SendRequestAsync("/reserve", json).ContinueWith(task => task.Result.ok);
         }
 
+        /// <summary>
+        /// Start sending health ping to the Agones sidecar. 
+        /// </summary>
+        public void StartHealthChecks()
```
markmandel, line 229: Thanks for sending this through. I'm curious though - what's the use case for these functions?

Was there a particular scenario in which you would want to manually start the health check process, rather than have it be part of the component?

Also, when would you want to stop a health check?

sdks/unity/AgonesSdk.cs
markmandel: Also, changing the encapsulation would also be a breaking change. Is there a good reason to do that as well?

Davidnovarro: > Is there a good reason to change this name?
The name itself is misleading, for example, the user may set it to true externally an expect health checks to start, which won't work because the variable is **only** checked in the Start method, in case if it was set to true then health checks start.

> Also, changing the encapsulation would also be a breaking change. Is there a good reason to do that as well?

This variable was made public so unity will be able to serialize and show it int the Inspector it basically has no external use case, see code examples below.

```
void Start()
{
    sdk.healthEnabled = true; //There is a 50 % chance that the health checks will start due to Unity random scheduling. 
}

void Update()
{
    if(someCondition)
        sdk.healthEnabled = true; //This won't work. 

    if(someOtherCondition)
        sdk.healthEnabled = false; //This doesn't make sense because we can't renew the health checks, instead we can use sdk.Shutdown() 
}
```

sdks/unity/AgonesSdk.cs
Davidnovarro: > when would you want to stop a health check?

In my particular case, I do resue my game servers for a long period of time and they all have two external connections with Backend Servers (economy, rewards, etc..) and Party Servers (matchmaking, tickets, backfill).
In case if the GS loses the connection I mark it as non-healthy until the connection is restored.

There may be other use cases like the first match is complete but GS can't load the resources to continue.

> Was there a particular scenario in which you would want to manually start the health check process, rather than have it be part of the component?

Yes, like GS was able to restore the connection or load the resources or some background process was revived.

markmandel: > In case if the GS loses the connection I mark it as non-healthy until the connection is restored.

This sounds like a very dangerous approach -- I wouldn't advise it. Unhealthy game servers have failed a health check and are deemed to not be alive. If this GameServer was part of a Fleet, it would immediately be shut down and recreated.

> Yes, like GS was able to restore the connection or load the resources or some background process was revived.

I can't advocate for any of these scenarios.

This could be solved using custom labels for example, without abusing health checking abilities.

sdks/unity/AgonesSdk.cs
markmandel: To be honest, I'm not sure why you would be trying to dynamically decide if health checking is enable or not - you would enable health checking if you have it turned on on Agones. That's really it.

Health checking is not a tool for managing game server lifecycles.

sdks/unity/AgonesSdk.cs
Davidnovarro: Hmm, you are right, this PR can be removed.

markmandel:
Thanks for working through this with me though! Appreciate you taking the time 👍

PR #1601, opened 2020-06-01
Additional commands for prometheus and grafana

akremsa:
**What type of PR is this?**
> /kind cleanup

While I was working on prometheus\grafana-related tasks I found that it would be nice to add some new commands inside a Makefile.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -445,6 +445,10 @@ setup-prometheus:
 		$(HELM_ARGS) \
 		-f $(mount_path)/build/prometheus.yaml
 
+uninstall-prometheus: $(ensure-build-image)
+	$(DOCKER_RUN) \
+		helm delete --purge prom
+
 # setup grafana in the current cluster with datasource and dashboards ready for use with agones
 # by default Persistent Volume Claims are requested.
 setup-grafana: PVC ?= true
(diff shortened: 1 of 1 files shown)


build/Makefile
```diff
@@ -445,6 +445,14 @@ setup-prometheus:
 		$(HELM_ARGS) \
 		-f $(mount_path)/build/prometheus.yaml
 
+run-prometheus: $(ensure-build-image)
+	$(DOCKER_RUN) \
+		kubectl --namespace metrics port-forward $(shell kubectl get pods --namespace metrics -l 'app=prometheus,component=server' -o name) 9090
```
cyriltovena, line 450: Don't we have a portforward target already for Prometheus and Grafana ?

https://github.com/googleforgames/agones/blob/b134abbd4f2e4a9ab9fd57d3cadf19821e4fb7e3/build/README.md#make-grafana-portforward

akremsa: Ok, good. Didn't notice that.

aLekSer:
Some E2E tests are failing due to change in consul lock name in https://github.com/googleforgames/agones/pull/1611 . Previously it was `LockE2E`, now we need to have separate two locks, one per each E2E test cluster name.

aLekSer:
These two would be useful.

Issue #1154, opened 2019-10-26
Update example containers to fix security vulnerabilities


roberthbailey:
While working on https://github.com/googleforgames/agones/issues/1126 and verifying that the new images were successfully uploaded, I noticed that many of our images have vulnerabilities that are identified by Google's automated image analysis.

Most of the vulnerabilites are probably just because we are using older dependencies in the base image layers (e.g. linux distro with outdated glibc) and should be easy to fix by going through the Dockerfiles and updating them to pick up patches.

aLekSer:
Have enabled containers security analysis, now I am able to see them. Useful links to enable this feature:
https://cloud.google.com/container-registry/docs/enabling-disabling-container-analysis
https://cloud.google.com/container-registry/docs/get-image-vulnerabilities

markmandel:
Just looking through the examples - with the exclusion of the Unity sample (which requires a Unity licence to update, which requires so work to get 😄 ) everything else looks relatively up to date.

I'm figuring we close this, and if need be open a specific ticket to update the Unity example ?

roberthbailey:
> I'm figuring we close this, and if need be open a specific ticket to update the Unity example?

sounds good to me.

markmandel:
Looks like we marked this stale in July 😄 - going close it, since there is no objections.

PR #1609, opened 2020-06-03
Update Debian version for CPP example

aLekSer:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

 /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Use more recent and updated version of Debian OS which includes less vulnerabilities.

For #1154

### DIFF
--- a/examples/cpp-simple/Dockerfile
+++ b/examples/cpp-simple/Dockerfile
@@ -34,12 +34,10 @@ RUN cd cpp-simple && mkdir -p .build && \
     cmake .. -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=.bin && \
     cmake --build . --target install
 
-FROM debian:stretch
+FROM debian:buster
 RUN useradd -u 1000 -m server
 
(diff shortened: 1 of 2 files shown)


examples/cpp-simple/Dockerfile
```diff
@@ -34,12 +34,10 @@ RUN cd cpp-simple && mkdir -p .build && \
...
     cmake --build . --target install
 
-FROM debian:stretch
+FROM debian:buster
 RUN useradd -u 1000 -m server
 
-COPY --from=builder /project/cpp-simple/.build/.bin/cpp-simple /home/server/cpp-simple
-RUN chown -R server /home/server && \
-    chmod o+x /home/server/cpp-simple
+COPY --from=builder --chown=server:server /project/cpp-simple/.build/.bin/cpp-simple /home/server/cpp-simple
```
EricFortin, line 40: You removed the `chmod o+x` line. Was it intended?

aLekSer: I have tested the resulted file, and it has the same permissions. So this two strings should be equivalent.

Issue #1074, opened 2019-09-25
E2E tests should use a randomly created Namespace for testing


pooneh-m:
**Is your feature request related to a problem? Please describe.**
All E2E tests are using default namespace in the test cluster when running, which makes it harder to identify which test is causing issues in the shared cluster. Tests are also more error prone by mistakenly reusing resources from other tests in the same namespace.

**Describe the solution you'd like**
Test runs should be isolated to their own namespaces.

markmandel:
Interesting idea! Is this mainly to make it easier to parse logs?

pooneh-m:
Yes, and if there is a leak from testing like what we saw few days ago, it is easier to debug. WDYT?

pooneh-m:
Also, at the test cleanup, deleting the namespace deletes all the resources. There is no need to delete resources one by one from default namespace.

aLekSer:
@akremsa can start working on this. In two steps:

1.  Add new variable `NAMESPACE` for `make install` and `make test-e2e`
1.  Change `e2e.sh` in `e2e-runner` image and `cloudbuild.yaml`

PR #1623, opened 2020-06-11
Added flags which allow to pass namespace to e2e tests

akremsa:
**What type of PR is this?**
> /kind feature

We need this flag in order to be able to pass a randomly generated namespace to e2e tests.

Part of #1074


pooneh-m:
How about [creating a new namespace](https://github.com/googleforgames/agones/blob/master/test/e2e/framework/framework.go#L491 ) at the test runtime?

aLekSer:
May be we can use both options, a flag to generate the namespace and another option as in this PR. Do you mean we should add `ServiceAccount` and `ClusterRole` per `gameservers.namespaces` not in `make install` step but in the test?
https://github.com/googleforgames/agones/blob/54f6a3885d7f4e7ced32b57ae9ba64f51dd0332c/install/helm/agones/templates/serviceaccounts/sdk.yaml#L30

pooneh-m:
> May be we can use both options, a flag to generate the namespace and another option as in this PR. Do you mean we should add `ServiceAccount` and `ClusterRole` per `gameservers.namespaces` not in `make install` step but in the test?
> https://github.com/googleforgames/agones/blob/54f6a3885d7f4e7ced32b57ae9ba64f51dd0332c/install/helm/agones/templates/serviceaccounts/sdk.yaml#L30

That's setting up the `ServiceAccount` in `default` namespace. 

[CreateNamespace()](https://github.com/googleforgames/agones/blob/master/test/e2e/framework/framework.go#L491) also sets up a new namespaces with `ServiceAccount` but for test only. If your motivation to add the flag is for testing on a non-default namespace, using CreateNamespace() does that for you, with all the required setup.

pooneh-m:
The original idea for https://github.com/googleforgames/agones/issues/1074 was to create a namespace per test suite at runtime and use the new namespace instead of default to help with isolation and debugging e.g. [here](https://github.com/googleforgames/agones/blob/master/test/e2e/allocator_test.go#L64-L65). Is there a usecase for passing in a namespace as an argument?

aLekSer:
@pooneh-m, now I got your idea.
I think for backward compatibility, let's leave an option to use `default` or any other namespace. If namespace is an empty string then use random approach you are suggesting.
So running:
`make test-e2e GAMESERVERS_NAMESPACE=default`
is equivalent to current state.
`make test-e2e` new randomised approach.
`createNamespace(time.Now())` could be used for example.
This way we can switch back easily if something goes wrong.
And no changes to `e2e.sh` is needed in Cloud Build.

akremsa:
I will update this PR

akremsa:
@pooneh-m  @aLekSer please, review this PR again.

At this moment there is the following behavior:
_GAMESERVERS_NAMESPACE_ is empty by default, so random namespace will be created after you run `make test-e2e`.
If you specify a namespace like `make test-e2e GAMESERVERS_NAMESPACE="xbox"`, agones will run tests against specified namespace. If there is no such a namespace - error will be returned. In this case, you need to create a namespace beforehand.

aLekSer:
Overall looks good, but I think we should give some hints how to run `test-e2e` with and without `GAMESERVERS_NAMESPACE` flag in `build/README.md` and probably additional comment in Makefile itself is needed.

aLekSer:
E2E test failed:
```
time="2020-06-16 13:58:19.721" level=info msg="failing Allocate request" error="rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: authentication handshake failed: x509: cannot validate certificate for 35.233.192.59 because it doesn't contain any IP SANs\""
--- FAIL: TestTlsWebhook (391.55s)
    fleetautoscaler_test.go:636: error waiting for fleet condition on fleet simple-fleet-778hp
FAIL
time="2020-06-16 14:05:07.569" level=info msg="Namespace 1592315887 is deleted"
```

pooneh-m:
> @pooneh-m @aLekSer please, review this PR again.
Looks great!!! Thanks for the change.
> 
> At this moment there is the following behavior:
> _GAMESERVERS_NAMESPACE_ is empty by default, so random namespace will be created after you run `make test-e2e`.
> If you specify a namespace like `make test-e2e GAMESERVERS_NAMESPACE="xbox"`, agones will run tests against specified namespace. If there is no such a namespace - error will be returned. In this case, you need to create a namespace beforehand.
> 
> One remark: I've noticed that in case if an error is returned after namespace is created and tests fail with error - that recently created namespace will remain. Should I add better error validations to remove dangling namespaces or it's better to do that in a separate PR?

The defer should clean up the namespace, unless there is a failure with deleting the namespace. If the namespace is pre-created, then the creator of the namespace should clean up.

akremsa:
@pooneh-m sorry, could you rephrase? 

For example, there is a test which creates a namespace. An error happens inside CreateNamespace after namespace is created. Should I add a DeleteNamespace call to error checks inside CreateNamespace in order to delete that namespace if we return an error and namespace won’t be used at all?
That’s my understanding of this scenario, am I right?

pooneh-m:
> @pooneh-m sorry, could you rephrase?
> 
> For example, there is a test which creates a namespace. An error happens inside CreateNamespace after namespace is created. Should I add a DeleteNamespace call to error checks inside CreateNamespace in order to delete that namespace if we return an error and namespace won’t be used at all?
> That’s my understanding of this scenario, am I right?

I see. When `CreateNamespace` is called, there are multiple [steps](https://github.com/googleforgames/agones/blob/master/test/e2e/framework/framework.go#L501-L503) after the namespace is created and they may fail. In this case we should delete the namespace if other steps failed. However, I would say it is nice to have to do that level of clean up, but not necessary.

akremsa:
Ok, I get it. I will try to implement that in a different PR.

akremsa:
Regarding TestTlsWebhook - it works only with _default_ namespace. Let's check why...

aLekSer:
E2E test has failed:
```
--- FAIL: TestGameServerPassthroughPort (319.72s)
    gameserver_test.go:667: 
        	Error Trace:	gameserver_test.go:667
        	            	waiting for {udp-server [{ Passthrough <nil> 0 0 }] {false 0 0 0}  { 0 0} {{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] nil []  []} {[] [] [{udp-server gcr.io/agones-images/udp-server:0.21 [] []  [] [] [{PASSTHROUGH TRUE nil}] {map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}] map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}]} [] [] nil nil nil   IfNotPresent nil false false false}]  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil>}} <nil>} GameServer instance readiness timed out (): waiting for GameServer to be Ready 1592392955/udp-serverprfck: timed out waiting for the condition
        	Test:       	TestGameServerPassthroughPort
    gameserver_test.go:668: 
        	Error Trace:	gameserver_test.go:668
        	Error:      	Could not get a GameServer ready
        	Test:       	TestGameServerPassthroughPort
FAIL
```

test/e2e/main_test.go
```diff
@@ -49,19 +49,38 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
-	// run cleanup before tests, to ensure no resources from previous runs exist.
-	err = framework.CleanUp(defaultNs)
-	if err != nil {
-		log.WithError(err).Error("failed to cleanup resources")
-	}
+	if framework.Namespace == "" {
+		// use a random namespace
```
aLekSer, line 53: This is not a random function, as the output is easy predictable.
```suggestion
		// use a custom namespace - Unix timestamp
```

test/e2e/main_test.go
```diff
@@ -49,19 +49,38 @@ func TestMain(m *testing.M) {
...
+		framework.Namespace = strconv.Itoa(int(time.Now().Unix()))
+		log.Infof("Randomly generated namespace: %s", framework.Namespace)
```
aLekSer, line 55:
```suggestion
		log.Infof("Custom generated namespace: %s", framework.Namespace)
```

test/e2e/main_test.go
```diff
@@ -49,19 +49,38 @@ func TestMain(m *testing.M) {
...
+
+		if err := framework.CreateNamespace(framework.Namespace); err != nil {
+			log.WithError(err).Error("failed to create a random namespace")
```
aLekSer, line 58: And here

akremsa:
@aLekSer thanks for comments. Updated.

test/e2e/allocator_test.go
```diff
@@ -119,11 +128,26 @@ func TestAllocatorCrossNamespace(t *testing.T) {
 
 	// Create namespaces A and B
 	namespaceA := fmt.Sprintf("allocator-a-%s", uuid.NewUUID())
-	framework.CreateNamespace(t, namespaceA)
-	defer framework.DeleteNamespace(t, namespaceA)
+	err := framework.CreateNamespace(namespaceA)
```
pooneh-m, line 131: namespaceA can be framework.Namespace. 
namespaceA := framework.Namespace
No need to create a new one.

test/e2e/allocator_test.go
```diff
@@ -61,8 +61,17 @@ func TestAllocator(t *testing.T) {
 	tlsCA := refreshAllocatorTLSCerts(t, ip)
 
 	namespace := fmt.Sprintf("allocator-%s", uuid.NewUUID())
-	framework.CreateNamespace(t, namespace)
-	defer framework.DeleteNamespace(t, namespace)
+
```
pooneh-m, line 64: Instead of creating a new namespace, it looks like we can reuse `namespace := framework.Namespace`

test/e2e/allocator_test.go
akremsa: done

PR #1629, opened 2020-06-17
test-e2e - Added several Fatal calls

akremsa:
**What type of PR is this?**
> /kind bug
> /kind cleanup

Added those additional Fatal calls in order to avoid nil pointer exceptions in some places.

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -87,7 +87,7 @@ func TestSDKSetLabel(t *testing.T) {
 		return gs.ObjectMeta.Labels != nil, nil
 	})
 
-	if assert.Nil(t, err) {
+	if assert.NoError(t, err) {
 		assert.NotEmpty(t, gs.ObjectMeta.Labels["agones.dev/sdk-timestamp"])
 	}
 }
@@ -116,8 +116,8 @@ func TestHealthCheckDisable(t *testing.T) {
 	time.Sleep(10 * time.Second)
 
 	gs, err = framework.AgonesClient.AgonesV1().GameServers(framework.Namespace).Get(readyGs.ObjectMeta.Name, metav1.GetOptions{})
-	if !assert.NoError(t, err) {
-		assert.FailNow(t, "gameserver get failed")
+	if err != nil {
+		assert.FailNow(t, "gameserver get failed", err.Error())
 	}
 
 	assert.Equal(t, agonesv1.GameServerStateReady, gs.Status.State)
@@ -169,13 +169,13 @@ func TestUnhealthyGameServerAfterHealthCheckFail(t *testing.T) {
 	gs.Spec.Health.FailureThreshold = 1
 
 	gs, err := framework.CreateGameServerAndWaitUntilReady(framework.Namespace, gs)
-	if !assert.NoError(t, err) {
-		assert.FailNow(t, "Failed to create a gameserver")
+	if err != nil {
+		assert.FailNow(t, "Failed to create a gameserver", err.Error())
 	}
 
 	reply, err := e2eframework.SendGameServerUDP(gs, "UNHEALTHY")
-	if !assert.NoError(t, err) {
-		assert.FailNow(t, "Failed to send a message to a gameserver")
+	if err != nil {
+		assert.FailNow(t, "Failed to send a message to a gameserver", err.Error())
 	}
 	assert.Equal(t, "ACK: UNHEALTHY\n", reply)
 
@@ -186,8 +186,8 @@ func TestUnhealthyGameServerAfterHealthCheckFail(t *testing.T) {
 func TestUnhealthyGameServersWithoutFreePorts(t *testing.T) {
 	t.Parallel()
 	nodes, err := framework.KubeClient.CoreV1().Nodes().List(metav1.ListOptions{})
-	if !assert.NoError(t, err) {
-		assert.FailNow(t, "Failed to list nodes")
+	if err != nil {
+		assert.FailNow(t, "Failed to list nodes", err.Error())
 	}
 	assert.True(t, len(nodes.Items) > 0)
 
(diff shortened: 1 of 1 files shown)


test/e2e/gameserver_test.go
```diff
@@ -87,8 +87,9 @@ func TestSDKSetLabel(t *testing.T) {
 		return gs.ObjectMeta.Labels != nil, nil
 	})
 
-	assert.Nil(t, err)
-	assert.NotEmpty(t, gs.ObjectMeta.Labels["agones.dev/sdk-timestamp"])
+	if assert.Nil(t, err) {
```
markmandel, line 90: if we're here, can we flip this to `assert.NoError()` - which will give us the error results if it fails.

test/e2e/gameserver_test.go
```diff
@@ -210,7 +219,7 @@ func TestGameServerUnhealthyAfterDeletingPod(t *testing.T) {
 	t.Parallel()
 	gs := framework.DefaultGameServer(framework.Namespace)
 	readyGs, err := framework.CreateGameServerAndWaitUntilReady(framework.Namespace, gs)
-	if err != nil {
+	if !assert.NoError(t, err) {
 		t.Fatalf("Could not get a GameServer ready: %v", err)
```
markmandel, line 223: I think we should be consistent, if we are doing a NoError, then is should have an inner t.FatalNow -- don't think it makes sense to repeat the error value twice.

test/e2e/gameserver_test.go
```diff
@@ -260,12 +273,12 @@ func TestGameServerRestartBeforeReadyCrash(t *testing.T) {
 	messageAndWait := func(gs *agonesv1.GameServer, msg string, check func(gs *agonesv1.GameServer, pod *corev1.Pod) bool) error {
 		return wait.PollImmediate(3*time.Second, 3*time.Minute, func() (bool, error) {
 			gs, err := gsClient.Get(gs.ObjectMeta.Name, metav1.GetOptions{})
-			if err != nil {
+			if !assert.NoError(t, err) {
```
markmandel, line 276: Not sure this adds any value? (same with one below)

We're already capturing this below.

test/e2e/gameserver_test.go
```diff
@@ -276,7 +289,10 @@ func TestGameServerRestartBeforeReadyCrash(t *testing.T) {
 
 			// create a connection each time, as weird stuff happens if the receiver isn't up and running.
 			conn, err := net.Dial("udp", address)
-			assert.NoError(t, err)
+			if !assert.NoError(t, err) {
```
markmandel, line 292: We should probably return the err here instead to escape out, and let the final check handle this, i think

test/e2e/gameserver_test.go
```diff
@@ -320,7 +336,7 @@ func TestGameServerRestartBeforeReadyCrash(t *testing.T) {
 		}
 		return false
 	})
-	if err != nil {
+	if !assert.NoError(t, err) {
```
markmandel, line 339: So we've got a consistent pattern of doing this as established i.e. check if err != nil and then failing with  t.Fatalf or assert.Failx()

What's the value add of changing this here? assert.NoError() doesn't seem like it's doing anything extra here?

test/e2e/gameserver_test.go
```diff
@@ -818,20 +847,26 @@ func TestPlayerConnectAndDisconnect(t *testing.T) {
...
-	assert.NoError(t, err)
+	if !assert.NoError(t, err) {
+		t.Fatalf("Could not message GameServer: %v", err)
+	}
 	assert.Equal(t, "3\n", reply)
 
 	err = wait.Poll(time.Second, time.Minute, func() (bool, error) {
 		gs, err = framework.AgonesClient.AgonesV1().GameServers(framework.Namespace).Get(gs.ObjectMeta.Name, metav1.GetOptions{})
-		if err != nil {
+		if !assert.NoError(t, err) {
```
markmandel, line 869: Same as all the above - I don't think these are providing any kind of extra value, and could be confusing.

test/e2e/gameserver_test.go
akremsa: ok, I agree

test/e2e/gameserver_test.go
akremsa: There are places where we check errors with assert library in tests, that is why I decided to do the same:
https://github.com/googleforgames/agones/blob/master/pkg/util/apiserver/apiserver_test.go#L100
https://github.com/googleforgames/agones/blob/master/test/e2e/fleetautoscaler_test.go#L358
https://github.com/googleforgames/agones/blob/master/test/e2e/gameserverallocation_test.go#L356

As I see, currently there is no single approach and we use both options across the project. You are right, there is no big difference, but as far as this is a test file I decided to use a testing library here.
In which cases do you think we should use _err != nil_ or _assert.NoError()_ ?

akremsa:
@markmandel I've updated PR taking into account your comments. Please check.

test/e2e/gameserver_test.go
markmandel: I don't know if I have strong opinions on either one. @roberthbailey I feel like you had opinions on this at one stage.

I feel like consolidating to:

```
	if !assert.NoError(t, err) {
		assert.FailNow(t, "the is my message for why this failed")
	}
```

Feels like the solution that best takes advantage of the test libraries we use, but that I think Robbie has preferred `if err != nil` more?

Thoughts?

roberthbailey: It's redundant to both assert.NoError and then call Fail, since the former calls fail already: https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1327

The one place where I don't recommend using the assert calls in tests is in goroutines, since calling asserts there don't necessarily bubble up nicely back to the main testing thread and give you a good idea of why a test failed. In those cases, doing "normal" error checking (`if err != nil`) and returning regular errors from the goroutines and then doing the asserts when they come back to the main routine is generally nicer. This is where I generally add comments about doing the normal error checking.

markmandel: > It's redundant to both assert.NoError and then call Fail, since the former calls fail already:

But `FailNow` will stop the test at that specific moment, which I think is the point. Or am I missing the point of your comment?

akremsa: Yes, aborting the test in case of an error is an idea, because if we go further nil value will be used. As a result, I’ve noticed nil pointer exceptions in some tests along with assert-fail messages in case if test fails.

roberthbailey: If we assert twice then we will duplicate error logs in the test output. So if we need to fail now I'm in favor of doing a regular `if err != nil` check to reduce noise in the test logs.

akremsa:
Updated again. I've replaced `!assert.NoError` with `err != nil` and put error message to `FailNow` call.

PR #1635, opened 2020-06-22
Fixed GameServer State Diagram

suecideTech:
Reflect the explanation of FAQ on the diagram.
https://agones.dev/site/docs/faq/#how-can-i-return-an-allocated-gameserver-to-the-ready-state

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation
> /kind feature
> /kind hotfix

Looking at this figure, I misunderstood that GameServer should not transition from Allocated to Ready.
So, I added that you can transition from Allocated to Ready as an optional.

None.

None.


site/static/diagrams/gameserver-states.dot
```diff
@@ -28,6 +28,7 @@ digraph {
     Scheduled -> RequestReady [ label="SDK.ready()", color="purple" ]
     RequestReady -> Ready [ label="ready to be allocated", color="blue" ]
     Reserved -> Ready [label="if seconds > 0  \land failed to call  \lSDK.allocate()  " color="purple"]
+    Allocated -> Ready [ label="(optional)  \lSDK.ready()", color="purple" ]
```
roberthbailey, line 31: Mark - does `SDK.ready()` move it directly back to the Ready state or back to the RequestReady state?

Also, I don't think the "(optional)" prefix is required, since many of the arrows in the diagram are optional.

markmandel: Yes, you are 100% right. This should point to RequestReady (which then flows down to Ready).

Agreed on (optional) prefix as well.

Otherwise, this looks like a great addition, thanks for submitting!

suecideTech:
@roberthbailey @markmandel 
Thank you for review!

I fixed it, but the output figure by Graphviz has changed drastically.
I have no knowledge of Graphviz...sorry.
Anyone have an idea to improve this?

roberthbailey:
> I fixed it, but the output figure by Graphviz has changed drastically.

That is expected, since the graph is drawn in a way that attempts to optimize the layout (and you added a new edge).

markmandel:
Looks great to me!

markmandel:
That's a new flaky one:

```
--- FAIL: TestLocal (0.00s)
    localsdk_test.go:67: 
        	Error Trace:	localsdk_test.go:67
        	Error:      	Not equal: 
        	            	expected: &sdk.GameServer_ObjectMeta{Name:"local", Namespace:"default", Uid:"1234", ResourceVersion:"v1", Generation:1, CreationTimestamp:1592928132, DeletionTimestamp:0, Annotations:map[string]string{"annotation":"true"}, Labels:map[string]string{"islocal":"true"}, XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
        	            	actual  : &sdk.GameServer_ObjectMeta{Name:"local", Namespace:"default", Uid:"1234", ResourceVersion:"v1", Generation:1, CreationTimestamp:1592928131, DeletionTimestamp:0, Annotations:map[string]string{"annotation":"true"}, Labels:map[string]string{"islocal":"true"}, XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(*sdk.GameServer_ObjectMeta)(name:"local" namespace:"default" uid:"1234" resource_version:"v1" generation:1 creation_timestamp:1592928132 annotations:<key:"annotation" value:"true" > labels:<key:"islocal" value:"true" > )
        	            	+(*sdk.GameServer_ObjectMeta)(name:"local" namespace:"default" uid:"1234" resource_version:"v1" generation:1 creation_timestamp:1592928131 annotations:<key:"annotation" value:"true" > labels:<key:"islocal" value:"true" > )
        	            	 
        	Test:       	TestLocal
```

PR #1638, opened 2020-06-22
Adding support for refreshing TLS certs in the allocator

### LINKED ISSUE #1599: Automatically refresh all allocator TLS certs, not just client CA cert
Our infrastructure takes advantage of short lifetime certificates that are automatically renewed & loaded into pods. Currently we need to restart the allocator pods whenever such a renew happens, as the allocator does not automatically refresh all certs. 

It appears as if PR #1145 introduced auto-refresh of the client CA, and this feature request asks that this functionality is extended to all client & server certs in use by the allocator. 

The hot-reload implementation in #1145 is extended to all certs in `/home/allocator/tls` and `/home/allocator/client-ca`.

* A liveness probe on the allocator k8s config that will kill any pods with expired certs (I believe this will only work for server certs as last I checked we can't have mTLS liveness probes)
* An external service that will delete pods that are older than X days - forcing the reload.

devloop0:
**What type of PR is this?**
/kind feature

Adds support for refreshing TLS certs in the allocator (the allocator already supports refreshing client certs).

Closes #1599


cmd/allocator/main.go
```diff
@@ -107,6 +113,16 @@ func main() {
 		for {
 			select {
 			// watch for events
+			case event := <-watcherTLS.Events:
+				h.tlsMutex.Lock()
+				tlsCert, err := readTLSCert()
+				if err != nil {
+					logger.WithError(err).Infof("could load TLS cert; keeping old one")
+				} else {
+					h.tlsCert = tlsCert
```
pooneh-m, line 122: While we are here can we change `h.caCertPool = caCertPool` to be in an else clause as well?

pooneh-m:
Looks pretty good! Happy to see this change.

site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -101,14 +101,6 @@ kubectl get secret allocator-client-ca -o json -n agones-system | jq '.data["cli
 
 The last command creates a new entry in the secret data map called `client_trial.crt` for `allocator-client-ca` and stores it. You can also achieve this by `kubectl edit secret allocator-client-ca -n agones-system`, and then add the entry.
 
-## Restart pods
```
pooneh-m, line 104: Thanks for the change. Please release guard it by 
```
{{% feature expiryVersion="1.7.0" %}}
## Restart pods
...
{{% /feature %}}
```

cmd/allocator/main.go
devloop0: Done!

site/content/en/docs/Advanced/allocator-service.md
devloop0: Oops, my bad. Guarded it instead.

pooneh-m:
@devloop0 great work! Thanks for your contribution.

cmd/allocator/main.go
```diff
@@ -97,6 +97,12 @@ func main() {
 		logger.WithError(err).Fatalf("cannot watch folder %s for secret changes", certDir)
 	}
 
+	watcherTLS, _ := fsnotify.NewWatcher()
```
roberthbailey, line 100: It looks like this does the same thing as on line 94 above, but the second return value from NewWatcher is an error, which we are ignoring. Code should never ignore errors.

See https://github.com/golang/go/wiki/CodeReviewComments#handle-errors

cmd/allocator/main.go
```diff
@@ -107,13 +113,24 @@ func main() {
 		for {
 			select {
 			// watch for events
+			case event := <-watcherTLS.Events:
+				h.tlsMutex.Lock()
+				tlsCert, err := readTLSCert()
+				if err != nil {
+					logger.WithError(err).Error("could not load TLS cert; keeping old one")
+				} else {
+					h.tlsCert = tlsCert
```
roberthbailey, line 122: This is the only line that needs the mutex guard, right?

cmd/allocator/main.go
```diff
@@ -173,26 +190,43 @@ func newServiceHandler(kubeClient kubernetes.Interface, agonesClient versioned.I
 	}
 	h.caCertPool = caCertPool
 
+	tlsCert, err := readTLSCert()
+	if err != nil {
+		logger.WithError(err).Fatal("could not load TLS certs.")
+	}
+	h.tlsCert = tlsCert
```
roberthbailey, line 197: Since this line changes the value protected by the mutex, should the mutex be held here as well?

cmd/allocator/main.go
```diff
@@ -173,26 +190,43 @@ func newServiceHandler(kubeClient kubernetes.Interface, agonesClient versioned.I
...
+
 	return &h
 }
 
+func readTLSCert() (*tls.Certificate, error) {
+	tlsCert, err := tls.LoadX509KeyPair(tlsDir+"tls.crt", tlsDir+"tls.key")
+	if err != nil {
+		logger.WithError(err).Infof("failed to generate credentials")
```
roberthbailey, line 205: Both places that call this function log the error, so no need to log it here.

cmd/allocator/main.go
```diff
@@ -275,7 +309,9 @@ func getCACertPool(path string) (*x509.CertPool, error) {
 type serviceHandler struct {
 	allocationCallback func(*allocationv1.GameServerAllocation) (k8sruntime.Object, error)
 	certMutex          sync.RWMutex
```
roberthbailey, line 311: It's best to put the mutex next to the thing that it guards, so that it's obvious what is protected by the mutex. So:

type serviceHandler struct {
  allocationCallback ...

  certMutex sync.RWMutex
  caCertPool *x509.CertPool

  tlsMutex sync.RWMutex
  tlsCert *tls.Certificate
}

cmd/allocator/main.go
devloop0: Checked the errors in the creation of both watchers in the new PR (https://github.com/googleforgames/agones/pull/1638).

cmd/allocator/main.go
devloop0: Yup, fixed this in (https://github.com/googleforgames/agones/pull/1638).

cmd/allocator/main.go
devloop0: Now, newServiceHandler is called to create the serviceHandler, so no one can access the object yet.

cmd/allocator/main.go
devloop0: Removed it in (https://github.com/googleforgames/agones/pull/1638).

cmd/allocator/main.go
devloop0: Reorganized the members in (https://github.com/googleforgames/agones/pull/1638).

PR #1642, opened 2020-06-23
New feature: SDK cached gameserver

### LINKED ISSUE #1630: WatchGameServer should immediately provide the cached GameServer

I'm always frustrated when I call the `WatchGameServer` API on the sidecar, and I have to wait up to 30 seconds to receive my current `GameServer`, if no changes happen.

This is particularly frustrating if the transition to `Allocated` happened before I called `WatchGameServer`, which means I'm waiting 30 seconds of _player visible_ time for my `GameServer` annotations, so I can accept the allocation and start playing.

When connecting to the `WatchGameServer`, the current `GameServer` resource, if already cached in the sdk server, is immediately sent to the client.

The alternative is to _also_ call `GetGameServer`. The problem with that is that you must start the Watch first, or you may still miss an update and still have to wait 30 seconds. If you start the Watch first, and an update comes down it while you are calling `GetGameServer`, then you have to detect this and discard the result of `GetGameServer`. And if the socket connection is lost for some reason, I have to repeat the process, in case I missed my `Allocated` update while recalling `WatchGameServer`. I don't like that approach.

Another alternative would be for every SDK to do the above inside its `WatchGameServer` wrapper, which would be (I think) needlessly complicated. It also wouldn't help for my use-case, as we are not yet using the Unreal SDK, as (amongst other reasons) it doesn't implement `WatchGameServer`.

An even worse alternative would be for the resync time to be reduced from 30 seconds to 5 seconds. I'm sure I can find even worse alternatives if I try...

**Additional context**

In our specific case, we realised we should call `WatchGameServer` before we announce `Ready`, and then we don't have the player-visible delay, as long as nothing else goes wrong. But in general, I'd hope `WatchGameServer` worked more like `kubectl get --watch` than `kubectl get --watch-only`.

Also this seems like a four-line change to `func (s *SDKServer) WatchGameServer` that wouldn't break the sidecar's API either. (Context in case I have time to create a PR myself at some future point, if this isn't rejected as a feature)

akremsa:
**What type of PR is this?**
> /kind feature

Return GS immediately after WatchGameServer is called. Long discussion [here](https://github.com/googleforgames/agones/issues/1630).

Closes #1630 

Should _feature-stages.md_ be also updated?

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -59,7 +59,7 @@ KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 # Game Server image to use while doing end-to-end tests
 GS_TEST_IMAGE ?= gcr.io/agones-images/udp-server:0.21
 
-ALPHA_FEATURE_GATES ?= "PlayerTracking=true&GameServerCaching=true"
+ALPHA_FEATURE_GATES ?= "PlayerTracking=true&SDKWatchSendOnExecute=true"
 
 # Directory that this Makefile is in.
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
--- a/build/Makefile
+++ b/build/Makefile
@@ -59,7 +59,7 @@ KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 # Game Server image to use while doing end-to-end tests
 GS_TEST_IMAGE ?= gcr.io/agones-images/udp-server:0.21
 
-ALPHA_FEATURE_GATES ?= "PlayerTracking=true"
+ALPHA_FEATURE_GATES ?= "PlayerTracking=true&GameServerCaching=true"
 
 # Directory that this Makefile is in.
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -100,6 +100,7 @@ type SDKServer struct {
 	gsReserveDuration  *time.Duration
 	gsPlayerCapacity   int64
 	gsConnectedPlayers []string
+	cachedGameServer   *sdk.GameServer
 }
 
 // NewSDKServer creates a SDKServer that sets up an
@@ -491,7 +492,16 @@ func (s *SDKServer) GetGameServer(context.Context, *sdk.Empty) (*sdk.GameServer,
 		return nil, err
 	}
 
-	return convert(gs), nil
+	s.cachedGameServer = convert(gs)
+	return s.cachedGameServer, nil
+}
+
+func (s *SDKServer) getCachedGameServer() (*sdk.GameServer, error) {
+	if s.cachedGameServer != nil {
+		return s.cachedGameServer, nil
+	}
+
+	return s.GetGameServer(context.Background(), &sdk.Empty{})
 }
 
 // WatchGameServer sends events through the stream when changes occur to the
@@ -499,6 +509,19 @@ func (s *SDKServer) GetGameServer(context.Context, *sdk.Empty) (*sdk.GameServer,
 func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServerServer) error {
 	s.logger.Debug("Received WatchGameServer request, adding stream to connectedStreams")
 	s.streamMutex.Lock()
+
+	if runtime.FeatureEnabled(runtime.FeatureGameServerCaching) {
+		gs, err := s.getCachedGameServer()
+		if err != nil {
+			s.logger.WithError(errors.WithStack(err)).Error("error getting cached game server")
+		} else {
+			err := stream.Send(gs)
+			if err != nil {
+				s.logger.WithError(errors.WithStack(err)).Error("error sending cached game server")
+			}
+		}
+	}
+
 	s.connectedStreams = append(s.connectedStreams, stream)
 	s.streamMutex.Unlock()
 	// don't exit until we shutdown, because that will close the stream
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -510,15 +510,17 @@ func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServer
 	s.logger.Debug("Received WatchGameServer request, adding stream to connectedStreams")
 	s.streamMutex.Lock()
 
-	if runtime.FeatureEnabled(runtime.FeatureGameServerCaching) {
+	if runtime.FeatureEnabled(runtime.FeatureSDKWatchSendOnExecute) {
 		gs, err := s.getCachedGameServer()
 		if err != nil {
 			s.logger.WithError(errors.WithStack(err)).Error("error getting cached game server")
-		} else {
-			err := stream.Send(gs)
-			if err != nil {
-				s.logger.WithError(errors.WithStack(err)).Error("error sending cached game server")
-			}
+			return err
+		}
+
+		err = stream.Send(gs)
+		if err != nil {
+			s.logger.WithError(errors.WithStack(err)).Error("error sending cached game server")
+			return err
 		}
 	}
 
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -100,7 +100,6 @@ type SDKServer struct {
 	gsReserveDuration  *time.Duration
 	gsPlayerCapacity   int64
 	gsConnectedPlayers []string
-	cachedGameServer   *sdk.GameServer
 }
 
 // NewSDKServer creates a SDKServer that sets up an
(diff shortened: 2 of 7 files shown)


akremsa:
> Please add an update to the documentation. https://agones.dev/site/docs/guides/feature-stages/#feature-gates

That is what I was asking, thanks.

pkg/sdkserver/sdkserver.go
```diff
@@ -491,14 +492,36 @@ func (s *SDKServer) GetGameServer(context.Context, *sdk.Empty) (*sdk.GameServer,
...
 // WatchGameServer sends events through the stream when changes occur to the
 // backing GameServer configuration / status
 func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServerServer) error {
 	s.logger.Debug("Received WatchGameServer request, adding stream to connectedStreams")
 	s.streamMutex.Lock()
+
+	if runtime.FeatureEnabled(runtime.FeatureGameServerCaching) {
+		gs, err := s.getCachedGameServer()
+		if err != nil {
+			s.logger.WithError(errors.WithStack(err)).Error("error getting cached game server")
```
markmandel, line 516: I think it would be better to return the error here instead. (Maybe wrap it, maybe not?)

Seems like it shouldn't be swallowed like this. If it fails, there's a good reason.

pkg/sdkserver/sdkserver.go
```diff
@@ -491,14 +492,36 @@ func (s *SDKServer) GetGameServer(context.Context, *sdk.Empty) (*sdk.GameServer,
...
+		} else {
+			err := stream.Send(gs)
+			if err != nil {
+				s.logger.WithError(errors.WithStack(err)).Error("error sending cached game server")
```
markmandel, line 520: Same here, error should be returned, and we can wrap it with some context as well.

pkg/sdkserver/sdkserver_test.go
```diff
@@ -665,6 +665,54 @@ func TestSDKServerWatchGameServer(t *testing.T) {
 	assert.Equal(t, stream, sc.connectedStreams[1])
 }
 
+func TestSDKServerWatchGameServer_FeatureGameServerCaching(t *testing.T) {
```
markmandel, line 668: Nit: no _ please.

build/Makefile
```diff
@@ -59,7 +59,7 @@ KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 # Game Server image to use while doing end-to-end tests
 GS_TEST_IMAGE ?= gcr.io/agones-images/udp-server:0.21
 
-ALPHA_FEATURE_GATES ?= "PlayerTracking=true"
+ALPHA_FEATURE_GATES ?= "PlayerTracking=true&GameServerCaching=true"
```
markmandel, line 62: Not a huge fan of "GameServerCaching" as a name Feels a bit ambiguous as to it's meaning. Maybe something that has "SDK" in it?

Some ideas:
- SDKWatchSendImmeadiate
- SDKWatchSendOnExecute
- WatchGameServerImmeadiate

Something like that? Not sure I like any of them 100% more, but something in that area would probably be good. WDYT?

As I discovered the other day, this will also need to be added to the cloudbuild.yaml
https://github.com/googleforgames/agones/blob/master/cloudbuild.yaml#L234

site/content/en/docs/Guides/Client SDKs/_index.md
```diff
@@ -163,6 +163,8 @@ the `message GameServer`.
 For language specific documentation, have a look at the respective source (linked above), 
 and the {{< ghlink href="examples" >}}examples{{< /ghlink >}}.
 
+You can use `GameServerCaching` feature passed as a feature gate if you want a `GameServer` to be returned right after `GameServerFeature` is called.
```
markmandel, line 166: Can we also wrap this is a feature shortcode, so it only shows up in the next release please :+1: 

```suggestion
You can use `GameServerCaching` feature passed as a feature gate if you want a `GameServer` to be returned right after `SDK.WatchGameServer(...)` is called.
```

Might also want to link to  https://agones.dev/site/docs/guides/feature-stages/#feature-gates here as well for information on how to turn this on.

build/Makefile
steven-supersolid: Agree. It feels like game server caching is being used to achieve the feature of sending on watch. Could actually implement that without any caching and might want to consider as separate features.
Would throw in feature suggestions of `SendGameServerOnWatch` or `SendOnWatch` as the context of what is being watched can be gleaned from the SDK function WatchGameServer, and this could be extended to also send immediately if we add any future watchers.

pkg/sdkserver/sdkserver.go
akremsa: There is a line in _gameServer_ function:
`return gs, errors.Wrapf(err, "could not retrieve GameServer %s/%s", s.namespace, s.gameServerName)`

So we already have a context, I think it's fine to return an error as is but leave a log output.

akremsa:
PR is updated taking into account the comments above. Please check.

pkg/util/runtime/features.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	// FeatureContainerPortAllocation is a feature flag to enable/disable allocating ports to several containers in a pod
 	FeatureContainerPortAllocation Feature = "ContainerPortAllocation"
+
+	// FeatureSDKWatchSendOnExecute is a feature flag to enable/disable immediate game server return after SDK.WatchGameServer is called
+	FeatureSDKWatchSendOnExecute Feature = "SDKWatchSendOnExecute"
```
markmandel, line 41: I like this name :+1: good choice.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -42,6 +42,7 @@ The current set of `alpha` and `beta` feature gates are:
 | Example Gate (not in use) | `Example` | Disabled | None | 0.13.0 |
 | [Port Allocations to Multiple Containers]({{< ref "/docs/Reference/gameserver.md" >}}) | `ContainerPortAllocation` | Enabled | `Beta` | 1.7.0 |
 | [Player Tracking]({{< ref "/docs/Guides/player-tracking.md" >}}) | `PlayerTracking` | Disabled | `Alpha` | 1.6.0 |
+| [SDK Send Game Server on Watch]({{< ref "/docs/Guides/Client SDKs/_index.md#watchgameserver-function-gameserver" >}}) | `SDKWatchSendOnExecute` | Disabled | `Alpha` | 1.7.0 |
```
markmandel, line 45: Not sure I'm 100% on this wordage to make it clear the change, but not sure I can think of anything better.

"SDK Send GameServer on Watch execution" ?

pkg/sdkserver/sdkserver.go
```diff
@@ -491,14 +492,38 @@ func (s *SDKServer) GetGameServer(context.Context, *sdk.Empty) (*sdk.GameServer,
...
 }
 
 // WatchGameServer sends events through the stream when changes occur to the
 // backing GameServer configuration / status
 func (s *SDKServer) WatchGameServer(_ *sdk.Empty, stream sdk.SDK_WatchGameServerServer) error {
 	s.logger.Debug("Received WatchGameServer request, adding stream to connectedStreams")
 	s.streamMutex.Lock()
+
+	if runtime.FeatureEnabled(runtime.FeatureSDKWatchSendOnExecute) {
+		gs, err := s.getCachedGameServer()
```
markmandel, line 514: Not sure why we have this `getCacheGameServer()` - the Informer always has the up to date GameServer cached in memory, so we don't need to double-cache it I don't think, and it's likely a recipe for race conditions.

pkg/sdkserver/sdkserver.go
```diff
@@ -491,14 +492,38 @@ func (s *SDKServer) GetGameServer(context.Context, *sdk.Empty) (*sdk.GameServer,
...
+		if err != nil {
+			s.logger.WithError(errors.WithStack(err)).Error("error getting cached game server")
```
markmandel, line 516: Since this error is being pushed up the client, I don't think we need to log this - we don't log any other client facing errors.

We don't do it for any other client facing errors.

pkg/sdkserver/sdkserver.go
```diff
@@ -491,14 +492,38 @@ func (s *SDKServer) GetGameServer(context.Context, *sdk.Empty) (*sdk.GameServer,
...
+			return err
+		}
+
+		err = stream.Send(gs)
+		if err != nil {
+			s.logger.WithError(errors.WithStack(err)).Error("error sending cached game server")
```
markmandel, line 522: Same here

akremsa:
@markmandel  done

markmandel:
Are you rebased against master?

We did just move to Helm 3.

akremsa:
Yes, latest master ...

markmandel:
Yep, looks like you have. Weird. Investigating.

markmandel:
Ah found it. Residual Helm 2 installation.

markmandel:
@aLekSer need you to approve, so we  can merge :smile:

PR #1643, opened 2020-06-23
Fixed some formatting issues in the allocator code.

devloop0:
**What type of PR is this?**

/kind cleanup

Fixed some refactoring issues mentioned in: https://github.com/googleforgames/agones/pull/1638

Closes #

@roberthbailey these are the changes from the closed PR that you suggested. I responded to the comments you made in that one, but feel free to make other comments here as well.

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -91,13 +91,19 @@ func main() {
 	h := newServiceHandler(kubeClient, agonesClient, health)
 
 	// creates a new file watcher for client certificate folder
-	watcher, _ := fsnotify.NewWatcher()
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.WithError(err).Fatalf("could not create watcher for client certs")
+	}
 	defer watcher.Close() // nolint: errcheck
 	if err := watcher.Add(certDir); err != nil {
 		logger.WithError(err).Fatalf("cannot watch folder %s for secret changes", certDir)
 	}
 
-	watcherTLS, _ := fsnotify.NewWatcher()
+	watcherTLS, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.WithError(err).Fatalf("could not create watcher for tls certs")
+	}
 	defer watcherTLS.Close() // nolint: errcheck
 	if err := watcherTLS.Add(tlsDir); err != nil {
 		logger.WithError(err).Fatalf("cannot watch folder %s for secret changes", tlsDir)
@@ -114,15 +120,15 @@ func main() {
 			select {
 			// watch for events
 			case event := <-watcherTLS.Events:
-				h.tlsMutex.Lock()
 				tlsCert, err := readTLSCert()
 				if err != nil {
 					logger.WithError(err).Error("could not load TLS cert; keeping old one")
 				} else {
+					h.tlsMutex.Lock()
+					defer h.tlsMutex.Unlock()
 					h.tlsCert = tlsCert
 				}
 				logger.Infof("Tls directory change event %v", event)
-				h.tlsMutex.Unlock()
 			case event := <-watcher.Events:
 				h.certMutex.Lock()
 				caCertPool, err := getCACertPool(certDir)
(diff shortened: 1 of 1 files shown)


cmd/allocator/main.go
```diff
@@ -114,15 +120,15 @@ func main() {
...
 				} else {
+					h.tlsMutex.Lock()
+					defer h.tlsMutex.Unlock()
 					h.tlsCert = tlsCert
 				}
 				logger.Infof("Tls directory change event %v", event)
-				h.tlsMutex.Unlock()
 			case event := <-watcher.Events:
 				h.certMutex.Lock()
 				caCertPool, err := getCACertPool(certDir)
```
roberthbailey, line 134: [This is for line 203, but I can't comment there]

From https://github.com/googleforgames/agones/pull/1638

> Me: Since this line changes the value protected by the mutex, should the mutex be held here as well?
> You: No, newServiceHandler is called to create the serviceHandler, so no one can access the object yet.

Maybe it's not necessary in golang, but in other languages, setting a variable outside of a lock and then accessing it in a different thread (in a lock) doesn't guarantee that the second thread sees the value set in the first thread. This is because the lock is what instructs the program to introduce a memory fence to make sure that the value is visible across threads (e.g. move it outside of the cpu specific cache that isn't visible to other cpus on a multi-cpu machine).

cmd/allocator/main.go
```diff
@@ -91,13 +91,19 @@ func main() {
 	h := newServiceHandler(kubeClient, agonesClient, health)
 
 	// creates a new file watcher for client certificate folder
-	watcher, _ := fsnotify.NewWatcher()
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.WithError(err).Fatalf("could not create watcher for client certs")
```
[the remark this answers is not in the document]
devloop0: Oops, my bad. Fixed.

cmd/allocator/main.go
```diff
@@ -91,13 +91,19 @@ func main() {
...
+	}
 	defer watcher.Close() // nolint: errcheck
 	if err := watcher.Add(certDir); err != nil {
 		logger.WithError(err).Fatalf("cannot watch folder %s for secret changes", certDir)
 	}
 
-	watcherTLS, _ := fsnotify.NewWatcher()
+	watcherTLS, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.WithError(err).Fatalf("could not create watcher for tls certs")
```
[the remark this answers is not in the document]
devloop0: Fixed.

cmd/allocator/main.go
```diff
@@ -114,15 +120,15 @@ func main() {
 			select {
 			// watch for events
 			case event := <-watcherTLS.Events:
-				h.tlsMutex.Lock()
 				tlsCert, err := readTLSCert()
 				if err != nil {
 					logger.WithError(err).Error("could not load TLS cert; keeping old one")
 				} else {
+					h.tlsMutex.Lock()
+					defer h.tlsMutex.Unlock()
```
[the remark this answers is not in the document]
devloop0: Yeah, I thought defer would just unlock at the end of the current lexical scope (the linter actually pointed out this issue). Fixed.

cmd/allocator/main.go
devloop0: Yeah, I'm not exactly sure what Go's memory consistency model is (if it is sequentially consistent, I think this should be fine, but that's a pretty strong assumption). The ca cert initialization actually has the same problem, so I surrounded that with its corresponding lock as well. Fixed.

cmd/allocator/main.go
```diff
@@ -184,25 +190,28 @@ func newServiceHandler(kubeClient kubernetes.Interface, agonesClient versioned.I
 		logger.WithError(err).Fatal("starting allocator failed.")
 	}
 
+	h.certMutex.Lock()
```
roberthbailey, line 193: super nit: this could go right above line 198 so that the function call isn't done holding the lock (this makes it a lot clearer what is necessary to lock, which is just r/w access to `h.caCertPool`. Same below for `h.tlsCert`.

devloop0: Fixed! I agree it would make things more clear for someone reading the code.

PR #1645, opened 2020-06-24
Conditionally enable mtls for the allocator.

### LINKED ISSUE #1590: Proposal: Provide a flag to disable mTLS for agones-allocator
agones-allocator service supports [mTLS authentication](https://github.com/googleforgames/agones/blob/master/cmd/allocator/main.go#L127-L129) built into the service, which is used for `LoadBalancer` service and for enabling [multi-cluster allocation](https://agones.dev/site/docs/advanced/multi-cluster-allocation/) solution. 

However, a `NodePort` service used by [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) does not need an authentication mechanism built into the service. `Ingress` can provide mTLS support along with other ways of authentications. Also, Ingress in the cloud solutions handle the authentication. Side car proxies such as Envoy can provide mTLS support. 

Therefore, mTLS being built into the service prevents other solutions' adaptation.

Provide a flag to disable mTLS for agones-allocator.

* The service only enables mTLS if it is of type `LoadBalancer`.

* By default disable mTLS and change the service type to `NodePort` and document setting up `Ingress` resource. Currently, because the IP address is not allocated before Agones installation, Agones cannot issue a certificate that after install would be ready to use by the service and requires additional step.

devloop0:
**What type of PR is this?**
/kind feature

Customers can provide their own authentication (or none) instead of being forced to use mtls.

Partially addresses: #1590

### DIFF
--- a/cmd/allocator/main_test.go
+++ b/cmd/allocator/main_test.go
@@ -48,6 +48,7 @@ func TestAllocateHandler(t *testing.T) {
 				},
 			}, nil
 		},
+		mTLSEnabled: true,
 	}
 
 	request := &pb.AllocationRequest{
@@ -75,6 +76,7 @@ func TestAllocateHandlerReturnsError(t *testing.T) {
 		allocationCallback: func(gsa *allocationv1.GameServerAllocation) (k8sruntime.Object, error) {
 			return nil, k8serror.NewBadRequest("error")
 		},
+		mTLSEnabled: true,
 	}
 
 	request := &pb.AllocationRequest{}
@@ -93,7 +95,8 @@ func TestGetTlsCert(t *testing.T) {
 	assert.Nil(t, err, "expected (serverCert2, serverKey2) to create a cert")
 
 	h := serviceHandler{
-		tlsCert: &cert1,
+		tlsCert:     &cert1,
+		mTLSEnabled: true,
 	}
 
 	retrievedCert1, err := h.getTLSCert(nil)
@@ -123,6 +126,7 @@ func TestHandlingStatus(t *testing.T) {
 				Code: http.StatusUnprocessableEntity,
 			}, nil
 		},
+		mTLSEnabled: true,
 	}
 
 	request := &pb.AllocationRequest{}
@@ -146,6 +150,7 @@ func TestBadReturnType(t *testing.T) {
 		allocationCallback: func(gsa *allocationv1.GameServerAllocation) (k8sruntime.Object, error) {
 			return &corev1.Secret{}, nil
 		},
+		mTLSEnabled: true,
 	}
 
 	request := &pb.AllocationRequest{}
@@ -170,7 +175,8 @@ func TestVerifyClientCertificateSucceeds(t *testing.T) {
 	assert.True(t, certPool.AppendCertsFromPEM(crt))
 
 	h := serviceHandler{
-		caCertPool: certPool,
+		caCertPool:  certPool,
+		mTLSEnabled: true,
 	}
 
 	block, _ := pem.Decode(crt)
@@ -185,7 +191,8 @@ func TestVerifyClientCertificateFails(t *testing.T) {
 	crt := []byte(clientCert)
 	certPool := x509.NewCertPool()
 	h := serviceHandler{
-		caCertPool: certPool,
+		caCertPool:  certPool,
+		mTLSEnabled: true,
 	}
 
 	block, _ := pem.Decode(crt)
--- a/cmd/allocator/main_test.go
+++ b/cmd/allocator/main_test.go
@@ -48,7 +48,6 @@ func TestAllocateHandler(t *testing.T) {
 				},
 			}, nil
 		},
-		mTLSEnabled: true,
 	}
 
 	request := &pb.AllocationRequest{
@@ -76,7 +75,6 @@ func TestAllocateHandlerReturnsError(t *testing.T) {
 		allocationCallback: func(gsa *allocationv1.GameServerAllocation) (k8sruntime.Object, error) {
 			return nil, k8serror.NewBadRequest("error")
 		},
-		mTLSEnabled: true,
 	}
 
 	request := &pb.AllocationRequest{}
@@ -95,8 +93,7 @@ func TestGetTlsCert(t *testing.T) {
 	assert.Nil(t, err, "expected (serverCert2, serverKey2) to create a cert")
 
 	h := serviceHandler{
-		tlsCert:     &cert1,
-		mTLSEnabled: true,
+		tlsCert: &cert1,
 	}
 
 	retrievedCert1, err := h.getTLSCert(nil)
@@ -126,7 +123,6 @@ func TestHandlingStatus(t *testing.T) {
 				Code: http.StatusUnprocessableEntity,
 			}, nil
 		},
-		mTLSEnabled: true,
 	}
 
 	request := &pb.AllocationRequest{}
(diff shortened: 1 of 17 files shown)


pkg/util/runtime/features.go
```diff
@@ -34,6 +34,9 @@ const (
 	// FeaturePlayerTracking is a feature flag to enable/disable player tracking features.
 	FeaturePlayerTracking Feature = "PlayerTracking"
 
+	// FeatureMTLSEnabled is a feature flag to enable/disable mTLS in the allocator.
+	FeatureMTLSEnabled Feature = "MTLSEnabled"
```
pooneh-m, line 38: Disabling mtls is a new feature. We should also clarify that this is for allocator service. So, something like "AllocatorMTLSDisabled"?

pooneh-m:
Also, we should make TLS secret requirement conditional on the flag: https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/service/allocation.yaml#L136

pkg/util/runtime/features.go
roberthbailey: I agree. The default value of a feature flag should start with false, so changing this from Enable -> Disable will make a lot more sense.

roberthbailey, line 38: New feature gates need to be documented at https://github.com/googleforgames/agones/blob/master/site/content/en/docs/Guides/feature-stages.md

pkg/gameserverallocations/controller_test.go
```diff
@@ -1161,6 +1162,8 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
 
 func TestCreateRestClientError(t *testing.T) {
 	t.Parallel()
+	err := runtime.ParseFeatures("")
```
roberthbailey, line 1165: Can you explain why this needed to be added?

cmd/allocator/main_test.go
```diff
@@ -48,6 +48,7 @@ func TestAllocateHandler(t *testing.T) {
 				},
 			}, nil
 		},
+		mTLSEnabled: true,
```
roberthbailey, line 51: If you change the default for the feature gate, you shouldn't need to add anything here to keep the old behavior. You will only need to make changes where you want to test the new behavior.

pkg/util/runtime/features.go
devloop0: Makes sense to me. Fixed!

devloop0: Done.

cmd/allocator/main_test.go
devloop0: Changing it to mTLSDisabled let me remove this.

pkg/gameserverallocations/controller_test.go
devloop0: This test relies on createRemoteClusterDialOption which now relies on feature flags being enabled.

roberthbailey: @markmandel - it seems like we shouldn't have to call ParseFeatures with an empty list in test code; shouldn't the default behavior when checking for a feature be to use the compiled in list (e.g. we should only have to call this in tests where we are using a non-default feature flag)?

devloop0: The reason I needed to do this was because the key was originally MTLSEnabled, which should be true by default. But if the key wasn't in the map, the default value returned is false, which means that all of the logic that was being test was bypassed, since we would immediately return grpc.WithInsecure. By changing the feature semantics to MTLSDisabled, the default value returned by the map and the expected default value match now.

markmandel:
One thing I'm confused about - if I'm reading this right, the feature flag is what is turning the mTLS off or on?

Shouldn't there be a helm argument that sets this value that is part of the config as well? I know it's a bit of a double-set, but otherwise, would we need to introduce a configuration param to Helm (to be passed down to the allocator as an env/argument) to disable mTLS? Which would be a new feature? Would that be the plan long term?

I can definitely see arguments for not having both a feature flag and a helm config that essentially does both, but just wanted to be sure that was the long term plan.

devloop0:
I think the long-term plan should be to include a configuration parameter as part of the values.yaml file.
This offers us more functionality like completely avoiding the need to mount volumes (i.e. only conditionally enabling these lines: https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/service/allocation.yaml#L136). Currently, all this does is it avoids the need to verify and send certificates when the flag is set, but the certificates still need to present. I think having this flag in the values.yaml file is the best long-term solution.
The reason this isn't being done now is because it's hard to check what happens when yaml files are configured differently at test time.

markmandel:
So when we go to beta, then this feature flag will be on by default, so mTLS will be disabled for all users when this feature goes to beta? Because that's the beta feature level contract.

This seems like it's not going to work with our [feature gate flow](https://agones.dev/site/docs/guides/feature-stages/#description-of-stages). Also, when the feature goes GA, users will need to change their config from beta->GA because there will be a whole new configuration value - which seems to go against the spirit of the feature levels (i.e. beta should eventually close onto what GA will be before we switch).

Maybe this doesn't need a feature flag? Since it's a straight introduction of a new feature without it impacting existing functionality?

markmandel:
Will take a look, but looks like tests are failing because you'll need to run `make gen-install` :+1: 

(Since you edited the helm config, we check if the install.yaml in git still lines up)

markmandel:
Just had a look, This looks pretty good to me! - @devloop0 , @pooneh-m , @roberthbailey are we all happy not to have this as a feature flag, and have it be added as a single config value?

I can't personally see a way that this could break previous functionality, or how it might even change in the future? What do you all think?

markmandel:
```
Error: UPGRADE FAILED: "agones" has no deployed releases
Makefile:301: recipe for target 'install' failed
make: *** [install] Error 1
Error running handler: exit status 2
Child terminated, releasing lock
Cleanup aborted, lock in use
```

You'll want to rebase against master - just merged the changes that upgrade all our dev tools to Helm 3.

Sorry about that!

pkg/gameserverallocations/allocator.go
```diff
@@ -364,6 +366,10 @@ func (c *Allocator) allocateFromRemoteCluster(gsa *allocationv1.GameServerAlloca
 
 // createRemoteClusterDialOption creates a grpc client dial option with proper certs to make a remote call.
 func (c *Allocator) createRemoteClusterDialOption(namespace string, connectionInfo *multiclusterv1.ClusterConnectionInfo) (grpc.DialOption, error) {
+	if c.mTLSDisabled {
```
pooneh-m, line 369: If mTLS is disabled, authentication will be enabled through different means, e.g. Ingress. Allocators clients still should be able to secure connections to allocation endpoint. So for multi-cluster allocation please leave it as is.

pkg/gameserverallocations/controller_test.go
devloop0: Marking this as resolved since this is no longer an issue when using a helm configuration parameter.

pkg/gameserverallocations/allocator.go
devloop0: Replaced this with a TODO for considering how disableMTLS interacts with multi-cluster allocation.

install/helm/agones/templates/controller.yaml
```diff
@@ -120,6 +120,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: DISABLE_MTLS
```
pooneh-m, line 123: We don't need this here.

devloop0: Removed mTLS disabling from the gameserverallocator since we left that as a TODO for multi-cluster.

pooneh-m:
Thanks for the change @devloop0. Please also add a note to the issue that e2e testing for this environment variable needs to be added after the test infrastructure is supporting the environment variable testing. Then we need to document this value in the helm parameters.

devloop0:
Thanks! I'll be sure to add the note once this is merged.

markmandel:
Merging!

Issue #1436, opened 2020-03-27
Move support from Helm 2 ➡ Helm 3


markmandel:
Helm 3 was released in Nov 2019, and has several advantages over Helm 2 (No more Tiller! 🙌)

At some point, we should switch over, but it's unclear when that should happen.

The questions I have for the community to help drive this decision:

- Are you actively wishing to move to Helm 3?
- If Agones does make a breaking change switch to Helm 3, is that going to be an issue for you?
- Has anyone already moved to Helm 3, and is the current chart still working for you?

Feedback (and any other questions people think are relevant), are appreciated.

steven-supersolid:
We're using Google Cloud Shell to deploy which now comes with Helm 2 so we would need to upgrade, although that shouldn't be a problem.

Helm 2 is now in maintenance mode:
https://helm.sh/blog/2019-10-22-helm-2150-released/

The maintenance schedule is mentioned in the post:
> 6 months after Helm 3’s public release, Helm 2 will stop accepting bug fixes. Only security issues will be accepted.
> 12 months after Helm 3’s public release, support for Helm 2 will formally end.

The bug fix window has since been extended to 8 months:
https://helm.sh/blog/covid-19-extending-helm-v2-bug-fixes/

Key dates:
August 13th 2020 - only security fixes from this point
November 13th 2020 - support ends

danieloliveira079:
Hi there,

I am not sure how much that would help but I've managed to install Agones with `Helm3` using the current charts. A few considerations though:

- It was installed on a fresh provisioned local `Kubernetes 1.14.10` cluster created using [Kind](https://kind.sigs.k8s.io/).
- I never had `Helm2` installed on my laptop and installed the latest `Helm v3.1.0` just before trying it.
- I followed the migration steps described on https://github.com/helm/helm-2to3 and apparently, nothing was changed in regards to the Agones charts content. Running `git diff` showed nothing. The only issue I got was a complaining from helm saying that I didn't have some folders like `starters` and `repository`. However, after creating those 2 folders manually the issue was gone. That happened possibly because I didn't have `Helm2` installed.
- Installation command changed a bit. The following command works: `helm install --namespace agones-system my-release agones/agones`. Instead of `helm install --name my-release --namespace agones-system agones/agones`. The flag `--name` has been removed from the `helm install` command.

Considerations:
- The command `repo add agones https://agones.dev/chart/stable` is still valid.
[...]

PR #1647, opened 2020-06-25
Update developer tooling to Helm 3

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This updates the developer tooling to Helm 3.

Tiller can happily live alongside Helm 2, but you will want to remove the Agones helm 2 install, as it will overlap with the Helm 3
installation.

If you want to hand remove Agones - delete the agones-system namespace, and remove all agones CRDs. (`kubectl get crds | grep agones`)

The `tiller-deploy` deployment can also be deleted from the Kubernetes namespace if you need to hand remove Helm 2 as well.

Work on #1436

Before each build, one needs to manually remove the helm2 install of Agones, and then after each build of this PR, one should remove the helm 3 installation (`helm uninstall agones --namespace=agones-system`) otherwise we could end up with competing Agones installations.


build/Makefile
```diff
@@ -443,42 +446,41 @@ setup-prometheus: PVC ?= true
 setup-prometheus: PV_SIZE ?= 64Gi
 setup-prometheus: SCRAPE_INTERVAL=30s
 setup-prometheus:
-	$(DOCKER_RUN) \
-		helm upgrade --install --wait prom stable/prometheus --namespace metrics \
+	$(DOCKER_RUN) helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+	$(DOCKER_RUN) helm repo update
+	$(DOCKER_RUN) helm upgrade prom stable/prometheus \
+		--install --wait --namespace metrics --create-namespace \
```
roberthbailey, line 452: nit: I don't see a reason for a line break before `--set` any longer and then all flags will be on one line.

build/Makefile
```diff
@@ -443,42 +446,41 @@ setup-prometheus: PVC ?= true
...
-		helm delete --purge prom
+		helm uninstall prom --namespace=metrics
 
 # setup grafana in the current cluster with datasource and dashboards ready for use with agones
 # by default Persistent Volume Claims are requested.
 setup-grafana: PVC ?= true
 setup-grafana: PV_SIZE ?= 64Gi
 setup-grafana: PASSWORD ?= admin
 setup-grafana:
+	-$(DOCKER_RUN) kubectl create namespace metrics
```
roberthbailey, line 467: Why does this one create the namespace here instead of passing the `--create-namespace` flag like the other  `helm upgrade` commands?

roberthbailey: It looks like it's because of the kubectl apply before the helm upgrade.... Do we have to do it in that order or could we do the apply after the helm upgrade to remove this line?

roberthbailey: Also, what does the `-` at the start of this line for?

build/Makefile
```diff
@@ -443,42 +446,41 @@ setup-prometheus: PVC ?= true
...
-		helm upgrade --install --wait grafana stable/grafana --version=5.0.13  --namespace metrics \
+	$(DOCKER_RUN) helm upgrade grafana stable/grafana \
+		--install --wait --namespace metrics \
 		--set persistence.enabled=$(PVC),server.persistentVolume.size=$(PV_SIZE) \
 		--set adminPassword=$(PASSWORD) -f $(mount_path)/build/grafana.yaml
 
 uninstall-grafana: $(ensure-build-image)
 	$(DOCKER_RUN) \
-		helm delete --purge grafana
+		helm uninstall grafana --namespace=metrics
```
roberthbailey, line 478: Should this remove the resources that were created by the kubectl apply command as well?

build/Makefile
markmandel: That --set line is quite long:
https://github.com/googleforgames/agones/blob/0fce1b368cedd5314e823016e687d1afe9aa8b2f/build/Makefile#L453, at 135 chars.

That beings said , `$(HELM_ARGS) -f $(mount_path)/build/prometheus.yaml` could definitely be pulled into one line, so doing that now.

build/Makefile
markmandel: > It looks like it's because of the kubectl apply before the helm upgrade

I believe that is the reason yes. 

> Do we have to do it in that order or could we do the apply after the helm upgrade to remove this line?

I am honestly unsure (not that familiar with grafana), I'm just matching the order of what was done previously. I also have a strong feeling there was a long standing bug that if you didn't install prometheus first (which would create the namespace), then the namespace for grafana wouldn't exist, and this would fail.

> Also, what does the - at the start of this line for?

That means that if this command fails for any reason, the error will be ignored by the Makefile, and commands will continue. It's a nice way of saying "create this namespace if it doesn't already exist"

build/Makefile
markmandel: Yeah, we should.

build/Makefile
roberthbailey: Oh sorry, I didn't notice the line break in the code review tool and thought it was a super short line instead of a rather long one. :)

build/Makefile
roberthbailey: > Presumably helm does that for us as well (in this new flow) so it would be awesome if we could make this look like the other lines. I'm not sure if it's worth trying to fix in this PR though, or maybe tinker with it as a follow-up?

build/Makefile
roberthbailey: Do you want to fix that in this PR or in a subsequent one?

(since this is an improvement separate from upgrading to helm 3 and not a regression and I don't want to slow you down too much)

build/Makefile
markmandel: :shrug: maybe tinker with it in a subsequent PR?

build/Makefile
markmandel: Already done it :smile: just needed a `$(DOCKER_RUN) kubectl delete -f $(mount_path)/build/grafana/` will push up in a minute.

build/Makefile
roberthbailey: sgtm.

PR #1652, opened 2020-06-26
E2E: Invert ContainerPortAllocation FeatureGate

aLekSer:
ContainerPortAllocation now turned on by default, switching it off for E2E tests.

An idea behind is to invert all FeatureGates compared to default values for a second E2E test run.

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Closes #

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -228,11 +228,11 @@ steps:
     - GO111MODULE=on
 
 #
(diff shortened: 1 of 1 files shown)


markmandel:
That is a very good point! Nice catch!

PR #1656, opened 2020-06-26
Update Agones developer guide to point at faster targets

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

Running `make build` or `make test` is probably not reasonable for developers to use in a development iteration loop - so pointing new developers instead at `test-go` and `build-images`, which is what is needed most of the time.

None

None


markmandel:
> Also sometimes to safe even more time, if you have the same commit-id then make build-controller-image would be enough, if change touches only Agones controller itself.

This sounds like a good extra PR!

I tend to do:
`make -j 4 build-images && make -j 4 push` a lot!

PR #1657, opened 2020-06-27
Update Node.js dependencies in sdk and example

steven-supersolid:
**What type of PR is this?**
/kind cleanup

Update grpc-js dependency. Add required peer dependency google-auth-library
Update Nodejs example to use currently published SDK
Update & dedupe dependencies to reduce size

grpc-js has been update with some bug fixes


examples/nodejs-simple/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@google-cloud/agones-sdk": "1.5.0"
+    "@google-cloud/agones-sdk": "1.6.0"
```
markmandel, line 3: Is it worth also incrementing the example version?

https://github.com/googleforgames/agones/blob/master/examples/nodejs-simple/Makefile#L31

Outside of that, this is good to go.

markmandel:
I swear this only happens to you @steven-supersolid , it's so weird!

image

markmandel:
Would you mind rebasing to master, I've no idea why, but I can't hit the button, otherwise I would.

PR #1658, opened 2020-06-28
Alpha SDK and example for Node.js (Player tracking)

steven-supersolid:
**What type of PR is this?**
/kind feature

Adds a Node.js alpha SDK that extends the base SDK with the alpha features of Player Tracking. Also adds an example server to showcase the SDK.
Once the features change stage then they can be moved to the main SDK with the tests and leave the framework for future alpha features.

Relates to #1033
Branched from #1657


sdks/nodejs/spec/agonesSDK.spec.js
```diff
@@ -72,51 +72,51 @@ describe('agones', () => {
 		});
 	});
 
-	describe('allocate', () => {
+	describe('ready', () => {
```
steven-supersolid, line 75: Reordered to match the SDK, which was reordered to match the proto file

examples/nodejs-alpha/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "@google-cloud/agones-sdk": "file:../../sdks/nodejs"
```
steven-supersolid, line 3: This is because the npm module is not published yet and should be fine for local testing

sdks/nodejs/src/index.js
```diff
@@ -13,5 +13,7 @@
 // limitations under the License.
 
 const AgonesSDK = require('./agonesSDK');
+const AlphaAgonesSDK = require('./alphaAgonesSDK');
 
-module.exports = AgonesSDK;
+module.exports.sdk = AgonesSDK;
```
steven-supersolid, line 18: This is a breaking change how the SDK would have to be imported:
`const AgonesSDK = require('@google-cloud/agones-sdk');`
&rarr;
`const AgonesSDK = require('@google-cloud/agones-sdk').sdk;`

Also I now realise the case is inconsistent.

examples/nodejs-alpha/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1: I would suggest moving all of this into the Node.js example -- it's where it will end up eventually.

Having seperate examples for alpha/beta workloads is just going to compound the extra amount of work we need to do as we traverse one level to another.

If we need to hide these features behind a flag passed to the example, I think that's totally fine as well.

examples/nodejs-simple/Makefile
```diff
@@ -38,7 +38,7 @@ root_path = $(realpath $(project_path)/../..)
 #    |_|\__,_|_|  \__, |\___|\__|___/
 #                 |___/
 
-# build the cpp-simple binary
+# build the nodejs-simple binary
```
markmandel, line 41: Nice catch :+1:

sdks/nodejs/src/index.js
markmandel: Do we have to have this breaking change? Is there any way around it?

steven-supersolid: Sure, I have a couple of ideas

examples/nodejs-simple/README.md
```diff
@@ -12,6 +12,10 @@ It will:
 - After the shutdown duration (default 60 seconds), shut the server down
 - Parse options to get help or set the shutdown timeout duration
 
+If alpha features are enabled it will additionally:
```
markmandel, line 15: Nice!

sdks/nodejs/src/alphaAgonesSDK.js
```diff
@@ -0,0 +1,133 @@
+// Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1: Nit: 
```suggestion
// Copyright 2020 Google LLC All Rights Reserved.

```

sdks/nodejs/spec/alphaAgonesSDK.spec.js
```diff
@@ -0,0 +1,250 @@
+// Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1:
```suggestion
// Copyright 2020 Google LLC All Rights Reserved.
```

2020

examples/nodejs-simple/src/index.js
```diff
@@ -17,12 +17,18 @@ const util = require('util');
...
-const agonesSDK = new AgonesSDK();
-
 const DEFAULT_TIMEOUT = 60;
 const MAX_TIMEOUT = 2147483;
 
-const connect = async (timeout) => {
+const connect = async (timeout, enableAlpha) => {
+	let agonesSDK;
+
+	if (enableAlpha) {
```
markmandel, line 26: Possible other option - make it look like the Go SDK:
`s.Alpha().SetPlayerCapacity(capacity);` ([source](https://github.com/googleforgames/agones/blob/release-1.6.0/examples/simple-udp/main.go#L346)), where `s` is the SDK instance.

Or would that be a non-canonical Node pattern?

Just wanted to get your thorughts.

markmandel:
Just touching base, to see if this work is still ongoing?

steven-supersolid:
I can take another look this weekend and still mulling over what the canonical approach would be.
I think correctly we would distribute a separate npm module, e.g. 1.7.0-alpha and the new methods would be built in, however I also think that would be overkill here. The Go SDK approach is not so difficult to parse as to be avoided for JavaScript.

Why I chose this implementation is that there is an SDK class which has an interface of all available operations and once imported/required it is either a regular SDK or an alpha SDK. The alternative is that it is an SDK which has alpha property/method behind which the alpha features are located. I hadn't considered this and it is more of a composition pattern vs. my initial inheritance approach.

Random thoughts:
* If we were to introduce beta features then with this approach alpha would have to extend beta which doesn't seem quite right
* What if there is a change of interface in an alpha feature - should developers be able to call the original and the new method?
* Should it be clear with each method call that this is an alpha method?
* What if the alpha SDK requires some additional initialisation?

Perhaps overthinking but it would be useful to have a pattern to follow if we continue to roll out alpha and beta features.

markmandel:
> . The alternative is that it is an SDK which has alpha property/method behind which the alpha features are located. I hadn't considered this and it is more of a composition pattern vs. my initial inheritance approach.

This is how the Go SDK is working.

> If we were to introduce beta features then with this approach alpha would have to extend beta which doesn't seem quite right

Beta would be it's own implementation. So no inheritance required.

> What if there is a change of interface in an alpha feature - should developers be able to call the original and the new method?

Only the new. We can break Alpha (and beta) commands -- there is no guarantee of stability.

> Should it be clear with each method call that this is an alpha method?

That was the idea behind SDK.Alpha().Command() to make that clear.

> What if the alpha SDK requires some additional initialisation?

I can't this being a required at this stage, if so, we can handle that at that time.

Does that help?

sdks/nodejs/src/alpha.js
```diff
@@ -0,0 +1,129 @@
+// Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1:
```suggestion
// Copyright 2020 Google LLC All Rights Reserved.

```

sdks/nodejs/src/agonesSDK.js
```diff
@@ -14,15 +14,20 @@
...
 class AgonesSDK {
 	constructor() {
 		const services = grpc.loadPackageDefinition(servicesPackageDefinition);
-		this.client = new services.agones.dev.sdk.SDK(`localhost:${this.port}`, grpc.credentials.createInsecure());
+		const address = `localhost:${this.port}`;
+		const credentials = grpc.credentials.createInsecure();
+		this.client = new services.agones.dev.sdk.SDK(address, credentials);
 		this.healthStream = undefined;
 		this.streams = [];
+		this.alpha = new Alpha(address, credentials);
```
markmandel, line 30: Curious - any reason for `sdk.alpha` as a property, as opposed to `sdk.alpha()`

Just wondering if `sdk.alpha` could be overwritten by the user?

steven-supersolid: `sdk.alpha`  could be overwritten even if it is a method/function, but only on the instance

We could add an alpha method and would need to rename the property in JS but then might as well convert to a getter but at that point we have a getter for a public property which is redundant, and the getter could still be overwritten (as well as the property). There are ways of making things private but nothing straightforward so I would go with public which seems to be the JS way of doing things.

markmandel: Cool! Thanks for the explanation. That makes perfect sense :+1:

sdks/nodejs/spec/alphaAgonesSDK.spec.js
markmandel: This is the only thing that still needs changing, and then this good to merge in my opinion :+1:

markmandel:
Just a reminder:
https://github.com/googleforgames/agones/pull/1658/files#diff-278846dade2570eb0da62e878bab72dfR1

This is the only change that is left - this needs to be moved to `Copyright 2020` not 2019.

steven-supersolid:
Sorry I missed that one when I clicked the handy 'apply suggestion' button before!

PR #1659, opened 2020-06-30
Added missing FailNow calls to sdkserver unit tests

akremsa:
**What type of PR is this?**
> /kind bug
> /kind cleanup

There are places where in case of an error nil pointer exception will happen because assert.Fail() do not abort test execution. I followed the approach that we have discussed [previously](https://github.com/googleforgames/agones/pull/1629#discussion_r443875851).

I've noticed that [NewSDKServer](https://github.com/googleforgames/agones/blob/master/pkg/sdkserver/sdkserver.go#L107) function has return parameters `*SDKServer, error` , but actually it never returns an error. Why? It produces `if err != nil` checks where an error won't happen.

### DIFF
--- a/pkg/sdkserver/sdkserver_test.go
+++ b/pkg/sdkserver/sdkserver_test.go
@@ -333,7 +333,9 @@ func TestSidecarUpdateState(t *testing.T) {
 		t.Run(k, func(t *testing.T) {
 			m := agtesting.NewMocks()
 			sc, err := defaultSidecar(m)
-			assert.Nil(t, err)
+			if err != nil {
+				assert.FailNow(t, err.Error())
+			}
 			sc.gsState = agonesv1.GameServerStateReady
 
 			updated := false
@@ -373,7 +375,9 @@ func TestSidecarHealthLastUpdated(t *testing.T) {
 	m := agtesting.NewMocks()
 
 	sc, err := defaultSidecar(m)
-	assert.Nil(t, err)
+	if err != nil {
+		assert.FailNow(t, err.Error())
+	}
 	sc.health = agonesv1.Health{Disabled: false}
 	fc := clock.NewFakeClock(now)
 	sc.clock = fc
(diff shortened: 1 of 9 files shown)


pkg/sdkserver/sdkserver_test.go
```diff
@@ -333,7 +333,9 @@ func TestSidecarUpdateState(t *testing.T) {
 		t.Run(k, func(t *testing.T) {
 			m := agtesting.NewMocks()
 			sc, err := defaultSidecar(m)
-			assert.Nil(t, err)
+			if err != nil {
```
aLekSer, line 336: I think this should be wrapped in a helper function.

akremsa: It's just a matter of taste. I prefer the current style because if we introduce a function we need to copy it in every test file and when you read code you need to dig into that additional function to check what happens there. I prefer to keep things as simple as possible. I believe that function will become more and more complex in time with additional logic, logging parameters, and so on. I would like to avoid that.
Other opinions?

oleg-balunenko: As another option, you can use the package `require` from [testify](github.com/stretchr/testify) library.
Package `require` implements the same assertions as the `assert` package but stops test execution when a test fails.
I found function [NoError](https://pkg.go.dev/github.com/stretchr/testify@v1.6.1/require?tab=doc#NoError) very useful for such cases when we want all preconditions to be completed fully or fail the test.

akremsa: @oleg-balunenko good suggestion. Actually we can either continue following `if err != nil { FailNow() }` pattern or use another package to avoid these checks. I find assert methods descriptions a bit confusing, honestly.

markmandel: No issues with pulling in `require` on my end - a little less noise in the code when wanting to fail fast.

aLekSer: Yes, let us use `require`, I also have found it, while investigating alternatives, and by the way it would make code more compact.

PR #1665, opened 2020-06-30
Flaky TestLocal

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Fixed an issue with time slipping in this test, causing equality test to fail.

None

None


markmandel:
> I don't want to block this change, but using something like https://pkg.go.dev/github.com/google/go-cmp/cmp would make it easy to express things like "these should be equal except this one field" and might be worth considering using if this sort of pattern becomes common in our code base.

Ooh! that's nice!

I didn't see anything similar in https://pkg.go.dev/github.com/stretchr/testify@v1.6.1/assert?tab=doc - but might be worth popping through it again just in case -- the list of assertions has grown considerably since we first started using the package!

PR #1667, opened 2020-06-30
Flaky: TestSDKServerWatchGameServerFeatureSDKWatchSendOnExecute

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Switching default: case to a timeout, since due to the asynchronous nature of event processing, it could take a moment or two for the extra event to be processed.

This is particularly true on the CI system, where there is a lot of processing going on, and it may take more time than usual to process.

Closes #

### DIFF
--- a/pkg/sdkserver/sdkserver_test.go
+++ b/pkg/sdkserver/sdkserver_test.go
@@ -682,9 +682,8 @@ func TestSDKServerWatchGameServerFeatureSDKWatchSendOnExecute(t *testing.T) {
 	}
 
 	m := agtesting.NewMocks()
-	m.AgonesClient.AddReactor("list", "gameservers", func(action k8stesting.Action) (bool, runtime.Object, error) {
-		return true, &agonesv1.GameServerList{Items: []agonesv1.GameServer{*fixture}}, nil
-	})
+	fakeWatch := watch.NewFake()
+	m.AgonesClient.AddWatchReactor("gameservers", k8stesting.DefaultWatchReactor(fakeWatch, nil))
 
 	err := agruntime.ParseFeatures(string(agruntime.FeatureSDKWatchSendOnExecute) + "=true")
 	if !assert.NoError(t, err) {
@@ -700,6 +699,8 @@ func TestSDKServerWatchGameServerFeatureSDKWatchSendOnExecute(t *testing.T) {
 	stop := make(chan struct{})
 	defer close(stop)
 	sc.informerFactory.Start(stop)
+
+	fakeWatch.Add(fixture.DeepCopy())
 	assert.True(t, cache.WaitForCacheSync(stop, sc.gameServerSynced))
 	sc.gsWaitForSync.Done()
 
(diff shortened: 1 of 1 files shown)


pkg/sdkserver/sdkserver_test.go
```diff
@@ -718,7 +718,7 @@ func TestSDKServerWatchGameServerFeatureSDKWatchSendOnExecute(t *testing.T) {
 			} else {
 				assert.Fail(t, "Channel is closed!")
 			}
-		default:
+		case <-time.After(30 * time.Second):
```
roberthbailey, line 721: Does this still need to try 2 times? If you are going to add a timeout, maybe remove the for statement?

markmandel: This test is looking for two events to be passed through - the initial one on connection, then the second one from WatchGameServer.

I could rewrite it as two `select` statements, but that feels a little verbose?

roberthbailey: I didn't catch that part. I think it might be a bit clearer and/or more idiomatic to do a `for { }` loop with a break (when it times out or when you've gotten the expected events) but this is ok too.

markmandel: Would there be a concern that it would never escape the loop though?

roberthbailey: You would escape the loop as long as you add break / return (or panic) statements to each case, e.g.

```
for {
  select {
  case _, ok := <-stream.msgs:
    // do stuff - maybe assert that the right messages show up?
    break
  case <-time.After(timeout):
    // error, since we hit a timeout when we were expecting messages
    t.Fail() // should cause the test to end, so no need to break / return
  }
}
```

pkg/sdkserver/sdkserver_test.go
```diff
@@ -700,6 +699,8 @@ func TestSDKServerWatchGameServerFeatureSDKWatchSendOnExecute(t *testing.T) {
 	stop := make(chan struct{})
 	defer close(stop)
 	sc.informerFactory.Start(stop)
+
+	fakeWatch.Add(fixture.DeepCopy())
```
markmandel, line 703: Using a fake Watch so we can control explicitly when the game server is modified.

pkg/sdkserver/sdkserver_test.go
```diff
@@ -709,17 +710,31 @@ func TestSDKServerWatchGameServerFeatureSDKWatchSendOnExecute(t *testing.T) {
 	assert.Nil(t, waitConnectedStreamCount(sc, 1))
 	assert.Equal(t, stream, sc.connectedStreams[0])
 
+	// modify for 2nd event in watch stream
+	fixture.Status.State = agonesv1.GameServerStateAllocated
```
markmandel, line 714: @roberthbailey @akremsa - I reworked this test to check for:
1. We get what we expect in events on watch and modification
1. That we don't get too many events
1. We can't go into an infinite loop

PTAL, let me know what you think

akremsa: Nice update! I like the way you've updated a select section.

aLekSer: I also suggested to use state other than Ready.

PR #1668, opened 2020-06-30
Move CloudBuild to N1_HIGHCPU_32

markmandel:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**:

Let's try a bigger box, see if that helps with things.

**Which issue(s) this PR fixes**:

None, just general flakiness and build times.

**Special notes for your reviewer**:

Let's see what the logs say / build times are like.

PR #1671, opened 2020-07-01
Update Hugo and Docsy

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Due to #1668, saw that Hugo can fall over with:
`fatal error: concurrent map read and map write`

When there are lots of CPUs, so decided it was time to take the
opportunity to update Hugo.

None

Only big difference was to enable "unsafe" in the markdown engine, so that it would allow HTML in markdown.


site/config.toml
```diff
@@ -47,12 +47,8 @@ pygmentsStyle = "tango"
...
 blog = "/:section/:year/:month/:day/:slug/"
 
-## Configuration for BlackFriday markdown parser: https://github.com/russross/blackfriday
-[blackfriday]
-plainIDAnchors = true
-hrefTargetBlank = false
-angledQuotes = false
-latexDashes = true
+[markup.goldmark.renderer]
+unsafe= true
```
roberthbailey, line 51: nit: space before = sign?

markmandel: oh yeah, that's ugly. I'll fix that up. Copy paste from stack overflow issue!

PR #1685, opened 2020-07-09
Tests update: Use require package in fleets package

akremsa:
**What type of PR is this?**
> /kind bug
> /kind cleanup

Use require package to abort test in order to avoid nil pointer exceptions.

### DIFF
--- a/pkg/fleets/controller_test.go
+++ b/pkg/fleets/controller_test.go
@@ -34,6 +34,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	admv1beta1 "k8s.io/api/admission/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


markmandel:
I question if we should be replacing all asserts with require? Just thinking of other work that might be more valuable long term.

Or should we just replace the instances where we are doing variations of `FailNow()`?

akremsa:
One benefit which I see from using assert is that it will show ALL failed cases if they exist at once. For example:
```
a := "a"
b := "b"
assert.Equal(t, "zzz", a)
assert.Equal(t, "zzz", b)
```
This code will output 2 errors for each assert which can be useful to see all possible failed test cases at once.
I prefer to use require only for those cases where we want to abort test execution. 
On the other hand - sometimes it doesn't make sense to proceed with a test if one scenario is marked as Failed, but I think it's a developer's responsibility to write efficient tests.

markmandel:
> I prefer to use require only for those cases where we want to abort test execution.

100% agree :+1:

markmandel:
Does it make sense to close this PR? Since it looks like it won't be merged in it's current state? (sorry @akremsa !)

akremsa:
@markmandel do you mean failed builds? Otherwise, this PR is perfectly fine, no?

markmandel:
Let me come at this question another way - is all the places you've updated in this PR places where we want to abort execution of the test at the time of failure?

Maybe I misunderstood where we ended up.

akremsa:
"is all the places you've updated in this PR places where we want to abort execution of the test at the time of failure"
Yes, exactly. I abort tests in cases where there is no point to continue.

markmandel:
Ah! Got it. I will review this with that in mind. In which case, ignore my previous comment!

markmandel:
```
make[1]: *** [ensure-build-sdk-image] Error 2
includes/sdk.mk:171: recipe for target 'run-sdk-conformance-test-rest' failed
```

Issue #1686, opened 2020-07-09
Proposal: provide allocator-client.default secret in gameserver namespace


pooneh-m:
**Is your feature request related to a problem? Please describe.**
There are multiple steps for enabling multi-cluster allocation and one of the steps is creating an allocation client certificate. 

**Describe the solution you'd like**
This can be improved if Agones would create a self-signed TLS client certificate stored in k8s secret, named `allocator-client.default`. The secret would be available on the game server namespace by default, similar to agones-sdk service account and act as multi-cluster allocation client secret if in the `GameServerAllocationPolicy` the `secretName` is unspecified. 

Istio also has a similar functionality that would create `istio.default` TLS secret in every namespace in a k8s cluster.

pooneh-m:
@markmandel can you please take a look. I want to replace the multi-cluster allocation [client certificate](https://agones.dev/site/docs/advanced/multi-cluster-allocation/#establish-trust) creation by having a default secret created in the game server namespace. This is what istio citadel does (to have a default secret created) and it simplifies usage. The secret can be managed by a certificate manager solution on the cluster in the future.

markmandel:
So something like when we generate self-signed certs for the agones controller? ([source](https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/extensions.yaml#L14))

if so, that seems to make a lot of sense!

pooneh-m:
This is also a related customer request: https://github.com/googleforgames/agones/issues/1698

roberthbailey:
It looks like this was fixed by https://github.com/googleforgames/agones/pull/1702. Please re-open if there is outstanding work to do.

PR #1702, opened 2020-07-20
Added allocator-client.default secret.

devloop0:
**What type of PR is this?**
/kind feature

Adds a default client secret (self-signed) to the allocator.

#1686 


install/helm/agones/values.yaml
```diff
@@ -126,6 +126,7 @@ agones:
       serviceType: LoadBalancer
       annotations: {}
     generateTLS: true
+    generateClientTLS: false
```
pooneh-m, line 129: `true` by default?
Also, test it by getting a dependency to the client secret in the e2e: https://github.com/googleforgames/agones/blob/master/test/e2e/allocator_test.go#L61-L62

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -276,3 +276,23 @@ data:
...
+---
+# Default allocation client secret
+{{- if .Values.agones.allocator.generateClientTLS }}
+{{- $selfSignedCert := genSelfSignedCert "allocator-client.default" nil nil 3650 }}
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/tls
+metadata:
+  name: allocator-client.default
+  namespace: {{ .Release.Namespace }}
```
pooneh-m, line 289: This needs to be stored in the gameserver namespace.

pooneh-m: https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/serviceaccounts/sdk.yaml#L20

install/helm/agones/values.yaml
devloop0: Done!

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -276,3 +279,24 @@ data:
 {{- else }}
   tls-ca.crt: {{ .Files.Get "certs/allocator/server.crt" | b64enc }}
 {{- end }}
+
+---
+# Default allocation client secret
+{{- if .Values.agones.allocator.generateClientTLS }}
```
pooneh-m, line 285: Also update the doc: https://github.com/googleforgames/agones/blob/master/site/content/en/docs/Installation/Install%20Agones/helm.md

install/yaml/install.yaml
```diff
@@ -125,6 +125,23 @@ metadata:
...
...VEF6TVRZek4xb3dBRENDQVNJd0RRWUpLb1pJaHZjTgpBUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBTiszdHZMdlhsNVhneXdpSFgyOGw3K2Z3TUJHUTF3alRSdlNDdHF1CnQwaDZvUEtzRksydEZaQUlSWnFpbVVYL1UyR0pzWURqRUJRWVFTVSs1WVp0cTFZbGtVQmFEOUl5ZnEzNlNiNWcKUWw2eUFlU3AvblFPS3JoWC8zdFVwK1RnZ3RXcEpaTlZlSWlhM0kwVG1TbVFreTFnZmpGblMva01JdHFDZ2k2ZwpzRUlyNlFvVktEdHBzVFVHdXU3Wk55OHk4dHYvRjZObGZQelJpcGRkS3lyaFo2elhvaG9QSHRqSFM0TlhRMmxhCkdkQ3hveTIzV2VFODV0eHoyRWRSZkt0TXNWSmZmbjA3bEx0SDlhVnplcHduVzQzcG9jaE1zM1ZweFdhUW5JdjQKZ05yNGNHZ2thbm9oNDA5WUhpZkNnNkwyNFhVSUY0azVxNEpTcGZtY2h6RFFldDBDQXdFQUFhTS9NRDB3RGdZRApWUjBQQVFIL0JBUURBZ1dnTUIwR0ExVWRKUVFXTUJRR0NDc0dBUVVGQndNQkJnZ3JCZ0VGQlFjREFqQU1CZ05WCkhSTUJBZjhFQWpBQU1BMEdDU3FHU0liM0RRRUJDd1VBQTRJQkFRQUQ5VXIxY2dnSVVvMVExV1lIM0IycnROTVMKeHNPTGc3YXBaTE92cUdWWmxmMFg4MFhBMFlQVE5jaEFTdExVbU5TVHNkanQ3ZFJiZjI0M29saS8rTXNIQ25nQgprY1crajR3NjNvalU3M1E3TEJNSGZYSDMwYmRKRE9MNStvVnhnbUgyNmVQVEFhQXBLWXlQem1OblZubzFyOUNpCk51Y2QxK0NsSmx0ZjFmRi90LzBoSGpwV2REbGpRUktOQjd6THlzV25MdTN2UW9iSTBZK3E1TXgvNUQvb2xyeHoKQkpWR1pPRUgyK0pQQy8rb0lCMWFzSmJJcW9vemNmbEJ1SWY0d21sN3hxWVg2YWxPOXZqT2d1UEdYb05Pc3A1WQo0WUZyRHdCaXUrUkZ6Y2VtNmR3eE5yOUJNSFQ0cFcxU2g0M2REWkdKZDBPbnJ4VjdtT1pmQmo1cWNhekIKLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
```
pooneh-m, line 142: Set the helm config to false to avoid generating this every time. Then you need to have a client cert stored [here](https://github.com/googleforgames/agones/tree/master/install/helm/agones/certs/allocator) to be picked up by helm.

test/e2e/allocator_test.go
```diff
@@ -110,9 +107,6 @@ func TestAllocatorCrossNamespace(t *testing.T) {
 	requestURL := fmt.Sprintf(allocatorReqURLFmt, ip, port)
 	tlsCA := refreshAllocatorTLSCerts(t, ip)
 
-	// Create namespaces A and B
-	namespaceA := framework.Namespace // let's reuse an existing one
```
pooneh-m, line 114: The default namespace always has a game server. This test verifies that the game server is allocated from namespace B, because namespace A doesn't have a game server. If you want to change the namespace to default, you should also verify the gameserver origin. Otherwise, change the namespace creation logic to copy over the allocator-client.default cert. I prefer the latter.

install/helm/agones/values.yaml
devloop0: Set it to false from below.

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -229,7 +230,9 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 data:
-{{- if .Values.agones.allocator.generateTLS }}
+{{- if .Values.agones.allocator.generateClientTLS }}
+  client-ca.crt: {{ b64enc $selfSignedCert.Cert }}
```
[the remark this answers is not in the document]
devloop0: Done.

install/helm/agones/templates/service/allocation.yaml
devloop0: Done.

install/yaml/install.yaml
devloop0: Done.

test/e2e/allocator_test.go
devloop0: Done!

test/e2e/allocator_test.go
```diff
@@ -178,6 +183,30 @@ func TestAllocatorCrossNamespace(t *testing.T) {
...
+			Kind:       clientSecret.TypeMeta.Kind,
+			APIVersion: clientSecret.TypeMeta.APIVersion,
+		},
+		Type: corev1.SecretTypeOpaque,
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   clientSecret.ObjectMeta.Name,
+			Labels: clientSecret.ObjectMeta.Labels,
+		},
+		Data: clientSecret.Data,
+	}
```
pooneh-m, line 203: Can newSecret be removed and only store clientSecret `kubeCore.Secrets(toNamespace).Create(clientSecret)`

install/helm/agones/values.yaml
pooneh-m: Can you explain more? I am not sure what you mean.

test/e2e/allocator_test.go
devloop0: Nope, the returned secret includes some metadata that you can't include when creating the secret (I got an error that a secret to create can't include a resourceVersion). So I just kept the relevant fields I think.

install/helm/agones/values.yaml
devloop0: Changed the default value back to true and set the value to false in the Makefile.

test/e2e/allocator_test.go
pooneh-m: Then you just needs to set `Name` and `Data` on the new secret. `TypeMeta` is inferred from API and `SecretTypeOpaque` is the default secret type

devloop0: Yeah, I simplified the copying logic. I just needed to set the ResourceVersion and the Namespace correctly.

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -276,3 +282,27 @@ data:
...
+    app: {{ template "agones.fullname" $ }}
+    chart: "{{ $.Chart.Name }}-{{ $.Chart.Version }}"
+    release: "{{ $.Release.Name }}"
+    heritage: "{{ $.Release.Service }}"
+data:
+{{- if $.Values.agones.allocator.generateClientTLS }}
+  tls.crt: {{ b64enc $selfSigned.Cert }}
+  tls.key: {{ b64enc $selfSigned.Key }}
+{{- else }}
+  tls.crt: {{ $.Files.Get "certs/allocator/allocator-client.default/client.crt" | b64enc }}
```
pooneh-m, line 305: I have a security concern that if someone uses install.yaml instead of helm chart to install Agones for production workload, their cluster is exposed as the secrets are publicly available. Sorry, I should have caught that earlier.

How about not generating the certificate if generateClientTLS is not set to `true`? In that case is there any way to test the secret in the e2e?

devloop0: Could we leave a warning here? https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/NOTES.txt

pooneh-m: I don't expect every user would read documentation.

devloop0: Okay, then no. There would be no way to test this in the e2e script.

devloop0: As in maybe, we could add logic here: https://github.com/googleforgames/agones/blob/master/build/Makefile#L283 such that lines beginning with tls.crt, tls.key, allocator-client.default.crt, ..., could be whitelisted from the diff.

devloop0:
> Well done. Thank you!

Thanks for reviewing it!

PR #1713, opened 2020-07-23
Tests update: gameservers-controller tests update

akremsa:
**What type of PR is this?**
> /kind cleanup

Added missing test cases, abort tests in proper places using require package.

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -430,7 +430,7 @@ func (c *Controller) syncGameServerDeletionTimestamp(gs *agonesv1.GameServer) (*
 		if pod.ObjectMeta.DeletionTimestamp.IsZero() {
 			err = c.podGetter.Pods(pod.ObjectMeta.Namespace).Delete(pod.ObjectMeta.Name, nil)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/gameservers/controller.go
```diff
@@ -430,7 +430,7 @@ func (c *Controller) syncGameServerDeletionTimestamp(gs *agonesv1.GameServer) (*
 		if pod.ObjectMeta.DeletionTimestamp.IsZero() {
 			err = c.podGetter.Pods(pod.ObjectMeta.Namespace).Delete(pod.ObjectMeta.Name, nil)
 			if err != nil {
-				return gs, errors.Wrapf(err, "error deleting pod for GameServer %s, %s", gs.ObjectMeta.Name, pod.ObjectMeta.Name)
+				return gs, errors.Wrapf(err, "error deleting pod for GameServer. Name: %s, Namespace: %s", gs.ObjectMeta.Name, pod.ObjectMeta.Namespace)
```
aLekSer, line 433: Since GS `Name` is always equal to Pod `Name` this change add more information to the user who see this error.

Issue #1649, opened 2020-06-25
Upgrade to Kubernetes 1.16


roberthbailey:
As discussed on the monthly community meeting today, it's time to think about when we should upgrade our official support to Kubernetes 1.16. 

It looks like we *might* be on track to do it for the Agones 1.8 release. 

List of items to do for upgrading to 1.16 (this is copied from the [1.15 issue](https://github.com/googleforgames/agones/issues/1478) and may need to be updated):

- [x] Update e2e cluster to run against 1.16
  - [x] Update deployment manage script
  - [x] Recreate cluster with new scripts
  - [x] Update kubectl in `e2e-image/Dockerfile`
- [x] Update prow cluster to use 1.16 (even though we aren't using it yet, we should keep it in sync)
  - [x] Update deployment script
  - [x] Recreate cluster with new scripts
- [x] Update the dev tooling to create 1.16 clusters
  - [x] GKE
  - [x] Minikube
  - [x] Kind
  - [x] Update kubectl
- [x] Update terraform submodules
  - [x] GKE
  - [x] Azure
  - [x] EKS
- [x] Update documentation for creating clusters to 1.16
  - [x] Usage requirements
  - [x] GKE
  - [x] Minikube
  - [x] EKS
  - [x] AKS
  - [x] ~~Helm documentation~~ (no longer neccessary)
- [x] Update links to k8s documentation
  - [x] examples/fleet.yaml
  - [x] examples/fleetautoscaler.yaml
  - [x] examples/gameserver.yaml
  - [x] site/content/en/docs/Reference/fleet.md
  - [x] site/content/en/docs/Reference/fleetautoscaler.md
  - [x] site/content/en/docs/Reference/gameserver.md
  - [x] site/content/en/docs/Advanced/limiting-resources.md
- [x] Update to client-go 13.0 (based on [compatibility matrix](https://github.com/kubernetes/client-go#compatibility-matrix))
- [x] Move API references from beta to GA. For instance, `scheduling.k8s.io/v1beta1` to `scheduling.k8s.io/v1`. See https://github.com/googleforgames/agones/search?l=Go&q=v1beta1
- [x] Update `site/assets/templates/crd-doc-config.json`

roberthbailey:
The [Kubernetes v1.16 Release Notes](https://v1-16.docs.kubernetes.io/docs/setup/release/notes/#no-really-you-must-read-this-before-you-upgrade) have some points that we should keep in mind for this upgrade:

[...]

PR #1714, opened 2020-07-23
Use the scheduling.k8s.io/v1 API since the beta API will no longer be served by default starting with Kubernetes 1.16

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

 /kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Part of https://github.com/googleforgames/agones/issues/1649


aLekSer:
@roberthbailey could you please run `make gen-install-yaml` and add resulted HTML file to this PR.
`make test-gen-install-yaml` would succeed.
This should described in `build/README.md` I think.

aLekSer:
Note, I have updated previous comment with a correct command `make gen-api-docs`.

markmandel:
How you doing with this @roberthbailey - do you need someone to take it over? :smile:

roberthbailey:
I need to get my development environment back up and running with the new computer. So it's taking a while to run a simple docker command....

roberthbailey:
`make gen-api-docs` didn't create any new modified files. It looks like the test that failed was `test-install-yaml` so I ran `make gen-install` which I think will fix the broken tests.

markmandel:
Yay! v1 APIS!

PR #1717, opened 2020-07-25
Supports Rust Alpha SDK

yoshd:
**What type of PR is this?**

/kind feature

Add Rust Alpha SDK. This makes it easy to use the feature gate for player tracking in the Rust SDK.

For example, [GetPlayerCount](https://agones.dev/site/docs/guides/client-sdks/#alphagetplayercount) can be called like this:

```rs
let sdk = agones::Sdk::new()?;
let player_count = sdk.alpha().get_player_count()?;
```

Relates to #1033


test/sdk/rust/src/main.rs
```diff
@@ -112,6 +112,69 @@ fn run() -> Result<(), String> {
...
+        .alpha()
+        .get_connected_players()
+        .map_err(|e| format!("Could not run GetConnectedPlayers(): {}. Exiting!", e))?;
+    println!("Connected players: {:?}", player_ids);
+
+    let player_count = sdk
+        .alpha()
+        .get_player_count()
+        .map_err(|e| format!("Could not run GetConnectedPlayers(): {}. Exiting!", e))?;
+    println!("Cuurent player count: {}", player_count);
```
aLekSer, line 159:
```suggestion
    println!("Curent player count: {}", player_count);
```

test/sdk/rust/src/main.rs
```diff
@@ -112,6 +112,69 @@ fn run() -> Result<(), String> {
...
+        println!("Removed player");
+    } else {
+        panic!("Failed to remove player. Exiting!");
+    }
+
+    let player_count = sdk
+        .alpha()
+        .get_player_count()
+        .map_err(|e| format!("Could not GetPlayerCount(): {}. Exiting!", e))?;
+    println!("Cuurent player count: {}", player_count);
```
aLekSer, line 176:
```suggestion
    println!("Curent player count: {}", player_count);
```

test/sdk/rust/src/main.rs
yoshd: Oops, sorry, I'll fix.

yoshd: I fixed it

roberthbailey:
I am definitely a rust noobie and I think @markmandel has been writing some more rust lately so I'll let him take a look.

build/includes/sdk.mk
```diff
@@ -165,7 +165,7 @@ run-sdk-conformance-test-go:
 	$(MAKE) run-sdk-conformance-no-build SDK_FOLDER=go GRPC_PORT=9001 HTTP_PORT=9101 FEATURE_GATES=PlayerTracking=true TESTS=$(DEFAULT_CONFORMANCE_TESTS),$(ALPHA_CONFORMANCE_TESTS)
 
 run-sdk-conformance-test-rust:
-	$(MAKE) run-sdk-conformance-test SDK_FOLDER=rust
+	$(MAKE) run-sdk-conformance-test SDK_FOLDER=rust FEATURE_GATES=PlayerTracking=true TESTS=$(DEFAULT_CONFORMANCE_TESTS),$(ALPHA_CONFORMANCE_TESTS)
```
markmandel, line 168: This should probably follow what Go is doing - which is running one without the flag, and then one with:
https://github.com/googleforgames/agones/blob/8b4c100070cfacb31395015acdbd2b7d19262567/build/includes/sdk.mk#L163

sdks/rust/src/alpha.rs
```diff
@@ -0,0 +1,114 @@
...
+        let count = self.client.get_player_capacity(&req).map(|c| c.count)?;
+        Ok(count)
+    }
+
+    /// This changes the player capacity to a new value.
+    pub fn set_player_capacity(&self, capacity: i64) -> Result<()> {
+        let mut c = alpha::Count::new();
+        c.set_count(capacity);
+        let res = self.client.set_player_capacity(&c).map(|_| ())?;
+        Ok(res)
```
markmandel, line 50:
```suggestion
        self.client.set_player_capacity(&c)
```

Could this be reduced to a single line? It _looks_ like the types line up, but figured I would ask.

Same for similar things below.

test/sdk/rust/src/main.rs
```diff
@@ -112,6 +112,69 @@ fn run() -> Result<(), String> {
     sdk.set_label("test-label", &creation_ts.to_string())
         .map_err(|e| format!("Could not run SetLabel(): {}. Exiting!", e))?;
 
+    println!("Setting player capacity...");
```
markmandel, line 115: We could do a simple string comparison to see if "PlayerTracking=true" is contained in the FEATURE_GATES environment variable as a way to determine whether to run these tests or not?

Keep things simple, WDYT?

build/includes/sdk.mk
yoshd: Thank you for the review.
OK. As far as I can see, it seems better not to duplicate ports between tests in each language. 
I'll fix it like that.

test/sdk/rust/src/main.rs
yoshd: OK. It seems easy to implement so I'll fix.

sdks/rust/src/alpha.rs
yoshd: self.client.set_player_capacity returns grpcio::Result<super::alpha::Empty> in the code generated by protoc, but SDK returns Result<()>.
Therefore it cannot be changed as such.
https://github.com/googleforgames/agones/blob/8b4c100070cfacb31395015acdbd2b7d19262567/sdks/rust/src/grpc/alpha_grpc.rs#L132

I could change it to something like below, but I feel the brevity doesn't change much...
```rs
        self.client
            .set_player_capacity(&c)
            .map(|_| ())
            .map_err(|e| Error::from(e))
```

build/includes/sdk.mk
yoshd: I fixed it.

test/sdk/rust/src/main.rs
yoshd: I fixed it.

sdks/rust/src/alpha.rs
markmandel: Got it. That makes total sense.

PR #1719, opened 2020-07-27
Tests update: gameservers - health_test

akremsa:
**What type of PR is this?**

> /kind bug
> /kind cleanup

Added missing test cases, abort tests in proper places using require package.

### DIFF
--- a/pkg/gameservers/health_test.go
+++ b/pkg/gameservers/health_test.go
@@ -15,6 +15,7 @@
 package gameservers
 
 import (
+	"errors"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/gameservers/health_test.go
```diff
@@ -286,7 +288,44 @@ func TestHealthControllerSyncGameServer(t *testing.T) {
 	}
 }
 
+func TestHealthControllerSyncGameServer_UpdateFailed(t *testing.T) {
```
akremsa, line 291: Decided to put this case in a separate function in order to not to overload _TestHealthControllerSyncGameServer_ with extra _if_ statements to trigger and check an error.

markmandel: Minor nit: remove the underscore :smile: 

Otherwise - LGTM!

akremsa: This is my habit 😄

PR #1723, opened 2020-07-29
Tests update: portallocator_test

akremsa:
**What type of PR is this?**

> /kind bug
> /kind cleanup

Made several fixes in test cases, abort tests in proper places using require package.

### DIFF
--- a/pkg/gameservers/portallocator.go
+++ b/pkg/gameservers/portallocator.go
@@ -106,7 +106,6 @@ func (pa *PortAllocator) Run(stop <-chan struct{}) error {
 }
 
 // Allocate assigns a port to the GameServer and returns it.
-// Return ErrPortNotFound if no port is allocatable
 func (pa *PortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.GameServer {
 	pa.mutex.Lock()
 	defer pa.mutex.Unlock()
--- a/pkg/gameservers/portallocator_test.go
+++ b/pkg/gameservers/portallocator_test.go
@@ -24,6 +24,7 @@ import (
 	agtesting "agones.dev/agones/pkg/testing"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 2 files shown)


pkg/gameservers/portallocator.go
```diff
@@ -106,7 +106,6 @@ func (pa *PortAllocator) Run(stop <-chan struct{}) error {
 }
 
 // Allocate assigns a port to the GameServer and returns it.
-// Return ErrPortNotFound if no port is allocatable
```
aLekSer, line 109: This would be actually a warning:
```
11m         Warning   FailedScheduling          pod/simple-udp-qbx79-c2vll                           0/6 nodes are available: 2 node(s) had taints that the pod didn't tolerate, 4 node(s) didn't have free ports for the requested pod ports.
```

pkg/gameservers/portallocator_test.go
```diff
@@ -123,29 +125,30 @@ func TestPortAllocatorAllocate(t *testing.T) {
...
 		assert.NotEmpty(t, gs.Spec.Ports[0].HostPort)
 		assert.Len(t, pa.portAllocations, 3)
 	})
 
 	t.Run("ports are all allocated with multiple ports per GameServers", func(t *testing.T) {
 		m := agtesting.NewMocks()
-		maxPort := int32(19) // make sure we have an even number
-		pa := NewPortAllocator(10, maxPort, m.KubeInformerFactory, m.AgonesInformerFactory)
+		minPort := int32(10)
+		maxPort := int32(20) // make sure we have an even number
```
aLekSer, line 150:
```suggestion
		maxPort := int32(19) // make sure we have an even number
```
Now we have odd number of ports. Not sure what we should change this comment or set back the number and consequently other comments with `20`.

akremsa: I thought this comment is related to maxPort that is why I’ve changed it.

pkg/gameservers/portallocator_test.go
```diff
@@ -547,3 +563,13 @@ func countTotalAllocatedPorts(pa *PortAllocator) int {
 	}
 	return count
 }
+
+func countTotalPort(pa *PortAllocator) int {
```
aLekSer, line 567:
```suggestion
func countTotalPorts(pa *PortAllocator) int {
```

pkg/gameservers/portallocator_test.go
```diff
@@ -547,3 +563,13 @@ func countTotalAllocatedPorts(pa *PortAllocator) int {
...
+	count := 0
+	for _, node := range pa.portAllocations {
+		for range node {
+			count++
+		}
```
aLekSer, line 572:
```suggestion
		count += len(node)
```

pkg/gameservers/portallocator_test.go
markmandel: Outside of this comment, I think this PR is good to merge?

akremsa: Yes.
Are you aware of a context of this comment?
If we all are unsure about it I think we can remove one and merge.

markmandel: Looks like I wrote it.... and I can't remember anymore :confused:

akremsa: Updated

PR #1725, opened 2020-07-29
Fix findOpenPorts portAllocator function

aLekSer:
There could be the possible case when we set PortPolicy as an empty
string on update of GameServer.

**What type of PR is this?**

/kind bug

There is a way to cause infinite loop (recursion) in `Allocate()` function using `kubectl edit gameservers`, if you update the `PortPolicy` to an empty string "" and one more field. Then you would receive the state when Agones controller could not allocate any ports and got stuck in this state:
```
kubectl get gs
NAME                     STATE            ADDRESS         PORT   NODE                                     AGE
simple-udp-2skbm-657tc   PortAllocation                                                                   18m
simple-udp-2skbm-894s9   PortAllocation                                                                   18m
simple-udp-7xxfz         PortAllocation   35.230.21.139          gke-test-cluster-default-5847533f-6z66   73m
``` 

none

This is not expected scenario of editing `GameServer` but I assume that restarting agones controller with deleting bad GameServer would recover from this situation.

### DIFF
--- a/pkg/gameservers/portallocator_test.go
+++ b/pkg/gameservers/portallocator_test.go
@@ -253,7 +253,7 @@ func TestPortAllocatorAllocate(t *testing.T) {
 		assert.True(t, cache.WaitForCacheSync(stop, pa.nodeSynced))
 
 		err := pa.syncAll()
-		require.Nil(t, err)
+		require.NoError(t, err)
 
 		// single port empty
 		fd := fixture.DeepCopy()
--- a/pkg/gameservers/portallocator_test.go
+++ b/pkg/gameservers/portallocator_test.go
@@ -40,37 +40,6 @@ var (
 	n3 = corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "node3", UID: "node3"}}
 )
 
-func TestEmptyPortPolicy(t *testing.T) {
-	t.Parallel()
-	fixture := dynamicGameServerFixture()
-
-	t.Run("test portPolicy as an empty string", func(t *testing.T) {
-		m := agtesting.NewMocks()
-		pa := NewPortAllocator(10, 50, m.KubeInformerFactory, m.AgonesInformerFactory)
-		nodeWatch := watch.NewFake()
-		m.KubeClient.AddWatchReactor("nodes", k8stesting.DefaultWatchReactor(nodeWatch, nil))
-
-		stop, cancel := agtesting.StartInformers(m, pa.nodeSynced)
-		defer cancel()
-
-		// Make sure the add's don't corrupt the sync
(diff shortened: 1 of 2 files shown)


aLekSer:
Note that in that case with `amount=0` we never get from the recursion as line https://github.com/googleforgames/agones/pull/1725/files#diff-310ad1666561c0a99ea0effb347aea2bR149 would always return `false`.

pkg/gameservers/portallocator_test.go
```diff
@@ -39,6 +39,37 @@ var (
...
+		nodeWatch.Add(&n2)
+		assert.True(t, cache.WaitForCacheSync(stop, pa.nodeSynced))
+
+		err := pa.syncAll()
+		assert.Nil(t, err)
+
+		// single port empty
+		fd := fixture.DeepCopy()
+		fd.Spec.Ports[0].PortPolicy = ""
+		pa.Allocate(fd)
```
akremsa, line 67: _Allocate_ returns GameServer pointer, so it is worth checking it for nil.

pkg/gameservers/portallocator_test.go
```diff
@@ -39,6 +39,37 @@ var (
...
+		assert.Nil(t, err)
```
akremsa, line 68: Which error do you check here? Seems this assert is redundant.

pkg/gameservers/portallocator_test.go
```diff
@@ -39,6 +39,37 @@ var (
...
+		defer cancel()
+
+		// Make sure the add's don't corrupt the sync
+		// (no longer an issue, but leave this here for posterity)
+		nodeWatch.Add(&n1)
+		nodeWatch.Add(&n2)
+		assert.True(t, cache.WaitForCacheSync(stop, pa.nodeSynced))
+
+		err := pa.syncAll()
+		assert.Nil(t, err)
```
akremsa, line 62: Use require here, please. If an error happens here there is no point to continue.

pkg/gameservers/portallocator_test.go
```diff
@@ -39,6 +39,37 @@ var (
 	n3 = corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "node3", UID: "node3"}}
 )
 
+func TestEmptyPortPolicy(t *testing.T) {
```
akremsa, line 42: I think this test name is not accurate. It should describe which function do you test.
There is a test function TestPortAllocatorAllocate, so you can put this scenario there as a separate `t.Run("empty port policy")` statement.

aLekSer:
Thanks for the review, I have applied both series of comments.

PR #1730, opened 2020-08-01
Build Terraform: Use docker image project default

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

This fixes the issue if your local gcloud setup points to a different
project than the build shell, but pulling the current project from the
build image configuration.

N/A

Also snuck in the change that we don't need to specifically remove the
helm module before removing the entire cluster as well.

### DIFF
--- a/build/includes/terraform.mk
+++ b/build/includes/terraform.mk
@@ -15,6 +15,9 @@
 # The GKE development cluster name
 GCP_TF_CLUSTER_NAME ?= agones-tf-cluster
 
+# the current project
+current_project := $(shell $(DOCKER_RUN) bash -c "gcloud config get-value project 2> /dev/null")
+
 ### Deploy cluster with Terraform
 terraform-init:
 terraform-init: $(ensure-build-image)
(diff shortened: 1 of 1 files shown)


build/includes/terraform.mk
```diff
@@ -80,10 +79,7 @@ endif
...
 
+gcloud-terraform-destroy-cluster: GCP_PROJECT ?= $(current_project)
 gcloud-terraform-destroy-cluster:
-ifndef GCP_PROJECT
-	$(eval GCP_PROJECT=$(shell sh -c "gcloud config get-value project 2> /dev/null"))
-endif
-	$(DOCKER_RUN) bash -c 'cd $(mount_path)/build/terraform/gke && \
-	terraform destroy -var project=$(GCP_PROJECT) -target module.helm_agones.helm_release.agones -auto-approve && sleep 60 && \
-	terraform destroy -var project=$(GCP_PROJECT) -auto-approve'
+	$(DOCKER_RUN) bash -c 'cd $(mount_path)/build/terraform/gke && terraform destroy -var project=$(GCP_PROJECT) -auto-approve'
```
roberthbailey, line 84: Do we no longer need the sleep and re-run?

build/includes/terraform.mk
```diff
@@ -15,6 +15,9 @@
 # The GKE development cluster name
 GCP_TF_CLUSTER_NAME ?= agones-tf-cluster
 
+# the current project
+current_project := $(shell $(DOCKER_RUN) bash -c "gcloud config get-value project 2> /dev/null")
```
roberthbailey, line 19: This will _always_ use the project from the docker image, whereas before you could override the project if you wanted. I doubt folks were doing that, but maybe give a heads up in the developer slack channel when this goes in as an fyi.

markmandel: This still allows users to override GCP_PROJECT like they would have done previously. `current_project` is a placeholder for the default value.

build/includes/terraform.mk
markmandel: The helm3 terraform module solves the issue with cleaning up helm resources before deleting the cluster! 👍

PR #1741, opened 2020-08-06
Add firewall name variable to GKE Terraform

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This gives you the ability to specify what the firewall name should be for the configured Agones+GKE cluster, while keeping the default value.

This means that you can now have a single firewall rule if you desire in your network configurations, rather than one per cluster, when using our Terraform modules to configure your GKE clusters.

N/A

N/A


install/terraform/modules/gke/cluster.tf
```diff
@@ -46,10 +46,10 @@ local.zone)}
 }
 
 resource "google_container_cluster" "primary" {
-  name     = local.name
-  location = local.zone
-  project  = local.project
-  network  = local.network
+  name       = local.name
```
markmandel, line 49: `terraform fmt` changed this - figured it made sense to leave it in.

install/terraform/modules/gke/cluster.tf
```diff
@@ -146,7 +146,7 @@ resource "google_container_cluster" "primary" {
 }
 
 resource "google_compute_firewall" "default" {
-  name    = "game-server-firewall-firewall-${local.name}"
+  name    = length(var.firewallName) == 0 ? "game-server-firewall-${local.name}" : var.firewallName
```
roberthbailey, line 149: Why this way instead of setting the default name in variables.tf?

markmandel: Ah! Excellent question. Had to do it this way, because you can't use use [Terraform string templates](https://www.terraform.io/docs/configuration/expressions.html#string-templates) in the default values of `var` declarations in Terraform.

Which took me a while to work out how to resolve!

roberthbailey: And we don't want the firewall name to be a fixed string by default?

markmandel: The default had the cluster name in it, so I figured it made sense to keep it that way.

(Probably to make it relatively unique, so it didn't collide with anything?)

markmandel: @aLekSer can speak to this, because you wrote the original version, from memory?

install/terraform/modules/gke/cluster.tf
```diff
@@ -146,7 +146,7 @@ resource "google_container_cluster" "primary" {
...
   project = local.project
   network = local.network
```
roberthbailey, line 152: Should the firewall open both tcp and udp? I seem to have both rules in my project, but maybe it's because I was testing out the [simple tcp example](https://github.com/googleforgames/agones/tree/master/examples/simple-tcp).

markmandel: I expect most people to only use UDP, so I wouldn't default to more open.

install/terraform/modules/gke/cluster.tf
aLekSer: Yes, that's true, in the other way we would have only one firewall, and creating the second cluster would fail.

PR #1743, opened 2020-08-07
Add note on GKE cluster versions

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

Added note for users who are using GKE, to make sure to check the different GKE versions that are available, and ensure they match up with their reliability needs.

N/A

N/A


site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -142,6 +142,16 @@ gcloud compute firewall-rules create game-server-firewall \
...
 ```
 
+{{< alert title="Note" color="info">}}
+Before planning your production GKE infrastructure, it is worth reviewing the 
+[different types of GKE clusters that can be created](https://cloud.google.com/kubernetes-engine/docs/concepts/types-of-clusters),
+such as Zonal or Regional, as each has different reliability and cost values, and ensuring this aligns with your
+Service Level Objectives or Agreements. 
+
+This is particularly true for GKE masters, which can go down temporarily to adjust for cluster resizing,
+[automatic upgrades](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades) and potentially crashes.
```
roberthbailey, line 152: maybe say repairs instead of crashes?

https://cloud.google.com/kubernetes-engine/docs/concepts/maintenance-windows-and-exclusions

markmandel: That's a nicer word :+1:

PR #1755, opened 2020-08-15
Helm default values in docs (related to controller limits) match chart default values

pgilfillan:
**What type of PR is this?**
/kind documentation

A small documentation fix: the values `agones.controller.numWorkers`, `agones.controller.apiServerQPS` and `agones.controller.apiServerQPSBurst` in the Helm chart were bumped up to 100, 400, 500 respectively in
https://github.com/googleforgames/agones/pull/563 but the docs still referred to the old default values.

N/A

N/A


pgilfillan:
It was marked as draft initially to wait for a co-worker to review it (company policy when using the company signed CLA), but that's done now.

I'm not sure why the build failed the second time, but not the first. The only difference between the two was I changed the commit to use a different email, so I assume it's not related to the change itself, but I don't have permission to retry the job. I can't see anything obvious in the logs of the failed job.

markmandel:
Looks like some flakiness in the sdk-conformance test. Re-running!

PR #1763, opened 2020-08-19
Adding initial README for new Unreal SDK

comerford:
**What type of PR is this?**

/kind documentation

Porting over the README for the Unreal SDK from Netspeak Games with a few tweaks, intent is to add to this once I actually migrate to this SDK with notes on usage. This preserves the documentation from the original repo and adds some notes about the migration and the fact that the SDK is new in 1.8.0


domgreen:
Thanks @comerford we actually moved the `README.md` under the docs that are generated for the site at the request of @markmandel 

I personally am a fan of having a readme in the plugin but would mean that the docs and the readme need to be kept in sync.

See https://github.com/googleforgames/agones/blob/master/site/content/en/docs/Guides/Client%20SDKs/unreal.md as this is where it was ported to.

comerford:
ah, no worries then - I will close this out :)

domgreen:
> ah, no worries then - I will close this out :)

I would be interested in seeing any updates you made :D always good to get a second pair of eyes

comerford:
I am currently installing the plugin, if I hit any speed bumps I will circle back to the main docs and update for sure - thanks for this, love that it is more feature complete and the game devs are super happy that it looks more Unreal-native

PR #1764, opened 2020-08-19
Add TCPUDP protocol

### LINKED ISSUE #1523: Support using the same port for both TCP/UDP forwarding
# Is your feature request related to a problem? Please describe.
SRCDS uses both TCP/UDP on the same port where the TCP is for RCON access, but the dynamic port policy will also assign a random port for both TCP RCON port and main game communication UDP port. This is really infuriating that I have to resort to using static allocation policy.

# Describe the solution you'd like
**Introduce a new protocol "Any" for GameServer, that will allocate one port that listen on both TCP and UDP**. If that semantically doesn't sound, because maybe in the future there will be more protocol supports, then make it an array:
```yaml
      ports:
        - name: srcds
          containerPort: {{ $.Values.srcds.port }}
          protocol:
            - UDP
            - TCP
          portPolicy: Static
          hostPort: {{ $.Values.srcds.port }}
```

Bmandk:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix

This feature adds a protocol for TCPUDP. Some servers, especially legacy servers, use both TCP and UDP protocol on the same hostPort. When a user specifies a TCP and UDP protocol separately with `portPolicy: Dynamic`, each port is allocated a different port number. With the TCPUDP protocol, the allocator allocates both TCP and UDP on the same port number.

Closes #1523

A lot of firsts for me in here (Golang, writing tests, doing a PR in general), so very open to feedback.
I am using my own gcr registry for the e2e image, as this needed a new server example. This should probably live in the official registry, and the references in the Makefiles should be changed before merging.


Bmandk:
The build failed because of make-docker step. Could this be related to the registries with the new server image?

examples/simple-tcpudp/main.go
```diff
@@ -0,0 +1,201 @@
...
+		sender, txt := udpReadPacket(conn, b)
+		parts := strings.Split(strings.TrimSpace(txt), " ")
+
+		switch parts[0] {
+		// shuts down the gameserver
+		case "EXIT":
+			// respond here, as we os.Exit() before we get to below
+			udpRespond(conn, sender, "ACK UDP: "+txt+"\n")
+			exit(s)
+		// turns off the health pings
```
aLekSer, line 101: Should we remove this comment? And the blank line below?

aLekSer:
It is failing with linter:
```
test/e2e/framework/framework.go:705: 705-743 lines are duplicate of `test/e2e/framework/framework.go:663-701` (dupl)
func (f *Framework) TcpUdpGameServer(namespace string) *agonesv1.GameServer {
```
By the way please run `make gen-api-docs` and commit an html file which I expect should be updated.

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -81,6 +81,7 @@ properties:
               enum:
```
aLekSer, line 81: Please update the description. Something like:
```suggestion
              title: Protocol being used. Defaults to UDP. TCP and TCPUDP are other options.
              type: string
              enum:
```

aLekSer (changes requested):
Please apply some comments, and run `make lint`, you can use `\\nolint:` directive to suppress an error message.

Bmandk:
Ah, just figured out I can `make test` locally, my bad (Maybe something that could be added to the contribution guide?)

Unsure why this is failing now though, the message is a bit unclear to me.

examples/simple-tcpudp/main.go
Bmandk: Was a leftover from initial code, I've removed it.

aLekSer:
This is failing because `install.yaml` should be updated:
Run `make gen-install` and add `install.yaml` to this commit, than you should have no error on `make test-install-yaml`.
I forgot to mention this in a previous comment.
```
make: *** [test-install-yaml] Error 1
69a70,71
>                             - TCPUDP
>                             - TCPUDP
317a320
>                     - TCPUDP
Makefile:286: recipe for target 'test-install-yaml' failed
+ /root/test.sh
+ cd ./sdks/csharp/test
+ dotnet test -c Release
```

test/e2e/framework/framework.go
```diff
@@ -117,6 +118,7 @@ func newFramework(kubeconfig string, qps float32, burst int) (*Framework, error)
 const (
 	kubeconfigFlag      = "kubeconfig"
 	gsimageFlag         = "gameserver-image"
+	gsimagetcpudpFlag   = "gameserver-image-tcpudp"
```
aLekSer, line 121:
```suggestion
	gsimageTcpUdpFlag   = "gameserver-image-tcpudp"
```

aLekSer:
New test is failing, I believe we could push the docker image to fix this E2E test:
```
--- FAIL: TestGameServerTcpUdpProtocol (144.59s)
    gameserver_test.go:748: Could not ping TCP GameServer: dial tcp 34.82.230.193:7578: connect: connection timed out
time="2020-08-19 20:49:10.627" level=info msg="Scaling fleet" fleet=simple-fleet-8b6lb patch="[{ \"op\": \"replace\", \"path\": \"/spec/replicas\", \"value\": 1 }]" scale=1
```

Bmandk:
> New test is failing, I believe we could push the docker image to fix this E2E test:
> 
> ```
> --- FAIL: TestGameServerTcpUdpProtocol (144.59s)
>     gameserver_test.go:748: Could not ping TCP GameServer: dial tcp 34.82.230.193:7578: connect: connection timed out
> time="2020-08-19 20:49:10.627" level=info msg="Scaling fleet" fleet=simple-fleet-8b6lb patch="[{ \"op\": \"replace\", \"path\": \"/spec/replicas\", \"value\": 1 }]" scale=1
> ```

It's actually failing on my local test as well, currently testing a new solution.

build/Makefile
```diff
@@ -58,6 +58,7 @@ KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 
 # Game Server image to use while doing end-to-end tests
 GS_TEST_IMAGE ?= gcr.io/agones-images/udp-server:0.21
+GS_TEST_IMAGE_TCPUDP ?= gcr.io/agones-dev-286708/tcpudp-server:0.1
```
aLekSer, line 61: Nice workaround. Please don't forget to switch to 
```suggestion
GS_TEST_IMAGE_TCPUDP ?= gcr.io/agones-images/tcpudp-server:0.1
```
Once your test would be stable.

examples/simple-tcpudp/gameserver.yaml
```diff
@@ -0,0 +1,36 @@
...
+  ports:
+  - name: default
+    portPolicy: Dynamic
+    containerPort: 7654
+    protocol: TCPUDP
+  template:
+    spec:
+      containers:
+      - name: simple-tcpudp
+        image: gcr.io/agones-dev-286708/tcpudp-server:0.1
```
aLekSer, line 29:
```suggestion
        image: gcr.io/agones-images/tcpudp-server:0.1
```

Bmandk:
So the problem is one of two scenarios.

The first could be that the TCP message is sent when the timeout is reached. The server receives the ping and responds, but since the timeout is reached in the test client, the client closes at the same time. 

The second possibility is that the message is sent, but for some reason my server only responds when the client disconnects, but I am not sure why that would happen. Maybe my server logic is flawed, and could use an extra pair of eyes on the tcpudp example server.

Below are logs from my local server, where the client connects, and after the timeout (set to 60 seconds here) the message is received (or at least logged).

e3jBret 1

Bmandk:
Unsure why the TCP dial is failing on the cloud build, it is working on my own test cluster. The logs don't give much, so are hard to debug for me, and not sure how to move forward from here.

examples/simple-tcpudp/Makefile
```diff
@@ -0,0 +1,46 @@
+# Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1: nit: 
```suggestion
# Copyright 2020 Google LLC All Rights Reserved.
```

examples/simple-tcpudp/gameserver.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1: As I go through, all the 2019 -> 2020 :smile:

examples/simple-tcpudp/main.go
```diff
@@ -0,0 +1,196 @@
+// Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1: You might hate me for this - but does it make sense to pass a flag to https://github.com/googleforgames/agones/blob/master/examples/simple-udp/main.go and when enabled, turn on a TCP listener as well?

Rather than going to all the trouble to create a whole new exampe?

install/terraform/modules/aks/aks.tf
```diff
@@ -127,6 +127,20 @@ resource "azurerm_network_security_rule" "gameserver" {
   network_security_group_name = azurerm_network_security_group.agones_sg.name
 }
 
+resource "azurerm_network_security_rule" "gameserver" {
```
markmandel, line 130: Can we remove this for now?

Two reasons:
1) It's a bit outside the scope of this PR I think. Likely better handled in it's own PR.
2) We likely don't want to default to having TCP open in firewall rules. We should probably look at how we can make firewall config through terraform configurable (and hence, it's own PR)

There's probably a longer discussion that needs to be had around Terrform config and what should be default and what shouldn't be - and I don't want to see it hold up your work here. (This is also why smaller PRs are better in general).

If we want to change things for the e2e cluster, let's add it [here](https://github.com/googleforgames/agones/blob/master/build/terraform/e2e/module.tf), with an extra `gke-tcp-game-server-firewall` firewall configuration. This is the script that is run to create our e2e cluster and it's configurations.

install/terraform/modules/eks/eks.tf
```diff
@@ -47,6 +47,15 @@ resource "aws_security_group" "worker_group_mgmt_one" {
       "0.0.0.0/0",
     ]
   }
+  ingress {
```
markmandel, line 50: Same here, let's do terraform updates in it's own PR.

pkg/gameservers/portallocator.go
```diff
@@ -148,6 +148,8 @@ func (pa *PortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.GameServer
 		if len(allocations) == amount {
 			pa.gameServerRegistry[gs.ObjectMeta.UID] = true
 
+			var extraPorts []agonesv1.GameServerPort
```
markmandel, line 151: Nice :+1:

test/e2e/framework/framework.go
```diff
@@ -117,6 +120,7 @@ func newFramework(kubeconfig string, qps float32, burst int) (*Framework, error)
 const (
 	kubeconfigFlag      = "kubeconfig"
 	gsimageFlag         = "gameserver-image"
+	gsimageTCPUDPFlag   = "gameserver-image-tcpudp"
```
markmandel, line 123: This might be a good reason to wrap the functionality into the gameserver-image - wdyt?

install/terraform/modules/aks/aks.tf
Bmandk: Makes perfect sense - I will remove the terraform changes and add it in the e2e config.

install/terraform/modules/eks/eks.tf
Bmandk: Will remove

install/terraform/modules/aks/aks.tf
Bmandk: But should the TCP rule not be allowed with this? Or does it work differently on the cloud build, which would cause the test to fail?

test/e2e/framework/framework.go
Bmandk: Yes, I was also not happy with the way this specific change was done.

My reasoning of making a new example was that there already is a udp and a tcp example. I'm on board with making it a flag instead. Will make a commit with this.

markmandel: Yeah, I 100% agree. Just thinking that we already have one image that we use for testing - seems like a lot of work to add another one.

@aLekSer wdyt?

aLekSer: OK, I am totally agree with having one example for testing, with a flag passed through to `simple-udp` to enable TCP listener. So we don't need this `gsimageTCPUDPFlag ` in this case, we could turn on the TCP in a specific E2E test.

roberthbailey: I wonder if we should change the same of our example servers (and consolidate them). Right now this flag defaults to `gcr.io/agones-images/udp-server:0.21` but it's going to be confusing if we add tcp functionality to the _udp_ server. What about renaming it to `simple-game-server` and collapsing https://github.com/googleforgames/agones/tree/master/examples/simple-tcp in with https://github.com/googleforgames/agones/tree/master/examples/simple-udp?

markmandel: Maybe call it something like "simple-server" ?

Works for me :+1:

markmandel: Ah, missing your comment @roberthbailey - "simple-game-server" also works for me as well. Don't mind. :+1:

Bmandk: Makes sense, although I'm a bit apprehensive to make the changes as there are likely to be a bunch of places it needs to change, and there could be some hidden ones I would be missing.

roberthbailey: The simplest might be to publish the image with two names for now until we clean up the other references.

markmandel: Also doesn't have to be done all in this PR as well. We can take multiple steps :+1:

Bmandk: For history sake - I've opened this PR to do exactly that: https://github.com/googleforgames/agones/pull/1771

Bmandk:
Now that #1771 has been merged to master, what would be the best approach? Merge master in here and write the tests, or make a completely new branch and PR?

markmandel:
I would probably be inclined to take your current branch, rebase it against master and either push it up to here, or start a new PR - up to you on what your easiest path forward is for you.

I wouldn't recommend using Github as your merging tool - better you have local control.

build/terraform/e2e/module.tf
```diff
@@ -62,3 +62,16 @@ resource "helm_release" "consul" {
     value = "ClusterIP"
   }
 }
+
+resource "google_compute_firewall" "tcp" {
```
Bmandk, line 66: Seem to be running into the same problem that the TCP ports are not open, so not sure why this is not working

aLekSer:
@googlebot I consent.

install/helm/agones/templates/crds/_gameserverspecvalidation.yaml
```diff
@@ -76,11 +76,12 @@ properties:
               - Static
               - Passthrough
             protocol:
-              title: Protocol being used. Defaults to UDP. TCP is the only other option
+              title: Protocol being used. Defaults to UDP. TCP and TCPUDP are other options
```
roberthbailey, line 79: nit: End the sentence with a period.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -71,6 +71,9 @@ const (
 	// This will mean that users will need to lookup what port has been opened through the server side SDK.
 	Passthrough PortPolicy = "Passthrough"
 
+	// ProtocolTCPUDP Protocol exposes the hostPort allocated for this container for both TCP and UDP
```
roberthbailey, line 74: nit: End comment sentence with a period.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -198,7 +201,7 @@ type GameServerPort struct {
 	ContainerPort int32 `json:"containerPort,omitempty"`
 	// HostPort the port exposed on the host for clients to connect to
 	HostPort int32 `json:"hostPort,omitempty"`
-	// Protocol is the network protocol being used. Defaults to UDP. TCP is the only other option
+	// Protocol is the network protocol being used. Defaults to UDP. TCP and TCPUDP are other options
```
roberthbailey, line 204: nit: End comment sentence with a period.

pkg/gameservers/portallocator.go
```diff
@@ -161,6 +163,27 @@ func (pa *PortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.GameServer
 				if p.PortPolicy == agonesv1.Passthrough {
 					gs.Spec.Ports[i].ContainerPort = a.port
 				}
+
+				// create a second port for UDP
```
roberthbailey, line 167: This comment says "udp" but the code inside the block creates a port with "-tcp" appended, which I find a bit confusing. Can you clarify this comment?

pkg/gameservers/portallocator_test.go
```diff
@@ -110,6 +110,17 @@ func TestPortAllocatorAllocate(t *testing.T) {
...
 		assert.Equal(t, 11, countTotalAllocatedPorts(pa))
+
+		// single port to two ports, tcpudp
+		gsCopy = fixture.DeepCopy()
+		gsCopy.Spec.Ports[0] = agonesv1.GameServerPort{Name: "tcpudp", PortPolicy: agonesv1.Dynamic, Protocol: agonesv1.ProtocolTCPUDP}
+
+		gs = pa.Allocate(gsCopy)
+		require.NotNil(t, gs)
+		assert.Equal(t, gsCopy.Spec.Ports[0].HostPort, gsCopy.Spec.Ports[1].HostPort)
+		assert.Equal(t, corev1.ProtocolTCP, gsCopy.Spec.Ports[0].Protocol)
```
roberthbailey, line 121: Can you also add asserts for the port names (since they have a specific format with this port policy)?

test/e2e/framework/framework.go
```diff
@@ -488,6 +497,63 @@ func SendUDP(address, msg string) (string, error) {
...
+	// use first tcp port
+	for _, p := range gs.Spec.Ports {
+		if p.Protocol == corev1.ProtocolTCP {
+			return SendGameServerTCPToPort(gs, p.Name, msg)
+		}
+	}
+	return "", errors.New("No UDP ports")
+}
+
+// SendGameServerTCPToPort sends a message to a gameserver at the named port and returns its reply
```
roberthbailey, line 517: nit: End comment sentence with a period.

test/e2e/framework/framework.go
```diff
@@ -488,6 +497,63 @@ func SendUDP(address, msg string) (string, error) {
...
+// returns error if no Ports were allocated or a port of the specified name doesn't exist
```
roberthbailey, line 518: nit: Start new sentence with a capital letter and end with a period.

test/e2e/framework/framework.go
```diff
@@ -488,6 +497,63 @@ func SendUDP(address, msg string) (string, error) {
...
+func SendGameServerTCPToPort(gs *agonesv1.GameServer, portName string, msg string) (string, error) {
+	if len(gs.Status.Ports) == 0 {
```
roberthbailey, line 520: I think it would be more robust to move the check below the for loop and instead check for port not being nil. That will catch both an empty ports array as well as a caller passing in a port name that isn't in the port array.

test/e2e/framework/framework.go
```diff
@@ -488,6 +497,63 @@ func SendUDP(address, msg string) (string, error) {
...
+	for _, p := range gs.Status.Ports {
+		if p.Name == portName {
+			port = p
+		}
+	}
+	address := fmt.Sprintf("%s:%d", gs.Status.Address, port.Port)
+	return SendTCP(address, msg)
+}
+
+// SendTCP sends a message to an address, and returns its reply if
```
roberthbailey, line 533: nit: End comment sentence with a period.

test/e2e/framework/framework.go
```diff
@@ -488,6 +497,63 @@ func SendUDP(address, msg string) (string, error) {
...
+// it returns one in 30 seconds
+func SendTCP(address, msg string) (string, error) {
+	conn, err := net.Dial("tcp", address)
+	if err != nil {
+		log.Fatalf("Could not Dial: %v", err)
+	}
+
+	err = conn.SetReadDeadline(time.Now().Add(30 * time.Second))
```
roberthbailey, line 541: nit: Should be "cuddled":

```
if err := conn.SetReadDeadline(...); err != nil {
  log.Fatalf(...)
}
```

test/e2e/framework/framework.go
```diff
@@ -488,6 +497,63 @@ func SendUDP(address, msg string) (string, error) {
...
+	if err != nil {
+		log.Fatalf("Timed out: %v", err)
+	}
+
+	defer func() {
+		err = conn.Close()
```
roberthbailey, line 547: `err` here is never used. Maybe add a log statement if err != nil?

test/e2e/framework/framework.go
```diff
@@ -488,6 +497,63 @@ func SendUDP(address, msg string) (string, error) {
...
+	}()
+
+	// writes to the tcp connection
+	fmt.Fprintf(conn, msg+"\n")
+
+	response, _ := bufio.NewReader(conn).ReadString('\n')
```
roberthbailey, line 553: This is ignoring the error from `ReadString`. Please handle the error case.

test/e2e/gameserver_test.go
```diff
@@ -702,6 +702,52 @@ func TestGameServerPassthroughPort(t *testing.T) {
...
 
+func TestGameServerTcpUdpProtocol(t *testing.T) {
+	t.Parallel()
+	gs := framework.DefaultGameServer(framework.Namespace)
+
+	gs.Spec.Ports[0].Protocol = agonesv1.ProtocolTCPUDP
+	gs.Spec.Ports[0].Name = "tcpudp-port"
+	gs.Spec.Template.Spec.Containers[0].Env = []corev1.EnvVar{{Name: "TCP", Value: "TRUE"}}
+
+	// gate
```
roberthbailey, line 713: This comment doesn't make sense to me. Can you expand on it or remove it?

test/e2e/gameserver_test.go
```diff
@@ -702,6 +702,52 @@ func TestGameServerPassthroughPort(t *testing.T) {
...
+	_, valid := gs.Validate()
+	assert.True(t, valid)
+
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(framework.Namespace, gs)
+	if err != nil {
+		assert.FailNow(t, "Could not get a GameServer ready", err.Error())
+	}
+
+	tcpPort := readyGs.Spec.Ports[0]
+	assert.Equal(t, corev1.ProtocolTCP, tcpPort.Protocol)
```
roberthbailey, line 723: Consider asserting the port name as well (or at least that the port name ends with the correct string).

test/e2e/framework/framework.go
Bmandk: Can you clarify what you mean by "cuddled"?

roberthbailey: Yes; it's what I put into the comment - you assign the error variable in the if statement to restrict the scope of `err` to just the if block. Then is the convention when the variable isn't needed outside of the if block to keep the variable's scope as small as possible.

aLekSer (approved):
My comments were resolved.

aLekSer:
`make test-install-yaml` have failed. Please run `make gen-install-yaml` and commit install.yaml.

build/terraform/e2e/module.tf
markmandel: This will need to be some variation of `gke-` because reasons -- but also makes me wonder if we should allow you to specific the protocols (UDP, TCP, etc) supported when creating the cluster and accompanying firewall?

build/terraform/e2e/module.tf
```diff
@@ -62,3 +62,16 @@ resource "helm_release" "consul" {
...
+  name    = "game-server-firewall-e2e-test-cluster-tcp"
```
markmandel, line 67:
```suggestion
  name    = "game-server-firewall-tcp"
```

markmandel (changes requested):
Had only one change that needs to be made, but looks like its otherwise good to go :+1:

markmandel (approved):
Just under the wire for this release!

PR #1765, opened 2020-08-19
Pass port into autoscaler url from webhook policy

andrewgrundy:
**What type of PR is this?**
>
> /kind breaking
>
/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The port in the URL created by createURL was hardcoded to 8000 rather than the port specified in the WebhookPolicy.
I have made the port default to 8000 when a port is not given and added a test to check the URL produced is correct when a port is given.

Closes #

### DIFF
--- a/pkg/fleetautoscalers/fleetautoscalers.go
+++ b/pkg/fleetautoscalers/fleetautoscalers.go
@@ -91,14 +91,19 @@ func buildURLFromWebhookPolicy(w *autoscalingv1.WebhookPolicy) (u *url.URL, err 
 		w.Service.Namespace = "default"
 	}
 
-	return createURL(scheme, w.Service.Name, w.Service.Namespace, *w.Service.Path), nil
+	return createURL(scheme, w.Service.Name, w.Service.Namespace, *w.Service.Path, w.Service.Port), nil
 }
 
 // moved to a separate method to cover it with unit tests and check that URL corresponds to a proper pattern
-func createURL(scheme, name, namespace, path string) *url.URL {
+func createURL(scheme, name, namespace, path string, port *int32) *url.URL {
+	var hostPort int32 = 8000
+	if port != nil {
+		hostPort = *port
+	}
+
 	return &url.URL{
 		Scheme: scheme,
-		Host:   fmt.Sprintf("%s.%s.svc:8000", name, namespace),
+		Host:   fmt.Sprintf("%s.%s.svc:%d", name, namespace, hostPort),
 		Path:   path,
 	}
 }
--- a/site/content/en/docs/Reference/fleetautoscaler.md
+++ b/site/content/en/docs/Reference/fleetautoscaler.md
@@ -92,7 +92,9 @@ The `spec` field is the actual `FleetAutoscaler` specification and it is compose
       - `namespace` is the kubernetes namespace where webhook is deployed. Optional
                       If not specified, the "default" would be used
       - `path` is an optional URL path which will be sent in any request to this service. (i. e. /scale)
+{{% feature publishVersion="1.9.0" %}}
       - `port` is optional, it is the port for the service which is hosting the webhook. The default is 8000 for backward compatibility. If given, it should be a valid port number (1-65535, inclusive).
+{{% /feature %}}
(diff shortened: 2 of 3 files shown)


pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -91,14 +91,19 @@ func buildURLFromWebhookPolicy(w *autoscalingv1.WebhookPolicy) (u *url.URL, err
...
 	}
 
-	return createURL(scheme, w.Service.Name, w.Service.Namespace, *w.Service.Path), nil
+	return createURL(scheme, w.Service.Name, w.Service.Namespace, *w.Service.Path, w.Service.Port), nil
 }
 
 // moved to a separate method to cover it with unit tests and check that URL corresponds to a proper pattern
-func createURL(scheme, name, namespace, path string) *url.URL {
+func createURL(scheme, name, namespace, path string, port *int32) *url.URL {
+	var hostPort int32 = 8000
```
markmandel, line 99: This should probably be 80, but there's no way we can change that without breaking backward compatibility for some people, is there?

markmandel: Looks like K8s defaults to 443 because certs
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#servicereference-v1-admissionregistration-k8s-io

So maybe 80 isn't a good idea :thinking: :shrug:

roberthbailey: It seems like no matter what would be best we need to stick with the existing number for compatibility.

aLekSer: We have already documented that if not specified, the port would be 8000.
https://agones.dev/site/docs/reference/fleetautoscaler/
```
service is a reference to the service for this webhook. Either service or url must be specified. If the webhook is running within the cluster, then you should use service. Port 8000 will be used if it is open, otherwise it is an error.
```

aLekSer (changes requested):
Overall is great, but please update document file:
https://github.com/googleforgames/agones/blob/1474c2097160e1b3fdf964f8e42f04c6faf3d239/site/content/en/docs/Reference/fleetautoscaler.md#L94
Add port value description.

aLekSer:
Thanks for this Pull Request, it is definitely should be merged soon, please update docs file and sign the CLA.

aLekSer (approved):
Thanks for the update.

aLekSer:
📝 Please visit https://cla.developers.google.com/ to sign.

site/content/en/docs/Reference/fleetautoscaler.md
```diff
@@ -92,6 +92,7 @@ The `spec` field is the actual `FleetAutoscaler` specification and it is compose
       - `namespace` is the kubernetes namespace where webhook is deployed. Optional
                       If not specified, the "default" would be used
       - `path` is an optional URL path which will be sent in any request to this service. (i. e. /scale)
+      - `port` is optional, it is the port for the service which is hosting the webhook. The default is 8000 for backward compatibility. If given, it should be a valid port number (1-65535, inclusive).
```
aLekSer, line 95: Basically this way:
```suggestion
{{% feature publishVersion="1.9.0" %}}
      - `port` is optional, it is the port for the service which is hosting the webhook. The default is 8000 for backward compatibility. If given, it should be a valid port number (1-65535, inclusive).
{{% /feature %}}
```

pkg/fleetautoscalers/fleetautoscalers.go
markmandel: Cool - just wanted to check.

site/content/en/docs/Reference/fleetautoscaler.md
markmandel: Looks like this might be the last thing before we can merge.

(Confirmed with @thisisnotapril that we can manually approve CLA) @andrewgrundy, you good to make the change?

markmandel:
Oops, I think you got caught as I was updating the e2e cluster. Let's try that again!

markmandel:
sdk-conformance flakiness :cry:

PR #1785, opened 2020-09-03
GolangCI-lint updating version to 1.30

aLekSer:
Add exportloopref which is a successor of obsoleted scopelint.
Disable unnecessaryDefer which are located right before return.

**What type of PR is this?**

 /kind cleanup

Updating Golangci-lint version and added a successor of scopelint which is obsolete.


markmandel (approved):
Nice!

PR #1794, opened 2020-09-05
Upgrade client-go and apimachinery to 0.16.15

markmandel:
**What type of PR is this?**
>

/kind breaking

> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Upgrading the supporting libraries to be inline with the updated Kubernetes version to 1.16.x.

Work on #1649

Worth noting that cache.WaitForCacheSync has changed its internal implementation to now use `err := wait.PollImmediateUntil(...)`, so there is no more implicit 100ms sleep before the sync. Therefore it didn't _actually_ wait to populate from Watch or allow events to fire - it just made room for it to occur.

So had to now using assert/require.Eventually() to make sure that the system is in a state that matches what we expect before testing.


markmandel:
sdk-conformance failure :cry:

go.mod
```diff
@@ -3,50 +3,42 @@ module agones.dev/agones
...
-	k8s.io/kube-openapi v0.0.0-20190709113604-33be087ad058 // indirect
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0
+	k8s.io/api v0.16.15
+	k8s.io/apiextensions-apiserver v0.16.15
+	k8s.io/apimachinery v0.16.15
+	k8s.io/client-go v0.16.15
 	k8s.io/utils v0.0.0-20200124190032-861946025e34 // indirect
 )
+
+replace google.golang.org/grpc v1.23.0 => google.golang.org/grpc v1.20.1 // apiserver updated grpc, but we aren't using that, so it's fine.
```
roberthbailey, line 44: We should update at some point; want to file a bug to follow up on this?

pkg/gameservers/controller_test.go
```diff
@@ -1452,30 +1458,43 @@ func TestControllerGameServerPod(t *testing.T) {
 	}
 
 	t.Run("no pod exists", func(t *testing.T) {
-		c, gs, _, stop, cancel := setup()
+		c, gs, _, _, cancel := setup()
```
roberthbailey, line 1461: It looks like stop is never used from `setup()` now so it can just be removed as a return var.

pkg/gameservers/portallocator_test.go
```diff
@@ -445,14 +447,17 @@ func TestPortAllocatorSyncDeleteGameServer(t *testing.T) {
 		return true, nl, nil
 	})
 
-	stop, cancel := agtesting.StartInformers(m, pa.gameServerSynced, pa.nodeSynced)
+	_, cancel := agtesting.StartInformers(m, pa.gameServerSynced, pa.nodeSynced)
```
roberthbailey, line 450: I didn't search for other uses of this function, but could it just have a single return value now?

markmandel: Good question - This does get used when we populated the informer with a List* fake, so it has value in those places -- just not where we populate or manipulate via fake Watches. So I think we should leave it as is.

go.mod
markmandel: Agreed - makes sense. We should take this as a motivator to keep grpc up to date :+1:

pkg/metrics/controller_test.go
```diff
@@ -141,22 +168,31 @@ func TestControllerGameServersTotal(t *testing.T) {
...
 	generateGsEvents(1, agonesv1.GameServerStateUnhealthy, "", c.gsWatch)
 
-	c.sync()
+	expected := 96
+	assert.Eventually(t, func() bool {
+		list, err := c.gameServerLister.GameServers(gs.ObjectMeta.Namespace).List(labels.Everything())
+		require.NoError(t, err)
+		return len(list) == expected
+	}, 5*time.Second, time.Second)
+	// write a good message if something goes wrong
```
aLekSer, line 177: Not sure I can follow this comment. Could you please remove or explain more what is it for?

pkg/metrics/controller_test.go
```diff
@@ -269,10 +340,36 @@ func TestControllerGameServersNodeState(t *testing.T) {
 	c.gsWatch.Add(gameServerWithNode("node2"))
 
 	c.run(t)
-	c.sync()
+	reader := metricexport.NewReader()
+
+	// wait until we have a some nodes and gameservers
```
aLekSer, line 345:
```suggestion
	// wait until we have some nodes and gameservers
```

pkg/gameservers/controller_test.go
markmandel: Just found where we do use this (and shows an example of what you mentioned below)
https://github.com/markmandel/agones/blob/66ce69bf64c1e58b0e38fd642556e282ee70d060/pkg/gameservers/controller_test.go#L1504-L1512

pkg/metrics/controller_test.go
markmandel: Updated comment, PTAL!

markmandel:
SDK conformance failure

roberthbailey (approved):

pkg/metrics/controller_test.go
aLekSer: Now it is much more verbose, I like it.

aLekSer (approved):
Thanks for this PR, tests also were refactored and look much better now.

PR #1801, opened 2020-09-08
Add lock to to sdk-conformance compare()

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Likely isn't an issue, but figure it wouldn't hurt, and probably should be there anyway.

N/A

N/A


aLekSer (approved):
Agree that this mutex could help.

PR #1805, opened 2020-09-11
Flaky: TestControllerGameServersNodeState

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

This involves two things:

(1) Adding a newFakeControllerWithMock to the metric test utilities to allow for pre-population -- which with the client-go updates make it far easier to write deterministic tests.

(2) Updated TestControllerGameServersNodeState with the new mock construction.

This included removing the c.run() command, which can run c.collect() at any point in time, and any number of times before the tests being checked, and replaced this a c.collect() so we can control exactly when metric collection occurs during the tests.

N/A

Ran this a few hundred times, no failures.

### DIFF
--- a/pkg/metrics/controller_test.go
+++ b/pkg/metrics/controller_test.go
@@ -30,7 +30,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


aLekSer (approved):
Thanks for the test fix, this func `c.collect()` looks much more appropriate for the tests now.

Issue #1779, opened 2020-09-01
Flaky SDK Conformance Tests


markmandel:
**What happened**:

For a long time the SDK Conformance tests will occasionally flake.

**What you expected to happen**:

Them to pass all the time.

**How to reproduce it (as minimally and precisely as possible)**:

That's the problem - they are flaky :smile: 

This bug is a place to track the flakiness responses, and see if we can see a common pattern.

**Anything else we need to know?**:

**Environment**:
- Agones version: Dev
- Kubernetes version (use `kubectl version`): 1.16
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): Helm
- Troubleshooting guide log(s):

Context: https://console.cloud.google.com/cloud-build/builds/b61c074d-d6cd-4f68-b289-74a2d0f7ad2b?project=258182270954
```
/usr/bin/make run-sdk-conformance-no-build SDK_FOLDER=go GRPC_PORT=9001 HTTP_PORT=9101 FEATURE_GATES=PlayerTracking=true TESTS=ready,allocate,setlabel,setannotation,gameserver,health,shutdown,watch,reserve,getplayercapacity,setplayercapacity,playerconnect,playerdisconnect,getplayercount,isplayerconnected,getconnectedplayers
/usr/bin/make ensure-image IMAGE_TAG=agones-build-sdk-node:2ca282f4dc BUILD_TARGET=build-build-sdk-image SDK_FOLDER=node
/usr/bin/make ensure-image IMAGE_TAG=gcr.io/agones-images/agones-sdk:1.9.0-b831bb7 BUILD_TARGET=build-agones-sdk-server-image
/usr/bin/make ensure-image IMAGE_TAG=agones-build-sdk-go:6a1f217343 BUILD_TARGET=build-build-sdk-image SDK_FOLDER=go
+ /root/sdktest.sh
+ cd /go/src/agones.dev/agones/test/sdk/nodejs
+ npm run testSDK
/usr/bin/make ensure-image IMAGE_TAG=gcr.io/agones-images/agones-sdk:1.9.0-b831bb7 BUILD_TARGET=build-agones-sdk-server-image

> @ testSDK /go/src/agones.dev/agones/test/sdk/nodejs
> node ./testSDKClient.js

attempting to connect
connected!
{"message":"Ready request has been received!","sdkName":"node","severity":"info","source":"*sdkserver.LocalSDKServer","time":"2020-09-01T10:31:19.057025478Z"}
[...]

PR #1808, opened 2020-09-11
CI: Wait for tests step before sdk-conformance

aLekSer:
Both these steps would trigger building with `go run` or `npm run` in
parallel on the same directories.

**What type of PR is this?**
>
> /kind breaking

/kind bug

CI is fluctuating and fails too frequently.

Work on #1779

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -259,6 +259,7 @@ steps:
   args: [ "-j", "5", "--output-sync=target", "run-sdk-conformance-tests"]
   waitFor:
     - build-images
+    - tests
 
 #
 # Cache the CPP conformance build directory, to speed up subsequent builds (considerably)


aLekSer:
Too many simultaneous E2E tests:

```
Setting up lock at path: LockE2E/.lock
Attempting lock acquisition
Handling connection for 8500
Lock acquisition failed: failed to acquire lock: Unexpected response code: 500 (invalid session "a700b588-431a-01e4-b6f1-99e7d9461afb")
```

markmandel:
This is looking like it's solving the issue? Although it does look like it pushes the CI time to completion from around 18 minutes to around 30 - which is a big jump (that being said, we're probably at that already with retries, so :shrug:)

Run it a few more times, and if it keeps passing, I say let's merge it. WDYT?

aLekSer:
Well, yes, longest running time is bad, we can fix this side effect somehow with caching. We can see SDK conformance would not come up at a later time, if it would be as frequent as now, we can remove this `"wait_for": "tests"` or might move tests which could not run together in a separate cloud build step, or add File Lock before running the same test twice, say NodeJS SDK test would wait for release of file lock in NodeJS SDK conformance test?

markmandel (approved):
Hasn't failed yet!

PR #1777, opened 2020-08-31
Added a new 'disableTLS' flag and changed 'disableMTLS' to only disab…

devloop0:
…le client auth.

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

**What this PR does / Why we need it**:
Adds an option to just disable client auth (so traditional TLS). (disableMTLS -> plainTLS, disableTLS -> full plaintext).

**Which issue(s) this PR fixes**:

Closes #

**Special notes for your reviewer**:

PR #1815, opened 2020-09-15
adding timeout to remote cluster allocate call and adding total timeout to allocate

### LINKED ISSUE #1700: Multicluster: Add gRPC dial timeout
**What happened**:

When using the multicluster allocator, dialling remote clusters which are unhealthy can cause allocation requests to block for 2 minutes+.

Some sensible default, or configuration option to control dial timeout of multicluster allocation gRPC requests.

Deploy a `gameserverallocationpolicy` that allocates to some blocking endpoint. We have observed this issue occurring with a DNS name that resolves to the IP of a google LB that has no configured upstreams (because the cluster has been taken down for maintenance, for example.)

```
apiVersion: multicluster.agones.dev/v1
kind: GameServerAllocationPolicy
metadata:
  name: test
spec:
  connectionInfo:
    allocationEndpoints:
    - my-blackhole-server.com
    clusterName: test
    secretName: test
  priority: 10
  weight: 100
```

Observe that allocation requests following this policy will block for ~150 seconds before responding.

Our use-case is slightly more involved than the minimum repro scenario, we'd like to have a second, low-priority, cluster configured as "failover" for when the first cluster is unhealthy or down for maintenance. The behaviour we've observed is that when the high-priority cluster blackholes requests (such is the default behaviour when a GKE cluster with GCP loadbalancers is taken down), allocation requests take ~150 seconds to succeed as they block waiting for the high-priority cluster to respond.

**Environment**:
- Agones version: `1.7.0` 
- Kubernetes version: `1.16.10`
- Cloud provider or hardware configuration: GKE + bare-metal

kdima:
**What type of PR is this?**
/kind feature

This PR adds an explicit timeout to the Allocate call. This timeout includes both the RPC itself and creating the connection. I also added the total timeout for Allocate retries. 
I think total timeout should be configurable as a command line parameter but I am not sure where to add that. Is cmd/allocator/metrics.go:43 struct the right place?

Closes #1700


markmandel:
Thanks for your contribution! Should this timeout be documented somewhere?

pkg/gameserverallocations/allocator.go
```diff
@@ -82,6 +82,13 @@ const (
 	batchWaitTime         = 500 * time.Millisecond
 )
 
+const (
+	// Timeout for an individual Allocate RPC call can take
+	remoteAllocateTimeout = 10 * time.Second
```
pooneh-m, line 87: Also the client timeout needs to be set on the request context: https://grpc.io/blog/deadlines/#go

pooneh-m (commented):
@kdima thanks for the change.

robbieheywood:
Looks really good Dima - +1 to making this a CLI argument if possible (we'd probably want to set this lower than the defaults here - maybe something like `1s`). Also, is there any way to add a metric for this (i.e. number of allocation timeouts)?

pkg/gameserverallocations/allocator.go
kdima: I think I am setting it https://github.com/googleforgames/agones/blob/cd28618557d90d2a7c009f2f7420790286c26b37/pkg/gameserverallocations/allocator.go#L153
Unless you mean something else?

pkg/gameserverallocations/allocator.go
```diff
@@ -82,6 +82,13 @@ const (
...
+	// Total timeout for allocation including retries
+	totalRemoteAllocationTimeout = 30 * time.Second
```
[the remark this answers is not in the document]
kdima: Thanks for catching this. Added.

pooneh-m (commented):
Thanks for the update. Can we add testing for it?
For the env flag and Helm config, here is an example of plumbing through a config value: https://github.com/googleforgames/agones/pull/1777/files

pooneh-m:
Thanks for the change. You are getting a linter issue:
```
pkg/gameserverallocations/controller_test.go:914:21: string `mocked` has 3 occurrences, make it a constant (goconst)
		expectedGSName := "mocked"
		                  ^
pkg/gameserverallocations/allocator.go:139: File is not `goimports`-ed (goimports)
		totalAllocationTimeout:      totalAllocationTimeout,
make: *** [lint] Error 1
```

cmd/allocator/metrics.go
```diff
@@ -34,19 +35,23 @@ const (
...
-	GCPProjectID      string
-	StackdriverLabels string
+	TLSDisabled            bool
+	MTLSDisabled           bool
+	PrometheusMetrics      bool
+	Stackdriver            bool
+	GCPProjectID           string
+	StackdriverLabels      string
+	TotalAllocationTimeout time.Duration
+	AllocationTimeout      time.Duration
```
pooneh-m, line 54: The naming is not clear when it is set for the agones-allocator service. It seems that it applies to a single cluster allocation. How about renaming it to *RemoteAllocationTimeout?

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -132,6 +132,10 @@ spec:
           value: {{ .Values.agones.allocator.disableMTLS | quote }}
         - name: DISABLE_TLS
           value: {{ .Values.agones.allocator.disableTLS | quote }}
+        - name: ALLOCATION_TIMEOUT
+          value: {{ .Values.agones.allocator.allocationTimeout | quote }}
+        - name: TOTAL_ALLOCATION_TIMEOUT
+          value: {{ .Values.agones.allocator.totalAllocationTimeout | quote }}
```
pooneh-m, line 138: Needs documentation: https://github.com/googleforgames/agones/blob/master/site/content/en/docs/Installation/Install%20Agones/helm.md

pkg/gameserverallocations/controller_test.go
```diff
@@ -1157,6 +1158,86 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
 			assert.Equal(t, expectedGSName, result.Status.GameServerName)
 		}
 	})
+	t.Run("Allocation is retried", func(t *testing.T) {
```
pooneh-m, line 1161: There are tests for retry:
- https://github.com/googleforgames/agones/blob/master/pkg/gameserverallocations/controller_test.go#L985
- https://github.com/googleforgames/agones/blob/master/pkg/gameserverallocations/controller_test.go#L1082

You want to add a test that remote allocation does not respond for the specified timeout. For example set the timeout to 10ms and send a request to a non existing service or to a service which sleeps for 20ms before sending a response.

pooneh-m (commented):
Thanks. It is coming together nicely.

cmd/allocator/metrics.go
kdima: Will change

pkg/gameserverallocations/controller_test.go
kdima: I have added a test that sets the total timeout to 0 and so allows us to validate that total timeout affects the retries.

kdima:
I have a couple of questions.
I can see that both controller instantiates allocator [here](https://github.com/googleforgames/agones/blob/43af4cafe110621ccae64a38326cdc0c7f007c5d/pkg/gameserverallocations/controller.go#L67)
Should I add a command line parameter to controller as well? 

In [Retry Loop](https://github.com/googleforgames/agones/blob/43af4cafe110621ccae64a38326cdc0c7f007c5d/pkg/gameserverallocations/allocator.go#L352) I return `ResourceExhausted` error code as this is the only way to break out of `Retry` function. Not sure if you like this flow?

Unrelated to this PR if  the allocation callback returns `ResourceExhausted` error the retries will immediately stop. I am not sure if this is intended behaviour.

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -235,7 +235,9 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.ping.udp.loadBalancerIP`                    | The [Load Balancer IP][loadBalancer] of the UDP Service load balancer. Only works if the Kubernetes provider supports this option.                | ""           |
 | `agones.ping.udp.loadBalancerSourceRanges`          | The [Load Balancer SourceRanges][loadBalancer] of the UDP Service load balancer. Only works if the Kubernetes provider supports this option.      | `[]`         |
 | `agones.allocator.disableMTLS`                      | Turns off client cert authentication for incoming connections to the allocator.            | `false`                |
-| `agones.allocator.disableTLS`                       | Turns off TLS security for incoming connections to the allocator. | `false`                |
+| `agones.allocator.disableTLS`                       | Turns off TLS security for incoming connections to the allocator.    | `false`                |
+| `agones.allocator.allocationTimeout`                | Remote allocation call timeout.                                      | `10s`                  |
```
kdima, line 239: I am not sure if this is the right place to add it as it is inside the `feature publishVersion="1.9.0" ` if statement

pkg/gameserverallocations/allocator.go
```diff
@@ -356,6 +367,11 @@ func (c *Allocator) allocateFromRemoteCluster(gsa *allocationv1.GameServerAlloca
 			}
 			break
 		}
+		select {
+		case <-ctx.Done():
+			return status.Errorf(codes.DeadlineExceeded, "remote allocation retry timeout exceeded")
+		default:
```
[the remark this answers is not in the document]
kdima: I changed the code a fair bit, see if it makes sense now

site/content/en/docs/Installation/Install Agones/helm.md
pooneh-m: It should similarly go out with 1.10 release. The 1.9-RC is already merged.

pkg/gameserverallocations/allocator.go
```diff
@@ -336,13 +342,19 @@ func (c *Allocator) allocateFromRemoteCluster(gsa *allocationv1.GameServerAlloca
 	request.MultiClusterSetting.Enabled = false
 	request.Namespace = connectionInfo.Namespace
 
+	ctx, cancel := context.WithTimeout(context.Background(), c.totalRemoteAllocationTimeout)
+	defer cancel() // nolint: errcheck
 	// Retry on remote call failures.
 	err = Retry(remoteAllocationRetry, func() error {
 		for i, ip := range connectionInfo.AllocationEndpoints {
+			select {
```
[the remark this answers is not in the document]
pooneh-m: Thanks! Makes sense.

pkg/gameserverallocations/allocator.go
```diff
@@ -336,13 +342,19 @@ func (c *Allocator) allocateFromRemoteCluster(gsa *allocationv1.GameServerAlloca
...
+			case <-ctx.Done():
+				return status.Errorf(codes.ResourceExhausted, "remote allocation retry timeout exceeded")
```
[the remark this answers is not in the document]
kdima: This is the weird part. 
```
In Retry Loop I return ResourceExhausted error code as this is the only way to break out of Retry function. Not sure if you like this flow?
```
Logically it should be DeadlineExceeded but than `Retry` function will just keep retrying as it considers `DeadlineExceeded` a retryable error. The only error code that breaks you out of `Retry` is `ResourceExhausted` which makes sense I think.

kdima:
One significant bit that I think is left is whether I should add same flags to controller, as controller also instantiates allocator [here](https://github.com/googleforgames/agones/blob/43af4cafe110621ccae64a38326cdc0c7f007c5d/pkg/gameserverallocations/controller.go#L67)
I am not quite clear what `allocator` instantiated in controller does and how it is different from allocator instantiated in `main.go`.

pkg/gameserverallocations/allocator.go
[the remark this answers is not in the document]
pooneh-m: [Here is the logic](https://github.com/googleforgames/agones/blob/master/pkg/gameserverallocations/allocator.go#L558) that makes `ResourceExhausted` non-retryable.

However, `DeadlineExceeded` is a retryable error.  If the parent context's timeout has reached the **total** allocation timeout, even if the retry happens, it fails before attempting to remote allocate. The child timeout, should be retried for `DeadlineExceeded` error. We should add a unit test for the inner timeout and count the attempts.

pkg/gameserverallocations/controller_test.go
```diff
@@ -1157,6 +1158,75 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
...
+			Spec: allocationv1.GameServerAllocationSpec{
+				MultiClusterSetting: allocationv1.MultiClusterSetting{
+					Enabled: true,
+				},
+				Required: metav1.LabelSelector{MatchLabels: map[string]string{agonesv1.FleetNameLabel: fleetName}},
+			},
+		}
+
+		_, err = executeAllocation(gsa, c)
+		assert.Error(t, err)
```
pooneh-m, line 1227: Check the error code?

pooneh-m:
> One significant bit that I think is left is whether I should add same flags to controller, as controller also instantiates allocator [here](https://github.com/googleforgames/agones/blob/43af4cafe110621ccae64a38326cdc0c7f007c5d/pkg/gameserverallocations/controller.go#L67)
> I am not quite clear what `allocator` instantiated in controller does and how it is different from allocator instantiated in `main.go`.

For allocation there are two APIs, one provided by `agones-allocator` acting as a data-plane, with gRPC API and cert auth and another provided by `agones-controller` as an Extension API server, for which you can allocate using `kubectl create GameServerAllocation`. 

For multicluster allocation, the cross cluster allocation happens using `agones-allocator` that you have covered here. However, if a MatchMaker uses `GameServerAllocation` API from the controller, the default timeouts are used. Which API does your service use? If the controller, you should add the environment variables [here](https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/controller.yaml#L74).

pkg/gameserverallocations/allocator.go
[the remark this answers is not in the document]
kdima: I am a bit confused here. Do you mean that I should return `DeadlineExceeded` error from select statement?
If that is the case I don't think it is the best approach because even though the remote allocate will not get executed we will still keep doing retries and will wait the full retry timeout as defined in `remoteAllocationRetry` variable.
I.e. my understanding is that if I change the code to:
`return status.Errorf(codes.ResourceExhausted, "remote allocation retry timeout exceeded")`
following will happen:
1) call out to `remoteAllocationCallback` happens and lets say it takes up the whole `totalRemoteAllocationTimeout`. At this point no future calls to allocate should be made and we should ideally exit `allocateFromRemoteCluster` function
2) `Retry` log get executed, so we wait for some period of time and do a retry
3) select statement is executed, we return `DeadlineExceeded`
4) `Retry` logic sees `DeadlineExceeded` waits exponential backoff and executes out logic again.
This will happen 7 times and take up a bunch of time as exponential backoff keeps getting longer.
Or have I misunderstood your comment?

pkg/gameserverallocations/controller_test.go
kdima: done

kdima:
> > One significant bit that I think is left is whether I should add same flags to controller, as controller also instantiates allocator [here](https://github.com/googleforgames/agones/blob/43af4cafe110621ccae64a38326cdc0c7f007c5d/pkg/gameserverallocations/controller.go#L67)
> > I am not quite clear what `allocator` instantiated in controller does and how it is different from allocator instantiated in `main.go`.
> 
> For allocation there are two APIs, one provided by `agones-allocator` acting as a data-plane, with gRPC API and cert auth and another provided by `agones-controller` as an Extension API server, for which you can allocate using `kubectl create GameServerAllocation`.
> 
> For multicluster allocation, the cross cluster allocation happens using `agones-allocator` that you have covered here. However, if a MatchMaker uses `GameServerAllocation` API from the controller, the default timeouts are used. Which API does your service use? If the controller, you should add the environment variables [here](https://github.com/googleforgames/agones/blob/master/install/helm/agones/templates/controller.yaml#L74).

Thanks! That makes sense.

I mixed up the github ui flow and one of my replies is not showing up on the main page. This is my understanding of the logic inside the Retry loop https://github.com/googleforgames/agones/pull/1815/files#r497595728

pkg/gameserverallocations/allocator.go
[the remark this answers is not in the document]
pooneh-m: @kdima yes we should return `DeadlineExceeded`.
There are two contexts one is the parent context, with timeout `c.totalRemoteAllocationTimeout` and the second one is the child context created from the parent as `allocationCtx, cancel := context.WithTimeout(ctx, remoteAllocationTimeout)`. If the parent context is timed out, in the loop the select each time would go through the context.Done and not the default. If the child context is timed out, `grpcClient.Allocate` should return immediately. 

So in response to your comments:
No. "call out to remoteAllocationCallback" does not happen because each time in the loop ctx.Done will return the error. Since returning error is immediate we will not wait.

Please add a test case for inner timeout and count the attempts.

pooneh-m:
> I mixed up the github ui flow and one of my replies is not showing up on the main page. This is my understanding of the logic inside the Retry loop https://github.com/googleforgames/agones/pull/1815/files#r497595728

Thanks, I responded to your comment. Basically we should always do the right thing for the user and work around the code to get it working. `DeadlineExceeded` is the right error type for when the context is timed out.

pkg/gameserverallocations/allocator.go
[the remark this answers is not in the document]
kdima: You are correct the remoteAllocationCallBack will not be called, I missed that part. I still think that returning `DeadlineExceeded` results in the unexpected behaviour.
In this commit I have changed the error code to `DeadlineExceeded` and added debug print statement to my test case. In this test case the `totalRemoteAllocationTimeout` is set to 0, but the call to `allocateFromRemoteCluster` will always take approx 6 seconds. You can see that in the test case print out:
`Total time is 6.406179836s`
The reason it takes 6.4s seconds is that even though we return `DeadlineExceeded` the `Retry` logic will always retry 7 times, with timeouts starting from 100ms and exponential factor of 2 as defined in `remoteAllocationRetry` var. If you think that this is correct and expected behaviour I will keep the `DeadlineExceeded` error code. From my perspective this felt like unexpected behaviour.

kdima:
> > I mixed up the github ui flow and one of my replies is not showing up on the main page. This is my understanding of the logic inside the Retry loop https://github.com/googleforgames/agones/pull/1815/files#r497595728
> 
> Thanks, I responded to your comment. Basically we should always do the right thing for the user and work around the code to get it working. `DeadlineExceeded` is the right error type for when the context is timed out.

I think we are forced to return `ResourceExhausted` from inside the `Retry` loop, I commented here https://github.com/googleforgames/agones/pull/1815#discussion_r503831435

If my explanation in that comment makes sense and we still want to have  `DeadlineExeeded` returned to the user (which I can see why that makes sense) we could achieve this is if I return a special error so instead of
```
select {
  case <-ctx.Done():
    return status.Errorf(codes.ResourceExhausted, "remote allocation retry timeout exceeded")
  default:
}

```
Have something like
```
var TotalTimeoutExceededError = status.Errorf(codes.ResourceExhausted, "remote allocation retry timeout exceeded")
select {
  case <-ctx.Done():
    return TotalTimeoutExceededError
  default:
}

```
and then in `applyMultiClusterAllocation` function I can have something like

```
result, err = c.allocateFromRemoteCluster(gsa, connectionInfo, gsa.ObjectMeta.Namespace)
if err == TotalTimeoutExceededError {
  err = status.Errorf(codes.DeadlineExceeded, "remote allocation retry timeout exceeded")
}
if err != nil {
  c.loggerForGameServerAllocation(gsa).WithField("allocConnInfo", connectionInfo).WithError(err).Error("remote-allocation failed")
}
```

this feels quite hacky though.

pooneh-m:
> > > I mixed up the github ui flow and one of my replies is not showing up on the main page. This is my understanding of the logic inside the Retry loop https://github.com/googleforgames/agones/pull/1815/files#r497595728
> > 
> > 
> > Thanks, I responded to your comment. Basically we should always do the right thing for the user and work around the code to get it working. `DeadlineExceeded` is the right error type for when the context is timed out.
> 
> I think we are forced to return `ResourceExhausted` from inside the `Retry` loop, I commented here [#1815 (comment)](https://github.com/googleforgames/agones/pull/1815#discussion_r503831435)
> 
> If my explanation in that comment makes sense and we still want to have `DeadlineExeeded` returned to the user (which I can see why that makes sense) we could achieve this is if I return a special error so instead of
> 
> ```
> select {
>   case <-ctx.Done():
>     return status.Errorf(codes.ResourceExhausted, "remote allocation retry timeout exceeded")
>   default:
> }
> ```
> 
> Have something like
> 
> ```
> var TotalTimeoutExceededError = status.Errorf(codes.ResourceExhausted, "remote allocation retry timeout exceeded")
> select {
>   case <-ctx.Done():
>     return TotalTimeoutExceededError
>   default:
> }
> ```
> 
> and then in `applyMultiClusterAllocation` function I can have something like
> 
> ```
> result, err = c.allocateFromRemoteCluster(gsa, connectionInfo, gsa.ObjectMeta.Namespace)
> if err == TotalTimeoutExceededError {
>   err = status.Errorf(codes.DeadlineExceeded, "remote allocation retry timeout exceeded")
> }
> if err != nil {
>   c.loggerForGameServerAllocation(gsa).WithField("allocConnInfo", connectionInfo).WithError(err).Error("remote-allocation failed")
> }
> ```
> 
> this feels quite hacky though.

You can return a custom error detail, e.g. TotalTimeoutExceededError, with error code as DEADLINE_EXCEEDED and in the [retry logic](https://github.com/googleforgames/agones/blob/master/pkg/gameserverallocations/allocator.go#L558-L560), condition on the error to stop retrying.

kdima:
I added custom `ErrTotalTimeoutExceeded` that follow the pattern similar to the one from `ErrNoGameServerReady` this way error is reported as `DeadlineExceeded` which I agree is much better for the user. And we still have the test that verifies that no allocation calls are made after the total timeout is exceeded.

pkg/gameserverallocations/allocator.go
```diff
@@ -62,6 +62,8 @@ var (
 	ErrNoGameServerReady = errors.New("Could not find a Ready GameServer")
 	// ErrConflictInGameServerSelection is returned when the candidate gameserver already allocated
 	ErrConflictInGameServerSelection = errors.New("The Gameserver was already allocated")
+	// ErrTotalTimeoutExceeded is used to signal that total retry timeout has been exceeded and no additional retries should be made
+	ErrTotalTimeoutExceeded          = status.Errorf(codes.DeadlineExceeded, "remote allocation retry timeout exceeded")
```
pooneh-m, line 66: s/retry/total

pooneh-m:
> I added custom `ErrTotalTimeoutExceeded` that follow the pattern similar to the one from `ErrNoGameServerReady` this way error is reported as `DeadlineExceeded` which I agree is much better for the user. And we still have the test that verifies that no allocation calls are made after the total timeout is exceeded.

Looks great, thanks for the change! You are missing a test coverage for per request timeout. Please add a test that validates that retries happen when the allocation request has exceeds the `remoteAllocationTimeout` timeout limit.

pkg/gameserverallocations/controller_test.go
```diff
@@ -1157,6 +1159,147 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
...
+		st, ok := status.FromError(err)
+		assert.True(t, ok)
+		assert.Equal(t, st.Code(), codes.DeadlineExceeded)
+		assert.Equal(t, 0, calls)
+	})
+	t.Run("No allocations called after total timeout", func(t *testing.T) {
```
pooneh-m, line 1232: The description of the test should be updated?

pkg/gameserverallocations/controller_test.go
```diff
@@ -1157,6 +1159,147 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
...
+		c, m := newFakeControllerWithTimeout(10*time.Second, 10*time.Second)
```
pooneh-m, line 1233: Should we use `newFakeController` if the values are not used?

pkg/gameserverallocations/controller_test.go
```diff
@@ -1157,6 +1159,147 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
...
+		fleetName := addReactorForGameServer(&m)
+
+		calls := 0
+		c.allocator.remoteAllocationCallback = func(ctx context.Context, endpoint string, dialOpt grpc.DialOption, request *pb.AllocationRequest) (*pb.AllocationResponse, error) {
+			if calls == 0 {
+				calls++
```
pooneh-m, line 1239: you can move calls to the top of the func and condition of calls == 1. Please also leave a comment on what the function is doing.

pkg/gameserverallocations/controller_test.go
```diff
@@ -1157,6 +1159,147 @@ func TestMultiClusterAllocationFromRemote(t *testing.T) {
...
+				return nil, status.Errorf(codes.DeadlineExceeded, "remote allocation call timeout")
+			}
+			calls++
+			return &pb.AllocationResponse{}, nil
+		}
+
+		// Allocation policy reactor
+		secretName := clusterName + "secret"
+		m.AgonesClient.AddReactor("list", "gameserverallocationpolicies", func(action k8stesting.Action) (bool, k8sruntime.Object, error) {
```
pooneh-m, line 1248: Can you refactor the reactors to a method call?

pooneh-m (approved):
Great addition. Thank you!

pooneh-m:
It looks like your changes is causing `TestAllocatorCrossNamespace` to fail. Please take a look.

PR #1818, opened 2020-09-15
Pin npm autoprefixer package for site generation

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The 10.0.0 release of autoprefixer broke site generation, so pinning the version to a 9.x.x release.

N/A

This will unclog CI as well.

### DIFF
--- a/build/includes/website.mk
+++ b/build/includes/website.mk
@@ -37,8 +37,9 @@ site-static: ensure-build-image
 	# for some reason, this only work locally
 	docker run --rm $(common_mounts) --workdir=$(mount_path)/site $(DOCKER_RUN_ARGS) $(build_tag) \
 		bash -c "npm list postcss-cli || npm install postcss-cli"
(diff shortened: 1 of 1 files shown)


roberthbailey (approved):
At some point this is going to come back and bite us if we forget to ever check to see if they fix things (or figure out how to upgrade to post 10.0.0) but this seems like a reasonable short term fix. Can you open a tracking bug to get back onto a non-pinned version in the future?

markmandel:
Should we create a ticket to update this dependency -- or should the ticket be to upgrade hugo+docsy, and therefore the whole suite of dependencies?

(I expect at some point Hugo/Docsy will work out how to work with the new version)

roberthbailey:
Either way is fine, just as long as we don't stay pinned forever and forget about it.

aLekSer (approved):

PR #1825, opened 2020-09-18
Add player tracking and shutdown to the supertuxkart example server

sudermanjr:
Signed-off-by: Andrew Suderman <andrew@sudermanjr.com>

**What type of PR is this?**

/kind feature

Introduces a lot of changes to the Supertuxkart server example:

* Tails the logfile constantly
* Uses regex to match log lines for triggering actions
* adds tests for those regexes (makes adding them easier to do)
* Adds the ability to turn on player tracking. Will add the supertuxkart username to the list of players using the alpha sdk features
* Some cleanup to logging. Add a prefix called [wrapper] so you can tell which logs are which. Will output all of the stk logs to stdout so you can see them when it's running
* Detects when zero people are in the lobby and sends a shutdown to the sdk

If you want to test it out without having to build it yourself, I have a version pushed at `quay.io/fairwinds/supertuxkart-agones:0.0.13`


examples/supertuxkart/main.go
```diff
@@ -57,58 +69,118 @@ func main() {
...
-	for i := 0; i <= 10; i++ {
-		time.Sleep(time.Second)
 
-		logFile, err := os.Open(path.Join(home, logLocation))
-		if err != nil {
-			log.Printf(">>> Could not open server log: %v", err)
-			continue
-		}
+	t := &tail.Tail{}
+	for i := 0; i < 10; i++ {
```
roberthbailey, line 83: If you are tailing the file why keep the sleep & loop?

examples/supertuxkart/main.go
```diff
@@ -57,58 +69,118 @@ func main() {
...
-		println(">>> Output from server_config.log:")
-		print(output)
-
-		if strings.Contains(output, "Listening has been started") {
-			ready = true
-			log.Printf(">>> Moving to READY!")
+	}
+	defer t.Cleanup()
+	for line := range t.Lines {
+		fmt.Println(line.Text)
```
roberthbailey, line 96: why use `fmt` here instead of `log`?

examples/supertuxkart/main.go
```diff
@@ -57,58 +69,118 @@ func main() {
...
+	if playerJoin.MatchString(line) {
+		matches := playerJoin.FindSubmatch([]byte(line))
+		player := string(matches[1])
+		log.Printf("player %s joined\n", player)
+		return "PLAYERJOIN", &player
+	}
+	if playerLeave.MatchString(line) {
+		matches := playerLeave.FindSubmatch([]byte(line))
+		player := string(matches[1])
+
```
roberthbailey, line 175: nit: extra blank line

examples/supertuxkart/entrypoint.sh
```diff
@@ -17,4 +17,4 @@
 set -e
 set +x
 
-./wrapper -i "./cmake_build/bin/supertuxkart --server-config=$(pwd)/server_config.xml"
+./wrapper -i "./cmake_build/bin/supertuxkart --server-config=$(pwd)/server_config.xml" -player-tracking="$ENABLE_PLAYER_TRACKING"
```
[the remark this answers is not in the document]
sudermanjr: The golang buit-in `flag` library requires just one `-` - https://golang.org/pkg/flag/

https://golang.org/pkg/flag/

examples/supertuxkart/main.go
sudermanjr: Since the line is coming directly from supertuxkart, it already has a prefix on it.  Adding the wrapper logging to the general logging makes for double prefixes and will muddy the logs.

The general intent is to keep the supertuxkart logs visible, while adding in additional logging from the wrapper

examples/supertuxkart/main.go
sudermanjr: I left this in to ensure that the log file has time to get created by stk. My understanding is that tail will error if the file doesn't exist yet.

examples/supertuxkart/main.go
```diff
@@ -34,20 +35,31 @@ const logLocation = "/.config/supertuxkart/config-0.10/server_config.log"
...
+	log.SetPrefix("[wrapper] ")
 	input := flag.String("i", "", "the command and arguments to execute the server binary")
+
+	// Since player tracking is not on by default, it is behind this flag.
+	// If it is off, still log messages about players, but don't actually call the player tracking functions.
+	enablePlayerTracking := flag.Bool("player-tracking", false, "If true, player tracking will be enabled.")
 	flag.Parse()
 
-	fmt.Println(">>> Connecting to Agones with the SDK")
+	log.Println("connecting to Agones with the SDK")
```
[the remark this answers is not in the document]
sudermanjr: Not sure I agree, but happy to change it.

examples/supertuxkart/entrypoint.sh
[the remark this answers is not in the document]
roberthbailey: From that site: "One or two minus signs may be used; they are equivalent."

We should make the two flags consistent (the previous flag is passed with two minus signs): either update the `server-config` flag to use one or update the `player-tracking` to use two.

examples/supertuxkart/main.go
roberthbailey: That makes total sense. Can you add a comment above this line so that we don't lose that information after the PR merges?

examples/supertuxkart/entrypoint.sh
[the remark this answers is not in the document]
sudermanjr: The previous flag is passed with one dash

```
 -i "./cmake_build/bin/supertuxkart --server-config=$(pwd)/server_config.xml"
```

The `--` is the argument being passed to supertuxkart. The `-i` is the flag in the golang.

examples/supertuxkart/main.go
sudermanjr: No problem.

examples/supertuxkart/entrypoint.sh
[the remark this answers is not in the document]
sudermanjr: Here's the two flags defined in main.go

https://github.com/FairwindsOps/agones/blob/4af8a1f865017a34da7067510aec22318391733a/examples/supertuxkart/main.go#L39-L43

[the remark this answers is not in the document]
roberthbailey: ugh, i totally misread that. thanks for clarifying.

roberthbailey (approved):

roberthbailey:
```
--- FAIL: TestGameServerPodNoErrors (0.00s)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16d09cd]
```

PR #1831, opened 2020-09-24
[Doc] Mark multicluster allocation feature as stable

### LINKED ISSUE #1780: Change the multi-cluster allocation feature to stable version
Multi-cluster allocation is currently in [Beta version](https://github.com/googleforgames/agones/blob/master/site/content/en/docs/Guides/feature-stages.md). `GameServerAllocationPolicy` CRD and `agones-allocator` service APIs, which the feature depends on are already in stable version. 

There has been some usage and feedback for the feature that it seems we are ready to move forward with the feature to mark it as a stable feature.

Change the documented version for multi-cluster allocation to stable.

pooneh-m:
**What type of PR is this?**
>

 /kind documentation

Changing the multicluster allocation version to stable

Closes #1780

### DIFF
--- a/site/content/en/docs/Guides/feature-stages.md
+++ b/site/content/en/docs/Guides/feature-stages.md
@@ -27,7 +27,6 @@ The current set of `alpha` and `beta` feature gates are:
 {{% feature publishVersion="1.9.0" %}}
 | Feature Name | Gate    | Default | Stage | Since |
 |--------------|---------|---------|-------|-------|
-| Multicluster Allocation<sup>*</sup> | N/A | Enabled | `Beta` | 1.6.0 |
 | Example Gate (not in use) | `Example` | Disabled | None | 0.13.0 |
 | [Port Allocations to Multiple Containers]({{< ref "/docs/Reference/gameserver.md" >}}) | `ContainerPortAllocation` | Enabled | `Beta` | 1.7.0 |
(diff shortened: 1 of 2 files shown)


markmandel:
Freezing this, since I think we likely don't want to move something to stable while between RC+Stable releases.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -27,7 +27,6 @@ The current set of `alpha` and `beta` feature gates are:
 {{% feature publishVersion="1.9.0" %}}
 | Feature Name | Gate    | Default | Stage | Since |
 |--------------|---------|---------|-------|-------|
-| Multicluster Allocation<sup>*</sup> | N/A | Enabled | `Beta` | 1.6.0 |
```
roberthbailey, line 30: Would it be better to mark this as generally available / stable and leave it in the table instead of removing it?

pooneh-m: This table seems to be for non-stable features. @markmandel?

roberthbailey: Yeah, based on our docs you are correct.

For stable features:
"The feature is enabled and the corresponding feature gate no longer exists."

roberthbailey (approved):

roberthbailey:
After I clicked approve I realized that both of these changes to the website should be gated behind our next agones release (so that they don't become retroactively stable)

pooneh-m:
I'll send a followup PR to release gate it.

Issue #1037, opened 2019-09-04
Pass arbitrary metrics from gameserver -> opencensus


markmandel:
**Is your feature request related to a problem? Please describe.**
There are always game specific metrics that also need to be tracked. To be duplicating the effort that already exists in Agones to connect some kind of metrics SDK into a game server seems like a waste of effort. 

**Describe the solution you'd like**

Some kind of way of exposing OpenCensus (eventually Open Telemetry) to the game server binary through the SDK in some fashion. Honestly not sure what the best way is - requires some investigation.

One thought is to build our own abstracted metrics API that (may) be close to OpenCensus - this allows us to replace the metrics backend if we need to (i.e. when OpenCensus becomes OpenTelemetry), without having to impact users. This is more work for us, but gives us portability.

**Describe alternatives you've considered**
N/A

**Additional context**
https://opencensus.io/
https://opentelemetry.io/

geeckmc:
+1

markmandel:
@domgreen How do ingest without reinventing the wheel?

markmandel:
Capturing conversation from chat:

> Mark Mandel
The more we talk about this, the more I'm convinced this should be a standalone OSS project that has a sidecar (probably open telemetry) and grpc/rest interface (maybe rest+proto?) And proper game engine sdks that do appropriate Async/threaded operations to send data to keep it off the main loop (edited) 
>
> Mark Mandel
I don't think it has to be integrated with Agones at all. In fact it's probably more applicable if it isn't.

markmandel:
Also, one could potentially also argue #2716 covers enough of this to for this issue to also be covered as well.

Issue #1036, opened 2019-09-04
Metric: Game Server Frame Rate over time


markmandel:
**Is your feature request related to a problem? Please describe.**
Every dedicated game server needs to track if it's running at the optimal, stable framerate, and also be able to alert if something is failing. We don't have any functionality to accomplish this right now.

**Describe the solution you'd like**
Some kind of integration into the Game Server SDK to track framerate, that sends this data out as metrics, and an accompanying dashboard.

I see there being two potential solutions (naming / organisation could be changed - sacrificial draft):

1) `SDK.frame()` - needs to be called every frame. Counts the framerate and averages it over time. Basically we control how this is calculated. More work for us, but we control it. 

2) `SDK.frameRate(int framesPerSecond)` - needs to be called every second, with the current frames per second, as reported by the engine. This is less work for us, but each user will need to calculate it themselves - which not all engines do. (what happens if someone calls this more than once per second?)

We could also potentially have both? But I'm leaning towards No. 1 as it's easier all around for us, and likely easier to integrate into commercial engines as a drop in component. But 2 is also handy if users are integrating with us via a sidecar container with their game server.

We may also want to expose the calculated frameRate through the SDK as well - as both debugging and information for the gameserver as well.

**Describe alternatives you've considered**
Honestly, couldn't think of anything better?

**Additional context**
https://answers.unity.com/questions/64331/accurate-frames-per-second-count.html
https://answers.unrealengine.com/questions/813911/how-to-get-the-fps-in-c.html?sort=oldest

roberthbailey:
Would it make sense to build a more generic framework for collecting metrics rather than trying to plumb individual metrics through the SDK.

markmandel:
I think we will need this anyway - but for metrics that are ubiquitous, I think it makes sense to have a standard path for them to go through, so we can build layers on top of that - i.e. dashboards etc.

Or did you mean something different? Everything has to be collected from within the GameServer somehow. We could namespace the metrics part mind you, or some kind of separation from the main SDK package? Is that what you are referring to?

markmandel:
I'll add another note - I just wrote out a bunch of metrics tickets -- I am definitely not an expert in metrics, I just wanted us to track this metric information, and so wrote tickets to identify the work. I most _definitely_ defer to those that have more experience in metrics and monitoring to dictate how this should all be setup, and what is ideal. Please take this as a sacrificial draft.

roberthbailey:
/cc @bbf 

[...]

Issue #1035, opened 2019-09-04
Metrics: Players


markmandel:
Dependent on #1033 

**Is your feature request related to a problem? Please describe.**
Need to integrate player tracking into metrics for aggregate dashboards.

**Describe the solution you'd like**
Metrics tracking and dashboards on:
- [ ] Game Server capacities
- [ ] Number of players per GameServer
- [ ] Empty player slots per GameServer
- [ ] Player connection length (how long they stay connected)
- [ ] Any others I can't think of right now.

**Describe alternatives you've considered**
Forcing users to do this on their own. A lot of work for what should be common metric requirements for a large part of our user base.

**Additional context**

cyriltovena:
Metrics and tracking are usually different systems

geeckmc:
+1

markmandel:
I think we can close this, as we have tracking work in #2716 and there is actually also metrics for the alpha player tracking now as well.

markmandel:
I think we can close this as well, since #2765 did some of the work, and we have counts and lists currently underway.

Issue #1239, opened 2019-12-17
Game Server Allocation advanced filtering: player count, state, reallocation


markmandel:
Dependent on #1033

(Not a detailed design ticket, just putting down the basic idea to get the conversation started)

**Is your feature request related to a problem? Please describe.**
In both session based games and persistent games, it would be useful to be able to do something akin to:
_Just give me a GameServer that has room for 2 (or n number) players, regardless whether it is Allocated or not_

Which would pass back a GameServer if it had capacity available, and if not, would Allocate a whole new one (with enough initial capacity), and return that instead.

This would be useful for:
- Automating matchmaking backfill operations
- matchmaking players to persistent worlds that have capacities (#1034 helps here too)

**Describe the solution you'd like**

Some kind of adjustment to `GameServerAllocation` that accounts for capacity of players that are available. Maybe a way to choose what State(s) are in the Allocation pool? (Ready or Allocated??)

This probably requires much thought, and would want to tie into #1197 so they don't conflict with each other.

**Describe alternatives you've considered**
Having a totally different Allocation path for player capacity based allocations -- but then we will likely end up with a huge number of paths for different types of allocations.,

**Additional context**
Can't think of anything else.

markmandel:
## Objective

Already defined above


## Background

These are terms that are currently used in Player Tracking:



*   Player Capacity: Max number of players that could end up in a GameServer 
*   Player Count: The number of players that are in a GameServer currently
*   Player Availability: Capacity - Count. (this is new for this ticket)


## Requirements and Scale



*   Is a high throughput operation, so will need to have an implementation that allows for this.
*   Will need similar scaling characteristics to current Allocation policy.
*   Will sit behind a feature flag of “PlayerAvailableAllocation”
*   Will still need to adhere to `Packing` and `Distributed` optimisation patterns
*   Will still need to be filterable on `required` and `preferred` match labels/expressions


## Design Ideas

Very similar to what we did with [FleetAutoscaler](https://agones.dev/site/docs/reference/agones_crd_api_reference/#autoscaling.agones.dev/v1.FleetAutoscalerSpec), we add a `policy` section to the `GameServerAllocation` which defines the strategy that the GameServerAllocation will take when allocating. There can then be a subject configuration block under `policy` that would include any needed configuration parameters.

This should be extensible to other types of allocation strategies down the line, and also doesn’t require a breaking change to the existing allocation implementation.

[...]

PR #1841, opened 2020-10-07
[Docs] Adding proposal for exposing metrics per GameServer

domgreen:
Initial proposal for discussion ... TLDR idea is to use the SDK to send a small number of metrics from the DGS to the sidecar for exposing via OpenCensus. 

**What type of PR is this?**
/kind documentation

Proposal around exposing metrics (Agones/Abitary) from game servers via the sidecar.

Relates to:
#1035 
#1036
#1037

# Game Metrics from Dedicated Game Servers
## Summary

Engineers often want to observe "metrics" which can be defined as "raw measurements, generally with the intent to produce continuous summaries of those measurements" from within their Dedicated Game Servers.

A proposal to do this would be to allow the DGS to send a small number (100ish) metrics via the SDK to the running agones sidecar. This requires a simple addition to the sidecar to allow it to expose these metrics via the already used OpenCensus integrations. We can update the SDK to have a metrics sub API (much like alpha and beta stages) to support metric work.

Metrics from dedicated servers often break down into two categories, firtly metrics that Agones already knows about count of players, free capactity etc. and arbitary metrics from within the DGS itself such as frame rate, number of sessions, total rings collected etc.

The main concerns with exposing arbitary metrics is to not expose too many, not impact the running DGS and not to reinvent the wheel but choose the correct technologies to work together.

We could theoretically break this into two proposals:
- expose Agones related metrics (things the sidecar knows about)
- expose DGS metrics (things Agones doesnt need to know about)

### Related Links

- [Agones - Metrics: Players](https://github.com/googleforgames/agones/issues/1035)
- [Agones - Metric: Game Server Frame Rate Over Time](https://github.com/googleforgames/agones/issues/1036)
- [Agones - Pass arbitary metrics from gameserver -> opencensus](https://github.com/googleforgames/agones/issues/1037)

## Goals

- Expose metrics around the current connected players
- Expose arbitary metrics from within the dedicated game server
  - eg. frame rate, number of concurrent sessions, number of failed connections
  - to confirm if this is a desire of the Agones project
- Agnostic to the choice of running engine
  - eg. supports by all engines from C#, Rust through to Unreal and Unity
- Support the basics of Counters, Gauge (UpDownCounter) and labels on these time series
- Minimal impact on the Mem/CPU of the running DGS

## Non-Goals

- Expose metrics around the infrastrucutre usage of the DGS
  - this can be acheived with existing projects
- Reinvent an existing solution
- Support an infinate number of arbitary metrics
- Additional aggregation of metrics
- Storing metrics

## Proposal: Limited number of metrics exposed via sidecar

Initial proposal would be to allow the sidecar to expose metrics itself, initally this would be the specific to the PlayerTracking data such as current number of connected users, free space that can be allocated per DGS and other such metrics that are already known to Agones. This would use the existing OpenCensus (and later OTel) implementation that is being used within Agones to allow these metrics to be scraped by other projects such as Prometheus. Once these metrics are exposed we can simply add another `ServiceMonitor` as is currently done with the controller to communicate with the Prometheus Operator metrics are avaliable to scrape. This would probably require the addition of a agones-sidecar service to route the traffic or defer back to a `PodMonitor`.

With the metrics known to Agones this is simple as state above however, metrics that are not nativley known by Agones (frame rate/sessions etc.) we would need a way to expose these metrics via the SDK to the sidecar. Here, we could simply add a small API that supports the basics of metrics (Counters, Guages, Labels) that sends the data to the sidecar that is exposed in the same way as the Agones native metrics are exposed.

It would be up to the Game Engineers to call this API when data they wish to record changes and would therefore be this would allow them to record the metrics that they are interested in from within the context of their own games.

I would envisage this API being a subset of the Agones SDK such as Alpha and Beta sections currently are. I would also look to limit (could be configurable) the number of metrics a game could send to the sidecar this would reduce the burden on the sidecar and reduce explosion of metrics that are reported.

One advantage of this approach is that we end up storing metrics outside of the DGS and in the sidecar causing much less of a memory impact on the running game. We also create a basic API so that all game running in the Agones project can report metrics in the same way.

The main problem however, is wether this is actually the responsibility of Agones to report on metrics from within a running game server and if so do we then also expose future APIs around events and other aspects of obersvability? I would argue that a mature platform would allow the engineers running their game server to pick up key observability points that they are interested in but could be swayed that this is the responsibility of another project much like CPU/Mem can be obtained directly from kubernetes APIs without the need to intefer with the DGS iteself.

Pro:
- Small extension the Agones SDK
- SDK means all game engines can report in a uniform way
- Native Agones and arbitary metrics can be collected in the same way
- low impact on the memory of the DGS
- Prometheus scraping is already documented and advised as part of the Agones project
  - no need for extra tech as with the alternatives

Con:
- Adding the the SDK surface area
- Push based from the DGS to sidecar (not sure this is a con tbh)

### Expose metrics via logs

This would be my proposal for getting arbitary metrics from a DGS if we were to seperate the two concerns of Agones vs none Agones metrics. In that case I would expose `/metrics` in the sidecar as standard for the data known to Agones (connected users, spaces free etc.).

In this alternative we would use the inbuilt loggers from the engines to log in a given format that then log shippers would be able to turn into metrics. An example of this is the [Prometheus sink](https://vector.dev/docs/reference/sinks/prometheus) for <vector.dev> which would allow you to transform logs and expose them to Prometheus.

The main concern around this is that we probably would not be able to form a decent standard of how to log metrics and would be up to the engineers running the servers and their game teams to discuss the best approach per individual case.

Pros:
- small amount of work in sidecar to expose Agones metrics
- Agones deals with only Agones related metrics

Cons:
- more work for engineers wanting to get metrics from within DGS
- no nice API to program against
- would need to form a standard within each company to send metrics via logs
- some games engines (Unreal) have logs that are designed to be human readable therefore not JSON compatible

### Expose a OpenTelemitry/Prometheus endpoint from within the DGS

This appoach would be a more standardised way of exposing metrics to via a running service and would most probably be supported in engines such as vanilla Rust, Go, Javascript etc. but other games servers Unreal for example does not support the idea of running a web server within the game to expose these metrics.

This also doesnt take into consideration that sidecar has metrics such as the number of connected players, free space on the DGS etc. this means we would probably end up implementing a `/metrics` endpoint within the DGS and its sidecar.

Pros:
- pull based
- supported out of the box in some "engines"
- would leave the engineers to choose technologies (Prom, OTel etc.)
- no work on SDK needed

Cons:
- unsupported in game engines such as Unity and Unreal
- memory impact on the running DGS

### Expose metrics via an agreed file format

- [Prometheus Docs - Text Format](https://prometheus.io/docs/instrumenting/exposition_formats/)

This approach would instead of sending metrics over the wire could instead write metrics to a specific file mounted on the pod in a well known format see the link above. This could then be picked up by something to expose or ship the metrics to a needed place for aggregation.

Pros:
- less memory impact (assumed)

Cons:
- would be down to the engineer to expose it in the needed format
- above link only Prometheus format, may not suite all needs
- sidecar would be exposing a `/metrics` endpoint anyway

### DIFF
--- a/docs/proposals/20201006_game_metrics.md
+++ b/docs/proposals/20201006_game_metrics.md
@@ -1,0 +1,116 @@
+# Game Metrics from Dedicated Game Servers
+## Summary
+
+Engineers often want to observe "metrics" which can be defined as "raw measurements, generally with the intent to produce continuous summaries of those measurements" from within their Dedicated Game Servers.
+
+A proposal to do this would be to allow the DGS to send a small number (100ish) metrics via the SDK to the running agones sidecar. This requires a simple addition to the sidecar to allow it to expose these metrics via the already used OpenCensus integrations. We can update the SDK to have a metrics sub API (much like alpha and beta stages) to support metric work.
+
+Metrics from dedicated servers often break down into two categories, firtly metrics that Agones already knows about count of players, free capactity etc. and arbitary metrics from within the DGS itself such as frame rate, number of sessions, total rings collected etc.
+
+The main concerns with exposing arbitary metrics is to not expose too many, not impact the running DGS and not to reinvent the wheel but choose the correct technologies to work together.
+
+We could theoretically break this into two proposals:
+- expose Agones related metrics (things the sidecar knows about)
+- expose DGS metrics (things Agones doesnt need to know about)
+
+### Related Links
+
+- [Agones - Metrics: Players](https://github.com/googleforgames/agones/issues/1035)
+- [Agones - Metric: Game Server Frame Rate Over Time](https://github.com/googleforgames/agones/issues/1036)
+- [Agones - Pass arbitary metrics from gameserver -> opencensus](https://github.com/googleforgames/agones/issues/1037)
+
+## Goals
+
+- Expose metrics around the current connected players
+- Expose arbitary metrics from within the dedicated game server
+  - eg. frame rate, number of concurrent sessions, number of failed connections
+  - to confirm if this is a desire of the Agones project
+- Agnostic to the choice of running engine
+  - eg. supports by all engines from C#, Rust through to Unreal and Unity
+- Support the basics of Counters, Gauge (UpDownCounter) and labels on these time series
+- Minimal impact on the Mem/CPU of the running DGS
+
+## Non-Goals
+
+- Expose metrics around the infrastrucutre usage of the DGS
+  - this can be acheived with existing projects
+- Reinvent an existing solution
+- Support an infinate number of arbitary metrics
+- Additional aggregation of metrics
+- Storing metrics
+
+## Proposal: Limited number of metrics exposed via sidecar
+
+Initial proposal would be to allow the sidecar to expose metrics itself, initally this would be the specific to the PlayerTracking data such as current number of connected users, free space that can be allocated per DGS and other such metrics that are already known to Agones. This would use the existing OpenCensus (and later OTel) implementation that is being used within Agones to allow these metrics to be scraped by other projects such as Prometheus. Once these metrics are exposed we can simply add another `ServiceMonitor` as is currently done with the controller to communicate with the Prometheus Operator metrics are avaliable to scrape. This would probably require the addition of a agones-sidecar service to route the traffic or defer back to a `PodMonitor`.
+
+With the metrics known to Agones this is simple as state above however, metrics that are not nativley known by Agones (frame rate/sessions etc.) we would need a way to expose these metrics via the SDK to the sidecar. Here, we could simply add a small API that supports the basics of metrics (Counters, Guages, Labels) that sends the data to the sidecar that is exposed in the same way as the Agones native metrics are exposed.
+
+It would be up to the Game Engineers to call this API when data they wish to record changes and would therefore be this would allow them to record the metrics that they are interested in from within the context of their own games.
+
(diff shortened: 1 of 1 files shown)


aLekSer:
Well I think that this proposal might be better filled out as a GitHub ticket? So that we can gather the comments here in a discussion under the ticket? 
Structured similar to this one https://github.com/googleforgames/agones/issues/1239

domgreen:
> Well I think that this proposal might be better filled out as a GitHub ticket. So that we can gather the comments here in a discussion under the ticket. WDYT?
> Structured similar to this one [GameServerAllocation based on Player capacity](https://github.com/googleforgames/agones/issues/1239)

Will paste into the issue :+1: I do however, like keeping proposals (accepted/reject) around for future reference, has been really nice on other projects I've been on.

markmandel:
> Will paste into the issue +1 I do however, like keeping proposals (accepted/reject) around for future reference, has been really nice on other projects I've been on.

Thanks - I'm not necessarily against proposals being like this, but if we do this, there should be community consensus that that is the format we are taking, as well as the appropriate organisational structure to be in place to support it. Right now, this would force our maintainers and users to look in two spots at once for any design decisions, which would likely be confusing.

We also do keep design tickets around for posterity - you can see them here: https://github.com/googleforgames/agones/issues?q=is%3Aissue+is%3Aclosed+label%3Akind%2Fdesign

domgreen:
@markmandel sorry I misunderstood the initial request and put it in the description of the PR ... closing this in favour of https://github.com/googleforgames/agones/issues/1845 ... will double check that I didnt miss any docs around submitting design docs :+1:

markmandel:
No worries! Thanks so much!

PR #1844, opened 2020-10-13
Make the force_update option configurable in Helm/Terraform

### LINKED ISSUE #1767: Upgrading from 1.7.0 to 1.8.0 using the helm module for terraform fails with force_update=true
**What happened**:
Changed the agones_version to "1.8.0", ran a terraform plan (OK), terraform apply (failed) with the following messaging:
```bash
Error: failed to replace object: Service "agones-ping-http-service" is invalid: spec.clusterIP: Invalid value: "": field is immutable && failed to replace object: Service "agones-controller-service" is invalid: spec.clusterIP: Invalid value: "": field is immutable && failed to replace object: Service "agones-allocator" is invalid: spec.clusterIP: Invalid value: "": field is immutable

  on .terraform/modules/helm_agones/install/terraform/modules/helm3/helm.tf line 29, in resource "helm_release" "agones":
  29: resource "helm_release" "agones" {
```

To workaround this (after digging into similar issues reported with helm), I changed the [force_update setting](https://github.com/googleforgames/agones/blob/master/install/terraform/modules/helm3/helm.tf#L32) to false (to do this I had to fork the module and use my own version, because the setting is not configurable in terraform at present. This allowed me to successfully upgrade to 1.8.0:

```bash
helm list -n agones-system
NAME    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
agones  agones-system   8               2020-08-20 13:17:59.956502806 +0100 IST deployed        agones-1.8.0    1.8.0
```

Ideally, this upgrade would "just work". I assume the force_update setting is required for other reasons and setting it to false generally is not an option?

Changing the force_update flag to be configurable is one workaround that would be much less clunky than having to fork and use your own version of the module.

Based on what I have read elsewhere, the issue is that the setting of these immutable fields is blank and is generated when run first time, but then can't be blank for a second/subsequent run. Is another option to populate the settings perhaps?

I am not 100% sure if this would occur with a config that has always used helm3, this was originally a helm2 config and I migrated to helm3, so this may be a hangover from that process.

I checked, and if I revert to using the official module from Agones, I once again get the error, so I can repro locally. 

**Environment**:
- Agones version: 1.8.0
`Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}`
`Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.9-eks-4c6976", GitCommit:"4c6976793196d70bc5cd29d56ce5440c9473648e", GitTreeState:"clean", BuildDate:"2020-07-17T18:46:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}`

- Cloud provider or hardware configuration: AWS - EKS
- Install method (yaml/helm): helm/terraform
- Others:

comerford:
**What type of PR is this?**
/kind bug

As mentioned in #1767 there are immutability errors thrown regularly when making changed to the agones helm release via terraform when the force_update flag is true. Since it is still likely the desirable default, the easiest way to provide a path to a workaround is to make this configurable. The only other known way to resolve this issue is to delete and reinstall the entire release, not a desirable outcome

https://github.com/googleforgames/agones/issues/1767
Closes #1767 

This is a very minor change, and the default being the same as the original static value should make this a trivial non-impacting change. Hence, I have not hidden the new docs behind a version, and I think this is a candidate to be cherry-picked back to 1.8.x also.


site/content/en/docs/Installation/Terraform/aks.md
```diff
@@ -50,6 +50,7 @@ Configurable parameters:
 - log_level - possible values: Fatal, Error, Warn, Info, Debug (default is "info")
 - cluster_name - the name of the AKS cluster (default is "agones-terraform-example")
 - agones_version - the version of agones to install (an empty string, which is the default, is the latest version from the [Helm repository](https://agones.dev/chart/stable))
+- force_update - whether or not to force the replacement/update of resource (default is true, false may be required to prevent immutability errors when updating the configuration)
```
markmandel, line 53: This is awesome!

The only thing to change is this new `force_update` section -- can we get it wrapped in a `feature` shortcode
(Example: https://agones.dev/site/docs/contribute/#within-a-page) ? Then it will only show up for the next release version, and people won't get confused when pulling down release branches.

(Might be good to put it at the end of the list, so it doesn't mess too much with the formatting)

```
{{% feature expiryVersion="1.10.0" %}}
- force_update - whether or not to force the replacement/update of resource (default is true, false may be required to prevent immutability errors when updating the configuration)
{{% /feature %}}
```

Should do the trick :+1:  (also same below)

comerford: Done - I initially left out the short code because I would love if this could be backported to earlier versions given how small and isolated the change is, but I get the need to hide it in the meantime

site/content/en/docs/Installation/Terraform/aks.md
```diff
@@ -58,6 +58,9 @@ Configurable parameters:
 - gameserver_minPort - the lower bound of the port range which gameservers will listen on (default is "7000")
 - gameserver_maxPort - the upper bound of the port range which gameservers will listen on (default is "8000")
 - gameserver_namespaces - a list of namespaces which will be used to run gameservers (default is `["default"]`). For example `["default", "xbox-gameservers", "mobile-gameservers"]`
+{{% feature expiryVersion="1.10.0" %}}
```
markmandel, line 61: I am totally sorry - this should be "publishVersion" not "expiryVersion".

This is telling the system that it should not be shown, not that it _should_ be shown :facepalm:.

I totally told you the wrong thing.

comerford: switched, with a nice typo in the commit message :)

markmandel (approved):

PR #1850, opened 2020-10-18
Update prometheus and grafana

### LINKED ISSUE #1829: Metrics: link to helm repository is deprecated, install command as well
Metrics: link to helm repository is deprecated, install command as well

8398a7:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

This is because the prometheus and grafana charts are deprecated.

Closes #1829

I used the agones chart in 1.9.0 to verify this.
Existing grafana dashboards also seem to work fine.

### DIFF
--- a/site/content/en/docs/Guides/metrics.md
+++ b/site/content/en/docs/Guides/metrics.md
@@ -125,7 +125,10 @@ Prometheus is an open source monitoring solution, we will use it to store Agones
 Let's install Prometheus using the [helm stable](https://github.com/helm/charts/tree/master/stable/prometheus) repository.
 
 ```bash
-helm upgrade --install --wait prom stable/prometheus --namespace metrics \
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm repo update
+
+helm upgrade --install --wait prom prometheus-community/prometheus --version 11.16.2 --namespace metrics \
     --set server.global.scrape_interval=30s \
     --set server.persistentVolume.enabled=true \
     --set server.persistentVolume.size=64Gi \
@@ -191,7 +194,10 @@ kubectl apply -f ./build/grafana/
 Now we can install [grafana chart](https://github.com/helm/charts/tree/master/stable/grafana) from stable repository. (Replace `<your-admin-password>` with the admin password of your choice)
 
 ```bash
-helm install --wait --name grafana stable/grafana --version=5.0.13 --namespace metrics \
+helm repo add grafana https://grafana.github.io/helm-charts
+helm repo update
+
+helm upgrade --install --wait grafana grafana/grafana --version=5.7.10 --namespace metrics \
   --set adminPassword=<your-admin-password> -f ./build/grafana.yaml
 ```
 


build/grafana.yaml
```diff
@@ -37,6 +37,6 @@ datasources:
     datasources:
     - name: Prometheus
       type: prometheus
-      url: http://prom-prometheus-server.metrics.svc.cluster.local:80/
+      url: http://prometheus-server.metrics.svc.cluster.local:80/
```
8398a7, line 40: This is the effect of changing the name of the installation from prom to prometheus.

If there is no reason, I think the same as the chart name is good.
It is because the name can be shortened.

In the case of grafana, it's the same as the chart name.

site/content/en/docs/Guides/metrics.md
```diff
@@ -191,7 +194,10 @@ kubectl apply -f ./build/grafana/
 Now we can install [grafana chart](https://github.com/helm/charts/tree/master/stable/grafana) from stable repository. (Replace `<your-admin-password>` with the admin password of your choice)
 
 ```bash
-helm install --wait --name grafana stable/grafana --version=5.0.13 --namespace metrics \
+helm repo add grafana https://grafana.github.io/helm-charts
+helm repo update
+
+helm upgrade --install --wait grafana grafana/grafana --version=5.7.10 --namespace metrics \
```
8398a7, line 200: The same commands can be used to install and upgrade, so it has been modified to use upgrade.

site/content/en/docs/Guides/metrics.md
```diff
@@ -125,7 +125,10 @@ Prometheus is an open source monitoring solution, we will use it to store Agones
 Let's install Prometheus using the [helm stable](https://github.com/helm/charts/tree/master/stable/prometheus) repository.
 
 ```bash
-helm upgrade --install --wait prom stable/prometheus --namespace metrics \
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm repo update
+
+helm upgrade --install --wait prometheus prometheus-community/prometheus --version 11.16.2 --namespace metrics \
```
8398a7, line 131: The version was fixed in grafana, so I fixed this one too.

8398a7:
Just out of curiosity, won't grafana and prometheus be tied to the agones chart?

```yaml
...

grafana:
  enabled: true

  service:
    port: 3000
...
prometheus:
  enabled: true

  alertmanager:
    enabled: false
...
```

It's as if you can disable it by using enabled.
It would be easier to deploy if you could also include the dashboard configmap from the agones chart.

build/grafana.yaml
markmandel: I'm a bit hesitant to change this name, because of backward compatibility.

If people are pulling this file down from release branches as part of their install process, this could break things for them, no?

8398a7: That's certainly going to be a problem in that case.
I'll change it to the original release name.

8398a7: Thanks for your comment. Fixed.

markmandel (approved):
Just tested! Working great!

PR #1863, opened 2020-10-22
Add QPS settings to Allocation endpoints

### LINKED ISSUE #1852: Allocator throttled by default K8s Client requests per second
**What happened**:

Allocation through the Allocator gRPC service is showing far lower performance than what was previously tested through a `GameServerAllocation` directly through the Kubernetes API. (https://github.com/googleforgames/agones/pull/804 showed 80-160 QPS)

Instead this is showing 4-6 QPS, which is **far lower** 

Similar QPS as directly calling a `GameServerAllocation` when creating allocations through the gRPC API.

https://github.com/googleforgames/agones/tree/master/test/load/locust-files

I think I have a likely culprit for the slowdown through gRPC.

Through the controller we set the QPS and Burst values: https://github.com/googleforgames/agones/blob/master/cmd/controller/main.go#L137-L138

But we don't do the same thing on the allocator: https://github.com/googleforgames/agones/blob/master/cmd/allocator/main.go#L313-L332

Seems like we need a `agones.allocator.apiServerQPSBurst` and `agones.allocator.apiServerQPSBurst` on the Helm chart with similar values that are passed down into the Allocator service.

**Environment**:
- Agones version: 1.9.0
- Kubernetes version (use `kubectl version`): 1.16.0
- Others:

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Allocation endpoints where throttled to the default ~4qps for a Kubernetes client.

Matching the controller settings on standard QPS and Burst to allow higher throughput.

Closes #1852

Code can be reviewed, but we should wait until we get the benchmarking tool from @ilkercelikyilmaz to confirm throughput before merging.


roberthbailey (commented):
LGTM (it matches https://github.com/googleforgames/agones/blob/master/cmd/controller/main.go).

markmandel:
> LGTM (it matches https://github.com/googleforgames/agones/blob/master/cmd/controller/main.go).

Copy paste driven development :grin:

cmd/allocator/metrics.go
```diff
@@ -65,6 +71,8 @@ func parseEnvFlags() config {
 	viper.SetDefault(remoteAllocationTimeoutFlag, 10*time.Second)
 	viper.SetDefault(totalRemoteAllocationTimeoutFlag, 30*time.Second)
 
+	pflag.Int32(apiServerSustainedQPSFlag, 100, "Maximum sustained queries per second to send to the API server")
+	pflag.Int32(apiServerBurstQPSFlag, 200, "Maximum burst queries per second to send to the API server")
```
pooneh-m, line 75: viper.GetInt(apiServerBurstQPSFlag) to follow the pattern?
Is there any case that the defaults are used? Why not set it to the helm default?

pooneh-m (commented):
Thanks for the fix.

pooneh-m:
Can you please add the Helm config documentations as well?

cmd/allocator/metrics.go
markmandel: No reason - I literally just copied exactly what was in the controller/main.go - figured might as well keep it all consistent.
https://github.com/googleforgames/agones/blob/master/cmd/controller/main.go#L258

markmandel:
> Can you please add the Helm config documentations as well?

:facepalm: I knew I forgot something.

cmd/allocator/metrics.go
pooneh-m: I think we should use the same default across stack, if the default values are not meant to be different, regardless of the controller. 

If we set the default to 100 here and 400 as the environment variable, I can imagine looking at the code, one may assume the default is 100 as they have not set that on the environment variable.

pooneh-m:
Looks pretty good. Thanks for the change!

cmd/allocator/metrics.go
roberthbailey: I also thought this was confusing, but it's consistent with the agones controller. Maybe a cleanup there would be better than propagating confusing dueling default values?

markmandel: Do we feel we should cleanup here, or do it in a subsequent PR? I don't mind either way tbh.

pooneh-m: For allocator, because we are making the change in this PR, we should fix it in this PR. For controller, we can fix it in a later PR.

markmandel:
Just did a test with the tooling from #1867 

```
root@fba8ccd77f67:/go/src/agones.dev/agones/test/load/allocation/grpc# TESTRUNSCOUNT=1 ./runAllocation.sh 40 100
Run number:  1
started: 2020-10-27 23:25:28.81615566 +0000 UTC m=+0.000918033
(failed(client=2,allocation=79): rpc error: code = Unknown desc = error updating allocated gameserver: Operation cannot be fulfilled on gameservers.agones.dev "load-test-fleet-bx24p-t7sd9": the object has been modified; please apply your changes to the latest version and try again
...
(failed(client=6,allocation=90): rpc error: code = Unknown desc = error updating allocated gameserver: Operation cannot be fulfilled on gameservers.agones.dev "load-test-fleet-bx24p-z25b7": the object has been modified; please apply your changes to the latest version and try again
finished: 2020-10-27 23:26:31.579609788 +0000 UTC m=+62.764372203
```

So that was 3980 allocation in 1 minute 3 seconds, so 63 QPS. (This is also from my home laptop while I'm on a hangout :smile:)

So looks like once I handle the above comments, this PR should be good to go!

cmd/allocator/metrics.go
markmandel: Thought I hit this, and apparently I missed it. Fix incoming!

pooneh-m (approved):

PR #1865, opened 2020-10-23
Add TypeMeta to GameServerAllocation when doing convertion

### LINKED ISSUE #1864: Allocating multicluster using GameServerAllocation API fails with missing Kind
**What happened**:
In multicluster allocation, when using GameServerAllocation API to allocate, error "Object 'Kind' is missing in" is returned on the successful call:

```
cat <<EOF | kubectl apply -oyaml -f -                                                                    
kind: GameServerAllocation
metadata:
  namespace: b
  name: hello
spec:
  multiClusterSetting:
    enabled: true
EOF

error: error when creating "STDIN": Object 'Kind' is missing in '{"metadata":{"creationTimestamp":null},"spec":{"multiClusterSetting":{"policySelector":{}},"required":{},"scheduling":"","metadata":{}},"status":{"state":"Allocated","gameServerName":"fleet-deployment-1-conf-1-l8wdr-khtvq","ports":[{"name":"default","port":7013}],"address":"34.82.237.195","nodeName":"gke-test-cluster-default-pool-4e30e684-l0vn"}}
'
```

No error

pooneh-m:
**What type of PR is this?**

/kind bug

Adds `TypeMeta` to `GameServerAllocation` response when doing conversion in multicluster allocation as the `kubectl ` command line fails if the `Kind` is not set.

Closes #1864

### DIFF
--- a/pkg/allocation/converters/converter_test.go
+++ b/pkg/allocation/converters/converter_test.go
@@ -231,6 +231,10 @@ func TestConvertGSAToAllocationResponse(t *testing.T) {
 		{
 			name: "status state is set to allocated",
 			in: &allocationv1.GameServerAllocation{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "allocation.agones.dev",
+					APIVersion: "v1",
+				},
 				Status: allocationv1.GameServerAllocationStatus{
 					State:          allocationv1.GameServerAllocationAllocated,
 					GameServerName: "GSN",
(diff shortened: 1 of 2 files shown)


pkg/allocation/converters/converter_test.go
```diff
@@ -355,6 +375,10 @@ func TestConvertAllocationResponseToGSA(t *testing.T) {
 				Ports: []*pb.AllocationResponse_GameServerStatusPort{},
 			},
 			want: &allocationv1.GameServerAllocation{
```
roberthbailey, line 377: Should there be a test where the typemeta isn't defined as part of `in` and we verify it gets added to `want`?

pooneh-m: `TestConvertAllocationResponseToGSA` is exactly doing it. The is an empty input and want has the `TypeMeta`.

roberthbailey: Sorry I missed that. I thought all of the additions were to input, but I see that this last one is for output.

roberthbailey (approved):

PR #1870, opened 2020-10-27
[Unreal SDK] Added a response code check to some functions

dotcom:
**What type of PR is this?**
/kind bug

Added a response code check to some functions about Unreal SDK.


domgreen (approved):
LGTM

Looks like a good addition :+1:

markmandel:
Only questions I have - is this a breaking change?

Other than that, I'm happy to push this through :+1:

dotcom:
@markmandel 
No. The user interface has not changed. It can still be used. This makes it less confusing for the user.
Users can execute an Error callback in case of a bad response code in addition to communication errors.

markmandel (approved):
seal of approval

PR #1871, opened 2020-10-27
Fix memory leak in client-go/workqueue

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Since the Agones state is constantly changing, we got hit by this memory leak in client-go rather badly:
https://github.com/kubernetes/kubernetes/issues/83895

A fix has been implemented, but is only available in 1.17.x client-go:
https://github.com/kubernetes/kubernetes/pull/83945

N/A

This PR does a backport of the fix into our `vendor` directly. Since in the 1.11.0 release we'll be upgrading to support Kubernetes 1.17.x I didn't do anything to ensure the fix would remain in place long term.

### DIFF
--- a/vendor/k8s.io/client-go/util/workqueue/delaying_queue.go
+++ b/vendor/k8s.io/client-go/util/workqueue/delaying_queue.go
@@ -178,6 +178,9 @@ func (q *delayingType) waitingLoop() {
 	// Make a placeholder channel to use when there are no items in our list
 	never := make(<-chan time.Time)
 
+	// Make a timer that expires when the item at the head of the waiting queue is ready
+	var nextReadyAtTimer clock.Timer
+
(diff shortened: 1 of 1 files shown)


markmandel:
@ilkercelikyilmaz if you are able, it would be great if you could run one of your long running tests to confirm?

ilkercelikyilmaz:
I ran the load test with this fix. The memory leak  re:time.NewTimer is gone. Also controller is now bringing up the game servers much faster (~3 mins as oppose to ~6 mins).

ilkercelikyilmaz (approved):
LGTM.
I ran thee load test. Leak is gone and now the controller performs much better.

pooneh-m (approved):

PR #1874, opened 2020-10-29
Add allocator resources

### LINKED ISSUE #1873: Adding allocator resources

The allocator currently does not allow you to specify resources.
This may result in resources being placed on insufficient nodes.

I will allow you to specify it in the chart.

**Additional context**

8398a7:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

I've made the allocator resources configurable.

Closes #1873 

### DIFF
--- a/site/content/en/docs/Installation/Install Agones/helm.md
+++ b/site/content/en/docs/Installation/Install Agones/helm.md
@@ -248,6 +248,7 @@ The following tables lists the configurable parameters of the Agones chart and t
 
 | Parameter                                           | Description                                                                                     | Default                |
 | --------------------------------------------------- | ----------------------------------------------------------------------------------------------- | ---------------------- |
+| `agones.allocator.resources`                        | Allocator pods resource requests/limit                                                          | `{}`                   |
 | `agones.allocator.apiServerQPS`                     | Maximum sustained queries per second that an allocator should be making against API Server      | `400`                  |
 | `agones.allocator.apiServerQPSBurst`                | Maximum burst queries per second that an allocator should be making against API Server          | `500`                  |
 | `agones.allocator.allocationTimeout`                | Remote allocation call timeout.                                      | `10s`                  |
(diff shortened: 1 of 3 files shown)


markmandel:
This LGTM! @pooneh-m can you see anything we should care about? (I don't think so, but figured I'd double check)

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -248,6 +248,7 @@ The following tables lists the configurable parameters of the Agones chart and t
 
 | Parameter                                           | Description                                                                                     | Default                |
 | --------------------------------------------------- | ----------------------------------------------------------------------------------------------- | ---------------------- |
+| `agones.allocator.resources`                        | Allocator pods resource requests/limit                                                          | `{}`                   |
```
pooneh-m, line 251: Would it be helpful to add a small section with an example of the value for `resources` or a reference link?

markmandel: Ooh, I like it (we should add it to the other config options. How about this:

```suggestion
| `agones.allocator.resources`                        | Allocator pods [resource requests/limit](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)                                                          | `{}`                   |
```

Which also has explanations and examples of resource limits and requests. Sound good?

pooneh-m: +1 also, I think a simple example would also help with how to set the value. Since it is a compound structure, it is not easy to see how the input should look like.

markmandel: Definitely wouldn't hurt - especially as we have a few config options line this. Maybe add an extra example to:

> Specify each parameter using the --set key=value[,key=value] argument to helm install. For example,

(Or is this something you would have to use a values.yaml for? Trying to find some helm docs we could even point to)

8398a7: As far as I know, resources often uses values.yaml.
I haven't seen many examples of using set to configure it.

What you see most often are commented out settings.

```yaml
resources: {}
  requests:
    # cpu: 1
    # memory: 256Mi
```

8398a7: I've added instructions on how to set the values.

PR #1877, opened 2020-10-29
Move the loadBalancerIP to helm installation as best effort

pooneh-m:
**What type of PR is this?**
>

> /kind cleanup

Change the `make install` to use allocator loadBalancerIP helm configuration as best effort. If the service is already installed and is just upgrading, then loadBalancerIP is reused and valid allocator certificate gets issued for the service. Otherwise, the current behavior remains.

Closes #

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -298,6 +298,7 @@ install: PING_SERVICE_TYPE := "LoadBalancer"
 install: ALLOCATOR_SERVICE_TYPE := "LoadBalancer"
 install: CRD_CLEANUP := true
 install: LOG_LEVEL := "debug"
+install: EXTERNAL_IP ?= $(shell $(DOCKER_RUN) kubectl get services agones-allocator -n agones-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 install: FEATURE_GATES ?= $(ALPHA_FEATURE_GATES)
 install: HELM_ARGS ?=
 install: $(ensure-build-image) install-custom-pull-secret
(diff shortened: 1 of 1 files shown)


build/Makefile
```diff
@@ -312,14 +313,14 @@ install: $(ensure-build-image) install-custom-pull-secret
...
+		--set agones.allocator.http.loadBalancerIP=$(EXTERNAL_IP) \
 		$(HELM_ARGS) \
 		agones $(mount_path)/install/helm/agones/
 
-update-allocation-certs: EXTERNAL_IP ?= $(shell $(DOCKER_RUN) kubectl get services agones-allocator -n agones-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 update-allocation-certs: NAMESPACE ?= default
 update-allocation-certs:
 	-mkdir -p $(agones_path)/build/allocation
-	$(MAKE) install HELM_ARGS="--set agones.allocator.http.loadBalancerIP=$(EXTERNAL_IP)"
+	$(MAKE) install
```
markmandel, line 323: nit: `install` could be a prerequiste of `update-allocation-certs`.

build/Makefile
```diff
@@ -312,14 +313,14 @@ install: $(ensure-build-image) install-custom-pull-secret
 		--set agones.controller.logLevel=$(LOG_LEVEL) \
 		--set agones.crds.cleanupOnDelete=$(CRD_CLEANUP) \
 		--set agones.featureGates=$(FEATURE_GATES) \
+		--set agones.allocator.http.loadBalancerIP=$(EXTERNAL_IP) \
 		$(HELM_ARGS) \
 		agones $(mount_path)/install/helm/agones/
 
-update-allocation-certs: EXTERNAL_IP ?= $(shell $(DOCKER_RUN) kubectl get services agones-allocator -n agones-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 update-allocation-certs: NAMESPACE ?= default
 update-allocation-certs:
```
markmandel, line 321:
```suggestion
update-allocation-certs: install
```

Instead of having it inline below.

markmandel (commented):
Small nit, but otherwise, looks like an excellent change!

build/Makefile
pooneh-m: Thanks for the suggestion. Applied the change.

markmandel (approved):
Nice!

PR #1897, opened 2020-11-12
Upgrade Go version to latest

markmandel:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**:

It had been a while, and there were some security fixes, so figured it was worth doing again.

**Which issue(s) this PR fixes**:

N/A

**Special notes for your reviewer**:

I've not incremented example image versions, since I figured this wasn't
a breaking change, and the new version could be picked up as other
changes happen to those images.

Issue #1261, opened 2020-01-07
Update Example Images with the latest everything


markmandel:
Bring Agones example images with all the latest things:
1. Languages used (for example, we're still using Go 1.11 in a bunch of places)
2. SDK
3. Operating system (for example, bring Alpine images up to 3.11)

To be completed:

- [x] simple-udp
- [x] allocator-service
- [x] simple-tcp
- [x] unity-simple
- [x] rust-simple
- [x] autoscaler-webhook
- [x] xonotic
- [x] cpp-simple
- [x] nodejs-simple
- [x] crd-client

markmandel:
Due to #1293 all the image versions got incremented. I figured it would make sense to do this work at the same time as we re-build each of these, and make sure they all still work (pretty sure the node one is currently broken)

aLekSer:
Also we could consider #909 - rebase images from alpine to gcr.io/distroless/static ?

markmandel:
> Also we could consider #909 - rebase images from alpine to gcr.io/distroless/static ?

I think the spirit of that ticket was more for the Agones images themselves.

I don't know if we want to do that for example images? Using Alpine does make it easier to introspect what is happening in the images (e.g. by running a shell in them), which is useful for learning.

Some images also run Debian (Xonotic, Supertuxkart), and moving them to scratch I expect would be super, super hard.

aLekSer:
/assign @akremsa
Alexey is working on this right now

markmandel:
I believe everything except for Unity should be complete.

markmandel:
Let's start this one allll over again :smile: 

Once #1900 gets completed, we should do the following updates to example images:

- [ ] Update Unity on the Unity Sample, possibly upgrade Ubuntu base image to latest LTS
- [ ] #1890 - Deprecate and remove the udp-server and tcp-server images
- [ ] Rebuild the autoscaler webhook example with new go version
- [ ] Rebuild the simple-game-server example with the new go version
- [ ] Update Super Tux Kart version (should have that AI connection fix now). Also has a new go version.
- [ ] Rebuild the Xonotic example with the new go version

markmandel:
Should we close this issue? It seems like a bit of a mess now.

It might be better to create new issues as needed instead. What do we think?

Marking as stale to see if anyone disagrees on closing the issue.

markmandel:
Marked stale last month. Going to close and we can reopen specific tickets as needed.

PR #1900, opened 2020-11-16
Updated Go to 1.14.12

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Security fixes really.

Specifically didn't upgrade to 1.15.x due to issues with SSL certificates on fleet autoscaler webhooks. See #1897.

I've also not incremented example image versions, since I figured this wasn't a breaking change, and the new version could be picked up as other changes happen to those images.

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -23,7 +23,7 @@ RUN apt-get update && \
 
 # install go
 WORKDIR /usr/local
-ENV GO_VERSION=1.14.4
+ENV GO_VERSION=1.14.12
 ENV GOPATH /go
 ENV GO111MODULE=on
 RUN wget -q https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
--- a/build/build-sdk-images/go/Dockerfile
+++ b/build/build-sdk-images/go/Dockerfile
@@ -20,7 +20,7 @@ RUN apt-get update && \
 
 # install go
 WORKDIR /usr/local
-ENV GO_VERSION=1.14.4
+ENV GO_VERSION=1.14.12
 ENV GO111MODULE=on
 ENV GOPATH /go
 RUN wget -q https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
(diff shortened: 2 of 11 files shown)


roberthbailey:
> I've also not incremented example image versions, since I figured this wasn't a breaking change, and the new version could be picked up as other changes happen to those images.

If this is to pick up security fixes, shouldn't we rebuild and push new images (with the next release)? Otherwise, what is the point of updating the go version. Something like the supertuxcart example image, which isn't under active development isn't going to get a new version on its own any time soon. And things like the simple-game-server aren't likely to be changed but are used in all of our examples and it would be good to have users running a secured version.

roberthbailey (approved):
Code changes look good, just want to resolve the question about container image versions.

markmandel:
> If this is to pick up security fixes, shouldn't we rebuild and push new images (with the next release)? Otherwise, what is the point of updating the go version. Something like the supertuxcart example image, which isn't under active development isn't going to get a new version on its own any time soon. And things like the simple-game-server aren't likely to be changed but are used in all of our examples and it would be good to have users running a secured version.

If I'm honest, I didn't do it, because it's a lot of busywork, and the security of an example images doesn't seem to me to be a huge priority? So figured, it would be taken care of on a best effort basis, rather than being a blocker for a release.

SuperTuxKart should actually get updated - a new version came out a little while ago, I believe with the AI connection fixes. I don't think Xonotic has though.

roberthbailey:
That's fair. How about filing a help-wanted issue to update the example images? Collecting the list of things that need to be rebuilt would be pretty straightforward.

markmandel:
Sounds like a plan - do we restart https://github.com/googleforgames/agones/issues/1261 or start a new one? :smile:

roberthbailey:
That issue is sooooo close.... But since it's already open, I think we should just re-use it.

markmandel:
How's this?
https://github.com/googleforgames/agones/issues/1261#issuecomment-729176894

roberthbailey:
>How's this?
> #1261 (comment)

Great. Thanks!

PR #1902, opened 2020-11-17
Adding rest to allocation endpoint

### LINKED ISSUE #1495: grpc-gateway powered REST API for MultiCluster Allocation

For users for which gRPC isn't a viable option, it would be handy to have a REST based API as well for the gPRC API.

We use grpc-gateway to generate the REST API, but also use the same certificates to authenticate the endpoint as we would for the gRPC endpoint.

We can also generate a swagger.json so that people can generate REST clients as well.

Having no REST endpoint. But we're seeing bugs around SSL cert management in other languages, and this would provide a workaround for any other issues we have in the future.

**Additional context**

I would recommend completing #1331 first.

It might also be worth updating grpc-gateway at some point as well, we're quite a few releases behind.

kdima:
/kind feature

This should solve https://github.com/googleforgames/agones/issues/1495 

Closes #1495 

We are running both grpc and rest on the same port so there is no need to update the services yaml.


cmd/allocator/main.go
```diff
@@ -179,10 +185,31 @@ func main() {
...
+	server := &http.Server{
+		Addr:      ":8443",
+		TLSConfig: cfg,
+		Handler:   grpcHandlerFunc(grpcServer, mux),
+	}
+
 	go func() {
-		err := grpcServer.Serve(listener)
-		logger.WithError(err).Fatal("allocation service crashed")
+		err = server.ListenAndServeTLS("", "")
```
markmandel, line 211: Just to confirm - this is blank here, because the TLS config is handled above?

cmd/allocator/main.go
```diff
@@ -64,6 +63,18 @@ const (
 	sslPort = "8443"
 )
 
+// grpcHandlerFunc returns an http.Handler that delegates to grpcServer on incoming gRPC
+// connections or otherHandler otherwise. Copied from https://github.com/philips/grpc-gateway-example.
+func grpcHandlerFunc(grpcServer *grpc.Server, otherHandler http.Handler) http.Handler {
```
markmandel: Oh that's clever!

cmd/allocator/main.go
```diff
@@ -64,6 +63,18 @@ const (
 	sslPort = "8443"
 )
 
+// grpcHandlerFunc returns an http.Handler that delegates to grpcServer on incoming gRPC
+// connections or otherHandler otherwise. Copied from https://github.com/philips/grpc-gateway-example.
+func grpcHandlerFunc(grpcServer http.Handler, otherHandler http.Handler) http.Handler {
```
markmandel, line 68: Oh that's clever!

@roberthbailey we probably should have done this on the sdk sidecar! oh well!

markmandel (commented):
Looks like you are getting caught on some linting issues - will come back again when you get this sorted.

But looks like a nice addition :fire: 

Only other thought - would need to have some documentation here as well: https://agones.dev/site/docs/advanced/allocator-service/ (maybe also an example on multi-cluster as well).

But since this is draft, that may be your intent for when it's ready to go :+1:

cmd/allocator/main.go
kdima: Yeap, sicne we set custom `GetCertificate` function that gets used when the file names passed to this function are empty.
From the docs
```
// Filenames containing a certificate and matching private key for the
// server must be provided if neither the Server's TLSConfig.Certificates
// nor TLSConfig.GetCertificate are populated. 
```

cmd/allocator/main.go
kdima: The other option is apparently to use a dedicated muxer but since Agones doesn't have one pulled in atm I figured it is easier to use this approach.

kdima:
Thanks for the review!

> 
> Only other thought - would need to have some documentation here as well: https://agones.dev/site/docs/advanced/allocator-service/ (maybe also an example on multi-cluster as well).
> 
I have added docs, but not sure if it is sufficient.

cmd/allocator/main.go
```diff
@@ -179,10 +185,31 @@ func main() {
...
+		Addr:      ":8443",
+		TLSConfig: cfg,
+		Handler:   grpcHandlerFunc(grpcServer, mux),
+	}
+
 	go func() {
-		err := grpcServer.Serve(listener)
-		logger.WithError(err).Fatal("allocation service crashed")
+		err = server.ListenAndServeTLS("", "")
+		logger.WithError(err).Fatal("could not listen on REST")
```
markmandel, line 212: I feel like this error message could be better somehow... :thinking: 

How about?
```suggestion
		logger.WithError(err).Fatal("Unable to start HTTPS listener")
```
WDYT?

cmd/allocator/main.go
```diff
@@ -255,24 +282,10 @@ func readTLSCert() (*tls.Certificate, error) {
 // getServerOptions returns a list of GRPC server options.
 // Current options are TLS certs and opencensus stats handler.
 func (h *serviceHandler) getServerOptions() []grpc.ServerOption {
-	if h.tlsDisabled {
```
markmandel, line 258: Not sure I'm following why this was removed? I figure it's because the TLS details are above?

@pooneh-m will have better insight into if there is a better way to refactor this I expect.

site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -6,11 +6,11 @@ description: >
   Agones provides an mTLS based allocator service that is accessible from outside the cluster using a load balancer. The service is deployed and scales independent to Agones controller.
 ---
 
-To allocate a game server, Agones in addition to {{< ghlink href="pkg/apis/allocation/v1/gameserverallocation.go" >}}GameServerAllocations{{< /ghlink >}}, provides a gRPC service with mTLS authentication, called `agones-allocator`.
+To allocate a game server, Agones in addition to {{< ghlink href="pkg/apis/allocation/v1/gameserverallocation.go" >}}GameServerAllocations{{< /ghlink >}}, provides a gRPC and REST service with mTLS authentication, called `agones-allocator`.
```
markmandel, line 9: Our documentation gets updates on each commit to master, so we'll need to wrap the new documentation in `{{% feature %}}` shortcodes for the next release `1.11.0`, so it only shows up on the next release. 

You can see details here:

It may be easiest to take a copy of the existing doc and have two large `feature` sections, one for the old, and one for the new.

site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -140,6 +142,26 @@ go run examples/allocator-client/main.go --ip ${EXTERNAL_IP} \
...
+
+# allocator-client.default secret is created only when using helm installation. Otherwise generate the client certificate and replace the following.
+# In case of MacOS replace "base64 -d" with "base64 -D"
+kubectl get secret allocator-client.default -n "${NAMESPACE}" -ojsonpath="{.data.tls\.crt}" | base64 -d > "${CERT_FILE}"
+kubectl get secret allocator-client.default -n "${NAMESPACE}" -ojsonpath="{.data.tls\.key}" | base64 -d > "${KEY_FILE}"
+kubectl get secret allocator-tls-ca -n agones-system -ojsonpath="{.data.tls-ca\.crt}" | base64 -d > "${TLS_CA_FILE}"
+
+curl --key ${KEY_FILE} --cert ${CERT_FILE} --cacert ${TLS_CA_FILE} -H "Content-Type: application/json" --data '{"namespace":"'${NAMESPACE}'"}' https://${EXTERNAL_IP}/gameserverallocation -XPOST
+```
+
```
markmandel, line 164: Is it worth providing a sample output? It might be nice to know the format of the JSON?

site/content/en/docs/Advanced/multi-cluster-allocation.md
```diff
@@ -103,6 +103,17 @@ go run examples/allocator-client/main.go --ip ${EXTERNAL_IP} \
     --multicluster true
 ```
 
+If using REST use
+
+```bash
+go run examples/allocator-client/main.go --ip ${EXTERNAL_IP} \
```
markmandel, line 109: Should this be `curl` if using REST? (Also, we'll need to wrap it in the `feature` shortcode like above please.)

test/e2e/allocator_test.go
```diff
@@ -99,6 +103,68 @@ func TestAllocator(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestRestAllocator(t *testing.T) {
```
markmandel, line 106: Yay! E2E test!

markmandel (commented):
This is awesome - can't wait to have it in. :cake: Some comments and questions.

site/content/en/docs/Advanced/allocator-service.md
kdima: I have wrapped the updates docs in the `{{feature}}` I think it might be easier to review the changes this way? But happy to copy the whole doc to have the old and new version if you think that it is better that way?

site/content/en/docs/Advanced/allocator-service.md
kdima: Added

cmd/allocator/main.go
kdima: I have changed the code a bit after your comment as originally I incorrectly always called `ListenAndServeTLS` and now I call `ListenAndServe` when `h.tlsDisabled`.

Here I have changed the behaviour of `getServeroptions` in the past if `h.tlsDisabled` then KeepAlive and StatsHandler would not be added to the ServerOptions. I think it is ok to always have them whether TLS exists or not.
At this point `getServeroptions` is not aware of TLS as we manage it inside the `http.Server` `TLSConfig`

site/content/en/docs/Advanced/multi-cluster-allocation.md
kdima: Good catch. Fixed

cmd/allocator/main.go
roberthbailey: Is this backwards compatible? I'm wondering if we could take the existing grpc port on the sidecar and add a similar handler to handle both grpc and http.

If so, we could migrate the default ports to be the same for both and then deprecate the ability to select a separate http port. After a grace period we could remove the ability to select the http port.

kdima: My understanding is that it is backwards compatible in the sense that grpc clients will not be aware of the change when `grpcHandlerFunc` is added. I double checked it by running the example allocator-client.

site/content/en/docs/Advanced/allocator-service.md
markmandel: This looks good :+1:

markmandel (commented):
This LGTM! Happy to approve once @pooneh-m has a chance to review. :+1:

cmd/allocator/main.go
```diff
@@ -64,6 +63,18 @@ const (
 	sslPort = "8443"
 )
 
+// grpcHandlerFunc returns an http.Handler that delegates to grpcServer on incoming gRPC
+// connections or otherHandler otherwise. Copied from https://github.com/philips/grpc-gateway-example.
+func grpcHandlerFunc(grpcServer http.Handler, otherHandler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// This is a partial recreation of gRPC's internal checks https://github.com/grpc/grpc-go/pull/514/files#diff-95e9a25b738459a2d3030e1e6fa2a718R61
+		if r.ProtoMajor == 2 && strings.Contains(r.Header.Get("Content-Type"), "application/grpc") {
```
pooneh-m, line 71: I am still not clear why `ProtoMajor == 2` and not 3 for the gRPC server.
Can you please add more documentations?

pooneh-m (commented):
Thanks for the change. I left some minor comments, but looks pretty good.

cmd/allocator/main.go
kdima: `ProtoMajor` refers to the HTTP protocol major version part. We are checking if request is HTTP/1.1 or HTTP/2. Will add that as a comment to the code.

cmd/allocator/main.go
```diff
@@ -255,24 +289,10 @@ func readTLSCert() (*tls.Certificate, error) {
...
-	cfg := &tls.Config{
-		GetCertificate: h.getTLSCert,
-	}
-
-	if !h.mTLSDisabled {
-		cfg.ClientAuth = tls.RequireAnyClientCert
-		cfg.VerifyPeerCertificate = h.verifyClientCertificate
-	}
-
 	// Add options for creds and  OpenCensus stats handler to enable stats and tracing.
```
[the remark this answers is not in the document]
kdima: Not sure what you mean. Do you mean that I should remove/change the comment that says:
`// Add options for creds and  OpenCensus stats handler to enable stats and tracing.`
or that I need to document changes to options?
I don't think this change changes the command line options as `h.tlsDisabled` and `h.mTLSDisabled` are still used.

site/content/en/docs/Advanced/allocator-service.md
kdima: I have modified both allocator-service.md and multi-cluster-allocation.md to have a factored out variable setting. Let me know if you think it looks ok now?
I think as a user I don't mind a bit of code duplication in the example scripts as it allows me to copy paste the script from documentation and have it run straight away, whereas with factored out env setting I need to do 2 steps.

pooneh-m: eventually the code snippets will be turned into a script for running multiple times but having a clear documentation on the code snippets may make it easier for the first timer to understand the flow. @markmandel do you have any recommendation on this? I am not experienced in the user experience for documentation.

cmd/allocator/main.go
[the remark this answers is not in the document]
kdima: Makes sense! Fixed it now.

site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -139,6 +148,63 @@ go run examples/allocator-client/main.go --ip ${EXTERNAL_IP} \
...
+To start, take a look at the allocation gRPC client examples in {{< ghlink href="examples/allocator-client/main.go" >}}golang{{< /ghlink >}} and {{< ghlink href="examples/allocator-client-csharp/Program.cs" >}}C#{{< /ghlink >}} languages. In the following, the {{< ghlink href="examples/allocator-client/main.go" >}}golang gRPC client example{{< /ghlink >}} is used to allocate a Game Server in the `default` namespace.
+
+```bash
+#!/bin/bash
+
+# allocator-client.default secret is created only when using helm installation. Otherwise generate the client certificate and replace the following.
+# In case of MacOS replace "base64 -d" with "base64 -D"
+kubectl get secret allocator-client.default -n "${NAMESPACE}" -ojsonpath="{.data.tls\.crt}" | base64 -d > "${CERT_FILE}"
+kubectl get secret allocator-client.default -n "${NAMESPACE}" -ojsonpath="{.data.tls\.key}" | base64 -d > "${KEY_FILE}"
+kubectl get secret allocator-tls-ca -n agones-system -ojsonpath="{.data.tls-ca\.crt}" | base64 -d > "${TLS_CA_FILE}"
```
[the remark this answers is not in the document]
kdima: Moved the lines and renamed the section to be: `Reuse following snippet for both grpc and rest examples:`
Not sure whether this is good heading for it so open to suggestions.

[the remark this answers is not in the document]
pooneh-m: Maybe something like `Set the environment variables and store the client secrets before allocating using gRPC or REST APIs`?

Also, please make the same change for the `multi-cluster-allocation.md` file to keep them in sync.

pooneh-m (approved):

site/content/en/docs/Advanced/allocator-service.md
markmandel: Looks good to me - I get it :+1:

Issue #1824, opened 2020-09-17
Upgrade to Kubernetes 1.17


roberthbailey:
We aren't ready for this yet, but now that we are wrapping up the upgrade to 1.16 we have some steps we want to proactively capture for 1.17 so it's time to start writing them down. 

List of items to do for upgrading to 1.17 (this was started from the list in the [1.16 issue](https://github.com/googleforgames/agones/issues/1649) and likley need further updates):

- [x] Update e2e cluster to run against 1.17
  - [x] Recreate cluster with new scripts (after updating the GKE terraform module)
  - [x] Update kubectl in `e2e-image/Dockerfile`
- [x] Update prow cluster to use 1.17 (even though we aren't using it yet, we should keep it in sync)
  - [x] Recreate cluster with new scripts
- [x] Update the dev tooling to create 1.17 clusters
  - [x] GKE
  - [x] Minikube
  - [x] Kind
  - [x] Update kubectl
- [x] Update terraform submodules
  - [x] GKE
  - [x] Azure
  - [x] EKS
- [x] Update documentation for creating clusters to 1.17
  - [x] Usage requirements
  - [x] GKE
  - [x] Minikube
  - [x] EKS
  - [x] AKS
- [x] Update links to k8s documentation
  - [x] examples/fleet.yaml
  - [x] examples/fleetautoscaler.yaml
  - [x] examples/gameserver.yaml
  - [x] site/content/en/docs/Reference/fleet.md
  - [x] site/content/en/docs/Reference/fleetautoscaler.md
  - [x] site/content/en/docs/Reference/gameserver.md
  - [x] site/content/en/docs/Advanced/limiting-resources.md
  - [x] site/assets/templates/crd-doc-config.json
- [x] Update client-go
- [x] Regenerate CRD Kubernetes client libraries
- [x] Move API references from beta to v1 (see the [Kubernetes v1.16 Release Notes](https://v1-16.docs.kubernetes.io/docs/setup/release/notes/#no-really-you-must-read-this-before-you-upgrade))
  - [x] CRDs
  - [x] Admission Webhooks

markmandel:
We should target this at our 1.11.0 release - 1.17 is available across all three cloud providers.

markmandel:
Wondering if we want to add running a long running version of #1867 to ensure there are no memory leaks in client-go when we upgrade?

This is the second time (#1871) we've been bitten by a memory leak in client-go.

roberthbailey:
[...]

PR #1906, opened 2020-11-20
Update minikube dev tooling

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Needed to update minikube to Kubernetes 1.17.x and I figured I would also go through the minikube dev experience and update it.

This includes:

* Switch to default to the Docker driver, since everyone should have Docker installed.
* Removing the Windows hacks, because they were awful and I feel bad I even wrote them in the first place.
* Migrate tooling to use new minikube functionality
* Update minikube commands to up to date release conformity.
* Updated the documentation

Work on #1824

N/A

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -102,8 +102,6 @@ Table of Contents
 ### Linux
 - Install Make, either via `apt install make` or `yum install make` depending on platform.
 - [Install Docker](https://docs.docker.com/engine/installation/) for your Linux platform.
-- (optional) Minikube will require [VirtualBox](https://www.virtualbox.org) and will need to be installed if you wish
-  to develop on Minikube 
 
 ### Windows
 Building and developing Agones requires you to use the 
@@ -121,16 +119,12 @@ as this makes it easy to create a (relatively) cross platform development and bu
 - Agones will need to be cloned somewhere on your `/c` (or drive of your choice) path, as that is what Docker will support mounts from
 - All interaction with Agones must be on the `/c` (or drive of your choice) path, otherwise Docker mounts will not work
 - Now the `make` commands can all be run from within your WSL shell
-- (optional) Minikube is supported via the [HyperV](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/index) 
-  driver - the same virtualisation platform as the Docker installation.
-- **Note**: If you want to dev and test with Minikube, you **must** run WSL as Administrator, otherwise Minikube can't control HyperV.
+- The Minikube setup on Windows has not been tested. Pull Requests would be appreciated!
 
 ### macOS
 
 - Install Make, `brew install make`, if it's not installed already
 - Install [Docker for Mac](https://docs.docker.com/docker-for-mac/install/)
-- (optional) Minikube will require [VirtualBox](https://www.virtualbox.org) and will need to be installed if you wish
-  to develop on Minikube 
 
 ## GOPATH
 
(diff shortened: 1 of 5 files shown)


build/README.md
```diff
@@ -653,20 +638,22 @@ Since Minikube runs locally, there are some targets that need to be used instead
 
 #### `make minikube-test-cluster`
 Switches to an "agones" profile, and starts a kubernetes cluster
```
aLekSer, line 640: I see that additional command which change profile was removed. Was it redundant?

build/includes/windows.mk
```diff
@@ -31,20 +31,6 @@ sha_dir = $(shell sha256sum  $(1) | cut -d' ' -f1 | sha256sum | head -c 10 )
...
 # Minikube executable
 MINIKUBE ?= minikube.exe
-# Default minikube driver
-MINIKUBE_DRIVER ?= hyperv
-# set docker env for minikube
-MINIKUBE_DOCKER_ENV ?= eval $$($(MINIKUBE) docker-env --shell=bash) && \
- export DOCKER_CERT_PATH=$$(echo $$DOCKER_CERT_PATH | $(win_to_wsl_path))
-
-# minikube shell mount for certificates
-minikube_cert_mount = $(cert_path):$(cert_path)
```
aLekSer, line 41: I think this one should stay there

build/includes/linux.mk
```diff
@@ -31,13 +31,6 @@ sha_dir = $(shell sha256sum  $(1) | cut -d' ' -f1 | sha256sum | head -c 10 )
 
 # Minikube executable
 MINIKUBE ?= minikube
-# Default minikube driver
-MINIKUBE_DRIVER ?= virtualbox
-# set docker env for minikube
-MINIKUBE_DOCKER_ENV ?= eval $$($(MINIKUBE) docker-env)
-
-# minikube shell mount for certificates
-minikube_cert_mount := ~/.minikube:$(HOME)/.minikube
```
aLekSer, line 40: This one also might be needed and be different for different OSs. WDYT?

build/includes/minikube.mk
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+MINIKUBE_DRIVER ?= docker
+
+# minikube shell mount for certificates
+minikube_cert_mount := ~/.minikube:$(HOME)/.minikube
```
aLekSer, line 18: It seems that Windows would not have this folder. Am I missing something? I can test on my other PC.

aLekSer (commented):
Left small comments regarding `minikube_cert_mount ` which might need to be OS-specific.

build/README.md
markmandel: This is now powered by the `-p` argument in `minikube start` -  which also creates the profile. `minikube profile foo` will not create the profile if it doesn't exist.

https://minikube.sigs.k8s.io/docs/commands/start/

build/includes/windows.mk
markmandel: So the previous Windows support was super hacky, brittle and I expect it doesn't work anymore.

We were rewriting windows path to linux paths in make files. If we decide to continue supporting it, it needs a whole new rewrite. So I've removed it for now. If we need to add new os specific tooling, then lets revisit then. Right now, between OSX and Linux, it's all the same.

build/includes/linux.mk
markmandel: For OSX and Linux it's the same.

It _may_ be the same for minikube under WSL, but it requires testing. But I figure we can revisit if we decide it's possible to support windows without hackery.

build/includes/minikube.mk
markmandel: This is why I changed the README to:
https://github.com/googleforgames/agones/pull/1906/files#diff-4b71ec736aee936feb417e5193e800a1014a8e2df2a5b5b92fd78fbd3d781dbdR122

> The Minikube setup on Windows has not been tested. Pull Requests would be appreciated!

I expect it will all fall apart. Last time I looked at this (many years ago now), kubectl on WSL wouldn't work with minikube because the .kubectl referenced certs hosted at `C:\` which doesn't work for Linux.

I expect the entire Windows dev environment setup would need a review anyway. (do we even have anyone developing on windows?)

roberthbailey (approved):
LGTM. Please address @aLekSer's comments before merging.

build/includes/minikube.mk
aLekSer: Got an idea. That would be a leftover from this PR.

markmandel: Since we have the git history to go back and look at if , I'm not convinced it's necessary to leave this in, given that we don't actually know what the correct path is for windows+minikube dev (or if we even want to support it).

This feels to me like leaving in commented out code in the current source repo.

markmandel:
@roberthbailey do you want to tiebreak on this PR?

build/includes/minikube.mk
roberthbailey: +1 to not leaving stuff in that is "dead code" in the hope that we will fix it later. It's better to clean it up and we can always look at repo history if necessary to pull it back in.

markmandel: In which case, I will merge this, and we can revisit Windows at a later date :+1:

Issue #1799, opened 2020-09-08
Update apiextensions.k8s.io/v1beta1 to v1 for all CRDs


aLekSer:
This should be done as part of Upgrade to Kubernetes 1.17.

**What should be done**

We have non structural schema errors on some CRDs:
```
kubectl describe crd gameserverallocationpolicies.multicluster.agones.dev 
...
 Message:               spec.validation.openAPIV3Schema.type: Required value: must not be empty at the root
    Reason:                Violations
    Status:                True
    Type:                  NonStructuralSchema
  Stored Versions:
    v1
Events:  <none>
```
So we need to add `type: object` root element in validation schema field.

`spec.validation`  need to be moved to a new `spec.versions.schema` section.
A working CRD example could be found here: 
https://github.com/googleforgames/agones/issues/1194#issuecomment-686560198

Next Go import should be updated `"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"`
More details could be found in this PR https://github.com/kubernetes/kubernetes/pull/79604 .

Robert's comment: CRDs went to v1 in 1.16, so I think we should wait until we move to 1.17 before we switch to using the v1 api type.

markmandel:
Another thing to do in here - CRD v1 has defaulting support via the schema. Before closing this issue, we should look at if we can move defaulting to schemas, and possibly do away with some/part of our Mutating Webhooks.

markmandel:
Have moving of pretty much everything I can find from `v1beta1` :arrow_right: `v1` in a WIP branch.

Will push more up when #1909 is merged.

PR #1909, opened 2020-11-24
Move CRDs from v1beta1 to v1

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

This does the work to move all the CRDs from v1beta1 to v1.

This includes:

- Inclusion of a typed schema to all CRDs
- Disabled pruning in appropriate places
- Cleanup on Allocation policy CRD (mostly reordering elements)
- Minor tweaks to e2e tests as necessary.

Work on #1799

### DIFF
--- a/install/helm/agones/templates/crds/_gameserverspecschema.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecschema.yaml
@@ -12,22 +12,37 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{/* Validation for a gameserver spec */}}
-{{- define "gameserver.validation" }}
+{{/* schema for a gameserver spec */}}
+{{- define "gameserver.schema" }}
+type: object
 required:
 - spec
 properties:
+  {{- if .metadata | default false }}
+  metadata:
+    type: object
+    properties:
+      labels:
+        type: object
+        x-kubernetes-preserve-unknown-fields: true
+      annotations:
+        type: object
+        x-kubernetes-preserve-unknown-fields: true
+  {{- end}}
   spec:
+    type: object
     required:
     - template
     properties:
       template:
         type: object
+        x-kubernetes-preserve-unknown-fields: true
         required:
         - spec
         properties:
           spec:
             type: object
+            x-kubernetes-preserve-unknown-fields: true
             required:
             - containers
             properties:
@@ -36,6 +51,7 @@ properties:
                 minItems: 1
                 items:
                   type: object
+                  x-kubernetes-preserve-unknown-fields: true
                   required:
                   - image
                   properties:
@@ -61,6 +77,9 @@ properties:
         items:
           type: object
           properties:
+            name:
+              title: Name is the descriptive name of the port
+              type: string
             portPolicy:
               title: the port policy that will be applied to the game server
               description: |
@@ -82,6 +101,11 @@ properties:
               - UDP
               - TCP
               - TCPUDP
+            container:
+              title: |
+                Container is the name of the container on which to open the port. Defaults to the game server container.
+                This field is beta-level and is enabled by default, could be disabled by the "ContainerPortAllocation" feature.
+              type: string
             containerPort:
               title: The port that is being opened on the game server process
               type: integer
@@ -149,16 +173,13 @@ properties:
             type: integer
             minimum: 1
             maximum: 2147483648
-      alpha:
+      players:
         type: object
-        title: Alpha properties for the GameServer
+        title: Configuration of player capacity
+        nullable: true
         properties:
-          players:
-            type: object
-            title: Configuration of player capacity
-            properties:
-              initialCapacity:
-                type: integer
-                title: The initial player capacity of this Game Server
-                minimum: 0
+          initialCapacity:
+            type: integer
+            title: The initial player capacity of this Game Server
+            minimum: 0
 {{- end }}
--- a/install/helm/agones/templates/crds/_gameserverstatus.yaml
+++ b/install/helm/agones/templates/crds/_gameserverstatus.yaml
@@ -1,0 +1,66 @@
+# Copyright 2020 Google LLC All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{/* schema for a gameserver status */}}
+{{- define "gameserver.status" }}
+status:
+  type: object
+  title: The status values for the GameServer
+  properties:
+    state:
+      type: string
+      enum:
+        - PortAllocation
+        - Creating
+        - Starting
+        - Scheduled
+        - RequestReady
+        - Ready
+        - Shutdown
+        - Error
+        - Unhealthy
+        - Reserved
+        - Allocated
+    ports:
+      type: array
+      nullable: true
+      items:
+        type: object
+        properties:
+          name:
+            type: string
+          port:
+            type: integer
+    address:
+      type: string
+    nodeName:
+      type: string
+    reservedUntil:
+      type: string
+      nullable: true
+      format: date-time
+    players:
+      type: object
+      nullable: true
+      properties:
+        count:
+          type: integer
+        capacity:
+          type: integer
+        ids:
+          type: array
+          nullable: true
+          items:
+            type: string
+{{- end}}
--- a/install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
+++ b/install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
@@ -14,12 +14,10 @@
 
 {{- if .Values.agones.crds.install }}
 
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  creationTimestamp: null
   labels:
-    controller-tools.k8s.io: "1.0"
     component: crd
     app: {{ template "agones.name" . }}
     chart: {{ template "agones.chart" . }}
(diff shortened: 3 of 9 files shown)


install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
```diff
@@ -14,12 +14,10 @@
 
 {{- if .Values.agones.crds.install }}
 
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
markmandel, line 17: @pooneh-m would love for you to give an extra set of eyes to this, just to triple check things :smile: 

I reordered the schema to match the Go structure to make things easier to debug, and also removed some of the extraneous pieces in there.

I expect it all to be fine, and the e2e tests all pass, but figured it was worth an extra check :+1:

markmandel:
As an extra FYI, after this PR, I'll submit the code changes to replace the underlying go library calls from v1beta1 to v1.

I just figured this was enough to do in one go.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -12,22 +12,37 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{/* Validation for a gameserver spec */}}
-{{- define "gameserver.validation" }}
+{{/* schema for a gameserver spec */}}
```
roberthbailey, line 15: Question about this file: it looks like it was moved (not copied). Is that going to break compatibility at all? It seems like we should add support for v1 CRDs first, wait a release or two, and then drop support for v1beta1 CRDs (I'm assuming that is possible). 

If not, then this change may make it impossible to do in-place upgrades to the next agones release.

install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
roberthbailey: We don't have any skew tests, but it would be worth doing a quick smoke check installing an old version of agones, running a fleet, and then doing a helm upgrade to see what happens (I think all of our e2e tests assume fresh install).

install/helm/agones/templates/crds/_gameserverspecschema.yaml
markmandel: > Question about this file: it looks like it was moved (not copied). Is that going to break compatibility at all?

Should make absolutely no difference at all. I just renamed it to be semantically inline with what the details are called in CRD v1.

> It seems like we should add support for v1 CRDs first, wait a release or two, and then drop support for v1beta1 CRDs (I'm assuming that is possible).

That outside of our control. Also the beta definitions of CRDs in Agones are already being converted into v1 CRDs by Kubernetes (in fact, that's how I did some of the initial conversions - looked at the details of how Kubernetes was doing it).

> If not, then this change may make it impossible to do in-place upgrades to the next agones release.

Out e2e cluster seems to be happily doing upgrades from v1 ↔ v1beta1 ↔ v1 again without any concerns, so it looks like in-place upgrades are not an issue.

install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
markmandel: Since our e2e tests do `helm update` - it _should_ be fine, but I can do a quick smoke check, it's a good idea to be sure.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
markmandel: In fact - I just realised something fun to note.

Kubernetes does the conversion in both directions. I.e. the code in Agones still references the CRDs via the beta interface, even though we are installing them as v1.

It's quite a clever little system they have going on in there :+1:

install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
markmandel: Just ran the test - installed Agones 1.10.0 via helm, and then did a `helm update` over the top with this PR's version, and all installed without issue.

markmandel: By default our helm install will remove all previous GameServers on update -- it's quite possible that if you still have a previous GameServer installed, it may not be possible to update via helm without deleting GameServers first. Are we concerned with this scenario? (It's not one we advocate people do in our [update documentation](https://agones.dev/site/docs/installation/upgrading/#upgrading-agones-single-cluster) )

roberthbailey: I'm not concerned about that scenario, since we tell folks it isn't a supported upgrade path.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
roberthbailey: Yes, that's part of the reason it took CRDs so long to reach v1.... lots of details to get right. :)

install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
markmandel: > I'm not concerned about that scenario, since we tell folks it isn't a supported upgrade path.

Agreed - just wanted to check. :+1:

markmandel: Just checking, is there anything blocking then for this PR, or are we waiting for @pooneh-m to do an extra check?

theminecoder:
Personally I would love to see at some point these crds expose the full api of sub objects rather then just hiding it behind basic openapi objects.

Would make it much handier if I could get the validation of the GameServer part of the spec while writing the Fleet object

markmandel:
> Personally I would love to see at some point these crds expose the full api of sub objects rather then just hiding it behind basic openapi objects.
> 
> Would make it much handier if I could get the validation of the GameServer part of the spec while writing the Fleet object

Sorry, I'm not following what you mean @theminecoder - could you expand? Or maybe provide an example?

theminecoder:
> > Personally I would love to see at some point these crds expose the full api of sub objects rather then just hiding it behind basic openapi objects.
> 
> > 
> 
> > Would make it much handier if I could get the validation of the GameServer part of the spec while writing the Fleet object
> 
> 
> 
> Sorry, I'm not following what you mean @theminecoder - could you expand? Or maybe provide an example?

When you write a Deployment object/resource/manifest you get the full validation of the deployment and the pod template that sits inside it, which makes it possible for ides and such to suggest and check fields and values which make writing the objects a ton easier. 

Currently your crds turn into basic objects when you get into template sub objects so all that validation and hinting is lost and writing the objects requires people to either know the fields by heart or open the website for documentation every time.

As a side note and what led me to realise this, I have been trying to use the crds inside a controller of my own (java based) but since the fields just turn into a basic objects the class generation just breaks as soon as it hits those sub objects unless I manually intervene to fix the typings.

markmandel:
> When you write a Deployment object/resource/manifest

Write where? In an IDE? Via code?

> You get the full validation of the deployment and the pod template that sits inside it

Where do you get this validation? The K8s API validates, so I'm not sure you mean when you make requests against the API?

> Currently your crds turn into basic objects when you get into template sub objects so all that validation and hinting is lost and writing the objects requires people to either know the fields by heart or open the website for documentation every time.

Not sure if it answers your question, but you could pull the full CRD out of K8s once it's been installed. `kubectl get <crd-name> -o yaml` ? Does that help?

CRDs also don't have the ability (AFAIK) to make references to other CRD types, unless there is new functionality that I'm missing?

> As a side note and what led me to realise this, I have been trying to use the crds inside a controller of my own (java based) but since the fields just turn into a basic objects the class generation just breaks as soon as it hits those sub objects unless I manually intervene to fix the typings.

If I'm understanding correctly, this is a code generation issue? Or an IDE issue? (What IDE are you using? How does it know about CRD validation?)

If you could provide examples of what you would like the above CRD definitions to look like (would need to be backward compatible as well, which is always the fun task), that might make things a bit clearer as well?

install/helm/agones/templates/crds/gameserverallocationpolicy.yaml
pooneh-m: Thanks @markmandel for the change. LGTM.

markmandel:
Gentle bump on re-review @roberthbailey @pooneh-m please.

test/e2e/fleet_test.go
```diff
@@ -300,13 +302,16 @@ func TestFleetRollingUpdate(t *testing.T) {
 						if maxUnavailable == 0 {
 							maxUnavailable = 1
 						}
-						// This difference is inevitable, also could be seen with Deployments and ReplicaSeets
+						// This difference is inevitable, also could be seen with Deployments and ReplicaSets
 						shift = maxUnavailable
+					} else {
+						shift = 1 // not sure why this solves the issue???
```
roberthbailey, line 308: What issue does this solve?

test/e2e/fleet_test.go
```diff
@@ -300,13 +302,16 @@ func TestFleetRollingUpdate(t *testing.T) {
...
 					}
 					assert.Nil(t, err)
 
-					if len(list.Items) > targetScale+maxSurge+maxUnavailable+shift {
-						err = fmt.Errorf("New replicas should be less than target + maxSurge + maxUnavailable %d %d", len(list.Items), targetScale+maxSurge+maxUnavailable+shift)
+					expectedTotal := targetScale + maxSurge + maxUnavailable + shift
+					if len(list.Items) > expectedTotal {
+						err = fmt.Errorf("new replicas should be less than target + maxSurge + maxUnavailable. Replicas: %d, Expected: %d", len(list.Items), expectedTotal)
```
roberthbailey, line 314: why does the printed string not include `+ shift` while the actual check does? (It was that way before too, but it seems strange)

install/helm/agones/templates/crds/_gameserverstatus.yaml
```diff
@@ -0,0 +1,66 @@
+# Copyright 2018 Google LLC All Rights Reserved.
```
roberthbailey, line 1: As this is a new file this should be 2020

markmandel: :thinking: this was technically a file rename, not a file create - should the licence change?

test/e2e/fleet_test.go
markmandel: Hmnn. That's odd, I thought I removed this. I was running the wrong feature gate'd test against the wrong install. Nice catch, I should be able to remove it.

install/helm/agones/templates/crds/_gameserverstatus.yaml
markmandel: Oh wait, I'm totally wrong. Ignore me. I'll make this change.

test/e2e/fleet_test.go
markmandel: :shrug: I'll add it it. I was leaving it as it was. But it makes sense to change it.

markmandel:
Thanks for that, this should be good to go now :crossed_fingers:

pooneh-m (approved):

roberthbailey (approved):

Issue #1899, opened 2020-11-16
Migrate to using SANs for webhook certificates for Go 1.15


markmandel:
Context: https://github.com/googleforgames/agones/pull/1897

Go 1.15 no longer supports Common Name based certs for HTTPS without a specific flag. Common Name is a 20 year old deprecation, and we probably shouldn't be promoting.

There is a backward compatible environment flag, but it will be going away eventually. This flag, for the life of me I could not make it work -- which blocked a migration to latest 1.15.

I tried to write steps to create certs with SANs, but my OpenSSL knowledge is not good enough.

So, as a plan to migrate to Go 1.15:

1. [x] Update [webhook with cert creation](https://agones.dev/site/docs/getting-started/create-webhook-fleetautoscaler/#2-create-x509-root-and-webhook-certificates) instructions to use SANs over Common Name, and ensure it works with current Agones version.
2. [x] Provide release note announcement on deprecation of Common Name certificates for webhooks.
3. [ ] Migrate build system to Go 1.15 and no longer support SAN based certificates.

markmandel:
We could really move to 1.16 at this point! 
https://golang.org/doc/go1.16

markmandel:
Worth noting - if you upgrade to 1.16, it's possible the certs we use for validation and mutation webhooks when installing thorugh `index.yaml` may not container SAN's and may therefore break the install path for `install.yaml` 

`helm` should be fine, since it generates SAN powered certs.

PR #1910, opened 2020-11-25
Adding SAN to fleet autoscaler certs and updating documentation

kdima:
**What type of PR is this?**
/kind cleanup

This PR solves part of #1899. I have updated the certs used for webhook integration test to include SAN record and also updated the docs with the description on how to generate such certs.

It partially completes #1899 

I have tested this change by running the `TestFleetAutoscalerTLSWebhook` integration test on go 1.15.5.


test/e2e/fleetautoscaler_test.go
```diff
@@ -461,76 +461,76 @@ func TestAutoscalerWebhook(t *testing.T) {
...
-Os8osgmh2DlutyIk7piEx9K6+yeW0h5T1SAsYVvHAoGAXlhRhU7ji0tolYrNruAh
-7cwK9Qe6t/p6LlqaprZsTOJMEx/oJUj+nKsTArrGdfp1X83YZCBTfWGmhs5ZBw+E
-jqnH6j9fcRCk7MySKZMBTdrQlUqfXFo3dm7Hp9Vu2Tb3FspFn6jcjsGyCwcUoT+e
-W9iNePwxwHpvbQ15iu9e0mUCgYAEsBbXT8x35LsMm6G1CQn+W4zsGjaswBzwuI06
-47sThpQfJsni7OTGt42WvcLIFhfM5393tIftSKHZETCb2a7/M3FuC70oOVJJKpui
-HBOBOWDT+rpjRZ9LE9v9/J/wcDzP4okhftRWyqn/8eJD5MyrM+6WnYHu0Vq8Hr3/
-h2ccwwKBgQCha4ox+SaXzlYROr1qge8xgqK+lpg1i2f32PgK3Ipodjk3esAuzeNM
-L5o5pDLorHu4EFtveIneRsV+kf8YPVuid18lYzMAJBqlXvcUik2Izk57cWB/P9so
-3/03jXI8iT8BbIU+PoII2EvQEPeAI07BYMU9cvsiFvFoB6z162DJhw==
+MIIEpQIBAAKCAQEAzTtFY02SAY4jHiryJbBRT4+2wn1OlqL4WTWUFtKaWEjm+gAn
```
markmandel, line 464: Oh that's awesome - so we now also know that using a SAN works with the current Agones install and Go toolchain.

Fantastic!

site/content/en/docs/Getting Started/create-webhook-fleetautoscaler.md
```diff
@@ -307,14 +307,30 @@ Every webhook that you wish to install a trusted certificate will need to go thr
...
+CN = autoscaler-tls-service.default.svc
+[v3_req]
+keyUsage = digitalSignature
+extendedKeyUsage = serverAuth
+subjectAltName = @alt_names
+[alt_names]
+DNS.1 = autoscaler-tls-service.default.svc
+```
+
+Generate the certificate signing request, use valid hostname which is `autoscaler-tls-service.default.svc` as `Common Name (eg, fully qualified host name)` in the config file.
```
markmandel, line 326:
```suggestion
Generate the certificate signing request, use valid hostname which in this case will be `autoscaler-tls-service.default.svc` as `Common Name (eg, fully qualified host name)` as well as `DNS.1` in the `alt_names` section of the config file.

Check the [Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records) to see how Services get assigned DNS entries.
```

Just wanted to expand that a little, and I've always wanted to link back to the documentation so people can see how DNS entries work. What do you think?

markmandel (commented):
Thank you for doing this!!!!

Minor doc note, but otherwise this is awesome.

site/content/en/docs/Getting Started/create-webhook-fleetautoscaler.md
```diff
@@ -307,14 +307,30 @@ Every webhook that you wish to install a trusted certificate will need to go thr
...
+[req]
+distinguished_name = req_distinguished_name
+req_extensions = v3_req
+prompt = no
+[req_distinguished_name]
+CN = autoscaler-tls-service.default.svc
+[v3_req]
+keyUsage = digitalSignature
+extendedKeyUsage = serverAuth
+subjectAltName = @alt_names
```
pooneh-m, line 321: You can also use the existing configurations in openssl.cnf and only set SAN. Here is an example of doing it in the old version of Agones documentations: https://1-4-0.agones.dev/site/docs/advanced/allocator-service/#server-tls-certificate

markmandel: oooh!

I have no strong opinions either way. Whichever is easier to explain?

Do we feel like having a file better, because it's easier for automation tools to utilise?

site/content/en/docs/Getting Started/create-webhook-fleetautoscaler.md
kdima: I think that is a good idea. Cahnged the doc.

site/content/en/docs/Getting Started/create-webhook-fleetautoscaler.md
kdima: I am happy to change it to use the `/etc/ssl/openssl.cnf`, but my thinking is that if you use that we are appending extra config to a fairly chunk example file. Whereas if we use a custom `cert.conf` it is easier for people to follow what is actually happening.

markmandel: > I am happy to change it to use the /etc/ssl/openssl.cnf, but my thinking is that if you use that we are appending extra config to a fairly chunk example file. Whereas if we use a custom cert.conf it is easier for people to follow what is actually happening.

Yeah, I would agree. Also wouldn't want people editing/thinking they have to edit the `/etc/ssl/openssl.cnf` file either.

kdima: Shall I keep it as is then?

kdima:
Am I OK to merge this or is there somenthing I should fix up before hand?

pooneh-m (approved):

markmandel:
Making this PR as breaking, not because it actually breaks things, but to remind us to make a note in the release notes to get people to update their webhook certs.

PR #1915, opened 2020-12-07
Update Kind dev tooling to 1.17.3

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Updates to the next K8s verison.

Work on #1824

### DIFF
--- a/build/includes/kind.mk
+++ b/build/includes/kind.mk
@@ -26,7 +26,7 @@ kind-test-cluster: DOCKER_RUN_ARGS+=--network=host
 kind-test-cluster: $(ensure-build-image)
 	@if [ -z $$(kind get clusters | grep $(KIND_PROFILE)) ]; then\
 		echo "Could not find $(KIND_PROFILE) cluster. Creating...";\
-		kind create cluster --name $(KIND_PROFILE) --image kindest/node:v1.16.9 --wait 5m;\
+		kind create cluster --name $(KIND_PROFILE) --image kindest/node:v1.17.11 --wait 5m;\
 	fi
 
 # deletes the kind agones cluster


build/includes/kind.mk
```diff
@@ -26,7 +26,7 @@ kind-test-cluster: DOCKER_RUN_ARGS+=--network=host
 kind-test-cluster: $(ensure-build-image)
 	@if [ -z $$(kind get clusters | grep $(KIND_PROFILE)) ]; then\
 		echo "Could not find $(KIND_PROFILE) cluster. Creating...";\
-		kind create cluster --name $(KIND_PROFILE) --image kindest/node:v1.16.9 --wait 5m;\
+		kind create cluster --name $(KIND_PROFILE) --image kindest/node:v1.17.13 --wait 5m;\
```
roberthbailey, line 29: Did you verify that this works? On https://hub.docker.com/r/kindest/node/tags?page=1&ordering=last_updated I see 1.17.11 but not 1.17.13.

markmandel: I haven't, I don't have Kind installed. Honestly, I've no idea if the dev tooling works. Maybe we should remove it entirely.

roberthbailey: I'm ok with removing it. If anyone is using it it would be straightforward to add back in (and then we would know who to get help from to keep it maintained).

markmandel: Well, I've installed Kind (that was actually really easy) - let's see if this thing works!

markmandel: Switch to the 1.17.13 and ran through the tooling - all works!

roberthbailey (approved):

PR #1946, opened 2021-01-12
Add nodeSelector property to Agones helm chart for Allocator

josephbmanley:
**What type of PR is this?**
I would think
/kind feature

When trying to deploy to a cluster with both Linux and Windows containers, Agones needs to be configurable to just select the Linux nodes. The easiest way to do so is with a nodeSelector.

### DIFF
--- a/site/content/en/docs/Installation/Install Agones/helm.md
+++ b/site/content/en/docs/Installation/Install Agones/helm.md
@@ -168,7 +168,6 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.allocator.generateTLS`                      | Set to true to generate TLS certificates or false to provide certificates in `certs/allocator/*`| `true`                 |
 | `agones.allocator.disableMTLS`                      | Turns off client cert authentication for incoming connections to the allocator.            | `false`                |
 | `agones.allocator.disableTLS`                       | Turns off TLS security for incoming connections to the allocator. | `false`                |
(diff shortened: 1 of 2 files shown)


site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -168,6 +168,7 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.allocator.generateTLS`                      | Set to true to generate TLS certificates or false to provide certificates in `certs/allocator/*`| `true`                 |
 | `agones.allocator.disableMTLS`                      | Turns off client cert authentication for incoming connections to the allocator.            | `false`                |
 | `agones.allocator.disableTLS`                       | Turns off TLS security for incoming connections to the allocator. | `false`                |
+| `agones.allocator.nodeSelector`                     | Allocator [node labels][nodeSelector] for pod assignment                                        | `{}`                   |
```
markmandel, line 171: Looks good - but can we drop this in the section below with the header "New Configuration Features".

This means it won't show up in current documentation, and will only show up in development, and the next release :+1: 

#1947 should also unblock CI as well. Thanks!

josephbmanley: Updated in latest commit.

markmandel (approved):
:shipit:

PR #1953, opened 2021-01-15
Add a "why" section for the Allocator Service documentation

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

We get questions about which tool to use when allocating a GameServer - so here are some docs to help answer that question!

N/A

N/A


markmandel:
@pooneh-m would particularly like your input here, see if I missed anything.

pooneh-m (approved):
Thanks for the addition! It is useful.

PR #1956, opened 2021-01-20
CRD OpenAPI Spec for ObjectMeta & PodTemplateSpec

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This PR provides the mechanisms to pull core Kubernetes resource OpenAPI schemas out of the API, and customise and convert them into a format that is embeddable in our own CRD specifications.

This has been used to extract OpenAPI schemas for `ObjectMeta`, and `PodTemplateSpec`, and then embedded in `GameServer` as well as parent CRDs, creating a layer of validation for Pods definitions that are implemented via a GameServer.

Also included in a new helm config option of `gameservers.podPreserveUnknownFields` in case a user needs to escape
the field pruning on a Pod, or if PR causes a bug of any kind.

Work on #1298

- The code generation portion is split into it's own commit, so it's easier to review just the logic changes
- We should make special note of this in the release notes, so that people know it exists, and the workaround if it causes issues.
- This is not comprehensive validation, and therefore I have not closed the the referenced issue.


markmandel:
Well this is a weird one, helm hangs on doing the upgrade.

This is what I see when I enable debugging: - it just stops at the blank line, and never continues. The helm release is still marked as active, and successfully deployed.

```
       helm upgrade --install --wait --namespace=agones-system \
        --create-namespace \
        --debug \
        --set agones.image.tag=1.12.0-40bb824,agones.image.registry=gcr.io/agones-mark \
        --set agones.image.controller.pullPolicy="Always",agones.image.sdk.alwaysPull=true \
        --set agones.image.controller.pullSecret= \
        --set agones.allocator.http.serviceType="LoadBalancer" \
        --set agones.controller.logLevel="debug" \
        --set agones.featureGates="PlayerTracking=true&SDKWatchSendOnExecute=true&RollingUpdateOnReady=true" \
        --set agones.allocator.http.loadBalancerIP=34.82.197.230 \
         \
history.go:52: [debug] getting history for release agones
upgrade.go:121: [debug] preparing upgrade for agones
upgrade.go:129: [debug] performing update for agones
upgrade.go:308: [debug] creating upgraded release for agones
client.go:173: [debug] checking 32 resources for changes
client.go:440: [debug] Looks like there are no changes for ServiceAccount "agones-allocator"
client.go:440: [debug] Looks like there are no changes for ServiceAccount "agones-controller"
client.go:440: [debug] Looks like there are no changes for ServiceAccount "agones-sdk"

^Cmake: *** [Makefile:319: install] Error 2
```

markmandel:
Repro'd locally, but looks like an upgrade to Helm solved it. :crossed_fingers:

roberthbailey (commented):
My only concern here is that by expanding object meta (in particular) we add documentation for a _ton_ of new fields which users really shouldn't care about. The extra validation may be useful, but I don't expect anyone using Agones will need anything other than labels and annotations, and now there are a bunch of other documented and validated fields each place we have metadata that only serve to distract from the fields that are relevant.

roberthbailey:
https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/ presents the fields nicely. Maybe folks will just read that and the validation is primarily for machine consumption.

build/boilerplate.yaml.txt
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Google LLC All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http:#www.apache.org/licenses/LICENSE-2.0
```
aLekSer, line 7:
```suggestion
#     http://www.apache.org/licenses/LICENSE-2.0
```

markmandel: Oooh! Nice catch. Will fix!

test/e2e/gameserver_test.go
```diff
@@ -778,6 +783,65 @@ func TestGameServerResourceValidation(t *testing.T) {
...
 	assert.Equal(t, "container", statusErr.Status().Details.Causes[0].Field)
+
+	// test that values are still being set correctly
+	m50 := resource.MustParse("50m")
+	gs.Spec.Template.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory] = mi64
+	gs.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceMemory] = mi64
+	gs.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = m50
+	gs.Spec.Template.Spec.Containers[0].Resources.Limits[corev1.ResourceCPU] = m50
+
+	//gate
```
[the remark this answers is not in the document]
markmandel: Not sure where I picked up that habit/nomenclature from :thinking: 

Basically it's saying that it's not a check that is pertinent for the functionality, but instead a check to confirm that the data is in the correct format to being with. I'll change the comment to be more clear.

markmandel:
> My only concern here is that by expanding object meta (in particular) we add documentation for a _ton_ of new fields which users really shouldn't care about. The extra validation may be useful, but I don't expect anyone using Agones will need anything other than labels and annotations, and now there are a bunch of other documented and validated fields each place we have metadata that only serve to distract from the fields that are relevant.

I would posit that we don't need to change any documentation above what we have now. Prior to 1.11 we did support the full ObjectMeta, so 1.11 is technically a regression in functionality (who knows who might want to add a finaliser to their Pod/GameServer one day????), and we have links to the existing Kubernetes reference documentation for ObjectMeta.

We currently document ObjectMeta in the [GameServer reference](https://agones.dev/site/docs/reference/gameserver/) by pointing to https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#podtemplate-v1-core and letting people go from there.

Similarly https://agones.dev/site/docs/reference/agones_crd_api_reference/#agones.dev/v1.GameServer points to the K8s reference as well, similarly in https://agones.dev/site/docs/reference/agones_crd_api_reference/#agones.dev/v1.GameServerTemplateSpec

So I think our current docs have good coverage in this area.

> https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/ presents the fields nicely. Maybe folks will just read that and the validation is primarily for machine consumption.

That's a really nice page :+1: maybe that's something we can add in our docs instead of / in addition to the reference docs.

markmandel:
Note to self when writing the release notes:
- Note that there is a big CRD validation change, and that everything should be backward compatible, but in case it isn't, there is a config option to turn off pruning and validation.
- Note that with this big a CRD change, if you are using Helm, you might need to update your Helm version and/or do a clean install. to apply it to your cluster if you already have a version of Agones installed.

roberthbailey (approved):

PR #1958, opened 2021-01-20
Refactored sdk functions to always return &alpha.Bool{} instead of nil

### LINKED ISSUE #1957: [Go SDK] PlayerConnect() at capacity panics
**What happened**:

When attempting to call `sdk.alpha.PlayerConnect(id)` when `initialCapacity` was set to 0 causes a panic.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9b61d1]
goroutine 1 [running]:
agones.dev/agones/sdks/go.(*Alpha).PlayerConnect(0xc000134cc0, 0xd67031, 0xe, 0xc0000b46e0, 0x14, 0x0)
        /go/pkg/mod/agones.dev/agones@v1.11.0/sdks/go/alpha.go:55 +0xe1
```

Should not panic, and return both `ok` with a false value and the error.

- Create a GameServer with an initialCapacity of 0 (or use an already full GameServer)
- Attempt to call `PlayerConnect(id)`

There are a few places here:
https://github.com/googleforgames/agones/blob/master/pkg/sdkserver/sdkserver.go#L571

That return null for the ok value, and it should always return a value, otherwise a panic will occur.

We should also test for this in our e2e test:
https://github.com/googleforgames/agones/blob/master/test/e2e/gameserver_test.go#L867

Should also look to see if we can do unit tests on this and/or local SDK tests, as it seems like the same issue exists there too:
https://github.com/googleforgames/agones/blob/master/pkg/sdkserver/localsdk.go#L402

**Environment**:
- Agones version: 1.11
- Kubernetes version (use `kubectl version`): 1.17
- Others:

Thanks to @danieloliveira079 for reporting the issue in Slack.

justjoeyuk:
**What type of PR is this?**
> /kind bug

This PR fixes the issue presented in #1957 wherein the SDK returns a `nil` value for an `ok` value instead of an `alpha.Bool{ Bool: false }`. This PR ensures that the SDK always returns a value for `ok` and that users can expect an `ok` value to always be non-nil.

Work on #1957

This is my first contribution - so I may have made a few errors of judgement in places. I'm happy to hear feedback and make the relevant changes.


justjoeyuk:
@googlebot I signed it!

markmandel:
Thanks for this! 

Can we switch `Closes #1957` to `Work on #1957`?

This is a great first start, but we'll likely also want at least an e2e test to confirm that this actually closes the bug in question, so we won't want to close the issue until that is done.

Unless you want to add an e2e test to this PR, or we can have it happen in a separate PR, either way is totally fine :+1:

markmandel:
> 
> Can we switch `Closes #1957` to `Work on #1957`?
> 

If you are happy with it, I can make this change via GitHub (forgot I can do this) - if we want to push this through before RC next week. 

But wanted to ask permission first before messing with your PR.

Let me know!

markmandel (approved):
LGTM!

PR #1963, opened 2021-01-22
Build changes f33

bnwhorton:
**What type of PR is this?**
> /kind feature

A few changes to allow the build to work on F33 with a small hack

Closes #

Update a few settings to allow the build to work in f33 with podman
Hack: I added a ln -s to podman in $HOME/bin to trick the Makefile, If it would be useful I can look at adding vars to the makefile to allow the option of docker vs podman


build/Makefile
```diff
@@ -79,12 +79,12 @@ helm_cache := ~/.cache/helm
...
 agones_package = agones.dev/agones
 mount_path = /go/src/$(agones_package)
-common_mounts = -v $(build_path)/.config/gcloud:/root/.config/gcloud \
-				-v $(kubeconfig_path):/root/.kube \
-				-v $(helm_config):/root/.config/helm \
-				-v $(helm_cache):/root/.cache/helm \
-				-v $(agones_path):$(mount_path) \
-				-v $(build_path)/.gomod:/go/pkg/mod
+common_mounts = -v $(build_path)/.config/gcloud:/root/.config/gcloud:Z \
+				-v $(kubeconfig_path):/root/.kube:Z \
```
markmandel, line 83: Can you explain what this `:Z` stuff does?

bnwhorton: Tells docker or podman to try to fix any SElinx label issues with the mount point.

https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

markmandel: So this will edit user permissions on the file system? This seems... scary? Just looking at the linked page:

> Use extreme caution with these options.

:grimacing: 

I have reservations just based on that. What's the exact problem this solves? Can you not mount folders with Docker otherwise?

bnwhorton: By default newly created directory wont be selinux labeled properly for bind mounting in to a container. This checks and updates the labels. Another way to approach this so to check for selinux and relabel the dirs in the Makefile. I would defiantly use caution during any hostbind mounts, if possible I would recommend switching to volumes. What that page suggests is if you hostbind /etc or /sys it will cause issues, in general I would highly recommend avoiding to mount anything in /*.

Would it be better if I explored these other options?

bnwhorton: If the dirs are not using outside of the containers, I would really recommend going to volume mounts and not hostbind any dirs.

markmandel: Just mentioned in meeting - we do reuse the local cache and/or configs on the host machine to make things easier, so there are places we could use volume mounts, but definitely not all (i.e. we reuse the .kube config from the host).

Unless someone takes it for a spin before me, I'll take it for a spin on my debian machine and see what the implications are for my local filesystem.

markmandel: I still haven't done this - sorry. Still kinda freaked out about what changes it might make to my local system, so been hanging back in case it borks something and I don't have time to fix it.

markmandel: > Another way to approach this so to check for selinux and relabel the dirs in the Makefile

You mentioned this! We currently have OS checks and command overrides -- maybe that is a better and safer option?

https://github.com/googleforgames/agones/blob/main/build/Makefile#L180-L187

markmandel:
Should we close this PR, it's been a long time open?

markmandel:
Been two weeks, so I'm going to close this PR. Hopefully we can find some alternate ways to support dev on SElinx supporting distros.

PR #1991, opened 2021-02-10
Add "gke" prefix to TCP e2e firewall

markmandel:
**What type of PR is this?**
>
> /kind breaking
> kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Make sure we have the prefix of "gke" on the TCP firewall for the e2e cluster.

Keeps things consistent.

N/A

N/A


roberthbailey (approved):
Does this have any negative side-effects if you run helm upgrade? For fresh installs this doesn't concern me.

markmandel:
This only affects Terraform, and only on the e2e infrastructure, so no impact on Helm!

PR #1944, opened 2021-01-06
[UrealSDK] Creating requests should work in all versions of UE4

domgreen:
**What type of PR is this?**
/kind hotfix

**What this PR does / Why we need it**:

We are replacing the usage if `TSharedRef<IHttpRequest>`  with `FHttpRequestRef` to allow it to work across versions.

It seems that this basically a typedef under the hood to make this work nicely for the current versions.

**Which issue(s) this PR fixes**:
Closes #1940

**Special notes for your reviewer**:
- Have tested with `4.25.1`
- Need to have someone test with `4.26` prior to merging.

PR #1999, opened 2021-02-18
Add WebSocket capability to WatchGameServer REST API

highlyunavailable:
**What type of PR is this?**

/kind feature

This PR adds a websocket proxy for the streaming endpoint `/watch/gameserver`. This allows clients that do not support reading from HTTP streams before they are complete, but _do_ support WebSocket streaming (e.g. Unreal) to use the `/watch/gameserver` endpoint to stream changes.

No issue, but I do want to use this to contribute to the Unreal plugin to add game server watch capabilities.

The PR is split up into multiple commits. I'm unsure as to what the correct practice is for go vendoring, but it wouldn't build without me running `go mod vendor`, so that is included as its own commit.

I'm also not sure how I'd test this - it work in manual tests of course but it doesn't appear there are any REST API tests in the first place so I'm not quite sure where to add them if it's needed.


site/content/en/docs/Guides/Client SDKs/rest.md
```diff
@@ -174,6 +174,9 @@ Response:
 {"result":{"object_meta":{"name":"local","namespace":"default","uid":"1234","resource_version":"v1","generation":"1","creation_timestamp":"1533766607","annotations":{"annotation":"true"},"labels":{"islocal":"true"}},"status":{"state":"Ready","address":"127.0.0.1","ports":[{"name":"default","port":7777}]}}}
 {"result":{"object_meta":{"name":"local","namespace":"default","uid":"1234","resource_version":"v1","generation":"1","creation_timestamp":"1533766607","annotations":{"annotation":"true"},"labels":{"islocal":"true"}},"status":{"state":"Ready","address":"127.0.0.1","ports":[{"name":"default","port":7777}]}}}
 ```
+
+The Watch GameServer stream is also exposed on the same `/watch/gameserver` URL as a WebSocket endpoint using the same newline-delimited JSON format for clients that support WebSocket but not HTTP streaming, such as Unreal.
```
markmandel, line 178: Yay docs!

Can we wrap this in a `feature` shortcode (details on how to do that [here](https://agones.dev/site/docs/contribute/#documentation-for-upcoming-features)) so it only shows up in 1.13.0

Do we want to also be explicit that it's the same port as the HTTP rest api, etc. just to make it super, super clear.

Apparently you can test websockets with `curl` ([gist](https://gist.github.com/htp/fbce19069187ec1cc486b594104f01d0)) - shall we also include an example, that would also be awesome.

markmandel (commented):
This looks cool! :smile: :partying_face: 

So in regards to test, there are integration tests - they are bit janky, because reasons.

The entry point for the REST tests are here: 
- https://github.com/googleforgames/agones/blob/main/build/includes/sdk.mk#L177-L183
- https://github.com/googleforgames/agones/blob/main/build/build-sdk-images/restapi/sdktest.sh

So you can run them locally with `make run-sdk-conformance-test-rest`

:thinking: since the REST sdk test already covers the `watch` event might want to have a specific test folder just for this websocket watch api, rather than try and integrate it with the REST test, as each integration test suite just does a "did this run at least once" check.

It could likely just be a bash command, if we can connect via curl?

Let me know if that list of comments made sense, and if I can point you in the right direction from there :+1: 

(Shame Unreal won't let you put in place http request interception middleware. Unity will :hand_over_mouth: )

markmandel:
Looks like you also ran into some linting issues.

If you want to run our linter locally: `make lint` will do the job. (All the make commands are run from the `build` folder)

See: https://github.com/googleforgames/agones/blob/main/build/README.md

highlyunavailable:
Thanks for your comments and guidance - I believe I've addressed all the issues found so far. Let me know if you see anything else!

markmandel:
Haven't forgotten about this - will review tomorrow :+1:

highlyunavailable:
No problem! I'm not in a hurry - I can use this code and I've got the Unreal plugin addition written but I have a hard dependency on #1944 being merged before I PR that because I'm on 4.26.

test/sdk/websocket-watch/ws-watch-test.go
```diff
@@ -0,0 +1,102 @@
+// Copyright 2019 Google LLC All Rights Reserved.
```
markmandel, line 1: Super minor nit, this should be 2021

markmandel (commented):
One minor nit on t he new file, but otherwise, this is good to go :+1:

test/sdk/websocket-watch/ws-watch-test.go
highlyunavailable: Nit complete! Thanks.

highlyunavailable:
I have also squashed this into a single commit, so it should be good to go now.

markmandel (approved):
Awesome.

PR #2002, opened 2021-02-22
Add RollTable to Companies that use Agones

NullSoldier:
**What type of PR is this?**
/kind documentation

**What this PR does / Why we need it**:
I was recommended to add this to the company list from a comment on our companies blog article here https://medium.com/rolltableapp/scalable-game-servers-using-agones-node-and-digital-ocean-90e6fb9ee5d

> Mark Mandel 6 days ago
> This is awesome! Thanks for sharing your experience with Agones!!!
> If you are interested, we'd love to have the Rolltable logo on https://agones.dev homepage - so please file a PR if this is something you are interested in showcasing!

**Which issue(s) this PR fixes**:
None

**Special notes for your reviewer**:
Agones is awesome and we love the project at RollTable.

PR #2003, opened 2021-02-23
add Vela Games to companies that use Agones

comerford:
**What type of PR is this?**
/kind documentation

Spotted an addition in #2002 so wanted to add Vela Games also (having gotten internal approval to do so)

Basically copied #2002 and local build looked fine - if anything different is required, just let me know


markmandel (approved):
Thank you!

PR #2006, opened 2021-02-25
Add support for empty ports

ItsKev:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> 
/kind feature
> /kind hotfix

This PR adds support for empty ports.

Closes https://github.com/googleforgames/agones/issues/749

This is my first time working with GO. If I can improve anything feel free to point me in the right direction.
Somehow the install.yaml shows a lot of changes. Looking through them in GoLand however, does not show me any changes at all on certain lines.


markmandel:
Looks like Helm is being grumpy with the change to the CRD :grimacing: it's been doing that as of late with the new schema changes, I can handhold that one through e2e testing.

At first glance, this looks good (I'll do a deeper review at some point) - the only thing I would advocate for - somewhere in https://github.com/googleforgames/agones/blob/main/test/e2e/gameserver_test.go write an e2e test that confirms you can create a GameServer with no ports, and nothing falls over.

Thanks for doing this work!

markmandel:
I'm getting this weird error I get sometimes with Github

image

So I'll wait for the e2e test to come in and walk it through the e2e tests then.

If you run into issues / would like some help, please feel free to drop into #development on Slack, and ask questions as well!

ItsKev:
Gonna write the e2e test tomorrow 👍

ItsKev:
I've added an e2e test but was not able to test it locally against a Google Cloud cluster. I get the same error while trying to use `make install` that happens in the tests here.

markmandel:
This looks good, looks like you are hitting https://github.com/helm/helm/issues/5806 which won't get fixed until K8s 1.20 :confused: We're hitting this in a couple of other places.

I'll hand hold this through e2e test shortly.

markmandel:
Hey - can you give me permission to update your branch please? I can't update it to master.

image

ItsKev:
The allow edits by maintainers is not possible for organization repositories.

@markmandel I've invited you to write to the repository. If this is not how you handle things I could fork the project with my private account and create a PR again.

markmandel:
> The allow edits by maintainers is not possible for organization repositories.

AAAAAAAAAAAAAAAAAAAAAAAAAAAH! That's why I can't update SOME PR's!!! This has been driving me up the wall for eons.

Don't give me write permissions on your repo -- that's seems icky.

Ping me on Slack when you want to update this PR to master, and I can handhold it through e2e tests :+1:

markmandel:
Actually, if you are happy, I'll take the write permissions on the repo -- let me know whichever way works best for you.

markmandel:
Can you run `make gen-install` again please? Your new changes combined with the recent updates to the CRDs because of #2007  aren't in that file. Thanks!

markmandel:
Just a heads up - RC is next Tuesday, but let's see if we can squeeze this in.

test/e2e/framework/framework.go
```diff
@@ -210,6 +210,11 @@ func NewFromFlags() (*Framework, error) {
 
 // CreateGameServerAndWaitUntilReady Creates a GameServer and wait for its state to become ready.
 func (f *Framework) CreateGameServerAndWaitUntilReady(ns string, gs *agonesv1.GameServer) (*agonesv1.GameServer, error) {
+	return f.CreateGameServerAndWaitUntilReadyWithOptionalPortCheck(ns, gs, true)
+}
+
+// CreateGameServerAndWaitUntilReadyWithOptionalPortCheck Creates a GameServer and wait for its state to become ready.
+func (f *Framework) CreateGameServerAndWaitUntilReadyWithOptionalPortCheck(ns string, gs *agonesv1.GameServer, portCheck bool) (*agonesv1.GameServer, error) {
```
markmandel, line 217: I'm trying to decide if `CreateGameServerAndWaitUntilReadyWithOptionalPortCheck` is necessary.

We already cover this by checking on ports in [this test](https://github.com/googleforgames/agones/blob/170384f993e4919d31950e5f995924de22fec8ef/test/e2e/gameserver_test.go#L54) for example (I'm sure there are others).

@roberthbailey WDYT? 

I get the impetus to check all other tests, but I'm not sure it's required.

test/e2e/gameserver_test.go
```diff
@@ -769,6 +769,22 @@ func TestGameServerTcpUdpProtocol(t *testing.T) {
...
+func TestGameServerWithoutPort(t *testing.T) {
+	t.Parallel()
+	gs := framework.DefaultGameServer(framework.Namespace)
+	gs.Spec.Ports = nil
+
+	_, valid := gs.Validate()
+	assert.True(t, valid)
+
+	readyGs, err := framework.CreateGameServerAndWaitUntilReadyWithOptionalPortCheck(framework.Namespace, gs, false)
+	if err != nil {
```
markmandel, line 781: Nit:

This `if err != nil` block can be changed to:
`require.NoError(err)`  (I know we have the `assert.FailNow()` pattern in a bunch of places, but that was before we had access to `require` :smile: )

test/e2e/gameserver_test.go
```diff
@@ -769,6 +769,22 @@ func TestGameServerTcpUdpProtocol(t *testing.T) {
...
+		assert.FailNow(t, "Could not get a GameServer ready", err.Error())
+	}
+
+	assert.Nil(t, readyGs.Spec.Ports)
```
markmandel, line 785: Just a thought. This might be better, just because nil vs an empty array are both valid return results. Not a strong opinion.

https://pkg.go.dev/github.com/stretchr/testify/assert#Empty

```suggestion
	assert.Empty(t, readyGs.Spec.Ports) 
```

markmandel (commented):
Yay we passed!

Just had a couple of minor things that I think could be cleaned up. Would also like a second opinion on `CreateGameServerAndWaitUntilReadyWithOptionalPortCheck` :thinking:

test/e2e/gameserver_test.go
ItsKev: Done

test/e2e/gameserver_test.go
ItsKev: Yes, you are right. assert.Empty itself does a null check.

Done

test/e2e/framework/framework.go
roberthbailey: Looking at your new test code, I'm guessing that you added this because without passing in false, line 227 (in the old file) was returning an error, because there was no port. 

But the test should be able to inspect the passed in game server and tell if a port should be created (by seeing if spec.Ports is nil). And then you can modify the check for ports to read something like "if the spec has a port and none was allocated, return an error" instead of what we have currently ("assume that there will always be a port so check to see if one was allocated").

markmandel: https://github.com/googleforgames/agones/blob/main/test/e2e/framework/framework.go#L227 - just for reference :smile: 

That's a really good point @roberthbailey , that could become:

```go
	if len(readyGs.Status.Ports) == len(gsCopy.Spec.Ports) ​{
		return nil, fmt.Errorf("Ready GameServer instance has no port: %v", readyGs.Status)
	}
```

Don't think we need nil checks either, since `len(nil)` is `0`

Which is far more elegant :+1:

roberthbailey: I think you mean != 

Also, the error message needs to be updated to print the expected number of ports and the observed number of ports:


```go
	if len(readyGs.Status.Ports) != len(gsCopy.Spec.Ports) ​{
		return nil, fmt.Errorf("Ready GameServer instance has %d port(s), want %v", readyGs.Status, gsCopy.Spec.Ports)
	}
```

markmandel: Ah yes! I did, excellent catch. I like it!

ItsKev: Thanks for the input. I've updated the PR accordingly. 👍

ItsKev:
The PR got reviewed and is ready to be merged if my changes are accepted.

/assign @cyriltovena

markmandel (approved):
I like it! Let's get this merged!

Issue #1856, opened 2020-10-20
Contentions errors from allocator gRPC service under high load


pooneh-m:
**What happened**:
Performance testing revealed that `agones-allocator` service, returns contention errors under heavy load when there are multiple replicas for the service.

**What you expected to happen**:
The service should allocate game servers under heavy load with no contention error.

**How to reproduce it (as minimally and precisely as possible)**:
Run a performance test with 50 parallel clients and 4000 gameservers.

**Anything else we need to know?**:
The `agones-allocator` pods are caching game servers in memory. Because the state of game servers are changed by different pods, the cache could quickly go out of sync with the state of game servers in the cluster. Either the cache should be changed to a key-value store shared between the pods or allocators watch game server changes and use k8s API to get a ready game server using its labels without caching them. 

Documentations should also provide recommendations on the # of replica for packed and distributed allocation.

**Environment**:
- Agones version: 1.9.0
- Kubernetes version (use `kubectl version`): 1.16
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): Helm
- Troubleshooting guide log(s):
- Others:

markmandel:
I'm assuming this is primarily a problem when we are using the Packing algorithm for allocation (would be good to have performance metrics on Distributed vs Packed error rate and throughput), as since we are trying to bin pack the Allocated game servers, which means that the sorted caching on each allocator binary will generally target the same GameServers.

Some thoughts on this:

- Doing some local performance testing, we can get ~100+ allocations a second (with a quick fix to #1852, PR coming soon). Is this acceptable performance. If so, we could leave this issue alone, and say if you want more, you need to run more clusters.
- I also expect that putting some light randomisation in the allocator (probably [here](https://github.com/googleforgames/agones/blob/master/pkg/gameserverallocations/find.go#L34)) e.g. take the first  _n_/% that are sorted, but randomise them before attempting the allocation, would reduce the amount of contention considerably (maybe _n_ defaults to allocation replica count + 1 for randomisation?). The downside here is we lose some tightness on the packing. This might be a configuration knob we can provide where the user can choose the tradeoff of throughput vs bin packing.

I must say - this has taken me on quite travel down memory lane to remember how allocation works!

markmandel:
One area of research we should also confirm -- how much is the gRPC endpoint actually loadbalanced?

It's quite possible there is only one pod being used at any given point and time with the current load balancing setup.

markmandel:
Another thought, at this point:
[...]

PR #2032, opened 2021-03-28
Allocation: Drop stale GameServer on conflict

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Previously, if an attempt at allocation would fail, it would get placed automatically back into the local cache.

If the failure was because of a Conflict, there is no point in putting it back into the cache, as it will never succeed until it gets updated from the Kubernetes API.

Work on #1856

Running at high load (TESTRUNSCOUNT=1 ./runAllocation.sh 200 20), reduced the number of failed allocations due to contention from 298 down to 180.

### DIFF
--- a/pkg/gameserverallocations/allocator.go
+++ b/pkg/gameserverallocations/allocator.go
@@ -43,6 +43,7 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


roberthbailey (approved):

roberthbailey:
There was an error in the rust conformance test:

```
error[E0658]: `match` is not allowed in a `const fn`
   --> /go/src/agones.dev/agones/test/sdk/rust/.cargo/registry/src/github.com-1ecc6299db9ec823/socket2-0.4.0/src/lib.rs:156:9
    |
156 | /         match address {
157 | |             SocketAddr::V4(_) => Domain::IPV4,
158 | |             SocketAddr::V6(_) => Domain::IPV6,
159 | |         }
    | |_________^
    |
    = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information

error: aborting due to previous error

For more information about this error, try `rustc --explain E0658`.
error: could not compile `socket2`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
includes/sdk.mk:88: recipe for target 'run-sdk-command' failed
make[2]: *** [run-sdk-command] Error 101
```

I'm going to try again and see if it's reproducible.

roberthbailey:
The test failed with the same rust error, so it looks like it isn't a flake.

roberthbailey:
I'm able to reproduce the error locally.

markmandel:
> I'm able to reproduce the error locally.

I'm struggling to reproduce locally. Care to share your steps?

I must have a cache library version of something locally I'm sure.

markmandel:
Ah - worked it out, I had a Cargo.lock sitting around that pointed to an older version of the library.

PR #2035, opened 2021-03-30
Use the correct feature flag name (and guard it properly).

roberthbailey:
**What type of PR is this?**
>
> /kind breaking

 /kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**: Fixes a test case, which was not testing what it was intended to test. 

Closes #

### DIFF
--- a/pkg/fleets/controller_test.go
+++ b/pkg/fleets/controller_test.go
@@ -196,7 +196,9 @@ func TestControllerSyncFleet(t *testing.T) {
 	})
 
 	t.Run("gameserverset with different image details", func(t *testing.T) {
-		assert.NoError(t, utilruntime.ParseFeatures(string(utilruntime.FeaturePlayerTracking)+"=true"))
+		utilruntime.FeatureTestMutex.Lock()
+		defer utilruntime.FeatureTestMutex.Unlock()
+		assert.NoError(t, utilruntime.ParseFeatures(string(utilruntime.FeatureRollingUpdateOnReady)+"=true"))
 
 		f := defaultFixture()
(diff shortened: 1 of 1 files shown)


pkg/fleets/controller_test.go
```diff
@@ -196,7 +196,9 @@ func TestControllerSyncFleet(t *testing.T) {
 	})
 
 	t.Run("gameserverset with different image details", func(t *testing.T) {
-		assert.NoError(t, utilruntime.ParseFeatures(string(utilruntime.FeaturePlayerTracking)+"=true"))
+		utilruntime.FeatureTestMutex.Lock()
```
roberthbailey, line 199: This was the only place in our test code that didn't use the feature test mutex, which could have caused a race condition.

pkg/fleets/controller_test.go
```diff
@@ -196,7 +196,9 @@ func TestControllerSyncFleet(t *testing.T) {
...
+		defer utilruntime.FeatureTestMutex.Unlock()
+		assert.NoError(t, utilruntime.ParseFeatures(string(utilruntime.FeatureRollingUpdateOnReady)+"=true"))
```
roberthbailey, line 201: Since this was setting the wrong feature to true, the test was always using the default feature value for rolling update on ready, which made it break when I changed the default.

roberthbailey:
Our helm install is broken again: `Error: UPGRADE FAILED: another operation (install/upgrade/rollback) is in progress`

roberthbailey:
Tests failed with `fatal error: concurrent map read and map write`.

markmandel (approved):
Good catch!

Issue #54, opened 2018-01-09
Running windows game server


cyriltovena:
Game production usually works on windows first then port to linux eventually at the end of the development. Having windows support would help the adoption rate.

What does it takes to run a windows game server ?

- We should find a way to detect that the game server resource is windows ? May be a resource parameter ?
- Does windows support sidecar concept ? Apparently from [here at the bottom](http://blog.kubernetes.io/2017/09/windows-networking-at-parity-with-linux.html) it says windows support only one container per pod, but should work on windows server 1709. We should try it.
- If yes, create a sidecar using [windows nano](https://hub.docker.com/r/microsoft/nanoserver/).
- add a windows game server example.

Testing will be difficult as windows support is still in beta since k8s 1.5 but apparently greatly improve in 1.9.

Documentation :
https://kubernetes.io/docs/getting-started-guides/windows/
http://blog.kubernetes.io/2017/09/windows-networking-at-parity-with-linux.html
https://github.com/kubernetes/community/tree/master/sig-windows
https://docs.microsoft.com/en-us/windows-server/get-started/whats-new-in-windows-server-1709

markmandel:
Today - there is an update!
https://kubernetes.io/blog/2019/03/25/kubernetes-1-14-release-announcement/

Kuqd:
Kind issue to look at: https://github.com/kubernetes-sigs/kind/issues/410

markmandel:
Big news!!!

https://cloud.google.com/kubernetes-engine/docs/release-notes?hl=en

> The ability to create clusters with node pools running Microsoft Windows Server is now in Beta. This feature is currently only available in the Rapid release channel.

EricFortin:
I have taken a stab at this in the past few days. I somewhat succeeded and thought I would recap my work in this issue so that we may be able to make it happen.

**Disclaimer: I was mainly interested in seeing if this could even work. I hacked a few things and work around others.**

### Building sidecar and simple udp images for Windows

#### SDK

* I built the sdk binary from Agones build system for Windows with following command:  `make build-agones-sdk-binary-windows`. *I had to switch to Go to at least 1.13.6 because of a bug in Windows: https://github.com/golang/go/issues/35447*
* I then wrote a simple `Dockerfile` to build an image. I skipped some security(like creating a user) in order to go faster.

Here's an example of my final `Dockerfile`:

```
[...]

PR #2037, opened 2021-03-31
Build Agones Windows images by default.

jeremyje:
**What type of PR is this?**
/kind feature
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Builds Windows container images by default so that they are easier to use out of the box.
This is done by turning on the `WITH_WINDOWS=1` flag in the Makefile.

Issue #54 is mostly complete.

This will add some latency to the Agones build.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -49,8 +49,10 @@ MINIKUBE_PROFILE ?= agones
 GO_BUILD_TAGS ?= none
 BUILDX_WINDOWS_BUILDER = windows-builder
 WINDOWS_DOCKER_PUSH_ARGS =
-WINDOWS_VERSIONS = ltsc2019 # 1909 (For GKE WINDOWS_SAC support)
-BUILD_WINDOWS_IMAGE =
+# Versions of Windows Server to support, default is Windows Server 2019.
+# For the full list of tags see https://hub.docker.com/_/microsoft-windows-servercore.
+# GKE-Windows version map: https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-cluster-windows#version_mapping
+WINDOWS_VERSIONS = ltsc2019
 
 # Specify stress test level 1..100
 # STRESS_TEST_LEVEL=n requires capacity between 50*n up to 100*n simple-game-server Game Servers.
@@ -64,6 +66,9 @@ KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 GS_TEST_IMAGE ?= gcr.io/agones-images/simple-game-server:0.2
 
 ALPHA_FEATURE_GATES ?= "PlayerTracking=true&NodeExternalDNS=true"
+
+# Build with Windows support
+WITH_WINDOWS=1
 
 # Directory that this Makefile is in.
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
@@ -94,7 +99,7 @@ build_tag = agones-build:$(build_version)
 controller_tag = $(REGISTRY)/agones-controller:$(VERSION)
 sidecar_tag = $(REGISTRY)/agones-sdk:$(VERSION)
 
-ifdef WITH_WINDOWS
+ifeq ($(WITH_WINDOWS), 1)
 sidecar_linux_amd64_tag = $(sidecar_tag)-linux_amd64
 else
 sidecar_linux_amd64_tag = $(sidecar_tag)
@@ -216,7 +221,7 @@ build: build-images build-sdks
 
 # build the docker images
 build-images: build-controller-image build-agones-sdk-image build-ping-image build-allocator-image
-ifdef WITH_WINDOWS
+ifeq ($(WITH_WINDOWS), 1)
 build-images: build-agones-sdk-image-windows
 endif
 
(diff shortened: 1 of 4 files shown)


markmandel:
Interesting - I thought this was going to add a lot of time to the build process, but it completed in 00:23:06 - which is about average for our usual build time - so that's pretty great!

build/Makefile
```diff
@@ -49,7 +49,7 @@ MINIKUBE_PROFILE ?= agones
 GO_BUILD_TAGS ?= none
 BUILDX_WINDOWS_BUILDER = windows-builder
 WINDOWS_DOCKER_PUSH_ARGS =
-WINDOWS_VERSIONS = ltsc2019 # 1909 (For GKE WINDOWS_SAC support)
+WINDOWS_VERSIONS = ltsc2019 # 20H2 (For GKE WINDOWS_SAC support)
```
markmandel, line 52: Wondering if we can add a comment/link here, about what these terms mean?

I'm not sure what the change to 20H2 means?

jeremyje: 1909 is going end of life soon. GKE now supports 20H2.

build/Makefile
```diff
@@ -65,6 +65,9 @@ GS_TEST_IMAGE ?= gcr.io/agones-images/simple-game-server:0.2
 
 ALPHA_FEATURE_GATES ?= "PlayerTracking=true&RollingUpdateOnReady=true&NodeExternalDNS=true"
 
+# Build with Windows support
+WITH_WINDOWS=1
```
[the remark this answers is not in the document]
jeremyje: Good point I'll make that change in a few hours.

build/Makefile
markmandel: I don't know what those words mean 🙂 can we add a comment to point to the relevant docs somewhere?

Is the comment of `20H2` mean that we need to change something eventually? I'd just like this comment to be explicitly clear.

cloudbuild.yaml
```diff
@@ -314,8 +314,8 @@ substitutions:
   _RUST_SDK_BUILD_CACHE_KEY: rust-sdk-build
 
 tags: ['ci']
-timeout: "2h"
-images: ['gcr.io/$PROJECT_ID/agones-controller', 'gcr.io/$PROJECT_ID/agones-sdk', 'gcr.io/$PROJECT_ID/agones-ping']
+timeout: 7200s
+images: ['gcr.io/$PROJECT_ID/agones-controller', 'gcr.io/$PROJECT_ID/agones-ping']
```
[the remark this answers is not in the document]
markmandel: So what having the `agones-sdk` in the list of artifacts causing an error of some kind?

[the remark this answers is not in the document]
jeremyje: I moved the `--purge` flag on manifest create. Hopefully that works.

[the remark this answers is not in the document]
jeremyje: It unfortunately did not work. I think what's likely happening is that GCB see's that the local copy of the image is different than the pushed version so it basically overwrites the manifest with the linux-only image. When you pull an image locally you only get a copy that's compatible with the host operating system. This is likely a bug that needs to be reported to GCB.

That said the `images:` clause here is redundant with the push image phase in the cloud build flow. If you look at any of the GCB logs you'll see at the end that a push flow is happening after cloudbuild.yaml is finished.

build/Makefile
markmandel: Is this the right link for details on what these terms mean?
https://cloud.google.com/kubernetes-engine/docs/concepts/node-images#windows-server

If so, can we add a comment above for reference this line for reference?

markmandel:
Outside that adding one comment around what the terms mean for Windows container versions, this LGTM. 

@roberthbailey WDYT?

build/Makefile
jeremyje: Added some docs about versioning.

markmandel: Perfect! Thanks so much!

examples/simple-game-server/Makefile
```diff
@@ -25,13 +25,18 @@
...
+WINDOWS_VERSIONS = ltsc2019
 WINDOWS_DOCKER_PUSH_ARGS = # When pushing set to --push.
+# Build with Windows support
+WITH_WINDOWS=1
 
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 project_path := $(dir $(mkfile_path))
 server_tag = $(REGISTRY)/simple-game-server:0.2
-ifdef WITH_WINDOWS
+ifeq ($(WITH_WINDOWS), 1)
```
roberthbailey, line 39: Just so I'm clear, this means that on the next release, the section "Images available with this release:" (e.g. [here](https://agones.dev/site/blog/2021/03/16/1.13.0-kubernetes-1.18-websocket-watch-sdk-empty-ports/)) will include: `gcr.io/agones-images/simple-game-server:0.2-linux_amd64` and `gcr.io/agones-images/simple-game-server:0.2-windows_amd64-ltsc2019` and `gcr.io/agones-images/simple-game-server:0.2` (where the last one will now be a multi-arch image. Is that correct?

For the multi-arch image, does pulling it use more disk space (pulling the windows & linux versions) or is docker smart enough to only pull for your architecture (I think the latter is true but haven't checked in a long time)? I want to make sure the release notes explain how this will affect users. 

The same question applies to the sdk sidecar image (above) since that is the other image that we will now be publishing this same way.

roberthbailey (commented):
I can't seem to add to the conversation above, but when you say 

> 1909 is going end of life soon. GKE now supports 20H2.

and I look at https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-cluster-windows#version_mapping it appears that for k8s 1.18 (which is the version we currently support) you should still be using 1909 Core. And when we upgrade to k8s 1.19 is when we would switch to 20H2 Core. 

I'm not sure how ltsc2019 maps to either of those, and the note was just in the comment before, but I can't tell which version we are building for in our CI and will build for our upcoming release (when we will build these images for the first time). We should make sure we are building targeting 1909 (even though it's almost EOL) so that it matches our supported k8s version on GKE.

examples/simple-game-server/Makefile
jeremyje: Windows Server 1909 will stop working on May 11, 2021. It's not productive to support it. The builds will not build Windows SAC flavors of the image only LTSC.
When docker queries an image registry if an image manifest is returned it'll select the most appropriate image based on the host's characteristics. I believe there's an API to basically have select the correct image for you without a second request. as well.

Additional space will be used for each version of the image but the windows base layers are stored on mcr.microsoft.com so Windows Server Core is not stored in gcr.io. Only the layers on top this is due to the foreign layers bit Microsoft sets on these images.

jeremyje: I think it'll be better just to remove references to Windows SAC altogether. It appears to be adding confusion.

roberthbailey (approved):

PR #2039, opened 2021-04-01
Update app labels to all use "agones.name" and not "agones.fullname"

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

 /kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**: If you inspect the labels on the installed resources from helm, the `app` label is currently used inconsistently. In most places it is just the charge name (e.g. `agones`) but in some resources it is the "fullname", which prepends the chart name with the release name (e.g. `my-release-agones`). Since we have a separate label with the release name and most places just use the chart name, this brings everything in line to consistently use the chart name for all occurrences of the `app` label. 

Closes #

### DIFF
--- a/install/helm/agones/templates/extensions.yaml
+++ b/install/helm/agones/templates/extensions.yaml
@@ -145,7 +145,7 @@ metadata:
   name: {{ template "agones.fullname" . }}-cert
   namespace: {{ .Release.Namespace }}
   labels:
-    app: {{ template "agones.fullname" . }}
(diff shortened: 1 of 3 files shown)


roberthbailey:
I forgot to generate install.yaml: `Makefile:298: recipe for target 'test-install-yaml' failed`.

I've pushed another commit.

markmandel (approved):

PR #2056, opened 2021-04-14
Update the recommended machine type to use when creating GKE clusters.

roberthbailey:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
> /kind cleanup

 /kind documentation

> /kind feature
> /kind hotfix

**What this PR does / Why we need it**: Ref https://github.com/googleforgames/agones/issues/2055

**Which issue(s) this PR fixes**:

Closes #

**Special notes for your reviewer**:

PR #2063, opened 2021-04-16
Update the machine type for GKE clusters in build scripts and terraform modules.

### LINKED ISSUE #2055: Update recommended machine type for GKE
n1 machine types are being phased out. We should update our recommended machine type, either to n2 or e2. 

I think e2 should be fine for the getting started guides, and presumably anyone using agones for production is already changing the machine type from the default in any case.

roberthbailey:
**What type of PR is this?**
>

 /kind breaking

> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**: Removes all remaining references to n1-standard machine type from the repository. 

I've marked this change as breaking since it changes the machine type in the terraform files. 

Closes #2055

**Special notes for your reviewer**: This PR also addresses a comment from #2056 that was made after the PR was merged.

### DIFF
--- a/site/content/en/docs/Installation/Terraform/gke.md
+++ b/site/content/en/docs/Installation/Terraform/gke.md
@@ -80,6 +80,7 @@ The GKE cluster created from the example configuration will contain 3 Node Pools
 
 Configurable parameters:
 
+{{% feature expiryVersion="0.14.0" %}}
 - project - your Google Cloud Project ID (required)
 - name - the name of the GKE cluster (default is "agones-terraform-example")
 - agones_version - the version of agones to install (an empty string, which is the default, is the latest version from the [Helm repository](https://agones.dev/chart/stable))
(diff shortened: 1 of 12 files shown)


site/content/en/docs/Installation/Terraform/gke.md
```diff
@@ -80,6 +80,7 @@ The GKE cluster created from the example configuration will contain 3 Node Pools
 
 Configurable parameters:
 
+{{% feature expiryVersion="0.14.0" %}}
 - project - your Google Cloud Project ID (required)
 - name - the name of the GKE cluster (default is "agones-terraform-example")
 - agones_version - the version of agones to install (an empty string, which is the default, is the latest version from the [Helm repository](https://agones.dev/chart/stable))
```
pooneh-m, line 86: You could just put this line for n1-standard-4 in the feature expiry and replace it with the new note.

roberthbailey: I thought that doing it for a single line in a list breaks rendering for the whole list (but I could be wrong). Since the release candidate is tomorrow, I'm leaning towards merging this since it'll be removed soon in any case. WDYT?

pooneh-m: That's true for the table, but I am not sure about a list. You can locally verify using `make site-server`.
I am fine with either. You already have my approval.

roberthbailey: I did a local test and you are correct - you can just wrap the one list item with the feature expiry.

roberthbailey: Since this will be cleaned up in a week and I'd like to get this in before the release candidate gets cut, I'm going to merge as-is.

roberthbailey: Looking more closely, while it works, it changes the spacing of the list elements, which I find off-putting. So I think that doing it around the entire list is preferable.

PR #2070, opened 2021-04-21
Rename MetaPatch to Metadata for AllocationRequest

### LINKED ISSUE #2042: Allocation endpoint: Deprecate `metaPatch` for `metadata`
Context:
https://github.com/googleforgames/agones/blob/release-1.13.0/proto/allocation/allocation.proto#L53

We have a concept disconnect currently between a `GameServerAllocation` and the Allocation Service, in that when passing labels or annotations on a GameServerAllocation it is done with a `metadata` element, and the Allocation Service it is done with a `metaPatch`.

I've seen a few users get caught up by this (and we have docs for it now too).

I would advocate we deprecate (but keep the functionality) of `metaPatch` in the gRPC Allocation service proto and replace it with `metadata` to avoid a change in concept name all the way from `GameServerAllocation` ➡️  Allocation Endpoint

Probably something like:

```proto
message AllocationRequest {
  // The k8s namespace that is hosting the targeted fleet of gameservers to be allocated
  string namespace = 1;

  // If specified, multi-cluster policies are applied. Otherwise, allocation will happen locally.
  MultiClusterSetting multiClusterSetting = 2;

  // The required allocation. Defaults to all GameServers.
  LabelSelector requiredGameServerSelector = 3;

  // The ordered list of preferred allocations out of the `required` set.
  // If the first selector is not matched, the selection attempts the second selector, and so on.
  repeated LabelSelector preferredGameServerSelectors = 4;

  // Scheduling strategy. Defaults to "Packed".
  SchedulingStrategy scheduling = 5;
  enum SchedulingStrategy {
    Packed = 0;
    Distributed = 1;
  }

  // Deprecated: Please use metadata instead.
  MetaPatch metaPatch = 6;

  // metadata is optional custom metadata that is added to the game server at
  // allocation You can use this to tell the server necessary session data
  MetaPatch metadata = 7;
}
```

And then I believe we could handle the conversion of both in the [converter](https://github.com/googleforgames/agones/blob/release-1.13.0/pkg/allocation/converters/converter.go) library.

lambert-wang:
**What type of PR is this?**
/kind cleanup

Renames the metaPatch field in AllocationRequest to metadata while retaining backwards compatibility.

Closes #2042


pkg/allocation/converters/converter.go
```diff
@@ -50,10 +50,10 @@ func ConvertAllocationRequestToGSA(in *pb.AllocationRequest) *allocationv1.GameS
 		}
 	}
 
-	if in.GetMetaPatch() != nil {
+	if in.GetMetadata() != nil {
```
pooneh-m, line 53: you should accept either of MetaPatch or Metadata being set, until MetaPatch is removed.
If both are set, reject the call.

site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -173,7 +173,7 @@ You should expect to see the following output:
 
 ### Sending Data to the Game Server
 
-The service accepts a `metaPatch` field, which can be used to apply `labels` and `annotations` to the allocated `GameServer`.
+The service accepts a `metadata` field, which can be used to apply `labels` and `annotations` to the allocated `GameServer`.
```
markmandel, line 176: This will need to be have a `feature` shortcode wrapper please. 

See https://agones.dev/site/docs/contribute/#within-a-page for an example.

🤔 Do we want to have a note here about the deprecated metapatch?

test/e2e/allocator_test.go
```diff
@@ -76,7 +76,7 @@ func TestAllocator(t *testing.T) {
 		RequiredGameServerSelector:   &pb.LabelSelector{MatchLabels: map[string]string{agonesv1.FleetNameLabel: flt.ObjectMeta.Name}},
 		PreferredGameServerSelectors: []*pb.LabelSelector{{MatchLabels: map[string]string{agonesv1.FleetNameLabel: flt.ObjectMeta.Name}}},
 		Scheduling:                   pb.AllocationRequest_Packed,
-		MetaPatch:                    &pb.MetaPatch{Labels: map[string]string{"gslabel": "allocatedbytest"}},
+		Metadata:                     &pb.MetaPatch{Labels: map[string]string{"gslabel": "allocatedbytest"}},
```
markmandel, line 79: 🤔 do we want to test both metadata and metapatch -- just to ensure the deprecated version continues to work?

pkg/allocation/converters/converter.go
```diff
@@ -50,10 +50,19 @@ func ConvertAllocationRequestToGSA(in *pb.AllocationRequest) *allocationv1.GameS
...
 	}
 
-	if in.GetMetaPatch() != nil {
+	// Accept both metadata (preferred) and metapatch until metapatch is fully removed.
+	metadata := in.GetMetadata()
+	if metadata == nil {
+		metadata = in.GetMetaPatch()
+	} else if in.GetMetaPatch() != nil {
+		// If both are set, reject the call.
+		return nil
```
pooneh-m, line 59: you need to throw a meaningful exception. The exception should be thrown before making this call at the input validation or you need to return error from the converter.

test/e2e/allocator_test.go
pooneh-m: +1 also test for the input validation case when both MetaPatch and Metadata is set. You can do more rigorous testing in unit tests.

lambert-wang: Additional test cases for the metapatch and metadata field were added to `converter_test.go`

pkg/allocation/converters/converter_test.go
```diff
@@ -147,13 +150,72 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
...
+		},
 	}
 	for _, tc := range tests {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
-			out := ConvertAllocationRequestToGSA(tc.in)
+			out, err := ConvertAllocationRequestToGSA(tc.in)
+			if !cmp.Equal(err, tc.wantError, cmpopts.EquateErrors()) {
```
markmandel, line 215: Any reason not to use `assert` or the `require` test assertion libraries we already use?

Seems like it does the same thing, and we don't need to add an extra library.

pkg/allocation/converters/converter_test.go
```diff
@@ -147,13 +150,72 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
...
+				t.Errorf("\"%s\" error: got %v, want %v", tc.name, err, tc.wantError)
+			}
+
 			if !assert.Equal(t, tc.want, out) {
 				t.Errorf("mismatch with want after conversion: \"%s\"", tc.name)
```
markmandel, line 220: This line could also be cleaned up if you are so inclined while you are in here 😄 the inner message could be moved into the `assert.Equals(...)` statement.

https://pkg.go.dev/github.com/stretchr/testify/assert#Equal

pkg/allocation/converters/converter_test.go
lambert-wang: I was following an example from the Google internal unit test examples, but I should probably just use `assert` so I don't need to include and additional libraries. I'll clean this up.

markmandel:
LGTM! @pooneh-m anything on your end?

pkg/allocation/converters/converter.go
```diff
@@ -77,7 +88,7 @@ func ConvertGSAToAllocationRequest(in *allocationv1.GameServerAllocation) *pb.Al
 			Enabled: in.Spec.MultiClusterSetting.Enabled,
 		},
 		RequiredGameServerSelector: convertInternalLabelSelectorToLabelSelector(&in.Spec.Required),
-		MetaPatch: &pb.MetaPatch{
+		Metadata: &pb.MetaPatch{
```
pooneh-m, line 91: This wouldn't work for [multi-cluster allocation](https://github.com/googleforgames/agones/blob/d8748dfb82dde85994209b2b8505ef9ee652a818/pkg/gameserverallocations/allocator.go#L350-L353). If one cluster has the field and another one does not have the field yet, it will ignore the `metadata`. In that case, we should do a double write to both `MetaPatch` and `Metadata`. So my earlier comment to return error for double write, wouldn't work. You can instead on the API add a comment that `MetaPatch` is ignored if `Metadata` is set.

markmandel: Oh yeah - that's a good catch!

lambert-wang: I've added the double write in the GSA to allocation request conversion and updated the tests accordingly.

pooneh-m (approved):
Thanks for the change.

PR #2075, opened 2021-04-24
Update logger dependency

shawticus:
<!--  Thanks for sending a pull request!  Here are some tips for you:

**What type of PR is this?**
/kind cleanup

Uses an ancient version the joonix logger which has the same API but a different constructor func. Updates the logger dependency and called NewFormatter instead of FluentdFormatter at line 37 of pkg/util/runtime.go

Benefit here is that it doesn't conflict with with other uses of the formatter.

API is visible here: https://github.com/joonix/log
Change was here: https://github.com/joonix/log/commit/7b18c88b187dd428c41cb92f038fb0c9e68fa65a#diff-cd0d0ec8574c5b5d734dde8b50b74f5fff4d5cec387501cb0976ab8b0824c8a9
Code that needs to be changed in Agones is just this one call: https://github.com/googleforgames/agones/blob/094ec10ecce6fa5e5c206e74cbb8616d8712f480/pkg/util/runtime/runtime.go#L37


shawticus:
@googlebot I signed it!

go.mod
```diff
@@ -11,27 +11,29 @@ require (
...
-	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
+	golang.org/x/net v0.0.0-20210316092652-d523dce5a7f4
+	golang.org/x/sys v0.0.0-20210423185535-09eb48e85fd7 // indirect
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
-	golang.org/x/tools v0.0.0-20190920225731-5eefd052ad72
-	google.golang.org/genproto v0.0.0-20191009194640-548a555dbc03
-	google.golang.org/grpc v1.26.0
+	golang.org/x/tools v0.1.0
+	google.golang.org/genproto v0.0.0-20210423144448-3a41ef94ed2b
+	google.golang.org/grpc v1.36.1
```
markmandel, line 36: Upgrading gRPC is a big change and I don't think would be not part of updating logging tooling? Please revert this change.

go.mod
```diff
@@ -11,27 +11,29 @@ require (
 	github.com/aws/aws-sdk-go v1.16.20 // indirect
 	github.com/fsnotify/fsnotify v1.4.7
 	github.com/go-openapi/spec v0.19.3
-	github.com/golang/protobuf v1.3.2
+	github.com/golang/protobuf v1.5.2
```
markmandel, line 14: Not sure why we're upgrading protobuf. Seems like a big change for upgrading a logging library.

markmandel (changes requested):
Thanks for the PR, but unfortunately we can't accept it in it's current state.

While the upgrade to the logging library is a nice cleanup, the upgrades to grpc and related tooling is a much bigger change -- and I expect unrelated to the logging library update?

(There's also several other updates that are unrelated to logging in here. Might be worth either removing them or updating the PR title and description).

Out of curiosity - is there a reason for the logrus library update?

shawticus:
@markmandel thanks for being nice. While I am feeling great about golang as a language I still am really confused by the package ecosystem still. I'm going to do some learning and reopen this PR when I know what I'm doing :)

PR #2080, opened 2021-04-28
Minor updates to the release checklist.

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Closes #

### DIFF
--- a/docs/governance/templates/release_issue.md
+++ b/docs/governance/templates/release_issue.md
@@ -18,7 +18,7 @@ and copy it into a release issue. Fill in relevant values, found inside {}
 ## Steps
 
 - [ ] Run `make shell` and run `gcloud config configurations activate agones-images`.
-- [ ] Review [closed issues with no milestone](https://github.com/googleforgames/agones/issues?q=is%3Aissue+is%3Aclosed+no%3Amilestone) and add relevant ones tothe current milestone.
+- [ ] Review [closed issues with no milestone](https://github.com/googleforgames/agones/issues?q=is%3Aissue+is%3Aclosed+no%3Amilestone) and add relevant ones to the current milestone.
   - Issues tagged as `invalid`, `duplicate`, `question`, or `wontfix` can be ignored
 - [ ] Review closed issues in the current milestone to ensure that they have appropriate tags.
 - [ ] Review [merged PRs that have no milestone](https://github.com/googleforgames/agones/pulls?q=is%3Apr+is%3Amerged+no%3Amilestone+) and add them to the current milestone.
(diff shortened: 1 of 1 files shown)


docs/governance/templates/release_issue.md
```diff
@@ -79,7 +78,7 @@ and copy it into a release issue. Fill in relevant values, found inside {}
 - [ ] If full release move [helm `tag` value][values] is set to {version}+1-dev
 - [ ] If full release move the [helm `Chart` version values][chart] is to {version}+1-dev
 - [ ] If full release move the [`sdks/nodejs/package.json`][nodejs] to {version}+1-dev
```
steven-supersolid, line 80: I think this is the correct step but we should replace with the following command
```
cd sdks/nodejs
npm version preminor --preid=dev
```
I.e. change to:
```
- [ ] If full release, change to the `sdks/nodejs` directory and run the command `npm version preminor --preid=dev` to update the package version
```

steven-supersolid, line 80:
```suggestion
- [ ] If full release, change to the `sdks/nodejs` directory and run the command `npm version preminor --preid=dev` to update the package version
```

markmandel:
Changes LGTM! @roberthbailey do we want to include @steven-supersolid 's suggestions here?

docs/governance/templates/release_issue.md
roberthbailey: That formats the version slightly differently, e.g.

`version": "1.16.0-dev.0"` instead of `version": "1.16.0-dev`

roberthbailey:
> @roberthbailey do we want to include @steven-supersolid 's suggestions here?

I've tinkered with the `npm version` command a bit and I can't figure out how to get it to generate version strings that match what we have been doing by hand. I think it makes sense to merge this without that change (so that the typos get fixed) and then open a separate PR to apply @steven-supersolid's suggestions.

docs/governance/templates/release_issue.md
steven-supersolid: Removing the trailing `.0` does not seem to be supported by npm and I think is due to how semver is handled internally (and by the semver package)
https://docs.npmjs.com/cli/v6/commands/npm-version
https://github.com/npm/node-semver

I think it is this way to support publishing multiple dev versions prior to a release, because it is not possible to re-publish an existing version. Perhaps idiomatic Node.js

An alternative would be specify the version explicitly rather than rely on the increment feature of npm - may be better for this project?
```
npm version 1.16.0-dev
```

That seems clearer

I'm hoping we can avoid the situation where we manually edit `package.json` as we must then run `npm install` to update `package-lock.json`

markmandel (approved):

docs/governance/templates/release_issue.md
roberthbailey: Marking as resolved, since I've moved this comment thread over to #2120.

PR #2088, opened 2021-05-01
i think public ips in aks must be default

dzmitry-lahoda:
also could move to vars, with default true

### DIFF
--- a/install/terraform/modules/aks/aks.tf
+++ b/install/terraform/modules/aks/aks.tf
@@ -68,7 +68,10 @@ resource "azurerm_kubernetes_cluster" "agones" {
     vm_size             = var.machine_type
     os_disk_size_gb     = var.disk_size
     enable_auto_scaling = false
-    #enable_node_public_ip = true
+    # VMSS is used, so it is unpredictable how NICs will be given to VMs
+    # So let Azure to create NICs with Public IPs as games examples require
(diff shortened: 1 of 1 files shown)


markmandel:
I have to ask - is this something that HAS to be set to true for all container types? Or is it just that you think so 😄 

Should this be a variable that people can set, with a sensible default? Rather than a setting for all clusters?

dzmitry-lahoda:
Without public IP you have to create NIC(Network Interface) + Public IP resources manually(or via SDK). When flag is set to true, Azure does that automatically.  It is much more sane for default to have that set to true. 

When the autoscale flag (which is set to false) is set to true - it makes even more sense.

So the flag is not must to set (you can create NIC and Public IP manually after Azure allocated VM for cluster), but it is a much more reasonable default I think (so it should be variable set to true).

Behaviour is same for both Linux and Windows.

markmandel:
>  It is much more sane for default to have that set to true.

Assuming I am vaguely understanding how AKS works 😄  Sounds like then we should have a variable (like you originally also suggested), wtih a default of "true" for enabling public IP addresses.

We should defintely update the documentation https://agones.dev/site/docs/installation/terraform/aks/ as well.

Not necessarily for this PR, but should https://agones.dev/site/docs/installation/creating-cluster/aks/be updated as well?

PR #2092, opened 2021-05-05
Add Afterverse logo

jose-cieni-playkids:
**What type of PR is this?**
/kind documentation

Adds Afterverse logo to Companies that use Agones

None

I set the width to 120px, as our logo is smaller in height.
If this is a problem, let me know and I'll change it back to 100px.


markmandel:
Oooh! Thanks!

🤔 I'm thinking it would best to be 100px, just because fairness.

I did some developer tools editing, and I think it looks fine at 100px. WDYT?

image

(We have a few other wide logos, and they are 100px, so probably be good to stick to things)

Unless you have a vertical lockup you want to use? I see you use the `AV` icon on a couple of places. Just a thought - but entirely up to you!

jose-cieni-playkids:
@markmandel I think we're only using the AV logo for profile icons, but I'm not sure... no worries though, we'll just change it back to 100px 😄  Thank you!

markmandel (approved):

PR #2102, opened 2021-05-14
addin http probe to simple game server

dzmitry-lahoda:
This PR add netcat command to sample server in Go. The commands sends HTTP GET request to URL provided. This is used to check possibility of environment to reach HTTP URLs to download resources.


markmandel:
I appreciate the effort, but I'm curious - what's the reasoning for taking on this feature in the simple-server?

I'm not sure I'm seeing how this helps the project specifically, considering we have to maintain this going forward.

PR #2103, opened 2021-05-14
aks setup improvements

dzmitry-lahoda:
# md

- go plugin to aks is 2 years old, it does not works (error) and MS have done that stuff into cluster, so removed. the person who did plugin working in Playfab
- so put public IP into console and documented how to reach it

# tf

- made public IP variable (default true)
- tf validate fmt passed

### DIFF
--- a/install/terraform/modules/aks/aks.tf
+++ b/install/terraform/modules/aks/aks.tf
@@ -63,12 +63,12 @@ resource "azurerm_kubernetes_cluster" "agones" {
   kubernetes_version = "1.18.14"
 
   default_node_pool {
-    name                = "default"
-    node_count          = var.node_count
-    vm_size             = var.machine_type
-    os_disk_size_gb     = var.disk_size
-    enable_auto_scaling = false
-    #enable_node_public_ip = true
+    name                  = "default"
+    node_count            = var.node_count
+    vm_size               = var.machine_type
+    os_disk_size_gb       = var.disk_size
+    enable_auto_scaling   = false
+    enable_node_public_ip = var.enable_node_public_ip
     #vnet_subnet_id     = azurerm_subnet.aks.id
   }
 
--- a/site/content/en/docs/Installation/Creating Cluster/aks.md
+++ b/site/content/en/docs/Installation/Creating Cluster/aks.md
@@ -67,7 +67,14 @@ az network nsg rule create \
 
 ### Getting Public IPs to Nodes
 
-Resource Group where the AKS resources are installed on the [portal](https://portal.azure.com) are named like `MC_resourceGroupName_AKSName_westeurope`). Find `Virtual machine scale set` -> `Instances` -> `Public IP address`. For more information on Public IPs for VM NICs, see [this document](https://docs.microsoft.com/azure/virtual-network/virtual-network-network-interface-addresses). 
+
+
+To find a resource's public IP, search for [Virtual Machine Scale Sets](https://portal.azure.com/#blade/HubsExtension/BrowseResourceBlade/resourceType/Microsoft.Compute%2FvirtualMachineScaleSets) -> click on the set name(inside  `MC_resourceGroupName_AKSName_westeurope` group) -> click `Instances` -> click on the instance name -> view `Public IP address`.
+
+To get public IP via API [look here](https://github.com/Azure/azure-libraries-for-net/issues/1185#issuecomment-747919226).
+
+
+For more information on Public IPs for VM NICs, see [this document](https://docs.microsoft.com/azure/virtual-network/virtual-network-network-interface-addresses). 
 
 ## Next Steps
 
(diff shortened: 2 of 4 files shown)


markmandel (commented):
No issues from me. Might want to see if we can find someone who is familiar with AKS to take a gander.


But if we can't, this should be a definite improvement regardless 👍🏻  thanks for the work!

site/content/en/docs/Installation/Creating Cluster/aks.md
```diff
@@ -65,9 +65,9 @@ az network nsg rule create \
...
 
-Nodes in AKS don't get a Public IP by default. To assign a Public IP to a Node, find the Resource Group where the AKS resources are installed on the [portal](https://portal.azure.com) (it should have a name like `MC_resourceGroupName_AKSName_westeurope`). Then, you can follow the instructions [here](https://docs.microsoft.com/en-us/azure/site-recovery/concepts-public-ip-address-with-site-recovery) to create a new Public IP and assign it to the Node/VM. For more information on Public IPs for VM NICs, see [this document](https://docs.microsoft.com/azure/virtual-network/virtual-network-network-interface-addresses). If you are looking for an automated way to create and assign Public IPs for your AKS Nodes, check [this project](https://github.com/dgkanatsios/AksNodePublicIPController).
+Resource Group where the AKS resources are installed on the [portal](https://portal.azure.com) are named like `MC_resourceGroupName_AKSName_westeurope`). Find `Virtual machine scale set` -> `Instances` -> `Public IP address`. For more information on Public IPs for VM NICs, see [this document](https://docs.microsoft.com/azure/virtual-network/virtual-network-network-interface-addresses).
```
eddie-knight, line 70: For clarity, "instances" is only viewable after clicking a particular set within the `Virtual Machine Scale Sets` UI, and the public IP is only viewable after selecting the individual instance.

I suppose that's implied. But if explicitness is desired, the instructions should be more like this:

> To find a resource's public IP, search for `Virtual Machine Scale Sets` -> click on the set name -> click `Instances` -> click on the instance name -> view `Public IP address`

eddie-knight: Also, is the entire first sentence obsolete now? Considering that Azure now has `--enable-node-public-ip`, this section isn't really serving an explicit purpose aside from letting people know how to view the IP from the UI (don't need the resource group name for that).

Last thought... Should we mention here that the IP can be viewed via `kubectl get nodes`, to avoid any confusion leading people around the UI unnecessarily?

dzmitry-lahoda: will retain only one simples way to see IP from UI. so I do not think it is obsolete until https://github.com/Azure/AKS/issues/2318  .

dzmitry-lahoda: i tried to get nodes - azure does not implements that spec, see issues mentioned.

mateuushz: is there a way to get the public IP listed via kubectl get gameservers? right now is picking up the private ip

install/terraform/modules/aks/aks.tf
```diff
@@ -63,12 +63,12 @@ resource "azurerm_kubernetes_cluster" "agones" {
   kubernetes_version = "1.18.14"
```
mateuushz, line 63: `kubernetes_version = "1.20.5"`

1.18.4 deprecated soon.

dzmitry-lahoda: i see, 1.18.* will not help. but as per docs Agones was not tested with higher. I think it should be other PR.

site/content/en/docs/Installation/Creating Cluster/aks.md
dzmitry-lahoda: see linked issues above

eddie-knight: > will retain only one simples way to see IP from UI. so I do not think it is obsolete until [Azure/AKS#2318](https://github.com/Azure/AKS/issues/2318) .

Sorry, didn't mean to say the entire thing is obsolete. But maybe we don't need the first sentence, regarding the resource group?

kaboing:
+1 from me

install/terraform/modules/aks/aks.tf
markmandel: Only 1.18 is supported on Agones right now, so we can't go and change the version, without the process to upgrade:
https://agones.dev/site/docs/installation/

markmandel:
Thanks for the comments all! I'll hold off on approval and merging until I get a 👍🏻 on the changes.

install/terraform/modules/aks/aks.tf
dzmitry-lahoda: afaik MS will prevent me to deploy 1.18 cluster this month. i do not yet have production cluster for k8s. so may be I must hedge that by preallocating:) 
what are plans for 1.19 for agones?

markmandel: > Agones will update its support to n-1 version of what is available across the majority of major cloud providers - GKE, EKS and AKS, while also ensuring that all Cloud providers can support that version.
https://agones.dev/site/docs/installation/

From review, we are ready to move to 1.19.

markmandel:
@eddie-knight you've been most vocal 😄 you happy with these changes?

install/terraform/modules/aks/aks.tf
roberthbailey: I checked at the start of this release cycle and we weren't quite ready. Given how soon the next release is, I was planning to do the update to 1.19 for the following release.

eddie-knight:
Yep, looks great!

markmandel (approved):

WeetA34:
Hello @dzmitry-lahoda, @markmandel 

i have two things to report:
- to get public IP working fine (external-ip properly set)on AKS, kubernetes version must be at least 1.18.19
- game server port are not opened as the network security group is not attached to the subnet

Regards
Stéphane

PR #2110, opened 2021-05-26
Return HTTP 201 on GameServerAllocation

### LINKED ISSUE #2108: Creating a GameServerAllocation returns a 200 Ok instead of a 201 Created
**What happened**: I use the kubernetes c# client to create a GameServerAllocation. I received an exception with the message ""Operation returned an invalid status code 'OK'". 
Using kubectl, I can see that the GameServerAllocation was indeed created. 

**What you expected to happen**: I expect the c# client not to throw an exception. 

**How to reproduce it (as minimally and precisely as possible)**: Create a GameServerAllocation with the official kubernetes c# client, use the `CreateNamespacedCustomObjectWithHttpMessagesAsync` to get the exception. Also, you can use any other client and check the http status code response. 

**Anything else we need to know?**: It seems that the c# client only accepts a `201 Created` status back from Agones, instead a `200 Ok` is sent. It seems that the go client, for instance, is more forgiving for this. 

**Environment**:
- Agones version: 1.14.0
- Kubernetes version (use `kubectl version`): 1.18.14
- Cloud provider or hardware configuration: Azure
- Others:

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The standard for CRD creation is to return a HTTP 201 response. The Go client is more flexible and was happy with a 200 response, but the C# client will error out if anything but a 201 response.

This fixes things so we send the correct HTTP response code when creating a GameServerAllocation.

Closes #2108

N/A

### DIFF
--- a/pkg/gameserverallocations/controller.go
+++ b/pkg/gameserverallocations/controller.go
@@ -133,11 +133,17 @@ func (c *Controller) processAllocationRequest(ctx context.Context, w http.Respon
 	if err != nil {
 		return err
 	}
-	if status, ok := result.(*metav1.Status); ok {
-		w.WriteHeader(int(status.Code))
+	var code int
+	switch obj := result.(type) {
+	case *metav1.Status:
+		code = int(obj.Code)
+	case *allocationv1.GameServerAllocation:
+		code = http.StatusCreated
+	default:
+		code = http.StatusOK
 	}
 
-	err = c.serialisation(r, w, result, scheme.Codecs)
+	err = c.serialisation(r, w, result, code, scheme.Codecs)
 	return err
 }
 
@@ -182,14 +188,18 @@ func (c *Controller) allocationDeserialization(r *http.Request, namespace string
 	return gsa, nil
 }
 
-// serialisation takes a runtime.Object, and serislises it to the ResponseWriter in the requested format
-func (c *Controller) serialisation(r *http.Request, w http.ResponseWriter, obj k8sruntime.Object, codecs serializer.CodecFactory) error {
+// serialisation takes a runtime.Object, and serialise it to the ResponseWriter in the requested format
+func (c *Controller) serialisation(r *http.Request, w http.ResponseWriter, obj k8sruntime.Object, statusCode int, codecs serializer.CodecFactory) error {
 	info, err := apiserver.AcceptedSerializer(r, codecs)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to find serialisation info for %T object", obj)
 	}
 
 	w.Header().Set("Content-Type", info.MediaType)
+	// we have to do this here, so that the content type is set before we send a HTTP status header, as the WriteHeader
+	// call will send data to the client.
+	w.WriteHeader(statusCode)
+
 	err = info.Serializer.Encode(obj, w)
 	return errors.Wrapf(err, "error encoding %T", obj)
 }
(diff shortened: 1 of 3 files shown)


markmandel:
Testing with `kubectl` as well:


```
$ kubectl -v 9 create -f ./examples/simple-game-server/gameserverallocation.yaml -o yaml
....
I0526 20:22:28.749263    1257 request.go:1068] Request Body: {"apiVersion":"allocation.agones.dev/v1","kind":"GameServerAllocation","metadata":{"namespace":"default"},"spec":{"required":{"matchLabels":{"agones.dev/fleet":"simple-game-server"}}}}
I0526 20:22:28.749302    1257 round_trippers.go:424] curl -k -v -XPOST  -H "Content-Type: application/json" -H "User-Agent: kubectl/v1.18.15 (linux/amd64) kubernetes/73dd5c8" -H "Accept: application/json" 'https://34.83.43.189/apis/allocation.agones.dev/v1/namespaces/default/gameserverallocations'
I0526 20:22:28.911903    1257 round_trippers.go:444] POST https://34.83.43.189/apis/allocation.agones.dev/v1/namespaces/default/gameserverallocations 201 Created in 162 milliseconds
```

You can see the 201 on the last line 🥳

pkg/gameserverallocations/controller.go
```diff
@@ -133,11 +133,17 @@ func (c *Controller) processAllocationRequest(ctx context.Context, w http.Respon
...
-	if status, ok := result.(*metav1.Status); ok {
-		w.WriteHeader(int(status.Code))
+	var code int
+	switch obj := result.(type) {
+	case *metav1.Status:
+		code = int(obj.Code)
+	case *allocationv1.GameServerAllocation:
+		code = http.StatusCreated
+	default:
+		code = 200
```
roberthbailey, line 143: It looks like the old behavior was to not set a code in the header at all unless the result was a metav1.Status whereas now we always set a code in the header. Is it better to only set the code when necessary, and not send any code at all for a successful response?

markmandel: There are probably a few things here, which are fun ways in which golang handles headers, that I found along the way.

1. This likely solves a bug that nobody ran into. Which is, when we previously set the `WriteHeader` in the current version, since we're writing the client at this point, it kicks the encoding type to `text/plain` - even though, no matter what we want `application/json` for all things (and we set it later in the code logic, but it will be ignored). This will cause all the k8s clients and `kubectl` to just fail. I found this out the hard way when implementing this fix.
2. If you don't set a HTTP status code header - then by default, Go will make it 200 on first write to the client. So you end up in the same place as you would do normally, but in this case it's now explicit rather than implicit.

Although a good point is - I shouldn't have it as "200" - I should use `http.StatusOK`. I'll make that change now.

roberthbailey: Thanks for the explanation.

pkg/gameserverallocations/controller_test.go
```diff
@@ -103,6 +103,7 @@ func TestControllerAllocator(t *testing.T) {
 			rec := httptest.NewRecorder()
 			err = c.processAllocationRequest(ctx, rec, r, "default")
 			assert.NoError(t, err)
+			assert.Equal(t, http.StatusCreated, rec.Code)
```
roberthbailey, line 106: We shouldn't block this bug fix merging, but it seems like we have a test that covers the other two case statements (using the metav1 status and returning a `http.StatusOK`).

roberthbailey (approved):

PR #2116, opened 2021-06-01
support custom autoscaler resync interval

### LINKED ISSUE #1955: Configurable custom resync period for FleetAutoscaler
I would like to be able to change "resyncPeriod" of Fleet Autoscaler from 30 seconds.

### Testing

I want to shorten the time for Test using Fleet AutoScaler.

### Development environment

GameServer lifecycle is sometimes shorter than production.
In this case, there may not be enough GameServer and errors occur.
This problem can be avoided by inscreasing the buffer size, but this is not desirable due to its cost.

### Production environment

Our service is supposed to have a large number of accesses in an instant.
We want to have an option to shorten the "resyncPeriod" to be ready for spikes.

I want to set the "resyncPeriod" from helm.
I suggest using "SharedInformerOption" and "customResync" mechanism.

### Testing
Wait 30 seconds or more for tests to complete

### Development environment
Inform the development teams that frequent allocation may cause errors

### Production environment 
Adjust the buffer size manually when access spikes are expected

jie-bao:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature

/kind feature

Add a new flag "fas-resync-interval" for customizing fleetAutoScaler resync interval

Closes #1955 

N/A

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -26,6 +26,7 @@ import (
 
 	"agones.dev/agones/pkg"
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
+	autoscalingv1 "agones.dev/agones/pkg/apis/autoscaling/v1"
 	"agones.dev/agones/pkg/client/clientset/versioned"
 	"agones.dev/agones/pkg/client/informers/externalversions"
 	"agones.dev/agones/pkg/fleetautoscalers"
@@ -49,6 +50,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	extclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/resource"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
@@ -77,6 +79,7 @@ const (
 	logLevelFlag                 = "log-level"
 	logSizeLimitMBFlag           = "log-size-limit-mb"
 	kubeconfigFlag               = "kubeconfig"
+	fasResyncIntervalFlag        = "fas-resync-interval"
 	defaultResync                = 30 * time.Second
 )
 
@@ -158,7 +161,9 @@ func main() {
 	wh := webhooks.NewWebHook(httpsServer.Mux)
 	api := apiserver.NewAPIServer(httpsServer.Mux)
 
-	agonesInformerFactory := externalversions.NewSharedInformerFactory(agonesClient, defaultResync)
+	logger.WithField("fasResyncInterval", ctlConf.FasResyncInterval).Info("Setting fleet autoscaler resync interval")
+	customResyncInterval := externalversions.WithCustomResyncConfig(map[v1.Object]time.Duration{&autoscalingv1.FleetAutoscaler{}: ctlConf.FasResyncInterval})
+	agonesInformerFactory := externalversions.NewSharedInformerFactoryWithOptions(agonesClient, defaultResync, customResyncInterval)
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, defaultResync)
 
 	server := &httpServer{}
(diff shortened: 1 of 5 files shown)


markmandel:
I like the general idea of this - but think it would be better if we did this on a per-fleet autoscaler level 🤔 - rather than Agones wide.

Then you could choose the appropriate rate (or maybe even per-event?) as the requirements for your Fleet deployment, rather than entirely Agones wide.

WDYT?

jie-bao:
> I like the general idea of this - but think it would be better if we did this on a per-fleet autoscaler level 🤔 - rather than Agones wide.
> 
> Then you could choose the appropriate rate (or maybe even per-event?) as the requirements for your Fleet deployment, rather than entirely Agones wide.
> 
> WDYT?

I like the per fleet autoscaler idea too. However, it's too hard to implement it.

According to the Kubernetes informer model, we can only assign one re-sync period to one informer. If we want to implement it on a per fleet autoscaler level, we will have to create an informer for each fleet rather than leverage the sharedInformer. It will make the controller overcomplicated. In addition, it's hard to limit the CPU usage caused by fleet autoscaler in Agones controller. Image a user can create 10 fleet and set the re-sync interval to 1s for all of them.

We can take reference to K8S horizontal pod autoscaler. It can set the sync period by the `--horizontal-pod-autoscaler-sync-period` in controller manager. But can't set the sync period for per HorizontalPodAutoscaler resource. I think it's reasonable for keeping the same design.

jie-bao:
> I like the general idea of this - but think it would be better if we did this on a per-fleet autoscaler level 🤔 - rather than Agones wide.
> 
> Then you could choose the appropriate rate (or maybe even per-event?) as the requirements for your Fleet deployment, rather than entirely Agones wide.
> 
> WDYT?

I have another idea that if we want to let FAS re-sync for per fleet, we can involve game server informer into FAS controller. Then each time a game server is added/deleted, we do a re-sync for that fleet. WDYT?

jie-bao:
@googlebot I signed it!

markmandel:
> I like the per fleet autoscaler idea too. However, it's too hard to implement it.

> According to the Kubernetes informer model, we can only assign one re-sync period to one informer. If we want to implement it on a per fleet autoscaler level, we will have to create an informer for each fleet rather than leverage the sharedInformer. It will make the controller overcomplicated. In addition, it's hard to limit the CPU usage caused by fleet autoscaler in Agones controller. Image a user can create 10 fleet and set the re-sync interval to 1s for all of them.

We don't _have_ to use the Informer sync for notification of the 30s. We could keep track of our own polling timers, and with each change to a FAS, make sure they are correct and up to date. That would get us to the same place, and have configurable resync periods.

> I have another idea that if we want to let FAS re-sync for per fleet, we can involve game server informer into FAS controller. Then each time a game server is added/deleted, we do a re-sync for that fleet. WDYT?

I thought I'd also mentioned this (if not, we should file it as a separate bug) - but yes, this is also a good idea as having as a configuration option (probably behind a feature gate) - but we don't need to tie it to the Fleet controller (in fact, I don't think this is a good idea). We can subscribe the FAS controller to Fleet events, and resync any FAS's that match against that specific Fleet, and then resync those FAS's.

I am curious what this would look like in a real world scenario, and how much thrashing of GameServers might occur - but it's a good feature to look into for sure.

jie-bao:
> We don't _have_ to use the Informer sync for notification of the 30s. We could keep track of our own polling timers, and with each change to a FAS, make sure they are correct and up to date. That would get us to the same place, and have configurable resync periods.

Given your idea, I think we can add a new property named like "syncInterval" to FleetAutoscaler resource definition. After FAS controller starts, for each FAS, create a go routine getting fleet status directly from Kube API server? That might be feasible. My only concern is how to protect Kube API server load from too many API queries, since based on this design the API throughput depends on Fleet deployers, not the Agones deployer. However it might not be a big problem in normal cases.

Besides, do you think it's still valuable for this PR? Since kube-controller has similiar implement for horizontal-pod-autoscaler?

markmandel:
> Given your idea, I think we can add a new property named like "syncInterval" to FleetAutoscaler resource definition. 

We should probably write up a comprehensive design on this in #1955 - we want to make sure it's future proof, rather than a one off config option. I'm thinking something like (off the top of my head):

```
syncPolicy:
  type: interval
  interval:
    duration: 30s
```

(Similar to what we do with the [FleetAutoscaler](https://agones.dev/site/docs/reference/fleetautoscaler/))

Which can then be expanded when/if we want to do an event based system.

> After FAS controller starts, for each FAS, create a go routine getting fleet status directly from Kube API server? That might be feasible. My only concern is how to protect Kube API server load from too many API queries, since based on this design the API throughput depends on Fleet deployers, not the Agones deployer. However it might not be a big problem in normal cases.

The client side Lister and Informer are client side caches - so doing a loop through all Fleet Autoscalers to reconcile them all on each FAS change doesn't put any extra load on the K8s API (which is why we have a local cache!). So no huge issue there.

> Besides, do you think it's still valuable for this PR? Since kube-controller has similiar implement for horizontal-pod-autoscaler?

Not quite sure what "it" is referencing here -- but I hope I helped answer the questions above?

jie-bao:
> We should probably write up a comprehensive design on this in #1955 - we want to make sure it's future proof, rather than a one off config option.

OK, thanks for your thoughts, we can move to that issue for further discussion.

markmandel:
Any objection to closing this PR for now, as we work on the design in the ticket?

jie-bao:
Close this PR and go further discussion in #1955

PR #2120, opened 2021-06-01
Change the way we generate new version numbers for nodejs

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

**What this PR does / Why we need it**: Address feedback from #2080

Closes #

I've moved the suggested edit from #2080 here so that we can continue the discussion about how to generate version numbers for the nodejs package.

### DIFF
--- a/docs/governance/templates/release_issue.md
+++ b/docs/governance/templates/release_issue.md
@@ -35,7 +35,7 @@ and copy it into a release issue. Fill in relevant values, found inside {}
 - [ ] Ensure the [helm `tag` value][values] is correct (should be {version} if a full release, {version}-rc if release candidate)
 - [ ] Ensure the [helm `Chart` version values][chart] are correct (should be {version} if a full release, {version}-rc if release candidate)
 - [ ] Update SDK Package Versions
-    - [ ] Ensure the [`sdks/nodejs/package.json`][nodejs] version is correct (should be {version} if a full release, {version}-rc if release candidate)
+    - [ ] Update the package version in [`sdks/nodejs/package.json`][package.json] and [`sdks/nodejs/package-lock.json`][package-lock.json] by running `npm version {version}` if a full release or `npm version {version}-rc` if release candidate
     - [ ] Ensure the [`sdks/csharp/sdk/AgonesSDK.nuspec` and `sdks/csharp/sdk/csharp-sdk.csproj`][csharp] versions are correct (should be {version} if a full release, {version}-rc if release candidate)
 - [ ] Run `make gen-install`
 - [ ] Run `make test-examples-on-gcr` to ensure all example images exist on gcr.io/agones-images-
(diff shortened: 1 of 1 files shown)


docs/governance/templates/release_issue.md
```diff
@@ -77,7 +77,7 @@ and copy it into a release issue. Fill in relevant values, found inside {}
 - [ ] If full release, then increment the `base_version` in [`build/Makefile`][build-makefile]
 - [ ] If full release move [helm `tag` value][values] is set to {version}+1-dev
 - [ ] If full release move the [helm `Chart` version values][chart] is to {version}+1-dev
-- [ ] If full release move the [`sdks/nodejs/package.json`][nodejs] to {version}+1-dev
+- [ ] If full release, change to the `sdks/nodejs` directory and run the command `npm version preminor --preid=dev` to update the package version
```
roberthbailey, line 80: Copying comments over:

@steven-supersolid:

I think this is the correct step but we should replace with the following command
```
cd sdks/nodejs
npm version preminor --preid=dev
```
I.e. change to:
```
- [ ] If full release, change to the `sdks/nodejs` directory and run the command `npm version preminor --preid=dev` to update the package version
```

@roberthbailey:

That formats the version slightly differently, e.g.

`version": "1.16.0-dev.0"` instead of `version": "1.16.0-dev"`

@steven-supersolid:

Removing the trailing `.0` does not seem to be supported by npm and I think is due to how semver is handled internally (and by the semver package)
https://docs.npmjs.com/cli/v6/commands/npm-version
https://github.com/npm/node-semver

I think it is this way to support publishing multiple dev versions prior to a release, because it is not possible to re-publish an existing version. Perhaps idiomatic Node.js

An alternative would be specify the version explicitly rather than rely on the increment feature of npm - may be better for this project?
```
npm version 1.16.0-dev
```

That seems clearer

I'm hoping we can avoid the situation where we manually edit `package.json` as we must then run `npm install` to update `package-lock.json`

steven-supersolid:
Given that we want to use project specific versions, e.g. `-dev`, `-rc` I think it's simpler to use the explicit version command rather than the version increment. e.g. `npm version 1.16.0-dev`. This also gives us the same version system as currently in place without the suffix `.0`

docs/governance/templates/release_issue.md
roberthbailey: I've updated the checklist using the npm version command, since that seems to do what we want (although, it means we have to explicitly specify the version number instead of letting the tooling increment it for us). But this is better than what we do today, which is editing the package.json file by hand, and not updating package-lock.json.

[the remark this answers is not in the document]
roberthbailey: I've also updated the instruction earlier in the checklist so we get the correct version in package-lock.json for the release candidate.

steven-supersolid (approved):

ilkercelikyilmaz:
LGTM

roberthbailey:
Adding @ilkercelikyilmaz for approval since Mark is out today.

roberthbailey:
Now that the release is finished, I'm going to update these instructions so we use them on the next release.

ilkercelikyilmaz (approved):
LGTM

PR #2121, opened 2021-06-02
Release 1.15.0-rc

sawagh:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind release

This includes all of the changes for the 1.15.0 release candidate.

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,48 @@
 # Changelog
 
+## [v1.15.0-rc](https://github.com/googleforgames/agones/tree/v1.15.0-rc) (2021-06-01)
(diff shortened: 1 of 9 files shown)


roberthbailey (approved):

steven-supersolid (approved):

steven-supersolid:
Ah I just realised when we edit `package.json` here we should instead be running `npm version 1.15.0-rc` (or equivalent) so it syncs `package-lock.json`. This does mean can't do the release with only a text editor though #2120

PR #2139, opened 2021-06-18
add copy to clipboard function to code on website

### LINKED ISSUE #2096: Add "copy to clipboard" buttons to example commands on the website
It's easy to make a mistake when copy/pasting from the agones documentation site. One example is when setting up the allocator service: https://agones.dev/site/docs/advanced/allocator-service/

This could be solved by adding a button that allows the user to copy to clipboard, rather than relying on them selecting the command themselves. If we do this, we should look at other pages that would also benefit from having commands be easily copied (although we may need to create other bugs since the surface area is pretty large). 

Leave as-is, maybe simplify the service setup so that it takes fewer steps?

cindy52:
**What type of PR is this?**
 /kind feature

Add "copy to clipboard" buttons to example commands on the website

Closes #2096 

NA

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -45,7 +45,7 @@ RUN go get -u golang.org/x/tools/cmd/goimports
 RUN gcloud components update && gcloud components install app-engine-go
 
 # the kubernetes version for the file
-ENV KUBERNETES_VER 1.18.15
+ENV KUBERNETES_VER 1.19.12
 
 # overwrite kubectl as we want a specific version
 RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBERNETES_VER}/bin/linux/amd64/kubectl && \
--- a/build/e2e-image/Dockerfile
+++ b/build/e2e-image/Dockerfile
@@ -15,7 +15,7 @@ RUN wget -q https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
 ENV PATH /usr/local/go/bin:/go/bin:$PATH
 
 # install kubectl without gcloud as we want the last version
-ENV KUBECTL_VER 1.18.15
+ENV KUBECTL_VER 1.19.12
 RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VER}/bin/linux/amd64/kubectl && \
     chmod go+rx ./kubectl && \
     mv ./kubectl /usr/local/bin/kubectl
--- a/build/terraform/prow/module.tf
+++ b/build/terraform/prow/module.tf
@@ -25,7 +25,7 @@ resource "google_container_cluster" "prow-build-cluster" {
   project            = var.project
   location           = "us-west1-c"
   description        = "Prow cluster to run tests for Agones"
-  min_master_version = "1.18"
+  min_master_version = "1.19"
   initial_node_count = 8
   node_config {
     machine_type = "e2-standard-4"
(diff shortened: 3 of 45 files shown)


roberthbailey:
Thanks for the pull request! It seems like this wasn't that hard to add with the new version of docsy. :)

I'm looking at the website preview and I'm kinda disappointed in the way that prismjs renders the copy button (obviously not your fault). 

I've been looking for examples of any other sites using prism for this functionality but I'm having trouble finding any. I've found examples like https://tutsplus.github.io/clipboard/ and https://chrissainty.com/copy-to-clipboard-in-blazor/ and https://medium.com/@nrrrdcore/tiny-design-bite-transitioning-tooltip-text-with-pseudo-elements-hover-states-82fbe00e8c33 which to me look sooo much nicer. I wonder if we should check with the docsy authors to see if they have suggestions on how we might be able to style this more nicely.

markmandel:
Oh yeah, that's uh... super ugly 😬 

image

Weirdly also, I see the copy button on "Install Agones using YAML" but not on any of the Quickstarts

image

Not quite sure why that is 🤔

roberthbailey:
Here's another example of a "nicer" looking copy button: https://kind.sigs.k8s.io/docs/user/quick-start/#installation

cindy52:
I think it's also related to our current code block style. I found an example here https://github.com/jenkins-x/jx-docs with different style and the copy button looks much nicer.

cindy52:
Took a look at the kind.sigs.k8s site code, looks like they defined their own copyText function https://github.com/kubernetes-sigs/kind/blob/b6bc112522651d98c81823df56b7afa511459a3b/site/assets/js/inline.js#L67 instead of using prism.

cindy52:
> Here's another example of a "nicer" looking copy button: https://kind.sigs.k8s.io/docs/user/quick-start/#installation
 I took a look at the kind.sigs.k8s site code, looks like they defined their own copyText function https://github.com/kubernetes-sigs/kind/blob/b6bc112522651d98c81823df56b7afa511459a3b/site/assets/js/inline.js#L67 instead of using prism.

cindy52:
> Oh yeah, that's uh... super ugly 😬
> 
> image
> 
> Weirdly also, I see the copy button on "Install Agones using YAML" but not on any of the Quickstarts
> 
> image
> 
> Not quite sure why that is 🤔

I think the copy button only show up when the code is in the block of '''bash some_code'''. All the code in quickstart is only one line thus in a block of  '''some_code '''.

markmandel:
> I think the copy button only show up when the code is in the block of '''bash some_code'''. All the code in quickstart is only one line thus in a block of '''some_code '''.

Aha! Sounds like we will want to review that then, so we can make sure to add copy buttons in relevant places around the site.

cindy52:
@roberthbailey , @markmandel , please take another look. I've adjusted the copy button a little bit. It looks way better this time.

markmandel:
That looks better! 👍🏻 Nice!

Do we want to use this PR to do a review of all the places we have commands that are copy-pastable?

cindy52:
> That looks better! 👍🏻 Nice!
> 
> Do we want to use this PR to do a review of all the places we have commands that are copy-pastable?

Do we want to add copy button to all the commands (including just one line) or just bash scripts (multiple lines)?

markmandel:
👋🏻 what happened with this PR? Looked like it was on a good path? 😞

roberthbailey:
See https://github.com/googleforgames/agones/pull/2149

PR #2141, opened 2021-06-18
Fix shutdown problems in ping application.

s-shin:
**What type of PR is this?**

/kind bug

* Make sure HTTP server gracefully shuts down.
    * Since `ListenAndServe()` always returns a non-nil error, the application exited by `logger.Fatal()` before graceful shutdown process is completed.
* Remove unnecessary error log.
    * In shutdown process of `udpServer`, an error log was always emitted, but it's not actually an error in intentional shutdown triggered by signal.

None maybe.

N/A

### DIFF
--- a/cmd/ping/main.go
+++ b/cmd/ping/main.go
@@ -91,7 +91,9 @@ func serveHTTP(ctlConf config, h healthcheck.Handler) func() {
 
 	go func() {
 		logger.Info("starting HTTP Server...")
-		logger.WithError(srv.ListenAndServe()).Fatal("could not start HTTP server")
+		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
+			logger.WithError(err).Fatal("could not start HTTP server")
+		}
 	}()
 
 	return func() {
@@ -101,6 +103,7 @@ func serveHTTP(ctlConf config, h healthcheck.Handler) func() {
 		if err := srv.Shutdown(ctx); err != nil {
 			logger.WithError(err).Fatal("could not shut down HTTP server")
 		}
+		logger.Info("HTTP server was gracefully shut down")
 	}
 }
 
--- a/cmd/ping/main.go
+++ b/cmd/ping/main.go
@@ -43,11 +43,11 @@ var (
 func main() {
 	ctlConf := parseEnvFlags()
 	if err := ctlConf.validate(); err != nil {
-		logger.WithError(err).Fatal("could not create controller from environment or flags")
+		logger.WithError(err).Fatal("Could not create controller from environment or flags")
 	}
 
 	logger.WithField("version", pkg.Version).WithField("featureGates", runtime.EncodeFeatures()).
-		WithField("ctlConf", ctlConf).Info("starting ping...")
+		WithField("ctlConf", ctlConf).Info("Starting ping...")
 
 	ctx := signals.NewSigKillContext()
 
@@ -61,7 +61,7 @@ func main() {
 	defer cancel()
 
 	<-ctx.Done()
-	logger.Info("shutting down...")
+	logger.Info("Shutting down...")
 }
 
 func serveUDP(ctx context.Context, ctlConf config) *udpServer {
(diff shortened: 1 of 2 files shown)


cmd/ping/main.go
```diff
@@ -91,7 +91,9 @@ func serveHTTP(ctlConf config, h healthcheck.Handler) func() {
 
 	go func() {
 		logger.Info("starting HTTP Server...")
-		logger.WithError(srv.ListenAndServe()).Fatal("could not start HTTP server")
+		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
+			logger.WithError(err).Fatal("could not start HTTP server")
```
markmandel, line 95:
```suggestion
			logger.WithError(err).Fatal("Could not start HTTP server")
```

We are super inconsistent with this 😬 , but log statements should be start with acapitall letter. I know you are copy pasting from the previous log message, but I'd like to take the opportunity to fix this while you are in there 😄 

Otherwise, this is great. Happy to merge with this small nit change. Thanks!

cmd/ping/udp.go
```diff
@@ -137,17 +141,17 @@ func (u *udpServer) rateLimitedEchoResponse(b []byte, sender net.Addr) {
 	if vis.limit.Allow() {
 		b = bytes.TrimRight(b, "\x00")
 		if _, err := u.conn.WriteTo(b, sender); err != nil {
-			u.logger.WithError(err).Error("error sending returning udp packet")
+			u.logger.WithError(err).Error("Error sending returning udp packet")
 		}
 	} else {
-		logger.WithField("addr", sender.String()).Warn("rate limited. No response sent.")
+		logger.WithField("addr", sender.String()).Warn("Rate limited. No response sent")
```
s-shin, line 147: Should the last period be ommited, right?

roberthbailey: That certainly matches the style in this file (although no where else do I see multiple sentences, so maybe "Rate limited, no response set" would be more consistent).

markmandel:
```
Copying file://agonessdk-server-1.16.0-8f05962.zip [Content-Type=application/zip]...
ResumableUploadAbortException: Upload complete with 1536 additional bytes left in stream; this can happen if a file changes size while being uploaded
CommandException: 1 file/object could not be transferred.
/ [0/1 files][    0.0 B/125.4 MiB]   0% Done                                    
```

Hmnn, that happened again 🤔

cmd/ping/udp.go
s-shin: I found two examples in the current codebase.

* https://github.com/googleforgames/agones/blob/v1.15.0/cmd/allocator/main.go#L168
    * Uses `;` (the file itself looks almost old style though.)
* https://github.com/googleforgames/agones/blob/v1.15.0/cmd/controller/main.go#L118
    * Uses `.`.

I think the whole consistency would be better to be fixed in another PR, how about you?

cmd/ping/main.go
```diff
@@ -85,22 +85,25 @@ func serveHTTP(ctlConf config, h healthcheck.Handler) func() {
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		if _, err := w.Write([]byte(ctlConf.HTTPResponse)); err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
-			logger.WithError(err).Error("error responding to http request")
+			logger.WithError(err).Error("Error responding to http request")
```
aLekSer, line 88: Please revert all capitalisation back, it was done intentional here
https://github.com/golang/go/wiki/CodeReviewComments#error-strings
```suggestion
			logger.WithError(err).Error("error responding to http request")
```

roberthbailey:
This failed due to upgrading the e2e cluster to 1.19: `Error from server (NotFound): statefulsets.apps "consul" not found`

It needs to be updated to the main branch, which will cause the tests to re-run (with the proper stateful set name).

cmd/ping/udp.go
markmandel: Yeah, this format doesn't bother me much. I'm more concerned with initial capitalisation. 

@roberthbailey any major objections to approving and merging? Otherwise I'm happy to merge on my end.

markmandel (approved):

cmd/ping/udp.go
roberthbailey: This is an improvement over what we had before, so I'm happy with merging it and working more towards consistency in a follow up PR.

PR #2143, opened 2021-06-21
Feature gates for advanced Allocation filtering

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Adding the feature gates for the upcoming features.

Work on #1239

Specifically didn't add the gates to either the Makefile for the cloudbuild.yaml until the full feature has all it's functionality in place and it will pass tests.

### DIFF
--- a/pkg/util/runtime/features.go
+++ b/pkg/util/runtime/features.go
@@ -43,6 +43,13 @@ const (
 
 	// NodeExternalDNS is a feature flag to enable/disable node ExternalDNS and InternalDNS use as GameServer address
 	NodeExternalDNS Feature = "NodeExternalDNS"
+
+	// FeatureStateAllocationFilter is a feature flag that enables state filtering on Allocation.
+	FeatureStateAllocationFilter Feature = "StateAllocationFilter"
+
+	// FeaturePlayerAllocationFilter is a feature flag that enables the ability for Allocations to filter based on
+	// player capacity.
+	FeaturePlayerAllocationFilter Feature = "PlayerAllocationFilter"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


markmandel:
```
Already have image (with digest): gcr.io/cloud-builders/gsutil
Copying file://agonessdk-server-1.16.0-e7d5bb6.zip [Content-Type=application/zip]...
ResumableUploadAbortException: Upload complete with 1536 additional bytes left in stream; this can happen if a file changes size while being uploaded
CommandException: 1 file/object could not be transferred.
/ [0/1 files][    0.0 B/125.4 MiB]   0% Done                                    
```

That's a new one!

pkg/util/runtime/features.go
```diff
@@ -43,18 +43,27 @@ const (
...
 	// featureDefaults is a map of all Feature Gates that are
 	// operational in Agones, and what their default configuration is.
 	// alpha features are disabled.
 	featureDefaults = map[Feature]bool{
-		FeatureExample:               true,
-		FeaturePlayerTracking:        false,
-		FeatureSDKWatchSendOnExecute: true,
-		FeatureRollingUpdateOnReady:  true,
-		NodeExternalDNS:              false,
+		FeatureExample:                true,
```
roberthbailey, line 60: I wonder if we should keep these sorted somehow (alphabetical and/or based on true/false). Since they end up in a map the order they are inserted doesn't matter, but it may be useful to humans to have them in a logical ordering.

roberthbailey (approved):

pkg/util/runtime/features.go
markmandel: 🤔 they are currently in order of initial implementation (I think?).

I don't have strong opinions here tbh. Happy to take advice?

PR #2148, opened 2021-06-21
Refactor ReadyGameServerCache to AllocationCache

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

A bit of a chunky PR, but it does several things that were tied together.

1. Since the cache will need to track both Ready and Allocated GameServers, renamed everything to be "Allocation" related, rather than just be about Ready GameServers
2. Implement functionality to be able to cache either Ready (for the stable installation), or Allocated + Ready Gameservers for when the Feature Flag of FeatureStateAllocationFilter is enabled.
3. There were some functionality and tests in some strange places, so took the opportunity to clean that up, since I had to refactor a bunch of stuff anyway.
4. Made sure tests would continue to pass in this interim state.

Work on #1239

The design on #1239 has it adding an annotation to the GameServer when allocating. I didn't put this behind the feature branch, because I didn't think it changed anything or was a risk. But please let me know if you think it should be behind the feature branch.


pkg/gameserverallocations/allocation_cache_test.go
```diff
@@ -0,0 +1,343 @@
...
+			// new feature flags
+			if v.features != "" {
+				require.NoError(t, runtime.ParseFeatures(v.features))
+			}
+
+			cache, m := newFakeAllocationCache()
+
+			gsList := v.list
+
+			m.AgonesClient.AddReactor("list", "gameservers", func(action k8stesting.Action) (bool, k8sruntime.Object, error) {
```
roberthbailey, line 144: I noticed that these functions don't have a case where they return a non-nil error. Do we have test coverage for what happens when the k8s calls are failing?

pkg/gameserverallocations/find_test.go
```diff
@@ -155,29 +160,38 @@ func TestFindGameServerForAllocationPacked(t *testing.T) {
...
 			},
+			features: fmt.Sprintf("%s=false&%s=false", runtime.FeaturePlayerAllocationFilter, runtime.FeatureStateAllocationFilter),
 		},
 	}
 
 	for k, v := range fixtures {
 		t.Run(k, func(t *testing.T) {
+			runtime.FeatureTestMutex.Lock()
+			defer runtime.FeatureTestMutex.Unlock()
+			// deliberately not resetting the Feature state, to catch any possible unknown regressions with the
```
roberthbailey, line 171: This doesn't make sense to me, since the tests are setting the feature flags to false, which should be exercising the old behavior.

pkg/gameserverallocations/allocator_test.go
```diff
@@ -0,0 +1,76 @@
...
+	assert.Equal(t, agonesv1.GameServerStateAllocated, gs.Status.State)
+	assert.NotNil(t, gs.ObjectMeta.Annotations["agones.dev/last-allocated"])
+
+	gs, err = allocator.applyAllocationToGameServer(ctx, allocationv1.MetaPatch{Labels: map[string]string{"foo": "bar"}}, &agonesv1.GameServer{})
+	assert.NoError(t, err)
+	assert.Equal(t, agonesv1.GameServerStateAllocated, gs.Status.State)
+	assert.Equal(t, "bar", gs.ObjectMeta.Labels["foo"])
+	assert.NotNil(t, gs.ObjectMeta.Annotations["agones.dev/last-allocated"])
+
+	gs, err = allocator.applyAllocationToGameServer(ctx, allocationv1.MetaPatch{Annotations: map[string]string{"foo": "bar"}}, &agonesv1.GameServer{})
```
roberthbailey, line 61: What does it test to run this call again with the same parameters?

pkg/gameserverallocations/controller_test.go
```diff
@@ -60,6 +60,11 @@ const (
 func TestControllerAllocator(t *testing.T) {
 	t.Parallel()
 
+	// TODO:(markmandel) remove once GameServerSelector is in place, since right now enabling the feature flag causes flaky tests
```
roberthbailey, line 63: This comment is a bit scary. Does it make the tests flaky because they sometime run with the feature flag on and sometimes with it off?

pkg/gameserverallocations/allocation_cache.go
```diff
@@ -196,53 +217,38 @@ func (c *ReadyGameServerCache) ListSortedReadyGameServers() []*agonesv1.GameServ
...
+func (c *AllocationCache) SyncGameServers(ctx context.Context, key string) error {
+	c.loggerForGameServerKey(key).Debug("Refreshing Allocation Gameserver cache")
 
-	return c.syncReadyGSServerCache()
+	return c.syncCache()
 }
 
-// syncReadyGSServerCache syncs the gameserver cache and updates the local cache for any changes.
-func (c *ReadyGameServerCache) syncReadyGSServerCache() error {
+// syncCache syncs the gameserver cache and updates the local cache for any changes.
```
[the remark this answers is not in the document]
markmandel: We don't for two reasons:
1. the underlying `gameServerCacheEntry` has appropriate locking in place for individual records.2
2. The workerqueue is only running a single routine, so this method only ever gets called a single time per invocation.


(Also, all this code has been running in production for ages with no reported issues)

pkg/gameserverallocations/allocation_cache_test.go
```diff
@@ -0,0 +1,343 @@
...
+			list:     []agonesv1.GameServer{gs1, gs2, gs3},
+			features: fmt.Sprintf("%s=true", runtime.FeatureStateAllocationFilter),
+			test: func(t *testing.T, list []*agonesv1.GameServer) {
+				require.Len(t, list, 3)
+				assert.Equal(t, "gs3", list[0].ObjectMeta.Name)
+				assert.Equal(t, "gs1", list[1].ObjectMeta.Name)
+				assert.Equal(t, "gs2", list[2].ObjectMeta.Name)
+			},
+		},
+		"nil player status (PlayerAllocationFilter)": {
```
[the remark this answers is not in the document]
markmandel: Good catch! Fixed.

pkg/gameserverallocations/allocation_cache_test.go
```diff
@@ -0,0 +1,343 @@
...
+				}
+			},
+		},
+	}
+
+	for k, v := range fixtures {
+		k := k
+		v := v
+		t.Run(k, func(t *testing.T) {
+			// deliberately not resetting the Feature state, to catch any possible unknown regressions with the
```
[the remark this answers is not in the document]
markmandel: Yes (admittedly this is a movement of an old test, as is much of this file) - since we don't know what features will be enabled/disabled across the whole set of features, we rely on go's random ordering of tests to find issues with combinations of feature flags over time.

Testing for every combination of feature flags everywhere for past and future flags is a combinatorial explosion, and probably not worth the effort (unless you have a clever way of doing this that I can't think of - in which case I'm all ears!)

pkg/gameserverallocations/allocation_cache_test.go
```diff
@@ -0,0 +1,343 @@
...
+	gs6 := agonesv1.GameServer{ObjectMeta: metav1.ObjectMeta{Name: "gs6", Namespace: defaultNs, UID: "6"},
+		Status: agonesv1.GameServerStatus{NodeName: "node1", State: agonesv1.GameServerStateReady, Players: &agonesv1.PlayerStatus{
+			Count:    2,
+			Capacity: 10,
+		}},
+	}
+
+	fixtures := map[string]struct {
+		list     []agonesv1.GameServer
+		test     func(*testing.T, []*agonesv1.GameServer)
```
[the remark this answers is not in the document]
markmandel: Looking through - the issue here is, in quite a few cases it can be an "either a or b solution is fine for the first two eleements, but the last element has to be c" because the ordering could go either way with the sorting - so having a closure for that makes things easier.

So some tests has things like `assert.Contains(t, []string{"gs2", "gs4"}, list[0].ObjectMeta.Name)` 

Reviewing it though, `testify` has a better way of testing these scenarios:
https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch

So I'll make that change, so it's more obvious what's going on.

pkg/gameserverallocations/allocation_cache_test.go
markmandel: 🤔 I don't think we do very often, just because it ties into the workqueue, and the workerqueue retrying on error is based on the K8s libraries, which have their own tests and is pretty rock solid.

Any suggestion here though, I can see the point, just not quite sure how to tackle it effectively?

It's possible one could argue this would get picked up in e2e tests, which is a more dynamic environment, that does have errors that fire just because of the nature of k8s?

pkg/gameserverallocations/allocator.go
```diff
@@ -567,6 +570,34 @@ func (c *Allocator) allocationUpdateWorkers(ctx context.Context, workerCount int
...
+	if gs.ObjectMeta.Annotations == nil {
+		gs.ObjectMeta.Annotations = make(map[string]string, len(fam.Annotations))
+	}
+	// apply annotations patch
+	for key, value := range fam.Annotations {
+		gs.ObjectMeta.Annotations[key] = value
+	}
+
+	// add last allocated, so it always gets updated, even if it is already Allocated
+	gs.ObjectMeta.Annotations["agones.dev/last-allocated"] = time.Now().String()
```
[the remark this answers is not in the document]
markmandel: Format of string is:  "2006-01-02 15:04:05.999999999 -0700 MST", which includes the timezone, so I think we should be good there for readability.

For the sake of the purposes we need it for allocation, we actually only care that the value changes, so that the object gets updated within K8s, without a field that has changed, if we update the `GameServer` with no differences from what is stored in etcd, it becomes a noop, which means it can't be watched for by the SDK or other services.

pkg/gameserverallocations/allocator_test.go
markmandel: You know, I have no idea 😄 removed!

pkg/gameserverallocations/controller_test.go
markmandel: That is understandably scary on review 😃 - let me adjust it so it is less scary.

TL;DR: once all the code is in place the test will pass with any feature flag enabled/disabled - but not yet.

pkg/gameserverallocations/find_test.go
markmandel: Ah yep, I see how this would look weird - apologies.

I have more tests coming for this test as the this feature continues to grow, which use different feature flags, so I just copied it across. Happy to scrap the individual setting of feature flags these features to false for overall, and then adjust it as needed down the line.

That being said, reviewing all the tests I have, and plan to have for this - I never don't set a feature flag, so I'll make it required for this unit test. Let me know what you think.

markmandel:
Review updates in place, PTAL!

pkg/gameserverallocations/allocation_cache_test.go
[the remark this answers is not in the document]
roberthbailey: I don't think we need to test every combination everywhere, but it places where we have two explicit code paths and it's easy to add coverage (by adding a new table driven test entry) it seems like not too much work to ensure that all test runs do both old and new.

pkg/gameserverallocations/allocation_cache_test.go
roberthbailey: That's what I was wondering - if this function returned an error once or twice and then the normal result would the test actually be any different? It sounds like no because the k8s libraries would just retry and cover up the errors with retries.

pkg/gameserverallocations/allocator.go
```diff
@@ -567,6 +570,34 @@ func (c *Allocator) allocationUpdateWorkers(ctx context.Context, workerCount int
...
+			gs.ObjectMeta.Labels[key] = value
+		}
+	}
+
+	if gs.ObjectMeta.Annotations == nil {
+		gs.ObjectMeta.Annotations = make(map[string]string, len(fam.Annotations))
+	}
+	// apply annotations patch
+	for key, value := range fam.Annotations {
+		gs.ObjectMeta.Annotations[key] = value
```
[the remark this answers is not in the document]
roberthbailey: Sounds like it's not worth it unless it becomes an issue.

pkg/gameserverallocations/allocator.go
```diff
@@ -567,6 +570,34 @@ func (c *Allocator) allocationUpdateWorkers(ctx context.Context, workerCount int
 	return updateQueue
 }
 
+// applyAllocationToGameServer patches the inputted GameServer with the allocation metadata changes, and updates it to the Allocated State.
+// Returns the updated GameServer.
+func (c *Allocator) applyAllocationToGameServer(ctx context.Context, fam allocationv1.MetaPatch, gs *agonesv1.GameServer) (*agonesv1.GameServer, error) {
+	// patch ObjectMeta labels
+	if fam.Labels != nil {
```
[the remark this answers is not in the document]
roberthbailey: `mp` would make the most sense to me.

roberthbailey (approved):

PR #2166, opened 2021-07-01
GSA: Switch LabelSelector to GameServerSelector

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This embeds the LabelSelector into a GameServerSelector struct, along with the new player and state selector features we need for advanced allocation, and utilising it in the `preferred` and `required` sections of a `GameServerAllocation`.

This means the CRD doesn't change, but the Go api surface does.

This also moves the logic of validation and selection matching into the GameServerSelector so it can be more easily reused.

GameServerAllocations are used pretty much everywhere, so a good chunk of the PR is making that api change.

Work on #1239

Sorry for the chunky PR.


roberthbailey:
This is a large PR... I'll try to get to it before the weekend.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -82,6 +82,137 @@ type GameServerAllocationSpec struct {
...
+	}
+
+	return true
+}
+
+// Validate validates that the selection fields have valid values
+func (s *GameServerSelector) Validate(field string, causes []metav1.StatusCause) ([]metav1.StatusCause, bool) {
+	valid := true
+
+	_, err := metav1.LabelSelectorAsSelector(&s.LabelSelector)
```
markmandel, line 163: Just highlighting this change: We end up calling `LabelSelectorAsSelector` twice. Once here in `Validate()`, where we check the error and then again in `Matches()`, where we don't, since we did it before.

I think it's worth it because it creates a much cleaner API surface, but let me know your thoughts.

pkg/allocation/converters/converter.go
```diff
@@ -64,7 +64,7 @@ func ConvertAllocationRequestToGSA(in *pb.AllocationRequest) *allocationv1.GameS
 	}
 
 	if ls := convertLabelSelectorToInternalLabelSelector(in.GetRequiredGameServerSelector()); ls != nil {
-		gsa.Spec.Required = *ls
+		gsa.Spec.Required = allocationv1.GameServerSelector{LabelSelector: *ls}
```
markmandel, line 67: This code will get updates once again once the proto changes, but this allows current functionality to continue working as expected.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -144,5 +265,10 @@ func (gsa *GameServerAllocation) Validate() ([]metav1.StatusCause, bool) {
 			Message: fmt.Sprintf("Invalid value: %s, value must be either Packed or Distributed", gsa.Spec.Scheduling)})
 	}
 
+	causes, _ = gsa.Spec.Required.Validate("spec.required", causes)
```
roberthbailey, line 268: This will overwrite causes if it's non empty. Both here and below you need to append to causes (like is done on line 263). 

(update: re-reading the validate function, I can see why this works, but it's really confusing to read. I think it's simpler to have validate not take causes and have this function do the appending).

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -82,6 +82,137 @@ type GameServerAllocationSpec struct {
...
+}
+
+// Validate validates that the selection fields have valid values
+func (s *GameServerSelector) Validate(field string, causes []metav1.StatusCause) ([]metav1.StatusCause, bool) {
+	valid := true
+
+	_, err := metav1.LabelSelectorAsSelector(&s.LabelSelector)
+	if err != nil {
+		valid = false
+		causes = append(causes, metav1.StatusCause{
```
roberthbailey, line 166: I find it a bit strange that the function takes in a slice of causes, appends to it, and then also returns it (all using the same name). Why does the function need to be given causes? Shouldn't validate only _output_ causes?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -82,6 +82,137 @@ type GameServerAllocationSpec struct {
...
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: "GameServerState value can only be Allocated or Ready",
+				Field:   field,
+			})
+		}
+	}
+
+	if runtime.FeatureEnabled(runtime.FeaturePlayerAllocationFilter) && s.Players != nil {
+		if s.Players.MinAvailable < 0 {
```
roberthbailey, line 185: Is there any case where we should check for the upper bound? Or any valid int value is acceptable?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -82,6 +82,137 @@ type GameServerAllocationSpec struct {
...
+		if s.Players == nil {
+			s.Players = &PlayerSelector{}
+		}
+	}
+}
+
+// Matches checks to see if a GameServer matches a given GameServerSelector's criteria
+func (s *GameServerSelector) Matches(gs *agonesv1.GameServer) bool {
+
+	// Assume at this point, this has already been run through Validate(), and it can be converted.
```
roberthbailey, line 124: Given that this is an exported function, we are trusting that callers are doing the right thing rather than being defensive and checking the error here. 

WDYT about having an error here cause a panic (instead of returning an error)? Then we make it super obvious that things are broken without polluting the method signature, and we don't have code that ignores errors from functions that it calls.

pkg/allocation/converters/converter_test.go
```diff
@@ -88,22 +88,25 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
...
+								MatchLabels: map[string]string{
+									"e": "f",
+								},
+							}},
+						{
+							LabelSelector: metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									"g": "h",
+								},
+							}},
```
roberthbailey, line 109: What about adding conversion tests for the new fields?

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: I agree! I will make this change! Same as comment below 👍🏻

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: Done!

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: Not that I can think of. The extra thing is `if s.Players.MinAvailable > s.Players.MaxAvailable` to make sure min is not greater than max.

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: I like this idea a lot! Implementing!

pkg/allocation/converters/converter_test.go
markmandel: Coming soon to a theatre near you! But once the allocation proto changes are in, because there's nothing to convert things to otherwise.

markmandel:
Updated. I'm 50/50 on whether this should be included in this RC or not 🤔 

Do we give some end users some pain now in preparation, or wait until the feature is ready to go?

pkg/apis/allocation/v1/gameserverallocation.go
roberthbailey: lgtm

roberthbailey (approved):

PR #2170, opened 2021-07-03
Used array of FStringFormatArg to process FString::Format to fix erro…

### LINKED ISSUE #2169: Errors in Unreal Engine SDK BuildAgonesRequest
**What happened**: Using UnrealEngine SDK -> Tried calling UAgonesComponent::SetCapacity and would crash

**What you expected to happen**: Not to crash

**How to reproduce it (as minimally and precisely as possible)**: Call UAgonesComponent::SetCapacity

**Anything else we need to know?**: You need to use an array of FStringFormatArg to process FString::Format

**Environment**:
- Agones version: 1.15 (current)
- Kubernetes version (use `kubectl version`): na
- Cloud provider or hardware configuration: na
- Install method (yaml/helm): na
- Troubleshooting guide log(s): na
- Others: Windows, Unreal Engine C++

WilSimpson:
…rs in BuildAgonesRequest

**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Closes #2169


WilSimpson:
@googlebot I signed it!

WilSimpson:
@googlebot I fixed it.

WilSimpson:
I'm unable to update for my other personal email address used wil@wilsimpson.me since it has been used by someone else for their Google account so I cannot verify that one. I have however committed with my GitHub email so it all matches for one. /assign @pooneh-m

roberthbailey:
@WilSimpson - Thanks for the bug fix! Can you please sign the CLA?

roberthbailey:
This look reasonable to me, but I'd like to get a quick review from @domgreen who has been helping maintain our unreal code. 

/assign @domgreen

WilSimpson:
> 
> 
> @WilSimpson - Thanks for the bug fix! Can you please sign the CLA?

Long story short I can't for this fork due to issues with the original committing email. Worst case I can create a new fork and commit with a different email.

roberthbailey:
> Worst case I can create a new fork and commit with a different email.

We can't merge a change without the CLA being signed, so if that is the only way to appease the CLA, then it sounds like the best path forward.

markmandel:
> Long story short I can't for this fork due to issues with the original committing email. Worst case I can create a new fork and commit with a different email.

If you rebase this PR down to a single commit authored with your @gmail.com address, and force push over the top, then the CLA bot should accept it without issue.

Also calling @highlyunavailable who has been doing lots of Unreal stuff as of late.

markmandel:
Just a heads up - Release Candidate is next Tuesday, so if you would like this fix in the next release, that'll be the target to aim for 👍🏻

WilSimpson:
Thanks for the information about re-basing Mark, I did not know that! I've re-based to the current branch now and submitted the fix. Unfortunately I wasn't able to get to your message until today.

Also, here is a link to the Unreal documentation about FString::Format for completeness: https://docs.unrealengine.com/4.26/en-US/API/Runtime/Core/Containers/FString/Format/2/

highlyunavailable:
This looks correct though I'm not sure why it's causing you trouble. This is a pretty central method to Agones for Unreal and the `{}` version is in production for me right now with no issues. Are you using an old version of Unreal that this is causing an issue for? This should be an implicit constructor (see below) so I'm not sure why constructing your own array would help.

image

I don't see why it would be _wrong_ to make these changes but I'd like to understand what else happened to cause them to be necessary.

WilSimpson:
@highlyunavailable I'm using UE5 currently, but looking at documentation, the current release should also be using `TArray<FStringFormatArg>` or `TMap<FString, FStringFormatArg>` but what the current agones sdk is using is `TArray<FString>` per your screenshot.

Even if it isn't needed in the current UE release of 4.26 (due to how each struct and class is defined they may be equal) however in future release it may need to be required, such as UE5.

highlyunavailable:
Ah, that might explain it. As I said, it doesn't seem incorrect to do this, I just wasn't sure why it was a problem since it's working in 4.26.

Amusingly, the TArray source in 4.26 calls out that this might be an issue:


```c++
TArray(std::initializer_list<InElementType> InitList)
{
  // This is not strictly legal, as std::initializer_list's iterators are not guaranteed to be pointers, but
  // this appears to be the case on all of our implementations.  Also, if it's not true on a new implementation,
  // it will fail to compile rather than behave badly.
  CopyToEmpty(InitList.begin(), (SizeType)InitList.size(), 0, 0);
}
```

I have identified two optimizations that might help here: Instead of creating a new array ahead of the format call, pass the values as new `FStringFormatArg`s.

```c++
Request->SetURL(FString::Format(TEXT("http://localhost:{0}/{1}"), {FStringFormatArg(HttpPort), FStringFormatArg(Path)}));
```

The implicit constructor of `{}` then creates a single TArray of FStringFormatArg instead of having to create an empty TArray, allocate space for 1, then re-allocate for a 2nd item. Not a huge win, but it's still avoiding at least 1 allocation per call, which is always nice, and the compiler might then be smart enough to optimize it further.

Note that I am also no longer de-referencing the FString values - the constructor for FStringFormatArg can take FStrings directly.

Unfortunately, FString::Format isn't smart enough to take `TArray<FStringFormatArg, TInlineAllocator<2>>`, but this absolutely compiles and _should_ avoid any heap allocation for the array:

```c++
Request->SetURL(
    FString::Format(TEXT("http://localhost:{0}/{1}"),
                    static_cast<FStringFormatOrderedArguments>(TArray<FStringFormatArg, TInlineAllocator<2>>{
                        FStringFormatArg(HttpPort),
                        FStringFormatArg(Path)
                    })));
```

Definitely test it with UE5 though!

highlyunavailable (changes requested):
See comment in main thread about allocations. Also check other areas that `FString::Format` is used to see if those are problematic as well.

markmandel:
Just checking in - wasn't sure where this ended up on changes/approvals?

highlyunavailable:
I had some concerns around how the arrays were being constructed + additional usages of {Args} for `FString::Format` that weren't being addressed, so I asked for some changes.

WilSimpson:
I have updated the changes as requested by @highlyunavailable. Everything runs smoothly on UE5ea2. I did not test in 4.26, however I do not believe these changes should effect 4.26 as it follows their guidelines as well.

There were not issues with passing a single element, however I converted them all for completeness and to avoid future errors. Since there was so much repetitive code, I included a few static method helper functions. I believe the few cycles added by making the additional private static methods out weigh the readability for the repetitiveness. Let me know if you disagree.

highlyunavailable (approved):
Looks fine to me. There may be some formatting nits but I can confirm it builds fine on 4.26.2 as well.

markmandel (approved):
Yay! Looks like we have a winner. Approving and merging!

Issue #1183, opened 2019-11-13
Adding AGONES_SDK_GRPC_HOST to NewSDK


castaneai:
In local development, I'd like to run Agones SDK server on docker-compose.
However, the current Agones SDK always connect to `localhost`. 

GameServer and Local SDK server cannot be run on different containers.

https://github.com/googleforgames/agones/blob/10719304a746990a9d4cdad9a1a05814713a9beb/sdks/go/sdk.go#L45-L49

One solution is to add `AGONES_SDK_GRPC_HOST` that can override localhost. What do you think?

roberthbailey:
Out of curiosity, what value would you put in for the server address running on docker compose? I'm assuming compose doesn't had a concept of pods to put the two containers into the same network namespace, which is why you would need to override the address.

castaneai:
For example, in the following `docker-compose.yml` I will set `AGONES_SDK_GRPC_HOST=agones-sdk-server`. 
(docker-compose provides DNS resolution with the same name as the service name.)

```yaml
version: "3"
services:
  gameserver:
    ...
  agones-sdk-server:
    ...
```

markmandel:
Could you do host networking for all your containers in compose?

(Not saying that's a great idea, but it's a workaround for now)

roberthbailey:
Can you steer containers in compose to land on the same underlying node? If not, host networking wouldn't help.

markmandel:
Docker compose is single node isn't it? (I haven't used it for a while) - so essentially local development only?

roberthbailey:
Good question. I've never used it, but based on https://docs.docker.com/compose/production/ it looks like you can run across multiple nodes using docker swarm: 

> If you want to scale up your application, you can run Compose apps on a Swarm cluster.

markmandel:
@castaneai I assume you are using this for just local development?

It sounds like you are trying to replicate a Pod system locally for development?

castaneai:
Yes, I'm using docker-compose for local development (on single node).
I have no intention to run on Swarm.
The reason I want to separate services in this way is because Docker recommends running only one process on one container.

roberthbailey:
In that case, does @markmandel's suggestion above about using host network unblock you?

castaneai:
That's true for Linux. However, host network is not supported on Mac and Windows. 😢 

https://docs.docker.com/network/host/
[...]

PR #2172, opened 2021-07-07
documentation - local sdk docker

dzmitry-lahoda:
**What type of PR is this?**
/kind documentation

Adds docker as option to run local sdk

### DIFF
--- a/site/content/en/docs/Guides/Client SDKs/local.md
+++ b/site/content/en/docs/Guides/Client SDKs/local.md
@@ -27,6 +27,7 @@ You will find the executables for the SDK server, one for each type of operating
 - `sdk-server.windows.amd64.exe` - Windows
 - `sdk-server.darwin.amd64` - macOS
 - `sdk-server.linux.amd64` - Linux
+- `docker run --env local=true gcr.io/agones-images/agones-sdk:1.15.0` - Docker (runs in Windows container)
 
 To run in local mode, pass the flag `--local` to the executable.
 


site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -27,6 +27,7 @@ You will find the executables for the SDK server, one for each type of operating
 - `sdk-server.windows.amd64.exe` - Windows
 - `sdk-server.darwin.amd64` - macOS
 - `sdk-server.linux.amd64` - Linux
+- `docker run --env local=true gcr.io/agones-images/agones-sdk:1.15.0` - Docker (runs in Windows container)
```
markmandel, line 30: What do you mean here by "runs in Windows container" ?

Also, if you want to do this, we should also cover how to mount the config file into the container, since the Docker image has no access to the local filesystem otherwise.

roberthbailey: I just downloaded the zip file and it only contains the 3 files that are already mentioned in the documentation. So as-is, this is a bit misleading. 

It looks like you are suggesting an alternative approach where instead of running the windows executable, you are running the container image in docker configured for local mode.

This sounds like an interesting alternative to what is currently documented but I think it needs more explanation so that developers understand _why_ they would want to do this. When is this preferable to running the sdkserver directly? When is it preferable to using something like minikube or kind if you are running the sdk server and/or game server as a container image?

dzmitry-lahoda: yep, seems should update. use case - you have traditional desktop server - in docker it fails. so when it in docker, need to access agones. windows does not have host docker network, and avoiding patching gameserver is awesome. so can put agones and gameserver into same docker compose (not tried kind/minikube yet). what is awesome - agones sdk has windows docker making that possible, but sure. need update, may be will provide after file cli fix, so can run clean solution e2e https://github.com/googleforgames/agones/pull/2174

markmandel: This can also be run `docker run gcr.io/agones-images/agones-sdk:1.15.0 --local=true` which I think looks cleaner anyway, and means you don't have to wait for the env fix for FILE, you could append `--file` to this command as well.

But I concur with Robbie - this probably warrants it's own "Running with a local container" (or similar naming) section, covering `docker run` and `docker compose`.

If you are going down that path, https://github.com/googleforgames/agones/issues/1183#issuecomment-554820747 might also be a useful trick on how to share networks between the containers.

markmandel: Touching base - just wanted to check in and see if you were planning on updating this PR?

markmandel:
👋🏻  just checking in to see if you plan to continue this PR? Marking it as stale for now, but let us know if you plan to return to it. Otherwise, let's close it off at the end of next week.

markmandel:
Closing this. If there is more to add, please reopen.

PR #2171, opened 2021-07-07
Custom fleet autoscaler resync interval

jie-bao:
**What type of PR is this?**
/kind feature

**What this PR does / Why we need it**:
Add a new property `sync` to FleetAutoScaler for supporting custom sync interval.
See more details in: #1955 

**Which issue(s) this PR fixes**:

Closes #1955 

**Special notes for your reviewer**:
None

PR #2174, opened 2021-07-07
Allow FILE env variable for local SDK server

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Looks like we forgot to bind the file argument flag to an environment variable input as well, even though we documented that you could also do this.

Well now you can!

N/A

N/A


markmandel:
Flaky. Handholding helm because #2171 has CRD changes.

roberthbailey (approved):

PR #2177, opened 2021-07-09
Allow disabling of all allocator secrets in helm chart

### LINKED ISSUE #2175: Provide an easier way to bring your own certificates via helm chart installation
There doesn't seem to be a great way to bring your own certificates when using the helm installation method. As it stands now, it's possible to either allow automatic certificate generation, or disable generation of certs, which then puts in place static certificates that exist in the helm chart.

The only way to bring your own certificates is to clone the chart, edit the files, and then install from the clone.

I have a working branch that allows disabling the creation of all k8s secrets (specifically for the allocator). This enables me to provision the secrets up-front in kubernetes, and then start up agones using those secrets (and thus certificates). I think this is the ideal situation.

The other possibility that could be done in addition to the proposed solution (or instead of), is to allow specifying the certs in your values file, which could then be passed in using different methods with helm (and other helm-wrapper tools). I went with the first solution in my test because it allows the handling of secrets outside of helm, which opens up more sophisticated secrets management methods.

**Additional context**
The diff on my proposed solution. If there are no objections, I'll gladly open a PR.

sudermanjr:
Signed-off-by: Andrew Suderman <andrew@sudermanjr.com>

**What type of PR is this?**
/kind feature

Provide a flag to completely disable allocator certificate generation

Closes #2175 

TODO:
- [x] Add documentation

### DIFF
--- a/site/content/en/docs/Advanced/allocator-service.md
+++ b/site/content/en/docs/Advanced/allocator-service.md
@@ -101,6 +101,12 @@ echo $TLS_CA_VALUE | base64 -d > ca.crt
 # echo $TLS_CA_VALUE | base64 -D > ca.crt
 ```
 
+{{% feature publishVersion="0.16.0" %}}
+### Bring Your Own Certificates (advanced)
+
+If you would like to completely manage the tls secrets outside of helm, you can create them in the namespace where agones is going to be installed, and then set the helm value `agones.allocator.disableSecretCreation` to `true`. This method will also work with the cert-manager method, as long as your certificate and secret are created ahead of time, and you populate the `allocator-tls-ca` and `allocator-client-ca` yourself.
+{{% /feature %}}
+
 ## Client Certificate
 
 Because agones-allocator uses an mTLS authentication mechanism, a client must provide a certificate that is accepted by the server.
--- a/install/helm/agones/templates/service/allocation.yaml
+++ b/install/helm/agones/templates/service/allocation.yaml
@@ -250,6 +250,7 @@ roleRef:
 
 {{- end }}
 
+{{- if not .Values.agones.allocator.disableSecretCreation }}
 ---
 # Allocation CA
 {{- $selfSigned := genSelfSignedCert "" nil nil 3650 }}
@@ -334,5 +335,6 @@ metadata:
 data:
   tls.crt: {{ b64enc $selfSigned.Cert }}
   tls.key: {{ b64enc $selfSigned.Key }}
+{{- end }}
 {{- end }}
 {{- end }}
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -153,6 +153,7 @@ agones:
       loadBalancerIP: ""
       loadBalancerSourceRanges: []
       annotations: {}
+    disableSecretCreation: false
     generateTLS: true
     generateClientTLS: true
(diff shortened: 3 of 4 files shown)


sudermanjr:
@roberthbailey I am realizing I don't like the name of this flag that I added. What do you think about `disableSecretCreation`?

site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -101,6 +101,10 @@ echo $TLS_CA_VALUE | base64 -d > ca.crt
 # echo $TLS_CA_VALUE | base64 -D > ca.crt
 ```
 
+### Bring Your Own Certificates (advanced)
```
roberthbailey, line 104: This won't work until you use the helm chart from the upcoming release, so it should be "hidden" in a release guard: https://agones.dev/site/docs/contribute/#within-a-page

site/content/en/docs/Advanced/allocator-service.md
```diff
@@ -101,6 +101,10 @@ echo $TLS_CA_VALUE | base64 -d > ca.crt
...
+
+If you would like to completely manage the tls secrets outside of helm, you can create them in the namespace where agones is going to be installed, and then set `agones.allocator.disableSecretCreation` to `true`. This method will also work with the cert-manager method, as long as your certificate and secret are created ahead of time, and you populate the `allocator-tls-ca` and `allocator-client-ca` yourself.
```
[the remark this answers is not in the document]
roberthbailey: What happens during install if the secrets aren't already in place? Does the helm install fail? Or does the install succeed and agones is non functional?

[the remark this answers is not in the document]
sudermanjr: If you don't set the `wait` or `atomic` flags when using helm, the installation will succeed and the allocator pods will fail to start.

site/content/en/docs/Advanced/allocator-service.md
sudermanjr: Neat. Fixed

sudermanjr: Can/should I do the same with the helm values table in the docs?

roberthbailey: Yes. With tables you have to copy/paste the whole table, which is super annoying. Or maybe that is what the **New Configuration Features:** table is for? @markmandel ?

markmandel: That's why it's there! So you don't have to copy paste the whole thing 😄  because it is super annoying.

sudermanjr: Looks like that table hasn't been merged since 1.15. Do you want me to combine that table into the main one and create a new one for 1.16?

roberthbailey: That would be great! I wonder if our release checklist is missing a step...

roberthbailey: Please change the publish version for the New Configuration Features to 1.16.0 and move the rows that were already in that table up to the visible table. Thanks!

sudermanjr: Done

roberthbailey (approved):

PR #2199, opened 2021-07-27
Update proto and allocator for advanced allocation

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Update the proto for allocation and related converters to allow advanced allocation to also occur through the Allocation endpoint, and not just the `GameServerAllocation` CRD.

This has been tested for backward compatability with the previous proto version (ran allocation e2e from `main` branch against an install of this PR).

Work on #2148

The only outstanding item for #2148 is updates to the documentation!


markmandel:
/cc @ilkercelikyilmaz please take a look!

I've tested the proto changes for backward compatibility (see comment above), but would be good to get your extra eyes on this!

markmandel:
Helm got stuck in "pending-upgrade" - that's a new one. Might be time to update helm as well.

pkg/allocation/converters/converter.go
```diff
@@ -132,28 +133,84 @@ func convertLabelSelectorToInternalLabelSelector(in *pb.LabelSelector) *metav1.L
...
 
+func convertGameServerSelectorToInternalGameServerSelector(in *pb.GameServerSelector) *allocationv1.GameServerSelector {
+	if in == nil {
+		return nil
+	}
+	result := &allocationv1.GameServerSelector{
+		LabelSelector: metav1.LabelSelector{MatchLabels: in.GetMatchLabels()},
+	}
+
+	if runtime.FeatureEnabled(runtime.FeatureStateAllocationFilter) {
```
roberthbailey, line 144: Should this also check for `in.GameServerState != nil`?

pkg/allocation/converters/converter_test.go
```diff
@@ -15,27 +15,35 @@
...
 func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
+	allocated := agonesv1.GameServerStateAllocated
+	ready := agonesv1.GameServerStateReady
+
 	tests := []struct {
 		name               string
 		in                 *pb.AllocationRequest
 		want               *allocationv1.GameServerAllocation
 		skipConvertFromGSA bool
+		features           string
```
roberthbailey, line 42: Since you put this right after name in the test cases, I think you should put it second in the struct as well (so that ordering is consistent and expected).

proto/allocation/allocation.proto
```diff
@@ -87,6 +87,24 @@ message MetaPatch {
...
 // LabelSelector used for finding a GameServer with matching labels.
 message LabelSelector {
+  // Labels to match.
+  map<string, string> matchLabels = 1;
+}
+
+// GameServerSelector used for finding a GameServer with matching filters.
+message GameServerSelector {
   // Labels to match.
 	map<string, string> matchLabels = 1;
```
roberthbailey, line 97: The indentation seems a bit off here.

proto/allocation/allocation.proto
```diff
@@ -87,6 +87,24 @@ message MetaPatch {
...
\ No newline at end of file
+  enum GameServerState {
+     READY = 0;
+     ALLOCATED = 1;
+  };
+  GameServerState gameServerState = 2;
+  PlayerSelector players = 3;
+}
+
+// PlayerSelector is filter for player capacity values.
```
roberthbailey, line 106: Consider adding a comment that minAvailable must be less than maxAvailable.

pkg/allocation/converters/converter.go
markmandel: It's not a pointer, the proto generates this struct ([source](https://github.com/markmandel/agones/blob/1ac724d635c27a77081be9f714671f697a7fe48e/pkg/allocation/go/allocation.pb.go#L435-L444)):

```go
// GameServerSelector used for finding a GameServer with matching filters.
type GameServerSelector struct {
	// Labels to match.
	MatchLabels          map[string]string                  `protobuf:"bytes,1,rep,name=matchLabels,proto3" json:"matchLabels,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
	GameServerState      GameServerSelector_GameServerState `protobuf:"varint,2,opt,name=gameServerState,proto3,enum=allocation.GameServerSelector_GameServerState" json:"gameServerState,omitempty"`
	Players              *PlayerSelector                    `protobuf:"bytes,3,opt,name=players,proto3" json:"players,omitempty"`
	XXX_NoUnkeyedLiteral struct{}                           `json:"-"`
	XXX_unrecognized     []byte                             `json:"-"`
	XXX_sizecache        int32                              `json:"-"`
}
```

So no nil check needed.

pkg/allocation/converters/converter_test.go
markmandel: Yeah, looks like I have done that in past. Good catch to be consistent 👍🏻

proto/allocation/allocation.proto
markmandel: I didn't see that in my IDE 👍🏻 Thanks!

proto/allocation/allocation.proto
markmandel: I like it!

roberthbailey (approved):

PR #2205, opened 2021-07-29
Support graceful termination

bennetty:
**What type of PR is this?**
/kind feature

To support terminationGracePeriodSeconds. Terminate only when SIGKILL is received.
Currently when terminationGracePeriodSeconds is defined for a gameserver, the agones SDK sidecar terminates immediately when it receives a SIGTERM.

Closes #


markmandel:
Doesn't this mean that the sidecar will _always_ wait for `terminationGracePeriodSeconds` - which for game servers that shut down when getting an interrupt, that's okay.

Or maybe to ask another question: Why is it bad that the sidecar shuts down on interrupt (which I would argue is when it is supposed to?)

bennetty:
@markmandel yeap, the default terminationGracePeriodSeconds is 30s. this is in the context of node upgrades without downtime, having the sidecar running after SIGTERM would allow the state of the gameserver to be updated while waiting for the terminationGracePeriodSeconds.

roberthbailey:
> Why is it bad that the sidecar shuts down on interrupt (which I would argue is when it is supposed to?)

I think the scenario is:

1. Node gets scheduled for upgrade
2. Node starts the draining process, which includes terminating all pods running on the node
3. Pod receives signal to start shutting down, since it is being terminated
4. Game server continues to run to allow the in progress game to conclude (up to `terminationGracePeriodSeconds`)
5. Game servers all exit cleanly
6. Node is empty and gets removed

Right now step 4 fails, because the game cannot continue to run successfully if the sidecar exits immediately when the kubelet tries to drain all running pods from the node.

markmandel:
> Right now step 4 fails, because the game cannot continue to run successfully if the sidecar exits immediately when the kubelet tries to drain all running pods from the node.

Why does this step fail?

> would allow the state of the gameserver to be updated while waiting for the terminationGracePeriodSeconds.

Why do you need to update the state of the GameServer?

roberthbailey:
To preserve the current behavior, could users set `terminationGracePeriodSeconds` to 0 on their game servers?

roberthbailey:
>> Right now step 4 fails, because the game cannot continue to run successfully if the sidecar exits immediately when the kubelet tries to drain all running pods from the node.
>
> Why does this step fail?

I can think of a few reasons where the game server would require making calls to the sidecar to continue to run or to exit cleanly. One is to allow new players to join by watching the game server state (say it's a server waiting for the last 2 players to start a match). Another is setting labels/annotations to communicate status out of the game server. 

But @bennetty likely has an actual concrete use case that would be more enlightening than me making up fictional ones.

bennetty:
it depends on the state of the gameserver being terminated, if the gameserver has already been allocated, the sidecar does not need to be running, 
however if the gameserver is ready, and it starts being drained(for the period of the terminationGracePeriodSeconds) agones would still allocate that gameserver as its still seen as Ready

markmandel:
> however if the gameserver is ready, and it starts being drained(for the period of the terminationGracePeriodSeconds) agones would still allocate that gameserver as its still seen as Ready

So what I'm hearing is: We should not allocate to GameServers that are running on Nodes that have been marked for deletion. This seems like a good idea 😄 (I'm assuming we can see this somewhere on the Node information, I hope).

My biggest concern with the PR in its current state: The sidecar will wait for the kill signal even when it doesn't have to. So shutdown of GameServers will suddenly take a long time where previously they would respond to the interrupt signal correctly.

markmandel:
Also - editing the central signal handler - this will run to _every component of Agones_ will now no longer fail fast on interrupt as they should.

bennetty:
agreed!  would it be better if 
1. seperate signal handler for the sidecar
2. adding a flag to switch between the 2 signal handler for the sidecar?

markmandel:
🤔 You know what I'm really thinking? What we actually want is for the sidecar to stay up and running - as long as any process connected to it is still connected to it. Essentially do a graceful termination.

I think we're trying to come at that problem at an oblique angle, and this would be the crux of the issue.

To make things entertaining, there are two API surfaces:

1. For gRPC, this is not an issue, since we have a connection, and could identify when it dicsonnects.
2. For REST - it IS an issue, since each request in unique and not a connection and I'm not sure how to solve this one.
3. For websockets, same, we can wait for disconnection.

Would that be correct?

roberthbailey:
> You know what I'm really thinking? What we actually want is for the sidecar to stay up and running - as long as any process connected to it is still connected to it. Essentially do a graceful termination.

What if instead of saying "any process connected to it" we said "the gameserver container it shares the pod with is still running"? We can use the fact that the game server is running to act as a proxy for the case where it _might_ want to make connections to the sidecar.

markmandel:
> What if instead of saying "any process connected to it" we said "the gameserver container it shares the pod with is still running"?

Given that any container that is in the Pod could be connected to the SDK, I'm not sure we can make that assumption?

It might be the only thing we can do in some scenarios (e.g. REST).

theminecoder:
What I imagine is that the SDK server shouldn't think about shutting down until Shutdown() is called or its terminated by `terminationGracePeriodSeconds`

bennetty:
@markmandel thats accurate! other containers in the pod will be able to make calls to the SDK during the `terminationGracePeriodSeconds`

@theminecoder yeah i think that would be ideal!

markmandel:
> What I imagine is that the SDK server shouldn't think about shutting down until Shutdown() is called or its terminated by terminationGracePeriodSeconds

Oh that is an interesting point! Since we know the GameServer state in the SDK, we can use that as graceful termination!

So the flow on SIGINT/SIGTERM could be something like this:

1. If currently in `Shutdowns state, then terminate straight away.
2. If not in `Shutdown` state, wait for shutdown state.
3. On `SIGKILL` terminate regardless.

I'm trying to think of a nice way of saying "if not in a shutdown state, and all grpc connections disconnect, then terminate", but I don't think we can do that reliably with the REST api in place.

But the above definitely makes sense to me 👍🏻 excellent suggestion!

roberthbailey:
It may be obvious, but this means that we should separate the signal handler for the SDK from the other processes, since we want them to have different behaviors.

bennetty:
first stab at what we discussed, looking for suggestions to better handle both gameserver status and signal! any suggestions are welcomed! :)

markmandel:
Reminder to check: Fleet scaledown - does it move the `GameServer` to `shutdown`, or does it just `Delete` it?

If it only does a `Delete` this will make Fleet scale down slower. Also noting that if does only delete, moving it to shutdown is extra requests to the K8s API. (haven't looked yet, just noting that we should check this).

Same question for deletion on a Fleet.

Meta question: Should this be behind a feature gate?

bennetty:
@markmandel fleet scaledown moves gameserver to shutdown first. however when deleting a fleet, the gameserver isnt moved to shutdown first(but it might be best practice to change gameserver status to Shutdown when SIGTERM is received?)

i think it should be behind a feature gate due to the new behaviour with fleet deletion?

bennetty:
@markmandel i was thinking that maybe we dont need this behind a feature gate? all will work as per usual _if_  gameserver state is moved to `shutdown` before termination

markmandel:
> @markmandel i was thinking that maybe we don't need this behind a feature gate? all will work as per usual _if_ gameserver state is moved to `shutdown` before termination

I'm more leaning towards a feature gate because there are changes in behaviour, and I'm not sure how they will interact 100%.

On Fleet shutdown, we're utilising [owner references](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#owners-dependents) to delete all child objects, we don't explicitly delete them (which would be an extra burden on the k8s api). This change will greatly reduce the speed of Fleet removal (which _may_ not be a problem, but might for some people). I'm not 100% sure how to solve this one. We could include an update to move to `Shutdown` if not already in [this code](https://github.com/googleforgames/agones/blob/3db9b97a39c7be12bd7f28b8a090c6f58a2ae146/pkg/gameservers/controller.go#L413-L451) in the `GameServer` controller - but that is extra API overhead, but may well be worth it (and in most cases, `GameServers` should already be in a `Shutdown` state, so we can make it a no-op if they are).

> but it might be best practice to change gameserver status to Shutdown (from the gameserver binary) when SIGTERM is received?

That's a change in behaviour that not everyone is going to catch. I'd like to avoid that where possible.

I'd always like to err on the side of conservatism when there is potential/actual behavioural changes.

Just also noting [here](https://github.com/googleforgames/agones/blob/main/pkg/gameserversets/controller.go#L492-L508) is where the GameServer Fleet gets scaled down (confirming that that does get moved to `Shutdown` as you mentioned before, so that's a 👍🏻 )

I'm also concerned about race conditions - it's entirely possible that a SDK server may get the SIGKILL notification before it gets notified of the `Shutdown` state change, but we could make sure that handle that if it occurs in either order. (that's actually a good point, we should be able to handle either order, and would solve a race condition in the update strategy above with Fleet deletion).

So I do feel like there are quite a few unknown unknowns here that we're only going to work out once we get this into production - so definitely leaning towards a Feature gate.

But seems like we're slowly moving ourselves towards a solution at least!

Thoughts?

bennetty:
@markmandel cool, makes sense! i have added a feature gate for the feature :)

markmandel:
Not your fault it's failing right now - see #2224

roberthbailey:
Tests are passing now, but I've lost context on whether the review feedback has been addressed. @bennetty - do you have a status update on this PR?

bennetty:
@roberthbailey yes! i think we agreed on the behaviour and to place it behind a feature flag so i have attempted to add a feature flag to the feature :)

bennetty:
@roberthbailey this PR is ready for review!

pkg/sdkserver/sdkserver.go
```diff
@@ -824,3 +832,25 @@ func (s *SDKServer) updateConnectedPlayers(ctx context.Context) error {
 	s.recorder.Event(gs, corev1.EventTypeNormal, "PlayerCount", fmt.Sprintf("Set to %d", gs.Status.Players.Count))
 	return nil
 }
+
+// NewSDKServerContext returns a Context that cancels when os.Kill is received
```
roberthbailey, line 836: I don't see a check in the code for os.Kill being received. It looks like in both cases cancel will only be called if the game server state is shutdown. Is that what is intended? 

I read the comment as `( (os.kill) || (os.interrupt && gs.state.shutdown) )` and not `( (os.kill || os.interrupt) && (gs.state.shutdown) )` (the latter of which seems to be what is implemented).

cmd/sdk-server/main.go
```diff
@@ -77,7 +78,11 @@ func main() {
 		time.Sleep(time.Duration(ctlConf.Delay) * time.Second)
 	}
 
-	ctx := signals.NewSigKillContext()
+	ctx := context.Background()
+	if !runtime.FeatureEnabled(runtime.FeatureGracefulTerminationFilter) {
```
roberthbailey, line 82: I'm not sure I understand why this conditional is here. 

* If the feature is enabled and the sidecar is running is test or local mode this changes the behavior to use the background context instead of the sigkillcontext. 
* If the feature is enabled, the context gets replaced below with the newsdkservercontext. That function can be passed `context.Background()` instead of `ctx` to have the same effect. 

Do we want to change the context that is used in the test and local configurations? Is there a reason for them to use the sigkillcontext or the sdkserver context? I'm wondering why the new code path is only installed when running as a sidecar below and not always, e.g.

```
ctx := signals.NewSigKillContext()
if runtime.FeatureEnabled(runtime.FeatureGracefulTerminationFilter) {
  ctx = s.NewSDKServerContext(context.Background())
}
```

pkg/sdkserver/sdkserver.go
```diff
@@ -440,6 +445,9 @@ func (s *SDKServer) Shutdown(ctx context.Context, e *sdk.Empty) (*sdk.Empty, err
 	s.logger.Debug("Received Shutdown request, adding to queue")
```
roberthbailey, line 445: Can you update the function comment (line 438) to explain that shutdown will block on the game server being shutdown (if the graceful termination feature is enabled)?

markmandel:
CI failure looks like a flake (cpp sdk conformance), running it again.

pkg/sdkserver/sdkserver.go
bennetty: Updated the comment to reflect `( (os.kill || os.interrupt) && (gs.state.shutdown) )` which is intended as per the discussions above. If the gs state never reaches shutdown, at the end of the gracefulTermination period, it will get killed regardless, this allows us to clean up the gameserver without having to wait the entire gracefulTermination period

cmd/sdk-server/main.go
bennetty: thanks for the suggestion! :) updated the conditional

roberthbailey (commented):
LGTM. I'm going to wait for @markmandel to take a look as well since he's a bit more familiar with these code paths.

cmd/sdk-server/main.go
```diff
@@ -25,14 +25,15 @@ import (
 	"strings"
 	"time"
 
+	"agones.dev/agones/pkg/util/signals"
+
```
markmandel, line 29: Small nit: Let's remove the empty line please.

pkg/util/runtime/features.go
```diff
@@ -53,21 +53,25 @@ const (
 
 	// FeatureCustomFasSyncInterval is a feature flag that enables custom the FleetAutoscaler rsync interval
 	FeatureCustomFasSyncInterval Feature = "CustomFasSyncInterval"
+
+	// FeatureGracefulTerminationFilter is a feature flag that enables SDK to support gracefulTermination
+	FeatureGracefulTerminationFilter Feature = "GracefulTerminationFilter"
```
markmandel, line 58: Not sure what the `Filter` part is alluding too here.

Maybe `FeatureSDKGracefulTermination` ?

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -34,6 +35,20 @@ The current set of `alpha` and `beta` feature gates are:
...
+| [SDK Send GameServer on Watch execution]({{< ref "/docs/Guides/Client SDKs/_index.md#watchgameserverfunctiongameserver" >}}) | `SDKWatchSendOnExecute` | Enabled | `Beta` | 1.12.0 |
+| Fix for RollingUpdate [Scale down](https://github.com/googleforgames/agones/issues/1625) | `RollingUpdateOnReady` | Enabled | `Beta` | 1.14.0 |
+| [Utilize Node ExternalDNS](https://github.com/googleforgames/agones/issues/1921) and additional [details]({{< ref "/docs/FAQ/_index.md" >}}) | `NodeExternalDNS` | Disabled | `Alpha` | 1.12.0 |
+| [Custom resync period for FleetAutoscaler](https://github.com/googleforgames/agones/issues/1955) | `CustomFasSyncInterval` | Disabled | `Alpha` | 1.17.0 |
+| [GameServer state filtering on GameServerAllocations](https://github.com/googleforgames/agones/issues/1239) | `StateAllocationFilter` | Disabled | `Alpha` | 1.14.0 |
+| [GameServer player capacity filtering on GameServerAllocations](https://github.com/googleforgames/agones/issues/1239) | `PlayerAllocationFilter` | Disabled | `Alpha` | 1.14.0 |
+| [Graceful Termination for GameServer SDK](https://github.com/googleforgames/agones/pull/2205) | `GracefulTerminationFilter` | Disabled | `Alpha` | 1.18.0 |
```
markmandel, line 50: I'd like to see some additional details on what is happening here in our documentation, especially as this does affect the overall GameServer lifecycle.

I'm thinking that adding a section on how everything shuts down here https://development.agones.dev/site/docs/guides/client-sdks/#shutdown would be a good first place to start.

pkg/sdkserver/sdkserver.go
```diff
@@ -435,11 +440,15 @@ func (s *SDKServer) Allocate(ctx context.Context, e *sdk.Empty) (*sdk.Empty, err
...
 
 // Shutdown enters the Shutdown state change for this GameServer into
-// the workqueue so it can be updated
+// the workqueue so it can be updated. If gracefulTermination feature is enabled,
+// Shutdown will block on GameServer being shutdown.
 func (s *SDKServer) Shutdown(ctx context.Context, e *sdk.Empty) (*sdk.Empty, error) {
 	s.logger.Debug("Received Shutdown request, adding to queue")
 	s.stopReserveTimer()
 	s.enqueueState(agonesv1.GameServerStateShutdown)
+	if runtime.FeatureEnabled(runtime.FeatureGracefulTerminationFilter) {
```
markmandel, line 449: I'd like to see some unit tests for this please 😃

pkg/sdkserver/sdkserver.go
```diff
@@ -824,3 +833,25 @@ func (s *SDKServer) updateConnectedPlayers(ctx context.Context) error {
...
+
+// NewSDKServerContext returns a Context that cancels when SIGTERM or os.Interrupt
+// is received and the GameServer's Status is shutdown
+func (s *SDKServer) NewSDKServerContext(ctx context.Context) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+	c := make(chan os.Signal, 2)
+	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		<-c
```
markmandel, line 845: Random thought:

Any reason why you didn't take the `ctx` from `NewSigKillContext` and then wrap it in your own ctx and wait for Shudown to happen after it was `Done()` ?

Probably a bit easier to test too.

pkg/sdkserver/sdkserver.go
```diff
@@ -824,3 +833,25 @@ func (s *SDKServer) updateConnectedPlayers(ctx context.Context) error {
 	s.recorder.Event(gs, corev1.EventTypeNormal, "PlayerCount", fmt.Sprintf("Set to %d", gs.Status.Players.Count))
 	return nil
 }
+
+// NewSDKServerContext returns a Context that cancels when SIGTERM or os.Interrupt
+// is received and the GameServer's Status is shutdown
+func (s *SDKServer) NewSDKServerContext(ctx context.Context) context.Context {
```
markmandel, line 839: I'd like to see some unit tests for this being used please 😃, to make sure that things are working as expected.

pkg/sdkserver/sdkserver.go
bennetty: thanks for the suggestion! thats much better :)

pkg/sdkserver/sdkserver.go
bennetty: had a stab at some tests but would like some suggestions please :)

pkg/sdkserver/sdkserver.go
```diff
@@ -824,3 +830,19 @@ func (s *SDKServer) updateConnectedPlayers(ctx context.Context) error {
...
+		<-ctx.Done()
+		for {
+			gsState := <-s.gsStateChannel
+			if gsState == agonesv1.GameServerStateShutdown {
+				cancel()
+			}
+		}
+	}()
+	return sdkCtx
+}
```
markmandel, line 848: 🤔 I feel like this would work, and be a bit more concise - if the parent context cancels, it should flow down into the child.

```suggestion
// NewSDKServerContext returns a Context that cancels when SIGTERM or os.Interrupt
// is received and the GameServer's Status is shutdown
func (s *SDKServer) NewSDKServerContext(ctx context.Context) context.Context {
	ctx, cancel := context.WithCancel(ctx)
	go func() {
		for {
			gsState := <-s.gsStateChannel
			if gsState == agonesv1.GameServerStateShutdown {
				cancel()
			}
		}
	}()
	return ctx
}
```

pkg/sdkserver/sdkserver_test.go
```diff
@@ -1345,6 +1346,192 @@ func TestSDKServerPlayerConnectAndDisconnect(t *testing.T) {
...
+		case <-time.After(timeout):
+			assert.Fail(t, "should have gone to Reserved by now")
+		}
+	}
+
+	_, err = sc.Ready(sdkCtx, &sdk.Empty{})
+	require.NoError(t, err)
+	assert.Equal(t, agonesv1.GameServerStateRequestReady, sc.gsState)
+	//	Mock interruption signal
+	cancel()
```
markmandel, line 1399: Nice! We can do this now since we're passing in the interrupt channel 👍🏻

pkg/sdkserver/sdkserver.go
markmandel: Tests look good! 👍🏻

markmandel (commented):
Looking good! Only one suggestion on the context building function, but otherwise, LGTM!

markmandel:
I think this is a flake, retrying.

```
time="2021-09-29 17:56:00.425" level=info msg="error waiting for fleet condition" error="timed out waiting for the condition" fleet=simple-fleet-h2zkc
--- FAIL: TestFleetAutoScalerRollingUpdate (332.53s)
    fleetautoscaler_test.go:231: error waiting for fleet condition on fleet simple-fleet-h2zkc
FAIL
time="2021-09-29 17:56:00.940" level=info msg="sent UDP packet" address="34.127.24.242:7644" test=TestGameServerUnhealthyAfterReadyCrash
time="2021-09-29 17:56:00.988" level=info msg="Namespace 1632937671 is deleted"
FAIL	agones.dev/agones/test/e2e	489.681s
FAIL
make[1]: ***
```

pkg/sdkserver/sdkserver.go
bennetty: i tried this but in the event that the parent context is cancelled(main signal handler), the child will be cancelled as well(and not wait for shutdown state)🤔

markmandel: aaah yes, of course. Good point. In which case, let's get this merged 👍🏻

markmandel (approved):

markmandel:
Seeing this error across builds:

```
 ---> 1a7b4a6ed3e4
Step 4/11 : RUN wget -q https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh -O /cmake-3.14.1-Linux-x86_64.sh
 ---> Running in 77510f566e3e
includes/sdk.mk:106: recipe for target 'build-build-sdk-image' failed
make[4]: Leaving directory '/workspace/build'
includes/build-image.mk:54: recipe for target 'ensure-image' failed
includes/sdk.mk:123: recipe for target 'ensure-build-sdk-image' failed
make[3]: *** [ensure-image] Error 2
make[2]: *** [ensure-build-sdk-image] Error 2
Unable to find image 'agones-build-sdk-cpp:37cd08900e' locally
docker: Error response from daemon: pull access denied for agones-build-sdk-cpp, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
make[1]: *** [run-sdk-command] Error 125
includes/sdk.mk:88: recipe for target 'run-sdk-command' failed
includes/sdk.mk:72: recipe for target 'run-sdk-command-cpp' failed
make: *** [run-sdk-command-cpp] Error 2
```

Looks like https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh -O /cmake-3.14.1-Linux-x86_64.sh doesn't exist!

markmandel:
Ah think I found it - 

```
/usr/bin/make ensure-image IMAGE_TAG=agones-build-sdk-base:48cd8562c3 BUILD_TARGET=build-build-sdk-image-base
The command '/bin/sh -c wget -q https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh -O /cmake-3.14.1-Linux-x86_64.sh' returned a non-zero code: 5
make[4]: *** [build-build-sdk-image] Error 5
make[4]: Entering directory '/workspace/build'
```

markmandel:
Running once more. I can run the command locally, I wonder if cmake.org had a temporary blip.

markmandel:
😖 I can run ``make run-sdk-conformance-test-cpp` locally. I'm very confused.

markmandel:
Rerunning again, because very confused.

markmandel:
Made an update, let's see how it goes this time 🤞🏻

PR #2230, opened 2021-08-24
remove allocated 2 times from dot file

dzmitry-lahoda:


roberthbailey (approved):
I don't see any difference in the rendered image between https://961990f-dot-preview-dot-agones-images.appspot.com/site/docs/reference/gameserver/ and https://agones.dev/site/docs/reference/gameserver/ but this seems like a good cleanup.

dzmitry-lahoda:
generating code from this is broken, enum. and i prefer reading dor files directly than looking at image for some reason.

markmandel:
Just tested to make sure - image generation worked on my machine, so not sure what went wrong? Did you possibly not delete the file before running `make site-images` ?

PR #2234, opened 2021-08-25
Unlock mutex before returning on error in SDKServer.updateState

skystar-p:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

In https://github.com/googleforgames/agones/blob/15678b65acd58c8fe9de8eb11e336375be3b4d79/pkg/sdkserver/sdkserver.go#L294, function is not unlocking `gsUpdateMutex` before returning function due to an error. This PR add a simple line that unlocks mutex.

Couldn't find related issue.


skystar-p:
@googlebot I signed it!

markmandel (approved):
Yeah, this looks like a legit bug lurking in the darkness. Approved the fix!

PR #2246, opened 2021-08-31
Add nil check to fleet autoscaler validation for sync field

lambertwang:
**What type of PR is this?**
>
> /kind breaking
/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Adds a nil check to the fleet autoscaler validatio routine. This prevents a null pointer exception if the sync field is absent

Closes #


markmandel (approved):
I'm still not sure how it could end up being nil, but better to not panic.

Issue #2263, opened 2021-09-14
Allow the allocator service to use the go http/2 server


roberthbailey:
**Is your feature request related to a problem? Please describe.**
https://github.com/googleforgames/agones/pull/1902 added support for REST to the allocator service, with a neat trick to reuse the same port for both gRPC and REST using a mux. However, in doing so, the allocator service switched from using `grpc.(*Server).Serve` to `grpc.(*Server).ServeHTTP`, the latter of which is [experimental](https://github.com/grpc/grpc-go/blob/2608e38e6386be7400720fecf2ece176c4cbc1b2/server.go#L933-L960) and has [noticeably worse performance](https://github.com/grpc/grpc-go/issues/586#issuecomment-286257439). 

There are certainly cases where performance isn't an issue and it's convenient to run the protocols on the same port (only one exposed port / firewall rule). 

However, we have a use case where we only need to run gRPC and the experimental implementation used by `grpc.(*Server).ServeHTTP` is not compatible with the gRPC client (whereas the implementation in `grpc.(*Server).Serve` works as expected).

**Describe the solution you'd like**
I'd like to add a flag to the allocator service to restore the old behavior: if set to true, the flag would disable the mux (and therefore the REST API) and use `grpc.(*Server).Serve`. The flag would default to false, keeping the current behavior for compatibility. 

**Describe alternatives you've considered**
I've looked into https://github.com/soheilhy/cmux as a replacement mux that might still allow us to use `grpc.(*Server).Serve`. It seems [a bit complicated to set up](https://github.com/grpc/grpc-go/issues/586#issuecomment-589170186), and I'm not sure how well it would work with our setup with client certs. So I'm opting for the simpler solution and we can revisit using cmux later if it makes sense.

markmandel:
The only thought I had:

Rather than having an on/off switch - maybe we should make two options to configure the ports for the gRPC and REST service:

So:
1. If the ports are the same (default), then things are the same as now
2. If the ports are different, then you there is no muxing, and is served on different ports (maybe on different services)

Looking at our config settings, I think we actually could align on this nicely, since we have:

* `agones.allocator.http.port`
* agones.allocator.http.serviceType
* agones.allocator.http.loadBalancerIP
* agones.allocator.http.loadBalancerSourceRanges
etc

We could do the same thing, but as `agones.allocator.grpc`.

Probably some details to work out, but thought I would float the idea, see if it would stick. WDYT?

roberthbailey:
[...]

PR #2268, opened 2021-09-17
Separate the Helm value for the allocator service name from its service account name

rcreasey:
**What type of PR is this?**
/kind bug

Separates the allocation service name from allocation service account name during a helm installation.  There are instances where I want to use the default `agones-allocator` service account name, yet customize the actual service name.  Additionally, should the service account name require updates, this will also impact the allocation service name (which could cause breaking changes to workloads depending on that fqdn).

### DIFF
--- a/site/content/en/docs/Installation/Install Agones/helm.md
+++ b/site/content/en/docs/Installation/Install Agones/helm.md
@@ -174,9 +174,10 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.allocator.annotations`                      | [Annotations][annotations] added to the Agones allocator pods                                   | `{}`                   |
 | `agones.allocator.resources`                        | Allocator pods [resource requests/limit][resources]                                             | `{}`                   |
 | `agones.allocator.nodeSelector`                     | Allocator [node labels][nodeSelector] for pod assignment                                        | `{}`                   |
+| `agones.allocator.serviceName         `             | Service name for the allocator                                                                  | `agones-allocator`     |
 | `agones.serviceaccount.controller.name`             | Service account name for the controller                                                         | `agones-controller`    |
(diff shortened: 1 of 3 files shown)


roberthbailey:
I'm all for making the names separate, but can you do something like

```
allocator:
  install: true
  ...
  service:
    name: agones-allocator
  ...
```

instead, since that will align with how we are planning to restructure the service configs in https://github.com/googleforgames/agones/issues/2263?

rcreasey:
> I'm all for making the names separate, but can you do something like
> 
> ```
> allocator:
>   install: true
>   ...
>   service:
>     name: agones-allocator
>   ...
> ```
> 
> instead, since that will align with how we are planning to restructure the service configs in #2263?

For sure.  This was actually my first reaction to making the change, but didn't see nested scope on any of the other variables.  Will update accordingly.

roberthbailey:
Looks like a test flake. Can you rebase on main and re-push your branch? (I can't update your branch for you)

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -174,9 +174,10 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.allocator.annotations`                      | [Annotations][annotations] added to the Agones allocator pods                                   | `{}`                   |
 | `agones.allocator.resources`                        | Allocator pods [resource requests/limit][resources]                                             | `{}`                   |
 | `agones.allocator.nodeSelector`                     | Allocator [node labels][nodeSelector] for pod assignment                                        | `{}`                   |
+| `agones.allocator.service.name`                     | Service name for the allocator                                                                  | `agones-allocator`     |
```
roberthbailey, line 177: For the new field, please add it to the table below (line 192) so that it doesn't appear in the docs until the next release. Thanks!

roberthbailey (approved):

PR #2270, opened 2021-09-17
Docs: High Density Integration Pattern

### LINKED ISSUE #1239: Game Server Allocation advanced filtering: player count, state, reallocation

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

Documentation explaining how to use the new advanced allocation features to run multiple concurrent game sessions in a single `GameServer` instance.

Closes #1239

This is tricky to explain, because it is a little complicated. All feedback appreciated to try and explain this concept!

If you have thoughts on if there are improvements / better ways to explain, please add them to this PR 👍🏻


markmandel:
https://d23d739-dot-preview-dot-agones-images.appspot.com/site/docs/integration-patterns/high-density-gameservers/ if people are looking for the preview

site/content/en/docs/Integration Patterns/high-density-gameservers.md
```diff
@@ -0,0 +1,79 @@
...
+`GameServers` at allocation time, and then return it back into the pool of allocatable `GameServers` if and when the 
+game server process deems that is has room to host another game session. 
+
+<a href="../../../diagrams/high-density.puml.png" target="_blank">
+<img src="../../../diagrams/high-density.puml.png" alt="High Density Allocation Diagram" />
+</a>
+
+{{< alert title="Info" color="info">}}
+To watch for Allocation events, there is the initial `GameServer.status.state` change from `Ready` to `Allocated`,
+but it is also useful to know that the value of `GameServer.metadata.annotations["agones.dev/last-allocated"]` will
```
roberthbailey, line 31: Is last allocated something that gets set automatically? I didn't see a clear explanation for how it's used in the UML diagram. Then I found this note which was helpful, but the example below doesn't set it, which makes me wonder how the game server can use it. Can you add a bit of text to explain where this comes from and what values to expect? Are we looking for true/false, a timestamp, something else?

site/content/en/docs/Integration Patterns/high-density-gameservers.md
```diff
@@ -0,0 +1,79 @@
...
+        agones.dev/fleet: simple-udp
+      gameServerState: Ready # Allocate out of the Ready Pool (which would be default, so backward compatible)
+  metadata:
+    labels:
+      agones.dev/sdk-available: "false" # this removes it from the pool
+```
+
+{{< alert title="Info" color="info">}}
+It's important to note that the labels that the `GameServer` process use to add itself back into the pool of 
+allocatable instances, must start with the prefix `agones.dev/sdk-`, since only labels that have this prefix are
```
roberthbailey, line 71: This prefix is a bit unfortunate in this example. When I first read the label, it sounds like "the sdk is available" instead of "a label set via the sdk indicates that the game server is saying it is available". 

Not that it's worth changing, but I'm curious why you did a trailing `-` instead of a trailing `/` here. Both would provide a namespace for the labels.

markmandel: > Not that it's worth changing, but I'm curious why you did a trailing - instead of a trailing / here. Both would provide a namespace for the labels.

In retrospect it would have been better with a `/`, but it was just a decision that got made early that stuck.

> When I first read the label, it sounds like "the sdk is available" instead of "a label set via the sdk indicates that the game server is saying it is available".

Since the label can be anything (best name I could come up with was `available`), if there is a better name you can think of, definitely open to suggestions!

site/content/en/docs/Integration Patterns/high-density-gameservers.md
markmandel: Good point! I've added new sections to the sequence diagram, as well as added to the `alert` box as well. PTAL!

markmandel:
New preview: https://2ad4e07-dot-preview-dot-agones-images.appspot.com/site/docs/integration-patterns/high-density-gameservers/

site/content/en/docs/Integration Patterns/high-density-gameservers.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Example `GameServerAllocation`
+
+The below `Allocation` will first attempt to find a `GameServer` from the `Fleet` `simple-udp` that is already 
+Allocated and also has the label `agones.dev/sdk-available` with the value of `true`.
+
+The above condition indicates that the matching game server process behind the matched `GameServer` record is able to 
+accept another game session at this time.
+
+If the above Allocated `GameServer` does not exist, then allocated a Ready `GameServer` from the `simple-udp` `Fleet`.
```
roberthbailey, line 44: nit: what about

If an Allocated `GameServer` does not exist with the desired labels, then use the next selector to allocate a Ready `GameServer` from the `simple-udp` `Fleet`.

site/content/en/docs/Integration Patterns/high-density-gameservers.md
```diff
@@ -0,0 +1,80 @@
...
+
+Whichever condition is met, once allocation is made against a `GameServer`, its label of `agones.dev/sdk-available` 
+will be set to the value of `false`, thereby removing it from any future allocations with the below schema.
```
roberthbailey, line 47: nit: what about

... value of `false` and it will no longer match the first selector, thereby removing it from ...

site/content/en/docs/Integration Patterns/high-density-gameservers.md
roberthbailey: It's verbose, but I wonder if something like `agones.dev/sdk-gs-ready-for-allocation` or `agones.dev/sdk-gs-ready-for-reallocation` or `agones.dev/sdk-gs-ready-for-new-session` might be clearer.

site/content/en/docs/Integration Patterns/high-density-gameservers.md
markmandel: I like it!

site/content/en/docs/Integration Patterns/high-density-gameservers.md
markmandel: I like the emphasis on that is won't match the first selector - that's an important point 👍🏻

site/content/en/docs/Integration Patterns/high-density-gameservers.md
markmandel: I like where this is going. I'm a fan of `agones.dev/sdk-gs-ready-for-new-session` - mainly because it's clear that it's a game session that is available within the current game server.

WDYT of `agones.dev/sdk-gs-session-available` ? Not _quite_ as verbose (35 chars vs 39) , but I think hits all the marks?

markmandel: I also like a variation on this type of label, because then SDK command becomes: `SDK.SetLabel("gs-session-available", "true")` - which I think makes more semantic sense 👍🏻

roberthbailey: I was leaning towards using "ready" in the label since it matches the terminology we use already (ready game servers) to indicate that you were moving into the state where you expected an allocation to come along. But I like your suggestion just as well. In some ways it's better to have a different word since this isn't an agones state and this page talks about "game sessions" instead of "game servers".

markmandel: > I was leaning towards using "ready" in the label

Oooh, I see where you are going there.  I like the bringing together or known concepts 👍🏻 

`SDK.SetLabel("gs-session-ready", "true")` (rather than available) ?

Which would become `agones.dev/sdk-gs-session-ready = true` WDYT?

roberthbailey: Sounds great.

markmandel:
Updated! I _think_ this should be good to go now!

roberthbailey (commented):
I found a minor typo and have one last grammatical suggestion.

site/content/en/docs/Integration Patterns/high-density-gameservers.md
roberthbailey: Re-reading the new version with this label, I like it even more than I thought I would. :)

site/content/en/docs/Integration Patterns/high-density-gameservers.md
```diff
@@ -0,0 +1,82 @@
...
+{{< alpha title="Allocation Player Filter and Allocation State Filter" gate="PlayerAllocationFilter,StateAllocationFilter" >}}
+
+Depending on the setup and resource requirements of your game server process, sometimes it can be a more economical 
+use of resources to run multiple concurrent game sessions from within a single `GameServer` instance.
+
+The tradeoff here is that this requires more management on behalf of the integrated game server process and external 
+systems, since it works around the common Kubernetes and/or Agones container lifecycle.
+
+Utilising the new allocation `gameServerState` filter as well as the existing ability to edit the 
+`GameServer` Labels at both [allocation time]({{% ref "/docs/Reference/gameserverallocation.md" %}}), and from
```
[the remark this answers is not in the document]
markmandel: Done!

site/static/diagrams/high-density.puml
```diff
@@ -0,0 +1,118 @@
...
+to determine there is an allocation event.
+end note
+
+alt <i>GameServer</i> can accept more concurrent sessions
+  Binary -> SDK: SDK.SetLabel("gs-session-ready", "true")
+  SDK --> GameServer: Sets <i>metadata.labels["agones.dev/sdk-gs-session-ready"] = "true"</i>.
+end alt
+note right
+  If the <i>GameServer</i> can accept
+  more concurrent sessions, reset the
```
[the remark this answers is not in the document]
markmandel: Done!

roberthbailey (approved):

PR #2272, opened 2021-09-21
Allow the ports for gRPC and REST to be configured for the allocator service

### LINKED ISSUE #2263: Allow the allocator service to use the go http/2 server

roberthbailey:
**What type of PR is this?**
>

/kind breaking

> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**: Makes it possible to disable the REST endpoint in the allocator service or to run gRPC and REST on different ports. This allows using the non-experimental gRPC server code.

Closes #2263

~~This is currently a work-in-progress. I still need to update the documentation before this can be merged.~~

I've added documentation. 

This is a breaking change because I'm renaming some helm parameters (which can be seen in the updates to the Makefile and the allocator service documentation). In the 1.18.0 release we should notify folks that if they rely on these helm parameters that they now have different names.

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -19,6 +19,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -43,6 +44,7 @@ import (
 	"go.opencensus.io/plugin/ocgrpc"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/status"
 	"gopkg.in/fsnotify.v1"
@@ -53,14 +55,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-var (
-	logger = runtime.NewLoggerWithSource("main")
-)
+var logger = runtime.NewLoggerWithSource("main")
 
 const (
 	certDir = "/home/allocator/client-ca/"
 	tlsDir  = "/home/allocator/tls/"
-	sslPort = "8443"
 )
 
 // grpcHandlerFunc returns an http.Handler that delegates to grpcServer on incoming gRPC
@@ -80,7 +79,7 @@ func main() {
 	conf := parseEnvFlags()
 
 	logger.WithField("version", pkg.Version).WithField("ctlConf", conf).
-		WithField("featureGates", runtime.EncodeFeatures()).WithField("sslPort", sslPort).
+		WithField("featureGates", runtime.EncodeFeatures()).
 		Info("Starting agones-allocator")
 
 	logger.WithField("logLevel", conf.LogLevel).Info("Setting LogLevel configuration")
@@ -90,6 +89,10 @@ func main() {
 	} else {
 		logger.WithError(err).Info("Specified wrong Logging.SdkServer. Setting default loglevel - Info")
 		runtime.SetLevel(logrus.InfoLevel)
+	}
+
+	if !validPort(conf.GRPCPort) && !validPort(conf.HTTPPort) {
+		logger.WithField("grpc-port", conf.GRPCPort).WithField("http-port", conf.HTTPPort).Fatal("Must specify a valid gRPC port or an HTTP port for the allocator service")
 	}
 
 	health, closer := setupMetricsRecorder(conf)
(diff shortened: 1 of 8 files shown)


install/helm/agones/templates/service/allocation.yaml
```diff
@@ -26,31 +26,46 @@ metadata:
...
   annotations:
-{{ toYaml .Values.agones.allocator.http.annotations | indent 4 }}
+{{ toYaml .Values.agones.allocator.service.annotations | indent 4 }}
 {{- end }}
 spec:
   selector:
     multicluster.agones.dev/role: allocator
   ports:
-    - port: {{ .Values.agones.allocator.http.port }}
+{{- if .Values.agones.allocator.service.http.enabled }}
```
roberthbailey, line 37: The logic here is similar to what is in `cmd/allocator/main.go` but is a bit harder to read due to the templating language.

There are 4 cases to cover:
1. If http & grpc are both enabled and use the same port, only configure the http port
1. If http & grpc are both enabled and use different ports, configure the http port and the grpc port
1. If only http is enabled, only configure the http port
1. If only grpc is enabled, only configure the grpc port

The logic to cover the 4 cases has three branches:

1. If the http port is enabled, then use it (covers cases 1-3 & skipping covers 4)
   1. If the grpc port is also enabled and it is different from the http port, use it as well (case 2)
1. If the http port is disabled but the grpc port is enabled, use it (case 4)

Locally I tested this by modifying my `values.yaml` file and running `make gen-install` to see the resulting yaml and verify that the correct number or ports were defined (1 or 2) with the correct values.

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -161,8 +184,19 @@ spec:
         - name: FEATURE_GATES
           value: {{ .Values.agones.featureGates | quote }}
         ports:
+        {{- if .Values.agones.allocator.service.http.enabled }}
         - name: https
```
rcreasey, line 188: I think that `name` should also be a changeable attribute if we're already changing the port values.  There are some use cases where changing the port to `80` and leaving the name as `https` would be disadvantageous.  If not, then perhaps we should statically name this to just be `http` to be consistent with the nomenclature of the helm values label.

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -161,8 +184,19 @@ spec:
...
-          containerPort: 8443
+          containerPort: {{ .Values.agones.allocator.service.http.targetPort }}
+        {{- if .Values.agones.allocator.service.grpc.enabled }}
+        {{- if ne .Values.agones.allocator.service.grpc.port .Values.agones.allocator.service.http.port }}
+        - name: grpc
```
rcreasey, line 192: Related to the above, it might be worth having this be defined via the helm value as well.

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -161,8 +184,19 @@ spec:
...
+          containerPort: {{ .Values.agones.allocator.service.grpc.targetPort }}
+        {{- end }}
+        {{- end }}
+        {{- else if .Values.agones.allocator.service.grpc.enabled }}
+        - name: grpc
+          containerPort: {{ .Values.agones.allocator.service.grpc.targetPort }}
+        {{- end }}
         volumeMounts:
         - mountPath: /home/allocator/tls
```
rcreasey, line 201: While it may not be directly related to this PR, it's worth noting that if you do have `disableTLS` or `disableMTLS` set to false, the `volumeMount` and corresponding `volume` should be gated off that value as well.

rcreasey (commented):
Leaving feedback, but not requiring them to be changed for the PR.

install/helm/agones/templates/service/allocation.yaml
roberthbailey: Done.

install/helm/agones/templates/service/allocation.yaml
roberthbailey: Done.

install/helm/agones/templates/service/allocation.yaml
roberthbailey: I'll do this in a follow up PR to keep this change more focused.

roberthbailey: I've created https://github.com/googleforgames/agones/issues/2276 to track this feature request.

rcreasey (approved):

markmandel (commented):
Couple of nits on logging, but otherwise LGTM

cmd/allocator/main.go
```diff
@@ -201,30 +240,44 @@ func main() {
...
 		}
 	}()
+}
 
-	// Finally listen on 8080 (http) and block the main goroutine
-	// this is used to serve /live and /ready handlers for Kubernetes probes.
-	err = http.ListenAndServe(":8080", http.DefaultServeMux)
-	logger.WithError(err).Fatal("allocation service crashed")
+func runGRPC(h *serviceHandler, grpcPort int) {
+	logger.Infof("Running the grpc handler on port %d", grpcPort)
```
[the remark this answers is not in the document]
roberthbailey: done.

cmd/allocator/main.go
```diff
@@ -181,17 +184,53 @@ func main() {
...
-	if err != nil {
+	if err := pb.RegisterAllocationServiceHandlerServer(context.Background(), mux, h); err != nil {
+		panic(err)
+	}
+
+	runHTTP(h, httpPort, grpcHandlerFunc(grpcServer, mux))
+}
+
+func runREST(h *serviceHandler, httpPort int) {
+	logger.Infof("Running the rest handler on port %d", httpPort)
```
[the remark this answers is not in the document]
roberthbailey: done.

cmd/allocator/main.go
```diff
@@ -92,6 +91,10 @@ func main() {
 		runtime.SetLevel(logrus.InfoLevel)
 	}
 
+	if !validPort(conf.GRPCPort) && !validPort(conf.HTTPPort) {
+		logger.Fatalf("Must specify a valid gRPC port or an HTTP port for the allocator service. grpc port = %d, http port = %d", conf.GRPCPort, conf.HTTPPort)
```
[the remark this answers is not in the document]
roberthbailey: done.

markmandel (approved):
Flake: 
```
    --- FAIL: TestFleetRollingUpdate/Use_fleet_Patch_false_10% (118.20s)
        fleet_test.go:331: 
            	Error Trace:	fleet_test.go:331
            	            	new replicas should be less than target + maxSurge + maxUnavailable + shift. Replicas: 11, Expected: 10
            	Test:       	TestFleetRollingUpdate/Use_fleet_Patch_false_10%
```

Retrying!

PR #2287, opened 2021-10-02
Update the link to look for issues for the release.

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

The query for closed issues was starting to return a large number (>130) issues. Scanning through some of the issues past the first page I noticed that while we generally caught recently opened (and closed) issues for a release, there were issues without a milestone for many old releases (1.17, 1.14, and even 1.1) that had been missed using the old link.


markmandel (approved):
Nice fix 👍🏻

markmandel:
One day we'll work out why the sdk conformance tests are a bit flaky.

PR #2288, opened 2021-10-04
Added check for empty healthcheck post-body

### LINKED ISSUE #2256: Sidecar REST endpoint should return 400 if healthcheck body is empty
A healthcheck POST to the REST endpoint with an empty body should return a 400 instead of a 200. The body must contain `{}` for the healthcheck to be registered by the agones sidecar. 

**Additional context**
See https://github.com/googleforgames/agones/issues/2132#issuecomment-914986274.

sankalp-r:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Closes #2256


roberthbailey:
I've created https://github.com/googleforgames/agones/pull/2289 to fix this failing test.

markmandel (approved):
Clever! I was wondering how we would solve this one, since it's all generated code!

pkg/util/https/https.go
```diff
@@ -66,3 +66,15 @@ func LogRequest(logger *logrus.Entry, r *http.Request) *logrus.Entry {
 		WithField("headers", r.Header).
 		WithField("requestURI", r.RequestURI)
 }
+
+// RequestWrapper perform checks on request
+func RequestWrapper(h http.Handler) http.Handler {
```
markmandel, line 71: I might have been a bit quick to approve - I think this could go into main.go / move into th sdk package - since it's not a generic https utility that's used across the project.

But still like the approach!

markmandel (changes requested):

pkg/util/https/https.go
roberthbailey: +1

What about just making this a function in `cmd/sdk-server/main.go` called something like `func healthcheckWrapper(...)` (not exported outside of that main pkg)?

sankalp-r: +1

@roberthbailey I agree with you.
We can put the wrapper function in `cmd/sdk-server/main.go`

roberthbailey: Sounds great. Let me know once it's ready for review, and we can try to get it merged before the release candidate tomorrow.

sankalp-r: I have made the suggested changes, but build is failing.

roberthbailey (approved):

PR #2312, opened 2021-10-14
Minikube: more options for connectivity

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

Minikube is a bit of a nightmare, since depending on your OS and your driver, you may or may not be able to connect to a UDP GameServer, with or without workarounds.

These changes capture two things:

1. A few more workaround to try and some dead ends to avoid.
2. Which OS+driver combos are known to work.
3. That it may not actually be possible, in which case, it's time to try something else.

N/A

N/A


roberthbailey:
@mvlabat - want to take another look?

PR #2315, opened 2021-10-18
Add error callback to WatchGameServer in Node.js SDK 

### LINKED ISSUE #2304: Uncatchable error in NodeJS Agones SDK when calling shutdown()
**What happened**:
Error logged from NodeJS Agones SDK:
Error: 14 UNAVAILABLE: Connection dropped

from [watchGameServer](https://github.com/googleforgames/agones/blob/release-1.17.0/sdks/nodejs/src/agonesSDK.js#L119). This happened after calling shutdown intentionally.

I should be able to catch this error, and I cannot because it is wrapped in a callback. The error callback is executed asynchronously, so It throws into a separate execution context and crashes the node process as an unhandled exception.

Instead, watchGameServer should take an onError callback or something, so that I can handle this sort of error.

Just call watchGameServer, then call shutdown.

**Environment**:
- Agones version: 1.17.0
- Kubernetes version (use `kubectl version`): v1.22.2
- Others:

steven-supersolid:
**What type of PR is this?**

/kind feature

Introduces an optional error callback to `watchGameServer` which if present will be called instead of throwing an error when the underlying stream emits an error
Updated the example code to use this callback should an error occur, although it is not intentionally triggered in the example

Closes #2304 

This only addresses possible errors thrown from `watchGameServer`


roberthbailey:
I've added a hold label to your PR. Please remove it (or ping the PR) once the documentation is ready.

PR #2367, opened 2021-11-11
expose Helm chart values for custom certs

### LINKED ISSUE #2364: allow passing certificates as values instead of files in the Helm chart
**Problem**

For some manifests, certificates are read in from the `certs` directory.  This makes it very inconvenient for a user to supply their own certificates.   Here's an example from the `allocator-tls` secret:

```yaml
data:
{{- if .Values.agones.allocator.generateTLS }}
  tls.crt: {{ b64enc $cert.Cert }}
  tls.key: {{ b64enc $cert.Key }}
{{- else }}
  tls.crt: {{ .Files.Get "certs/allocator/server.crt" | b64enc }}
  tls.key: {{ .Files.Get "certs/allocator/server.key" | b64enc }}
{{- end }}
```

**Solution**

This problem could be solved by exposing values for providing certificate data.  Using the example above:

```yaml
data:
{{- if .Values.agones.allocator.generateTLS }}
  tls.crt: {{ b64enc $cert.Cert }}
  tls.key: {{ b64enc $cert.Key }}
{{- else }}
  tls.crt: {{ default (.Files.Get "certs/allocator/server.crt") .Values.agones.allocator.tlsCrt | b64enc }}
  tls.key: {{ default (.Files.Get "certs/allocator/server.key") .Values.agones.allocator.tlsKey | b64enc }}
{{- end }}
```

The user could supply the certificate and key with:

```sh
helm install my-release agones/agones \
  --set-file agones.allocator.tlsCrt=mydir/tls.crt,agones.allocator.tlsKey=mydir/tls.key
```

rahil-p:
**What type of PR is this?**

/kind feature

**What this PR does:**

This PR provides a convenient alternative for installing Agones with custom certificates.  Instead of having to pull and modify the contents of the chart, custom certificates may be passed in as values:

```shell
helm install my-release agones/agones \
  --set agones.allocator.generateTLS=false \
  --set-file agones.allocator.tlsCert=mydir/tls.crt \ # or --set agones.allocator.tlsCert="$(cat mydir/tls.crt)"
  --set-file agones.allocator.tlsKey=mydir/tls.key \
  --set-file "agones.allocator.clientCAs.my-ca\.crt"=mydir/ca.crt
```

Closes #2364


site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -117,7 +117,9 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.controller.healthCheck.failureThreshold`    | Number of times before giving up (in seconds)                                                   | `3`                    |
 | `agones.controller.healthCheck.timeoutSeconds`      | Number of seconds after which the probe times out (in seconds)                                  | `1`                    |
 | `agones.controller.resources`                       | Controller [resource requests/limit][resources]                                                 | `{}`                   |
-| `agones.controller.generateTLS`                     | Set to true to generate TLS certificates or false to provide your own certificates in `certs/*` | `true`                 |
+| `agones.controller.generateTLS`                     | Set to true to generate TLS certificates or false to provide your own certificates              | `true`                 |
```
roberthbailey, line 120: The new lines should be added to the "New Configuration Features" table below, which will be published along with the next release (in about a week). Changes here apply to the live website, and since this feature isn't part of the chart until the next release, we want to hide it until then.

install/helm/agones/templates/extensions.yaml
```diff
@@ -38,7 +38,7 @@ spec:
         {{- if .Values.agones.controller.generateTLS }}
   caBundle: {{ b64enc $ca.Cert }}
         {{- else }}
-  caBundle: {{ .Files.Get "certs/server.crt" | b64enc }}
+  caBundle: {{ default (.Files.Get "certs/server.crt") .Values.agones.controller.tlsCert | b64enc }}
```
roberthbailey, line 41: Would it work to put "certs/server.crt" into values.yaml and just use the new variable here?

From https://helm.sh/docs/chart_template_guide/functions_and_pipelines/:

> In an actual chart, all static default values should live in the values.yaml, and should not be repeated using the default command (otherwise they would be redundant).

I'm not particularly familiar with the `--set-file` option, but it looks like if the new parameter is a file name and we keep the `.Files.Get` here referencing the variable, then it might work (but using the `--set` flag instead of `--set-file` because you wouldn't pass the file contents into the variable).

rahil-p: The new parameter expects the file's contents.  The `--set-file` option will set a value to the contents of the file at the provided path - it's equivalent to `--set value=$(cat value.txt)`.

The idea behind the new parameter is to allow the user to pass the certificate directly as a value.  When the value isn't provided, `.Files.Get` will be used as a fallback to read the file from the `certs` directory. I included this so that the old way of passing in certificates (i.e. modifying the files in `certs`) continues to be supported.

Let me know if that makes sense and/or if you'd like to see any changes.

rahil-p:
I pushed a commit that removes reading in defaults from the `certs` directory. This simplifies the template. It also should provide some security in case of a user failing to pass in their own certificate - instead of silently falling back on the repository certificate, the server will log an error and exit after failing to load empty credentials.

This is breaking for releases that disable `allocator.generateTLS` or `controller.generateTLS`.  If this isn't the right call, let me know and I'll revert it.

roberthbailey:
I like these changes, but I'm on the fence about whether this will break too many folks. 

@markmandel - WDYT?

markmandel:
🤔 I'm on the fence with the breaking change as well. I _feel_ like not many people use a custom `certs` directory (I'm actually not sure how you would do so to be honest, I assume you would have to grab the chart source, and run it locally???)

But it's hard to judge impact. Maybe we should have this PR merge with the non-breaking behaviour for now, and file an issue for the breaking change, and see if anyone objects?

rahil-p:
I'd guess that workflow is the most common - it would probably look something like this during deployment:

```sh
helm pull agones/agones --untar
cp -TR my-certs/ agones/certs/
helm install my-release ./agones --set agones.allocator.generateTLS=false
```

I think a new issue is a good call - it could benefit from some more discussion.

markmandel:
Flake in `TestAutoscalerBasicFunctions`

markmandel:
@roberthbailey you've been tracking this more closely than I have. If you are happy with this PR, I'm happy to squeeze this in before RC today.

roberthbailey (approved):

Issue #2385, opened 2021-11-30
Flakiness: Autoscaler tests


markmandel:
**What happened**:

Autoscaler tests have a tendency to be flaky.

Usually a `AssertFleetCondition` fails.

**What you expected to happen**:

Not be flaky 😄 

**How to reproduce it (as minimally and precisely as possible)**:

Usually by running CI.

You might be able to get it to run via running lots of the e2e tests over and over. I've found `TestAutoscalerStressCreate` is often a decent candidate.

**Anything else we need to know?**:

I have a two current theories I need to confirm:

1. The flakiness is in when the `CustomFasSyncInterval` is enabled, but I wasn't checking on this before when tracking the issues, so need to go back and look.
2. At least part of the problem lies in the fact that the `fasThreads` map in the autoscaler controller was using the FleetAutoscaler `name` for the key. This has no guarantee to be unique, and could therefore cause race conditions (thinking removal comes after a new addition with the same name, which happens in some e2e tests). We should use the UID of the object instead to guarantee uniqueness of the key.

**Environment**: Cloud Build
- Agones version:  dev.
- Kubernetes version (use `kubectl version`): 1.21
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): helm
- Troubleshooting guide log(s): N/A
- Others:

Digging into some fixes here: 
https://github.com/markmandel/agones/tree/wip/autoscaler-flakiness

markmandel:
I've no idea how this is passing autoscaling tests but:
https://github.com/googleforgames/agones/blob/f484a5398010a9a267284a2200821990f8802337/test/e2e/fleetautoscaler_test.go#L727

That image doesn't exist! Only 0.4 does.

Will fire a PR tomorrow, unless someone beats me to it.

markmandel:
Writing this down while I'm thinking about it, so I can track the thought. I think we need a mutex what handling the custom sync handler, as I think one could have a race condition in which you could end up with multiple attempts to shut down a `fasThread` (send a message to `terminateSignal`), but if that happens, one will lock indefinitely as there is nothing to catch the signal anymore.

Actually, while I think about this - rather than mutexing things - we can use a cancellable Context instead, since multiple calls to cancellation function won't deadlock, and we no longer need a mutex, as the request will be idempotent.

Code in question:

https://github.com/markmandel/agones/blob/eedf687fcf427659753d49a3fdde1dd7293ba2db/pkg/fleetautoscalers/controller.go#L423-L428

[...]

PR #2386, opened 2021-12-02
Fix webhook image, and e2e.AssertFleetCondition

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Found two issues:
* The webhook image points to a non-existent image.
* Fixes to AssertFleetCondition to help see flaky issues, introduced a bug where it could pass even if something failed.

Work on #2385

Also removed the t.Helper() as I found it got in the way of debugging issues, and added some extra event logging on failure,

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -39,7 +39,6 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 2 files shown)


roberthbailey:
For the first issue:

> The webhook image points to a non-existent image.

Does this mean that our test isn't actually pulling the image and exercising it? How would the test _ever_ pass if the image doesn't exist?

test/e2e/framework/framework.go
```diff
@@ -456,10 +451,8 @@ func (f *Framework) CleanUp(ns string) error {
 func (f *Framework) CreateAndApplyAllocation(t *testing.T, flt *agonesv1.Fleet) *allocationv1.GameServerAllocation {
 	gsa := GetAllocation(flt)
 	gsa, err := f.AgonesClient.AllocationV1().GameServerAllocations(flt.ObjectMeta.Namespace).Create(context.Background(), gsa, metav1.CreateOptions{})
-	if !assert.NoError(t, err) {
-		assert.FailNow(t, "gameserverallocation could not be created")
-	}
-	assert.Equal(t, string(allocationv1.GameServerAllocationAllocated), string(gsa.Status.State))
+	require.NoError(t, err)
```
roberthbailey, line 454: Should this keep the message text as well ("gameserverallocation could not be created")?

test/e2e/framework/framework.go
```diff
@@ -315,15 +314,14 @@ func (f *Framework) WaitForFleetCondition(t *testing.T, flt *agonesv1.Fleet, con
 			f.LogEvents(t, log, gs.ObjectMeta.Namespace, &gs)
 		}
 
-		return err
+		return resultErr
```
roberthbailey, line 317: Is this the line that you think may have made the test flaky?

markmandel:
> Does this mean that our test isn't actually pulling the image and exercising it? How would the test _ever_ pass if the image doesn't exist?

You ask an excellent question. To be honest - I am not 100% sure. My only theory is that `AssertFleetCondition` wasn't failing when it should have done, and that more often than not that would allow the full e2e test suite to pass, since `AssertFleetCondition` is used in a large majority of our e2e tests.

I only found this because I was trying to solve some other race conditions, and realised things weren't failing when they should.

This should solve the timeout issues we were seeing in our e2e tests, since you would likely get multiple of these 5 minute waits back to back, and an individual test will timeout after 10 minutes.

test/e2e/framework/framework.go
markmandel: 🤷🏻 I can put it back in if you feel strongly. Since we get a stacktrace you can find where the error is coming from, and it should be a rare failure, I wasn't too fussed about whether there was a message or not.

test/e2e/framework/framework.go
markmandel: So this is the line that would let the `WaitForFleetCondition` seem to pass.

The line `gsList, err := f.ListGameServersFromFleet(flt)` would have overwritten the `err` variable (and it would be strange for it to fail) -- so if there was a timeout on the `wait.PollImmediate`, it wouldn't be caught and would allow the whole process to flow through, which is bad.

markmandel: Sorry - to answer the original question: I don't think this affects flakiness per-se -- but it does mean that the root cause of why e2e tests where failing was often harder to track down, because only a test that relies on the expected state of this test would fail after the fact.

roberthbailey: Got it.

roberthbailey (approved):

PR #2417, opened 2021-12-29
Generating TLS certificates for agones controller and admissions cont…

### LINKED ISSUE #2389: generateTLS with lookup of existing secret when using agones chart as dependency
I would like to include agones as dependency When using agones as a dependency subchart there is problem as described in [Install agones using Helm](https://agones.dev/site/docs/installation/install-agones/helm/#controller-tls-certificates) with regenerating certs within any `helm upgrade`, even when nothing changes. 

The alternative is to use `certs.sh` and `generateTls: false`, but it is additional step to remember when deploying agones, and pain when using agones as dependency - has to pull dependency independently and extract script from `tgz`.

There is possibility to use `lookup` template function introduced in `Helm 3.1` to check if secret with cert exists and use it when upgrading.
[Helm issue 3053](https://github.com/helm/helm/issues/3053#issuecomment-782837107)
```
apiVersion: v1
kind: Secret
metadata:
  name: foobar-secret
type: Opaque
{{- $previous := lookup "v1" "Secret" .Release.Namespace "foobar-secret" }}
data:
  {{- if $previous }}
  foobarPassword: {{ $previous.data.foobarPassword }}
  {{- else if .Values.foobarPassword }}
  foobarPassword: {{ .Values.foobarPassword | b64enc | quote }}
  {{- else }}
  foobarPassword: {{ randAlphaNum 40 | b64enc | quote }}
  {{- end }}
```

It is also possible to use annotation `"helm.sh/hook": "pre-install"` as described here [https://medium.com/nuvo-group-tech/move-your-certs-to-helm-4f5f61338aca](https://medium.com/nuvo-group-tech/move-your-certs-to-helm-4f5f61338aca), but it seems to be older solution

**Additional context**

gorylpe:
**What type of PR is this?**
/kind feature

This PR is extending generating tls certificates functionality for painless integration of agones to umbrella charts.

Closes #2389 

Helm upgrades was tested on microk8s using command
```
microk8s.helm3 upgrade --install agones . --set agones.ping.install=false --set agones.allocator.install=false
```
with various state transitions including:
- changing from reusing and not reusing generated tls certs
- disabling generation of tls and then reusing certs provided from file
and checking whether allocation and fleet status check (force admission controller to run and use certs for requests) is working

I think allocator also needs this feature but it would be a little bit more complex.

### DIFF
--- a/install/helm/agones/templates/extensions.yaml
+++ b/install/helm/agones/templates/extensions.yaml
@@ -11,11 +11,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-{{- $ca := genCA "admission-controller-ca" 3650 }}
-{{- $cn := printf "agones-controller-service" }}
-{{- $altName1 := printf "agones-controller-service.%s"  .Release.Namespace }}
-{{- $altName2 := printf "agones-controller-service.%s.svc" .Release.Namespace }}
-{{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca }}
+{{- $certCert := "" }}
+{{- $certKey := "" }}
+{{- if .Values.agones.controller.generateTLS }}
+  {{- $previous := lookup "v1" "Secret" .Release.Namespace (include "agones.fullname" . | printf "%s-cert") }}
+  {{- if $previous | and .Values.agones.controller.reuseGeneratedTLS }}
+    {{- $certCert = index $previous.data "server.crt" | b64dec }}
+    {{- $certKey = index $previous.data "server.key" | b64dec }}
+  {{- else }}
+    {{- $ca := genCA "admission-controller-ca" 3650 }}
+    {{- $cn := printf "agones-controller-service" }}
+    {{- $altName1 := printf "agones-controller-service.%s"  .Release.Namespace }}
+    {{- $altName2 := printf "agones-controller-service.%s.svc" .Release.Namespace }}
+    {{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca }}
+    {{- $certCert = $cert.Cert }}
+    {{- $certKey = $cert.Key }}
+  {{- end }}
+{{- end }}
 ---
 {{- if .Values.agones.registerApiService }}
 apiVersion: apiregistration.k8s.io/v1
@@ -35,11 +47,11 @@ spec:
   service:
     name: agones-controller-service
     namespace: {{ .Release.Namespace }}
-        {{- if .Values.agones.controller.generateTLS }}
-  caBundle: {{ b64enc $ca.Cert }}
-        {{- else }}
+{{- if .Values.agones.controller.generateTLS }}
+  caBundle: {{ b64enc $certCert }}
+{{- else }}
   caBundle: {{ default (.Files.Get "certs/server.crt") .Values.agones.controller.tlsCert | b64enc }}
-        {{- end }}
+{{- end }}
   version: v1
 {{- end}}
 {{- if .Values.agones.registerWebhooks }}
@@ -66,7 +78,7 @@ webhooks:
         namespace: {{ .Release.Namespace }}
         path: /validate
 {{- if .Values.agones.controller.generateTLS }}
-      caBundle: {{ b64enc $ca.Cert }}
+      caBundle: {{ b64enc $certCert }}
 {{- else }}
       caBundle: {{ default (.Files.Get "certs/server.crt") .Values.agones.controller.tlsCert | b64enc }}
 {{- end }}
@@ -123,7 +135,7 @@ webhooks:
         namespace: {{ .Release.Namespace }}
         path: /mutate
 {{- if .Values.agones.controller.generateTLS }}
-      caBundle: {{ b64enc $ca.Cert }}
+      caBundle: {{ b64enc $certCert }}
 {{- else }}
       caBundle: {{ default (.Files.Get "certs/server.crt") .Values.agones.controller.tlsCert | b64enc }}
 {{- end }}
@@ -161,8 +173,8 @@ metadata:
 type: Opaque
 data:
 {{- if .Values.agones.controller.generateTLS }}
-  server.crt: {{ b64enc $cert.Cert }}
-  server.key: {{ b64enc $cert.Key }}
+  server.crt: {{ b64enc $certCert }}
+  server.key: {{ b64enc $certKey }}
 {{- else }}
   server.crt: {{ default (.Files.Get "certs/server.crt") .Values.agones.controller.tlsCert | b64enc }}
   server.key: {{ default (.Files.Get "certs/server.key") .Values.agones.controller.tlsKey | b64enc }}
--- a/install/helm/agones/templates/extensions.yaml
+++ b/install/helm/agones/templates/extensions.yaml
@@ -11,21 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-{{- $certCert := "" }}
-{{- $certKey := "" }}
+{{- $cert := "" }}
+{{- $key := "" }}
 {{- if .Values.agones.controller.generateTLS }}
   {{- $previous := lookup "v1" "Secret" .Release.Namespace (include "agones.fullname" . | printf "%s-cert") }}
   {{- if $previous | and .Values.agones.controller.reuseGeneratedTLS }}
-    {{- $certCert = index $previous.data "server.crt" | b64dec }}
-    {{- $certKey = index $previous.data "server.key" | b64dec }}
+    {{- $cert = index $previous.data "server.crt" | b64dec }}
+    {{- $key = index $previous.data "server.key" | b64dec }}
   {{- else }}
-    {{- $ca := genCA "admission-controller-ca" 3650 }}
-    {{- $cn := printf "agones-controller-service" }}
+    {{- $cn := printf "agones-controller-service.%s.svc" .Release.Namespace }}
     {{- $altName1 := printf "agones-controller-service.%s"  .Release.Namespace }}
     {{- $altName2 := printf "agones-controller-service.%s.svc" .Release.Namespace }}
-    {{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca }}
-    {{- $certCert = $cert.Cert }}
-    {{- $certKey = $cert.Key }}
+    {{- $certObj := genSelfSignedCert $cn nil (list $altName1 $altName2) 3650 }}
+    {{- $cert = $certObj.Cert }}
+    {{- $key = $certObj.Key }}
   {{- end }}
 {{- end }}
 ---
@@ -48,7 +47,7 @@ spec:
     name: agones-controller-service
     namespace: {{ .Release.Namespace }}
 {{- if .Values.agones.controller.generateTLS }}
-  caBundle: {{ b64enc $certCert }}
+  caBundle: {{ b64enc $cert }}
 {{- else }}
   caBundle: {{ default (.Files.Get "certs/server.crt") .Values.agones.controller.tlsCert | b64enc }}
 {{- end }}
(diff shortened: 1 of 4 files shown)


roberthbailey:
@gorylpe - Can you please sign the CLA? Once that check is passing I'll be happy to review your PR.

gorylpe:
@roberthbailey Done. I'm sorry, I'm catching up with all docs, didn't know what to expect as I am rookie in open source

roberthbailey:
No worries. We just can't merge any PRs without the CLA signed so if that check fails it's always the first question we have to ask. :)

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -120,6 +120,7 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `agones.controller.healthCheck.timeoutSeconds`      | Number of seconds after which the probe times out (in seconds)                                  | `1`                    |
 | `agones.controller.resources`                       | Controller [resource requests/limit][resources]                                                 | `{}`                   |
 | `agones.controller.generateTLS`                     | Set to true to generate TLS certificates or false to provide your own certificates              | `true`                 |
+| `agones.controller.reuseGeneratedTLS`               | Set to true to reuse generated TLS certificates with lookup existing secrets                    | `true`                 |
```
roberthbailey, line 123: For newly added config params, please add them to the table **New Configuration Features** (down around line 211). This way they won't be published to the website until the next release, when the helm chart will be published that supports using them.

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -275,8 +276,10 @@ That means that you skipped the `--cleanup` flag and you should either delete th
 
 ## Controller TLS Certificates
 
-By default agones chart generates tls certificates used by the admission controller, while this is handy, it requires the agones controller to restart on each `helm upgrade` command.
-For most use cases the controller would have required a restart anyway (eg: controller image updated). However if you really need to avoid restarts we suggest that you turn off tls automatic generation (`agones.controller.generateTLS` to `false`) and provide your own certificates (`certs/server.crt`,`certs/server.key`).
+By default agones chart generates tls certificates used by the admission controller and reuses them on each `helm upgrade` command using `lookup` templating function.
```
roberthbailey, line 279: Please wrap this in a feature shortcode (see https://agones.dev/site/docs/contribute/#within-a-page) so that the old text is still published until the next release, when this behavior becomes the new default.

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -275,8 +276,10 @@ That means that you skipped the `--cleanup` flag and you should either delete th
...
+This prevents agones controller from restart during update of values not related to controller itself (especially when using agones as a subchart dependency in umbrella charts).
+To turn off this feature and generate tls certificates on each `helm upgrade` command (e.g. to force integrity between versions) set `agones.controller.reuseGeneratedTLS` to `false`.
+Another approach to avoid restarts and having more control we suggest that you turn off tls automatic generation (`agones.controller.generateTLS` to `false`) and provide your own certificates (`certs/server.crt`,`certs/server.key`).
```
roberthbailey, line 282: nit: "Another approach to avoid the controller restarting during an upgrade is to turn off automatic TLS certificate generation altogether (`agones.controller.generateTLS` to `false`) and provide your own certificates (`certs/server.crt`,`certs/server.key`).

install/helm/agones/templates/extensions.yaml
```diff
@@ -35,11 +47,11 @@ spec:
   service:
     name: agones-controller-service
     namespace: {{ .Release.Namespace }}
-        {{- if .Values.agones.controller.generateTLS }}
-  caBundle: {{ b64enc $ca.Cert }}
-        {{- else }}
+{{- if .Values.agones.controller.generateTLS }}
+  caBundle: {{ b64enc $certCert }}
```
roberthbailey, line 51: Before we were using the cert of the ca here (`ca.Cert`) and I think this is using the cert that was generated instead (`cert.Cert`). Shouldn't we be putting the cert from the certificate authority here and not the cert generated by the certificate authority?

Same for below on lines 81 and 126, whereas line 176 seems consistent with before.

roberthbailey (commented):
@gorylpe - sorry for the delay in getting this reviewed, but I've now left some comments for you.

Our next release candidate is scheduled for next week on Tuesday, so if we want to get this merged in time for the upcoming release we will need it approved and merged prior to that.

site/content/en/docs/Installation/Install Agones/helm.md
markmandel: This also seems like a change in behaviour - i.e. we currently regen certs, and this means that it won't.

I would advocate for a default of `false`, so that there isn't a breaking change in behaviour in the helm charts.

gorylpe: Okay, working on it

install/helm/agones/templates/extensions.yaml
gorylpe: When not generating certs, there was used `server.crt`
```
caBundle: {{ default (.Files.Get "certs/server.crt") .Values.agones.controller.tlsCert | b64enc }}
```
which is not ca.cert, but cert.Cert equivalent. And this is when ambiguity happen.
`server.crt` and `server.key` are self signed cert, which can be, as far as I know, CA for itself. But the generated certs were using separated CA. I would prefer to maintain `Secret` clean, and not creating additional branches to it, so there is 2 possible ways I can follow:
- changing script for `server.key` and `server.crt` to also generate `ca.crt`
- generating self signed cert and use it as caBundle - `genSelfSignedCert` helm func
Is there any preference you would like me to implement?

site/content/en/docs/Installation/Install Agones/helm.md
gorylpe: done

site/content/en/docs/Installation/Install Agones/helm.md
gorylpe: done

site/content/en/docs/Installation/Install Agones/helm.md
gorylpe: done

install/helm/agones/templates/extensions.yaml
gorylpe: I used second approach

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -275,12 +276,24 @@ That means that you skipped the `--cleanup` flag and you should either delete th
 
 ## Controller TLS Certificates
 
+{{% feature expiryVersion="1.21.0" %}}
```
roberthbailey, line 279: Since we didn't merge this PR before the last release, this needs to be updated to the next release (1.22).

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -275,12 +276,24 @@ That means that you skipped the `--cleanup` flag and you should either delete th
...
 By default agones chart generates tls certificates used by the admission controller, while this is handy, it requires the agones controller to restart on each `helm upgrade` command.
 For most use cases the controller would have required a restart anyway (eg: controller image updated). However if you really need to avoid restarts we suggest that you turn off tls automatic generation (`agones.controller.generateTLS` to `false`) and provide your own certificates (`certs/server.crt`,`certs/server.key`).
 
 {{< alert title="Tip" color="info">}}
 You can use our script located at {{< ghlink href="install/helm/agones/certs/cert.sh" >}}cert.sh{{< /ghlink >}} to generate them.
 {{< /alert >}}
+{{% /feature %}}
+{{% feature publishVersion="1.21.0" %}}
```
roberthbailey, line 287: This one also needs to be updated to 1.22.0

install/helm/agones/templates/extensions.yaml
```diff
@@ -11,11 +11,22 @@
...
-{{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca }}
+{{- $cert := "" }}
+{{- $key := "" }}
+{{- if .Values.agones.controller.generateTLS }}
+  {{- $previous := lookup "v1" "Secret" .Release.Namespace (include "agones.fullname" . | printf "%s-cert") }}
+  {{- if $previous | and .Values.agones.controller.reuseGeneratedTLS }}
+    {{- $cert = index $previous.data "server.crt" | b64dec }}
+    {{- $key = index $previous.data "server.key" | b64dec }}
+  {{- else }}
+    {{- $cn := printf "agones-controller-service.%s.svc" .Release.Namespace }}
```
roberthbailey, line 22: Shouldn't this still be `agones-controller-service` (as it was before)? Right now the common name is the same as the altName2.

install/helm/agones/templates/extensions.yaml
```diff
@@ -11,11 +11,22 @@
...
+    {{- $altName1 := printf "agones-controller-service.%s"  .Release.Namespace }}
+    {{- $altName2 := printf "agones-controller-service.%s.svc" .Release.Namespace }}
+    {{- $certObj := genSelfSignedCert $cn nil (list $altName1 $altName2) 3650 }}
```
roberthbailey, line 25: Maybe you already answered this (it's been a while since I reviewed this PR), but can you explain why this isn't using the `genCA` command any longer?

To @markmandel's earlier review comment - ideally after this change nothing looks different than it did before _unless_ you set the new helm value. From what I can tell here, this changes the way that certs are generated even if you _don't_ set the new value to true.

install/helm/agones/templates/extensions.yaml
gorylpe: https://kubernetes.io/docs/tasks/extend-kubernetes/setup-extension-api-server/

`5. Create a server cert/key for the api-server to use for HTTPS. This cert should be signed by the above CA. It should also have a CN of the Kube DNS name. This is derived from the Kubernetes service and be of the form <service name>.<service name namespace>.svc`

And it is also said inside file generating certs `install/helm/agones/certs/cert.sh` - `echo "Common Name should be: agones-controller-service.agones-system.svc"`

install/helm/agones/templates/extensions.yaml
gorylpe: Like I said in this comment (https://github.com/googleforgames/agones/pull/2417#discussion_r798486250) - there is difference with approach in this helm file and `install/helm/agones/certs/cert.sh`.
 
After my change nothing looks different in terms of how parts of this helm chart works. 
The only difference is that I use self signed certificate feature (just like certificate generated by `install/helm/agones/certs/cert.sh`) that it can be its own certification authority - that's why it is used in place of `caBundle`.  This change also unify usage of certs generated by helm and by mentioned above script.

markmandel:
I was reviewing #2453 and it seemed like a more elegant solution to your issue? (basically, with those changes, you can control the Secret and can manage your own cert creation / destruction). I _think_ it gets you to the same place, with a bit more work on your end maybe?

I am also questioning the requirements here for this functionality, so if you will indulge my curiosity, I'd like to dig deeper into the use case before we add this additional complexity to the installation experience.

In my mind, the ideal path would be that you would install Agones one time along with the cluster creation (probably all via terraform). This would create the _development platform_ that you would then host and run your game servers on.

From there, you could then apply your `Fleets` - as your application deployment through a seperate system (e.g. CI/CD pipeline).

Combining Fleet updates with Agones updates seems at cross-purposes in my mind. I'm trying to envisage a scenario in which I would want to both update Agones _and_ update a Fleet at the same time in a production environment.

The only exception I could possibly see is in a dev environment (or similar), you wanted to add / remove namespaces that Agones could support. And while Helm is handy for updating that, if you wanted to manage it without a short Agones downtime, you could script adding/removing the correct RBAC permissions as needed.

Would love to get your thoughts on the above though - it's entirely possible I'm missing/misinterpreting a use case here.

gorylpe:
Okay, so basically use case I had in my mind is to attach Agones as dependency to our helm chart. The reason I want to do this is to stick Agones version with one of our apps which uses Kubernetes API to access Agones. And when we update our helm with new minor version I do not want Agones to restart, its version won't change as often as our app. 

But right now I realized that I cannot do this, because, after rethinking installation by terraform, Agones also install CRD, and helm doesn't provide any way to define order of installing dependencies therefore my use case is invalid I think.

So, for the Agones independent installation, the #2453 seems much more elegant

gorylpe:
It can be closed closed right now, because my use case is not entirely valid, therefore #2453 solution will be more accurate

markmandel:
Sounds good. Closing!

PR #2436, opened 2022-01-19
Remove curl|bash from dockerfile to address vulnerability issues

cindy52:
Files affected are: 
```
build/build-sdk-images/node/Dockerfile,
build/build-sdk-images/restapi/Dockerfile,
build/build-image/Dockerfile,
examples/nodejs-simple/Dockerfile
```

**What type of PR is this?**
>
> /kind breaking
> /kind bug

 /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Resolve the unvendored shell script is being executed using curl | bash.


build/includes/sdk.mk
```diff
@@ -103,7 +106,8 @@ build-build-sdk-image-base:
 # Builds the docker image used by commands for a specific sdk
 build-build-sdk-image: DOCKER_BUILD_ARGS= --build-arg BASE_IMAGE=$(build_sdk_base_tag)
 build-build-sdk-image: ensure-build-sdk-image-base
-		docker build --tag=$(SDK_IMAGE_TAG) $(build_path)build-sdk-images/$(SDK_FOLDER) $(DOCKER_BUILD_ARGS)
+		cd $(root_path) &&  docker build -f $(build_path)build-sdk-images/$(SDK_FOLDER)/Dockerfile $(DOCKER_BUILD_ARGS) --tag=$(SDK_IMAGE_TAG) .
+		# docker build --tag=$(SDK_IMAGE_TAG) $(build_path)build-sdk-images/$(SDK_FOLDER) $(DOCKER_BUILD_ARGS)
```
markmandel, line 110: Nit: cleanup the commented code please 😃

build/build-image/Dockerfile
```diff
@@ -84,8 +86,9 @@ RUN mkdir /tmp/hugo && \
     mv /tmp/hugo/hugo /usr/local/bin/ && \
     rm -r /tmp/hugo
 
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
-    apt-get install -y nodejs
+COPY /examples/nodejs-simple .
+RUN tar xvf node-v12.16.1-linux-x64.tar.gz
```
markmandel, line 90: Curious - any reason we're vendoring, and not pulling from a central website?

build/build-image/Dockerfile
```diff
@@ -68,7 +69,8 @@ RUN curl -L  ${HELM_URL} > /tmp/helm.tar.gz \
 RUN echo "source <(helm completion bash)" >> /root/.bashrc
 
 # install golang-ci linter
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOPATH/bin v1.42.1
+RUN go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0
+# RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOPATH/bin v1.42.1
```
markmandel, line 73: Nit: remove commented out instructions.

build/build-image/Dockerfile
```diff
@@ -39,7 +39,8 @@ RUN wget -q https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.zip && un
 ENV PATH /usr/local/go/bin:/go/bin:/opt/google-cloud-sdk/bin:$PATH
 
 # install go tooling for development, building and testing
-RUN go get -u golang.org/x/tools/cmd/goimports
+RUN go install golang.org/x/tools/cmd/goimports@latest
+# RUN go get -u golang.org/x/tools/cmd/goimports
```
markmandel, line 43: Nit: remove commented out instructions.

build/build-image/Dockerfile
```diff
@@ -108,7 +111,7 @@ RUN echo "export EDITOR=nano" >> /root/.bashrc
 RUN wget -nv -O terraform.zip https://releases.hashicorp.com/terraform/1.0.8/terraform_1.0.8_linux_amd64.zip && unzip ./terraform.zip && mv terraform /usr/local/bin/
 
 # code generation scripts
-COPY *.sh /root/
+COPY /scripts/build-image/*.sh /root/
```
markmandel, line 114: 🤔 

1. What's the impetus to move all the scripts into a seperate folder away from the Dockerfile? I figured having them next to each other made discovery easier? (now I need to go to a different folder, match the name of this folder to it's subfolder, and then work from there - which is a tad tricky).
2. If we do have a ./scripts/ folder, rather than having it sit as a new top level folder, it should probably sit within `/build` since that houses all our build scripts and related.

cindy52: I move them to the top level scripts folder because the docker file needs node package in example folder. I can't put the node package in build folder because the build folder is excluded from dockerfile at the example folder level. Right now I changed the make file in build/includes to cd to the agones repo to get a copy of node package, thus I need to move the script files in build folder to top level to avoid it excluded from docker. Let me know if you have better locations to put them. Thanks!

build/build-image/Dockerfile
cindy52: The security folks recommend to directly checkin the package to avoid the curl|bash vulnerability.

build/build-sdk-images/node/Dockerfile
```diff
@@ -14,14 +14,13 @@
 ARG BASE_IMAGE=agones-build-sdk-base:latest
 FROM $BASE_IMAGE
 
-## --allow-releaseinfo-change is a workaround for the new debian release
-## Can be removed when upgrading to bullseye
-RUN apt-get --allow-releaseinfo-change update && curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
-    apt-get install -y nodejs
+COPY ./examples/nodejs-simple .
```
steven-supersolid, line 17: I'm not sure why we do this here, should not be required to build the SDK or are we running the simple example as a kind of test?

examples/nodejs-simple/Dockerfile
```diff
@@ -14,17 +14,14 @@
...
 RUN useradd -u 1000 -m server
-RUN apt-get update && apt-get install -y curl  && apt-get clean
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
-    apt-get install -y nodejs
 
-WORKDIR /home/server
+COPY ./examples/nodejs-simple /home/server/examples/nodejs-simple
+WORKDIR /home/server/examples/nodejs-simple
+RUN tar xvf node-v12.16.1-linux-x64.tar.gz
+RUN cd node-v12.16.1-linux-x64 && ln -s $(pwd)/bin/node /usr/bin/node && ln -s $(pwd)/bin/npm /usr/bin/npm
```
steven-supersolid, line 21: I'm not 100% sure but would this install node under the workdir

examples/nodejs-simple/Dockerfile
```diff
@@ -14,17 +14,14 @@
 
 FROM debian:buster
 RUN useradd -u 1000 -m server
-RUN apt-get update && apt-get install -y curl  && apt-get clean
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
-    apt-get install -y nodejs
 
-WORKDIR /home/server
+COPY ./examples/nodejs-simple /home/server/examples/nodejs-simple
+WORKDIR /home/server/examples/nodejs-simple
```
steven-supersolid, line 19: Minor but if this was moved above the line above could the destination path be simplified in the `COPY` to a relative path?

build/build-image/Dockerfile
cindy52: Done

build/build-image/Dockerfile
cindy52: Done.

build/build-image/Dockerfile
markmandel: Where did node-v12.16.1-linux-x64.tar.gz come from - could we not download it from there?

I'm being cognizant of the fact that this repo (for better or worse), gets pulled in as a dependency for a go, rust, and terraform (probably others), and adding an extra 20MB to the repo for each version of node we want to vendor seems like something we should avoid?

build/build-image/Dockerfile
markmandel: > I move them to the top level scripts folder because the docker file needs node package in example folder

I still feel like we should be able to download this from somewhere, and then this goes away?

If the scripts were shared between images, I would get having some sort of `common` directory that each image could pull from, but we don't really do that anywhere - so this aspect is still confusing me.

It's already a bit complicated that you have to match `build/build-sdk-images/<lang>` to `sdk/<lang>` appropriately, but now we're adding another layer to that of `build/scripts/<lang>` - and I'm still not sure of the value add in doing so.

build/build-sdk-images/node/Dockerfile
```diff
@@ -14,14 +14,13 @@
...
+RUN tar xvf node-v12.16.1-linux-x64.tar.gz
+RUN cd node-v12.16.1-linux-x64 && ln -s $(pwd)/bin/node /usr/bin/node && ln -s $(pwd)/bin/npm /usr/bin/npm
```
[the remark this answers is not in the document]
cindy52: Thanks for the advise. I've changed the pr to use the manual installation instead.

cindy52:
@markmandel , @steven-supersolid , I have changed the way to install nodejs with apt-get but no bash as the manual installation. Please refer to #2438 instead. Closing it for now. Thanks!

PR #2409, opened 2021-12-22
Update the simple game server to move itself back to the Ready state after allocation

roberthbailey:
This allows us to test the "Reusing Allocated GameServers" integration pattern.

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

**What this PR does / Why we need it**:

**Which issue(s) this PR fixes**:

Closes #

**Special notes for your reviewer**:

PR #2440, opened 2022-01-20
Get the initial game server state before marking the simple game server ready

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Closes #

If you are allocating the simple game servers at a high rate, then you can move from `Scheduled` -> `Ready` -> `Allocated` between the call to `ready(s)` and the call in `shutdownAfterNAllocations` to get the initial game server state. This was always the case (and testing found that even in older versions of the simple game server using the `automaticShutdownDelayMin` flag some got stuck forever), but since the watch function tends to fire multiple times the game servers often eventually received the signal to exit. This got worse with [my recent change](https://github.com/googleforgames/agones/pull/2409) where I'm explicitly looking for the transition from `Ready` to `Allocated` (to count the number of times it happens rather than the number of times we get a callback with the Allocated state, which isn't accurate). Now, if that initial transition to `Allocated` is missed the game server will _never_ decide to exit (or move back to ready), leaving it stuck. 

Running another test with this code left zero stuck game servers in the `Allocated` state, so the race condition appears to be resolved.

### DIFF
--- a/examples/simple-game-server/Makefile
+++ b/examples/simple-game-server/Makefile
@@ -35,7 +35,7 @@ WITH_WINDOWS=1
 
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 project_path := $(dir $(mkfile_path))
-server_tag = $(REGISTRY)/simple-game-server:0.7
+server_tag = $(REGISTRY)/simple-game-server:0.8
 ifeq ($(WITH_WINDOWS), 1)
 server_tag_linux_amd64 = $(server_tag)-linux_amd64
 else
--- a/examples/simple-game-server/main.go
+++ b/examples/simple-game-server/main.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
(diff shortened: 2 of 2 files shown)


roberthbailey:
sdk-conformance:

```
make[2]: *** [includes/sdk.mk:145: run-sdk-conformance-no-build] Error 1
make[1]: *** [includes/sdk.mk:153: run-sdk-conformance-test] Error 2
make: *** [includes/sdk.mk:156: run-sdk-conformance-test-cpp] Error 2
```

ilkercelikyilmaz (approved):
LGTM

markmandel:
My only comment on this would be it would good in the PR description and commit to describe what the issue is that's being resolved. I'm assuming there is a race condition somewhere - I think it would be good to outline what it was, and how it is solved in the new code, so when we look at it 3 months down the line, we remember why we made the change  (my memory is awful, not sure how good yours is 😄).

roberthbailey:
> I'm assuming there is a race condition somewhere

As far as I can tell there is. I'll update the PR description with what I saw.

PR #2438, opened 2022-01-19
Remove curl|bash from dockerfile to address vulnerability issues

cindy52:
Files affected are:
build/build-sdk-images/node/Dockerfile,
build/build-sdk-images/restapi/Dockerfile,
build/build-image/Dockerfile,
examples/nodejs-simple/Dockerfile

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

 /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**:
Resolve the unvendored shell script is being executed using curl | bash.

PR #2444, opened 2022-01-21
Fix connection timeout on Rust SDK.

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

This is a fix for the Rust SDK flakiness!

The previous iteration of the code would fail as soon as the initial connection would fail at any timeout greater than 2 seconds.

We move the connection to being lazy, so that the connection is only attempted on first attempt, so we can retry connecting if there is a failure for up to 30 seconds.

N/A

This can either be merged before #2438, or if #2438 gets merged first, I'll need to remove the workaround as part of the PR before this gets merged.

### DIFF
--- a/examples/rust-simple/Makefile
+++ b/examples/rust-simple/Makefile
@@ -27,7 +27,7 @@ REPOSITORY ?= gcr.io/agones-images
 
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 project_path := $(dir $(mkfile_path))
-server_tag = $(REPOSITORY)/rust-simple-server:0.9
+server_tag = $(REPOSITORY)/rust-simple-server:0.10
 
 #   _____                    _
 #  |_   _|_ _ _ __ __ _  ___| |_ ___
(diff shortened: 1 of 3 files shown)


markmandel:
@Jake-Shadle would you mind reviewing my fix to the Rust SDK?

roberthbailey:
I don't know much rust but this seems like a reasonable change. One question is whether we should also update https://github.com/googleforgames/agones/tree/main/examples/rust-simple to a new version to pick up the SDK fixes.

Jake-Shadle:
LGTM!

markmandel:
> I don't know much rust but this seems like a reasonable change. One question is whether we should also update https://github.com/googleforgames/agones/tree/main/examples/rust-simple to a new version to pick up the SDK fixes.

That's an excellent point! I will add that change!

markmandel:
This should be good to go now. I'll let @roberthbailey decide which order to merge things in 😄

roberthbailey (approved):

PR #2451, opened 2022-01-25
Update the simple game server to avoid a race condition when transitioning from ready to allocated

roberthbailey:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The code previously assumed that each state transition for the game server could be observed via callbacks `WatchGameServer` which via some testing has been observed to not be true. 

Instead of looking for the changes, look for the new state. This changes the code from being edge triggered to level triggered, as per https://speakerdeck.com/thockin/edge-vs-level-triggered-logic

### DIFF
--- a/examples/simple-game-server/main.go
+++ b/examples/simple-game-server/main.go
@@ -153,15 +153,17 @@ func shutdownAfterNAllocations(s *sdk.SDK, readyIterations, shutdownDelaySec int
 	log.Printf("Initial game Server state = %s", gs.Status.State)
 
 	m := sync.Mutex{} // protects the following two variables
-	currentState := gs.Status.State
+	lastAllocated := gs.ObjectMeta.Annotations["agones.dev/last-allocated"]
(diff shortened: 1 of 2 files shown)


examples/simple-game-server/main.go
```diff
@@ -153,15 +153,17 @@ func shutdownAfterNAllocations(s *sdk.SDK, readyIterations, shutdownDelaySec int
 	log.Printf("Initial game Server state = %s", gs.Status.State)
 
 	m := sync.Mutex{} // protects the following two variables
-	currentState := gs.Status.State
+	lastAllocated := gs.ObjectMeta.Annotations["agones.dev/last-allocated"]
```
markmandel, line 156: Nice! That does make more sense 👍🏻

markmandel (approved):
LGTM!

PR #2476, opened 2022-02-12
Add NodePort to the helm template for agones-allocator service

### LINKED ISSUE #1699: Allow specifying agones-allocator nodePort via Helm values
We use minikube within a Vagrant VM for development. The main server could be launched locally (on host machine), which requires that services within the VM are accessible from the host machine. minikube doesn't support LoadBalancer services, so we have to specify NodePort for allocator-service. However, by default it allocates a random port, while it would be convenient to have it fixed. It is possible to do that in Kubernetes via `nodePort` option, but it is not exposed in the Helm chart.

Expose `nodePort` option for allocator service which would be used if `serviceType` is `NodePort`.

jiwonaid:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

To support type: NodePort for agones dev environment.

Closes #1699

Hello, this is my second pull request.

May I freely select between normal issues and `good first issue`? Is it okay to solve multiple good first issues?

### DIFF
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -164,11 +164,13 @@ agones:
         port: 443
         portName: https
         targetPort: 8443
+        nodePort: 0 # nodePort will be used if the serviceType is set to NodePort
       grpc:
         enabled: true
         port: 443
         portName: grpc
         targetPort: 8443
+        nodePort: 0 # nodePort will be used if the serviceType is set to NodePort
     serviceMetrics:
       name: agones-allocator-metrics-service
       annotations: {}
--- a/install/helm/agones/templates/service/allocation.yaml
+++ b/install/helm/agones/templates/service/allocation.yaml
@@ -38,12 +38,18 @@ spec:
     - port: {{ .Values.agones.allocator.service.http.port }}
       name: {{ .Values.agones.allocator.service.http.portName }}
       targetPort: {{ .Values.agones.allocator.service.http.targetPort }}
+      {{- if eq .Values.agones.allocator.service.serviceType "NodePort" }}
+      nodePort: {{ .Values.agones.allocator.service.http.nodePort }}
+      {{- end }}
       protocol: TCP
 {{- if .Values.agones.allocator.service.grpc.enabled }}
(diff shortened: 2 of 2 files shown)


jiwonaid:
> This seems like a straightforward extension to the helm chart, if someone would like to contribute a PR to implement it I'd be happy to review it.

/cc @roberthbailey

markmandel:
Looks like a flake, will run again when it gets fixed.

install/helm/agones/values.yaml
```diff
@@ -164,11 +164,13 @@ agones:
         port: 443
         portName: https
         targetPort: 8443
+        nodePort: 0
```
roberthbailey, line 167: Two thoughts about this.

1. Setting this to 0 seems like a bad default. If we are going to set it to an example value, it seems like it should be a port that would actually work if you just changed the `serviceType` above.
2. This file doesn't really have many comments, but it seems like if we are going to include a value here it is worth adding a comment that it will only be used if the service type is set to NodePort.

jiwonaid: Thank you for the feedback.

1. I think it is okay to use 0. When I type `kubectl create service nodeport --help` The nodePort parameter is described like this.

```bash
(docker-desktop) ☁  ~  kubectl create service nodeport --help
--node-port=0: Port used to expose the service on each node in a cluster.
```

I think this is because v1.ServiceSpec.ServicePort is using int32.

https://github.com/kubernetes/api/blob/b8c40e080bc5e830097df540d4ef804034cb5bdb/core/v1/types.go#L4536

I found some examples that using `NodePort: 0` in the test code of `kubectl` with `expectErr: false`.

https://github.com/kubernetes/kubectl/blob/c26d922e2ead95fb4850c73c3c41027da40f3e0e/pkg/cmd/create/create_service_test.go#L204-L221

It is a good idea to use an example NodePort with the default range of 30000-32767. However, it could make an error if the port is already taken by another Service.

2. Comments are added right behind the nodePort.

roberthbailey:
Failed on `push-images`:

```
received unexpected HTTP status: 503 Service Unavailable
make: *** [Makefile:442: push-ping-image] Error 1
```

roberthbailey (approved):

PR #2489, opened 2022-02-22
Fix panic when playertracking is false

jiwonaid:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

```markdown
| "PLAYER_CAPACITY"   | With one argument, gets the player capacity; with two arguments sets the player capacity |
| "PLAYER_CONNECT"    | Connects the specified player to the game server                                         |
| "PLAYER_DISCONNECT" | Disconnects the specified player from the game server                                    |
| "PLAYER_CONNECTED"  | Returns true/false depending on whether the specified player is connected                |
| "GET_PLAYERS"       | Returns a list of the connected players                                                  |
| "PLAYER_COUNT"      | Returns a count of the connected players                                                 |
```

When the FeatureFlag:PlayerTracking is false(default), all the commands above emit panic.
This is because alpha is using protobuf/proto. So, the values would be better to access with auto-generated Get() methods.

https://github.com/googleforgames/agones/blob/94128b43efa027378037cbdec81db15cfa5e3833/pkg/sdk/alpha/alpha.pb.go#L109-L114

Work on #1957

### DIFF
--- a/sdks/go/alpha.go
+++ b/sdks/go/alpha.go
@@ -40,7 +40,7 @@ func newAlpha(conn *grpc.ClientConn) *Alpha {
 // If the player capacity is set from outside the SDK, use SDK.GameServer() instead.
 func (a *Alpha) GetPlayerCapacity() (int64, error) {
 	c, err := a.client.GetPlayerCapacity(context.Background(), &alpha.Empty{})
-	return c.Count, errors.Wrap(err, "could not get player capacity")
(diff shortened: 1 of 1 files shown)


markmandel (approved):
Excellent change! Should have been this way in the first place. 👍🏻

PR #2505, opened 2022-03-08
static TCPUDP support for the container ports

### LINKED ISSUE #2314: TCPUDP not supported
**What happened**:

Pod "gameserver" is invalid: spec.containers[0].ports[0].protocol: Unsupported value: "TCPUDP": supported values: "SCTP", "TCP", "UDP"

```
apiVersion: agones.dev/v1
kind: GameServer
spec:
  ports:
     portPolicy: Static
     containerPort: 7000
     hostPort: 7000
     protocol: TCPUDP
```

**Environment**:
- Agones version: v1.18.0
- Kubernetes version (use `kubectl version`): v1.22.1

SaitejaTamma:
**What type of PR is this?**
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Closes #2314


markmandel:
Failure on linting. Gets us all. Run `make lint` to run lints locally on your code.

pkg/gameservers/controller.go
```diff
@@ -474,6 +477,35 @@ func (c *Controller) syncGameServerPortAllocationState(ctx context.Context, gs *
 	return gs, nil
 }
 
+// Allocate gameserver ports ( containerports ) for dynamic protocols such as TCPUDP
+// TODO: Going forward, this needs to be enhanced for other kinds of dynamic protocols
+func handleTCPUDPPortAllocationforStaticPortPolicy(gs *agonesv1.GameServer) {
```
markmandel, line 482: I thought we were going to move this logic into `GameServer.Pod(...)`?

https://github.com/googleforgames/agones/issues/2314#issuecomment-1055714041

SaitejaTamma: The logic for TCPUDP port allocation for static port policy is encapsulated in the method “**handleTCPUDPPortAllocationforStaticPortPolicy”**. This method is invoked from **“syncGameServerPortAllocationState”**. Since this is an explicit port allocation action this needs to happen before the event “port  allocated” is recorded, **_“ c.recorder.Event(gs, corev1.EventTypeNormal, string(gs.Status.State), "Port allocated") “_**

markmandel: Got it. That makes a certain amount of sense.

I'm not 100% sure, but is there a way to move this duplication of port logic into some kind of centralised function(s) that are shared between the two implementations? Feels like a lot of duplication.

But I can't see it myself, but I thought I'd ask the question?

pkg/gameservers/controller.go
```diff
@@ -474,6 +477,35 @@ func (c *Controller) syncGameServerPortAllocationState(ctx context.Context, gs *
 	return gs, nil
 }
 
+// Allocate gameserver ports ( containerports ) for dynamic protocols such as TCPUDP
+// TODO: Going forward, this needs to be enhanced for other kinds of dynamic protocols
```
markmandel, line 481: Not sure I follow this TODO? I feel like the comment should just be something like "Creates duplicate ports for handling TCPUDP protocol with static port policies" - or something simple like that? There isn't dynamic protocols in this?

pkg/gameservers/controller.go
markmandel, line 482:
```suggestion
func assignStaticTCPUDPPorts(gs *agonesv1.GameServer) {
```
Other options from "assign": "apply", "create" ? "handle" feels generic, but we don't use it anywhere else either.

I feel like we aren't "allocating" a port, since it's static - we are just assigning existing static ports. WDYT?

markmandel:
Another thought I had this morning, was should we have an e2e test for this so that we check actual connectivity?

The only tricky part is choosing a static port that otherwise is not used, we should have a look to see if there are other e2e tests for static ports? (I'm not sure if there are?)

Issue #2479, opened 2022-02-14
Flaky:  TestFleetRecreateGameServers/gameserver_shutdown


markmandel:
**What happened**:

It fails sometimes!

```
--- FAIL: TestFleetRecreateGameServers (0.00s)
    --- FAIL: TestFleetRecreateGameServers/gameserver_shutdown (89.80s)
        fleet_test.go:1160: Could not message GameServer: timed out attempting to send UDP packet to address: timed out waiting for the condition
```

**What you expected to happen**:

It to pass!

**How to reproduce it (as minimally and precisely as possible)**:

Run the test lots.

**Anything else we need to know?**:

If I had to guess, the test is expecting a response, but for whatever reason that UDP packet is lost on the way back (which is never guaranteed). The condition for the test should be on the Shutdown of the GameServer, not on the receipt of the message.

**Environment**: Cloud Build
- Agones version: dev
- Kubernetes version (use `kubectl version`): 1.21.0
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): helm
- Troubleshooting guide log(s):
- Others:

markmandel:
Test logs:

```
Step #21 - "e2e-feature-gates": time="2022-02-14 18:01:04.296" level=info msg="waiting for fleet condition" fleet=simple-fleet-rp9gw test=TestFleetRecreateGameServers/gameserver_shutdown
Step #21 - "e2e-feature-gates": time="2022-02-14 18:01:06.889" level=info msg="Checking Fleet Ready replicas" expected=10 fleet=simple-fleet-rp9gw fleetStatus="{Replicas:10 ReadyReplicas:0 ReservedReplicas:0 AllocatedReplicas:0 Players:0xc00013e9d0}" test=TestFleetRecreateGameServers/gameserver_shutdown
Step #21 - "e2e-feature-gates": time="2022-02-14 18:01:10.468" level=info msg="Checking Fleet Ready replicas" expected=10 fleet=simple-fleet-rp9gw fleetStatus="{Replicas:10 ReadyReplicas:6 ReservedReplicas:0 AllocatedReplicas:0 Players:0xc0005b0f20}" test=TestFleetRecreateGameServers/gameserver_shutdown
[...]

PR #2510, opened 2022-03-12
C# Health and WatchGameServer fixes

mleenhardt:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

This PR fixes a few issues we've observed in the C# SDK.
- The `IClientStreamWriter<Empty>` used for calls to `HealthAsync()` was created once during the SDK's constructor and never recreated. This made it so any subsequent calls to `HealthAsync()` would instantly fail if the stream had encountered a transient error that could normally be resolved by recreating a stream from the underlying channel. The fix we're proposing for this is to always check whether a stream needs to be created before attempting to write to it, and just null out the stream if an exception occurs, so the next caller can create a fresh stream and retry writing. We chose to account for the possibility of `HealthAsync()` being called concurrently from multiple threads and thus had to synchronize access to the underlying stream. We  however think that the typical use case for this call is likely to be called from a single thread a time, hence the cost for the lock that was added should be minimal.
- `BeginInternalWatchAsync` had several related issues.
  - It would fail and return if any transient issue happened to the underlying stream or if a user callback threw an exception.
  - It was rethrowing any caught exception but if the exception was thrown after the first yielding `await`, there was no way for the caller of `WatchGameServer()` to be notified that an exception happened and the watch had stopped.
  - It would dispose the whole SDK object upon failure / completion, which felt particularly unexpected from a user perspective.
  - We made it so similarly to `HealthAsync()`, it will attempt to recreate its stream if the current one throws any exception and will retry subscribing to GameServer events, until it gets cancelled.
- Finalizers were removed from `AgonesSDK` and `Alpha`. C# finalizers are only intended to dispose of unmanaged resources which neither of these classes holds. Besides, any exception thrown in a finalizer takes down the whole process so these should only be defined when absolutely required.

We observed the `HealthAsync()` issue in our deployed environments for a while now where a pod would come up, start instantly failing its health check and eventually get killed by the side car because it had never reported healthy. The second pod that would come online to replace the first one would most of the time be able to succeed its first and subsequent health checks. We believe that this was caused by a race condition between the game server pod coming online and its side car. If the game server attempted to send its health check too early before the side car was ready to handle connections, the underlying `HealthAsync()` stream would become unusable and couldn't be repaired for the reasons explained above.

### DIFF
--- a/sdks/csharp/sdk/AgonesSDK.cs
+++ b/sdks/csharp/sdk/AgonesSDK.cs
@@ -46,7 +46,7 @@ namespace Agones
 		/// without starting a new task for every subscription request.
 		/// </summary>
 		private event Action<GameServer> GameServerUpdated;
-		internal Delegate[] GameServerUpdatedCallbacks => GameServerUpdated.GetInvocationList();
+		internal Delegate[] GameServerUpdatedCallbacks => GameServerUpdated?.GetInvocationList();
 		private readonly ILogger _logger;
 		private readonly SemaphoreSlim _healthStreamSemaphore = new SemaphoreSlim(1, 1);
 		private readonly object _gameServerWatchSyncRoot = new object();
@@ -98,7 +98,7 @@ namespace Agones
 			{
 				return true;
 			}
-			LogError(null, $"Could not connect to the sidecar at {Host}:{Port}. Exited with connection state: {channel.State}.");
+			LogError($"Could not connect to the sidecar at {Host}:{Port}. Exited with connection state: {channel.State}.");
 			return false;
 		}
 
@@ -115,7 +115,7 @@ namespace Agones
 			}
 			catch (RpcException ex)
 			{
-				LogError(ex, "Unable to mark GameServer to 'Ready' state.");
+				LogError("Unable to mark GameServer to 'Ready' state.", ex);
 				return ex.Status;
 			}
 		}
@@ -135,7 +135,7 @@ namespace Agones
 			}
 			catch (RpcException ex)
 			{
-				LogError(ex, "Unable to mark the GameServer to 'Allocated' state.");
+				LogError("Unable to mark the GameServer to 'Allocated' state.", ex);
 				return ex.Status;
 			}
 		}
@@ -158,7 +158,7 @@ namespace Agones
 			}
 			catch (RpcException ex)
 			{
-				LogError(ex, "Unable to mark the GameServer to 'Reserved' state.");
+				LogError("Unable to mark the GameServer to 'Reserved' state.", ex);
 				return ex.Status;
 			}
 		}
@@ -178,7 +178,7 @@ namespace Agones
 			}
 			catch (RpcException ex)
 			{
-				LogError(ex, "Unable to get GameServer configuration and status.");
+				LogError("Unable to get GameServer configuration and status.", ex);
 				throw;
 			}
 		}
@@ -203,10 +203,11 @@ namespace Agones
 							{
 								GameServerUpdated?.Invoke(reader.Current);
 							}
-							catch
+							catch (Exception ex)
 							{
-								// Ignore any exception thrown here. We don't want a callback's exception to cause
+								// Swallow any exception thrown here. We don't want a callback's exception to cause
 								// our watch to be torn down.
+								LogWarning($"A {nameof(WatchGameServer)} callback threw an exception", ex);
 							}
 						}
 					}
@@ -221,7 +222,7 @@ namespace Agones
 				}
 				catch (RpcException ex)
 				{
-					LogError(ex, "An error occurred while watching GameServer events, will retry.");
+					LogError("An error occurred while watching GameServer events, will retry.", ex);
 				}
 			}
 		}
@@ -263,7 +264,7 @@ namespace Agones
 			}
 			catch (RpcException ex)
 			{
-				LogError(ex, "Unable to mark the GameServer to 'Shutdown' state.");
+				LogError("Unable to mark the GameServer to 'Shutdown' state.", ex);
 				return ex.Status;
 			}
 		}
@@ -287,7 +288,7 @@ namespace Agones
 			}
 			catch(RpcException ex)
 			{
-				LogError(ex, $"Unable to set the GameServer label '{key}' to '{value}'.");
+				LogError($"Unable to set the GameServer label '{key}' to '{value}'.", ex);
 				return ex.Status;
 			}
 		}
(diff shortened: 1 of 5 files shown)


sdks/csharp/test/csharp-sdk-test.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
mleenhardt, line 4: NET Core 2.2 support ended in December 2019 so hopefully this is fine to update to the previous NET Core LTS (current LTS is NET 6). Let me know if you'd rather not update. This is only the test project, the SDK library itself is still targeting NET Standard 2.0.

markmandel:
@rcreasey would you have a moment to take a look at this? You know C# way better than I do.

sdks/csharp/sdk/AgonesSDK.cs
```diff
@@ -57,11 +61,20 @@ public sealed class AgonesSDK : IAgonesSDK
 		{
 			_logger = logger;
 			RequestTimeoutSec = requestTimeoutSec;
-			cts = cancellationTokenSource ?? new CancellationTokenSource();
+			
+			if (cancellationTokenSource == null)
```
DazEdword, line 65: A general note for this and other `null` checks. IIRC the semantics of checking null with equality/inequality operator vs using `is` can be slightly different. I'm not familiar with this code, so I'll leave this as something that you might want to check?

https://stackoverflow.com/questions/40676426/what-is-the-difference-between-x-is-null-and-x-null

sdks/csharp/sdk/AgonesSDK.cs
```diff
@@ -174,24 +187,42 @@ public async Task<GameServer> GetGameServerAsync()
...
+						{
+							try
+							{
+								GameServerUpdated?.Invoke(reader.Current);
+							}
+							catch
+							{
+								// Ignore any exception thrown here. We don't want a callback's exception to cause
+								// our watch to be torn down.
+							}
```
DazEdword, line 210: I tend to be wary of empty catches. Perhaps you would consider adding something, even if it is a debug log?

sdks/csharp/sdk/AgonesSDK.cs
```diff
@@ -286,24 +324,74 @@ public async Task<Status> HealthAsync()
...
+				{
+					try
+					{
+						// Best effort to clean up.
+						healthStream.Dispose();
+					}
+					catch
+					{
+						// Intentionally empty
+					}
```
DazEdword, line 361: Similar note as above re empty catches. Debug log an option?

DazEdword (commented):
Haven't done C# in a while, but hope my thoughts help anyway.

rcreasey (approved):
These changes look good, but will defer the judgement for debug logging the watch exception.

sdks/csharp/sdk/AgonesSDK.cs
mleenhardt: They both compile down to the same IL when there is no overloaded `==` operator. Generally, I'd say `==` should be used when null checking because if somebody defined an overloaded operator for their type, it's probably best to use it.

sdks/csharp/sdk/AgonesSDK.cs
mleenhardt: Fair, I'll add a log

sdks/csharp/sdk/AgonesSDK.cs
mleenhardt: The reasoning was that we're already logging when the exception is thrown above, and that consumers wouldn't really care if we failed to dispose the underlying broken stream, but I can also add a log here.

markmandel:
Blerg, caught by:
* #2479

DazEdword (approved):
Lgtm!

rcreasey (approved):
LGTM.

markmandel (approved):

markmandel:
Flake: e2e-stable

```
--- FAIL: TestGameServerRestartBeforeReadyCrash (214.48s)
    gameserver_test.go:353: 
        	Error Trace:	gameserver_test.go:353
        	Test:       	TestGameServerRestartBeforeReadyCrash
```

markmandel:
@SaitejaTamma I'd like to get this into the RC today. Going to retry it.

This is a new flake:

```
panic: Fail in goroutine after TestFleetRollingUpdate/Use_fleet_Patch_true_25%_cycle_true has completed

goroutine 2447 [running]:
testing.(*common).Fail(0xc0008d3ba0)
	/usr/local/go/src/testing/testing.go:710 +0x1b4
testing.(*common).Errorf(0xc00012f5a0, {0x20a5142, 0x3}, {0xc00012f590, 0x1, 0x1})
	/usr/local/go/src/testing/testing.go:818 +0x85
github.com/stretchr/testify/assert.Fail({0x2324480, 0xc0008d3ba0}, {0xc000e10720, 0x5c}, {0x0, 0x0, 0x0})
	/go/src/agones.dev/agones/vendor/github.com/stretchr/testify/assert/assertions.go:262 +0x43d
github.com/stretchr/testify/assert.NoError({0x2324480, 0xc0008d3ba0}, {0x23239a0, 0xc000922640}, {0x0, 0x0, 0x0})
	/go/src/agones.dev/agones/vendor/github.com/stretchr/testify/assert/assertions.go:1332 +0x125
github.com/stretchr/testify/require.NoError({0x233d2e8, 0xc0008d3ba0}, {0x23239a0, 0xc000922640}, {0x0, 0x0, 0x0})
	/go/src/agones.dev/agones/vendor/github.com/stretchr/testify/require/require.go:1035 +0xa5
agones.dev/agones/test/e2e/framework.(*Framework).CycleAllocations.func1.1(0xc000406fc0)
	/go/src/agones.dev/agones/test/e2e/framework/framework.go:297 +0x1b4
created by agones.dev/agones/test/e2e/framework.(*Framework).CycleAllocations.func1
	/go/src/agones.dev/agones/test/e2e/framework/framework.go:294 +0x7e8
FAIL	agones.dev/agones/test/e2e	392.777s
```

But I figure we can hunt this down at some point.

markmandel:
Blerg, not doing well with the flakes again. retrying.:

e2e-feature-gates.
```
--- FAIL: TestFleetRollingUpdate (0.01s)
    --- FAIL: TestFleetRollingUpdate/Use_fleet_Patch_true_25%_cycle_true (114.91s)
        fleet_test.go:320: 
            	Error Trace:	fleet_test.go:320
FAIL
```

Issue #2216, opened 2021-08-10
Support Agones on ARM systems


wernerphysics:
I'm running Agones in a local Minikube cluster on an M1 Mac. When I create a basic xonotic game server from this repo's examples, I get an Image Pull Error. The error details specify that the image "gcr.io/agones-images/agones-sdk" does not have a linux/arm64/v8 manifest. The issue seems to be that the Docker runtime when running on M1 Mac expects only ARM images.

My suggestion is to provide an ARM64 build of the Agones SDK in the GCR repo so that developers on ARM machines can run Agones/Minikube locally.

markmandel:
I can see us needing to support ARM in the future long term, so this is a good ticket to track this work. 👍🏻 

Out of curiosity though: On an M1, can you _only_ run arm docker images? (i.e. minikube won't work with anything else?) or is there a path to running x86 images?

wernerphysics:
I looked into the issue some more and in fact you can run x86 images on M1 Mac Minikube. It works transparently through the Rosetta2 compatibility layer.

The issue: when the pod is created, it pulls agones-sdk:1.16.0 without an additional tag, which on a default M1 Mac install requests the nonexistent ARM64 image. I got it to work by editing the pod spec to specifically pull the agones-sdk:1.16.0-linux_amd64 image. After this step, both containers in the xonotic-example pod are working.

For what it's worth, I think this only works because of Apple's x86 emulation. It would still be helpful to add the ARM image tag for users on ARM platforms without robust emulation support.

markmandel:
>  It would still be helpful to add the ARM image tag for users on ARM platforms without robust emulation support.

I agree. Also combined with #2223 we should probably cross compile all the components.

davidpdrsn:
@wernerphysics I'm having the same issue. Could you elaborate on what you meant by

> I got it to work by editing the pod spec to specifically pull the agones-sdk:1.16.0-linux_amd64 image.

I'm still sort new to k8s 😅

markmandel:
If we want to do this in pieces, to start, let's get a ARM binary generated:
https://github.com/googleforgames/agones/blob/15678b65acd58c8fe9de8eb11e336375be3b4d79/build/Makefile#L382-L384

(Do you need an arm binary for [local sdk server development](https://agones.dev/site/docs/guides/client-sdks/loal/)?)

markmandel:
[...]

PR #2514, opened 2022-03-16
Sidecar arm64 bin

Ludea:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix

Closes https://github.com/googleforgames/agones/issues/2216

This PR currently build arm64 allocator binary

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -69,6 +69,8 @@ ALPHA_FEATURE_GATES ?= "PlayerTracking=true&StateAllocationFilter=true&PlayerAll
 
 # Build with Windows support
 WITH_WINDOWS=1
+# Build with arm64support
+WITH_ARM64=1
 
 # Directory that this Makefile is in.
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
@@ -135,6 +137,7 @@ endif
 
 go_build_cmd = go build -mod=vendor
 GO_BUILD_LINUX_AMD64=docker run --rm -e "CGO_ENABLED=0" -e "$(gomod_on)" -w $(workdir_path) $(common_mounts) $(build_tag) $(go_build_cmd)
+GO_BUILD_LINUX_ARM64=docker run --rm -e "CGO_ENABLED=0" -e "GOOS=linux" -e "GOARCH=arm64" -e "$(gomod_on)" -w $(workdir_path) $(common_mounts) $(build_tag) $(go_build_cmd)
 GO_BUILD_DARWIN_AMD64=docker run --rm -e "GOOS=darwin" -e "GOARCH=amd64" -e "$(gomod_on)" -w $(workdir_path) $(common_mounts) $(build_tag) $(go_build_cmd)
 GO_BUILD_WINDOWS_AMD64=docker run --rm -e "GOOS=windows" -e "GOARCH=amd64" -e "$(gomod_on)" -w $(workdir_path) $(common_mounts) $(build_tag) $(go_build_cmd)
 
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -18,6 +18,14 @@
 
 steps:
 
+#
+# Print Docker version
+#
+
+- name: "gcr.io/cloud-builders/docker"
+  id: version
```
markmandel, line 26:
```suggestion
  id: docker-version
```

👍🏻 hard to track otherwise.

build/Makefile
```diff
@@ -135,7 +137,9 @@ endif
 
 go_build_cmd = go build -mod=vendor
 GO_BUILD_LINUX_AMD64=docker run --rm -e "CGO_ENABLED=0" -e "$(gomod_on)" -w $(workdir_path) $(common_mounts) $(build_tag) $(go_build_cmd)
+GO_BUILD_LINUX_ARM64=docker run --rm -e "CGO_ENABLED=0" -e "GOOS=linux" -e "GOARCH=arm64" -e "$(gomod_on)" -w $(workdir_path) $(common_mounts) $(build_tag) $(go_build_cmd)
```
markmandel, line 140: Oh good catch there! We would want both darwin and linux arm builds! I did't think of that!

build/Makefile
```diff
@@ -382,19 +396,32 @@ push-controller-image: $(ensure-build-image)
 	docker push $(controller_tag)
 
 # build the static binary for the gameserver sidecar
-build-agones-sdk-binary: $(ensure-build-image) build-agones-sdk-binary-linux build-agones-sdk-binary-windows build-agones-sdk-binary-darwin
+ifeq ($(WITH_ARM64), 1)
```
markmandel, line 399: So for the SDK binary, I say let's always build all the sidecar binaries.

Since we have https://agones.dev/site/docs/guides/client-sdks/local/ - we might as well have them all, since we should ship a zip of all them with each release anyway.

So I think we can remove the `WITH_ARM64` condition on `build-agones-sdk-binary`.

build/Makefile
```diff
@@ -353,10 +357,20 @@ uninstall: $(ensure-build-image)
 
 
 # Build a static binary for the gameserver controller
+ifeq ($(WITH_ARM64), 1)
 build-controller-binary: $(ensure-build-image)
 	$(GO_BUILD_LINUX_AMD64) \
 	-tags $(GO_BUILD_TAGS) -o $(go_build_base_path)/cmd/controller/bin/controller \
 	$(go_rebuild_flags) $(go_version_flags) -installsuffix cgo $(agones_package)/cmd/controller
+	$(GO_BUILD_LINUX_ARM64) \
```
markmandel, line 365: Two fun things here:

1. The second build will overwrite the first, since the binary name for both builds is `controller`. We should probably replicate the pattern we have with the sdk and include some version of the arch in the file name.
1. I expect this is why the CI is failing, because we are sending an ARM64 binary to run in a AMD64 docker image.
1. If we change the filename, we'll need to also update the Docker images. I think for this PR we can hardcode that for now, and then come back around and fix it up when we do multi-arch images.
1. It would be good to copy the pattern of a seperate target for each arch type, like we do with the `sdk` targets, and then have those targets as dependencies of this one - that way they can be built in parrallel 😄

build/Makefile
```diff
@@ -446,10 +473,20 @@ build-ping-image: $(ensure-build-image) build-ping-binary build-licenses build-r
 	docker build $(agones_path)/cmd/ping/ --tag=$(ping_tag) $(DOCKER_BUILD_ARGS)
 
 # Build a static binary for the allocator service
+ifeq ($(WITH_ARM64), 1)
+build-allocator-binary: $(ensure-build-image)
```
markmandel, line 477: ⬆ Exactly the same comments here as with the controller binary.

markmandel (commented):
Thanks for doing the work to split this into a smaller PR! Much easier to review, and I think also made clear why CI is failing.

Lemme know if you have any questions on the comments below!

build/Makefile
Ludea: I really don't like harcoded variable.
WDYT about --build_arg option ?

Ludea: Will do harcoded var for now, and when building arm64 img, will use this pattern

markmandel:
On first pass, this looks good. I just want to run it locally, make sure it is doing what I expect.

@SaitejaTamma I would like to get this into today's RC.

markmandel:
I'd love to get this merged into RC, but I don't have time to today to take it for a spin (mostly I just wanted to see if the arm binary gets captured in the zip files)

@roberthbailey do you have time to run the code?

roberthbailey:
I pulled down the code and ran `make build-agones-sdk-binary` to see if the arm binay was in the resulting zip file.

The command created `agonessdk-server-1.22.0-2334829.zip` which I unzipped to find: `sdk-server.darwin.amd64`, `sdk-server.darwin.arm64`, `sdk-server.linux.amd64`, `sdk-server.linux.arm64`, and `sdk-server.windows.amd64.exe`. 

I compared this with what is in `agonessdk-server-1.21.0.zip ` from our releases page which doesn't include the arm64 binaries.

roberthbailey (approved):

PR #2516, opened 2022-03-16
e2e flow for High Density GameServers

### LINKED ISSUE #2408: GameServerAllocation is not working for "High Density GameServers"
**What happened**:
I followed the document "High Density GameServers" (at https://agones.dev/site/docs/integration-patterns/high-density-gameservers/) to run multiple concurrent game sessions in a single GameServer process.

I use Nodejs and the Nodejs SDK to implement the GameServer. The GameServer is implemented as follows:

```
const AgonesSDK = require('@google-cloud/agones-sdk');
const util = require('util');

const sleep = util.promisify(setTimeout);

const DEFAULT_TIMEOUT = 60;
const MAX_TIMEOUT = 2147483;

const connect = async () => {
	let agonesSDK = new AgonesSDK();

	let lifetimeInterval;
	let healthInterval;
	let serverName = ""
	
	try {
		console.log(`Connecting to the SDK server...`);
		await agonesSDK.connect();
		console.log('...connected to SDK server');
		

		let connectTime = Date.now();
		lifetimeInterval = setInterval(() => {
			let lifetimeSeconds = Math.floor((Date.now() - connectTime) / 1000);
			console.log(`Running for ${lifetimeSeconds} seconds`);
		}, 10000);
		healthInterval = setInterval(() => {
			agonesSDK.health();
			console.log('Health ping sent');
		}, 20000);


		agonesSDK.watchGameServer(async (result) => {
			let output = `GameServer Update:
	name: ${result.objectMeta.name}
	state: ${result.status.state}
	labels: ${result.objectMeta.labelsMap.join(' & ')}
	annotations: ${result.objectMeta.annotationsMap.join(' & ')}`;
			
			console.log(output);

			if (result.objectMeta.name == serverName && result.status.state == "Allocated" && output.indexOf("agones.dev/sdk-gs-session-ready,false") != -1){
				console.log("Set gs-session-ready to true")
				await agonesSDK.setLabel('gs-session-ready', "true");
			}
		}, (error) => {
			console.error('Watch ERROR', error);
			clearInterval(healthInterval);
			clearInterval(lifetimeInterval);
			process.exit(0);
		});

		await sleep(10000);
		console.log('Marking server as ready...');
		await agonesSDK.ready();

		let result = await agonesSDK.getGameServer();
		serverName = result.objectMeta.name;
		console.log('Server Info', result);

		await agonesSDK.setLabel('gs-session-ready', "false");

		result = await agonesSDK.getGameServer();
		console.log('Server Info', result);

		
		
		await sleep(864000 * 1000);
		console.log('Shutting down...');
		agonesSDK.shutdown();

		await sleep(10000);
		console.log('Closing connection to SDK server');
		clearInterval(healthInterval);
		agonesSDK.close();

		await sleep(10000);
		console.log('Exiting');
		clearInterval(lifetimeInterval);

		// Some parts of the SDK are still using the event loop so must exit manually until fixed
		process.exit(0);
	} catch (error) {
		console.error(error);
		clearInterval(healthInterval);
		clearInterval(lifetimeInterval);
		process.exit(0);
	}
};

connect()
```

I created a fleet as follows:
```
apiVersion: "agones.dev/v1"
kind: Fleet
metadata:
  name: sang
spec:
  replicas: 2
  strategy:
    type: Recreate
  template:
    spec:
      ports:
      - name: default
        containerPort: 26000
      health:
        initialDelaySeconds: 30
        periodSeconds: 60
      template:
        spec:
          containers:
          - name: sgs
            image: gcr.io/sangs-space/sang-gs:0.0.9
```

I use the follow GameServerAllocation to allocate GameServer to my GameClient.
```
kind: GameServerAllocation
spec:
  selectors:
    - matchLabels:
        agones.dev/fleet: sang
        agones.dev/sdk-gs-session-ready: "true" # this is important
      gameServerState: Allocated # new state filter: allocate from Allocated servers
    - matchLabels:
        agones.dev/fleet: sang
      gameServerState: Ready # Allocate out of the Ready Pool (which would be default, so backward compatible)
  metadata:
    labels:
      agones.dev/sdk-gs-session-ready: "false" # this removes it from the pool
```
      
 At the beginning, there are two GameServer whose status are Ready
 
```
NAME               STATE       ADDRESS         PORT   NODE                                    AGE
sang-6dcxr-dzv4v   Ready       34.81.194.170   7773   gke-agones-default-pool-0284a633-vjjb   39m
sang-6dcxr-qp99d  Ready   34.81.194.170   7360   gke-agones-default-pool-0284a633-vjjb   39m
```

The first time that I used the GameServerAllocation to request an Allocation, one of the GameServer's status change to Allocated.

```
NAME               STATE       ADDRESS         PORT   NODE                                    AGE
sang-6dcxr-dzv4v   Ready   34.81.194.170   7773   gke-agones-default-pool-0284a633-vjjb   39m
sang-6dcxr-qp99d  Allocated   34.81.194.170   7360   gke-agones-default-pool-0284a633-vjjb   39m
```

Then, I doubled check the label 'agones.dev/sdk-gs-session-ready' of the GameServer sang-6dcxr-qp99d  is changed to true.

The second time I used the GameServerAllocation to request an Allocation, the Allocator Service did not return the GameServer sang-6dcxr-qp99d, but return the other GameServer, which is sang-6dcxr-dzv4v. Moreover, the status of sang-6dcxr-dzv4v is changed to Allocated

```
NAME               STATE       ADDRESS         PORT   NODE                                    AGE
sang-6dcxr-dzv4v   Allocated   34.81.194.170   7773   gke-agones-default-pool-0284a633-vjjb   39m
sang-6dcxr-qp99d  Allocated   34.81.194.170   7360   gke-agones-default-pool-0284a633-vjjb   39m
```

The third time  I used the GameServerAllocation to request an Allocation, the Allocator Service did not return any GameServers. Even though, both of the GameServer had 'agones.dev/sdk-gs-session-ready' set to true.
 
- For the second call, the Allocator should return the GameServer sang-6dcxr-qp99d.

**Environment**:
- Agones version: 1.9
- Kubernetes version (use `kubectl version`): 1.21.6-gke.1500
- Others:

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

This PR includes two things, such that I believe we can close the below
issue:

1. An e2e test to confirm that the described flow in https://agones.dev/site/docs/integration-patterns/high-density-gameservers/
works as expected (it does 😄).
2. Add a section in the integration documentation to highlight the consistency models (i.e. things are not atomic operations, but
eventually consistent), and that people need to keep this in mind.

Closes #2408

N/A

### DIFF
--- a/test/e2e/gameserverallocation_test.go
+++ b/test/e2e/gameserverallocation_test.go
@@ -194,7 +194,8 @@ func TestHighDensityGameServerFlow(t *testing.T) {
 		require.Equal(t, string(allocationv1.GameServerAllocationAllocated), string(result.Status.State))
 
 		if expected != result.Status.GameServerName {
-			log.WithField("expected", expected).WithField("gsa", result).Debug("Re-allocation attempt failed. Retrying.")
+			log.WithField("expected", expected).WithField("gsa", result).Info("Re-allocation attempt failed. Retrying.")
+			return false
 		}
 
 		return true
--- a/test/e2e/gameserverallocation_test.go
+++ b/test/e2e/gameserverallocation_test.go
@@ -129,6 +129,78 @@ func TestCreateFleetAndGameServerStateFilterAllocation(t *testing.T) {
 	require.NotEqual(t, gs1.ObjectMeta.Annotations["agones.dev/last-allocated"], gs2.ObjectMeta.Annotations["agones.dev/last-allocated"])
 }
 
+func TestHighDensityGameServerFlow(t *testing.T) {
+	if !runtime.FeatureEnabled(runtime.FeatureStateAllocationFilter) {
+		t.SkipNow()
+	}
+	t.Parallel()
+	log := e2e.TestLogger(t)
+	ctx := context.Background()
+
+	fleets := framework.AgonesClient.AgonesV1().Fleets(framework.Namespace)
+	fleet := defaultFleet(framework.Namespace)
+	lockLabel := "agones.dev/sdk-available"
+	// to start they are all available
+	fleet.Spec.Template.ObjectMeta.Labels = map[string]string{lockLabel: "true"}
+
+	flt, err := fleets.Create(ctx, fleet, metav1.CreateOptions{})
+	require.NoError(t, err)
+	defer fleets.Delete(ctx, flt.ObjectMeta.Name, metav1.DeleteOptions{}) // nolint:errcheck
+
+	framework.AssertFleetCondition(t, flt, e2e.FleetReadyCount(flt.Spec.Replicas))
+
+	fleetSelector := metav1.LabelSelector{MatchLabels: map[string]string{agonesv1.FleetNameLabel: flt.ObjectMeta.Name}}
+	allocatedSelector := fleetSelector.DeepCopy()
+
+	allocated := agonesv1.GameServerStateAllocated
+	allocatedSelector.MatchLabels[lockLabel] = "true"
+	gsa := &allocationv1.GameServerAllocation{
+		Spec: allocationv1.GameServerAllocationSpec{
+			MetaPatch: allocationv1.MetaPatch{Labels: map[string]string{lockLabel: "false"}},
+			Selectors: []allocationv1.GameServerSelector{
+				{LabelSelector: *allocatedSelector, GameServerState: &allocated},
+				{LabelSelector: fleetSelector},
+			},
+		}}
+
+	// standard allocation
+	result, err := framework.AgonesClient.AllocationV1().GameServerAllocations(fleet.ObjectMeta.Namespace).Create(ctx, gsa, metav1.CreateOptions{})
+	require.NoError(t, err)
+	require.Equal(t, string(allocationv1.GameServerAllocationAllocated), string(result.Status.State))
+
+	gs, err := framework.AgonesClient.AgonesV1().GameServers(fleet.ObjectMeta.Namespace).Get(ctx, result.Status.GameServerName, metav1.GetOptions{})
+	require.NoError(t, err)
+	require.Equal(t, allocated, gs.Status.State)
+
+	// set the label to being available again
+	_, err = framework.SendGameServerUDP(t, gs, "LABEL available true")
+	require.NoError(t, err)
+
+	// wait for the label to be applied!
+	require.Eventuallyf(t, func() bool {
+		gs, err := framework.AgonesClient.AgonesV1().GameServers(fleet.ObjectMeta.Namespace).Get(ctx, result.Status.GameServerName, metav1.GetOptions{})
+		require.NoError(t, err)
+		log.WithField("labels", gs.ObjectMeta.Labels).Info("checking labels")
+		return gs.ObjectMeta.Labels[lockLabel] == "true"
+	}, time.Minute, time.Second, "GameServer did not unlock")
+
+	// Run the same allocation again, we should get back the preferred item.
+	expected := result.Status.GameServerName
+
+	// we will run this as an Eventually, as caches are eventually consistent
+	require.Eventuallyf(t, func() bool {
+		result, err = framework.AgonesClient.AllocationV1().GameServerAllocations(fleet.ObjectMeta.Namespace).Create(ctx, gsa, metav1.CreateOptions{})
+		require.NoError(t, err)
+		require.Equal(t, string(allocationv1.GameServerAllocationAllocated), string(result.Status.State))
+
+		if expected != result.Status.GameServerName {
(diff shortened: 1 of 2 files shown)


test/e2e/gameserverallocation_test.go
```diff
@@ -129,6 +129,78 @@ func TestCreateFleetAndGameServerStateFilterAllocation(t *testing.T) {
...
+		return gs.ObjectMeta.Labels[lockLabel] == "true"
+	}, time.Minute, time.Second, "GameServer did not unlock")
+
+	// Run the same allocation again, we should get back the preferred item.
+	expected := result.Status.GameServerName
+
+	// we will run this as an Eventually, as caches are eventually consistent
+	require.Eventuallyf(t, func() bool {
+		result, err = framework.AgonesClient.AllocationV1().GameServerAllocations(fleet.ObjectMeta.Namespace).Create(ctx, gsa, metav1.CreateOptions{})
+		require.NoError(t, err)
```
roberthbailey, line 193: If this fails does the enclosing require.Eventually retry? Or should a mismatch here cause the function to return false?

test/e2e/gameserverallocation_test.go
```diff
@@ -129,6 +129,78 @@ func TestCreateFleetAndGameServerStateFilterAllocation(t *testing.T) {
...
+		require.Equal(t, string(allocationv1.GameServerAllocationAllocated), string(result.Status.State))
+
+		if expected != result.Status.GameServerName {
+			log.WithField("expected", expected).WithField("gsa", result).Debug("Re-allocation attempt failed. Retrying.")
```
roberthbailey, line 197: You log that it will retry but this block also needs to return false.

test/e2e/gameserverallocation_test.go
```diff
@@ -129,6 +129,78 @@ func TestCreateFleetAndGameServerStateFilterAllocation(t *testing.T) {
...
+	// Run the same allocation again, we should get back the preferred item.
+	expected := result.Status.GameServerName
+
+	// we will run this as an Eventually, as caches are eventually consistent
+	require.Eventuallyf(t, func() bool {
+		result, err = framework.AgonesClient.AllocationV1().GameServerAllocations(fleet.ObjectMeta.Namespace).Create(ctx, gsa, metav1.CreateOptions{})
+		require.NoError(t, err)
+		require.Equal(t, string(allocationv1.GameServerAllocationAllocated), string(result.Status.State))
+
+		if expected != result.Status.GameServerName {
```
roberthbailey, line 196: I'm not clear on this check. This will fail if the allocation attempt succeeds but returns a different game server than the first allocation attempt, right? Does that warrant a retry or should that be considered a failure? It seems like the test could end up allocating all game servers in the fleet before coming back around and reallocating from the one from the first allocation and the test would still pass - but that isn't the intended behavior of the feature.

test/e2e/gameserverallocation_test.go
markmandel: No it does not. If there is an actual error on creating an allocation, something has gone horribly wrong, and therefore the test should fail at this point.

test/e2e/gameserverallocation_test.go
markmandel: oh! oops! Good catch! Eeep! That invalidates my test passing!

test/e2e/gameserverallocation_test.go
markmandel: Well, it has to work this way. Everything is eventually consistent (hence the change in the docs), so it should work correctly _eventually_ - the order actually doesn't matter.

We can't control when the allocation cache gets refreshed for example.

test/e2e/gameserverallocation_test.go
roberthbailey: Oh, I guess both this and the next check should alway succeed (we should get back an allocation and the resulting status should be allocated). So nevermind about this comment.

test/e2e/gameserverallocation_test.go
roberthbailey: I think it means your test passed without ever hitting this condition and needing to retry.

markmandel: Just made the changes locally - still passing 😅 . Updating incoming shortly.

test/e2e/gameserverallocation_test.go
markmandel: Locally this works on first try, but I expect on the e2e server under load, it might need a few retries.

roberthbailey (approved):

PR #2526, opened 2022-03-31
Check for DeletionTimestamp of fleet and gameserverset before scaling

### LINKED ISSUE #2524: Foreground deletion of a fleet managed by fleet autoscaler results in infinite pod recreation loop
**What happened**:

Deleting a fleet managed by a fleet autoscaler with foreground cascading deletion results in an infinite pod recreation loop.

This was found when using ArgoCD to try to manage fleet manifests as it uses foreground deletion of resources.

Fleet and all related resources (gameserverset, gameservers, pods) are deleted without recreation.

Apply the following manifest: [fleet.yaml.txt](https://github.com/googleforgames/agones/files/8383060/fleet.yaml.txt)

Then execute:
`$ kubectl delete fleet simple-game-server --cascade=foreground`

Pods will be terminated and recreated indefinitely until the gameserverset is manually removed with background cascading.

**Environment**:
- Agones version: 1.21.0
- Kubernetes version (use `kubectl version`): 1.21.5
- Install method (yaml/helm):  Helm

estebangarcia:
**What type of PR is this?**

/kind bug

It checks the DeletionTimestamp of GameServerSet and sets numServersToAdd to 0 and checks DeletionTimestamp of fleets managed by autoscaler and doesn't scale if they are marked for deletion.

This prevents agones from recreating gameservers from fleets marked for deletion by foreground cascading. As reported in #2524

Closes #2524

Tested on AWS EKS 1.21.5


markmandel (commented):
This straight makes sense! Thanks so much for the fix!

The only thing that would block this from merge is to add some unit tests to make sure we never mess up this functionality in the future.

(Do we need e2e tests? 🤔 )

estebangarcia:
Hey, @markmandel thanks for taking the time to review. I just added some unit tests to cover this.

I'm not sure if the e2e is necessary as it is a very small change, but I'll let you decide on that :)

pkg/fleetautoscalers/controller.go
```diff
@@ -292,6 +292,11 @@ func (c *Controller) syncFleetAutoscaler(ctx context.Context, key string) error
 		return err
 	}
 
+	// Don't do anything, the fleet is marked for deletion
+	if !fleet.DeletionTimestamp.IsZero() {
```
roberthbailey, line 296: Are there any other cases where this timestamp might be set to be non-zero?

According to https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion if the object is being deleted using foreground deletion it will also be apparent because the `metadata.finalizers` field will be set to `foregroundDeletion` in addition to having a deletion timestamp set. I'm curious if there is any reason to check the finalizers here in addition to the deletion timestamp.

pkg/gameserversets/controller.go
```diff
@@ -307,6 +307,12 @@ func (c *Controller) syncGameServerSet(ctx context.Context, key string) error {
 
 	numServersToAdd, toDelete, isPartial := computeReconciliationAction(gsSet.Spec.Scheduling, list, c.counter.Counts(),
 		int(gsSet.Spec.Replicas), maxGameServerCreationsPerBatch, maxGameServerDeletionsPerBatch, maxPodPendingCount)
+
+	// GameserverSet is marked for deletion then don't add gameservers.
+	if !gsSet.DeletionTimestamp.IsZero() {
```
roberthbailey, line 312: Similar question here - should we check for the finalizer being set?

roberthbailey (commented):
Should there also be a change in pkg/fleets/controller.go? If you delete a fleet with foreground cascading deletion and then proceed to update the fleet (e.g. change the container image) before the cascading deletion completes will the fleet controller create a new game server set which will then get orphaned?

markmandel:
> Should there also be a change in pkg/fleets/controller.go?

Oooh! That is an excellent question. I would argue yes - even if just to ensure no extra resource creation in general when a Fleet is being deleted (either foreground or background).

pkg/fleetautoscalers/controller.go
markmandel: 🤔 I'm _thinking_ that we don't need to check it, just because, regardless of if it's a foreground or background deletion, we shouldn't have activity in Fleets or related - it's a waste of resources.

estebangarcia: I'm not sure it's necessary to check for the finalizer specifically, we are mostly interested in knowing that the resource has been marked for deletion so as to not take any action that will end up in the recreation of gameservers unnecessarily.
The other case when DeletionTimestamp is set is for Orphan Cascading, which for FleetAutoScalers doesn't mean anything special as they are not owners of any resources and can be considered as a regular deletion, and for GameserverSets would mean it shouldn't add new game servers and the existing ones shouldn't be removed. For the latter, the current PR is setting `numServersToAdd` to 0 so it is consistent with that behaviour.

estebangarcia:
> Should there also be a change in pkg/fleets/controller.go?

I agree with this. I've just committed a change that considers this as well.

estebangarcia:
@markmandel the e2e tests on Cloud Build seemed to have failed I see a lot of `x509: certificate signed by unknown authority` errors but I don't know if that's expected or not.

markmandel:
Looks like the deity of flakiness has decided to target this PR. Rerunning.

estebangarcia:
I merged main to my fork, build is failing again. Going to need you to retry it when you can @markmandel

markmandel (approved):

markmandel:
We are past feature freeze - and I fixed the 404 in our site testing, so this is good to merge 👍🏻

PR #2547, opened 2022-04-20
Move some Info logging to Debug

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Was cleaning up a single instance where an Info log should have been Debug, so decided to do a scan of the codebase, and found a few different places where we have logging that should be Debug as it's either (a) verbose or (b) not relevant for production systems.

N/A

N.A


markmandel:
`TestGameServerRestartBeforeReadyCrash` - should really find a way to fix this flake.

roberthbailey (approved):

PR #2554, opened 2022-04-25
Fix for health controller race condition marking healthy pods as unhealthy

### LINKED ISSUE #2553: Healthcontroller falsly marks healthy pods as unhealthy
**What happened**:
The HealthController can mark healthy Gameservers as unhealthy. This is caused by a race condition.

This can happen under certain conditions:
- You can fit more GameServer per node than there are available ports. Pods would get unschedulable, the HealthController would pick up the `unschedulableWithNoFreePorts` condition.
- The health controller queue is backed up (hitting the rate limit) leading to delays greater than the minute in between `Update` event processing and `enqueue`. This can happen with a large buffer size.
- Game Servers occupying ports are naturally exiting
- Game Servers in the HealthController queue get scheduled on the newly free port
- The Game server can get allocated
- The HealthController processes the GameServer, marking it as unhealthy

See the following schema explaining the race condition (only relevant state transition represented, API calls ommited)
image

I propose to add a check to `HealthController::syncGameServer` to know if `hc.isUnhealthy` is still true at the moment the event is processed.

I could not reproduce simply in a e2e, let me know if this is required.

**Environment**:
- Agones version: 1.22
- Kubernetes version (use `kubectl version`): `v1.21.5`
- Others:

Thiryn:
**What type of PR is this?**

/kind bug

The HealthController processes events asynchronously and is rate limited.
It is possible for pods to not be unhealthy anymore after some time spent in the queue being rate limited. This check ensures the pod is still unhealthy when processing the event.

Closes #2553


pkg/gameservers/health.go
```diff
@@ -198,8 +198,28 @@ func (hc *HealthController) syncGameServer(ctx context.Context, key string) erro
 		return nil
 	}
 
-	if skip, err := hc.skipUnhealthy(gs); err != nil || skip {
-		return err
+	// retrieve the pod for the gameserver
+	pod, err := hc.podLister.Pods(gs.ObjectMeta.Namespace).Get(gs.ObjectMeta.Name)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
```
roberthbailey, line 204: I think this would be clearer (not withstanding the double negative) as:

```
if err != nil {
	if !k8serrors.IsNotFound(err) {
		// If the pod exists but there is an error, go back into the queue.
		return errors.Wrapf(err, "error retrieving Pod %s for GameServer to check status", gs.ObjectMeta.Name)
	}
	hc.baseLogger.WithField("gs", gs.ObjectMeta.Name).Debug("Could not find Pod")
}
```

which makes it clearer that one case _doesn't_ return and instead continues control flow down the function.

pkg/gameservers/health.go
```diff
@@ -198,8 +198,28 @@ func (hc *HealthController) syncGameServer(ctx context.Context, key string) erro
...
+			hc.baseLogger.WithField("gs", gs.ObjectMeta.Name).Debug("Could not find Pod")
+		} else {
+			// if it's something else, go back into the queue
+			return errors.Wrapf(err, "error retrieving Pod %s for GameServer to check status", gs.ObjectMeta.Name)
+		}
+	}
+
+	// If the pod doesn't exist, the GameServer is definitely not healthy
```
roberthbailey, line 212: This comment is confusing, because it's describing the "else" case for this conditional (not what is happening in the next block). What about:

"Make sure that the pod is in a healthy state"

Thiryn: I agree it is confusing. How about renaming the `skipUnhealthy` into `skipUnhealthyGameContainer`.
This is to avoid the following (accentuating the confusion with comments) :
```go
	// Make sure that the pod is in a healthy state
	if pod != nil {
		// skip if unhealthy
		if skip, err := hc.skipUnhealthy(gs, pod); err != nil || skip {
			return err
		}

		// skip if healthy
		if !hc.isUnhealthy(pod) {
			hc.baseLogger.WithField("gs", gs.ObjectMeta.Name).WithField("podStatus", pod.Status).Debug("GameServer is not unhealthy anymore")
			return nil
		}
	}

```

I propose
```go
	// Make sure that the pod has to be marked unhealthy
	if pod != nil {
		if skip, err := hc.skipUnhealthyGameContainer(gs, pod); err != nil || skip {
			return err
		}

		// If the pod is not unhealthy anymore, go back in the queue
		if !hc.isUnhealthy(pod) {
			hc.baseLogger.WithField("gs", gs.ObjectMeta.Name).WithField("podStatus", pod.Status).Debug("GameServer is not unhealthy anymore")
			return nil
		}
	}
```

roberthbailey: lgtm!

roberthbailey (approved):

Issue #2494, opened 2022-02-24
Upgrade to Kubernetes 1.22


markmandel:
List of items to do for upgrading to 1.22:

- [x] Update terraform submodules
  - [x] GKE
  - [x] Azure
  - [x] EKS
- [x] Update prow cluster to use 1.22 (even though we aren't using it yet, we should keep it in sync)
  - [x] Recreate cluster with new scripts: `cd build/terraform/prow; terraform apply -var project=agones-images`
- [x] Update e2e cluster to run against 1.22
  - [x] Recreate cluster with new scripts: `cd build/terraform/e2e; terraform apply -var project=agones-images`
- [x] Update kubectl in dev tooling
  - [x] Update kubectl in `build/build-image/Dockerfile`
  - [x] Update kubectl in `build/e2e-image/Dockerfile`
- [x] Update documentation for creating clusters to 1.22
  - [x] Config.toml `supported_k8s` and related (do `dev_` before main)
- [x] Update the dev tooling to create 1.22 clusters
  - [x] Minikube
  - [x] Kind
- [x] Update client-go
- [x] Update CRD API reference 
  - [x] Update links to k8s documentation in site/assets/templates/crd-doc-config.json
  - [x] Regenerate crd api reference docs - `make gen-api-docs`
- [ ] Regenerate Kubernetes resource includes (e.g. ObjectMeta, PodTemplateSpec)
  - [ ] Start a cluster with `make gcloud-test-cluster`, uninstall agones using `helm uninstall agones -n agones-system`, and then run  `make gen-embedded-openapi` and `make gen-install`

1804devs:
1. I edited  terraform submodules but i have a older chromebook so i wont be able to run terrform apply etc ..
2. I'll like to work on all 22 issues will i be able to without a ide

1804devs:
> markmandel

Thanks for reaching back. Im looking into github/ codespace to help . I'll keep you up todate

roberthbailey:
@1804devs - if you are looking for a good issue to get started, you might want to check out the [open issues labeled "good first issue"](https://github.com/googleforgames/agones/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).

roberthbailey:
Trying to recreate the e2e-test-cluster I get the following error from our terraform script:

```
$ terraform apply -var project=agones-images
module.gke_cluster.null_resource.test-setting-variables: Refreshing state... [id=231080905775840212]
[...]

PR #2557, opened 2022-04-25
Increase the e2e cluster size to match what was observed

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Work on #2494 

While recreating the e2e test cluster for the Kubernetes 1.22 update I noticed that the cluster size shrank from 12 nodes to 10 nodes. Here I'm bumping the size of the default node pool so that during the next upgrade it doesn't have to be manually adjusted.

### DIFF
--- a/build/terraform/e2e/module.tf
+++ b/build/terraform/e2e/module.tf
@@ -41,7 +41,7 @@ module "gke_cluster" {
     "name"             = "e2e-test-cluster"
     "zone"             = "us-west1-c"
     "machineType"      = "e2-standard-4"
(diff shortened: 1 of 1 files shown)


pooneh-m (approved):
Does the documentation need to be updated for this change?

roberthbailey:
> Does the documentation need to be updated for this change?

Since this just changes a value in the e2e configuration I don't think any documentation updates are necessary.

Issue #2570, opened 2022-05-05
The Player Capacity integration pattern example is still using the deprecated fields in the GameServerAllocation


roberthbailey:
On https://agones.dev/site/docs/integration-patterns/player-capacity/ we have an example game server allocation using the preferred and required fields in the spec:

```
apiVersion: "allocation.agones.dev/v1"
kind: GameServerAllocation
spec:
  preferred:
    - gameServerState: Allocated
      matchLabels:
        agones.dev/fleet: lobby
      players:
        minAvailable: 10
        maxAvailable: 15
  required:
    matchLabels:
      agones.dev/fleet: lobby
```

However https://agones.dev/site/docs/reference/gameserverallocation/ says that those fields are deprecated and selectors should be used instead. We should update the Player Capacity integration pattern example to use selectors instead.

PR #2577, opened 2022-05-12
Change deprecated preferred & required to selectors in Player Capacity docs

JJhuk:
**What type of PR is this?**
/kind documentation

Because `preferred` and `required` fields are deprecated in agones, so I changed them to `selectors` 

Closes https://github.com/googleforgames/agones/issues/2570

### DIFF
--- a/site/content/en/docs/Integration Patterns/player-capacity.md
+++ b/site/content/en/docs/Integration Patterns/player-capacity.md
@@ -30,16 +30,15 @@ to 15 players, and if it cannot find one, will allocate a Ready one from the sam
 apiVersion: "allocation.agones.dev/v1"
 kind: GameServerAllocation
 spec:
-  preferred:
-    - gameServerState: Allocated
-      matchLabels:
+  selectors:
+    - matchLabels:
         agones.dev/fleet: lobby
+      gameServerState: Allocated
       players:
         minAvailable: 10
         maxAvailable: 15
-  required:
-    matchLabels:
-      agones.dev/fleet: lobby
+    - matchLabels:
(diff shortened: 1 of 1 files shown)


site/content/en/docs/Integration Patterns/player-capacity.md
```diff
@@ -30,16 +30,15 @@ to 15 players, and if it cannot find one, will allocate a Ready one from the sam
 apiVersion: "allocation.agones.dev/v1"
 kind: GameServerAllocation
 spec:
-  preferred:
-    - gameServerState: Allocated
-      matchLabels:
+  selectors:
+    - matchLabels:
         agones.dev/fleet: lobby
+      gameServerState: Allocated
```
roberthbailey, line 36: I'm curious why you moved `gameServerState:` below `matchLabels:` since I would assume they would naturally go in the order (lexical) order as they were in the old preferred block.

JJhuk: In this [document](https://agones.dev/site/docs/reference/gameserverallocation/), All `selectors` order start with `matchLabels:` so I moved to `gameServerState:` below `matchLabels:` .
Do I happen to change order?

roberthbailey: You changed order in this file from how it was before, but you are right that it is consistent with the reference document (I think the reference doc uses that order to group all alpha fields together, whereas this doc does it in the order that would be output by k8s, which is lexical ordering). Since this is consistent with our existing documentation, I don't think you need to change it.

roberthbailey (approved):

PR #2589, opened 2022-05-20
Make Allocator batchWaitTime configurable

### LINKED ISSUE #2586: Make batchWaitTime configurable in the Allocator
We'd like to make the `batchWaitTime` configurable on the Alocator. On certain allocation pattern, the hard-coded 500ms sleep is too aggressive and degrading allocation latency.

Make the `batchWaitTime` configurable through an environment variable, with a default set to 500ms to avoid any breaking change.

* Change the hardcoded value `batchWaitTime` instead of making it configurable.
* Getting rid of the sleep altogether.

Those two solutions would lead to a change in existing behavior so I think they are not to be considered.

**Additional context**
From our internal testing the impact on CPU usage is negligible but the uplift in p95 allocation latency as well as the throughput in allocation per second is significant.
In any case the default value won't change so existing behaviour is kept intact.

valentintorikian:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix

This PR make the Allocator `batchWaitTime` configurable, giving more fine tuning capacity to the end user.
See #2586 

Closes #2586

### DIFF
--- a/cmd/allocator/metrics.go
+++ b/cmd/allocator/metrics.go
@@ -42,6 +42,7 @@ const (
 	apiServerSustainedQPSFlag        = "api-server-qps"
 	apiServerBurstQPSFlag            = "api-server-qps-burst"
 	logLevelFlag                     = "log-level"
+	allocationBatchWaitTime          = "allocation-batch-wait-time"
 )
 
 func init() {
@@ -62,6 +63,7 @@ type config struct {
 	LogLevel                     string
 	totalRemoteAllocationTimeout time.Duration
 	remoteAllocationTimeout      time.Duration
+	allocationBatchWaitTime      time.Duration
 }
 
 func parseEnvFlags() config {
@@ -79,6 +81,7 @@ func parseEnvFlags() config {
 	viper.SetDefault(remoteAllocationTimeoutFlag, 10*time.Second)
 	viper.SetDefault(totalRemoteAllocationTimeoutFlag, 30*time.Second)
 	viper.SetDefault(logLevelFlag, "Info")
+	viper.SetDefault(allocationBatchWaitTime, 500*time.Millisecond)
 
 	pflag.Int32(httpPortFlag, viper.GetInt32(httpPortFlag), "Port to listen on for REST requests")
 	pflag.Int32(grpcPortFlag, viper.GetInt32(grpcPortFlag), "Port to listen on for gRPC requests")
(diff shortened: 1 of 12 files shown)


cmd/allocator/metrics.go
```diff
@@ -42,6 +42,7 @@ const (
 	apiServerSustainedQPSFlag        = "api-server-qps"
 	apiServerBurstQPSFlag            = "api-server-qps-burst"
 	logLevelFlag                     = "log-level"
+	allocationBatchWaitTime          = "allocation-batch-wait-time"
```
markmandel, line 45: So this got me curious - since the value doesn't change (at least in a single install), why export is as a metric?

install/helm/agones/values.yaml
```diff
@@ -92,6 +92,7 @@ agones:
       periodSeconds: 3
       failureThreshold: 3
       timeoutSeconds: 1
+    allocationBatchWaitTime: 500ms
```
markmandel, line 95: 🤔 Just seeing if the idea makes sense - but should this be a single value across both the controller and allocator?

Probably not, since we duplicate other parameters between the two, but thought I would ask the questions just in case.

pkg/gameserverallocations/allocator.go
```diff
@@ -108,6 +107,7 @@ type Allocator struct {
 	remoteAllocationCallback     func(context.Context, string, grpc.DialOption, *pb.AllocationRequest) (*pb.AllocationResponse, error)
 	remoteAllocationTimeout      time.Duration
 	totalRemoteAllocationTimeout time.Duration
+	allocationBatchWaitTime      time.Duration
```
markmandel, line 110:
```suggestion
	batchWaitTime      time.Duration
```

Nit: Seems redundant to say allocation again, since we're already in the allocator.

markmandel (commented):
Looks pretty good! A couple of questions/nits, but otherwise should be good to go with the next release.

Only blocking issue, can you please add some docs to the "New Configuration Features" section on https://github.com/googleforgames/agones/blob/main/site/content/en/docs/Installation/Install%20Agones/helm.md for the new configuration paramter.

Thanks!

cmd/allocator/metrics.go
markmandel: OIC, this is just bad file naming. We should clean that up at some point.

PR #2612, opened 2022-06-06
Fix double build of SDK binary in Makefile

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Previously we were building each of the SDK binaries for both arm64 amd amd64 which could potentially run into each other during builds.

So being specific instead, which is faster, and also avoids race conditions.

N/A

N/A


roberthbailey (approved):

markmandel:
Weird we're getting allocation issues 🤔

PR #2629, opened 2022-06-14
Build image: Use apt to install gcloud

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Since there are now debian repos for install gcloud's cli, let's use that as the source of truth for gcloud installation within our build image.

N/A just doing some cleanup with the new gke-gcloud-auth-plugin on my linux systems, so I figured I would make it all match.

I'd say let's merge this after stable release next week - 96% sure I haven't broken anything, but would hate it to block stable release.

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -31,19 +31,17 @@ RUN wget -q https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
 
 # install gcloud + kubectl, because it's an easy way to test/dev against kubernetes.
 WORKDIR /opt
-RUN wget -q https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.zip && unzip -q google-cloud-sdk.zip && \
-    rm google-cloud-sdk.zip && \
-    /opt/google-cloud-sdk/install.sh --usage-reporting=true --path-update=true --bash-completion=true --rc-path=/root/.bashrc
 
-# update the path for both go and gcloud
-ENV PATH /usr/local/go/bin:/go/bin:/opt/google-cloud-sdk/bin:$PATH
+# credits https://cloud.google.com/sdk/docs/install#deb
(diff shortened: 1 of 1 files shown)


build/build-image/Dockerfile
```diff
@@ -31,20 +31,18 @@ RUN wget -q https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
...
+    apt-get install google-cloud-cli google-cloud-cli-gke-gcloud-auth-plugin google-cloud-cli-app-engine-go -y && \
+    echo "source /usr/share/google-cloud-sdk/completion.bash.inc" >> /root/.bashrc
+
+# update the path for go
+ENV PATH /usr/local/go/bin:/go/bin:$PATH
 
 # install go tooling for development, building and testing
 RUN go install golang.org/x/tools/cmd/goimports@latest
 
-# RUN gcloud components update
```
roberthbailey, line 44: Do we no longer need to try to update? Looking at the docs the command would be something like:

`sudo apt-get update && sudo apt-get --only-upgrade install google-cloud-cli google-cloud-cli-gke-gcloud-auth-plugin  google-cloud-cli-app-engine-go`

markmandel: We don't need to update since `apt` installs the latest version.

Also the line where we install all the components is here:
https://github.com/googleforgames/agones/pull/2629/files#diff-686b06cb4b89c224b0df5f76fa2bbc7b5757477074e6917a21b8040a9e6857a4R37

markmandel:
e2e-features

```
time="2022-06-23 17:38:06.523" level=info msg="gameserver 3/10 not found"
    --- FAIL: TestFleetRollingUpdate/Use_fleet_Patch_true_25%_cycle_true (98.00s)
        fleet_test.go:413: 
            	Error Trace:	fleet_test.go:413
time="2022-06-23 17:38:07.704" level=info msg="Checking Fleet Ready replicas" expected=10 fleet=simple-fl
```

build/build-image/Dockerfile
roberthbailey: Before we downloaded the latest zip but still ran components update afterwords. I wasn't sure whether something similar was still necessary but it sounds like you don't think so.

roberthbailey (approved):

PR #2639, opened 2022-06-22
New example to easily setup multi-cluster Agones on GKE with attached allocation-endpoint and monitoring to Google Cloud Monitoring

pszemko-t:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This is terraform script that can be used as an example for those that want to test setup that includes:
1. Multiple clusters (2 by default in 2 different regions)
2. Allocation endpoint 
3. Monitoring to Cloud Monitoring 

I must admit this is nothing new or ground breaking by any stretch of imagination :) this is just something that helped me while I was testing Agones  so I figured I might as well share.


examples/multiple-clusters-with-allocation-endpoint/terraform/modules/gke_with_agones/main.tf
```diff
@@ -0,0 +1,113 @@
+# Copyright 2022 Google LLC All Rights Reserved.
```
markmandel, line 1: Do we need to replicate this Terraform stuff we have in https://github.com/googleforgames/agones/tree/main/install/terraform/modules (this is duplicate right, I'm not missing something?)

Could we reference it relatively, or do it similarly to how we do things here:

https://github.com/googleforgames/agones/blob/6e57d34ed1c0250555b986ff83c7f9745ac11f8b/examples/terraform-submodules/gke/module.tf#L82-L100

pszemko-t: In short:
It is not 1:1 but we could use it. I would need to update stuff in https://github.com/googleforgames/agones/tree/main/install/terraform/modules  to make it work in the same way it currently works and in "my way" where it it needed. 

Bit longer explenation:
In the end it was "by choice" I wanted to avoid making any updates to /install/terraform/modules but those updates are nothing major so I guess I can do it but I will need to modify existing scripts in /install/terraform... a bit...

For GKE i use dedicate service account instead of default one. 
Some additional parameters in helm3 module agones.allocator.* are easy ones agones.metrics.* managable. also managable bit more puzzled with what to do with agones.serviceaccount.controller.* but I'll think of something. 

So do you think I should do it ?

markmandel: Yes, I do. if we are providing examples, we should eat our own dogfood. It's only going to confuse things if we have competing solutions.

If there are good reasons to expand the feature set of the existing solutions, and there is no downside, we should add them and make the developer experience as smooth as possible.

markmandel:
Just wanted to gently bump this, see what the plans where?

Tadzioss:
Sorry for lack of response... had some issues will finish it sometime this month. Again sorry.

markmandel:
No need for apologies - priorities shift all the time 😄 just wanted to touch base and see where things were at. We can always close this and revisit at a later date if need be.

Tadzioss:
That makes a lot of sense. Let's close it for now. I do not want it just to hang here without no apparent reason :)

markmandel:
Sure! No problem at all!

PR #2649, opened 2022-06-28
Health check error message

XiaNi:
**What type of PR is this?**
/kind feature

it turns faceless error like:
```
Error: write after end
    at new NodeError (node:internal/errors:371:5)
    at _write (node:internal/streams/writable:319:11)
    at ClientHttp2Stream.Writable.write (node:internal/streams/writable:334:10)
    at Http2CallStream.writeMessageToStream (/home/xiani/work/msa-xrengine/node_modules/@grpc/grpc-js/src/call-stream.ts:499:23)
    at /home/xiani/work/msa-xrengine/node_modules/@grpc/grpc-js/src/call-stream.ts:851:14
```
to
```
Error: health ping connection failure: write after end
    at /home/xiani/work/msa-xrengine/packages/instanceserver/src/start.ts:85:15
    at onwriteError (node:internal/streams/writable:415:3)
    at onwrite (node:internal/streams/writable:457:7)
    at processTicksAndRejections (node:internal/process/task_queues:82:21)
```
where developer can see who was trying to write and why.


steven-supersolid:
Nice addition! Could you check coverage with `npm run cover` and see if it is now reduced from 100%? Then add a test if so. I think we have some existing tests for errors that can be used as a reference

XiaNi:
@steven-supersolid recheck please, if it's ok to make health() function to be able to take optional errorCallback?
I made so to make users be able to catch stream write error if it's needed by providing errorCallback - if it's there - it will be called on error, otherwise error will be thrown as usual.
3 tests cases added to provide 100% coverage.

markmandel:
```
--- FAIL: TestGameServerAllocationDuringGameServerDeletion (0.00s)
    --- FAIL: TestGameServerAllocationDuringGameServerDeletion/scale_down (16.42s)
        fleet_test.go:947: 
            	Error Trace:	fleet_test.go:947
            	            				fleet_test.go:959
            	Error:      	Should NOT be empty, but was []
            	Test:       	TestGameServerAllocationDuringGameServerDeletion/scale_down
```
e2e feature gates - that's a new one, but looks unrelated to the change at hand.

sdks/nodejs/spec/agonesSDK.spec.js
```diff
@@ -187,6 +187,49 @@ describe('AgonesSDK', () => {
...
+		it('calls the server and handles stream write error if callback provided', async () => {
+			let stream = jasmine.createSpyObj('stream', ['write', 'on']);
+			stream.write.and.callFake((chunk, encoding, callback) => {
+				callback('error');
+			});
+			spyOn(agonesSDK.client, 'health').and.callFake(() => {
+				return stream;
+			});
+			try {
+				agonesSDK.health(() => {});
```
steven-supersolid, line 199: It may be worth adding an expect in here to test/demonstrate that the error is passed to the callback intact.
Could spy on the callback and I think and check what it was called with

sdks/nodejs/spec/agonesSDK.spec.js
```diff
@@ -187,6 +187,49 @@ describe('AgonesSDK', () => {
...
+			} catch (error) {
+				fail();
+			}
+		});
+		it('calls the server and re throws stream write error if no callback', async () => {
```
steven-supersolid, line 204: nit: For consistency with other tests, perhaps we should add a newline here to separate tests

sdks/nodejs/spec/agonesSDK.spec.js
```diff
@@ -187,6 +187,49 @@ describe('AgonesSDK', () => {
...
+			});
+			try {
+				agonesSDK.health();
+				fail();
+			} catch (error) {
+				expect(agonesSDK.client.health).toHaveBeenCalled();
+				expect(error).toEqual('error');
+			}
+		});
+		it('do not call error callback if there was no stream error', async () => {
```
steven-supersolid, line 220: nit
```suggestion
		it('does not call error callback if there was no stream error', async () => {

```

sdks/nodejs/src/agonesSDK.js
```diff
@@ -93,14 +93,27 @@ class AgonesSDK {
 		});
 	}
 
-	health() {
+	health(errorCallback) {
 		if (this.healthStream === undefined) {
 			this.healthStream = this.client.health(() => {
 				// Ignore error as this can't be caught
 			});
+			if (typeof this.healthStream.on === 'function') {
```
steven-supersolid, line 101: @XiaNi I'm just wondering a couple of things about this change:
1. Do we need to listen to the event if we are already passing a callback above? Or is this just to make testing easier?
2. If so, do we need the `typeof` check? Generally I would trust gRPC client types not to change unless we intentionally update the client and then would change this code too

steven-supersolid:
The callback is a great idea - we may have to update some documentation too though, e.g. in https://github.com/googleforgames/agones/blob/main/site/content/en/docs/Guides/Client%20SDKs/nodejs.md

sdks/nodejs/spec/agonesSDK.spec.js
XiaNi: done

sdks/nodejs/spec/agonesSDK.spec.js
XiaNi: done

XiaNi:
> The callback is a great idea - we may have to update some documentation too though, e.g. in https://github.com/googleforgames/agones/blob/main/site/content/en/docs/Guides/Client%20SDKs/nodejs.md

documentation is updated now too

sdks/nodejs/src/agonesSDK.js
steven-supersolid: Thanks for the detailed explanation. On point 2 I'd add an 'on' to the stream spy object so we can remove this check

steven-supersolid (approved):

markmandel (approved):

PR #2661, opened 2022-07-07
e2e tests and bug fixes: SDKGracefulTermination

### LINKED ISSUE #2647: End to end tests for SDKGracefulTermination

We've had [Graceful Termination for GameServer SDK](https://github.com/googleforgames/agones/pull/2205) in Alpha since 1.18.0, and it would be great to move it forward into Beta, however - we don't have any e2e tests for it, so it's hard for us to judge its stability.

We should write some e2e tests for it, so we feel more confident in the implementation, and can gradate the feature.

Delete the feature, but I'm pretty sure it's being used, and it seems like a good thing.

**Additional context**

I _think_ a potential way to test this would be (need to check this strategy sends the required signals):

1. Create a `GameServer`, with a `terminationGracePeriodSeconds` of 60s and wait for it to move to `Ready`
2. Delete the Ready GameServer via the K8s API: Rather than the Pod going away right away, we should see it wait (maybe ensure it still exists 20s later?)
3. Since it's still running, you should be able to send it a `SHUTDOWN` command, in which case, it should get shutdown relatively quickly (or at least faster than the remaining 40s left).

Some of the timing on this could be tweaked, but that's my general idea.

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

This commit does the following:

* Include SDKGracefulTermination in our e2e tests (previously missing)
* Add an e2e test for graceful shutdown
* Fix the bug wherein the `GameServer` sdk sidecar would only skip the
grace period if shutdown was fired from the SDK. Now it is skipped
whenever a GameServer moves to `Shutdown` state from anywhere.

Closes #2647

Best way I could figure to e2e test this was through Sleeps, and ensuring it was within the expected band. Not perfect, but it does the trick.


markmandel:
If you have feedback @bennetty, @theminecoder - would love to have your eyes on the changes.

theminecoder (approved):
LGTM!

pkg/sdkserver/sdkserver_test.go
```diff
@@ -1401,9 +1403,9 @@ func TestSDKServerGracefulTerminationInterrupt(t *testing.T) {
 	//	Assert gs is still requestReady
 	assert.Equal(t, agonesv1.GameServerStateRequestReady, sc.gsState)
 	// gs Shutdown
-	_, err = sc.Shutdown(sdkCtx, &sdk.Empty{})
-	require.NoError(t, err)
-	assert.Equal(t, agonesv1.GameServerStateShutdown, sc.gsState)
+	gs.Status.State = agonesv1.GameServerStateShutdown
+	fakeWatch.Modify(gs.DeepCopy())
```
roberthbailey, line 1407: It looks like all of the tests in this file now use `fakeWatch.Modify()` instead of calling `sc.Shutdown()`. Should we have unit tests for both?

Most of the lines to the shutdown state in https://agones.dev/site/docs/reference/gameserver/#gameserver-state-diagram are through the SDK.

test/e2e/gameserver_test.go
```diff
@@ -1147,3 +1147,50 @@ func TestPlayerConnectAndDisconnect(t *testing.T) {
...
+	gs.Spec.Template.Spec.TerminationGracePeriodSeconds = &minute
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(t, framework.Namespace, gs)
+	if err != nil {
+		t.Fatalf("Could not get a GameServer ready: %v", err)
+	}
+	assert.Equal(t, readyGs.Status.State, agonesv1.GameServerStateReady)
+	gameservers := framework.AgonesClient.AgonesV1().GameServers(framework.Namespace)
+	err = gameservers.Delete(ctx, readyGs.ObjectMeta.Name, metav1.DeleteOptions{})
+	require.NoError(t, err)
+	log.Info("Created GameServer, waited 20 seconds")
```
roberthbailey, line 1170: This log line is right after you are calling Delete (and before the 20 second sleep), which is confusing. Should it be: "Deleted GameServer, waiting 20 seconds..."

test/e2e/gameserver_test.go
```diff
@@ -1147,3 +1147,50 @@ func TestPlayerConnectAndDisconnect(t *testing.T) {
...
+	time.Sleep(20 * time.Second)
+	log.WithField("gs", gs).Info("Checking GameServer")
+	gs, err = gameservers.Get(ctx, readyGs.ObjectMeta.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	assert.Equal(t, readyGs.ObjectMeta.Name, gs.ObjectMeta.Name)
+
+	// move it to shutdown
+	gsCopy := gs.DeepCopy()
+	gsCopy.Status.State = agonesv1.GameServerStateShutdown
+	_, err = gameservers.Update(ctx, gsCopy, metav1.UpdateOptions{})
```
roberthbailey, line 1180: Why are you triggering the shutdown this way instead of sending `"EXIT"` to the game server so that it calls `sdk.Shutdown()` from within the game server?

The way I read the feature, the idea is that the sidecar waits to exit until the game server can clean up and call shutdown, so why not exercise that flow by having the game server call shutdown?

pkg/sdkserver/sdkserver_test.go
markmandel: > It looks like all of the tests in this file now use `fakeWatch.Modify()` instead of calling `sc.Shutdown()`. Should we have unit tests for both?

https://github.com/googleforgames/agones/blob/7492d86c2f55b3980050522f961f886507abbd96/pkg/sdkserver/sdkserver_test.go#L43

Has all the unit tests to ensure the SDK moves the GameServer to the correct state. We also have e2e tests to confirm that `sdk.Shutdown()` does move the `GameServer` to shutdown as well.

https://github.com/googleforgames/agones/blob/7492d86c2f55b3980050522f961f886507abbd96/test/e2e/gameserver_test.go#L693

So I don't think we need a separate `sdk.Shutdown()` unit test for this use case, since we already have tests showing that it will successfully move the `GameServer` to a `Shutdown` state.

This test confirms that no matter how a `GameServer` gets to a `Shutdown` state, it responds as expected.

test/e2e/gameserver_test.go
markmandel: Yes it should - good catch. Updating.

test/e2e/gameserver_test.go
markmandel: So we actually can't at this point, because the `simple-game-server` has already exited! 

What we are really testing here is what happens when a `GameServer` `Pod` container gets a SIGTERM  signal, as if a Node was about to shut down under the Pod (or really any other reason a SIGTERM is fired). To get that to happen, we have to delete the `GameServer` as we do above.

At that point, the `simple-game-server` handles the SIGTERM as an exit signal, and exits - so we can't send it more UDP packets. The sdk-server though acts as we would want - hanging out until the `GameServer` moves to `Shutdown` state. which keeps the `GameServer` around. 

> The way I read the feature, the idea is that the sidecar waits to exit until the game server can clean up and call shutdown,

Right - but that was the thing I want to fix in this PR. The `sdk-server` shouldn't be to wait until the GameServer calls _just_ `sdk.Shutdown()` - it should exit as soon as _anything_ moved the `GameServer` to `Shutdown` - which includes scale down from a Fleet -- otherwise the sidecar is always going to wait the grace period on scale down, regardless of if the GameServer is in `Shutdown` state or not.

So running e2e test through the SDK doesn't actually do anything for our testing. We only really care that the GameServer is responding to a Shutdown state from _anywhere_. That's what actually matters.

It's a subtle difference that ultimately makes a big different to overall processing times for Fleets in general - does that make sense?

roberthbailey: > So we actually can't at this point, because the `simple-game-server` has already exited!

That makes sense. If we wanted to test that flow we could modify the simple game server to wait N seconds after receiving a SIGTERM and then call shutdown, but real game servers that aren't designed to handle SIGTERM may just immediately exit as the simple-game-server does today and we don't want the sidecar hanging around in those cases. 

I'm confused a bit about your statement about things other than the SDK moving the gameserver to the shutdown state though. My understanding based on https://agones.dev/site/docs/reference/gameserver/#gameserver-state-diagram is that the SDK is the way that a game server moves to shutdown. But it sounds like the fleet controller may also do this as part of scaling down the fleet (to remove replicas) - but presumably only on ready game servers and not on reserved or allocated ones. Are there other scenarios that you know of that would also cause the game server to be moved into the shutdown state through an external mechanism?

test/e2e/gameserver_test.go
roberthbailey: In the new code I still see `log.Info("Created GameServer, waiting 20 seconds...")`

test/e2e/gameserver_test.go
markmandel: > My understanding based on https://agones.dev/site/docs/reference/gameserver/#gameserver-state-diagram is that the SDK is the way that a game server moves to shutdown

This is true. That is how the SDK works.

> But it sounds like the fleet controller may also do this as part of scaling down the fleet (to remove replicas) - but presumably only on ready game servers and not on reserved or allocated ones.

See:
https://github.com/googleforgames/agones/blob/7492d86c2f55b3980050522f961f886507abbd96/pkg/gameserversets/controller.go#L498-L514

> Are there other scenarios that you know of that would also cause the game server to be moved into the shutdown state through an external mechanism?

🤔 There could be an argument that we the SDK server might want to exit early on `Unhealthy` or `Error` state, bit given that that state should be the exception rather than the norm, I'm happy to wait and see how things go with those states, before making that change.

I did a quick search of the codebase, but outside the SDK and the `GameServerSet` controller, those are the only places that Agones moves a GameServer to `Shutdown` (there was a good reason to do it on Fleet scale down, but now I forget what it was - possibly to have a shutdown event in the GameServer? - although with this change, it's now very important)

But in theory, someone could build a system external to Agones that moves a GameServer to Shutdown as an explicit message to Shutdown a GameServer - since it is an open API after all.

markmandel: Oh! I know why we had to move to Shutdown on scale down! Otherwise you could accidentally delete an Allocated `GameServer`.

If you looked at the cached version of the GameServer, it might say it was in Ready state, but you wouldn't know. Deletion doesn't look at generational locking, so you _have_ to change the `GameServer` object to ensure that the copy you are looking at is actually the current state of the `GameServer` on etcd.

See - I knew there was an important reason.

test/e2e/gameserver_test.go
markmandel: 🤦🏻 I read the second part of the sentence, not the first.

pkg/sdkserver/sdkserver_test.go
roberthbailey: Makes sense.

test/e2e/gameserver_test.go
roberthbailey: This all makes sense, but makes me wonder if we should document this somewhere so it's discoverable (as I'm just learning about it now).

roberthbailey (approved):

PR #2662, opened 2022-07-07
Fix minikube dev tooling

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The update to multi-arch images broke the minikube dev tooling, this
fixes that.

N/A

Only caveat being, that I can only test on amd64 architecture. In theory, it might work on a M1, but if not, we can switch our which
images are loaded before being re-tagged based on host architecture.

### DIFF
--- a/build/includes/minikube.mk
+++ b/build/includes/minikube.mk
@@ -28,7 +28,7 @@ minikube_cert_mount := ~/.minikube:$(HOME)/.minikube
 # of the right version.
 minikube-test-cluster: DOCKER_RUN_ARGS+=--network=host -v $(minikube_cert_mount)
 minikube-test-cluster: $(ensure-build-image)
-	$(MINIKUBE) start --kubernetes-version v1.22.9 -p $(MINIKUBE_PROFILE) --vm-driver $(MINIKUBE_DRIVER)
+	$(MINIKUBE) start --kubernetes-version v1.22.9 -p $(MINIKUBE_PROFILE) --driver $(MINIKUBE_DRIVER)
 
 # Connecting to minikube requires so enhanced permissions, so use this target
 # instead of `make shell` to start an interactive shell for development on minikube.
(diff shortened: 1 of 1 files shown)


build/includes/minikube.mk
```diff
@@ -28,7 +28,7 @@ minikube_cert_mount := ~/.minikube:$(HOME)/.minikube
 # of the right version.
 minikube-test-cluster: DOCKER_RUN_ARGS+=--network=host -v $(minikube_cert_mount)
 minikube-test-cluster: $(ensure-build-image)
-	$(MINIKUBE) start --kubernetes-version v1.22.9 -p $(MINIKUBE_PROFILE) --vm-driver $(MINIKUBE_DRIVER)
+	$(MINIKUBE) start --kubernetes-version v1.22.9 -p $(MINIKUBE_PROFILE) --driver $(MINIKUBE_DRIVER)
```
markmandel, line 31: Snuck in a change - `wm-driver` is deprecated.

markmandel:
FYI: @mridulji since you tested this last and found the issue 👍🏻

build/includes/minikube.mk
```diff
@@ -39,9 +39,14 @@ minikube-shell: $(ensure-build-image)
...
 minikube-push:
 	$(MINIKUBE) image load $(sidecar_linux_amd64_tag) -p $(MINIKUBE_PROFILE)
-	$(MINIKUBE) image load $(controller_tag) -p $(MINIKUBE_PROFILE)
-	$(MINIKUBE) image load $(ping_tag) -p $(MINIKUBE_PROFILE)
-	$(MINIKUBE) image load $(allocator_tag) -p $(MINIKUBE_PROFILE)
+	$(MINIKUBE) image load $(controller_amd64_tag) -p $(MINIKUBE_PROFILE)
+	$(MINIKUBE) image load $(ping_amd64_tag) -p $(MINIKUBE_PROFILE)
+	$(MINIKUBE) image load $(allocator_amd64_tag) -p $(MINIKUBE_PROFILE)
+
+	$(MINIKUBE) image tag $(sidecar_linux_amd64_tag) $(sidecar_tag) -p $(MINIKUBE_PROFILE)
```
roberthbailey, line 46: As you noted, this seems likely to break if minikube is running on an ARM machine (e.g. an M1 mac).

roberthbailey (approved):

build/includes/minikube.mk
roberthbailey: I've sent https://github.com/googleforgames/agones/pull/2663 to update the docs so that this knowledge is more easily discoverable.

PR #2666, opened 2022-07-08
Building and pushing ARM64 images

### LINKED ISSUE #2517: Can't use `make` to build on m1 mac / ARM
**What happened**:

Attempting to run `make shell` results in the following error as the image attempts to build:

```shell
=> ERROR [ 7/31] RUN go install [golang.org/x/tools/cmd/goimports@latest](http://golang.org/x/tools/cmd/goimports@latest)                                                                  0.2s 
------                                                                                                                                         
 > [ 7/31] RUN go install [golang.org/x/tools/cmd/goimports@latest](http://golang.org/x/tools/cmd/goimports@latest):                                                                             
#10 0.187 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2':
```

We expect this is because the M1 pulls down the arm image of Debian, and therefore can't find the expected libraries, and there is an ARCH mismatch.

It to all work!

Run `make shell` (or any other build command) on an ARM system.

Short term fix (in theory):
Drop a `dev.mk` in the users's [local-includes](https://github.com/googleforgames/agones/tree/main/build/local-includes) folder with the following content:

```
DOCKER_BUILD_ARGS+= --platform linux/x86_64
DOCKER_RUN_ARGS+= --platform linux/x86_64
```
(And we should document it).

To ensure that we're running on the correct platform.

Long term fix:

Add a section to our [OS Include/detection code](https://github.com/googleforgames/agones/blob/main/build/Makefile#L173-L189), or specifically within the [macos.mk](https://github.com/googleforgames/agones/blob/main/build/includes/macos.mk) include, to automatically make this adjustment on a mac with M1 (not sure how to detect it though).

**Environment**:
- Agones version: dev
- Kubernetes version (use `kubectl version`): N/A
- Troubleshooting guide log(s): See above.
- Others:

mridulji:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

In order to build arm64 based images in Mac M1 device, we need to create a builder and use it to build images.

According to the docker documentation, to build the images for ARM64 arch CPU we would need to use `buildx` command which uses `builder`. So to create builder: 
```bash
docker buildx create --name mybuilder
```
We can see the supported platforms by:
```bash
docker buildx inspect --bootstrap
```
using builder in our command for building images:
image

So finally we are able to build images for Mac m1.

Closes #2517 

Reference 1: [build-and-run-multi-architecture-images](https://docs.docker.com/desktop/multi-arch/#build-and-run-multi-architecture-images) 
Reference 2: [Docker build with Mac M1](https://medium.com/geekculture/docker-build-with-mac-m1-d668c802ab96)

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -415,6 +415,8 @@ build-required-src-dist:
 	docker run --rm $(common_mounts) $(build_tag) $(mount_path)/build/build-required-src-dist.sh
 
 # Create docker builder used in creating arm64 images
+# According to the docker documentation, to build the images for ARM64 arch CPU 
+# we would need to use buildx command which uses builder https://docs.docker.com/desktop/multi-arch/
 create-arm64-builder:
 	-docker buildx create --name $(BUILDX_ARM64_BUILDER)
 
--- a/build/Makefile
+++ b/build/Makefile
@@ -426,6 +426,7 @@ endif
 
 build-controller-image-amd64: $(ensure-build-image) build-controller-binary build-licenses build-required-src-dist
 	docker build $(agones_path)/cmd/controller/ --tag=$(controller_amd64_tag) $(DOCKER_BUILD_ARGS)
+# creating docker builder and then using that builder to build controller image in buildx command
 build-controller-image-arm64: $(ensure-build-image) build-controller-binary
 	$(MAKE) create-arm64-builder 
 	docker buildx build --platform linux/arm64 --builder $(BUILDX_ARM64_BUILDER) --build-arg ARCH=arm64 $(agones_path)/cmd/controller/ --tag=$(controller_arm64_tag) $(DOCKER_BUILD_ARGS)
@@ -486,6 +487,7 @@ endif
 
 build-agones-sdk-image-amd64: $(ensure-build-image) build-agones-sdk-binary-linux-amd64
 	docker build $(agones_path)/cmd/sdk-server/ --tag=$(sidecar_linux_amd64_tag) $(DOCKER_BUILD_ARGS)
+# creating docker builder and then using that builder to build sdk image in buildx command
 build-agones-sdk-image-arm64: $(ensure-build-image) build-agones-sdk-binary-linux-arm64
 	$(MAKE) create-arm64-builder 
 	docker buildx build --platform linux/arm64 --builder $(BUILDX_ARM64_BUILDER) --build-arg ARCH=arm64 $(agones_path)/cmd/sdk-server/ --tag=$(sidecar_linux_arm64_tag) $(DOCKER_BUILD_ARGS) 
@@ -535,6 +537,7 @@ endif
 
 build-ping-image-amd64: $(ensure-build-image) build-ping-binary build-licenses build-required-src-dist
 	docker build $(agones_path)/cmd/ping/ --tag=$(ping_amd64_tag) $(DOCKER_BUILD_ARGS)
+# creating docker builder and then using that builder to build ping image in buildx command
 build-ping-image-arm64: $(ensure-build-image) build-ping-binary
 	$(MAKE) create-arm64-builder 
 	docker buildx build --platform linux/arm64 --builder $(BUILDX_ARM64_BUILDER) --build-arg ARCH=arm64 $(agones_path)/cmd/ping/ --tag=$(ping_arm64_tag) $(DOCKER_BUILD_ARGS) 
@@ -577,6 +580,7 @@ endif
 
 build-allocator-image-amd64: $(ensure-build-image) build-allocator-binary build-licenses build-required-src-dist
 	docker build $(agones_path)/cmd/allocator/ --tag=$(allocator_amd64_tag) $(DOCKER_BUILD_ARGS)
+# creating docker builder and then using that builder to build allocator image in buildx command
 build-allocator-image-arm64: $(ensure-build-image) build-allocator-binary
 	$(MAKE) create-arm64-builder 
 	docker buildx build $(agones_path)/cmd/allocator/ --builder $(BUILDX_ARM64_BUILDER) --tag=$(allocator_arm64_tag) --build-arg ARCH=arm64 --platform linux/arm64 $(DOCKER_BUILD_ARGS) 
--- a/build/Makefile
+++ b/build/Makefile
@@ -429,8 +429,7 @@ endif
 build-controller-image-amd64: $(ensure-build-image) build-controller-binary build-licenses build-required-src-dist
 	docker build $(agones_path)/cmd/controller/ --tag=$(controller_amd64_tag) $(DOCKER_BUILD_ARGS)
 # creating docker builder and then using that builder to build controller image in buildx command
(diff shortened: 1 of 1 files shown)


markmandel:
I'm struggling to work out why this change works / it's the appropriate path for the issue - can you write out a description of what is happening, and what the blocking issues you ran into where? (I read the article, but now I'm even more curious? I don't know if it made things clearer 😕 )

How does this relate to:
https://github.com/googleforgames/agones/blob/73cdabc4bf0f238ce2d913df8352d11b3a888445/build/Makefile#L189-L192

Or do they not relate at all?

Why is the builder only needed for Arm images?

This feels like a commit that needs a pretty solid description attached to it, so that if someone looks back to see why this was done, we have historical understanding.

mridulji:
>How does this relate to:
>https://github.com/googleforgames/agones/blob/73cdabc4bf0f238ce2d913df8352d11b3a888445/build/Makefile#L189-L192
>Or do they not relate at all?

This specifies the platform but in addition to this, we also need to use the `--builder` to be able to use `docker buildx` command.

build/Makefile
```diff
@@ -413,6 +414,10 @@ build-licenses:
 build-required-src-dist:
 	docker run --rm $(common_mounts) $(build_tag) $(mount_path)/build/build-required-src-dist.sh
 
+# Create docker builder used in creating arm64 images
```
markmandel, line 417: I'd love to see a comment/summary here on why the builder is necessary, maybe with pointing to one or more articles on the subject - so it's easy to look back and work out what is going on.

build/Makefile
```diff
@@ -421,8 +426,10 @@ endif
 
 build-controller-image-amd64: $(ensure-build-image) build-controller-binary build-licenses build-required-src-dist
 	docker build $(agones_path)/cmd/controller/ --tag=$(controller_amd64_tag) $(DOCKER_BUILD_ARGS)
+# creating docker builder and then using that builder to build controller image in buildx command
 build-controller-image-arm64: $(ensure-build-image) build-controller-binary
-	docker buildx build --platform linux/arm64 --build-arg ARCH=arm64 $(agones_path)/cmd/controller/ --tag=$(controller_arm64_tag) $(DOCKER_BUILD_ARGS)
+	$(MAKE) create-arm64-builder 
+	docker buildx build --platform linux/arm64 --builder $(BUILDX_ARM64_BUILDER) --build-arg ARCH=arm64 $(agones_path)/cmd/controller/ --tag=$(controller_arm64_tag) $(DOCKER_BUILD_ARGS)
```
markmandel, line 432: So we're only using this builder for arm64 images? Just for my own education, is there a reason we don't need it for the amd64 images?

build/Makefile
```diff
@@ -421,8 +426,10 @@ endif
 
 build-controller-image-amd64: $(ensure-build-image) build-controller-binary build-licenses build-required-src-dist
 	docker build $(agones_path)/cmd/controller/ --tag=$(controller_amd64_tag) $(DOCKER_BUILD_ARGS)
+# creating docker builder and then using that builder to build controller image in buildx command
 build-controller-image-arm64: $(ensure-build-image) build-controller-binary
-	docker buildx build --platform linux/arm64 --build-arg ARCH=arm64 $(agones_path)/cmd/controller/ --tag=$(controller_arm64_tag) $(DOCKER_BUILD_ARGS)
+	$(MAKE) create-arm64-builder
```
markmandel, line 431: Should this be:

`build-controller-image-arm64: $(ensure-build-image) build-controller-binary create-arm64-builder`

Rather than calling `create-arm64-builder` within the `build-controller-image-arm64` target?

(And same for other targets as well)

mridulji: Thanks for the suggestion, Yes it works 😃.

build/Makefile
mridulji: We are actually building multi-arch images for which builder is required in buildx command.

markmandel: We were chatting about this - just as a note - we are using the multi-arch build process to build out a single arch image (linux/arm64) -- but the important note is that for a M1 mac, there is no reason to use a builder on `docker build`  to build images.

build/Makefile
```diff
@@ -413,6 +414,12 @@ build-licenses:
 build-required-src-dist:
 	docker run --rm $(common_mounts) $(build_tag) $(mount_path)/build/build-required-src-dist.sh
 
+# Create docker builder used in creating arm64 images
+# According to the docker documentation, to build the images for ARM64 arch CPU 
+# we would need to use buildx command which uses builder https://docs.docker.com/desktop/multi-arch/
```
markmandel, line 419: Looking at https://docs.docker.com/desktop/multi-arch/#build-and-run-multi-architecture-images - I think we need to be clear on (and I am also wondering why it doesn't work 🤷🏻 that's so weird), why the default builder is not sufficient in this scenario?

Is there ongoing work to fix this, that we should keep an eye on?

build/Makefile
```diff
@@ -413,6 +414,12 @@ build-licenses:
...
+create-arm64-builder:
+	-docker buildx create --name $(BUILDX_ARM64_BUILDER)
```
markmandel, line 421: I noticed this silently fails -- what happens if this step fails? Will the `docker buildx` that uses this builder also fail? Since the image doesn't exist?

mridulji: >I noticed this silently fails

So it fails if builder with this name already exists. And if builder already present then `docker buildx` would work. Making this command optional became important to pass the `make build-images -j 4` command step of CI build process.

build/Makefile
mridulji: >I think we need to be clear on (and I am also wondering why it doesn't work 🤷🏻 that's so weird), why the default builder is not sufficient in this scenario?

Some clarifications I have got while going through this discussion about multi-arch images, One way could be to change the current approach of using `docker buildx` command and build the images for one platform at a time. Please let me know your thoughts.
Thanks🙂

build/Makefile
markmandel: Aaaah! Got it. let's make a comment to that effect.

build/Makefile
markmandel, line 419:
```suggestion
# we would need to use buildx command which uses builder https://docs.docker.com/desktop/multi-arch/
# We primarily need this for M1 Macs, see: https://medium.com/geekculture/docker-build-with-mac-m1-d668c802ab96
```

build/Makefile
markmandel, line 421:
```suggestion
	# Will fail once built for the first time, so we ignore failures.
	-docker buildx create --name $(BUILDX_ARM64_BUILDER)
```

markmandel (commented):
I tried this out on my Debian amd64 machine, and it all worked great.

I suggested some extra comments to add that I think will round this off nicely. Otherwise, I think this is good to go 👍🏻

markmandel (approved):
LGTM!

PR #2669, opened 2022-07-12
upgrade xonotic version

### LINKED ISSUE #2664: Upgrade Xonotic to 0.8.5
There's a new version of Xonotic!
https://xonotic.org/posts/2022/xonotic-0-8-5-release/

Source files: https://github.com/googleforgames/agones/blob/main/examples/xonotic

Would be nice to upgrade everything to the latest version!

mridulji:
**What type of PR is this?** 
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

**What this PR does / Why we need it**: This PR upgrades the xonotic version to 0.8.5 provided in the examples.

Closes #2664 

- Updated golang version in this example to resolve build failure.
- Added go.mod file to support newer version on golang
- Updated base image debian:stretch to debian:bullseye


examples/xonotic/go.mod
```diff
@@ -0,0 +1,19 @@
+module agones.dev/agones/examples/xonotic
```
roberthbailey, line 1: Can you still run `make build` from this directory to build this locally outside of the docker container?

examples/xonotic/go.mod
```diff
@@ -0,0 +1,19 @@
...
+	github.com/pkg/errors v0.9.1 // indirect
+	golang.org/x/net v0.0.0-20211209124913-491a49abca63 // indirect
+	golang.org/x/sys v0.0.0-20210616094352-59db8d763f22 // indirect
+	golang.org/x/text v0.3.6 // indirect
+	google.golang.org/genproto v0.0.0-20201110150050-8816d57aaa9a // indirect
+	google.golang.org/grpc v1.27.1 // indirect
+	google.golang.org/protobuf v1.26.0 // indirect
+)
+
+replace agones.dev/agones v0.0.0-dev => ../../
```
roberthbailey, line 19: @markmandel - what do you think about building this from head vs. a specific agones release? STK is currently built against 1.8.0, which is pretty old but also makes it clear the last time we updated it as well as making the build reproducible.

examples/xonotic/go.mod
mridulji: Before making these changes, I am not able to build image but `go build -o wrapper .` command creating the binary locally.
image

After making changes, I am able to build it both ways.

examples/xonotic/go.mod
markmandel: 🤔 hmnn, that's a good question.

The benefit I see of running against the development branch is that you can build out an example against something that is currently being worked on and not released.

That being said - pointing to a specific version ([like we do in STK](https://github.com/googleforgames/agones/blob/main/examples/supertuxkart/go.mod#L6)) is probably a better experience for the end user, so we should probably be consistent and do the same here.

It would also make the Docker build simpler as we don't have to force building from project root and copying in files (basically copy what we do in STK).

If we need to have in-development features, we can worry about it when it happens.

The only conflict I have - in simple-game-server we do probably want to continue building against a development version, as editing that binary is often necessary to implementing and testing new features. So that's likely the exception to the above rule.

Sound good?

mridulji: >pointing to a specific version ([like we do in STK](https://github.com/googleforgames/agones/blob/main/examples/supertuxkart/go.mod#L6)) is probably a better experience

I also feel that xonotic is pretty much similar STK example.

>If we need to have in-development features, we can worry about it when it happens.

Should we target to agones v1.8.0 (Works fine, just tested) or any other specific version?

markmandel: Let's use what is latest - `1.24.0`. Then we know when this example was built.

examples/xonotic/Makefile
```diff
@@ -41,6 +41,10 @@ image_tag = $(REPOSITORY)/xonotic-example:0.8
 build:
 	cd $(root_path) && docker build -f $(project_path)/Dockerfile --tag=$(image_tag) .
```
markmandel, line 42:
```suggestion
	docker build -f $(project_path)/Dockerfile --tag=$(image_tag) .
```

We no longer need all of Agones to build it, so we can avoid a big Docker context pull now!

STK example:
https://github.com/googleforgames/agones/blob/15e90007006df4b0ccde8e47a3ab2b66d38de8cf/examples/supertuxkart/Makefile#L40-L41

Other than that, I think this is good to go (I should test making it, but I'll do that one this is complete).

Not for this PR - but I noticed that `make build` doesn't build the windows image -- we should probably track that work somewhere.

mridulji: Thanks for the suggestion 😃. Please see if it looks good now.

markmandel (approved):

PR #2672, opened 2022-07-14
Bug: `make install` uses local helm and jq

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

In a difference between " and ', meant that the `make install` command that would check the helm install and uninstall if necessary was actually using the local installations of `helm` and `jq` rather than the ones in the Docker container.

This didn't fail in most places, as Cloud Build, and many of our dev machines have these installed already, and the command is set to be ignored if it fails for whatever reason (as a safety measure).

I only noticed the output on screen when helping a teammate debug some other issues.

N/A

N/A


roberthbailey (approved):
Gotta love quotes in shell.

PR #2681, opened 2022-07-25
TLS-Memoryleak

### LINKED ISSUE #2278: FleetAutoscaler keeps alive all TLS connections permanently causing memory leak on webhook server
**What happened**:
Over time my `https` server which hosts the FleetAutoscaler webhook goes OOM. This is caused by 1000s of never dying sockets on the server. This does NOT happen when I call it with cURL or a browser. It only happens with Agones calls the endpoint.

```
/app $ lsof -p $PID | grep socket
...
1       /app/zeus-rest  socket:[289294771]
1       /app/zeus-rest  socket:[289294783]
1       /app/zeus-rest  socket:[289292336]
1       /app/zeus-rest  socket:[289291653]
1       /app/zeus-rest  socket:[289291654]
1       /app/zeus-rest  socket:[289293769]
1       /app/zeus-rest  socket:[289294780]
...

/app $ lsof -p $PID | grep socket | wc -l
6397
/app $ lsof -p $PID | grep socket | wc -l
6403
/app $ lsof -p $PID | grep socket | wc -l
6418
```

I expect that when the FleetAutoscaler is called by Agones is either reuses the TLS client or it disconnects it. Keeping it alive and then making a new one seems naughty.

Create a TLS FleetAutoscaler endpoint with keepalive turned on and no timeout specified and watch the sockets multiply.

I suspect that this could be repaired by adding to 
`pkg/fleetautoscalers/fleetautoscalers.go`
```
var client = http.Client{
	Timeout: 15 * time.Second,
+++	Transport: &http.Transport{
+++                DisableKeepAlives: true,
+++        },
}
```
I fixed it by disabling KeepAlive on the server side. But it took me several hours to figure out the problem because I could not reproduce it with any clients of my own.

**Environment**:
- Agones version: 1.16
- Kubernetes version (use `kubectl version`): 1.21
- Cloud provider or hardware configuration: EKS and Minikube
- Others:

SaitejaTamma:
**What type of PR is this?**
>
> /kind breaking

 /kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Closes #2278 


pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -115,11 +115,8 @@ func setCABundle(caBundle []byte) error {
 	if ok := rootCAs.AppendCertsFromPEM(caBundle); !ok {
 		return errors.New("no certs were appended from caBundle")
 	}
-	client.Transport = &http.Transport{
-		TLSClientConfig: &tls.Config{
-			RootCAs: rootCAs,
-		},
-	}
+	TLSClientConfig := &tls.Config{}
```
markmandel, line 118: What we need to do is update this `client` global variable here:
https://github.com/googleforgames/agones/blob/0ec68ff6953e30868615f3ca333876bce7f3b894/pkg/fleetautoscalers/fleetautoscalers.go#L38

What is happening here is you are changing a local variable now, which isn't doing anything specific.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -115,11 +119,7 @@ func setCABundle(caBundle []byte) error {
 	if ok := rootCAs.AppendCertsFromPEM(caBundle); !ok {
 		return errors.New("no certs were appended from caBundle")
 	}
-	client.Transport = &http.Transport{
-		TLSClientConfig: &tls.Config{
-			RootCAs: rootCAs,
-		},
-	}
+	tlsConfig.RootCAs = rootCAs
```
markmandel, line 122: 🤔 

Does it still work if we do `client.Transport.TLSCClientConfig.RootCAs = rootCAs` instead? (then we could inline `tlsConfig` into `client` in the `var` declaration above?

SaitejaTamma: I tried this approach as well but instance is not created by default, TLSclientconfig do not present when we assign align like that. By default it is having the round trip instance not the transport instance.

markmandel: What if above becomes:

```go
var client = http.Client{
	Timeout: 15 * time.Second,
	Transport: &http.Transport{
		TLSClientConfig: &tls.Config{},
	},
}
```

Does that resolve the issue?

SaitejaTamma: If we  dont have a tlsConfig variable we will not be able to reference it to update the rootCAs at https://github.com/googleforgames/agones/blob/main/pkg/fleetautoscalers/fleetautoscalers.go#L118-L121.

markmandel: Oh I see the issue now, `http.RoundTripper` is an interface, so you can't access the variables within it.

I was playing with it here: https://go.dev/play/p/C6RTLx9VfWl

markmandel (approved):

PR #2695, opened 2022-08-03
Move StateAllocationFilter to beta

### LINKED ISSUE #2688: Move StateAllocationFilter to Beta
Let's track down the issues that are blocking the StateAllocationFilter from moving to Beta in here.

**Additional context**

- https://agones.dev/site/docs/guides/feature-stages/#feature-gates
- https://agones.dev/site/docs/integration-patterns/high-density-gameservers/
- https://agones.dev/site/docs/integration-patterns/player-capacity/

katsew:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This PR promotes the StateAllocationFilter feature from alpha to beta.

Closes #2688

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -244,7 +244,7 @@ steps:
 
 - name: 'e2e-runner'
   args:
-    - 'PlayerTracking=true&StateAllocationFilter=true&PlayerAllocationFilter=true&SDKGracefulTermination=true&CustomFasSyncInterval=false'
+    - 'PlayerTracking=true&StateAllocationFilter=false&PlayerAllocationFilter=true&SDKGracefulTermination=true&CustomFasSyncInterval=false'
     - 'e2e-test-cluster'
     - "${_REGISTRY}"
   id: e2e-feature-gates
--- a/pkg/allocation/converters/converter_test.go
+++ b/pkg/allocation/converters/converter_test.go
@@ -143,7 +143,8 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
 			},
 		},
 		{
-			name: "all fields are set",
+			name:     "all fields are set",
+			features: fmt.Sprintf("%s=false&%s=false", runtime.FeaturePlayerAllocationFilter, runtime.FeatureStateAllocationFilter),
 			in: &pb.AllocationRequest{
 				Namespace: "ns",
 				MultiClusterSetting: &pb.MultiClusterSetting{
@@ -243,7 +244,8 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
 			},
 		},
 		{
-			name: "empty fields to GSA",
+			name:     "empty fields to GSA",
+			features: fmt.Sprintf("%s=false&%s=false", runtime.FeaturePlayerAllocationFilter, runtime.FeatureStateAllocationFilter),
 			in: &pb.AllocationRequest{
 				Namespace:                    "",
 				MultiClusterSetting:          &pb.MultiClusterSetting{},
@@ -286,6 +288,29 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
 					Required: allocationv1.GameServerSelector{
 						GameServerState: &ready,
 					},
+					Scheduling: apis.Distributed,
+				},
+			},
+		},
+		{
+			name:     "empty fields to GSA with selectors fields",
+			features: fmt.Sprintf("%s=false&%s=false", runtime.FeaturePlayerAllocationFilter, runtime.FeatureStateAllocationFilter),
+			in: &pb.AllocationRequest{
+				Namespace:           "",
+				MultiClusterSetting: &pb.MultiClusterSetting{},
+				GameServerSelectors: []*pb.GameServerSelector{{}},
+				Scheduling:          pb.AllocationRequest_Distributed,
+				Metadata:            &pb.MetaPatch{},
+			},
+			want: &allocationv1.GameServerAllocation{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "",
+				},
+				Spec: allocationv1.GameServerAllocationSpec{
+					MultiClusterSetting: allocationv1.MultiClusterSetting{
+						Enabled: false,
+					},
+					Selectors:  []allocationv1.GameServerSelector{{}},
 					Scheduling: apis.Distributed,
 				},
 			},
@@ -351,6 +376,9 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
 					MultiClusterSetting: allocationv1.MultiClusterSetting{
 						Enabled: false,
 					},
+					Required: allocationv1.GameServerSelector{
+						GameServerState: &ready,
+					},
 					Scheduling: apis.Distributed,
 					MetaPatch: allocationv1.MetaPatch{
 						Labels: map[string]string{
@@ -388,6 +416,9 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
 						Enabled: false,
 					},
 					Scheduling: apis.Distributed,
+					Required: allocationv1.GameServerSelector{
+						GameServerState: &ready,
+					},
 					MetaPatch: allocationv1.MetaPatch{
 						Labels: map[string]string{
 							"a": "b",
--- a/site/content/en/docs/Guides/feature-stages.md
+++ b/site/content/en/docs/Guides/feature-stages.md
@@ -24,6 +24,7 @@ that can be found in the [Helm configuration]({{< ref "/docs/Installation/Instal
 
 The current set of `alpha` and `beta` feature gates are:
 
+{{% feature expiryVersion="1.26.0" %}}
 | Feature Name | Gate    | Default | Stage | Since |
 |--------------|---------|---------|-------|-------|
 | Example Gate (not in use) | `Example` | Disabled | None | 0.13.0 |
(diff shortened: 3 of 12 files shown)


cloudbuild.yaml
```diff
@@ -244,7 +244,7 @@ steps:
 
 - name: 'e2e-runner'
   args:
-    - 'PlayerTracking=true&StateAllocationFilter=true&PlayerAllocationFilter=true&SDKGracefulTermination=true'
+    - 'PlayerTracking=true&PlayerAllocationFilter=true&SDKGracefulTermination=true'
```
markmandel, line 247: Thanks for doing this!

```suggestion
    - 'PlayerTracking=true&StateAllocationFilter=false&PlayerAllocationFilter=true&SDKGracefulTermination=true'
```

Basically, we want to test the reverse.

Which makes me also realise that `CustomFasSyncInterval` should be `false` in there too. I'll enter that as a separate PR (unless you want to sneak it in).

This should also mean that the changes to the e2e tests below shouldn't be necessary (I don't think).

markmandel:
A unit test failed 🤔 we should see if that's reproducible or a new flake.

cloudbuild.yaml
katsew: I see. I'll reverse the flag and revert the test if it's unnecessary.

> Which makes me also realise that CustomFasSyncInterval should be false in there too. I'll enter that as a separate PR (unless you want to sneak it in).
I could sneak it in, but I think it's better to release in 1.25.0 if still make it in time, or release in patch version like 1.25.1.
What do you think?

katsew: > This should also mean that the changes to the e2e tests below shouldn't be necessary (I don't think).

The changes that I made is for unit tests running by `make test-go` not e2e tests, so It seems to be necessary.

katsew:
> A unit test failed 🤔 we should see if that's reproducible or a new flake.

Since failed test is related to gameserverallocation, looks like it related to my commit...? 🤔 
But also it seems to be a new flake.
What can I do for it?

cloudbuild.yaml
markmandel: > I could sneak it in, but I think it's better to release in 1.25.0 if still make it in time, or release in patch version like 1.25.1.

Don't worry about it - it's a small fix, I'll take care of it.

> The changes that I made is for unit tests running by make test-go not e2e tests, so It seems to be necessary.
Oh so it was! My bad. Ignore that comment then.

pkg/allocation/converters/converter_test.go
```diff
@@ -203,20 +203,25 @@ func TestConvertAllocationRequestToGameServerAllocation(t *testing.T) {
 							},
 						},
 					},
-					Required: allocationv1.GameServerSelector{LabelSelector: metav1.LabelSelector{
-						MatchLabels: map[string]string{
-							"c": "d",
+					Required: allocationv1.GameServerSelector{
```
markmandel, line 206: For this specific test group `"all fiends are set"` - rather than add `GameServerState` to everything, let's add a 

```
features: fmt.Sprintf("%s=false&%s=false", runtime.FeaturePlayerAllocationFilter, runtime.FeatureStateAllocationFilter),
```

To [this block](https://github.com/googleforgames/agones/pull/2695/files#diff-5b3da3128a01faaf200ec681805dab5edced5aeb32314af0647488b6a95b50c7R146). This test block is a repeat of `"all fields are set (StateAllocationFilter, PlayerAllocationFilter)"` but with the feature flags flipped (which by default they used to be) - so we should maintain the unit tests of the Alpha features.

We should do the same for any other tests that follow the same pattern (specifically turn off the feature flag for the original version of the test).

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -29,7 +29,7 @@ The current set of `alpha` and `beta` feature gates are:
 | Example Gate (not in use) | `Example` | Disabled | None | 0.13.0 |
 | [Player Tracking]({{< ref "/docs/Guides/player-tracking.md" >}}) | `PlayerTracking` | Disabled | `Alpha` | 1.6.0 |
 | [Custom resync period for FleetAutoscaler](https://github.com/googleforgames/agones/issues/1955) | `CustomFasSyncInterval` | Enabled | `Beta` | 1.25.0 |
-| [GameServer state filtering on GameServerAllocations](https://github.com/googleforgames/agones/issues/1239) | `StateAllocationFilter` | Disabled | `Alpha` | 1.14.0 |
+| [GameServer state filtering on GameServerAllocations](https://github.com/googleforgames/agones/issues/1239) | `StateAllocationFilter` | Enabled | `Beta` | 1.26.0 |
```
markmandel, line 32: Our documentation is updated on PR merge, so please see: https://agones.dev/site/docs/contribute/#within-a-page on how to only show this change on the new version of Agones.

You'll want to create two copies of the table, one with the old version in a `feature` shortcode that expires with 1.26.0 and one that publishes with `1.26.0`.

site/content/en/docs/Integration Patterns/high-density-gameservers.md
```diff
@@ -7,7 +7,7 @@ description: >
   How to run multiple concurrent game sessions in a single GameServer process.
 ---
 
-{{< alpha title="Allocation State Filter" gate="StateAllocationFilter" >}}
+{{< beta title="Allocation State Filter" gate="StateAllocationFilter" >}}
```
markmandel, line 10: These changes will also need to be `feature` shortcoded, such that they only get updated when we do our next release.

site/content/en/docs/Integration Patterns/player-capacity.md
```diff
@@ -8,8 +8,9 @@ description: >
 ---
 
 {{< alpha 
-    title="Player Tracking, Allocation Player Filter, and Allocation State Filter" 
-    gate="PlayerTracking,PlayerAllocationFilter,StateAllocationFilter" >}}
+    title="Player Tracking and Allocation Player Filter"
```
markmandel, line 11: These changes will also need to be `feature` shortcoded, such that they only get updated when we do our next release.

site/content/en/docs/Reference/gameserverallocation.md
```diff
@@ -37,7 +37,7 @@ spec:
         game: my-game
       matchExpressions:
         - {key: tier, operator: In, values: [cache]}
-      # [Stage:Alpha]
+      # [Stage:Beta]
```
markmandel, line 40: These changes will also need to be `feature` shortcoded, such that they only get updated when we do our next release.

site/content/en/docs/Reference/gameserverallocation.md
```diff
@@ -80,7 +80,7 @@ spec:
       game: my-game
     matchExpressions:
       - {key: tier, operator: In, values: [cache]}
-    # [Stage:Alpha]
+    # [Stage:Beta]
```
markmandel, line 83: These changes will also need to be `feature` shortcoded, such that they only get updated when we do our next release.

You may want to make a copy of the entire example rather than try and do it within specific lines.

markmandel:
> Since failed test is related to gameserverallocation, looks like it is related to my commit...? 
> But also it seems to be a flake. (I'm not sure it is existing flaky test or new)
> What can I do for it?

I had a look at the code - I don't think it's related to your change. Just making a note here on which test it is, and we'll see how often it pops up.

```
--- FAIL: TestControllerAllocator (2.64s)
    --- FAIL: TestControllerAllocator/successful_allocation (2.63s)
        controller_test.go:118: 
            	Error Trace:	controller_test.go:118
            	            				controller_test.go:125
            	Error:      	Should be true
            	Test:       	TestControllerAllocator/successful_allocation
            	Messages:   	Failed: UnAllocated vs Allocated
```

So for now, I suggest we ignore it.

site/content/en/docs/Guides/feature-stages.md
katsew: I must admit that I erased them when I resolved conflicts 🤦
I'll fix it and other docs that follow the same pattern, Thanks.

katsew: I added feature shortcode.

site/content/en/docs/Integration Patterns/high-density-gameservers.md
katsew: I added `feature` shortcode.

site/content/en/docs/Reference/gameserverallocation.md
katsew: I added `feature` shortcode wrapping `{{< tabpane >}}{{< /tabpane >}} `, 
because the document did not build as intended if I added it inside tabpane.

site/content/en/docs/Reference/gameserverallocation.md
katsew: I added `feature` shortcode like as you said.

pkg/allocation/converters/converter_test.go
katsew: I fixed tests and I also added a missing test.
Please let me know if I'm doing wrong.

site/content/en/docs/Integration Patterns/player-capacity.md
katsew: I added `feature` shortcode.

markmandel:
I'm 99% sure this is good to be approved - but could you resolve the conflict, that would be awesome, since then we can run it through CI with the resolved state.

katsew:
I resolved the conflict and also rebase commits down to a single commit.

markmandel (approved):

markmandel:
Flake: timeout in 

```
agones.dev/agones/test/e2e.TestFleetRollingUpdate.func1(0xc0009f6680)
	/go/src/agones.dev/agones/test/e2e/fleet_test.go:436 +0x165a
testing.tRunner(0xc0009f6680, 0xc0009a00c0)
```

e2e-stable.

PR #2712, opened 2022-08-18
Exposes metrics ports on pods in order to enable GCP Managed Prometheus

austin-space:
**What type of PR is this?**
/kind feature

**What this PR does / Why we need it**: GKE managed prometheus(GMP) makes some *interesting* decisions in how it implements a prometheus operator-like system. The most notable change is that there is no concept of a `ServiceMonitor` in GMP, only a `PodMonitoring` custom resource. As the name implies this monitor only has visibility on pods, not onto services. Fortunately I can setup `PodMonitoring` resources that closely imitate the `ServiceMonitor` resources in the included helm charts, the only issue is that the metrics port for the allocator service is not defined on the pod itself, just on the metrics service.

This is the least intrusive way of introducing this. Alternatively a GMP flag could be included in the values and I could add all of the pod monitors as well. 

**Which issue(s) this PR fixes**: none that I'm aware of


markmandel:
Thanks for digging into this! I'm now also learning how Google Cloud Managed Prometheus works!

i think in general this looks like the correct approach (exposing the container ports on the relevant pods)

It's worth noting that we have a metrics endpoint on the controller as well as the allocation system (controller's [service monitor](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/service-monitor.yaml)) - so this will need to cover both.

🤔 another thought I had, not sure if it sticks, but does it make any sense to allow someone to add arbitrary container ports to either the controller or the allocation Pods? Much in the same way we do `agones.controller.annotations` for example.

Just trying to think more generically, rather than a specific solution for this specific problem.

(Also docs are good 😄 )

WDYT?

austin-space:
Hey Mark, the controller metrics port is actually [already exposed](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/controller.yaml#L137) since it's the same port that is used for its other http traffic. There's probably some work that can be done to unify the way that these ports are setup/exposed since some have names and port numbers coming from the `values.yaml` file, and others are hardcoded in the templates. I can update the docs in this PR later today or tomorrow when I get some time. For docs are you thinking it would be better to have a quick blurb on GMP(something like "if you are using Google managed prometheus, you will need to set ... and set up your own podmonitoring resources(link to google docs)") or something more? Also should the podmonitoring resources required for managed prometheus be included in the helm chart(behind a `usingGoogleManagedPrometheus` flag or something)? I don't know how much you do or don't want to avoid provider specific stuff sneaking in there.

Thinking about the more generic solution: outside of this port, I'm not sure what other ports even have something interesting running on them and aren't exposed on the pod by default(as opposed to a service) so I don't know if that would be of too much value at this point in time(but might be if another one of these crops up).

markmandel:
That 100% makes sense.

My thought here then, would be to just leave the container port always open (i.e. don't bother putting in a helm configuration variable). It's inside the cluster anyway, so I don't think it really matters. @roberthbailey WDYT?

Regarding documentation - good point re: platform specific.

I'm thinking something generic like "if your metric collection agent needs to scrape container ports directly (such as with Google Cloud Managed Prometheus), the ports you would need to scrape can be found at {insert details}"

How does that sound?

roberthbailey:
> My thought here then, would be to just leave the container port always open (i.e. don't bother putting in a helm configuration variable). It's inside the cluster anyway, so I don't think it really matters. @roberthbailey WDYT?

I agree that it should be fine to just add the port all the time. If nothing scrapes it then it doesn't do any harm, but it's there if someone wants to scrape it (using GMP or a different prometheus scraper that uses `PodMonitoring`) without needing to re-install / re-configure Agones later.

markmandel:
Sounds like we have consensus!

If we could add some docs with a [feature tag](https://agones.dev/site/docs/contribute/#within-a-page) around it for the next version, that would be perfect 👍🏻

markmandel:
Just a heads up, we are one week away from our release candidate, so if you have time to implement the above comments, it would be awesome to get that in for that release.

austin-space:
Sorry, the week got away from me. I think everything should be good now. Let me know if you want any changes with the little blurb I added in the docs.

site/content/en/docs/Guides/metrics.md
```diff
@@ -17,6 +17,10 @@ We plan to support multiple exporters in the future via environment variables an
 
 If you are running a [Prometheus](https://prometheus.io/) instance you just need to ensure that metrics and kubernetes service discovery are enabled. (helm chart values `agones.metrics.prometheusEnabled` and `agones.metrics.prometheusServiceDiscovery`). This will automatically add annotations required by Prometheus to discover Agones metrics and start collecting them. (see [example](https://github.com/prometheus/prometheus/tree/main/documentation/examples/kubernetes-rabbitmq))
 
+{{% feature publishVersion="1.26.0" %}}
+If your Prometheus metrics collection agent requires that you scrape from the pods directly(such as with Google Cloud Managed Prometheus), then the metrics ports for the controller and allocator will both be named `http` and exposed on `8080`. In the case of the allocator, the port name and number can be overriden with the `agones.allocator.serviceMetrics.http.portName` and `agones.allocator.serviceMetrics.http.port` helm chart values.
```
markmandel, line 21:
```suggestion
If your Prometheus metrics collection agent requires that you scrape from the pods directly(such as with [Google Cloud Managed Prometheus](https://cloud.google.com/stackdriver/docs/managed-prometheus)), then the metrics ports for the controller and allocator will both be named `http` and exposed on `8080`. In the case of the allocator, the port name and number can be overriden with the `agones.allocator.serviceMetrics.http.portName` and `agones.allocator.serviceMetrics.http.port` helm chart values.
```

This would be my only feedback - otherwise, this looks good to go 👍🏻

markmandel:
```
13943a13944
>           containerPort:  8080
14156a14158
>         - name: http
```

oooh, I see what it is.

If you could run `make gen-install` in the `./build` directory (assuming you have Make and Docker installed) that will refresh the install.yaml to have your changes, and this should be good to go 👍🏻

markmandel:
Lemme know if you run into any issues doing that, and I can do it on my end, and submit a PR to your PR 😄

markmandel (approved):

Issue #2713, opened 2022-08-23
C# SDK Passing CancellationTokenSource as a constructor argument raises a NullReferenceException.


g2-ochiai-yuta:
**What happened**:
```C#
       public AgonesSDK(double requestTimeoutSec = 15.0, SDK.SDKClient sdkClient = null, CancellationTokenSource cancellationTokenSource = null, ILogger logger = null)
        {
            _logger = logger;
            RequestTimeoutSec = requestTimeoutSec;
            if (cancellationTokenSource == null)
            {
                cts = new CancellationTokenSource();
                ownsCts = true;
            }
            else
            {
                ownsCts = false;
            }

            // NullReferenceException is raised because the cts variable is not initialized.
            ctoken = cts.Token;
            channel = new Channel(Host, Port, ChannelCredentials.Insecure);
            client = sdkClient ?? new SDK.SDKClient(channel);
            alpha = new Alpha(channel, requestTimeoutSec, cancellationTokenSource, logger);
        }
```
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
```C#
IAgonesSDK sdk = new AgonesSDK(cancellationTokenSource: new CancellationTokenSource());
```

**Anything else we need to know?**:
https://github.com/googleforgames/agones/commit/60545e03ce3f45a629ab7f05d8000e383c05d4c9
![image](https://user-images.githubusercontent.com/7463440/186244694-7fa0a67b-fe96-4472-995d-30ba51cead37.png)

**Environment**:
- Agones version:  1.25
- Kubernetes version (use `kubectl version`): v1.23.5
- Cloud provider or hardware configuration: local (kind)
- Install method (yaml/helm): yaml
- Troubleshooting guide log(s):
- Others:

markmandel:
This was implemented in #2714 - not sure why it's open. Closing.

PR #2714, opened 2022-08-24
Bug: Passing arguments to the constructor results in an error

g2-ochiai-yuta:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

NullReferenceException is raised because the cts variable is not initialized.

Closes # https://github.com/googleforgames/agones/issues/2713

### DIFF
--- a/sdks/csharp/sdk/AgonesSDK.cs
+++ b/sdks/csharp/sdk/AgonesSDK.cs
@@ -69,6 +69,7 @@ namespace Agones
 			}
 			else
 			{
+				cts = cancellationTokenSource;
 				ownsCts = false;
 			}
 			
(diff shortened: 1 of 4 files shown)


roberthbailey:
Can you add a unit test that shows the fix (e.g. that would throw the NullReferenceException without your change to the source files but passes with them)? Having a test will prevent any regressions in the future.

g2-ochiai-yuta:
@roberthbailey 
I added unit test.

roberthbailey:
Thanks!

I've verified by running `make test-sdk SDK_FOLDER=csharp` that without the change the test fails. :)

roberthbailey (approved):

PR #2740, opened 2022-09-14
Implemented PodDisruptionBudget on relevant deployments

valentintorikian:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix

Added (optional) PodDisruptionBudget creation for the `ping` and `allocator` deployment. Disabled by default.
Will allow for better management of voluntary disruptions.

N/A

N/A


roberthbailey:
Thanks for the PR! I will take a look at it once the in-progress release is finalized (also, you will probably have to rebase and deal with conflicts in the documentation changes at that point).

roberthbailey:
@valentintorikian - As expected, you now need to rebase your documentation changes.

valentintorikian:
@roberthbailey  rebased

markmandel:
@roberthbailey just gently bumping if this is good to go? Looks like you were reviewing 😃

install/helm/agones/templates/ping.yaml
```diff
@@ -89,6 +89,22 @@ spec:
...
+{{- end }}
+{{- if .Values.agones.ping.pdb.enabled }}
+---
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: agones-ping-pdb
+spec:
+  minAvailable: {{ .Values.agones.ping.pdb.minAvailable }}
+  maxUnavailable: {{ .Values.agones.ping.pdb.maxUnavailable }}
```
roberthbailey, line 101: According to https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget the above two fields are mutually exclusive (e.g. you can only set one or the other). 

It isn't clear to me what happens if the spec looks like:

```
spec:
  minAvailable: 2
  maxUnavailable: 
```

which is what will be produced if the boolean to enable this is toggled to true. Is that spec accepted by k8s or rejected?

Does the template here need to be more complicated so that we only set a single value?

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -211,13 +211,20 @@ The following tables lists the configurable parameters of the Agones chart and t
...
-| -------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- | ---------------------- |
+
+| Parameter                             | Description                                                                                                                                                                                | Default |
+|---------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| `agones.allocator.pdb.enabled`        | Set to `true` to enable the creation of a [PodDisruptionBudget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) for the allocator deployment                              | `false` |
+| `agones.allocator.pdb.minAvailable`   | Description of the number of pods from that set that must still be available after the eviction, even in the absence of the evicted pod. Can be either an absolute number or a percentage. | `1`     |
```
roberthbailey, line 223: In each of the lines for minAvailable and maxUnavailable can you note that the two fields are mutually exclusive? Following the link to PDBs you can find this but it would be nice to highlight it here as well.

markmandel:
Just a gentle reminder that our Release Candidate comes out next week, so if you would like this in our next release, Tuesday next week is the deadline.

install/helm/agones/templates/ping.yaml
zmerlynn: I was curious, so I tested this. It works fine, and sends the JSON you might expect: `"spec":{"maxUnavailable":null,"minAvailable":0, [...]}`.

roberthbailey: Thanks @zmerlynn for checking that out. Do you think this is ok to submit as-is then?

zmerlynn: I think it should still be documented, but we could also merge and come back to that later.

zmerlynn (approved):
@roberthbailey I'm happy to fix up the documentation on mutual-exclusivity after this goes in.

zmerlynn:
> @roberthbailey I'm happy to fix up the documentation on mutual-exclusivity after this goes in.

It looks like we could even use an approach like https://stackoverflow.com/questions/53508229/helm-chart-with-dependent-values to just ensure it.

roberthbailey (approved):

PR #2742, opened 2022-09-16
Added a title to the 1.26.0 release

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Just nice to have something in the title 😃

Kept the slug the same, so the URL won't break.

N/A

Apologies for not catching this in PR review.


roberthbailey (approved):
Thanks for catching this.

PR #2754, opened 2022-09-28
If the user has specified cluster autoscaling behavior for their gameserver then don't overwrite it

### LINKED ISSUE #2747: Allow cluster autoscaler to scale down game server pods
[This is paraphrased from a slack conversation]

As described in https://agones.dev/site/docs/advanced/scheduling-and-autoscaling/#cluster-autoscaler-1, we currently prevent the cluster autoscaler from removing game server pods by adding the annotation `"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"`. 

For game servers that are OK being terminated (even while in the allocated state), we should make it possible for the cluster autoscaler to remove the pods to compact the cluster by removing underutilized nodes. Note that this imposes a pretty strict burden on the game server & it's configuration to have properly set graceful termination and for the game server to be able to actually gracefully terminate and finish player sessions within the configured time. As a result, we should err on the side of not enabling this by default and considering it an advanced use case. 

A helm configuration option to change the boolean value for this particular annotation. 

Note that removing the annotation will not have the desired effect, because the cluster autoscaler does not understand that Agones-managed pods are managed and since it believes that they are un-replicated bare pods it will not remove them. Instead, we need to change the annotation to `"cluster-autoscaler.kubernetes.io/safe-to-evict": "true"` to let the autoscaler know that it is safe to remove the pods. 

Instead of allowing the cluster autoscaler to compact the cluster and move workloads, we could write a custom controller that looks at the cluster and shuts down any game servers that can be evicted thereby compacting it from there. This would largely duplicate the functionality of the cluster autoscaler, but we could potentially add some Agones-aware logic to better choose which game servers to evict. 

**Additional context**
_n/a_

roberthbailey:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

This change allows the object metadata annotation that controls the cluster autoscaler behavior to be passed along through the pod template to the game server pod. Previously, the annotation was always set by Agones, which would override any attempt by a user to set this annotation. 

This change allows game servers to mark themselves as being able to be preempted and removed by the cluster autoscaler, which allows the cluster autoscaler to compact nodes when there is space to downsize a cluster. 

Closes #2747

### DIFF
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -99,6 +99,10 @@ const (
 	// becomes ready, so we can track when restarts should occur and when a GameServer
 	// should be moved to Unhealthy.
 	GameServerReadyContainerIDAnnotation = agones.GroupName + "/ready-container-id"
+	// PodSafeToEvictAnnotation is an annotation that the Kubernetes cluster autoscaler uses to
+	// determine if a pod can safely be evicted to compact a cluster by moving pods between nodes
+	// and scaling down nodes.
+	PodSafeToEvictAnnotation = "cluster-autoscaler.kubernetes.io/safe-to-evict"
 )
 
 var (
@@ -632,7 +636,7 @@ func (gs *GameServer) podObjectMeta(pod *corev1.Pod) {
 		pod.ObjectMeta.Labels = make(map[string]string, 2)
 	}
 	if pod.ObjectMeta.Annotations == nil {
-		pod.ObjectMeta.Annotations = make(map[string]string, 1)
+		pod.ObjectMeta.Annotations = make(map[string]string, 2)
 	}
 	pod.ObjectMeta.Labels[RoleLabel] = GameServerLabelRole
 	// store the GameServer name as a label, for easy lookup later on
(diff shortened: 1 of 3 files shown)


pkg/apis/agones/v1/gameserver.go
```diff
@@ -644,8 +644,11 @@ func (gs *GameServer) podObjectMeta(pod *corev1.Pod) {
 
 	if gs.Spec.Scheduling == apis.Packed {
 		// This means that the autoscaler cannot remove the Node that this Pod is on.
-		// (and evict the Pod in the process)
-		pod.ObjectMeta.Annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"] = "false"
+		// (and evict the Pod in the process). Only set the value if it has not already
+		// been configured in the pod template (to not override user specified behavior).
+		if _, exists := pod.ObjectMeta.Annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"]; !exists {
```
markmandel, line 649: Should "cluster-autoscaler.kubernetes.io/safe-to-evict" be a const? Might make it a bit easier to read.

As a possible other simplification (suggesting a possible const):

```go
if  _, exists := pod.ObjectMeta.Annotations[safeToEvicAnnotation]; !exists && gs.Spec.Scheduling == apis.Packed {
   pod.ObjectMeta.Annotations[safeToEvicAnnotation] = "false"
}
```

Make it one line? WDYT?

roberthbailey: I like that. I picked a name for the constant that was a bit longer but more closely matched the name used by the cluster autoscaler implementation (although they used "Key" instead of "Annotation" as the last part of their name which I found less descriptive).

site/content/en/docs/Advanced/scheduling-and-autoscaling.md
```diff
@@ -97,13 +97,42 @@ This affects the Cluster autoscaler, Allocation Scheduling, Pod Scheduling and F
...
 gameplay, Agones adds the annotation [`"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"`](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node)
 to the backing Pod.
+{{% /feature %}}
+
+{{% feature publishVersion="1.27.0" %}}
+When using the “Packed” strategy, Agones will ensure that the Cluster Autoscaler doesn't attempt to evict and move `GameServer` `Pods` onto new Nodes during
+gameplay by adding the annotation [`"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"`](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node)
+to the backing Pod.
+
+However, if a gameserver can tolerate being evicted (generally in combination with setting an appropriate graceful termination period
```
markmandel, line 111:
```suggestion
However, if a gameserver can tolerate [being evicted](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works) (generally in combination with setting an appropriate graceful termination period
```

Nit: Linking back to K8s documentation so that "eviction" is clear.

roberthbailey: good idea.

markmandel (approved):

PR #2644, opened 2022-06-25
Migrade from github.com/golang/protobuf to google.golang.org/protobuf

govargo:
**What type of PR is this?**

/kind breaking
/kind cleanup
/kind feature

**What this PR does / Why we need it**:

I found that the `github.com/golang/protobuf` will be deprecated in the future when I executed `make gen-all-sdk-grpc`.

```
+ protoc -I /go/src/agones.dev/agones/proto/googleapis -I /go/src/agones.dev/agones/proto/sdk sdk.proto --grpc-gateway_out=logtostderr=true:pkg/sdk
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
        A future release of golang/protobuf will delete this package,
        which has long been excluded from the compatibility promise.

+ protoc -I /go/src/agones.dev/agones/proto/googleapis -I /go/src/agones.dev/agones/proto/sdk/alpha alpha.proto --grpc-gateway_out=logtostderr=true:pkg/sdk/alpha
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
        A future release of golang/protobuf will delete this package,
        which has long been excluded from the compatibility promise.

+ protoc -I /go/src/agones.dev/agones/proto/googleapis -I /go/src/agones.dev/agones/proto/sdk/beta beta.proto --grpc-gateway_out=logtostderr=true:pkg/sdk/beta
beta.proto:20:1: warning: Import google/api/annotations.proto is unused.
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
        A future release of golang/protobuf will delete this package,
        which has long been excluded from the compatibility promise.
```

This PR has many changes for migration to `google.golang.org/protobuf`.

**Added**
1. Add proto files for `proto/grpc-gateway`
2. Add auto-generated files for grpc(`allocation_grpc.pb.go`, `alpha_grpc.pb.go`, `beta_grpc.pb.go`, `sdk_grpc.pb.go`, `protoc-gen-openapiv2/options/annotations.pb.h`, `protoc-gen-openapiv2/options/openapiv2.pb.h`, `protoc-gen-openapiv2/annotations.pb.cc`, `protoc-gen-openapiv2/openapiv2.pb.cc`)
4. Add `absl` installation for grpc build stability to sdk-base-image
5. Add `grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger` to proto files in order to define swagger.json definitions
[...]

PR #2757, opened 2022-09-30
Remove generation for swagger Go code and Add static swagger codes for test

govargo:
**What type of PR is this?**

/kind feature
/kind testing

This PR changes conformance test behaviour for REST API SDK.
Currently, REST API SDK test does the following steps.
1. Grab the current swagger file
2. Generate the Go client from that swagger file
3. Run the test with that generated client

With above steps, if the developer changes the swagger definition file, the test will success because step 2's generation code will be also executed.
So we cannot detect whether the changes of the swagger files cause the breaking or not.

In this PR, I change the test steps.
1. Generate swagger Go client code with manually
2. Run the test with that generated client

After this PR, the developer who changes the swagger definition file must execute the `make run-sdk-command COMMAND=gen SDK_FOLDER=restapi`

Related: #2644

None


govargo:
Test failed because the image `gcr.io/agones-images/restapi-simple-server:0.1` doesn't exist to the docker repository yet.

```
--- FAIL: TestGameServerRESTAPISimple (304.38s)
    gameserver_test.go:1215: Could not get a GameServer ready: waiting for {game-server [{udp-port Dynamic <nil> 7654 0 UDP}] {false 20 0 30}  { 0 0} {{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] [] [{game-server gcr.io/agones-images/restapi-simple-server:0.1 [] []  [] [] [] {map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}] map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}]} [] [] nil nil nil nil   IfNotPresent nil false false false}] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] [] <nil> nil}} <nil>} GameServer instance readiness timed out (): waiting for GameServer to be Ready 1664542313/game-server8jc6b: timed out waiting for the condition
FAIL
```

markmandel:
🤔 I'll be honest, I'm not sure this is necessary.

Reason being we have the SDK conformance tests - and it's exactly the same API surface on the local binary as it is with the remote - because it's actually the same binary.

What I think I'd instead recommend here is that we remove the runtime generation of the go code at test time with swagger - that way we know that if the swagger changes, the test code does not.

We can keep some kind of gen target/shell script if we ever want to manually refresh the test code from the swagger spec, but we could have a static implementation and leave it there.

This probably would be simpler, and require less moving parts to ensure the SDK API surface remains stable for #2644.

WDYT?

govargo:
I would like to confirm your idea first.

> we remove the runtime generation of the go code at test time with swagger - that way we know that if the swagger changes, the test code does not.

Are you referring to this test?
https://github.com/googleforgames/agones/blob/v1.26.0/build/build-sdk-images/restapi/sdktest.sh

> We can keep some kind of gen target/shell script if we ever want to manually refresh the test code from the swagger spec, but we could have a static implementation and leave it there.

Do you have an image of the test when doing it by the shell?
The following is the my image. If you have any images, please let me know.

```
# restapi-test.sh
# Ready test
curl -d "{}" -H "Content-Type: application/json" -X POST http://localhost:${AGONES_SDK_HTTP_PORT}/ready

# Health test
# curl -d "{}" -H "Content-Type: application/json" -X POST http://localhost:${AGONES_SDK_HTTP_PORT}/health

...(and all other methods)
```

markmandel:
Yep, totally makes sense to confirm first!

Yes - that's 100% the test I'm thinking of 👍🏻 

Compliance tests are built a little differently, but also self managed by the built toolchain.

You can see some (hopefully up to date) instructions here: https://github.com/googleforgames/agones/tree/v1.26.0/build/build-sdk-images

And you can see all the commands are here:
https://github.com/googleforgames/agones/blob/v1.26.0/build/includes/sdk.mk

My thought isn't to replace the REST API test, but instead tweak it.

Right now it does the following:

1. Grab the current swagger file
2. Generate the Go client from that swagger file
3. Run the test with that generated client

What I'm suggesting is, we change it to the following:
1. Remove the generation of the Go client from every run of the compliance test
2. Build a `gen` target that can be run manually as needed that vendors the Go swagger client.
3. On each test, run the Go code that is both the client and the test code Go code, that is static and doesn't change on each re-run.

Does that make more sense?

govargo:
Thank you your idea.

> 1. Remove the generation of the Go client from every run of the compliance test
> 2. Build a gen target that can be run manually as needed that vendors the Go swagger client.
> 3. On each test, run the Go code that is both the client and the test code Go code, that is static and doesn't change on each re-run.

I can imagene 1 and 2. Sterp 1 & 2 are about build toolchain refactor.
I cannot imagene about step 3. Is the `test code Go code` using [current test codes](https://github.com/googleforgames/agones/tree/main/test/sdk/restapi)?
Or should I create the new material in this PR?

markmandel:
What the compliance tests do is run the local SDK server in a "test mode" where is expects a set of commands - the code you linked, is sending all the commands to the local SDK server that it expects, which will then pass/fail depending on if it gets the commands it expects.

Basically because of linting and compilation failure, we put that linked Go code in a `.nolint` file, because we don't have the generated clients at linting time, so it all fails.

If we generate and keep the REST swagger client, then that problem should go away.

Does that make more sense?

markmandel:
You can see the test mode here: https://github.com/googleforgames/agones/blob/bf1b8a4b33fb04e01f80760c3db32b1c2dbd1b15/cmd/sdk-server/main.go#L108

govargo:
The generated swagger files failed due to lint problem.
I'll update this...

govargo:
Ready for review

build/build-sdk-images/restapi/gen.sh
```diff
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-mkdir /go/src/agones.dev/agones/swagger
 wget -q https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/2.4.10/swagger-codegen-cli-2.4.10.jar -O /tmp/swagger-codegen-cli.jar
```
markmandel, line 17: (non blocking question): Any desire to upgrade the swagger-codegen-cli ? No idea how much it matters.

Looks like this version came out in 2019

[2.4.10/](https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/2.4.10/)     2019-11-16 16:38

test/sdk/restapi/alpha/swagger/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
markmandel, line 1: Do we need this? If so, it needs an Apache licence.

test/sdk/restapi/alpha/swagger/git_push.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/sh
```
markmandel, line 1: 🤔 what's this for? Generated as well?

test/sdk/restapi/alpha/swagger/api/swagger.yaml
```diff
@@ -0,0 +1,203 @@
+---
```
markmandel, line 1: All this generated code will need a Apache header, and a note about it being generated.

There are some boilerplate files here: https://github.com/googleforgames/agones/tree/main/build for both go and yaml that we use for headers for generated code. Figured they might be useful.

Can't see if the swagger command line tool will do this for you, if not, you can steal from: https://github.com/googleforgames/agones/blob/main/build/build-sdk-images/go/gen.sh

markmandel (commented):
Looking good - a few comments inline.

Given that this is primarily for testing, can we drop the generation of all the .md files, maybe also the .sh files as well? It feels like a lot of noise. WDYT?

build/build-sdk-images/restapi/gen.sh
govargo: Updated to v3.0.35(latest)

test/sdk/restapi/alpha/swagger/git_push.sh
govargo: This is generated but un-necessary. Removed

test/sdk/restapi/alpha/swagger/api/swagger.yaml
govargo: Added Apache header to generated files.

markmandel:
No issues on my end - just wanted to check though, the two swagger.yaml files that are generated, do we need those, or could we clean those out too?

govargo:
> just wanted to check though, the two swagger.yaml files that are generated, do we need those, or could we clean those out too?

Updated
 * delete swagger.yaml
 * update Apache header for restapi/gen.sh from 2019 to 2022

markmandel (approved):

markmandel:
Love it!

PR #2760, opened 2022-10-12
Fix case of HTTP module reference

TroutZhang:
**What type of PR is this?**
/kind cleanup

This fixes the warning of the following in Unreal Engine:
"Module 'Http' (referenced via Target -> Agones.Build.cs) has incorrect text case. Did you mean 'HTTP'?"

### DIFF
--- a/sdks/unreal/Agones/Source/Agones/Agones.Build.cs
+++ b/sdks/unreal/Agones/Source/Agones/Agones.Build.cs
@@ -24,7 +24,7 @@ public class Agones : ModuleRules
 		PublicDependencyModuleNames.AddRange(new[]
 		{
 			"Core",
-			"Http",
+			"HTTP",
 			"Json",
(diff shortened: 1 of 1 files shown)


markmandel:
Calling all our unreal contributors 😄  @YannickLange , @domgreen , @highlyunavailable , @dotcom  - any issues?

Also, which version of Unreal are you seeing the warning on?

domgreen:
As per docs [here](https://docs.unrealengine.com/4.27/en-US/API/Runtime/HTTP/)  this change is correct :+1: 

Whilst I was there I checked if `Json`  should be `JSON` and upsetting that it shouldn't be changed [link](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Json/)

markmandel (approved):

markmandel:
Thanks @domgreen ! Approved!

PR #2765, opened 2022-10-18
Players in-game metric for when PlayerTracking is enabled

estebangarcia:
**What type of PR is this?**
/kind feature

Records a players in-game metric only when PlayerTracking is enabled.

We recently had a big playtest and needed a way to count players in-game in Prometheus.
We created a separate service that watches over all gameservers but thought that adding this as part of Agones made more sense.

**Which issue(s) this PR fixes**: #1035 (partially)


pkg/metrics/controller_metrics.go
```diff
@@ -52,6 +53,7 @@ var (
 	fasLimitedStats           = stats.Int64("fas/limited", "The fleet autoscaler is capped (0 indicates false, 1 indicates true)", "1")
 	gameServerCountStats      = stats.Int64("gameservers/count", "The count of gameservers", "1")
 	gameServerTotalStats      = stats.Int64("gameservers/total", "The total of gameservers", "1")
+	gameServerPlayersInGame   = stats.Int64("gameservers/players_in_game", "The total number of players connected to gameservers", "1")
```
markmandel, line 56: Per https://prometheus.io/docs/practices/naming/ I've suggest making this `gameserver_player_connected_total` for the total number of players.

Since you are in here, should we also add `gameserver_player_capacity_total` such that you can also track how much room there is over time?

site/content/en/docs/Guides/metrics.md
```diff
@@ -43,6 +43,7 @@ Follow the [Stackdriver Installation steps](#stackdriver-installation) to see yo
 | agones_gameservers_count                              | The number of gameservers per fleet and status                                                                                                                                              | gauge     |
 | agones_gameserver_allocations_duration_seconds        | The distribution of gameserver allocation requests latencies                                                                                                                                | histogram |
 | agones_gameservers_total                              | The total of gameservers per fleet and status                                                                                                                                               | counter   |
+| agones_gameservers_players_in_game                    | The total number of players connected to gameservers (Only available when [player tracking]({{< relref "player-tracking.md" >}}) is enabled)                                              | gauge     |
```
markmandel, line 46: Since docs are published on merge, we'll need you to create two instance of this table, each in `feature` shortcodes, one with the new metric(s) and one without.

That way the new metric will stay hidden until the next release 👍🏻 

See:

estebangarcia: Instead of duplicating the table, I added the `feature` shortcode within the existing one to hide the new metrics until Agones 1.28.
Let me know if this is suitable or if I should duplicate the table instead.

estebangarcia: Disregard the last comment. Had to duplicate the table as what I did wasn't working as expected.

markmandel:
Looks that way - retrying tests.

pkg/metrics/controller_metrics.go
```diff
@@ -42,19 +44,21 @@ var (
 	// fleetViews are metric views associated with Fleets
 	fleetViews = append([]string{fleetReplicaCountName, gameServersCountName, gameServersTotalName, gameServerStateDurationName}, fleetAutoscalerViews...)
```
markmandel, line 45: We'll want to add `gameServerPlayerConnectedTotal` and `gameServerPlayerCapacityTotal` to `fleetViews` - so that the system knows to reset them when Fleets get deleted,.

markmandel (commented):
Just saw one small thing to add, but otherwise this look good to go once we are out from feature freeze 👍🏻

pkg/metrics/controller_metrics.go
estebangarcia: Good catch. I just added that.

markmandel (approved):

Issue #2777, opened 2022-10-26
Support Agones on GKE Autopilot


zmerlynn:
**Is your feature request related to a problem? Please describe.**

Agones has a number of friction points on [GKE Autopilot](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview):
* Arbitrary host port selection is not allowed (`denied by autogke-no-host-port`)
* Use of the `cluster-autoscaler.kubernetes.io/safe-to-evict=false`  to minimize disruption (`denied by autogke-node-affinity-selector-limitation`):
  * On game server pods, Agones defaults to `cluster-autoscaler.kubernetes.io/safe-to-evict=false` annotation - this can be mitigated with #2747 
  * The Agones controller also defaults to `cluster-autoscaler.kubernetes.io/safe-to-evict=false', presumably because it is not HA and would like to minimize disruption, e.g. #1296 

Note that Autopilot issues were previously brought up in #2249 - the state of the world has changed considerably since then (mutating webhooks are now allowed), but there are still friction points.

**Describe the solution you'd like**

I propose the following:

- [x] We've recently added a feature to allow for Autopilot host port assignment using a special annotation. In order to allow this to be used with Agones, I propose:
  * A small cloud-provider library to allow for per-provider abstractions
  * In said cloud provider abstraction, we add a game server pod interceptor that can add the appropriate annotation
  * We also add a validator that the `portPolicy` is `Dynamic` (we may be able to support `Passthrough` later, but let's start here)
  * (This section is complete with #2782 and #2789)
- [ ] The controllers/webhooks/allocator can be made HA, at which point we won't need to annotate `safe-to-evict=false` (#2797).
- [x] So much testing.
- [ ] Add docs!

roberthbailey:
One question on

> On game server pods, Agones defaults to cluster-autoscaler.kubernetes.io/safe-to-evict=false annotation - this can be mitigated with https://github.com/googleforgames/agones/issues/2747

The current mitigation will require that any fleet spec being moved from non-autopilot to autopilot will need to be updated to add extra metadata. It would be nice in the longer term if we could better support upgrades (across all platforms) and switch the default behavior to allow a _safe_ eviction but give folks the chance to prevent it if that doesn't work for their game server.

zmerlynn:
> One question on
> 
> > On game server pods, Agones defaults to cluster-autoscaler.kubernetes.io/safe-to-evict=false annotation - this can be mitigated with #2747
> 
[...]

PR #2782, opened 2022-10-28
Sync Pod host ports back to GameServer in GCP

zmerlynn:
This is the start of the implementation for #2777:

* Most of this is mechanical and implements a thin cloud provider layer in pkg/cloud, instantiated with New(cloud). The cloud provider provides a single function so far: SyncPodPortsToGameServer.

* SyncPodPortsToGameServer is inserted as a hook while syncing IP/ports, to let different cloud providers handle port allocation slightly differently (in this case, GKE Autopilot)

* In GKE Autopilot, we look for a JSON string like `{"min":7000,"max":8000,"portsAssigned":{"7001":7737,"7002":7738}}` as an indication that the host ports were reassigned (per policy). As a side note to anyone watching, this is currently an unreleased feature. If we see this, we use the provided mapping to map the host ports in the GameServer.Spec.

With this change, it's possible to launch a GameServer and get a healthy GameServer Pod by adding the following annotation:

```
annotations:
  cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
  autopilot.gke.io/host-port-assignment: '{"min": 7000, "max": 8000}'
```

In future PRs, I will make this process automatic and not require the annotation.

/kind feature


pkg/cloud/cloud.go
```diff
@@ -0,0 +1,52 @@
...
+	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
+	"agones.dev/agones/pkg/cloud/gcp"
+	"agones.dev/agones/pkg/cloud/generic"
+	"agones.dev/agones/pkg/util/runtime"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Cloud provides a generic interface that abstracts cloud specific functionality.
+// Users should call New() to instantiate a specific cloud provider.
+type Cloud interface {
```
markmandel, line 29: The thought I had on this the other day was that this isn't going to be used per-_Cloud_, it's going to be used per _Cloud Product_.

I.e. The aim for this is Google Cloud Autopilot, not all of Google Cloud (it won't work on GKE for example), so I'm wondering if this should be more of a `Product` flag, rather than a `Cloud` flag?

zmerlynn: I was planning on layering in GCP and Autopilot detection and adding basically an "is autopilot" bool to the GCP provider for this case. (As the code stands right now, though, it won't do anything unless you force `--cloud-provider=gcp` _and_ have the right Pod annotation.)

(I am imagining we may want other provider functions for GKE Standard as well.)

markmandel: I'd be tempted to make it extremely explicit - so, something like `--cloud-product=gcp-autopilot` rather than doing auto detection.

zmerlynn: I can make it forceable for sure, but I understood one of the goals to be portability of configuration - which would lean towards auto-detection. If we differentiate at the product level, the configuration wouldn't even be portable between GKE Standard and Autopilot, without additional work on the install side.

zmerlynn: We talked offline about this: @markmandel's objection was more the granularity/naming of the abstraction vs the auto-detection. I agree with him, so I changed it to a --cloud-product / CloudProduct abstraction.

cmd/controller/main.go
```diff
@@ -264,6 +270,7 @@ func parseEnvFlags() config {
 	viper.SetDefault(logDirFlag, "")
 	viper.SetDefault(logLevelFlag, "Info")
 	viper.SetDefault(logSizeLimitMBFlag, 10000) // 10 GB, will be split into 100 MB chunks
+	viper.SetDefault(cloudProductFlag, "")
```
markmandel, line 273: Non blocking suggestion:

```suggestion
	viper.SetDefault(cloudProductFlag, "auto")
```

Only change if you like it.

pkg/gameservers/gameservers.go
```diff
@@ -72,14 +72,19 @@ func address(node *corev1.Node) (string, error) {
 
 // applyGameServerAddressAndPort gathers the address and port details from the node and pod
 // and applies them to the GameServer that is passed in, and returns it.
-func applyGameServerAddressAndPort(gs *agonesv1.GameServer, node *corev1.Node, pod *corev1.Pod) (*agonesv1.GameServer, error) {
+func applyGameServerAddressAndPort(gs *agonesv1.GameServer, node *corev1.Node, pod *corev1.Pod, syncPodPortsToGameServer func(*agonesv1.GameServer, *corev1.Pod) error) (*agonesv1.GameServer, error) {
```
markmandel, line 75: 🤔  curious why you are passing in the function, rather than passing through a `CloudProduct` interface? Easier testing?

pkg/gameservers/gameservers_test.go
```diff
@@ -107,23 +107,45 @@ func TestAddress(t *testing.T) {
...
-		require.NoError(t, err)
-		pod.Spec.NodeName = node.ObjectMeta.Name
+	noopMod := func(*corev1.Pod) {}
+	noopSyncer := func(*agonesv1.GameServer, *corev1.Pod) error { return nil }
+	for name, tc := range map[string]struct {
+		podMod       func(*corev1.Pod)
+		podSyncer    func(*agonesv1.GameServer, *corev1.Pod) error
+		wantHostPort int32
+	}{
+		"normal": {noopMod, noopSyncer, 9999},
```
markmandel, line 117: nice conversion to table based testing.

markmandel (commented):
Overall structure looks good! Just had a few small questions / comments. Nice work!

cmd/controller/main.go
zmerlynn: If I go that way, do you think the empty string should be invalid, then? (i.e. we rewrite --cloud-product to be set to something valid, but it defaults to `auto`)

pkg/gameservers/gameservers.go
zmerlynn: It was mostly for easier testing, yeah - also a general desire to pass as narrow interface as possible (I didn't create the interface for this because it's a single function).

pkg/gameservers/gameservers_test.go
zmerlynn: Thanks you!

cmd/controller/main.go
markmandel: That would be my thought, but also what happens if I enter "blueberry" 😄  does it default to auto detect?

Maybe if the value passed in is not known, then it throws a warning, and does an autodetect?

zmerlynn: I like this approach, will get a version up soon.

markmandel: Awesome. Otherwise, LGTM!

zmerlynn:
> Overall structure looks good! Just had a few small questions / comments. Nice work!

Thanks! Please take a look at the new version. :)

cmd/controller/main.go
zmerlynn: Oh, I somehow missed this comment

> Maybe if the value passed in is not known, then it throws a warning, and does an autodetect?

The version I uploaded throws an error if it doesn't know the product (and on "") - I think it's reasonable to do this on CLI flags, since overriding them is "explicit intent".

pkg/cloudproduct/gke/gke.go
```diff
@@ -0,0 +1,84 @@
...
+func Detect(ctx context.Context, kc *kubernetes.Clientset) string {
+	if !metadata.OnGCE() {
+		return ""
+	}
+	// Look for the workload defaulter - this is the current best method to detect Autopilot
+	if _, err := kc.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(
+		ctx, workloadDefaulterWebhook, metav1.GetOptions{}); err != nil {
+		logger.Infof("Assuming GKE Standard and defaulting to generic provider: "+
+			"failed to get MutatingWebhookConfigurations/%q (error expected if not on GKE Autopilot): %v",
+			workloadDefaulterWebhook, err)
```
markmandel, line 54: I think I somehow didn't submit this suggestion, sorry about that:

```suggestion
		logger.WithError(err)
		.WithField("reason", fmt.Sprintf("failed to get MutatingWebhookConfigurations/%q (error expected if not on GKE Autopilot), workloadDefaulterWebhook))
		.Info("Assuming GKE Standard and defaulting to generic provider")
```

Since we have structured logs.

(Or something like this).

zmerlynn: Ah, I've never worked with these funny structured logs! Done!

markmandel (approved):

PR #2789, opened 2022-11-02
Split port allocators, implement Autopilot port allocation/policies

zmerlynn:
**What type of PR is this?**

/kind feature

Continuing #2777: In the Agones on GKE Autopilot implementation, we have no need for the port allocator - the informer/etc. is an unnecessary moving piece. This PR allows for cloud products to provide their own port allocation
implementation. We do this by:

 * Splitting portallocator off to its own package. It was basically self-sufficient anyways, except it was a little too friendly with controller_test.go. I solved that by introducing a `TestInterface` for controller_test.go to upcast to.
 * Allow cloud product implementations to define their own port allocator.
 * Defining a new port allocator for GKE that does a simple per-port HostPort allocation, and adds the host-port-assignment annotation to the pod template.
 * Extend cloudproduct again to add a GameServer validator
 * And in Autopilot, reject if the PortPolicy is not `Dynamic`

### DIFF
--- a/pkg/cloudproduct/cloudproduct.go
+++ b/pkg/cloudproduct/cloudproduct.go
@@ -19,10 +19,14 @@ import (
 	"fmt"
 
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
+	"agones.dev/agones/pkg/client/informers/externalversions"
 	"agones.dev/agones/pkg/cloudproduct/generic"
 	"agones.dev/agones/pkg/cloudproduct/gke"
+	"agones.dev/agones/pkg/portallocator"
 	"agones.dev/agones/pkg/util/runtime"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -33,6 +37,12 @@ type CloudProduct interface {
 	// SyncPodPortsToGameServer runs after a Pod has been assigned to a Node and before we sync
 	// Pod host ports to the GameServer status.
 	SyncPodPortsToGameServer(*agonesv1.GameServer, *corev1.Pod) error
+
+	// ValidateGameServer is called by GameServer.Validate to allow for product specific validation.
+	ValidateGameServer(*agonesv1.GameServer) []metav1.StatusCause
+
+	// NewPortAllocator creates a PortAllocator. See gameservers.NewPortAllocator for parameters.
+	NewPortAllocator(int32, int32, informers.SharedInformerFactory, externalversions.SharedInformerFactory) portallocator.Interface
 }
 
 const (
--- a/pkg/cloudproduct/gke/gke.go
+++ b/pkg/cloudproduct/gke/gke.go
@@ -16,13 +16,17 @@ package gke
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
+	"agones.dev/agones/pkg/client/informers/externalversions"
+	"agones.dev/agones/pkg/portallocator"
 	"agones.dev/agones/pkg/util/runtime"
 	"cloud.google.com/go/compute/metadata"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -30,6 +34,8 @@ const (
 	workloadDefaulterWebhook     = "workload-defaulter.config.common-webhooks.networking.gke.io"
 	noWorkloadDefaulter          = "failed to get MutatingWebhookConfigurations/workload-defaulter.config.common-webhooks.networking.gke.io (error expected if not on GKE Autopilot)"
 	hostPortAssignmentAnnotation = "autopilot.gke.io/host-port-assignment"
+
+	errPortPolicyMustBeDynamic = "PortPolicy must be Dynamic on GKE Autopilot"
 )
 
 var logger = runtime.NewLoggerWithSource("gke")
@@ -83,4 +89,75 @@ func (*gkeAutopilot) SyncPodPortsToGameServer(gs *agonesv1.GameServer, pod *core
 		}
 	}
 	return nil
+}
+
+func (*gkeAutopilot) NewPortAllocator(minPort, maxPort int32,
+	_ informers.SharedInformerFactory,
+	_ externalversions.SharedInformerFactory) portallocator.Interface {
+	return &autopilotPortAllocator{minPort: minPort, maxPort: maxPort}
+}
+
+func (*gkeAutopilot) ValidateGameServer(gs *agonesv1.GameServer) []metav1.StatusCause {
+	var causes []metav1.StatusCause
+	for _, p := range gs.Spec.Ports {
+		if p.PortPolicy != agonesv1.Dynamic {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Field:   fmt.Sprintf("%s.portPolicy", p.Name),
+				Message: errPortPolicyMustBeDynamic,
+			})
+		}
+	}
+	return causes
+}
+
+type autopilotPortAllocator struct {
+	minPort int32
+	maxPort int32
+}
+
+func (*autopilotPortAllocator) Run(_ context.Context) error        { return nil }
+func (*autopilotPortAllocator) DeAllocate(gs *agonesv1.GameServer) {}
+
+func (apa *autopilotPortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.GameServer {
+	if len(gs.Spec.Ports) == 0 {
+		return gs // Nothing to do.
+	}
+
+	var ports []agonesv1.GameServerPort
+	for i, p := range gs.Spec.Ports {
+		if p.PortPolicy != agonesv1.Dynamic {
+			logger.WithField("gs", gs.Name).WithField("portPolicy", p.PortPolicy).Error(
+				"GameServer has invalid PortPolicy for Autopilot - this should have been rejected by webhooks. Refusing to assign ports.")
+			return gs
+		}
+		p.HostPort = int32(i + 1) // Autopilot expects _some_ host port - use a value unique to this GameServer Port.
+
+		if p.Protocol == agonesv1.ProtocolTCPUDP {
+			tcp := p
+			tcp.Name = p.Name + "-tcp"
+			tcp.Protocol = corev1.ProtocolTCP
+			ports = append(ports, tcp)
+
+			p.Name += "-udp"
+			p.Protocol = corev1.ProtocolUDP
+		}
+		ports = append(ports, p)
+	}
+
+	hpa := hostPortAssignment{Min: apa.minPort, Max: apa.maxPort}
+	hpaJSON, err := json.Marshal(hpa)
+	if err != nil {
+		logger.WithError(err).WithField("hostPort", hpa).WithField("gs", gs.Name).Error("Internal error marshalling hostPortAssignment for GameServer")
+		// In error cases, return the original gs - on Autopilot this will result in a policy failure.
+		return gs
+	}
+
+	// No errors past here.
+	gs.Spec.Ports = ports
+	if gs.Spec.Template.ObjectMeta.Annotations == nil {
+		gs.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+	gs.Spec.Template.ObjectMeta.Annotations[hostPortAssignmentAnnotation] = string(hpaJSON)
+	return gs
 }
--- a/pkg/cloudproduct/gke/gke_test.go
+++ b/pkg/cloudproduct/gke/gke_test.go
@@ -17,8 +17,8 @@ import (
 	"testing"
 
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
-	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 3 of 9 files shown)


zmerlynn:
(Rebased and removed comment about #2782.)

pkg/cloudproduct/gke/gke.go
```diff
@@ -82,3 +88,74 @@ func (*gkeAutopilot) SyncPodPortsToGameServer(gs *agonesv1.GameServer, pod *core
...
+	}
+
+	var ports []agonesv1.GameServerPort
+	for i, p := range gs.Spec.Ports {
+		if p.PortPolicy != agonesv1.Dynamic {
+			logger.Errorf("GameServer %s port %v has PortPolicy %q - this should have been rejected by webhooks, refusing to assign ports",
+				gs.Name, p, p.PortPolicy)
+			return gs
+		}
+		p.HostPort = int32(i + 1) // Autopilot expects _some_ host port - use a value unique to this GameServer Port.
```
markmandel, line 132: A thought I just had - not for this PR, but eventually we'll need a way to turn off some e2e tests for Autopilot (and maybe even add some Autopilot specific ones?)

Just something to think about.

pkg/cloudproduct/gke/gke_test.go
```diff
@@ -77,6 +77,198 @@ func TestSyncPodPortsToGameServer(t *testing.T) {
...
+					Field:   "another-bad-udp.portPolicy",
+				},
+			},
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			causes := (&gkeAutopilot{}).ValidateGameServer(tc.gs)
+			if diff := cmp.Diff(tc.want, causes); diff != "" {
+				t.Errorf("Causes differ (-want +got):\n%s", diff)
+			}
```
markmandel, line 157:
```suggestion
require.Equal(tc.want, causes)
```

Saves you some time and effort, and provides a diff if different. We use the testify libs pretty much everywhere.

pkg/cloudproduct/gke/gke_test.go
```diff
@@ -77,6 +77,198 @@ func TestSyncPodPortsToGameServer(t *testing.T) {
...
+							Protocol:      corev1.ProtocolUDP,
+						},
+					},
+				},
+			},
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			gs := (&autopilotPortAllocator{minPort: 8000, maxPort: 9000}).Allocate(tc.gs)
+			if diff := cmp.Diff(tc.wantGS, gs); diff != "" {
```
markmandel, line 265: Same as above, let's use testify.

pkg/portallocator/portallocator_test.go
```diff
@@ -626,10 +637,18 @@ func countTotalAllocatedPorts(pa *PortAllocator) int {
...
+func countTotalPorts(pa *portAllocator) int {
 	count := 0
 	for _, node := range pa.portAllocations {
 		count += len(node)
 	}
 	return count
 }
+
+func portAllocatorImpl(t *testing.T, pai Interface) *portAllocator {
+	pa, ok := pai.(*portAllocator)
```
markmandel, line 649:
```suggestion
	pa := pai.(*portAllocator)
```

Could you just do that? If the assertion fails, then a panic happens, and you get what I think is a pretty decent error message anyway? Minor nit though.

pkg/portallocator/portallocator_test.go
```diff
@@ -626,10 +637,18 @@ func countTotalAllocatedPorts(pa *PortAllocator) int {
...
+	if !ok {
+		t.Fatalf("toPortAllocator() called on portAllocator of type %t", pai)
```
markmandel, line 651: Non blocking nit, only change if you want: We do tend to use testify,so `require.Failf(...)` would be inline with that (not sure if I missed these in the previous PR).

pkg/cloudproduct/cloudproduct.go
```diff
@@ -33,6 +37,12 @@ type CloudProduct interface {
 	// SyncPodPortsToGameServer runs after a Pod has been assigned to a Node and before we sync
 	// Pod host ports to the GameServer status.
 	SyncPodPortsToGameServer(*agonesv1.GameServer, *corev1.Pod) error
+
+	// ValidateGameServer is called by GameServer.Validate to allow for product specific validation.
+	ValidateGameServer(*agonesv1.GameServer) []metav1.StatusCause
+
+	// PortAllocatorFactory creates a PortAllocator. c.f. gameservers.NewPortAllocator for parameters.
+	PortAllocatorFactory(int32, int32, informers.SharedInformerFactory, externalversions.SharedInformerFactory) portallocator.Interface
```
markmandel, line 45:
```suggestion
	NewPortAllocator(int32, int32, informers.SharedInformerFactory, externalversions.SharedInformerFactory) portallocator.Interface
```

Reason being, as `PortAllocatorFactory` makes me think I'm getting a Factory, not an actual allocator. Also `New` since it is essentially a constructor.

markmandel (commented):
Overall pattern looks good to me 👍🏻 . @roberthbailey would love your input too.

Everything else is minor stuff, naming changes, etc.

pkg/cloudproduct/gke/gke.go
zmerlynn: Yeah, I filed an internal bug to get e2es working on AP - I'll file an issue here soon, too.

pkg/cloudproduct/gke/gke_test.go
zmerlynn: Done. Sorry, old habits die hard. :)

pkg/cloudproduct/gke/gke_test.go
zmerlynn: Done

pkg/portallocator/portallocator_test.go
zmerlynn: Sounds good to me!

pkg/portallocator/portallocator_test.go
zmerlynn: I changed a couple from my previous PR and will scan for others here before I re-upload.

pkg/cloudproduct/cloudproduct.go
zmerlynn: You're right. I was overly factory happy here. :)

pkg/portallocator/portallocator_test.go
zmerlynn: Honestly that cleans up a lot. I realized along the way that I didn't need to define the `TestInterface` in the `cloudproduct` package either - I can define it in the test that's upcasting it.

zmerlynn:
> Overall pattern looks good to me 👍🏻 . @roberthbailey would love your input too.
> 
> Everything else is minor stuff, naming changes, etc.

Resolved, rebased, PTAL. :)

pkg/portallocator/doc.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
roberthbailey, line 1: nit: This looks like a new file, so we should use the current year for copyright instead of using the year the other files were created.

pkg/portallocator/portallocator_test.go
```diff
@@ -48,7 +48,7 @@ func TestPortAllocatorAllocate(t *testing.T) {
 
 	t.Run("test allocated port counts", func(t *testing.T) {
 		m := agtesting.NewMocks()
-		pa := NewPortAllocator(10, 50, m.KubeInformerFactory, m.AgonesInformerFactory)
+		pa := New(10, 50, m.KubeInformerFactory, m.AgonesInformerFactory).(*portAllocator)
```
roberthbailey, line 51: nit: It seems like we should be able to avoid casting here. What about a `new` (non-exported) method that returns `*portAllocator` that is called by `New` (which then just wraps the return value into the exported `portAllocator.Interface`)?

pkg/portallocator/portallocator.go
```diff
@@ -56,18 +51,34 @@ type PortAllocator struct {
 	nodeInformer       cache.SharedIndexInformer
 }
 
-// NewPortAllocator returns a new dynamic port
+// portallocator.Interface manages the dynamic port allocation strategy.
+//
+// The portallocator does not currently support mixing static portAllocations (or any pods with defined HostPort)
+// within the dynamic port range other than the ones it coordinates.
+type Interface interface {
```
roberthbailey, line 58: nit: Since this is the exported type for this package, can you move it above the non-exported concrete type?

pkg/portallocator/portallocator.go
```diff
@@ -236,20 +247,22 @@ func (pa *PortAllocator) DeAllocate(gs *agonesv1.GameServer) {
...
-// syncAll syncs the pod, node and gameserver caches then
+// SyncAll syncs the pod, node and gameserver caches then
 // traverses all Nodes in the cluster and all looks at GameServers
 // and Terminating Pods values make sure those
 // portAllocations are marked as taken.
 // Locks the mutex while doing this.
 // This is basically a stop the world Garbage Collection on port allocations, but it only happens on startup.
-func (pa *PortAllocator) syncAll() error {
+//
+// SyncAll is exposed on the TestInterface only.
```
roberthbailey, line 264: @markmandel - I'm not an expert on how we use informers in our test code, but I'm wondering if there is a way to do it without needing to expose these functions outside of this package. Do you have any thoughts?

pkg/cloudproduct/gke/gke_test.go
```diff
@@ -66,17 +66,201 @@ func TestSyncPodPortsToGameServer(t *testing.T) {
...
+				require.Equal(t, tc.wantGS, tc.gs)
+				require.Equal(t, oldPod, tc.pod)
 			}
 		})
 	}
 }
 
+func TestValidateGameServer(t *testing.T) {
+	for name, tc := range map[string]struct {
+		gs   *agonesv1.GameServer
```
roberthbailey, line 78: Each test case would be more concise if you just used the slice of `GameServerPorts` instead of the whole struct (the stuff around that part is just boilerplate for these tests). 

Also, are these tests sufficient for all edge cases? (I didn't check closely enough to know, so it's ok to say yes).

pkg/cloudproduct/gke/gke_test.go
```diff
@@ -66,17 +66,201 @@ func TestSyncPodPortsToGameServer(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			causes := (&gkeAutopilot{}).ValidateGameServer(tc.gs)
+			require.Equal(t, tc.want, causes)
+		})
+	}
+}
+
+func TestAutopilotPortAllocator(t *testing.T) {
+	for name, tc := range map[string]struct {
+		gs     *agonesv1.GameServer
```
roberthbailey, line 158: As on the previous one, it might be simpler to pass in a slice ports and specify the slice of ports you expect to get back. In this case, there would also need to be a bool to verify whether the annotation is set on the resulting game server. 

Same question as on the last one - is 2 test cases enough to catch all edge cases?

pkg/portallocator/doc.go
zmerlynn: Done

pkg/portallocator/portallocator_test.go
zmerlynn: Sounds good, done.

pkg/portallocator/portallocator.go
zmerlynn: Done

pkg/cloudproduct/gke/gke_test.go
zmerlynn: Agreed on ports. I think the test cases are more than adequate here, and I double checked the coverage layer and it seems to agree - in terms of validation there's not a whole lot. The harder one is actually below, which I'll look at longer.

pkg/portallocator/portallocator.go
zmerlynn: FWIW, I did not spend long trying to figure that out, since a change in structure of allocating the informers earlier was considerably more invasive. I assume there may be a way to pull the informers out and pass them down, but maybe we can handle that as a follow-on? (Though it frankly seems somewhat low priority - I tried my best to ensure that it would take pretty explicit intent to call these outside the package.)

pkg/cloudproduct/gke/gke_test.go
zmerlynn: I looked longer at this one as it's more complex, and added one other minor test that adds a TCPUDP after the normal ports. One thing is that the "2 test cases" are really compound .. I layered several pieces of complexity into the success case. Again, checking coverage, the coverage here is solid, with the exception of an error from `json.Marshal`. I have historically never bothered testing such errors, as they can be more work to induce than is worth it.

pkg/cloudproduct/cloudproduct.go
```diff
@@ -33,6 +37,12 @@ type CloudProduct interface {
 	// SyncPodPortsToGameServer runs after a Pod has been assigned to a Node and before we sync
 	// Pod host ports to the GameServer status.
 	SyncPodPortsToGameServer(*agonesv1.GameServer, *corev1.Pod) error
+
+	// ValidateGameServer is called by GameServer.Validate to allow for product specific validation.
+	ValidateGameServer(*agonesv1.GameServer) []metav1.StatusCause
+
+	// NewPortAllocator creates a PortAllocator. c.f. gameservers.NewPortAllocator for parameters.
```
markmandel, line 44: What's "c.f." ? See function?

pkg/portallocator/portallocator.go
markmandel: In this particular instance, you could call `portallocator.Run(ctx)` in the test, since it spawns no worker threads, and isn't doing anything other than syncing the state. Even if it's just `context.Background()`.

That way you don't have to export the `syncAll()` function, `Run` is already public, and you don't need to export the `NodeSynced()` function either.

I think that should work!

zmerlynn: Ah, I missed that. Totally works, thanks!

pkg/cloudproduct/cloudproduct.go
zmerlynn: "see also", except I screwed it up: https://en.wikipedia.org/wiki/Cf.

Will just change to "see".

roberthbailey (commented):
lgtm. Waiting for @markmandel to approve before merging.

zmerlynn:
(Rebased)

pkg/cloudproduct/gke/gke.go
```diff
@@ -84,3 +90,74 @@ func (*gkeAutopilot) SyncPodPortsToGameServer(gs *agonesv1.GameServer, pod *core
...
+			p.Name += "-udp"
+			p.Protocol = corev1.ProtocolUDP
+		}
+		ports = append(ports, p)
+	}
+
+	hpa := hostPortAssignment{Min: apa.minPort, Max: apa.maxPort}
+	hpaJSON, err := json.Marshal(hpa)
+	if err != nil {
+		logger.Errorf("Internal error marshalling hostPortAssignment %v for GameServer %s: %v", hpa, gs.Name, err)
```
markmandel, line 151: Sorry, missed this logging statement on first pass.

Nit, use structured logging:

```suggestion
		logger.WithError(err).WithField("hostPort", hpa).WithField("gs", gs.Name).Error("Internal error marshalling hostPortAssignment for GameServer")
```

pkg/cloudproduct/gke/gke.go
```diff
@@ -84,3 +90,74 @@ func (*gkeAutopilot) SyncPodPortsToGameServer(gs *agonesv1.GameServer, pod *core
...
+
+func (apa *autopilotPortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.GameServer {
+	if len(gs.Spec.Ports) == 0 {
+		return gs // Nothing to do.
+	}
+
+	var ports []agonesv1.GameServerPort
+	for i, p := range gs.Spec.Ports {
+		if p.PortPolicy != agonesv1.Dynamic {
+			logger.Errorf("GameServer %s port %v has PortPolicy %q - this should have been rejected by webhooks, refusing to assign ports",
```
markmandel, line 130: Sorry, another one to switch to structured logging 😄 

I swear I'll stop finding these at the last minute.

zmerlynn:
PTAL :)

pkg/cloudproduct/gke/gke.go
```diff
@@ -84,3 +90,74 @@ func (*gkeAutopilot) SyncPodPortsToGameServer(gs *agonesv1.GameServer, pod *core
...
+
+func (apa *autopilotPortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.GameServer {
+	if len(gs.Spec.Ports) == 0 {
+		return gs // Nothing to do.
+	}
+
+	var ports []agonesv1.GameServerPort
+	for i, p := range gs.Spec.Ports {
+		if p.PortPolicy != agonesv1.Dynamic {
+			logger.WithField("gs", gs.Name).WithField("PortPolicy", p.PortPolicy).Error(
```
markmandel, line 130: Super tiny nit:

```suggestion
			logger.WithField("gs", gs.Name).WithField("portPolicy", p.PortPolicy).Error(
```

(Which made me go back and look at what we do everywhere else, and all fields we have start with a lower case)

zmerlynn: Heh, I had it as `portPolicy` initially and changed it back, oops. :) Fixed.

markmandel (approved):

PR #2790, opened 2022-11-02
Fix (not really) problems reported by VSCode

zmerlynn:
**What type of PR is this?**

/kind cleanup

VSCode reports `main redeclared` between allocationload.go and runscenario.go due to the fact that they both look like `package main` binaries in the same directory, similar to e.g. [this poster on a different project](https://stackoverflow.com/questions/66970531/vs-code-go-main-redeclared-in-this-block).

To fix it, it's easy enough to just give these binaries their own package path and fix up the calling scripts.


markmandel:
Oh no, conflicts!

zmerlynn:
> Oh no, conflicts!

Fixed!

markmandel (approved):

PR #2792, opened 2022-11-04
Add fswatch library to watch/batch filesystem events, use in allocator

### LINKED ISSUE #1816: Only refresh certificates if the fsnotify event is relevant
Right now the certificates for the certs directory and client-ca are refreshed on any event fired from the fsnotify watcher: https://github.com/googleforgames/agones/blob/master/cmd/allocator/main.go#L147-L156

The watcher fires multiple events when a new k8s secret is installed into the directory. From the logs I see:

Create
Chmod
Rename
Remove
Remove
Remove

all being logged within 100ms for a single update to the cert. 

The current code takes a mutex, re-reads the file system, re-parses the certs, and drops the mutex for each event. It would make it easier to tell what is happening in the system if we only did the work to refresh certificates when necessary. 

Leave it alone; the refresh is currently working.

zmerlynn:
**What type of PR is this?**

/kind cleanup

This pull refactors the fsnotify code in allocator/main out to a shared library, and in that shared library implements a batched notification processor.

Closes #1816: This takes a slightly different approach than specified in the issue, instead choosing to just delay processing until after a batch processing period. I chose 1s - it's far longer than necessary, but still much shorter than it takes for the secret changes to propagate to the container anyways.

I considered the approach in #1816 of trying to parse the actual events, but it's too fiddly to get exactly right: e.g. maybe you only refresh on "write", but then "chmod" could make the file readable whereas it wasn't before, "rename" could expose a file that wasn't there before, etc.

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -37,6 +37,7 @@ import (
 	"agones.dev/agones/pkg/client/informers/externalversions"
 	"agones.dev/agones/pkg/gameserverallocations"
 	"agones.dev/agones/pkg/gameservers"
+	"agones.dev/agones/pkg/util/fswatch"
 	"agones.dev/agones/pkg/util/runtime"
 	"agones.dev/agones/pkg/util/signals"
 	gw_runtime "github.com/grpc-ecosystem/grpc-gateway/runtime"
@@ -49,7 +50,6 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/status"
-	"gopkg.in/fsnotify.v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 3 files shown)


pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,80 @@
...
+	go batchWatch(batchFor, watcher.Events, watcher.Errors, cancelChan, processEvent, func(error) {
+		logger.WithError(err).Errorf("error watching %s", path)
+	})
+	return cancel, nil
+}
+
+// batchWatch: watch for events; when an event occurs, keep draining events for duration `batchFor`, then call processEvent().
+// Intended for batching of rapid-fire events where we want to process the batch once, like filesystem update notifications.
+func batchWatch(batchFor time.Duration, events chan fsnotify.Event, errors chan error, cancelChan chan struct{}, processEvent func(), onError func(error)) {
+	// Pattern shamelessly stolen from https://blog.gopheracademy.com/advent-2013/day-24-channel-buffering-patterns/
```
markmandel, line 53: ooooh! Fancy!

pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,80 @@
...
+package fswatch
+
+import (
+	"time"
+
+	"agones.dev/agones/pkg/util/runtime"
+	"gopkg.in/fsnotify.v1"
+)
+
+var logger = runtime.NewLoggerWithSource("fswatch")
```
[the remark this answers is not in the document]
zmerlynn: Yup, this is what prompted that internal convo. I'll fix it and get a new version up soon. :)

pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,80 @@
...
+		close(cancelChan)
+		_ = watcher.Close()
+	}
+	if err := watcher.Add(path); err != nil {
+		cancel()
+		return nil, err
+	}
+
+	go batchWatch(batchFor, watcher.Events, watcher.Errors, cancelChan, processEvent, func(error) {
+		logger.WithError(err).Errorf("error watching %s", path)
```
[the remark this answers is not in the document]
zmerlynn: Love it.

pkg/util/fswatch/fswatch.go
zmerlynn: I almost started to hand roll it, and then I had this tickle that I've seen this pattern before... :)

zmerlynn:
Rebased, resolved. PTAL, thanks!

cmd/allocator/main.go
```diff
@@ -224,70 +224,38 @@ func main() {
 	h := newServiceHandler(kubeClient, agonesClient, health, conf.MTLSDisabled, conf.TLSDisabled, conf.remoteAllocationTimeout, conf.totalRemoteAllocationTimeout, conf.allocationBatchWaitTime)
 
 	if !h.tlsDisabled {
-		watcherTLS, err := fsnotify.NewWatcher()
+		cancelTLS, err := fswatch.Watch(logger, tlsDir, time.Second, func() {
+			tlsCert, err := readTLSCert()
+			if err != nil {
+				logger.WithError(err).Error("could not load TLS certs; keeping old one")
+			} else {
```
roberthbailey, line 231: nit: Now that this is nicely wrapped in a function, we can use the more idiomatic control flow of

```
if err != nil {
  logger.WithError(...)
  return
}
h.tlsMutex.Lock()
h.tlsCert = tlsCert
h.tlsMutex.Unlock()
logger.Info(...)
```

This also means that if there is an error we won't get an error log line followed by the info line.

roberthbailey: Another thing that occurred to me (which doesn't make as much difference here) is that instead of defering the Unlock call we did it inline before because this was inside of a large code block and not a small function. This could also be changed to

```
h.tlsMutex.Lock()
defer h.tlsMutex.Unlock()
h.tlsCert = tlsCert
```

if you think that is clearer (I'm on the fence, since it's just a single line that is in the lock's scope).

cmd/allocator/main.go
```diff
@@ -224,70 +224,38 @@ func main() {
...
-				}
-			}
-		}()
+		defer cancelTLS()
 
 		if !h.mTLSDisabled {
-			// creates a new file watcher for client certificate folder
-			watcher, err := fsnotify.NewWatcher()
+			cancelCert, err := fswatch.Watch(logger, certDir, time.Second, func() {
+				h.certMutex.Lock()
```
roberthbailey, line 245: Now that this in inside a nice tidy function, we can use the convention of defering the unlock call right after the lock call:

```
h.certMutex.Lock()
defer h.certMutex.UnLock()
```

which allows the code below to use typical return conventions:

```
... := getCACertPool(...)
if err != nil {
 ...
 return
}
h.caCertPol = caCertPool
log(...)
```

pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,81 @@
...
+package fswatch
+
+import (
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify.v1"
+)
+
+// Watch watched the filesystem path `path`. When anything changes, changes are
```
roberthbailey, line 25: nit: watched -> watches

zmerlynn: Done

zmerlynn:
Appreciate the nod to cleanliness. PTAL!

roberthbailey (approved):

zmerlynn:
Rebased, hopefully it doesn't flake again.

markmandel (approved):

PR #2795, opened 2022-11-08
Enable fieldalignment linter, then mostly ignore it

### LINKED ISSUE #2325: Conditionally enable fieldalignment linter in govet
In https://github.com/googleforgames/agones/pull/2323 we disabled the maligned linter as it has been deprecated. The replacement is govet 'fieldalignment' but turning that on for the whole codebase produces a number of errors. 

As I mentioned in in https://github.com/googleforgames/agones/pull/2323#pullrequestreview-784797859 it would be nice to enable the checks on parts of the code base (for structures where we care about memory compaction) but leave some structs (like config structs that are only created once) organized in a more human readable fashion rather than optimizing them for memory usage. 

Leave the check disabled.

zmerlynn:
This enables the fieldalignment linter by enabling all `govet` checks except shadowing. I then configured it to ignore large swaths of code (tests, cmd/, APIs), and nolint'd all the existing complaints because they seemed irrelevant.

Along the way:

* removed existing nolint:maligned, as `maligned` is no more.

* disabled `structcheck` and `deadcode` as they are deprecated (and I
believe have been subsumed by other linters)

* changed `gameServerCacheEntry` to `gameServerCache`. It is the
cache, not just an entry.

* fixed alignment of `gameServerSetCacheEntry`.

/kind cleanup
Closes #2325

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,7 +48,6 @@ output:
 linters:
   enable:
     - bodyclose
-    - deadcode
     - dupl
     - exportloopref
     - goconst
@@ -61,11 +60,16 @@ linters:
     - nakedret
     - revive
     - staticcheck
-    - structcheck
     - unconvert
     - unparam
 
 linters-settings:
+  govet:
+    disable-all: false
+    enable-all: true
+    disable:
+      # extremely noisy, also against common Go style in most cases
+      - shadow
   gocritic:
     enabled-tags:
       - performance
@@ -90,3 +94,11 @@ issues:
   exclude-use-default: false
   exclude:
     - Using the variable on range scope .* in function literal
+  exclude-rules:
+    # Skip fieldalignment checks on:
+    # - tests: memory footprint doesn't matter
+    # - cmd/: we assume these are one-off config entries
+    # - pkg/apis: Keep strong convention on TypeMeta/ObjectMeta/Spec/Status
+    - path: '(.+)_test\.go|^test/|^cmd/.*|^pkg/apis/.*'
+      # fieldalignment is in the `govet` linter, so exclude based on text instead of all of govet
+      text: '^fieldalignment: .*'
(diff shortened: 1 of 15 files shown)


gongmax (approved):

.golangci.yml
```diff
@@ -48,7 +48,6 @@ output:
 linters:
   enable:
     - bodyclose
-    - deadcode
```
roberthbailey, line 51: deadcode doesn't sound related to field alignment, which makes me curious as to why this line was removed.

pkg/gameserverallocations/cache.go
```diff
@@ -21,6 +21,8 @@ import (
 )
 
 // gameserver cache to keep the Ready state gameserver.
+//
+//nolint:govet // ignore fieldalignment, low count
```
roberthbailey, line 25: I'm curious about some of these comments (especially the low count ones in struct names like this that _sound_ like we would have lots of them). 

/cc @markmandel

.golangci.yml
gongmax: `deadcode` is replaced by `unused` which is enabled by default. https://golangci-lint.run/usage/linters/

zmerlynn: ^ It's in the "along the way" of the commit but I didn't copy that language to the PR after I update it.

pkg/gameserverallocations/cache.go
zmerlynn: This one in particular is embedded in AllocationCache. I changed the comment and the name of the structure to reflect that, since it was also confusing me.

The gameServerStateCache was different and I was wrong there after looking at it more - I thought it followed the pattern here. I went ahead and just fixed the alignment.

.golangci.yml
roberthbailey: 👍

roberthbailey (approved):

Issue #2279, opened 2021-09-28
The Documentation/Installation page on old releases shows an incorrect Kubernetes version


roberthbailey:
**What happened**:

During a recent community meeting someone asked the best way to tell which version of Kubernetes is supported by each Agones release. @markmandel suggested looking at the older releases of the website, specifically the https://agones.dev/site/docs/installation/ page, to see what version they suggest.

However, I just checked a few old versions and noticed that they are showing the wrong version of Kubernetes!

From the blog releases, we can see that:
* [1.13.0](https://agones.dev/site/blog/2021/03/16/1.13.0-kubernetes-1.18-websocket-watch-sdk-empty-ports/) added support for Kubernetes 1.18
* [1.16.0](https://agones.dev/site/blog/2021/07/20/1.16.0-kubernetes-1.19-golang-1.15/) added support for Kubernetes 1.19
* [1.17.0](https://agones.dev/site/blog/2021/08/31/1.17.0-kubernetes-1.20-deprecated-required/preferred-add-sync-property-to-fleetautoscaler/) added support for Kubernetes 1.20

However, if you look at the Install documentation for these releases, we see a different story:
* [1.12.0](https://1-12-0.agones.dev/site/docs/installation/) says to use Kubernetes 1.18 (should be 1.17)
* [1.13.0](https://1-13-0.agones.dev/site/docs/installation/) says to use Kubernetes 1.18 (correct)
* [1.14.0](https://1-14-0.agones.dev/site/docs/installation/) says to use Kubernetes 1.18 (correct)
* [1.15.0](https://1-15-0.agones.dev/site/docs/installation/) says to use Kubernetes 1.19 (should be 1.18)
* [1.16.0](https://1-16-0.agones.dev/site/docs/installation/) says to use Kubernetes 1.20 (should be 1.19)

**What you expected to happen**:

I expect that the website for an old release shows the correct Kubernetes version to install for that release.

markmandel:
Huh! That's super wierd!

🤔 I wonder if it's publishing the dev version of the site, which is showing the _next_ supported version.

uuugh - yeah it is: https://github.com/googleforgames/agones/blob/main/site/layouts/shortcodes/k8s-version.html

So we don't publish with env HUGO_ENV=production so we don't have duplicate content on search engines (dev content is set to not be indexable). (This is a docsy setting, which I think [changed recently](https://www.docsy.dev/docs/deployment/#build-environments-and-indexing))

A few thoughts:
[...]

PR #2808, opened 2022-11-11
Release Automation: Generate version of website to push

### LINKED ISSUE #2457: Release Automation: Generate version of website to push
Right now is a step in the [release process](https://github.com/googleforgames/agones/blob/main/docs/governance/templates/release_issue.md) to capture an old version of the website:

- [ ] If full release, run `make site-deploy SERVICE={version}-1`, (replace . with -)
   - For example, if you are creating the 1.18.0 release, then you would deploy the 1-17-0 service (release minus one, and then replace dots with dashes).

We have a make target to push a new version of the website during a release to capture a snapshot before we make changes. But we don't have one that will create the proper version string for the release; right now that is still done by a human. 

We should create a make target that can generate and format the proper version string.

mangalpalli:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

kind feature
> /kind hotfix

We need to have a make target to push a new version of the website during a release to capture a snapshot before we make changes. But we didn't have one that created the proper version string for the release and required to change the version string manually.

We created a make target that can generate and format the proper version string for the release.

Closes #2457


build/scripts/previousversion/main.go
```diff
@@ -0,0 +1,16 @@
...
+	semver "github.com/blang/semver/v4"
+)
+
+func main() {
+	baseVersion := flag.String("version", "", "the base version to be taken in as input for release")
+	flag.Parse()
+	v, _ := semver.Parse(*baseVersion)
+	releaseVersion := fmt.Sprintf("%d.%d.%d", v.Major, v.Minor-1, v.Patch)
+	fmt.Println(releaseVersion)
+}
```
markmandel, line 16: 🤔 do we put a small go.mod in this directory, rather than add at the root level (then we also don't need to vendor the code either)

Since this gets run once every 6 weeks, it probably doesn't need to be vendor ed. WDYT?

build/scripts/previousversion/main.go
```diff
@@ -0,0 +1,16 @@
...
+	"fmt"
+
+	semver "github.com/blang/semver/v4"
+)
+
+func main() {
+	baseVersion := flag.String("version", "", "the base version to be taken in as input for release")
+	flag.Parse()
+	v, _ := semver.Parse(*baseVersion)
+	releaseVersion := fmt.Sprintf("%d.%d.%d", v.Major, v.Minor-1, v.Patch)
```
markmandel, line 14: ooooh! So we missed a small part of the requirements in the ticket:

> (replace . with -)

But we can do this here! Use `-` rather than `.`, and we are good to go!

markmandel:
With the notes above - you should be able to deploy a 1-27-0 service on App Engine - which would be neat!

build/scripts/previousversion/main.go
mangalpalli: Hi, I believe that creating multiple go.mod might require additional efforts on maintaining the documentation and explaining the reason behind adding it seperately along with the major vendor file. Also, creating multiple such directories and executables might lead to us maintaining multiple go.mod dependencies, which i think might be confusing. Just my thought.

build/scripts/previousversion/main.go
mangalpalli: Updated it.

build/scripts/previousversion/main.go
markmandel: I'd probably argue that it's easier to manage and maintain to keep it seperate. This go.mod is for development tooling - whereas the root go.mod is for building production Agones.

🤔 maybe a good in-between would be to have a go.mod in `build/scripts` - then we can have a central place for any build-tooling Go dependencies, but they can be kept separate from the dependencies and Go version for the actual Agones code. How does that sound?

mangalpalli: Fair point, will try this.

markmandel:
This looks good! Only thing extra to add, would be to update https://github.com/googleforgames/agones/blob/main/docs/governance/templates/release_issue.md to use this process instead!

mangalpalli:
> This looks good! Only thing extra to add, would be to update https://github.com/googleforgames/agones/blob/main/docs/governance/templates/release_issue.md to use this process instead!

I will update it.

build/scripts/previousversion/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/agones/agones/build/scripts/previousversion
+
+go 1.19
```
markmandel, line 3: Looks good! I think this is cleaner, what do you think?

build/scripts/previousversion/main.go
```diff
@@ -0,0 +1,16 @@
+package main
```
markmandel, line 1: Ooh, this file needs an Apache header please.

markmandel (commented):
Looks good! just missing a licence, and it's good to go 👍🏻

build/scripts/previousversion/main.go
mangalpalli: I didn't get this, what do you mean by apache header?

markmandel: We need the open source Apache licence header at the top of all code files.

https://www.apache.org/licenses/LICENSE-2.0#apply

For example, see https://github.com/googleforgames/agones/blob/main/examples/xonotic/main.go

markmandel:
I'm not sure how I managed to do it, but it looks like I accidentally pushed to your branch! 😬  sorry about that!

(github tooling is being weird)

But I think I solved the issue you were having:

https://github.com/googleforgames/agones/pull/2808/files#diff-30a85f458fa13d688e195f0a550118ae55eabfa0bc4836a4379a1e71e3c8966aR59-R64

The magic pieces here are:

* Adding `&&` between each command - that way if one command fails, it will stop right away (makes things much easier to debug)
* The `GOFLAGS="-mod=mod"` - that is a Go setting, but `DOCKER_RUN_ARGS` is passed to the `docker run` command, so if we want to add a new environment variable, it needs to have a `-e` before it.

markmandel:
I think I accidentally broke your go.mod - feel free to revert my changes.

markmandel (approved):
This looks good!

If you feel confident, we can merge this now, so it's ready for you to use during the 1.28.0 stable release. (will need to move it out of draft).

Also, thanks for squashing all the commits! 😃

markmandel:
Okay - let's get stable out, and then we can merge it after that release is done 👍🏻 

@igooch you might also find this interesting, since you are also looking at #2279

markmandel (approved):
Approved. 

@mangalpalli can you move this out of draft, and we can merge it in 👍🏻

PR #2813, opened 2022-11-15
Terraform, GKE - Option to create regional cluster as well as option to create autoscaling nodepool

### LINKED ISSUE #1441: Terraform, GKE - Option to create a Regional Cluster
A best practice for production workload on GKE would be to run on Regional Clusters, and not Zonal Clusters. Regional clusters provide redundancy on the masters, as well as allowing nodepools to exist across zones as well.

Have an option to create the cluster as Regional within GKE on the Terraform config.
This should also turn off cluster auto balancing as well.

We should discuss if this should be the default or not. 

Default on is a breaking change, but drives people towards a production setup.
Default off is a nicer "try out" experience, but drives people towards a non-production setup. We could potentially solve this with documentation.

None.

**Additional context**

- https://cloud.google.com/kubernetes-engine/docs/concepts/regional-clusters
- https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#balancing_across_zones
- https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#im-running-cluster-with-nodes-in-multiple-zones-for-ha-purposes-is-that-supported-by-cluster-autoscaler

### LINKED ISSUE #1467: Terraform, GKE - add autoscaling Node Pools option

In order GKE Cluster to be more cost-efficient there should be an option to scale down GameServers Node Pool size.
Some of Agones users who used gcloud to provision their clusters has already autoscaling for Node Pools turned on. This functionality also is supported by Packed `GameServerAllocation` Scheduling.

[container_node_pool#autoscaling](https://www.terraform.io/docs/providers/google/r/container_node_pool.html#autoscaling) or
[container_cluster#cluster_autoscaling](https://www.terraform.io/docs/providers/google/r/container_cluster.html#cluster_autoscaling)

Currently we are using strict [node_count](https://www.terraform.io/docs/providers/google/r/container_node_pool.html#node_count) parameter, which should be omitted when we are using `autoscaling` parameter.
So we could add this configuration separately to what we have now.

If it is possible merge both options together and depending on variable use autoscaling or strict node pool size.

**Additional context**
Using the Cluster Autoscaler with Agones - #368
and Kubernetes Node Autoscaling #145

chiayi:
**What type of PR is this?**
/kind bug

It adds configuration to terraform so that regional clusters can be created.
It also added autoscaling option for `default` nodepools. 

Closes #1441 
Closes #1467 

I've also made changes to the README for /build to include the parameters.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -43,6 +43,7 @@ KUBECONFIG ?= ~/.kube/config
 # The (gcloud) test cluster that is being worked against
 GCP_CLUSTER_NAME ?= test-cluster
 GCP_CLUSTER_ZONE ?= us-west1-c
+GCP_CLUSTER_LOCATION ?= $(GCP_CLUSTER_ZONE)
 GCP_BUCKET_CHARTS ?= agones-chart
 # the profile to use when developing on minikube
 MINIKUBE_PROFILE ?= agones
--- a/build/includes/google-cloud.mk
+++ b/build/includes/google-cloud.mk
@@ -27,10 +27,17 @@ gcloud-init: ensure-build-config
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT ?= 4
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_MACHINETYPE ?= e2-standard-4
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_ENABLEIMAGESTREAMING ?= true
+gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_AUTOSCALE ?= false
+gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_MIN_NODECOUNT ?= 1
+gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_MAX_NODECOUNT ?= 5
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_WINDOWSINITIALNODECOUNT ?= 0
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_WINDOWSMACHINETYPE ?= e2-standard-4
 gcloud-test-cluster: $(ensure-build-image)
-	$(MAKE) gcloud-terraform-cluster GCP_TF_CLUSTER_NAME="$(GCP_CLUSTER_NAME)" GCP_CLUSTER_ZONE="$(GCP_CLUSTER_ZONE)" \
+	$(MAKE) gcloud-terraform-cluster GCP_TF_CLUSTER_NAME="$(GCP_CLUSTER_NAME)" \
+		GCP_CLUSTER_LOCATION="$(GCP_CLUSTER_LOCATION)" \
+		GCP_CLUSTER_AUTOSCALE="$(GCP_CLUSTER_NODEPOOL_AUTOSCALE)" \
+		GCP_CLUSTER_MIN_NODECOUNT="$(GCP_CLUSTER_NODEPOOL_MIN_NODECOUNT)" \
+		GCP_CLUSTER_MAX_NODECOUNT="$(GCP_CLUSTER_NODEPOOL_MAX_NODECOUNT)" \
 		GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT="$(GCP_CLUSTER_NODEPOOL_INITIALNODECOUNT)" \
 		GCP_CLUSTER_NODEPOOL_MACHINETYPE="$(GCP_CLUSTER_NODEPOOL_MACHINETYPE)" \
 		GCP_CLUSTER_NODEPOOL_ENABLEIMAGESTREAMING="$(GCP_CLUSTER_NODEPOOL_ENABLEIMAGESTREAMING)" \
@@ -73,9 +80,7 @@ clean-gcloud-prow-build-cluster: $(ensure-build-image)
 # Pulls down authentication information for kubectl against a cluster, name can be specified through GCP_CLUSTER_NAME
 # (defaults to 'test-cluster')
 gcloud-auth-cluster: $(ensure-build-image)
-	docker run --rm $(common_mounts) $(build_tag) gcloud config set container/cluster $(GCP_CLUSTER_NAME)
-	docker run --rm $(common_mounts) $(build_tag) gcloud config set compute/zone $(GCP_CLUSTER_ZONE)
-	docker run --rm $(common_mounts) $(build_tag) gcloud container clusters get-credentials $(GCP_CLUSTER_NAME)
+	docker run --rm $(common_mounts) $(build_tag) gcloud container clusters get-credentials $(GCP_CLUSTER_NAME) --zone  $(GCP_CLUSTER_LOCATION)
 
 # authenticate our docker configuration so that you can do a docker push directly
 # to the gcr.io repository
--- a/build/terraform/gke/module.tf
+++ b/build/terraform/gke/module.tf
@@ -63,9 +63,14 @@ variable "ping_service_type" {
   default = "LoadBalancer"
 }
 
-variable "zone" {
+variable "location" {
   default     = "us-west1-c"
-  description = "The GCP zone to create the cluster in"
+  description = "The GCP location to create the cluster in"
+}
+
+variable "zone" {
+  default     = ""
+  description = "The GCP zone to create the cluster in (deprecated, use `location`)"
 }
 
 variable "pull_policy" {
@@ -86,6 +91,18 @@ variable "image_pull_secret" {
 
 variable "log_level" {
   default = "info"
+}
+
+variable "autoscale" {
+  default = "false"
+}
+
+variable "min_node_count" {
+  default = "1"
+}
+
+variable "max_node_count" {
+  default = "5"
 }
 
 // Note: This is the number of gameserver nodes. The Agones module will automatically create an additional
@@ -116,6 +133,7 @@ module "gke_cluster" {
 
   cluster = {
     "name"                    = var.name
+    "location"		      = var.location
     "zone"                    = var.zone
     "machineType"             = var.machine_type
     "initialNodeCount"        = var.node_count
@@ -124,6 +142,9 @@ module "gke_cluster" {
     "windowsInitialNodeCount" = var.windows_node_count
     "project"                 = var.project
     "network"                 = var.network
+    "autoscale"		      = var.autoscale
+    "minNodeCount"	      = var.min_node_count
+    "maxNodeCount"	      = var.max_node_count
   }
 }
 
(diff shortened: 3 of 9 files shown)


build/Makefile
```diff
@@ -42,6 +42,7 @@ REGISTRY ?= $(release_registry)
 KUBECONFIG ?= ~/.kube/config
 # The (gcloud) test cluster that is being worked against
 GCP_CLUSTER_NAME ?= test-cluster
+GCP_CLUSTER_LOCATION ?=
 GCP_CLUSTER_ZONE ?= us-west1-c
```
zmerlynn, line 46: I think we could keep the Makefile/.mk changes a little less chatty by just collapsing to `GCP_CLUSTER_LOCATION` here:

```
GCP_CLUSTER_ZONE ?= us-west1-c
GCP_CLUSTER_LOCATION ?= $(GCP_CLUSTER_ZONE)
```

That would let you just use `LOCATION` in build/includes/google-cloud.mk and remove the conditionals, since `GCL_CLUSTER_LOCATION` would always be set.

build/terraform/gke/module.tf
```diff
@@ -63,6 +63,11 @@ variable "ping_service_type" {
   default = "LoadBalancer"
 }
 
+variable "location" {
+  default     = ""
+  description = "The GCP location to create the cluster in"
+}
+
```
zmerlynn, line 70: Should we just deprecate `zone`? IMO yes, but would like input from @roberthbailey. It would look like:
* on new line 67, change the default to `"us-west1-c"`
* on new line 72, change the default to `""`
* on new line 73, let's say `"The GCP zone to create the cluster in (deprecated, use 'location')"`

It looks like there is no formal deprecation procedure (https://github.com/hashicorp/terraform/issues/18381) though, so I think to actually enforce the deprecation we'd have to add output to the module if `zone` were set.

WDYT?

zmerlynn (changes requested):

build/Makefile
chiayi: This makes sense to me.

build/terraform/gke/module.tf
chiayi: I think the plan is to 
> make sure we keep things backwards compatible for at least 1 release and give folks a warning in the release notes that the variable name is changing.

So adding a deprecation note does follow that. But does this mean that `zone` would take precedence over `location` for the 1 release before being removed?

zmerlynn: `GCP_CLUSTER_ZONE` and `GCP_CLUSTER_LOCATION` are both set, but I think the change you just made to `gcloud-terraform-install` sets only location, right?

The logic is a little fussy, but I think we should check:

* if `location` is set but not `zone`, use `location` (good path)
* if `zone` is set but not `location`, use `zone` and warn (deprecated path)
* if `location` and `zone` are set but equal, use either and warn.
* if *both* are set, and not equal, throw an error

I'm not exactly sure how possible that is in .tf, but it seems like the right approach. If it's too much effort, though, holler.

chiayi: I will add back `zone` to `gcloud-terraform-install` so that both variables will be available. As for adding warnings and errors, not sure if that is possible within the terraform files?

zmerlynn: I don't think you need to add `zone` back in, I was pointing out that the Makefile is only sending the one - so there's no conflict between them.

gongmax:
I think it's better to also wire these options up to the [example module](https://github.com/googleforgames/agones/blob/release-1.27.0/examples/terraform-submodules/gke/module.tf) and update the website page in case customer is following the [instruction](https://agones.dev/site/docs/installation/terraform/gke/) to create the cluster

build/includes/google-cloud.mk
```diff
@@ -73,9 +80,7 @@ clean-gcloud-prow-build-cluster: $(ensure-build-image)
 # Pulls down authentication information for kubectl against a cluster, name can be specified through GCP_CLUSTER_NAME
 # (defaults to 'test-cluster')
 gcloud-auth-cluster: $(ensure-build-image)
-	docker run --rm $(common_mounts) $(build_tag) gcloud config set container/cluster $(GCP_CLUSTER_NAME)
-	docker run --rm $(common_mounts) $(build_tag) gcloud config set compute/zone $(GCP_CLUSTER_ZONE)
-	docker run --rm $(common_mounts) $(build_tag) gcloud container clusters get-credentials $(GCP_CLUSTER_NAME)
+	docker run --rm $(common_mounts) $(build_tag) gcloud container clusters get-credentials $(GCP_CLUSTER_NAME) --zone  $(GCP_CLUSTER_LOCATION) ;\
```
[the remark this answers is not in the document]
chiayi: Removed.

build/terraform/gke/module.tf
chiayi: Okay so I made it so that within the tf file, zone will be empty unless explicitly set. And it will only use zone if the variable is empty.

install/terraform/modules/gke/cluster.tf
```diff
@@ -56,7 +60,7 @@ resource "null_resource" "test-setting-variables" {
 
 resource "google_container_cluster" "primary" {
   name       = local.name
-  location   = local.zone
+  location   = local.zone == "" ? local.location : local.zone
```
zmerlynn, line 63: I would make this prioritize location.

It looks like warnings are not straightforward, so I think skipping that is fine. But at the very least, let's add an error if they're both non-empty and mismatched. I think it can be done like: https://medium.com/codex/terraform-variable-validation-b9b3e7eddd79

Later we can start erroring when `zone` is provided.

examples/terraform-submodules/gke/module.tf
```diff
@@ -53,9 +53,14 @@ variable "enable_image_streaming" {
   default = "true"
 }
 
-variable "zone" {
```
gongmax, line 56: We also need to update the [website page](https://github.com/googleforgames/agones/blob/main/site/content/en/docs/Installation/Terraform/gke.md) to indicate the change. Noted for changes to the website, we generally gate them so that they do not appear until the next release. Even though this doesn't require new features in Agones, it still makes sense to wait to publish this guidance until our next release.
https://agones.dev/site/docs/contribute/#within-a-page describes how to add the feature gate tags to the website pages.

install/terraform/modules/gke/cluster.tf
chiayi: If we make the location a priority, do we still want location to have a default value? This would also effect the validation done on the variable. 

As for cross variable validation, it's currently not supported but there seems to be hacky ways around it if we really need the check.

examples/terraform-submodules/gke/module.tf
chiayi: Done!

examples/terraform-submodules/gke/module.tf
```diff
@@ -53,9 +53,14 @@ variable "enable_image_streaming" {
...
+variable "location" {
   default     = "us-west1-c"
-  description = "The GCP zone to create the cluster in"
+  description = "The GCP location to create the cluster in"
+}
+
+variable "zone" {
+  default     = ""
+  description = "The GCP zone to create the cluster in (deprecated, use `location`)"
 }
```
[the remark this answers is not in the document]
chiayi: Added `autoscale`, `min_node_count` and `max_node_count`

install/terraform/modules/gke/cluster.tf
chiayi: So zone no longer has a default and will only be used as the location if zone is not empty.

site/content/en/docs/Installation/Terraform/gke.md
```diff
@@ -96,6 +96,14 @@ Configurable parameters:
 - gameserver_maxPort - the upper bound of the port range which gameservers will listen on (default is "8000")
 - gameserver_namespaces - a list of namespaces which will be used to run gameservers (default is `["default"]`). For example `["default", "xbox-gameservers", "mobile-gameservers"]`
 - force_update - whether or not to force the replacement/update of resource (default is true, false may be required to prevent immutability errors when updating the configuration)
+{{% feature publishVersion="1.28" %}}
+- location - the name of the [location](https://cloud.google.com/compute/docs/regions-zones) you want your cluster  to eb created in
```
gongmax, line 100: Better to callout the default values, like the other fields.

site/content/en/docs/Installation/Terraform/gke.md
```diff
@@ -96,6 +96,14 @@ Configurable parameters:
 - gameserver_maxPort - the upper bound of the port range which gameservers will listen on (default is "8000")
 - gameserver_namespaces - a list of namespaces which will be used to run gameservers (default is `["default"]`). For example `["default", "xbox-gameservers", "mobile-gameservers"]`
 - force_update - whether or not to force the replacement/update of resource (default is true, false may be required to prevent immutability errors when updating the configuration)
+{{% feature publishVersion="1.28" %}}
```
zmerlynn, line 99: Probably should add another version-guarded section at line 90 to mention that `zone` is deprecated.

zmerlynn (approved):

site/content/en/docs/Installation/Terraform/gke.md
chiayi: Added `feature expiry` to the zone. Not sure if this is the best way to do it.

site/content/en/docs/Installation/Terraform/gke.md
chiayi: Done!

gongmax (approved):

build/terraform/gke/module.tf
roberthbailey: I didn't see this comment thread earlier but I think where you ended up looks good - we use location and we can add to the release notes that folks should switch their configs to use location but we continue to support zone for N releases (maybe 1, maybe a few) to give people time to migrate their configs. 

We don't have a formal policy about compatibility of terraform configs (or helm for that matter) but we do try to do our best not to have knife edge changes that would break users upgrading across a single release.

install/terraform/modules/gke/variables.tf
```diff
@@ -30,7 +30,8 @@ variable "cluster" {
   type        = map
 
   default = {
-    "zone"                    = "us-west1-c"
+    "location"		      = "us-west1-c"
+    "zone"                    = ""
```
roberthbailey, line 34: Should we remove zone from here to only have location in our sample configs?

roberthbailey (commented):
I noticed while reviewing your change that in `build/includes/terraform.mk` there is a lot of duplication between the `gcloud-terraform-cluster` and `gcloud-terraform-install` targets. If you agree and think there would be a clean way to simplify them, that would be a good change for a follow-up PR.

install/terraform/modules/gke/variables.tf
chiayi: I think we can. There isn't really a reason to have zone here anymore.

roberthbailey (approved):

roberthbailey (approved):

PR #2816, opened 2022-11-16
Allow controller service account to update finalizers

### LINKED ISSUE #1740: Agones roles have insufficient permissions defined for clusters where OwnerReferencesPermissionEnforcement is enabled
**What happened**:
When trying to deploy a `GameServer`, I received an error.

A `GameServer` should deploy.

* create a k8s cluster
* enable `OwnerReferencesPermissionEnforcement` plugin
* install agones
* create a gameserver

**Environment**:
- Agones version: 1.7
- Kubernetes version (use `kubectl version`): 1.6 but likely any recent version
`pods "simple-udp-jcc4d" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>`

In the `agones-controller` role, explicitly adding the finalizers permissions (eg: `gameservers/finalizers`) at least is a workaround. I am not sure if that is the correct fix.

bostrt:
**What type of PR is this?**

/kind bug

In clusters where the OwnerReferencesPermissionEnforcement admission controller is enabled, the agones-controller service account needs to have access to update finalizers on fleets, gameserversets, and gameservers.

Closes #1740

I'm not sure if this is the fix you are looking for in regards to #1740 but it works great for me on OpenShift 4.11.


markmandel:
One thing I also realised - I'm fairly sure we only use finalizers on a `GameServer`, not on `Fleets` or `GameServerSets` - so do we need permission on all three, or just `GameServers` ?

bostrt:
> One thing I also realised - I'm fairly sure we only use finalizers on a GameServer, not on Fleets or GameServerSets - so do we need permission on all three, or just GameServers ?

When testing, I saw the same error when creating Fleets and GameServerSets but let me double-check!

markmandel:
But yeah - otherwise, no issues on my end with the change.

bostrt:
I just double-checked and all the new privileges will be needed from what I can tell. The owner references and finalizers begin at Pod and end at Fleet. Here's a link to more reference that's much to share in this comment :) https://gist.github.com/bostrt/aaf11005e72ec7456cf6b568653f38de

```shell
### POD 
$ oc get pod simple-game-server-nwfq2-b74cn -o yaml | yq '.kind,.metadata.finalizers,.metadata.ownerReferences'
Pod
null
- apiVersion: agones.dev/v1
  blockOwnerDeletion: true
  controller: true
  kind: GameServer
  name: simple-game-server-nwfq2-b74cn
  uid: cfe872cb-e221-4f68-890c-2b81d1eff6ac

### GAMESERVER
$ oc get gs simple-game-server-nwfq2-n756t -o yaml | yq '.kind,.metadata.finalizers,.metadata.ownerReferences'
GameServer
- agones.dev
- apiVersion: agones.dev/v1
  blockOwnerDeletion: true
  controller: true
  kind: GameServerSet
  name: simple-game-server-nwfq2
  uid: 340d0fe1-8c71-4442-b7c1-fdf235fc7088

### GAMESERVERSET
$ oc get gss simple-game-server-nwfq2 -o yaml | yq '.kind,.metadata.finalizers,.metadata.ownerReferences'
GameServerSet
null
- apiVersion: agones.dev/v1
  blockOwnerDeletion: true
  controller: true
  kind: Fleet
  name: simple-game-server
  uid: 4b3f1b14-9e5f-4e31-aabe-3627142ec52b

### FLEET
$ oc get fleet simple-game-server -o yaml | yq '.kind,.metadata.finalizers,.metadata.ownerReferences'
Fleet
null
null
```

markmandel:
Oh interesting! It's all about the `OwnerReference`, not about the actual finalizers! Fascinating.

@thoraxe if you have a chance to take a look and confirm this is good, that would be awesome, otherwise we can merge as is I think 👍🏻

roberthbailey (approved):

thoraxe:
I'm not sure I'm smart enough/knowledgeable enough to chime in. If this PR causes the reproducer to work, then I'm satisfied, for whatever that's worth!

PR #2822, opened 2022-11-22
fix: solve the native collection's memory leak detected by Unity 2021…

kingshijie:
….3.7f1

On Unity 2021.3.7f1, several error messages says "A Native Collection has not been disposed, resulting in a memory leak, Allocated from:xxx"; When look into this issue, it points into AgonesSdk and the UnityWebRequest's Dispose has not been called. After calling the Dispose() at the end of the method, the same error disappeared.

**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Fix the memory leak detected by the Unity; The whole error message is as follow: "A Native Collection has not been disposed, resulting in a memory leak, Allocated from:xxx";

You can reproduce the issue on Unity 2021.3.7f1, after the fix, the error will disappear.

Closes #

### DIFF
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -290,6 +290,8 @@ namespace Agones
                 Log($"Agones SendRequest failed: {api} {req.error}");
             }
 
+            req.Dispose();
+
             return result;
         }
 


davidyrgilbert:
In general, Unity recommends the "using" syntax for Webrequests to prevent issues like this one and guarantee a nicer cleanup. [[https://docs.unity3d.com/ScriptReference/Networking.UnityWebRequest.Dispose.html](https://docs.unity3d.com/ScriptReference/Networking.UnityWebRequest.Dispose.html)]
This probably goes beyoned this PR, but might be worth considering. 

Otherwise this looks like it'll fix my errors too, I've had the mentioned errors as well but didn't bother to track them down yet.

roberthbailey (approved):

roberthbailey:
@davidyrgilbert - I've created https://github.com/googleforgames/agones/issues/2830 with your suggestion and marked it as help wanted.

PR #2839, opened 2022-12-02
Remove trailing whitespace.

roberthbailey:
This allows the command can be copy/pasted.

**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

In the past two days I've copied this command a few times and gotten the error `unrecognized arguments: ` followed by `command not found: --enable-image-streaming`. It took me a few minutes to realize it's because there is an extra space, which means the backslash isn't escaping the newline which turns this into two commands (neither of which works). 

Closes #


zmerlynn (approved):

gongmax:
Maybe we can add some Markdown linter so this kind of issue can be captured automatically in future.

gongmax (approved):

PR #2840, opened 2022-12-02
GKE Autopilot: Separate game server workloads

zmerlynn:
This PR does a couple of things towards #2777:

* Enforce that on Autopilot the scheduling strategy is `Packed`

* If the user does not provide a nodeSelector or tolerations in the PodSpec, we add:

```
    nodeSelector:
      agones.dev/role: gameserver
    tolerations:
    - effect: NoSchedule
      key: agones.dev/role
      operator: Equal
      value: gameserver
```

This has the effect of splitting the game server pods off to their own nodes ala https://wdenniss.com/autopilot-workload-separation.

If the user has either defined, we assume they know what they're doing and skip this.

Note: In `Packed` we already set an affinity:

```
    affinity:
      podAffinity:
        preferredDuringSchedulingIgnoredDuringExecution:
        - podAffinityTerm:
            labelSelector:
              matchLabels:
                agones.dev/role: gameserver
            topologyKey: kubernetes.io/hostname
          weight: 100
```

We keep this affinity, but it becomes mostly a no-op after using a `nodeSelector`.

/kind feature

### DIFF
--- a/pkg/cloudproduct/cloudproduct.go
+++ b/pkg/cloudproduct/cloudproduct.go
@@ -41,6 +41,9 @@ type CloudProduct interface {
 	// ValidateGameServer is called by GameServer.Validate to allow for product specific validation.
 	ValidateGameServer(*agonesv1.GameServer) []metav1.StatusCause
 
+	// MutateGameServerPod is called by createGameServerPod to allow for product specific pod mutation.
+	MutateGameServerPod(*agonesv1.GameServer, *corev1.Pod) error
+
 	// NewPortAllocator creates a PortAllocator. See gameservers.NewPortAllocator for parameters.
 	NewPortAllocator(int32, int32, informers.SharedInformerFactory, externalversions.SharedInformerFactory) portallocator.Interface
 }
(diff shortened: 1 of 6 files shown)


roberthbailey (commented):
My only concern is that in pkg/gameservers/controller.go we create the pod and mutate it in very different places in the code. If we (for some reason) decide to set tolerations or a node selector on all game server pods in the future (e.g. in the call to `gs.Pod(sidecar)`) that will silently stop the gke autopilot provider from adding the extra fields. 

I think this could be caught by adding a unit test in pkg/gameservers/controller_test.go calling `createGameServerPod` _without_ the gke autopilot provider and verifying that we don't set any tolerations or node selectors. That would still make the code do the right thing if a user set one in the pod spec, but would cause a test to fail if we added code that did it for all pod specs.

zmerlynn:
> I think this could be caught by adding a unit test in pkg/gameservers/controller_test.go calling `createGameServerPod` _without_ the gke autopilot provider and verifying that we don't set any tolerations or node selectors. That would still make the code do the right thing if a user set one in the pod spec, but would cause a test to fail if we added code that did it for all pod specs.

Good call, done!

roberthbailey (approved):

Issue #2018, opened 2021-03-08
Flaky: C# conformance test because of invalid version string


markmandel:
**What happened**:

C# Sdk conformance flakes occasionally

```
/usr/share/dotnet/sdk/2.2.402/NuGet.targets(123,5): error : '1.13.0-0640590' is not a valid version string. [/go/src/agones.dev/agones/sdks/csharp/test/csharp-sdk-test.csproj]
/usr/share/dotnet/sdk/2.2.402/NuGet.targets(123,5): error : Parameter name: value [/go/src/agones.dev/agones/sdks/csharp/test/csharp-sdk-test.csproj]
includes/sdk.mk:88: recipe for target 'run-sdk-command' failed
```

**What you expected to happen**:

It to pass

**How to reproduce it (as minimally and precisely as possible)**:

Have a git hash that is not a valid NuGet Package version:
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning

**Anything else we need to know?**:

**Environment**: Cloud Build / Linux
- Agones version: Dev
- Kubernetes version (use `kubectl version`): 1.18
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): helm
- Troubleshooting guide log(s): N/A
- Others:

It _might_ be better to have a `{version}+{hash}` rather than a `{version}-{hash}` for the test.

gongmax:
I think it failed against the rule ["Numeric identifiers MUST NOT include leading zeroes"](https://semver.org/#spec-item-9)

gongmax:
The easiest fix I can think of is to append a character before the [git commit short hash](https://github.com/googleforgames/agones/blob/main/build/Makefile#L35) in the version. Will that introduce other issues? I'm not sure whether something else relies on format of the version.

markmandel:
I'm trying to work out where the version gets introduced to the csharp SDK, and I'm struggling.

igooch:
@Reousa  @rcreasey do you have any insight into this issue?

PR #2847, opened 2022-12-05
refactor: Implemented using the standard library

aimuz:

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind cleanup

Closes #

This pr may not be of much value, but I think it is a better practice to use the standard library implementation

### DIFF
--- a/pkg/util/signals/signals.go
+++ b/pkg/util/signals/signals.go
@@ -18,20 +18,12 @@ package signals
 
 import (
 	"context"
-	"os"
 	"os/signal"
 	"syscall"
 )
 
 // NewSigKillContext returns a Context that cancels when os.Interrupt or os.Kill is received
 func NewSigKillContext() context.Context {
-	ctx, cancel := context.WithCancel(context.Background())
-	c := make(chan os.Signal, 2)
-	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
-	go func() {
-		<-c
-		cancel()
-	}()
-
+	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	return ctx
 }


zmerlynn:
@aimuz Sadly retest doesn't work in this repo but I kicked off a new test.

This LGTM. We're in feature freeze and this is more of a cleanup, so I'm holding off until thaw to merge.

zmerlynn:
Bleh. Both of these failed on:

```
/usr/share/dotnet/sdk/3.1.425/NuGet.targets(128,5): error : '1.28.0-0361074' is not a valid version string. (Parameter 'value') [/go/src/agones.dev/agones/sdks/csharp/test/csharp-sdk-test.csproj]
```

which is #2018.

zmerlynn:
@aimuz You'll need to force-push this branch again to get the commit hash to change to get past that flake. It looks like you need to rebase anyways, so you could do that, and it'll retest after you push the rebase. If it again fails in the same place, you can re-push after `git commit --amend --date=now --no-edit` to just bump the time.

markmandel:
We can also wait until after stable, and update the branch when something else is merged. Should be fine.

pkg/util/signals/signals.go
```diff
@@ -18,20 +18,12 @@ package signals
...
 	"context"
-	"os"
 	"os/signal"
 	"syscall"
 )
 
 // NewSigKillContext returns a Context that cancels when os.Interrupt or os.Kill is received
 func NewSigKillContext() context.Context {
-	ctx, cancel := context.WithCancel(context.Background())
-	c := make(chan os.Signal, 2)
```
roberthbailey, line 29: @markmandel - I noticed that the library function creates a channel with size 1 regardless of how many signals are being tracked: https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/os/signal/signal.go;l=284. I assume that since it's part of the std library and the API is designed to handle multiple signals that it is safe (or even correct) to set the channel to size 1 but wanted to know if there was a reason that 5 years ago you set this to size 2: https://github.com/googleforgames/agones/commit/726af15d4d8b3d40435b74ea9bfa41d1607e2881#diff-e7b27b1d977569544a1feedd454f855f61c641a2f35bf9fe75d4bc15c7cefb35

markmandel: Honestly - not that I can remember! $10 says I copied from some other example that copied from another example?

I'd give more confidence to the Go std library than code I wrote 5 years ago 😄

zmerlynn (approved):

PR #2857, opened 2022-12-07
SafeToEvict: Implement Eviction API, add SetEviction cloud product hook

### LINKED ISSUE #2794: Explicit flag for "can be disrupted": eviction.safe
## Overview

In Agones today, you need intimate knowledge of sources of [voluntary pod disruption](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/) on your cluster to understand how to configure the `GameServer` specification best for your game server. For example:

*   Pod graceful termination (via [`terminationGracePeriodSeconds`](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/)) is allowed by all sources of voluntary disruption, but by default, [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#does-ca-respect-gracefultermination-in-scale-down) only permits 10m of graceful termination (and GKE does not allow this to be tuned). Additionally, pod graceful termination requires cooperation of the pod, as it [needs to intercept SIGTERM](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) (or [it must define a preStop hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/)).
*   The `safe-to-evict=false` pod annotation influences the behavior of [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node), but no other sources of voluntary disruption - but this flag is blocked on GKE Autopilot, as mentioned in #2777.
*   With #2807, you can use `PodDisruptionBudget` as a way to express “do not disrupt” for game server pods. This is roughly the equivalent of the `safe-to-evict=false` pod annotation, but works for anything that uses the standard eviction libraries (including GKE node upgrades, for up to an hour per node). However, `PodDisruptionBudget` [may be violated by pod preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#poddisruptionbudget-is-supported-but-not-guaranteed).
*  The safest way to control [pod preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption) is separating game servers to dedicated nodes, which is beyond the scope of this proposal (but ties nicely into something we can do in Autopilot ala #2840).

As long as you understand the complexity, you can today tune Agones to what you need. But some of the complexity changes between cloud products (GKE Autopilot is more opinionated than GKE Standard).

Here we propose a field `eviction` within the [`GameServer spec`](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/crds/_gameserverspecschema.yaml) that controls whether your game server understands [graceful termination](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/), and whether it should be disrupted for autoscaling events, upgrades, or allowed to run to completion without disruption.

## `eviction.safe`

We propose field `eviction` in `GameServer` with a single field `safe`, a string-typed enum with options `Always`, `Never`, `OnUpgrade`, like:

```
eviction:
  safe: Always
```

Each option uses a slightly different permutation of:
* the `safe-to-evict` annotation to block [Cluster Autoscaler based eviction](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node)
* and the `agones.dev/safe-to-evict` label selector to select the PDB from #2807, which blocks [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) and (for a limited time) [disruption from node upgrades](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets).
  * Note that PDBs do influence pod preemption as well, but it's not guaranteed.

As a quick reference:

| evictions.safe setting  |  `safe-to-evict` pod annotation |  `agones.dev/safe-to-evict` label |
|-----------------------|---------------------------------|----------------------------------|
| `Never` (default)         | `false`                                            | `false` (matches PDB)                  |
| `OnUpgrade`               | `false`                                            | `true` (does not match PDB)        |
| `Always`                       | `true`                                             | `true` (does not match PDB)        |

### `safe: Never` (the default)

Agones ensures the game server pod runs to completion by:
*   enabling the `safe-to-evict=false` pod annotation
*   applying the PDB from #2807

### `safe: OnUpgrade`

Agones blocks Cluster Autoscaler based evictions, but allows eviction from upgrades. On certain providers (GKE, for example), the graceful termination support for upgrades is much longer than is support by Cluster Autoscaler (1h vs 10m), so supporting eviction for upgrades alone may work well.

The game server binary must be prepared to receive `SIGTERM` for evictions. If the default [grace period of 30s](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) is insufficient, you must tune `terminationGracePeriodSeconds` in the pod spec as well. For broad support across cloud providers, the game server must exit within 1h after receiving `SIGTERM`.

### `safe: Always`

Allow evictions by both Cluster Autoscaler and upgrades. As with `OnUpgrade`, the game server binary must be prepared to receive `SIGTERM`. If the default [grace period of 30s](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) is insufficient, you must tune `terminationGracePeriodSeconds` in the pod spec as well.

By default, Cluster Autoscaler supports only 10m of graceful termination. For broad support across cloud providers, the game server must exit within 10m after receiving `SIGTERM` if running on a cluster where Cluster Autoscaler down-scaling is enabled (or running on a product that implicitly does, like GKE Autopilot). As with `OnUpgrade`, one hour of graceful termination is broadly supported if Autoscaler is not enabled (and in fact, if Autoscaler is not enabled, `safe: Always` is the same as `safe: OnUpgrade`).

## Backward Compatibility with `safe-to-evict=true` pod annotation; Existing overrides in template 

We propose that using a similar approach to #2754, for any case where Agones modifies the pod, we honor the provided [`GameServer.template`](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/crds/_gameserverspecschema.yaml#L33-L37) in preference to anything Agones would set automatically - for example, if you set the `agones.dev/safe-to-evict` label in `template`, Agones assumes you know better. We will [warn](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response) when such conflicts occur to make it obvious, though.

That said, one case we need to be particularly sensitive of is the case in #2754 itself: `safe` defaults to `Never`. Mixing the `safe: Never` `GameServer` default with the `safe-to-evict=true` pod annotation doesn’t make sense: the inflexible `PodDisruptionBudget` [prevents Cluster Autoscaler scale-down](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#does-ca-work-with-poddisruptionbudget-in-scale-down), even if `safe-to-evict=true` would allow it for other cases.

Prior to the proposal here, the `safe-to-evict` pod annotation is currently the only [documented Agones disruption control](https://agones.dev/site/docs/advanced/scheduling-and-autoscaling/#cluster-autoscaler-1). To maintain backwards compatibility, we match the original intent of the `safe-to-evict=true` pod annotation, which is to allow graceful terminatio: If the `safe-to-evict=true` pod annotation is present in the template, we enable `safe: Always` in the `GameServer`.

We do not make a similar inference for `terminationGracePeriodSeconds` (despite it also being in our documentation) as it is unnecessary.

With this one piece of backward compatibility resolved, it’s safe to enable the `safe: Never` default without breaking existing workloads.

## Feature gate: new `SafeToEvict` replaces `LifecycleContract`.

We propose a new feature gate `SafeToEvict` (which will replace `LifecycleContract`). Graduation to Beta should be relatively rapid as the feature has broad backward compatibility. The only situation where backward compatibility is violated is if the user is negatively impacted by a PDB on the game server pods, e.g. the user is relying on the default `safe-to-evict=false` pod annotation, but _also_ is assuming that node drain (e.g. via cluster upgrade) will evict. Given that our existing documentation doesn't allow for node drains, we think this is a safe case for backwards incompatibility. To be conservative about it, though, when we graduate `SafeToEvict` to `Beta`, we will flag it a potentially breaking change in Release Notes to call attention to it. To allow this situation, the user simply needs to set `eviction: safe: OnUpgrade`. 

### Why not default to `OnUpgrade` then?

Despite the possibility of introducing a behavior change, we still think `Never` is the right default. `OnUpgrade` requires the game server to understand `SIGTERM`, and typically terminate within an hour.

## Dependency on `SDKGracefulTermination` feature gate

Anything but `safe: Never` relies on the `SDKGracefulTermination` feature gate. If `SDKGracefulTermination` is disabled but the `GameServer` has `safe: Always` or `safe: OnUpgrade` (explicitly, or implicitly via backwards compatibility), we will warn but allow it.

In #2831, we proposed graduating `SDKGracefulTermination` to Beta, so this situation should only occur if `SDKGracefulTermination` is explicitly disabled and `SafeToEvict` is explicitly enabled, which seems like an odd choice.

zmerlynn:
Implements #2794, in three separate commits:

* API changes under `SafeToEvict` feature gate:
  * Add the `eviction` object and `safe` property
  * Implement the defaulting/validation for `eviction.safe` field. In particular, it enshrines the defaulting logic for the somewhat complicated backwards compatibility case (`safe-to-evict=true` is set but `eviction.safe` is not).

* Runs `gen-api-docs`, as a separate commit.

* Then we flow policy from `GameServer.Status.Eviction.Safe`:
  * Following the pattern in #2840, moves the enforcement of cloudproduct specific things up to the controller, adding a new `cloudproduct.SetEviction` hook to handle it.
    * As a result, when the `SafeToEvict` feature flag is on, we disable annotation in gameserver.go and test that we don't set anything (as a change detection test).
  * Adds a generic and GKE Autopilot version of how to set policy from the `eviction.safe` status. In Autopilot, we disallow `eviction.safe: OnUpgrade` - this configuration doesn't make sense since we can't use `safe-to-evict=false` to restrict the Autoscaler.

### DIFF
--- a/install/helm/agones/templates/crds/_gameserverspecschema.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecschema.yaml
@@ -154,6 +154,21 @@ properties:
             title: The initial player capacity of this Game Server
             minimum: 0
 {{- if .featureSafeToEvict }}
+      eviction:
+        type: object
+        title: Eviction tolerance of the game server
+        properties:
+          safe:
+            type: string
+            title: Game server supports termination via SIGTERM
+            description: |
+              - Never: The game server should run to completion. Agones sets Pod annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` and label `agones.dev/safe-to-evict: "false"`, which matches a restrictive PodDisruptionBudget.
+              - OnUpgrade: On SIGTERM, the game server will exit within `terminationGracePeriodSeconds` or be terminated; Agones sets Pod annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"`, which blocks evictions by Cluster Autoscaler. Evictions from node upgrades proceed normally.
+              - Always: On SIGTERM, the game server will exit within `terminationGracePeriodSeconds` or be terminated, typically within 10m; Agones sets Pod annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"`, which allows evictions by Cluster Autoscaler.
+            enum:
+            - Always
+            - OnUpgrade
+            - Never
       immutableReplicas:
         type: integer
         title: Immutable count of Pods to a GameServer. Always 1. (Implementation detail of implementing the Scale subresource.)
--- a/pkg/apis/agones/v1/apihooks.go
+++ b/pkg/apis/agones/v1/apihooks.go
@@ -15,6 +15,8 @@
 package v1
 
 import (
+	"agones.dev/agones/pkg/util/runtime"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -28,6 +30,9 @@ type APIHooks interface {
 
 	// MutateGameServerPodSpec is called by createGameServerPod to allow for product specific pod mutation.
 	MutateGameServerPodSpec(*GameServerSpec, *corev1.PodSpec) error
+
+	// SetEviction is called by gs.Pod to enforce GameServer.Status.Eviction.
+	SetEviction(EvictionSafe, *corev1.Pod) error
 }
 
 var apiHooks APIHooks = generic{}
@@ -45,3 +50,37 @@ type generic struct{}
 
 func (generic) ValidateGameServerSpec(*GameServerSpec) []metav1.StatusCause    { return nil }
 func (generic) MutateGameServerPodSpec(*GameServerSpec, *corev1.PodSpec) error { return nil }
+
+// SetEviction sets disruptions controls based on GameServer.Status.Eviction.
+func (generic) SetEviction(safe EvictionSafe, pod *corev1.Pod) error {
+	if !runtime.FeatureEnabled(runtime.FeatureSafeToEvict) {
+		return nil
+	}
+	if _, exists := pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; !exists {
+		switch safe {
+		case EvictionSafeAlways:
+			pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = True
+		case EvictionSafeOnUpgrade, EvictionSafeNever:
+			// For EvictionSafeOnUpgrade and EvictionSafeNever, we block Cluster Autoscaler.
+			pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = False
+		default:
+			return errors.Errorf("unknown eviction.safe value %q", string(safe))
+		}
+	}
+	if _, exists := pod.ObjectMeta.Labels[SafeToEvictLabel]; !exists {
+		switch safe {
+		case EvictionSafeAlways, EvictionSafeOnUpgrade:
+			// For EvictionSafeAlways and EvictionSafeOnUpgrade, we use a label value
+			// that does not match the agones-gameserver-safe-to-evict-false PDB. But
+			// we go ahead and label it, in case someone wants to adopt custom logic
+			// for this group of game servers.
+			pod.ObjectMeta.Labels[SafeToEvictLabel] = True
+		case EvictionSafeNever:
+			// For EvictionSafeNever, match gones-gameserver-safe-to-evict-false PDB.
+			pod.ObjectMeta.Labels[SafeToEvictLabel] = False
+		default:
+			return errors.Errorf("unknown eviction.safe value %q", string(safe))
+		}
+	}
+	return nil
+}
--- a/pkg/apis/agones/v1/apihooks_test.go
+++ b/pkg/apis/agones/v1/apihooks_test.go
@@ -1,0 +1,123 @@
+// Copyright 2022 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import (
+	"testing"
+
+	"agones.dev/agones/pkg/util/runtime"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestSetEviction(t *testing.T) {
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
+
+	emptyPodAnd := func(f func(*corev1.Pod)) *corev1.Pod {
+		pod := &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{},
+				Labels:      map[string]string{},
+			},
+		}
+		f(pod)
+		return pod
+	}
+	for desc, tc := range map[string]struct {
+		featureFlags string
+		safeToEvict  EvictionSafe
+		pod          *corev1.Pod
+		wantPod      *corev1.Pod
+	}{
+		"SafeToEvict feature gate disabled => no change": {
+			featureFlags: "SafeToEvict=false",
+			// intentionally leave pod nil, it'll crash if anything's touched.
+		},
+		"SafeToEvict: Always, no incoming labels/annotations": {
+			featureFlags: "SafeToEvict=true",
+			safeToEvict:  EvictionSafeAlways,
+			pod:          emptyPodAnd(func(*corev1.Pod) {}),
+			wantPod: emptyPodAnd(func(pod *corev1.Pod) {
+				pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = True
+				pod.ObjectMeta.Labels[SafeToEvictLabel] = True
+			}),
+		},
+		"SafeToEvict: OnUpgrade, no incoming labels/annotations": {
+			featureFlags: "SafeToEvict=true",
+			safeToEvict:  EvictionSafeOnUpgrade,
+			pod:          emptyPodAnd(func(*corev1.Pod) {}),
+			wantPod: emptyPodAnd(func(pod *corev1.Pod) {
+				pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = False
+				pod.ObjectMeta.Labels[SafeToEvictLabel] = True
+			}),
+		},
+		"SafeToEvict: Never, no incoming labels/annotations": {
+			featureFlags: "SafeToEvict=true",
+			safeToEvict:  EvictionSafeNever,
+			pod:          emptyPodAnd(func(*corev1.Pod) {}),
+			wantPod: emptyPodAnd(func(pod *corev1.Pod) {
+				pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = False
+				pod.ObjectMeta.Labels[SafeToEvictLabel] = False
+			}),
+		},
+		"SafeToEvict: Always, incoming labels/annotations": {
+			featureFlags: "SafeToEvict=true",
+			safeToEvict:  EvictionSafeAlways,
+			pod: emptyPodAnd(func(pod *corev1.Pod) {
+				pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = "just don't touch, ok?"
+				pod.ObjectMeta.Labels[SafeToEvictLabel] = "seriously, leave it"
+			}),
+			wantPod: emptyPodAnd(func(pod *corev1.Pod) {
+				pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = "just don't touch, ok?"
+				pod.ObjectMeta.Labels[SafeToEvictLabel] = "seriously, leave it"
+			}),
+		},
+		"SafeToEvict: OnUpgrade, incoming labels/annotations": {
+			featureFlags: "SafeToEvict=true",
+			safeToEvict:  EvictionSafeOnUpgrade,
+			pod: emptyPodAnd(func(pod *corev1.Pod) {
+				pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = "better not touch"
+				pod.ObjectMeta.Labels[SafeToEvictLabel] = "not another one"
+			}),
+			wantPod: emptyPodAnd(func(pod *corev1.Pod) {
+				pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = "better not touch"
+				pod.ObjectMeta.Labels[SafeToEvictLabel] = "not another one"
+			}),
+		},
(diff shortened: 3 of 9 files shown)


zmerlynn:
(We're changing the design a bit.)

zmerlynn:
PTAL. Note that the bottom commit is #2840, so you can ignore it here.

zmerlynn:
> PTAL. Note that the bottom commit is https://github.com/googleforgames/agones/pull/2840, so you can ignore it here.

#2840 is merged, rebased this one. Ready to go!

zmerlynn:
Rebased on top of #2868, which is now the last commit of this PR - let's work on merging that first.

The other 3 commits are detailed in the top comment and can be reviewed separately (the middle commit, for example, is purely mechanical).

zmerlynn:
Rebased now that #2868 is merged.

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -150,63 +187,25 @@ func TestGameServerApplyDefaults(t *testing.T) {
...
-					PeriodSeconds:       5,
-				},
-				sdkServer: SdkServer{
-					LogLevel: SdkServerLogLevelInfo,
-					GRPCPort: 9357,
-					HTTPPort: 9358,
-				},
-				alphaPlayerCapacity: &ten,
-			},
+			gameServer: defaultGameServerAnd(func(gss *GameServerSpec) {
```
markmandel, line 195: I like the name of the function 👍🏻

markmandel (approved):
I think this looks pretty good! @roberthbailey did you want to give it a check?

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -154,6 +154,27 @@ properties:
...
+            title: Game server supports termination via SIGTERM
+            description: |
+              - Never: The game server should run to completion. Agones sets Pod annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` and label `agones.dev/safe-to-evict: "false"`, which matches a restrictive PodDisruptionBudget.
+              - OnUpgrade: On SIGTERM, the game server will exit within `terminationGracePeriodSeconds` or be terminated; Agones sets Pod annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"`, which blocks evictions by Cluster Autoscaler. Evictions from node upgrades proceed normally.
+              - Always: On SIGTERM, the game server will exit within `terminationGracePeriodSeconds` or be terminated, typically within 10m; Agones sets Pod annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"`, which allows evictions by Cluster Autoscaler.
+            enum:
+            - Always
+            - OnUpgrade
+            - Never
+      safeToEvict:
```
roberthbailey, line 172: I think lines 172-177 should be removed - they look leftover from before the field was renamed.

pkg/apis/agones/v1/apihooks.go
```diff
@@ -45,3 +49,31 @@ type generic struct{}
...
+// SetEviction sets disruptions controls based on GameServer.Status.Eviction.
+func (generic) SetEviction(safe EvictionSafe, pod *corev1.Pod) error {
+	if !runtime.FeatureEnabled(runtime.FeatureSafeToEvict) {
+		return nil
+	}
+	if _, exists := pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; !exists {
+		if safe == EvictionSafeAlways {
+			pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = True
+		} else {
+			// For SafeToEvictOnUpgrade and SafeToEvictNever, we block Cluster Autoscaler.
```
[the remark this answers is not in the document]
zmerlynn: Yup. Fixed this and below - I used a symbolic rename but forgot to actually do a search&replace as well.

pkg/apis/agones/v1/apihooks.go
```diff
@@ -45,3 +49,31 @@ type generic struct{}
...
+			pod.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = False
+		}
+	}
+	if _, exists := pod.ObjectMeta.Labels[SafeToEvictLabel]; !exists {
+		if safe != EvictionSafeNever {
```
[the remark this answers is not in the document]
zmerlynn: Er, disregard that comment. :) Yeah, I think swapping makes sense. I was going by the smmetry of the True/False case for the annotation, but it does read a little weird if you just look at it from the conditionals.

pkg/apis/agones/v1/apihooks_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}{
+		"SafeToEvict feature gate disabled => no change": {
+			featureFlags: "SafeToEvict=false",
+			// intentionally leave pod nil, it'll crash if anything's touched.
+		},
+		"SafeToEvict: Always, no incoming labels/annotations": {
+			featureFlags: "SafeToEvict=true",
+			safeToEvict:  EvictionSafeAlways,
+			pod:          emptyPodAnd(func(*corev1.Pod) {}),
+			wantPod: emptyPodAnd(func(pod *corev1.Pod) {
```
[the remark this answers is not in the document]
zmerlynn: It did (I don't think Agones-Bot would let it past otherwise) .. I think it's because it's a multi-line, but agree it's a bit of an odd formatting. :shrug:

pkg/apis/agones/v1/gameserver.go
```diff
@@ -161,6 +178,8 @@ type GameServerSpec struct {
 	// (Alpha, PlayerTracking feature flag) Players provides the configuration for player tracking features.
 	// +optional
 	Players *PlayersSpec `json:"players,omitempty"`
+	// (Alpha, SafeToEvict feature flag) Eviction specifies the eviction tolerance of the GameServer.
```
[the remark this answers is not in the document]
zmerlynn: We seem pretty inconsistent here. Health, Scheduling, SdkServer can all be left unspecified as well. I added it because I agree with you, but I wonder if we should look at this more carefully.

pkg/cloudproduct/gke/gke.go
```diff
@@ -127,13 +128,50 @@ func (*gkeAutopilot) ValidateGameServerSpec(gss *agonesv1.GameServerSpec) []meta
 			Message: errSchedulingMustBePacked,
 		})
 	}
+	if gss.Eviction.Safe == agonesv1.EvictionSafeOnUpgrade {
```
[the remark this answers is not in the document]
zmerlynn: No - remember that the 100% availability PDB will also block Autoscaler eviction. But the overly restrictive PDB is kind of a big gun that's great for `Never`, but behaves badly for `OnUpgrade`, which is why we block it.

I added a reference to the SetEviction comment, and tried to add some clarifying language there.

pkg/cloudproduct/gke/gke.go
```diff
@@ -38,8 +38,9 @@ const (
 	noWorkloadDefaulter          = "failed to get MutatingWebhookConfigurations/workload-defaulter.config.common-webhooks.networking.gke.io (error expected if not on GKE Autopilot)"
 	hostPortAssignmentAnnotation = "autopilot.gke.io/host-port-assignment"
 
-	errPortPolicyMustBeDynamic = "PortPolicy must be Dynamic on GKE Autopilot"
-	errSchedulingMustBePacked  = "Scheduling strategy must be Packed on GKE Autopilot"
+	errPortPolicyMustBeDynamic = "portPolicy must be Dynamic on GKE Autopilot"
+	errSchedulingMustBePacked  = "scheduling strategy must be Packed on GKE Autopilot"
+	errSafeToEvictNotOnUpgrade = "safeToEvict OnUpgrade not supported on GKE Autopilot"
```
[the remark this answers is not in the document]
zmerlynn: Yup, and changed the variable name.

markmandel:
One thing I am noting - we are closing #2794 with this PR. Do we need documentation to go along with this work, before we can consider this ticket "closed"?

zmerlynn:
> One thing I am noting - we are closing #2794 with this PR. Do we need documentation to go along with this work, before we can consider this ticket "closed"?

Yeah, probably. I'll leave it open for now (removed the PR line).

pkg/apis/agones/v1/gameserver.go
```diff
@@ -356,6 +394,25 @@ func (gss *GameServerSpec) applySchedulingDefaults() {
...
+	if gss.Eviction.Safe == "" {
+		gss.Eviction.Safe = EvictionSafeNever
+	}
+}
+
+func (gs *GameServer) applyEvictionStatus() {
+	if !runtime.FeatureEnabled(runtime.FeatureSafeToEvict) {
+		return
+	}
+	gs.Status.Eviction = gs.Spec.Eviction
```
markmandel, line 410: I missed this bit before - but if the value of status.eviction is never going to change... why do we have a status value for eviction?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -356,6 +394,25 @@ func (gss *GameServerSpec) applySchedulingDefaults() {
...
+	if gs.Spec.Template.ObjectMeta.Annotations[PodSafeToEvictAnnotation] == "true" {
```
markmandel, line 411: Is this the reason why we have a status value?

This feels a little magical to me. I'm not sure it's necessary. I'd be tempted to get rid of gs.status.eviction and just rely on what the user specifies in gs.spec.eviction.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -373,6 +430,16 @@ func (gss *GameServerSpec) Validate(devAddress string) ([]metav1.StatusCause, bo
 		}
 	}
 
+	if !runtime.FeatureEnabled(runtime.FeatureSafeToEvict) {
+		if gss.Eviction.Safe != "" {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueNotSupported,
+				Field:   "safeToEvict",
```
roberthbailey, line 437: I found one other place where I think the field name needs to be updated.

pkg/apis/agones/v1/gameserver.go
zmerlynn: This is why there's a status, yes, to make explicit that we are inferring the .Eviction settings rather than using the Spec. The intention was to demystify this particular magic - we only base decisions on the Status. The inference from the annotation was mentioned in the design as necessary for backwards compatibility.

pkg/apis/agones/v1/gameserver.go
zmerlynn: (responded in https://github.com/googleforgames/agones/pull/2857#discussion_r1062784460)

pkg/apis/agones/v1/gameserver.go
markmandel: Got it. re-read the design notes - that makes sense. Objection rescinded  👍🏻

pkg/apis/agones/v1/gameserver.go
zmerlynn: Fixed!

roberthbailey:
The latest run seems to have flaked in e2e-feature-gates:

```
    --- FAIL: TestFleetRollingUpdate/Use_fleet_Patch_true_25%_cycle_true (107.33s)
        fleet_test.go:413: 
            	Error Trace:	fleet_test.go:413
```

This particular error seems to be happening more frequently as of late.

roberthbailey (approved):

PR #2863, opened 2022-12-10
Return better error message when mutation webhook gets invalid JSON

zmerlynn:
I know this looks a little ridiculous since we're returning the error twice, but it looks nicer on the command line. If there's garbage in the manifest (in this case I changed `portPolicy` to `true`, it goes from:

```
The GameServer "" is invalid
```

To:

```
The GameServer "" is invalid: : error unmarshalling original GameServer json: {"apiVersion":"agones.dev/v1","kind":"GameServer","metadata":{"creationTimestamp":null,"generateName":"simple-game-server-","namespace":"default"},"spec":{"health":{"failureThreshold":6},"immutableReplicas":1,"ports":[{"containerPort":false,"name":"default","portPolicy":"Dyanmic"}],"template":{"spec":{"containers":[{"image":"gcr.io/agones-images/simple-game-server:0.14","name":"simple-game-server","resources":{"limits":{"cpu":"20m","memory":"64Mi"},"requests":{"cpu":"20m","memory":"64Mi"}}}]}}}}: json: cannot unmarshal bool into Go struct field GameServerPort.spec.ports.containerPort of type int32
```

### DIFF
--- a/pkg/util/webhooks/webhooks.go
+++ b/pkg/util/webhooks/webhooks.go
@@ -94,13 +94,14 @@ func (wh *WebHook) handle(path string, w http.ResponseWriter, r *http.Request) e
 
 			review, err = oh.handler(review)
 			if err != nil {
-				causes := make([]metav1.StatusCause, 0)
 				review.Response.Allowed = false
 				details := metav1.StatusDetails{
-					Name:   review.Request.Name,
-					Group:  review.Request.Kind.Group,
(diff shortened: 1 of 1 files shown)


gongmax:
LGTM
The PR description seems not accurate. From the cli output, seems you changed `containerPort` to `false`, rather than `portPolicy` to `true` as stated in the description.

gongmax (approved):

PR #2865, opened 2022-12-12
Skip validation errors in mutating webhooks

### LINKED ISSUE #1770: GameServerTemplate validation: no description when used big port values
**What happened**:
Inconsistent error message when using a big value for `grpcPort` field  `grpcPort: 33332229357`.
```
The Fleet "" is invalid
```

Consistent error message without dependency on port field length.

Change `example/fleet.yaml` lines starting from 65 to :
```
      sdkServer:
        # sdkServer log level parameter has three options:
        #  - "Info" (default) The SDK server will output all messages except for debug messages
        #  - "Debug" The SDK server will output all messages including debug messages
        #  - "Error" The SDK server will only output error messages
        logLevel: Info
        grpcPort: 33332229357
        httpPort: 9358
```
`kubectl apply -f ` resulted fleet:
```
The Fleet "" is invalid
```

Do the same with smaller grpcPort value `2229357`:
```
spec.template.spec.sdkServer.grpcPort in body should be less than or equal to 65535
```

**Environment**:
- Agones version:  1.8 (master)
- Kubernetes version (use `kubectl version`): 1.15
- Install method (yaml/helm): Terraform Helm
- Others:

zmerlynn:
Request processing by kube-apiserver is roughly:
  { mutating webhooks } => { OpenAPI schema validation } => { validating webhooks }

In this PR, we disable validation errors (at least, JSON unmarshalling validation errors) during mutating webhook processing. It seems counter-intuitive not to return an error in mutating webhooks for a serious violation such as invalid JSON, but doing so allows OpenAPI schema validation to proceed. Schema validation gives us nice errors, so e.g.:

instead of:

```
The Fleet "" is invalid
```

we get:

```
The Fleet "fleet-example" is invalid: spec.template.spec.sdkServer.grpcPort: Invalid value: 33332229357: spec.template.spec.sdkServer.grpcPort in body should be less than or equal to 65535
```

This will become all the more important if we eventually move to CEL based validations.

I'm also changing the error in #2863 to an internal error - after this PR, any hard error from a handler (vs a cause) is unexpected.

Fixes #1770

### DIFF
--- a/pkg/fleetautoscalers/controller.go
+++ b/pkg/fleetautoscalers/controller.go
@@ -198,8 +198,9 @@ func (c *Controller) mutationHandler(review admissionv1.AdmissionReview) (admiss
 	fas := &autoscalingv1.FleetAutoscaler{}
 	err := json.Unmarshal(obj.Raw, fas)
 	if err != nil {
-		c.baseLogger.WithField("review", review).WithError(err).Error("validationHandler")
-		return review, errors.Wrapf(err, "error unmarshalling original FleetAutoscaler json: %s", obj.Raw)
+		// If the JSON is invalid during mutation, fall through to validation. This allows OpenAPI schema validation
+		// to proceed, resulting in a more user friendly error message.
+		return review, nil
 	}
 
 	fas.ApplyDefaults()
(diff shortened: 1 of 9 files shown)


gongmax:
Code LGTM. Is there any downside to not fall fast on an invalid request? Maybe not a big deal but potential increasing latency for those request?

zmerlynn:
> Is there any downside to not fall fast on an invalid request? Maybe not a big deal but potential increasing latency for those request?

Yes, there's a minor latency difference, I suspect. In theory there's also a minor load difference on the Kubernetes Control Plane. But, importantly, it only effects the error case, which is uncommon to optimize for.

gongmax (approved):

PR #2875, opened 2022-12-16
Move all actual Agones releases images to GAR

### LINKED ISSUE #2358: Consider moving agones system images from gcr.io to GCP's artifact registry
GCP's [Artifact Registry](https://cloud.google.com/artifact-registry/) is an alternate way that we can store container images in Google Cloud (we currently store them in Google Container Registry - gcr.io). 

If we put images into the Artifact Registry, we can then take advantage of the recently announced [GKE image streaming for fast application startup and autoscaling](https://cloud.google.com/blog/products/containers-kubernetes/introducing-container-image-streaming-in-gke). This may not be a huge improvement for our linux images (which are pretty small containers) but our windows binaries (the example game server and sdk sidecar) are both pretty large and may see a larger benefit. 

* Keep the status quo.
* We can push images to both locations during a migration period.

gongmax:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Closes #2358 

I created the registry `us-docker.pkg.dev/agones-images/release` manually.
Tested by following the instructions to [run a test GKE cluster](https://github.com/googleforgames/agones/tree/main/build#running-a-test-google-kubernetes-engine-cluster)


build/README.md
```diff
@@ -226,9 +226,9 @@ installing Agones next.
 
 To prepare building and pushing images, let's set the REGISTRY environment variable to point to our new project.
 You can [choose any registry region](https://cloud.google.com/container-registry/docs/pushing-and-pulling#choosing_a_registry_name)
```
roberthbailey, line 228: This link is for GCR (not artifact registry). If we are suggesting that folks use GAR instead of GCR are there other pre-requisites that need to be updated (like creating a registry in the project)?

build/README.md
```diff
@@ -417,7 +417,7 @@ command from within `make shell` or locally, to refresh your authentication toke
 The version of this build. Version defaults to the short hash of the latest commit.
 
 ### REGISTRY
-The registry that is being used to store docker images. Defaults to gcr.io/agones-images - the release + CI registry.
+The registry that is being used to store docker images. Defaults to us-docker.pkg.dev/agones-images/release
```
roberthbailey, line 420: Since we can now split out registries, I'm wondering if the default (e.g. if you forget to set the `REGISTRY` environment variable) should be something else to make it harder to accidentally push to the release registry. 

Should the registry default to something like `us-docker.pkg.dev/agones-images/dev`? And be overridden in the `make do-release` target to point to the release registry?

Ideally we'd have cloud build (or some similar build automation) creating our releases and restrict access to the release registry to only automation.

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -129,7 +129,7 @@ is to have them both run using the host network, like so:
 In one shell run:
 
 ```shell
-docker run --network=host --rm gcr.io/agones-images/agones-sdk:{{< release-version >}} --local
+docker run --network=host --rm us-docker.pkg.dev/agones-images/release/agones-sdk:{{< release-version >}} --local
```
roberthbailey, line 132: The website changes should be behind a release guard. 

build/README.md
gongmax: The `us-docker.pkg.dev/agones-images/release` registry only give `allUsers` reader access so `make push` won't succeed without override the `REGISTRY`.
Do we want to allow users to push image to any of the registry under `agones-images` project? Should we just enforce users to use their own registry for development?
In our current setting, the `gcr.io` registry has public visibility. Does it mean arbitrary users can push images to it? I understand it gives all users pull permission, put didn't find any information about push in the public docs.

roberthbailey: > `-docker.pkg.dev/agones-images/release` registry only give `allUsers` reader access so `make push` won't succeed without override the `REGISTRY`.

But for maintainers that have editor access on the project, we can still push to it (so we can cut releases) but this means we can also inadvertently push to it. I've done this in the past with the GCR bucket where I forgot to set the `REGISTRY` variable before running a push command and accidentally pushed into our official bucket (luckily my build looked like any other CI build so it didn't cause any problems). But now that CI is split out from releases, if I accidentally pushed it would put an extraneous build into the releases bucket. 

> Do we want to allow users to push image to any of the registry under agones-images project? Should we just enforce users to use their own registry for development?

Correct. Users should use their own. But as I mentioned above for folks that have access to the agones-images project so we can push official builds we can accidentally clobber official releases.

gongmax: Oh, I see your point. Didn't realize the editors have all the access. Makes perfect sense!

markmandel: I like the idea of setting it to something default that isn't a real registry.

A couple of suggestions:
1. Make it blank by default. So images are local only unless you specify a REGISTRY.
2. Make it "agones-images", so while it doesn't exist, it does group all Agones with a prefix.

I did check, and all images start wtih `agones-`, so I'd be tempted to lean towards No. 1, but No. 2 may be safer over time, just because of the extra prefix.

gongmax: If we leave it as blank or only make it a prefix, the docker commands (such as in `make push` `make install`) will default it as a docker registry, which leading to more confusion I think.

markmandel: 🤔 why do you think it will create more confusion? Either way, it still ends up failing if someone attempts to push the image.

But to that point - that would say to me that leaving is blank is actually the best option to indicate "this is only a local version, not a release version you've got in your local images".

If I run `docker images` and I see the following:

```shell
❯ docker images
REPOSITORY                TAG                          IMAGE ID       CREATED              SIZE
agones-allocator          1.29.0-c61a21a-amd64         8df912cba403   About a minute ago   68.4MB
agones-ping               1.29.0-c61a21a-amd64         ee443e7013f0   About a minute ago   24.4MB
agones-extensions         1.29.0-c61a21a-amd64         a38e43db5912   2 minutes ago        69.4MB
agones-controller         1.29.0-c61a21a-amd64         55e5cd993a40   2 minutes ago        69.4MB
agones-sdk                1.29.0-c61a21a-linux-amd64   8a0612f7cfe4   3 minutes ago        65.3MB
agones-build              f1d1057367                   cd6bfe544e90   4 weeks ago          3.57GB
agones-build-sdk-node     2a5f9e6f1a                   6967d5d17df4   7 weeks ago          1.83GB
agones-build-sdk-base     4e48d2360c                   5348986fd7ce   7 weeks ago          1.59GB
debian                    bookworm                     0ac8b312578d   5 months ago         119MB
```

I think that's fairly obvious that it isn't coming from a Docker Hub -- but what do you think?

Also, if we do set a prefix of `agones-images`  - then someone could create a user of `agones-images` on Docker Hub, and I think that _would_ definitely be confusing 😄 

Just to see it side by side with the other suggestion:

```
❯ docker images
REPOSITORY                TAG                          IMAGE ID       CREATED              SIZE
agones-images/agones-allocator          1.29.0-c61a21a-amd64         8df912cba403   About a minute ago   68.4MB
agones-images/agones-ping               1.29.0-c61a21a-amd64         ee443e7013f0   About a minute ago   24.4MB
agones-images/agones-extensions         1.29.0-c61a21a-amd64         a38e43db5912   2 minutes ago        69.4MB
agones-images/agones-controller         1.29.0-c61a21a-amd64         55e5cd993a40   2 minutes ago        69.4MB
agones-images/agones-sdk                1.29.0-c61a21a-linux-amd64   8a0612f7cfe4   3 minutes ago        65.3MB
agones-build                            f1d1057367                   cd6bfe544e90   4 weeks ago          3.57GB
agones-build-sdk-node                   2a5f9e6f1a                   6967d5d17df4   7 weeks ago          1.83GB
agones-build-sdk-base                   4e48d2360c                   5348986fd7ce   7 weeks ago          1.59GB
debian                                  bookworm                     0ac8b312578d   5 months ago         119MB
```

Eh - actually looking at it. I don't mind either 🤷🏻

gongmax: I had the impression that for the maintainers, we'd like to provide a valid default registry for the development with overriding the environment viable. Otherwise, there was no reason to have this [`release_registry`](https://github.com/googleforgames/agones/blob/main/build/Makefile#L30) defined for the development process at first place, i.e. we could separate it from the registry which is used for the real release (`make do-release`) and make it default to local at the beginning. It has nothing to do with the limitation of GCR that we cannot split out registries.

But I agreed with you, seems like even for the maintainers, we should prevent them to use `agones-images` project for development purpose. And make it a hard requirement to set the `REGISTRY` environment viable for local development that follows the guidance is reasonable.

gongmax: I removed `release_registry` from the `Makefile` and add instructions to indicates it's required to set this environment viable to follow the guidance to build, push, and install Agones for dev purpose.

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -226,6 +226,7 @@ The following tables lists the configurable parameters of the Agones chart and t
 
 | Parameter                             | Description                                                                                                                                                                                                                           | Default |
 |---------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| `agones.image.registry`                                  | Global image registry for all the Agones systrem images                                                                                                                                                                                   | `us-docker.pkg.dev/agones-images/release`
```
roberthbailey, line 229: s/systrem/system

roberthbailey (approved):

markmandel:
I'll take an extra look at this today - primarily I want to double check the release makefile commands and checklist 👍🏻

build/includes/google-cloud.mk
```diff
@@ -83,9 +83,9 @@ gcloud-auth-cluster: $(ensure-build-image)
 	docker run --rm $(common_mounts) $(build_tag) gcloud container clusters get-credentials $(GCP_CLUSTER_NAME) --zone  $(GCP_CLUSTER_LOCATION)
 
 # authenticate our docker configuration so that you can do a docker push directly
-# to the gcr.io repository
+# to the Google Artifact Registry repository
 gcloud-auth-docker: $(ensure-build-image)
-	docker run --rm $(common_mounts) $(build_tag) gcloud auth print-access-token | docker login -u oauth2accesstoken --password-stdin https://gcr.io
+	docker run --rm $(common_mounts) $(build_tag) gcloud auth print-access-token | docker login -u oauth2accesstoken --password-stdin https://us-docker.pkg.dev
```
markmandel, line 88: Thank you! I forgot that kept having to do this manually!

site/cloudbuild.yaml
```diff
@@ -29,6 +29,8 @@ steps:
   args: ['build', '-f', 'Dockerfile.build', '-t', 'make-docker', '.'] # we need docker and make to run everything.
 - name: "make-docker"
   dir: "build"
+  env:
+    - "REGISTRY=${_REGISTRY}"
```
markmandel, line 33: 🤔 what does this do? I don't _think_ this is required, but I could be wrong?

markmandel (commented):
Looks good - but would love confirmation on the change to site/cloudbuild.yaml before merging.

site/cloudbuild.yaml
markmandel: So this all comes from here:
https://github.com/googleforgames/agones/blob/176ffb21af4d621d8afd310b33e3e10b627d7bc1/build/includes/build-image.mk#L51

So pull and push build image targets are targets specifically for the CI system (local developers don't use them), so for those, I don't know if you want to set them to a default that will work on CI?

Also, if you have `_REGISTRY` you will need to specify it in the substitutions block of this cloudbuild.yaml

I'm curious if you tested running this cloud-build yaml against your dev project to see if it worked - because I would have expected it would have failed because of the missing `_REGISTRY` definition.

gongmax: Right, I understood they come from the `build-image.mk`. I'm just not sure where the [`REGISTRY`](https://github.com/googleforgames/agones/blob/176ffb21af4d621d8afd310b33e3e10b627d7bc1/build/includes/build-image.mk#L22) come from in this file. So if we do not specify the `REGISTRY` env viable in the `site/cloud-build.yaml`, will it try to pull the image from `gcr.io`?

You're right the `_REGISTRY` definition is missed. I only noticed it when I saw your last comments.

markmandel: OIC OIC! So the `REGISTRY` variable in the Makefile could be set in the parent Makefile or any includes that came before it.

So in the `main` version of the code, it come from here: https://github.com/googleforgames/agones/blob/176ffb21af4d621d8afd310b33e3e10b627d7bc1/build/Makefile#L40

>  So if we do not specify the REGISTRY env viable in the site/cloud-build.yaml, will it try to pull the image from gcr.io?

So in your code, the `REGISTRY` variable is removed from the parent `Makefile`, so I'm not sure where it would be getting it from -- I actually expect that it would fail silently, since there wouldn't be a registry attached.

https://github.com/googleforgames/agones/blob/176ffb21af4d621d8afd310b33e3e10b627d7bc1/build/includes/build-image.mk#L70-L72

So TL;DR - I would suggest either hardcode where it's pulled from somewhere in that Makefile, or create a substitution section like we have in the main cloudbuild.yaml and pull it from there.

gongmax: Thanks for the details! This makes sense now.

gongmax: Added the substitution in `site/cloudbuild.yaml`. Confirmed the cloud build step is looking at the `us-docker.pkg.dev/${PROJECT_ID}/ci`, and the corresponding `agones-build` image exists in `us-docker.pkg.dev/agones-images/ci`.

markmandel: Looks right! Okay, let's approve and hopefully nothing breaks 😀 

But looks good to me!

markmandel (approved):

PR #2885, opened 2023-01-04
Set seccompProfile of `Unconfined` on Autopilot unless overidden by user

zmerlynn:
GKE Autopilot defaults to a seccomp profile of `RuntimeDefault`, meaning the container uses the default seccomp profile of the container runtime. Upstream Kubernetes itself is marching towards `RuntimeDefault` as a default as well ([link](https://kubernetes.io/docs/tutorials/security/seccomp/#enable-the-use-of-runtimedefault-as-the-default-seccomp-profile-for-all-workloads)).

The problem is that seccomp currently incurs performance penalties on some workloads, in some cases pretty severe, e.g.: https://github.com/moby/moby/issues/41389. To future-proof Agones until seccomp performance improves, this PR sets a `seccompProfile` which disables seccomp on Autopilot alone, the equivalent of:

```
spec:
  securityContext:
    seccompProfile:
      type: Unconfined
```

We skip the defaulting if the user specifies a Pod `seccompProfile`. The user can also specify a container-level `seccompProfile` which overrides the Pod-level profile (either our default or their explicit config).

### DIFF
--- a/pkg/cloudproduct/gke/gke.go
+++ b/pkg/cloudproduct/gke/gke.go
@@ -131,7 +131,22 @@ func (*gkeAutopilot) ValidateGameServerSpec(gss *agonesv1.GameServerSpec) []meta
 }
 
 func (*gkeAutopilot) MutateGameServerPodSpec(gss *agonesv1.GameServerSpec, podSpec *corev1.PodSpec) error {
+	podSpecSeccompUnconfined(podSpec)
 	return nil
+}
+
+// podSpecSeccompUnconfined sets to seccomp profile to `Unconfined` to avoid serious performance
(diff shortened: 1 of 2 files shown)


markmandel (commented):
Two thoughts:

1. Should this be an Autopilot only default (for now?)
2. If it is global, we definitely need to document this somewhere, since at some point it will be a deviation from standard Kubernetes.

Also we have to be comfortable with this being our default value _forever_ across all Kubernetes installations on all Cloud Providers to set this. This feels a little scary to me because of that.

zmerlynn:
@markmandel I went ahead and just made this Autopilot only for now, since that's where the impact is. We can figure out the rest later. PTAL

markmandel (approved):
SGTM!

Issue #2797, opened 2022-11-09
Design: Make it possible to evict `agones-controller` by separating into controller/extensions


zmerlynn:
## Overview

The `agones-controller` binary currently hosts a collection of k8s related things:
* control loops for `GameServer`, `GameServerSet`, etc.
* mutating and validating webhooks for the same
* the `v1.allocation.agones.dev` extension API

Currently, `agones-controller` is run as a `Deployment` with `replicas=1`, and (unless configured otherwise) sets `safe-to-evict=false` to prevent arbitrary eviction. In this design, we propose:
* splitting `agones-controller` into `agones-controller` and `agones-extensions` (feel free to bikeshed a better name)
  * `agones-controller` would be a pure [controller](https://kubernetes.io/docs/concepts/architecture/controller/) and consist _only_ of the control loops (think: `agones-controller` would have a Kubernetes client, but would not be an endpoint for any services except a local healthcheck)
  * `agones-extensions` would be an HA service that handles webhooks and extension APIs (think: everything in [extensions.yaml](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/extensions.yaml))

Splitting `agones-controller` then allows us to:
* Scale the (new) `agones-extensions` `Deployment` arbitrarily on larger clusters (in case webhook / extension API traffic needs horizontal scaling). We should start with `replicas=2` and a `PodDisruptionBudget`, though, to allow for eviction during upgrades/autoscaling.
* Remove `safe-to-evict=false` from `agones-controller`, and _either_ just let it be, if we think pod restart time after eviction is fast enough, or configure it to run `replicas=2` and add leader-election, which is a common controller pattern.

### Phase 1: Split `agones-extensions` services out of `agones-controller` <a href="#user-content-phase1" id="phase1">#</a>

* Create a new `agones-extensions` binary and image with just the webhooks and extension APIs from `agones-controller`. It would not be deployed until we ....
* Add a new `SplitControllerAndExtensions` [feature flag](https://agones.dev/site/docs/guides/feature-stages/) to [`values.yaml`](https://github.com/googleforgames/agones/blob/main/install/helm/agones/values.yaml)
  * In `agones-controller`, `SplitControllerAndExtensions` would disable the pieces that `agones-extensions` serves.
[...]

PR #2891, opened 2023-01-09
Trim down `agones-extensions` and add flag to `agones-controller`

chiayi:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix

This part of phase 1 for [HA Agones]. Trims down some of the unnecessary parts of extensions as well as adding a `enableSplitControllerFlag` to controller. Default of flag is false. 

Since extensions is the webhooks, I trimmed things down based on the parts that were used when creating NewController, there may be parts within extensions that I could also be removing.


cmd/controller/main.go
```diff
@@ -80,6 +80,7 @@ const (
 	kubeconfigFlag               = "kubeconfig"
 	allocationBatchWaitTime      = "allocation-batch-wait-time"
 	defaultResync                = 30 * time.Second
+	enableSplitControllerFlag    = "enable-split-controller"
```
zmerlynn, line 83: This doesn't need to be a flag - you have access to the feature flag via `runtime.FeatureEnabled`

cmd/controller/main.go
```diff
@@ -213,14 +214,14 @@ func main() {
 		ctlConf.MinPort, ctlConf.MaxPort, ctlConf.SidecarImage, ctlConf.AlwaysPullSidecar,
 		ctlConf.SidecarCPURequest, ctlConf.SidecarCPULimit,
 		ctlConf.SidecarMemoryRequest, ctlConf.SidecarMemoryLimit, ctlConf.SdkServiceAccount,
-		kubeClient, kubeInformerFactory, extClient, agonesClient, agonesInformerFactory)
+		kubeClient, kubeInformerFactory, extClient, agonesClient, agonesInformerFactory, ctlConf.SplitController)
```
zmerlynn, line 217: Rather than burying the if (!split) { NewExtensions() } in the NewController call for each of these, I think it's clearer if you instead just lift the NewExtensions() call up to here (main.go).

I personally would just keep the block of NewController calls here, then around like 223 after all the controllers are initialized, have a block like:

```
if (!<feature flag is enabled>) {
  foo.NewExtensions()
  ...
}
```

Later when we go fold the feature flag to "true", we'd just delete that whole block, which makes that diff less messy.

zmerlynn (changes requested):

zmerlynn:
Great work, BTW!

cmd/controller/main.go
chiayi: Changed to a `runtime.FeatureEnabled(runtime.FeatureSplitControllerAndExtensions)`

cmd/controller/main.go
chiayi: Thanks for pointing this out. I should've realized this since it is similar what it's doing in extensions. 😅

cmd/controller/main.go
```diff
@@ -398,6 +405,7 @@ type config struct {
 	LogLevel                string
 	LogSizeLimitMB          int
 	AllocationBatchWaitTime time.Duration
+	SplitController         bool
```
zmerlynn, line 408: leftover?

cmd/extensions/main.go
```diff
@@ -138,19 +120,11 @@ func main() {
 		logger.WithError(err).Fatal("Could not create in cluster config")
 	}
 
-	clientConf.QPS = float32(ctlConf.APIServerSustainedQPS)
```
zmerlynn, line 141: QPS/Burst seem relevant for anything that carries a kubeClient .. but I'm wondering if `cloudproduct.Initialize` is the only thing using the kubeClient? If so, you could make it more obvious by separating it into its own block or function, i.e.:

{
  kubeClient, err = ...
  cloudProduct.Initialize()
}

then, especially if we don't need the informers, it makes it clear to the reader that we're not really using the client at all.

cmd/extensions/main.go
```diff
@@ -163,7 +137,6 @@ func main() {
 	// https server and the items that share the Mux for routing
 	httpsServer := https.NewServer(ctlConf.CertFile, ctlConf.KeyFile)
 	wh := webhooks.NewWebHook(httpsServer.Mux)
-	api := apiserver.NewAPIServer(httpsServer.Mux)
 
 	agonesInformerFactory := externalversions.NewSharedInformerFactory(agonesClient, defaultResync)
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, defaultResync)
```
zmerlynn, line 142: Is either informer in use anymore?

cmd/extensions/main.go
```diff
@@ -200,30 +173,14 @@ func main() {
 	// which is a requirements of Stackdriver, otherwise most of time series would be invalid for Stackdriver
 	metrics.SetReportingPeriod(ctlConf.PrometheusMetrics, ctlConf.Stackdriver)
 
-	// Add metrics controller only if we configure one of metrics exporters
-	if ctlConf.PrometheusMetrics || ctlConf.Stackdriver {
-		rs = append(rs, metrics.NewController(kubeClient, agonesClient, kubeInformerFactory, agonesInformerFactory))
```
zmerlynn, line 205: I'm not super familiar with metrics but it seems odd that we have none on the extensions side? Might be true.

Actually, I think this is an artifact of missing the allocation server, below.

cmd/extensions/main.go
```diff
@@ -200,30 +173,14 @@ func main() {
...
-	}
-
 	server.Handle("/", health)
 
-	gsCounter := gameservers.NewPerNodeCounter(kubeInformerFactory, agonesInformerFactory)
```
zmerlynn, line 210: You'll probably need this as well, when you add back the allocations server.

zmerlynn (changes requested):
Getting closer but still a couple of structural things.

cmd/controller/main.go
chiayi: Oohh yup 😅

cmd/extensions/main.go
zmerlynn: I thought I fixed this comment after I left the comment about the allocation service, but ignore me about `kubeClient` - I'm fairly sure you need it for that. (Which means definitely keep QPS/Burst.)

cmd/extensions/main.go
zmerlynn: (It will be if you keep the allocation service.)

cmd/extensions/main.go
chiayi: Added the QPSs back.

cmd/extensions/main.go
```diff
@@ -200,30 +173,14 @@ func main() {
...
 
-	gsController := gameservers.NewController(wh, health,
-		ctlConf.MinPort, ctlConf.MaxPort, ctlConf.SidecarImage, ctlConf.AlwaysPullSidecar,
-		ctlConf.SidecarCPURequest, ctlConf.SidecarCPULimit,
-		ctlConf.SidecarMemoryRequest, ctlConf.SidecarMemoryLimit, ctlConf.SdkServiceAccount,
-		kubeClient, kubeInformerFactory, extClient, agonesClient, agonesInformerFactory)
-	gsSetController := gameserversets.NewController(wh, health, gsCounter,
-		kubeClient, extClient, agonesClient, agonesInformerFactory)
-	fleetController := fleets.NewController(wh, health, kubeClient, extClient, agonesClient, agonesInformerFactory)
-	gasController := gameserverallocations.NewController(api, health, gsCounter, kubeClient, kubeInformerFactory,
```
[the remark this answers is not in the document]
chiayi: gameserverallocations is added back and is now `gameserverallocations.NewExtensions`.

cmd/extensions/main.go
chiayi: added back.

cmd/extensions/main.go
chiayi: I think I'm still confused on how the metrics controller is used, and specifically how it's used by the game allocation server. I don't see it being explicitly passed to gamerserverallocations and I can't seem to connect the allocation events to the metrics.

chiayi: Removed everything thats related to metrics controller.

zmerlynn (commented):
LGTM, but I think @markmandel or @roberthbailey should look as well.

zmerlynn:
Re-reviewed changes and still LGTM. @markmandel consider prioritizing after feature freeze, thanks!

zmerlynn (approved):
LGTM

cmd/controller/main.go
```diff
@@ -217,13 +217,22 @@ func main() {
 	gsSetController := gameserversets.NewController(wh, health, gsCounter,
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
 	fleetController := fleets.NewController(wh, health, kubeClient, extClient, agonesClient, agonesInformerFactory)
-	gasController := gameserverallocations.NewController(api, health, gsCounter, kubeClient, kubeInformerFactory,
-		agonesClient, agonesInformerFactory, 10*time.Second, 30*time.Second, ctlConf.AllocationBatchWaitTime)
 	fasController := fleetautoscalers.NewController(wh, health,
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
 
+	if !runtime.FeatureEnabled(runtime.FeatureSplitControllerAndExtensions) {
+		gameservers.NewExtensions(wh)
```
roberthbailey, line 224: I'd originally written
> Maybe this should happen later (since it requires more untangling) but I don't understand why `wh` needs to be passed to all of the control loops. It seems like they shouldn't require (or know anything about) the webhooks, and the whole creation of the webhooks and extension api should be hidden behind this feature gate. Then with the feature gate the controller wouldn't run anything related to the extensions at all. 

But after reading the rest of the PR and seeing how you refactored the controllers to not use the webhooks any longer, I think it should be part of this PR. You are moving the webhook registration out of the controller constructors so the constructors should no longer need to be passed in the variable. And once that change is made, then it doesn't make sense to even construct those variables outside of the feature gate block because they would not be used.

pkg/fleetautoscalers/controller.go
```diff
@@ -152,6 +148,18 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) {
+	extensionLogger = runtime.NewLoggerWithSource("FleetAutscalerExtensions")
```
roberthbailey, line 154: Creating a global variable to share state between this function and the validation function seems less than ideal. 

What about creating a struct that just holds the logger and hanging the mutationHandler function off of it (similarly to how it was previously hanging off of the controller struct)?

(this same comment applies to the other controller files as well)

pkg/fleets/controller.go
```diff
@@ -233,6 +241,18 @@ func (c *Controller) loggerForFleet(f *agonesv1.Fleet) *logrus.Entry {
 	return c.loggerForFleetKey(fleetName).WithField("fleet", f)
 }
 
+func loggerForFleetKey(key string) *logrus.Entry {
```
roberthbailey, line 244: These functions are duplicates of ones that already exist. Instead of copying them, how about extracting the existing functions to live outside of the controller struct and taking the logger as an argument (since that is the only field in the controller that they need)? You may need to change a lot of call sites, but it will prevent code duplication here and also make the function have a clear and small surface and purpose.

pkg/gameserverallocations/controller.go
```diff
@@ -51,8 +51,8 @@ type Controller struct {
 	allocator  *Allocator
 }
 
-// NewController returns a controller for a GameServerAllocation
-func NewController(apiServer *apiserver.APIServer,
+// NewExtensions returns a controller for a GameServerAllocation
+func NewExtensions(apiServer *apiserver.APIServer,
```
roberthbailey, line 55: Why is this being renamed? It is no longer a controller? The return type is still `*Controller`.

zmerlynn: I asked for the rename because it follows the pattern of the other "NewExtensions" and makes it clear which side of the divide. (I missed the return type from review fatigue.)

cmd/controller/main.go
```diff
@@ -217,13 +217,22 @@ func main() {
 	gsSetController := gameserversets.NewController(wh, health, gsCounter,
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
 	fleetController := fleets.NewController(wh, health, kubeClient, extClient, agonesClient, agonesInformerFactory)
-	gasController := gameserverallocations.NewController(api, health, gsCounter, kubeClient, kubeInformerFactory,
-		agonesClient, agonesInformerFactory, 10*time.Second, 30*time.Second, ctlConf.AllocationBatchWaitTime)
 	fasController := fleetautoscalers.NewController(wh, health,
 		kubeClient, extClient, agonesClient, agonesInformerFactory)
 
+	if !runtime.FeatureEnabled(runtime.FeatureSplitControllerAndExtensions) {
```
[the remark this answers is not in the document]
chiayi: Done! Re-ordered the lines so that `rs = append(...)` is here first.

cmd/extensions/main.go
```diff
@@ -403,16 +264,6 @@ type config struct {
 // validate ensures the ctlConfig data is valid.
 func (c *config) validate() []error {
```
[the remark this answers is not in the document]
chiayi: Removed.

cmd/extensions/main.go
```diff
@@ -34,52 +33,36 @@ import (
...
-	numWorkersFlag               = "num-workers"
-	apiServerSustainedQPSFlag    = "api-server-qps"
-	apiServerBurstQPSFlag        = "api-server-qps-burst"
-	logDirFlag                   = "log-dir"
-	logLevelFlag                 = "log-level"
-	logSizeLimitMBFlag           = "log-size-limit-mb"
-	kubeconfigFlag               = "kubeconfig"
-	allocationBatchWaitTime      = "allocation-batch-wait-time"
-	defaultResync                = 30 * time.Second
+	stackdriverLabels         = "stackdriver-labels"
```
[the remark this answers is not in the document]
chiayi: Removed as well.

cmd/extensions/main_test.go
```diff
@@ -13,50 +13,3 @@
 // limitations under the License.
 
 package main
```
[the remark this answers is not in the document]
chiayi: The file is now gone.

cmd/extensions/main.go
```diff
@@ -323,70 +229,25 @@ func parseEnvFlags() config {
...
 		APIServerBurstQPS:       int(viper.GetInt32(apiServerBurstQPSFlag)),
 		LogDir:                  viper.GetString(logDirFlag),
 		LogLevel:                viper.GetString(logLevelFlag),
 		LogSizeLimitMB:          int(viper.GetInt32(logSizeLimitMBFlag)),
-		StackdriverLabels:       viper.GetString(stackdriverLabels),
 		AllocationBatchWaitTime: viper.GetDuration(allocationBatchWaitTime),
 	}
 }
 
 // config stores all required configuration to create a game server controller.
```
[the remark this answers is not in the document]
chiayi: I changed it to `... game server extensions`.

cmd/extensions/main.go
```diff
@@ -25,7 +25,6 @@ import (
 	"time"
```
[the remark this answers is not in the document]
chiayi: Updated to use the `Extensions` keyword instead.

cmd/extensions/main.go
```diff
@@ -34,52 +33,36 @@ import (
...
+	logLevelFlag              = "log-level"
+	logSizeLimitMBFlag        = "log-size-limit-mb"
+	allocationBatchWaitTime   = "allocation-batch-wait-time"
+	kubeconfigFlag            = "kubeconfig"
+	defaultResync             = 30 * time.Second
+	apiServerSustainedQPSFlag = "api-server-qps"
+	apiServerBurstQPSFlag     = "api-server-qps-burst"
 )
 
 var (
```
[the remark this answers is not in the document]
chiayi: Updated. Please let me know if there's another way to do it.

cmd/extensions/main.go
```diff
@@ -146,11 +129,6 @@ func main() {
...
 	}
 
-	extClient, err := extclientset.NewForConfig(clientConf)
-	if err != nil {
-		logger.WithError(err).Fatal("Could not create the api extension clientset")
-	}
-
 	agonesClient, err := versioned.NewForConfig(clientConf)
 	if err != nil {
 		logger.WithError(err).Fatal("Could not create the agones api clientset")
```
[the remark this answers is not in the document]
chiayi: Done! I'll keep this in mind for the future.

cmd/extensions/main.go
```diff
@@ -170,60 +148,22 @@ func main() {
...
-		kubeClient, extClient, agonesClient, agonesInformerFactory)
 
-	rs = append(rs,
-		httpsServer, gsCounter, gsController, gsSetController, fleetController, fasController, gasController, server)
+	gameservers.NewExtensions(wh)
+	gameserversets.NewExtensions(wh)
+	fleets.NewExtensions(wh)
+	fleetautoscalers.NewExtensions(wh)
+
+	rs = append(rs, httpsServer, gasExtensions, server)
```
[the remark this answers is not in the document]
chiayi: Changed to the second suggestion. 👍

cmd/controller/main.go
chiayi: You're correct, originally I had the feature check within the controllers which used `wh`. I removed the `wh` as a param for controllers.

pkg/fleetautoscalers/controller.go
chiayi: There is now a `Extensions` struct that contains the logger. The `Extensions` struct is now the receiver for the handler functions as well.

pkg/fleets/controller.go
chiayi: Removed the duplicates and now takes in the logger as a param. Thanks!

pkg/gameserverallocations/controller.go
chiayi: Hmm, I see how it can be confusing without changing the struct name. We could either change the struct as well but not sure if that would cause any confusion on other parts or change it back?

cmd/extensions/main.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Controller for gameservers
+// Extensions for gameservers
```
markmandel, line 15:
```suggestion
// Extensions for the Agones system
```
Nit: Original comment was probably debt from when I initially wrote the controller.

cmd/extensions/main.go
```diff
@@ -104,7 +86,7 @@ func setupLogging(logDir string, logSizeLimitMB int) {
 	)
 }
 
-// main starts the operator for the gameserver CRD
+// main initializes the extensions service for the game servers
```
markmandel, line 89:
```suggestion
// main initializes the extensions services for Agones
```

Nit: Same as above.

pkg/fleetautoscalers/controller.go
```diff
@@ -152,6 +150,22 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{
+		extensionLogger: runtime.NewLoggerWithSource("FleetAutscalerExtensions"),
+	}
```
markmandel, line 158:
```suggestion
	ext := &Extensions{}
  ext.baseLogger = runtime.NewLoggerWithType(ext)
```

To be consistent with how controllers are setup. i.e.

https://github.com/googleforgames/agones/blob/efb2caa77645f1c5d452ccc5e590c13e250bc08c/pkg/gameservers/controller.go#L143

pkg/fleetautoscalers/controller.go
```diff
@@ -65,6 +65,11 @@ type fasThread struct {
 	cancel     context.CancelFunc
 }
 
+// Extensions struct contains what is needed to bind webhook handlers
+type Extensions struct {
+	extensionLogger *logrus.Entry
```
markmandel, line 70:
```suggestion
	baseLogger *logrus.Entry
```

To be consistent.

pkg/fleets/controller.go
```diff
@@ -125,12 +125,26 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{
+		extensionLogger: runtime.NewLoggerWithSource("FleetsExtensions"),
```
markmandel, line 132: Same as above, use `NewLoggerWithType`, with `ext.baseLogger`.

pkg/gameservers/controller.go
```diff
@@ -195,6 +196,19 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{
+		extensionLogger: runtime.NewLoggerWithSource("GameserversExtensions"),
```
markmandel, line 203: Same as above, use `NewLoggerWithType`, with `ext.baseLogger`.

pkg/gameserversets/controller.go
```diff
@@ -150,6 +151,19 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{
+		extensionLogger: runtime.NewLoggerWithSource("GameserversetExtensions"),
```
markmandel, line 158: Same as above, use `NewLoggerWithType`, with `ext.baseLogger`.

pkg/gameserverallocations/controller.go
markmandel: 🤔 I'd probably be inclined to move the whole type to being *Extension, since that would align with the responsibilities of all the other Extension types.

cmd/extensions/main.go
chiayi: Done.

cmd/extensions/main.go
chiayi: Done.

pkg/fleetautoscalers/controller.go
chiayi: Done.

pkg/gameserversets/controller.go
chiayi: Done.

chiayi:
I've made changes to `pkg/gameserverallocations/controller.go` and renamed *Controller to *Extensions.

cmd/extensions/main.go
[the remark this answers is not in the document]
roberthbailey: A couple more nits on these lines. 

- One the main func comment,  it should end with "service for Agones"
- On the error message for logging, the code is checking the log level and making sure it is a valid string, so the error should be something like "Unable to parse log level, using the default level (Info)" or something like that.

pkg/fleetautoscalers/controller.go
```diff
@@ -152,6 +150,22 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{}
```
roberthbailey, line 156: nit: This and the next line can be combined:

```
ext := &Extensions{baseLogger: runtime.NewLoggerWithSource("FleetAutscalerExtensions")}
```

pkg/fleetautoscalers/controller_test.go
```diff
@@ -1241,12 +1245,19 @@ func defaultWebhookFixtures() (*autoscalingv1.FleetAutoscaler, *agonesv1.Fleet)
...
-	c := NewController(wh, healthcheck.NewHandler(), m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
+	c := NewController(healthcheck.NewHandler(), m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
 	c.recorder = m.FakeRecorder
 	return c, m
 }
 
+// newFakeExtensions returns a extensions struct
+func newFakeExtensions() *Extensions {
+	wh := webhooks.NewWebHook(http.NewServeMux())
+	ext := NewExtensions(wh)
```
roberthbailey, line 1256: nit: Can just do `return NewExtensions(wh)` here.

pkg/fleets/controller.go
```diff
@@ -125,12 +125,26 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{}
```
roberthbailey, line 131: same here

pkg/fleets/controller_test.go
```diff
@@ -1402,12 +1400,19 @@ func TestControllerRollingUpdateDeployment(t *testing.T) {
...
-	c := NewController(wh, healthcheck.NewHandler(), m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
+	c := NewController(healthcheck.NewHandler(), m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
 	c.recorder = m.FakeRecorder
 	return c, m
 }
 
+// newFakeExtensions returns a fake extensions struct
+func newFakeExtensions() *Extensions {
+	wh := webhooks.NewWebHook(http.NewServeMux())
+	ext := NewExtensions(wh)
```
roberthbailey, line 1411: nit: combine with the following line

pkg/gameservers/controller.go
```diff
@@ -195,6 +196,19 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{}
```
roberthbailey, line 202: same here

pkg/gameservers/controller_test.go
```diff
@@ -1957,6 +1956,14 @@ func newFakeController() (*Controller, agtesting.Mocks) {
 	return c, m
 }
 
+// newFakeExtensions return a fake extensions struct
+func newFakeExtensions() *Extensions {
+	wh := webhooks.NewWebHook(http.NewServeMux())
+	ext := NewExtensions(wh)
```
roberthbailey, line 1962: nit: combine with the following line.

pkg/gameserversets/controller.go
```diff
@@ -150,6 +151,19 @@ func NewController(
 	return c
 }
 
+// NewExtensions binds the handlers to the webhook outside the initialization of the controller
+// initializes a new logger for extensions.
+func NewExtensions(wh *webhooks.WebHook) *Extensions {
+	ext := &Extensions{}
```
roberthbailey, line 157: same here

pkg/gameserversets/controller_test.go
```diff
@@ -995,9 +996,16 @@ func createGameServers(gsSet *agonesv1.GameServerSet, size int) []agonesv1.GameS
...
-	c := NewController(wh, healthcheck.NewHandler(), counter, m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
+	c := NewController(healthcheck.NewHandler(), counter, m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
 	c.recorder = m.FakeRecorder
 	return c, m
 }
+
+// newFakeExtensions returns an extensions struct
+func newFakeExtensions() *Extensions {
+	wh := webhooks.NewWebHook(http.NewServeMux())
+	ext := NewExtensions(wh)
```
roberthbailey, line 1008: nit: combine with the following line

roberthbailey (commented):
This is looking really good! I left a few comments for minor tweaks, but I think it's just about ready to be merged.

pkg/fleetautoscalers/controller.go
chiayi: Discussed offline, `NewExtensions` will match how `NewController` initializes the loggers.

pkg/fleetautoscalers/controller_test.go
chiayi: Combined.

pkg/fleets/controller_test.go
chiayi: Combined.

chiayi: Combined in other tests as well.

roberthbailey (approved):

pkg/fleetautoscalers/controller.go
```diff
@@ -152,6 +150,22 @@ func NewController(
...
+
+	ext.baseLogger = runtime.NewLoggerWithSource("FleetAutscalerExtensions")
```
markmandel, line 158: This should use NewLoggerWithType

markmandel (commented):
Just reviewing on phone - but please switch to using the BaseLoggerWithType to create the log entry 👍🏻

pkg/fleetautoscalers/controller.go
chiayi: I changed this one as well as the other `NewExtensions`.

zmerlynn (approved):
LGTM!

pkg/gameserversets/controller_test.go
```diff
@@ -995,9 +996,15 @@ func createGameServers(gsSet *agonesv1.GameServerSet, size int) []agonesv1.GameS
...
 	counter := gameservers.NewPerNodeCounter(m.KubeInformerFactory, m.AgonesInformerFactory)
-	c := NewController(wh, healthcheck.NewHandler(), counter, m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
+	c := NewController(healthcheck.NewHandler(), counter, m.KubeClient, m.ExtClient, m.AgonesClient, m.AgonesInformerFactory)
 	c.recorder = m.FakeRecorder
 	return c, m
 }
+
+// newFakeExtensions returns an extensions struct
+func newFakeExtensions() *Extensions {
+	ext := NewExtensions(webhooks.NewWebHook(http.NewServeMux()))
```
roberthbailey, line 1007: One other thing I noticed - in each of these files, this can be simplified to a single line (removing the ext var and just doing `return NewExtensions(webhooks.NewWebHook(http.NewServeMux()))`.

roberthbailey (approved):

markmandel (approved):

PR #2892, opened 2023-01-10
Validate the .scheduling field in the fleet and gameserverset

zmerlynn:
Adds a cloud product hook specific to the scheduling field, for use in the fleet / gameserverset APIs. This might seem narrowly scoped, but this field is the only one that propagates down outside the template.

Towards #2777

### DIFF
--- a/pkg/apis/agones/v1/fleet.go
+++ b/pkg/apis/agones/v1/fleet.go
@@ -190,6 +190,9 @@ func (f *Fleet) Validate() ([]metav1.StatusCause, bool) {
 	if len(gsCauses) > 0 {
 		causes = append(causes, gsCauses...)
 	}
+	if productCauses := apiHooks.ValidateScheduling(f.Spec.Scheduling); len(productCauses) > 0 {
+		causes = append(causes, productCauses...)
+	}
 	objMetaCauses := validateObjectMeta(&f.Spec.Template.ObjectMeta)
 	if len(objMetaCauses) > 0 {
 		causes = append(causes, objMetaCauses...)
--- a/test/e2e/gameserverallocation_test.go
+++ b/test/e2e/gameserverallocation_test.go
@@ -51,6 +51,11 @@ func TestCreateFleetAndGameServerAllocate(t *testing.T) {
 			fleet := defaultFleet(framework.Namespace)
 			fleet.Spec.Scheduling = strategy
 			flt, err := fleets.Create(ctx, fleet, metav1.CreateOptions{})
+			if strategy != apis.Packed && framework.CloudProduct == "gke-autopilot" {
+				// test that Autopilot rejects anything but Packed and skip the rest of the test
+				assert.ErrorContains(t, err, "configuration is invalid")
+				return
+			}
 			if assert.NoError(t, err) {
 				defer fleets.Delete(ctx, flt.ObjectMeta.Name, metav1.DeleteOptions{}) // nolint:errcheck
 			}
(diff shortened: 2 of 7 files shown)


pkg/apis/agones/v1/fleet.go
```diff
@@ -190,6 +190,9 @@ func (f *Fleet) Validate() ([]metav1.StatusCause, bool) {
 	if len(gsCauses) > 0 {
 		causes = append(causes, gsCauses...)
 	}
+	if productCauses := apiHooks.ValidateScheduling(f.Spec.Scheduling); len(productCauses) > 0 {
```
roberthbailey, line 193: Should these have new unit tests? Or do unit tests only make sense in the (existing) cloudproduct package where the behavior can return an error?

roberthbailey:
I can't pull the branch forward as it looks like there is a conflict you will need to resolve manually.

pkg/apis/agones/v1/fleet.go
zmerlynn: My feeling was that the logic here was small enough that unit testing in the cloud product was sufficient. It was also perhaps expediency for the fact that there is not yet a fake cloud product for unit tests. If you'd like, I can address that here.

zmerlynn:
> I can't pull the branch forward as it looks like there is a conflict you will need to resolve manually.

Resolved, PTAL!

pkg/apis/agones/v1/fleet.go
markmandel: +1 on having a unit test. Otherwise, this looks good to me 👍🏻

roberthbailey: I'm ok with unit tests in the provider package as long as we are still testing that this check is working properly in a unit test.

zmerlynn: The gke provider code path is exercised in `TestValidateGameServer`, which calls it through `ValidateGameServerSpec` -> `ValidateScheduling`. It's slightly different than the path at the cursor here, _but_ it's definitely exercised.

I looked at exercising `apiHooks` and I will say, it's a convenient structure and I still think it's better than the way I was previously handling it, but explicitly overriding `apiHooks` for tests is difficult: Unlike the feature flags, the apiHooks are called _broadly_ due to their presence in basic structure validation. I headed down a part of adding a mutex just like `runtime` handles its singleton, but it ended being fairly invasive and I'm still fighting. I'm looking at it and .. don't think it's the right approach.

Rather, I'm wondering if for both feature flags and cloud product hooks if we should instead be passing through some sort of context structure, which is the more Go-endorsed solution. It's equally invasive but it's rhythmic in that for the most part such a context would just flow through most functions, similar to passing `ctx` around.

But, short version: I don't think it's worth testing this exact line, and I think the GKE provider implementation is tested.

markmandel: I wouldn't mind seeing a test that was basically a "do causes get appended if there is a result from `apiHooks.ValidateScheduling`" just to confirm the branch works as expected.

I don't think it needs to be anything more complex than that.

test/e2e/gameserverallocation_test.go
```diff
@@ -44,6 +44,9 @@ func TestCreateFleetAndGameServerAllocate(t *testing.T) {
 	for _, strategy := range fixtures {
 		strategy := strategy
 		t.Run(string(strategy), func(t *testing.T) {
+			if strategy == apis.Distributed {
```
markmandel, line 47: Given it's difficult to stub out the global `apiHooks` - the other option could be than rather than / in addition to skipping these e2e tests, add e2e tests just for Autopilot that would test to see that we get the appropriate error on `apis.Distributed`.

Then the branch above is tested, without having to do the stubbing. WDYT?

zmerlynn: Yup, agreed, will add.

pkg/apis/agones/v1/fleet.go
zmerlynn: We compromised on an e2e test, which I added (by modifying one of the Skip cases).

zmerlynn:
PTAL, thanks!

roberthbailey (approved):

PR #2893, opened 2023-01-10
e2e framework: Allow variable timing based on cloud product

zmerlynn:
Most of the "wait for state" functions use a hardcoded 5m. To allow for autoscaling variance on Autopilot, allow for 10m on Autopilot.

Along the way: Remove New/NewWithRates as they are unused or unnecessary wrappers - in particular, they definitely don't need to be exported.

Towards #2777


gongmax (approved):
A dumb question, I saw there are hardcoded 5 mins in some other tests, eg the tests in `gameserver_test.go`. There is no need to make them adjustable for AP?

PR #2898, opened 2023-01-11
Removes explicit versions of Prometheus and Grafana

igooch:
**What type of PR is this?**

/kind cleanup

Removes explicit version settings for Grafana and Prometheus, so that the most recent versions are pulled, and we do not need to continually update the versions in the documentation.
Updates the Helm version in the Makefile to >3.10 so that it works with the most recent versions of Grafana and Prometheus. 
Adds Error handling suggestions to the Makefile for easier troubleshooting.

N/A

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -110,6 +110,10 @@ Table of Contents
       * [Error: cluster-admin-binding already exists](#error-cluster-admin-binding-already-exists)
       * [Error: releases do not exist](#error-releases-do-not-exist)
       * [I want to use pprof to profile the controller](#i-want-to-use-pprof-to-profile-the-controller)
+      * [Error: Kubernetes cluster unreachable](#error-kubernetes-cluster-unreachable-invalid-configuration-no-configuration-has-been-provided)
+      * [Error: project: required field is not set](#error-project-required-field-is-not-set-error-invalid-value-for-network-project-required-field-is-not-set)
+      * [Error: could not download chart](#error-could-not-download-chart-failed-to-download-httpsagonesdevchartstableagones-1280tgz)
+      * [Invalid argument "/agones-controller:$(VERSION)" for "-t, --tag" flag: invalid reference format](#invalid-argument-agones-controller1290-961d8ae-amd64-for--t---tag-flag-invalid-reference-format)
 
 ## Building on Different Platforms
 
@@ -158,7 +162,7 @@ free to make cup of tea or coffee at this point. ☕️
 The build image is only created the first time one of the make targets is executed, and will only rebuild if the build
 Dockerfile has changed.
 
-Assuming that the tests all pass, let's go ahead and compile the code and build the Docker images that Agones 
+Assuming that the tests all pass, let's go ahead and compile the code and build the Docker images that Agones
 consists of.
 
 To compile the Agones images, run `make build-images`. This is often all you need for Agones development.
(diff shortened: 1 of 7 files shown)


markmandel (commented):
Love it - but shall we merge after stable release?

igooch:
> Love it - but shall we merge after stable release?

Yep!

build/README.md
```diff
@@ -933,3 +939,20 @@ on [http://localhost:6061](http://localhost:6061).
 To view heap metrics, run `make pprof-heap-web`, which will start the web interface with a Heap usage graph.
 on [http://localhost:6062](http://localhost:6062).
 
+### Error: Kubernetes cluster unreachable: invalid configuration: no configuration has been provided
```
gongmax, line 942: `###` should be `####`, same for the following entries.

build/build-image/Dockerfile
```diff
@@ -57,7 +57,7 @@ RUN mkdir -p /go/src/k8s.io && cd /go/src/k8s.io && \
     git clone -b kubernetes-${KUBERNETES_VER} --depth=3 https://github.com/kubernetes/code-generator.git
 
 # install Helm package manager
-ENV HELM_VER 3.5.0
+ENV HELM_VER 3.10.2
```
gongmax, line 60: nit: 3.10.3 is the latest version

build/grafana/dashboard-allocations.yaml
```diff
@@ -170,7 +170,6 @@ data:
           "nullPointMode": "connected",
           "nullText": null,
           "options": {},
-          "pluginVersion": "6.2.4",
```
gongmax, line 173: Just for my learning purpose, why these `pluginVersion` are not needed?

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -96,7 +96,7 @@ The following tables lists the configurable parameters of the Agones chart and t
...
 | `agones.serviceaccount.controller`                       | Service account name for the controller. **Note**: Will be replaced with `agones.serviceaccount.controller.name` in Agones 1.16                                                                                         | `agones-controller`                |
-| `agones.serviceaccount.sdk`                              | Service account name for the sdk. **Note**: Will be replaced with `agones.serviceaccount.sdk.name` in Agones 1.16                                                                                                       | `agones-sdk`                       | 
+| `agones.serviceaccount.sdk`                              | Service account name for the sdk. **Note**: Will be replaced with `agones.serviceaccount.sdk.name` in Agones 1.16                                                                                                       | `agones-sdk`                       |
```
gongmax, line 99: Not related to your PR, but was this field already replaced since Agones is already way pass 1.16. Same for the line 98.

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -272,7 +272,7 @@ In order to use `helm test` command described in this section you need to set `h
 
 Check the Agones installation by running the following command:
 ```bash
-helm test my-release --cleanup
```
gongmax, line 275: For curiosity, was the option `--cleanup` removed from the new helm version so we removed it here?

gongmax (commented):
I'm not sure how the `prometheus-stack` is used (line 745 in build/Makefile) but should we also remove the explicit version (which is pretty outdated) there?

site/content/en/docs/Installation/Install Agones/helm.md
gongmax: Looking into the code, the replacement was already done so I think it should be fine to remove these two lines.

build/README.md
```diff
@@ -110,6 +110,10 @@ Table of Contents
       * [Error: cluster-admin-binding already exists](#error-cluster-admin-binding-already-exists)
       * [Error: releases do not exist](#error-releases-do-not-exist)
       * [I want to use pprof to profile the controller](#i-want-to-use-pprof-to-profile-the-controller)
+      * [Error: Kubernetes cluster unreachable](#error-kubernetes-cluster-unreachable-invalid-configuration-no-configuration-has-been-provided)
```
roberthbailey, line 113: Thanks for adding these. I noticed while checking the file that the headings for these new sections seem to be the wrong size (they use ### instead of #### like the ones above them in the same section). While you are fixing this file, could you adjust the headings for these sections at the bottom of the file?

site/content/en/docs/Installation/Install Agones/helm.md
igooch: Yes, the cleanup is now automatic so they go rid of the flag.

build/grafana/dashboard-allocations.yaml
igooch: It's also a reference to the Grafana version. Leaving it out here so that it will use the most recent stable version instead of a specific version.

gongmax (approved):

PR #2900, opened 2023-01-12
controller/extensions: Add explicit requests/limits for ephemeral storage

zmerlynn:
On Autopilot, ephemeral storage defaults to a request/limit of 1GiB, but on Agones we assume there's no ephemeral storage limit. The default persistent log limit is 10GB.

In this PR, we add a default `resources.{limits,requests}` block to the controller if the user has not already specified `resources` and set it to the persistent log maximum plus a fudge factor of 100MB to allow for the certs and some overhead. By default this comes out to 10100MB, which slides under the Autopilot 10GiB limit by virtue of GiB/MB differences.

Towards #2777

### DIFF
--- a/install/helm/agones/templates/controller.yaml
+++ b/install/helm/agones/templates/controller.yaml
@@ -146,9 +146,14 @@ spec:
           periodSeconds: {{ .Values.agones.controller.healthCheck.periodSeconds }}
           failureThreshold: {{ .Values.agones.controller.healthCheck.failureThreshold }}
           timeoutSeconds: {{ .Values.agones.controller.healthCheck.timeoutSeconds }}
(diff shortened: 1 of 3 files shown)


roberthbailey:
Looking at the generated install.yaml, it looks like the ephemeral storage is nudged up to `10100Mi` which doesn't match the PR description. Did you mean to subtract 100 instead of add 100?

zmerlynn:
> Looking at the generated install.yaml, it looks like the ephemeral storage is nudged up to `10100Mi` which doesn't match the PR description. Did you mean to subtract 100 instead of add 100?

I changed the commit text but forgot to backpropagate it to the PR description - fixed now.

roberthbailey (approved):

PR #2901, opened 2023-01-12
Wait on free ports on GKE Autopilot

zmerlynn:
On Autopilot, since port assignments are random, port conflicts happen. Just accept it and wait on the autoscaler to kick in - the problem will go away.

We do this by introducing a `WaitOnFreePorts()` cloud product hook to indicate whether we should pay attention to the case where the pod is unschedulable due to free ports.

It is very likely this condition occurs organically in any autoscaling scenario, but the fleet controller also hides the condition by rapidly deleting the Unhealthy GameServer.

Along the way, fix a typo I introduced in the e2e framework.

### DIFF
--- a/pkg/gameservers/health.go
+++ b/pkg/gameservers/health.go
@@ -24,6 +24,7 @@ import (
 	getterv1 "agones.dev/agones/pkg/client/clientset/versioned/typed/agones/v1"
 	"agones.dev/agones/pkg/client/informers/externalversions"
 	listerv1 "agones.dev/agones/pkg/client/listers/agones/v1"
+	"agones.dev/agones/pkg/cloudproduct"
 	"agones.dev/agones/pkg/util/logfields"
 	"agones.dev/agones/pkg/util/runtime"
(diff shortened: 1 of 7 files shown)


pkg/gameservers/health.go
```diff
@@ -109,6 +110,10 @@ func (hc *HealthController) isUnhealthy(pod *corev1.Pod) bool {
 // unschedulableWithNoFreePorts checks if the reason the Pod couldn't be scheduled
 // was because there weren't any free ports in the range specified
 func (hc *HealthController) unschedulableWithNoFreePorts(pod *corev1.Pod) bool {
+	// On some cloud products (GKE Autopilot), wait on the Autoscaler to schedule a pod with conflicting ports.
+	if cloudproduct.ControllerHooks().WaitOnFreePorts() {
```
gongmax, line 114: Share we add unit test for this?

zmerlynn: Sure, done. :)

gongmax (approved):

markmandel (approved):

PR #2904, opened 2023-01-13
e2e: Use gotestsum for CI e2e runs, change e2e ARGS handling

zmerlynn:
The main goal of this PR is to add [`gotestsum`](https://pkg.go.dev/gotest.tools/gotestsum#section-readme) as an optional way to run e2es. `gotestsum` wraps `go test -json` and, aside from some fancy formatting options, allows us to re-run flaky tests. As we start adding more permutations to the test matrix, this will become important. (Also important will become actually monitoring flakes, but that's the next step.)

We do this by:
* Installing `gotestsum` in the build image
* Adding a new `E2E_USE_GOTESTSUM` option to the Makefile that defaults to 2x reruns and super-terse output.
* Also added a `GOTESTSUM_TESTNAME_AND_VERBOSE` option meant for CI that shows the status of all of the tests then dumps the full verbose output for diagnostics.

One thing that changes slightly after this PR is the arguments involved in invoking e2es via `make`. Previously, both `ARGS` and `GO_E2E_TEST_ARGS` were passed to `go test`. In order to hack arguments for re-runs, `gotestsum` requires a delineation between the arguments to `go test` and the arguments to the test binary. However, so that usage is consistent when E2E_USE_GOTESTSUM is present or not, I changed both to do the equivalent of:

```
go test ${ARGS} -args ${GO_E2E_TEST_ARGS}
```

In other words, for the e2e targets:
* Use `ARGS` to pass args to `go test` (e.g. `-parallel 16`)
* Use `GO_E2E_TEST_ARGS` to pass args to the test binary (e.g. `--kubeconfig`)
* If `E2E_USE_GOTESTSUM` is on, use `GOTESTSUM_ARGS` to pass args to `gotestsum` (e.g. `--rerun-fails=5`)

/kind cleanup

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -163,9 +163,20 @@ ifdef NO_RACE_DETECTOR
   RACE_DETECTOR_ARGS=
 endif
 
-go_test_cmd = go test -mod=vendor $(RACE_DETECTOR_ARGS)
+go_test_args = -mod=vendor $(RACE_DETECTOR_ARGS)
+go_test_cmd = go test $(go_test_args)
 GO_TEST=$(DOCKER_RUN) $(go_test_cmd)
 GO_E2E_TEST_ARGS=--kubeconfig /root/.kube/$(kubeconfig_file)
+
+gotestsum_json=/tmp/agones.gotestsum.json
+ifdef GOTESTSUM_VERBOSE
+gotestsum_format=--format=testname \
+  --post-run-command="sh -c 'echo; echo --- RAW VERBOSE OUTPUT ---; jq -j \"select(.Output != null) | .Output\" < $(gotestsum_json); echo --- END RAW VERBOSE OUTPUT ---'"
+else
+gotestsum_format=--format=standard-quiet
+endif
+gotestsum_cmd=gotestsum $(gotestsum_format) --jsonfile=$(gotestsum_json) --hide-summary=skipped --rerun-fails=2 $(GOTESTSUM_ARGS)
+GOTESTSUM=$(DOCKER_RUN) $(gotestsum_cmd)
 
 PERF_OUTPUT_DIR=$(mount_path)/build/.perf
 
@@ -178,6 +189,7 @@ ifdef LOCAL_GO
 	GO_BUILD_WINDOWS_AMD64=GOOS=windows GOARCH=amd64 $(gomod_on) $(go_build_cmd)
 	GO_BUILD_DARWIN_AMD64=GOOS=darwin GOARCH=amd64 $(gomod_on) $(go_build_cmd)
 	GO_TEST=$(go_test_cmd) -v
+	GOTESTSUM=$(gotestsum_cmd)
 	GO_E2E_TEST_ARGS=
 	PERF_OUTPUT_DIR=$(build_path)/.perf
 	go_build_base_path=$(agones_path)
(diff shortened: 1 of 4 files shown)


build/Makefile
```diff
@@ -163,10 +163,21 @@ ifdef NO_RACE_DETECTOR
...
 GO_E2E_TEST_ARGS=--kubeconfig /root/.kube/$(kubeconfig_file)
 
+gotestsum_json=/tmp/agones.gotestsum.json
+ifdef GOTESTSUM_VERBOSE
+gotestsum_format=--format=testname \
+  --post-run-command="sh -c 'echo; echo --- RAW VERBOSE OUTPUT ---; jq -j \"select(.Output != null) | .Output\" < $(gotestsum_json); echo --- END RAW VERBOSE OUTPUT ---'"
+else
+gotestsum_format=--format=standard-quiet
+endif
+gotestsum_cmd=gotestsum $(gotestsum_format) --jsonfile=$(gotestsum_json) --hide-summary=skipped --rerun-fails=2 $(gotestsum_post_run_command) $(GOTESTSUM_ARGS)
```
gongmax, line 178: Is `gotestsum_post_run_command` set anywhere? Is it supposed to be the same thing as the `--post-run-command` in line 174?

gongmax: Is `GOTESTSUM_ARGS` needed? You mentioned "use `GOTESTSUM_ARGS` to pass args to gotestsum (e.g. --rerun-fails=5)", but looks like those fields (i.e. jsonfile, hide-summary, rerun-failes) are set explicitly here.

build/Makefile
```diff
@@ -301,31 +313,51 @@ test-e2e:
...
+	--gameserver-image=$(GS_TEST_IMAGE) \
+	--pullsecret=$(IMAGE_PULL_SECRET)
 test-e2e-failure: $(ensure-build-image)
-	echo "starting e2e controller failure test"
-	$(GO_TEST) -parallel=1 $(agones_package)/test/e2e/controller $(GO_E2E_TEST_ARGS) \
-    	--gameserver-image=$(GS_TEST_IMAGE) \
-    	--pullsecret=$(IMAGE_PULL_SECRET)
+	echo "Starting e2e controller failure test!"
+ifdef E2E_USE_GOTESTSUM
+	$(GOTESTSUM) --packages=$(agones_package)/test/e2e/controller -- $(go_test_args) $(ARGS) -parallel=1 -args \
```
gongmax, line 345: `&(ARGS)` has the `-parallel=16` set but here `-parallel` is explicitly set to 1. Should the `$(ARGS)` be removed?

build/Makefile
zmerlynn: `gotestsum_post_run_command` was a vestigal of a previous factoring - missed that, thanks. Removed.

The idea behind `GOTESTSUM_ARGS` was just as an entry point to allow the user to specify more args specifically for `gotestsum`. We default to the ones you mention, but I intentionally but it at the end so if you want to override any of those default things (e.g. `--rerun-fails=5`), you can do that.

build/Makefile
zmerlynn: I think you're referring to the `ARGS` coming from e2e.sh, but two things:

* in my mind, `ARGS` is generically the args we pass to `go test` - it just happens that `e2e.sh` explicitly overrides it, I think because sometime back someone decided to force the parallelism possibly higher than the default. But you can also `make test-e2e`, so this should remain generic to that as well
* _that said_, for this test specifically we want to run it serially regardless (here I am just reflecting what we do in the `go test` case). Why we chose to do this via `-parallel=1` vs just not removing the `t.Parallel()`, I'm not clear, but I chose not to change the behavior here.

zmerlynn:
BTW, formatting changes can be previewed in the [logs here](https://pantheon.corp.google.com/cloud-build/builds/edc33727-25c1-41ee-8011-8e2679a2ffe8;step=22?project=agones-images).

gongmax (approved):

markmandel (approved):

roberthbailey (approved):

roberthbailey:
In `e2e-feature-gates`:

```
Shutdown triggered or timeout during lock acquisition
```

roberthbailey:
The last 4 builds in CI have all failed with that same error, which looks like a CI infrastructure bug than anything triggered by this PR.

PR #2913, opened 2023-01-19
Add GKE Autopilot to e2e

zmerlynn:
Run e2e tests against Autopilot 1.24 cluster. Mark as `allowFailure` for now just to get a sense for whether it works - I will manually scan CI after this goes in.

We probably should script additions/mutations to this in the future...

Towards #2777


gongmax (approved):

cloudbuild.yaml
```diff
@@ -268,13 +270,50 @@ steps:
...
   waitFor:
     - push-images
     - build-e2e
 
+#
+# Run the e2e tests on GKE Autopilot with FeatureGates inverted compared to Stable
+#
+# Keep in sync with (the inverse of) pkg/util/runtime/features.go:featureDefaults
+
+- name: 'e2e-runner'
```
gongmax, line 287: Just to leave a note here, this will be refactored a little bit if we have consensus on the CI change proposed in the design of multiple version support. But definitely looks good for now.

zmerlynn: Yeah, this was meant as an incremental addition for sure.

PR #2914, opened 2023-01-19
Adds snapshot Hugo env to separate from default env

### LINKED ISSUE #2279: The Documentation/Installation page on old releases shows an incorrect Kubernetes version

igooch:
**What type of PR is this?**

/kind bug

Hugo changed the default environment to "production" which is indexable for search engines. This PR creates a new environment "snapshot" that is not indexable for when we create snapshots of previous versions of the website. This PR also updates the shortcodes to use the supported K8s version for the "snapshot" environment.

Closes #2279

~~This code has been tested using `make site-server ENV=RELEASE_VERSION=1.28.0 ARGS=` and `make site-server ENV=RELEASE_VERSION=1.28.0 HUGO_ENV=blah ARGS=` but I could not find another way to test how this code would be deployed in production.~~ Tested by running `make site-static` and then `make release-deploy-site` on gcloud App Engine.


build/includes/release.mk
```diff
@@ -63,7 +63,7 @@ release-deploy-site: DOCKER_RUN_ARGS += -e GOFLAGS="-mod=mod" --entrypoint=/usr/
 release-deploy-site:
 	version=$$($(DOCKER_RUN) run $(mount_path)/build/scripts/previousversion/main.go -version $(base_version)) && \
 	echo "Deploying Site Version: $$version" && \
-	$(MAKE) site-deploy SERVICE=$$version
+	$(MAKE) site-deploy SERVICE=$$version HUGO_ENV=production
```
markmandel, line 66:
```suggestion
	$(MAKE) site-deploy SERVICE=$$version HUGO_ENV=snapshot
```

Since this is what we call at release time to capture the site snapshot

build/includes/website.mk
```diff
@@ -25,12 +25,14 @@
 #
 
 # generate the latest website
+site-server: HUGO_ENV ?=snapshot
```
markmandel, line 28: This does change how the development server works - which is usually to show the latest development version -- wondering if it should stay that way?

```suggestion
```

Also, wouldn't this need to be `HUGO_ENV=snapshot` ? But given that we already have an `ENV` var for passing in custom parameters, I think we have the use case of changing the HUGE_ENV environment variables covered?

build/includes/website.mk
```diff
@@ -25,12 +25,14 @@
...
 site-server: ARGS ?=-F
 site-server: ENV ?= RELEASE_VERSION="$(base_version)" RELEASE_BRANCH=main
 site-server: ensure-build-image
 	docker run --rm $(common_mounts) --workdir=$(mount_path)/site $(DOCKER_RUN_ARGS) -p 1313:1313 $(build_tag) bash -c \
-	"$(ENV) hugo server --watch --baseURL=http://localhost:1313/ --bind=0.0.0.0 $(ARGS)"
+	"$(ENV) hugo server -e $(HUGO_ENV) --watch --baseURL=http://localhost:1313/ --bind=0.0.0.0 $(ARGS)"
 
+site-static: HUGO_ENV ?=snapshot
```
markmandel, line 35: Same as above - we have `ENV`, so not sure we need a `HUGO_ENV` setting?

markmandel:
Cloud Build is being extra grumpy - might need to `git commit --amend --date=now` and force push it up to restart the build process.

build/includes/website.mk
igooch: Good call, I can change it back. I mostly had this here for testing.

igooch: Actually testing this without the HUGO_ENV the site-server default environment is "development" -- should I change this to site-server: HUGO_ENV=production?

build/includes/website.mk
igooch: Ah OK, I didn't realize that Hugo recognizes all HUGO_* environment variables https://gohugo.io/getting-started/configuration/#configure-with-environment-variables. I'll go ahead and use the environment variable instead of the flag.

build/includes/website.mk
igooch: Leaving a note for my future self: The site-server is defaulting the environment to `development` because we are using an older version of Hugo. If we upgrade Hugo in the future the default environment behavior changes to `production` and we will need to explicitly call out the `development` environment.

build/includes/release.mk
```diff
@@ -60,6 +60,7 @@ example-image-markdown.%:
 # Deploys the site by taking in the base version and deploying the previous version
 release-deploy-site: $(ensure-build-image)
 release-deploy-site: DOCKER_RUN_ARGS += -e GOFLAGS="-mod=mod" --entrypoint=/usr/local/go/bin/go
+release-deploy-site: export ENV=HUGO_ENV=snapshot
```
markmandel, line 63:
```suggestion
```

Rather than doing this, make

```
$(MAKE) site-deploy SERVICE=$$version
```

into

```
$(MAKE) ENV=HUGO_ENV=snapshot site-deploy SERVICE=$$version
```

I expect it will work perfectly, and I think is more readable on what is happening.

Given that its a release time only command, it doesn't need to be configurable.

WDYT?

site/layouts/shortcodes/k8s-api.html
```diff
@@ -1,2 +1,7 @@
-{{- $version := cond (eq (getenv "HUGO_ENV") "production") $.Page.Site.Params.supported_k8s $.Page.Site.Params.dev_supported_k8s }}
-{{- $prefix := replace $version "." "-" }}https://v{{- $prefix }}.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v{{- $version }}/{{- .Get "href" }}
\ No newline at end of file
+<!-- $version must be assigned outside of if statement for variable scope purposes.
```
markmandel, line 1: There's somethign in this that creating a linebreak, you can see it here:

image

http://localhost:1313/docs/reference/fleet/

It could be HTML docs at the top, or it might be that `{{ end }}` should be `{{- end }}`. Whitespace and line breaks can be tricky in Hugo sometimes. Could also be linebreak we didn't have previously on the end of the shortcode.

site/layouts/shortcodes/k8s-version.html
```diff
@@ -1,5 +1,5 @@
-{{- if eq (getenv "HUGO_ENV") "production" }}
+{{- if or (eq (getenv "HUGO_ENV") "production") (eq (getenv "HUGO_ENV") "snapshot")}}
 {{- $.Page.Site.Params.supported_k8s }}
 {{- else }}
 {{- $.Page.Site.Params.dev_supported_k8s }}
-{{- end }}
```
markmandel, line 5: Pretty sure this is caused by the inclusion of the line break to the template here. We should probably remove it.

image

http://localhost:1313/docs/installation/creating-cluster/minikube/

site/layouts/shortcodes/minikube-k8s-minor-version.html
```diff
@@ -1,5 +1,5 @@
-{{- if eq (getenv "HUGO_ENV") "production" }}
+{{- if or (eq (getenv "HUGO_ENV") "production") (eq (getenv "HUGO_ENV") "snapshot")}}
 {{- $.Page.Site.Params.minikube_minor_supported_k8s }}
 {{- else }}
 {{- $.Page.Site.Params.dev_minikube_minor_supported_k8s }}
-{{- end }}
```
markmandel, line 5: I don't think the line break is causing an issue right now, but I'd still be tempted to remove the linebreak anyway just to prevent it from happening anyway.

markmandel (commented):
Pretty close - just fixing some rendering issues.

build/includes/release.mk
```diff
@@ -95,6 +95,11 @@ do-release: $(ensure-build-image)
 
 	@echo "Now go make the $(RELEASE_VERSION) release on Github!"
 
-build-release: 
-	cd $(agones_path) && gcloud builds submit . --substitutions _VERSION=$(base_version) --config=./build/release/cloudbuild.yaml $(ARGS)
+ifdef RC_RELEASE
```
markmandel, line 98: 🤔  somehow this got added back in. I don't think it was supposed to.

igooch: Good catch, I think that was part of the rebase merge conflict, I'll update.

markmandel: Other than this, this good to merge 👍🏻 Just checked the other formatting comments below.

markmandel (approved):

PR #2915, opened 2023-01-20
Move the e2e test cluster Consul installation to the top-level gke module

gongmax:
…dule

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Move the e2e test cluster Consul installation to the top-level gke module, so we can mitigate the issue that we cannot use the same Module file to create multiple e2e test cluster since [Terraform with Helm doesn’t support specifying the provider for multiple clusters in a single module file](https://github.com/hashicorp/terraform-provider-helm/issues/539).
Also add a gate variable so by default the Consul is not installed in the cluster.
 
Closes #

### DIFF
--- a/build/terraform/e2e/module.tf
+++ b/build/terraform/e2e/module.tf
@@ -44,38 +44,10 @@ module "gke_cluster" {
     "initialNodeCount"      = 10
     "enableImageStreaming"  = true
     "project"               = var.project
+    "installConsul"         = true
   }
 
   firewallName = "gke-game-server-firewall"
-}
-
-provider "helm" {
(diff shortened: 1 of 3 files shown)


zmerlynn:
I think #2912 has a better/different factoring of this now but I appreciate the proof of concept!

gongmax:
> I think #2912 has a better/different factoring of this now but I appreciate the proof of concept!

Yep, I looked into #2912 briefly. I agreed that is more elegant and looks like the correct direction to move forward. Close this accordingly.

PR #2916, opened 2023-01-20
Move pooneh-m to emeritus approvers

pooneh-m:
It's been a while since I have reviewed a PR in this repo. Removing myself from the active approvers list.


markmandel (approved):
And you are missed, but also understood!

PR #2923, opened 2023-01-24
Enable SplitControllerAndExtensions for e2e testing and changed value…

chiayi:
…s yaml for deployment

Important note: `extensions` values are inherited through the `controller` values using yaml inheritance. The `&controllerValues` represents the variable and what sub section(s) will be inherited. To override the field(s) in `extensions`, explicitly create the field of the same name **after** the inheritance. Declaring the field before inheritance will not work. 
If `controller` values need to be changed but `extensions` values do not, the field(s) would need to be explicitly set/copied into `extensions`, so essentially an override with the original value. 

example to override: 
```
  controller: &example
    numWorkers: 100
    apiServerQPS: 400
    apiServerQPSBurst: 500
    http:
      port: 8080
...
  extensions:
    <<: *example
    apiServerQPS: 600
```
In the example, the `apiServerQPS:600` declared in `extensions` will override the `apiServerQPS` inherited from `controller`, but `extensions` will still have `extensions.numWorkers`, `extensions.apiServerQPSBurst`, etc.. 

Important Note: `helm --set` sets the variable after the inheritance, thus meaning that changing a `controller` value will not change the value inherited by `extensions`.

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This part of phase 1 for [HA Agones]. This moves the `SplitControllerAndExtensions` into alpha.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -64,7 +64,7 @@ KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 GS_TEST_IMAGE ?= us-docker.pkg.dev/agones-images/examples/simple-game-server:0.14
 
 # Enable all alpha feature gates. Keep in sync with `false` (alpha) entries in pkg/util/runtime/features.go:featureDefaults
-ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&SafeToEvict=true&PodHostname=true&Example=true"
+ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&SafeToEvict=true&PodHostname=true&SplitControllerAndExtensions=true&Example=true"
 
 # Build with Windows support
 WITH_WINDOWS=1
@@ -391,7 +391,6 @@ install: LOG_LEVEL := "debug"
 install: EXTERNAL_IP ?= $(shell $(DOCKER_RUN) kubectl get services agones-allocator -n agones-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 install: FEATURE_GATES ?= $(ALPHA_FEATURE_GATES)
 install: SERVICE_MONITOR := false
-install: CONTROLLER_SAFE_TO_EVICT := false # TODO(#2797): Remove when `SplitControllerAndExtensions` => safe-to-evict=true
 install: HELM_ARGS ?=
 install: $(ensure-build-image) install-custom-pull-secret
 	-$(DOCKER_RUN) bash -c '[[ $$(helm status agones -n agones-system --output json | jq -r ".info.status") =~ (failed|pending-.*) ]] && helm uninstall agones --namespace=agones-system || true'
@@ -408,7 +407,6 @@ install: $(ensure-build-image) install-custom-pull-secret
 		--set agones.featureGates=$(FEATURE_GATES) \
 		--set agones.allocator.service.loadBalancerIP=$(EXTERNAL_IP) \
 		--set agones.metrics.serviceMonitor.enabled=$(SERVICE_MONITOR) \
-		--set agones.controller.safeToEvict=$(CONTROLLER_SAFE_TO_EVICT) \
 		$(HELM_ARGS) \
 		agones $(mount_path)/install/helm/agones/
 
--- a/install/helm/agones/templates/controller.yaml
+++ b/install/helm/agones/templates/controller.yaml
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{- $featureGates := include "agones.featureGates" . | fromYaml }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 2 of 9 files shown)


install/helm/agones/values.yaml
```diff
@@ -45,7 +45,7 @@ agones:
   createPriorityClass: true
   priorityClassName: agones-system
   cloudProduct: "auto"
-  controller:
+  defaultValues: &defaultValues
```
zmerlynn, line 48: - I would definitely name this something a little less generic - they're not default across other jobs (allocation, ping service).
- But I'm also kind of vacillating now as to whether to have a common values, or to just inherit extensions from controller. Basically let's imagine a user already has `agones.controller.foobar` set - do we want `agones.extensions.foobar` also set, or do we want the override exclusive to `controller`? I see both sides.

chiayi: I think it might actually make more sense to have `extensions` inherit from `controller` unless there are plans to regularly make changes to the controller values.

zmerlynn: works for me, let's try it.

zmerlynn (approved):
Great work!

zmerlynn:
One thing: I'd encourage you in the PR description to at least call out the defaulting mechanism you're using here. In particular, we're going to need to document how to override just `controller`, or just `extensions`.

build/Makefile
```diff
@@ -391,7 +391,6 @@ install: LOG_LEVEL := "debug"
 install: EXTERNAL_IP ?= $(shell $(DOCKER_RUN) kubectl get services agones-allocator -n agones-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 install: FEATURE_GATES ?= $(ALPHA_FEATURE_GATES)
 install: SERVICE_MONITOR := false
-install: CONTROLLER_SAFE_TO_EVICT := false # TODO(#2797): Remove when `SplitControllerAndExtensions` => safe-to-evict=true
```
roberthbailey, line 394: Should this be left in until `SplitControllerAndExtensions` is enabled by default (e.g. beta)?

install/helm/agones/templates/controller.yaml
```diff
@@ -36,7 +37,11 @@ spec:
   template:
     metadata:
       annotations:
+{{- if $featureGates.SplitControllerAndExtensions }}
+        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
```
roberthbailey, line 41: Is there a difference between explicitly setting this to true vs. just leaving it off of the deployment entirely? Is it better to be explicit about it? I noticed in the extensions deployment we left it off.

build/Makefile
chiayi: I think since we have a check within the yaml, we won't need this. But in the case that `safe-to-evict` is enabled, and we still want it to be false for the `controller` we would want this. Not sure when that will happen though. What do you think?

zmerlynn: This only really matters for e2e testing by CI. If you look at how we've configured CI for Autopilot, it's a bit asymmetric now, with the `SaveToEvict` and `SplitControllerAndExtensions` feature gates being enabled in both the `-stable` and `-feature-gates` runs - basically these feature gates are mandatory on Autopilot. I added `CONTROLLER_SAFE_TO_EVICT` last week to allow us to add Autopilot to CI without gating on @chiayi's work, but now that we have the feature gate, I would rather drop the option. Dropping the option also means that on `e2e-stable` with the default gates on GKE returns to the normal configuration, rather than as it is today, where it's testing a slightly off config (stable gates + safe-to-evict annotation).

install/helm/agones/templates/controller.yaml
zmerlynn: Let's omit it. Technically `true` is [broader in scope than the omission](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) but I don't think any of the conditions applies to us. (It's not a `kube-system` pod, it uses ephemeral not local storage, and the controller is backed by a deployment.)

zmerlynn:
> Important note: extensions values are inherited through the controller values using yaml inheritance. The &controllerValues represents the variable and what sub section(s) will be inherited. To override the field(s) in extensions, explicitly create the field of the same name after the inheritance. Declaring the field before inheritance will not work.
If controller values need to be changed but extensions values do not, the field(s) would need to be explicitly set/copied into extensions, so essentially an override with the original value.

@chiayi - Can you provide some examples? In particular, what you're describing sounds like the user modifying the chart's values.yaml, but typically the user is going to either [use --set or provide a values.yaml](https://helm.sh/docs/chart_template_guide/values_files/). I believe, but have not a ton of experience with, the second values.yaml is an overlay on top of the chart (that's what the docs say), but I'm not clear on the order of inheritance at that point.

build/Makefile
roberthbailey: Thanks for the explanation. Make sense!.

zmerlynn (approved):

roberthbailey (approved):

PR #2928, opened 2023-01-25
Added GKE Workload Identity flag to GKE Terraform modules

abmarcum:
This adds a workload_id flag for both the gke & gke-autopilot Terraform modules. If the option is set to true, this will enable the Workload Identity GKE feature through the associated Terraform GKE option. 

By default, this option is set to false and Workload Identity will not be enabled on the GKE cluster. 

This option is being added so that [Allocation Endpoints](https://github.com/googleforgames/agones/tree/main/examples/allocation-endpoint) can be run on GKE clusters provisioned by the Agones Terraform GKE modules.

### DIFF
--- a/install/terraform/modules/gke-autopilot/cluster.tf
+++ b/install/terraform/modules/gke-autopilot/cluster.tf
@@ -27,6 +27,7 @@ locals {
   location          = lookup(var.cluster, "location", "us-west1")
   network           = lookup(var.cluster, "network", "default")
   subnetwork        = lookup(var.cluster, "subnetwork", "")
+  workload_id       = lookup(var.cluster, "workload_id", false)
   releaseChannel    = lookup(var.cluster, "releaseChannel", "REGULAR")
   kubernetesVersion = lookup(var.cluster, "kubernetesVersion", "1.24")
 }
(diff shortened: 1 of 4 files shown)


install/terraform/modules/gke-autopilot/cluster.tf
```diff
@@ -72,6 +73,13 @@ resource "google_container_cluster" "primary" {
     }
   }
 
+  dynamic "workload_identity_config" {
```
zmerlynn, line 76: Did you test this on Autopilot? I ask because I believe Workload Identity is always present on Autopilot: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable

install/terraform/modules/gke/cluster.tf
```diff
@@ -35,6 +35,7 @@ locals {
   windowsInitialNodeCount = lookup(var.cluster, "windowsInitialNodeCount", "0")
   windowsMachineType      = lookup(var.cluster, "windowsMachineType", "e2-standard-4")
   autoscale               = lookup(var.cluster, "autoscale", false)
+  workload_id             = lookup(var.cluster, "workload_id", false)
```
zmerlynn, line 38: I recognize that `workload_id` matches the rest of the terraform universe, but our modules have been using camelCaps for a while, it seems. Additionally, I would go ahead and expand the shortening: `workloadIdentity`.

zmerlynn (changes requested):
Thanks for the contribution!

PR #2933, opened 2023-01-26
Allocation Endpoint: Fix Makefile to correctly call docker build 

abmarcum:
The Allocation Endpoint [container image](us-docker.pkg.dev/agones-images/examples/allocation-endpoint-proxy:0.2) will not run due to the allocation-endpoint-proxy binary being incompletely built.

When you try to run the container image, the following error occurs:

> exec: "/allocation-endpoint-proxy": permission denied:  

This PR fixes the permission error by modifying the Makefile to include the correct build path when building the allocation-endpoint-proxy binary. 

Once the PR has been merged, an account with the appropriate repository permissions will need to re-publish the Allocation Endpoint [container image](us-docker.pkg.dev/agones-images/examples/allocation-endpoint-proxy:0.2).

### DIFF
--- a/examples/allocation-endpoint/Makefile
+++ b/examples/allocation-endpoint/Makefile
@@ -40,7 +40,7 @@ root_path = $(realpath $(project_path)/../..)
 
 # build the allocation endpoint proxy binary
 build:
-	cd $(root_path) && docker build -f $(project_path)/server/Dockerfile --tag=$(server_tag) .
+	cd $(project_path)/server/ && docker build --tag=$(server_tag) .
 
 run:
 	docker run --network=host $(server_tag) $(args)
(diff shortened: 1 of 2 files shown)


markmandel (commented):
Since we're fixing the image, we should also:

Increment the version number (0.3) here (and I can push up a new build once merged):
https://github.com/googleforgames/agones/blob/10490f0d2f734d92b8f2e7787c9685a4d4bb2234/examples/allocation-endpoint/Makefile#L31

And also in the Terraform here:
https://github.com/googleforgames/agones/blob/10490f0d2f734d92b8f2e7787c9685a4d4bb2234/examples/allocation-endpoint/terraform/variable.tf#L23

So that it's fixed everywhere.

Good catch!

markmandel:
Took the Makefile changes for a spin - all works correctly 👍🏻

markmandel (approved):

PR #2938, opened 2023-01-27
Use GCS as the Terraform state backend

gongmax:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Use GCS as the backend of the e2e test cluster Terraform state, so the maintainers can have a central store of the state. This will also benefit the management of the additional e2e test clusters we will introduced in near future.

As a side work, the PR also remove the consul from the Autopilot e2e test cluster.

Closes #

Since we will have the central authority of the e2e test clusters state, we should discourage users to create e2e test cluster for their local dev purpose. Users should be able to just create one `test-cluster` and perform all the dev on it including the e2e test. Instruction changes will be in following PRs.


build/terraform/e2e/module.tf
```diff
@@ -28,6 +28,10 @@ terraform {
       version = "~> 2.3"
     }
   }
+  backend "gcs" {
+   bucket  = "agones-e2e-infra-bucket-tfstate"
```
markmandel, line 32: Given that a bucket name is unique to the project - this should probably be dynamic (use the project name as a prefix?) so it can be run against development projects without failing.

gongmax: Yeah, that's doable. But I have doubt whether it makes sense to create an e2e test cluster for development. I was thinking we should push user to just create one test-cluster and use it for the development purpose even the e2e test.

markmandel: With it the way it is, you can't test this Terraform script, except to test it in production.

gongmax: Yeah, I was expecting the backend part would be relatively stable since it's only a thing for our maintainers who have access to the infra and won't have many change (if any). But it's still a fair point. I will make it not hardcoded.

zmerlynn: Honestly just something like `{project}-tfstate` makes sense to me, that's what I was thinking of doing.

build/includes/google-cloud.mk
```diff
@@ -53,28 +53,28 @@ clean-gcloud-test-cluster: $(ensure-build-image)
...
-	$(MAKE) terraform-init DIRECTORY=e2e
+	$(MAKE) terraform-init BUCKET=$(GCP_PROJECT)-e2e-infra-bucket-tfstate PREFIX=terraform/state DIRECTORY=e2e
 	$(DOCKER_RUN) bash -c 'cd $(mount_path)/build/terraform/e2e && terraform destroy -var project=$(GCP_PROJECT) -auto-approve'
 
 # Creates a gcloud cluster for prow
 gcloud-prow-build-cluster: GCP_PROJECT ?= $(shell $(current_project))
 gcloud-prow-build-cluster: $(ensure-build-image)
 gcloud-prow-build-cluster:
-	$(MAKE) terraform-init DIRECTORY=prow
+	$(MAKE) terraform-init BUCKET=$(GCP_PROJECT)-prow-infra-bucket-tfstate PREFIX=terraform/state DIRECTORY=prow
```
zmerlynn, line 71: maybe not strictly necessary here, but given that we are at an inflection point of changing the e2e infrastructure _and_ I don't think any of us understands if the Prow configuration is even working, should we just instead turn down Prow? I am afraid that if we continue as is we'll give someone a false sense of "maybe it works" when I feel like it has reached the point of bitrot.

roberthbailey: I have no objections to removing the prow cluster.

markmandel: > I have no objections to removing the prow cluster.

Who are you, and what have you done with the real @roberthbailey ? 😁

build/terraform/e2e/module.tf
```diff
@@ -14,6 +14,7 @@
 
 
 // Run:
+//  terraform init -backend-config="bucket=<YOUR_GCP_ProjectID>-e2e-infra-bucket-tfstate" -backend-config="prefix=terraform/state"
```
zmerlynn, line 17: out of curiousity, what happens if you don't provide these variables? does terraform error out in an obvious way, or does it default to something bad?

zmerlynn (approved):
Left a minor question. It's non-blocking to merge.

build/terraform/e2e/module.tf
gongmax: If starting with a clean state, i.e. there is no tfstate file for this module in your dev environment, `terraform init` without those variables will succeed but the local root configuration directory for this module will be used as backend. But the following `terraform apply` will fail because the state doesn't match with the existing resources, unless the related resources (clusters, firewalls) are all deleted, which is basically starting from a brand new state.

zmerlynn: Good point. I think that's good enough, yeah.

PR #2940, opened 2023-01-28
Add types for nodejs sdk

### LINKED ISSUE #2937: TypeScript types for Node SDK
It would be rad to bundle types with the `@google-cloud/agones-sdk` NPM package :+1:

vasily-polonsky:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix

In npm package there are no types. Should be helpful for typescript projects.

Closes #2937

Many return types are not perfect, but I dont know where to get the proper types, that can be used here.
One more thing, I'm not sure if this is enough to get the types into the npm package or if some other changes are needed also

### DIFF
--- a/sdks/nodejs/src/agonesSDK.d.ts
+++ b/sdks/nodejs/src/agonesSDK.d.ts
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC All Rights Reserved.
+// Copyright 2023 Google LLC All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


steven-supersolid:
Nice, this will be really useful. Will check the types match

sdks/nodejs/src/agonesSDK.d.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    alpha: Alpha
+
+    get port(): string
+
+    connect(): Promise<void>
+
+    close(): void
+
+    ready(): Promise<Record<string, any>>
```
steven-supersolid, line 30: When using TS I usually use `Record<string, unknown>` here as would cast the type to something useful but maybe `any` is more flexible?
We could also supply full return types, e.g. interfaces/types for all these, or better to do in a separate PR?

vasily-polonsky: As for `unknown` and `any`, `unknown` is definitely better when you can distinguish type, there's no `instanceof` type check you can do, or another field like `isAxiosError` to distinguish a particular type. The problem I see with `unknown` is that in this particular case, you have to write
```
if(typeof awaitedValue === 'object') {
  if(typeof awaitedValue.prop1 === 'object') {
    awaitedValue.prop1.prop1A
  }
}
```
with `any` you can just use awaitedValue?prop1?prop1A with the same security.
Maybe I am missing something?

vasily-polonsky: Can we provide full return types? That would be much better. I just started working with Agones last week, but for me the real result only returns for `getGameServer` and `watchGameServer` methods, for the rest sdk returns an empty object, maybe something is wrong in my cluster.
if we could specify the exact types, I would definitely do in this PR

markmandel:
@steven-supersolid lemme know when this is all good, and happy to approve and merge 👍🏻

steven-supersolid (approved):
LGTM. Great stuff!

vasily-polonsky:
Sorry, read the pull request process, and thought it should be merged automatically after approval, but it wasn't. Thought it was because of commits ahead in the main branch, merged the main branch here, but that action removed the approval. What should I do?

steven-supersolid:
Still LGTM

zmerlynn:
> Sorry, read the pull request process, and thought it should be merged automatically after approval, but it wasn't. Thought it was because of commits ahead in the main branch, merged the main branch here, but that action removed the approval. What should I do?

Tag an approver and we can handle it. I'll do so now!

zmerlynn (approved):

Issue #2716, opened 2022-08-24
Arbitrary Counts and Lists for GameServers, SDKs and Allocation


markmandel:
# Objective

With the recent work with Player Tracking, as well as it’s cross over into High Density Game Server support / re-allocation of Allocated `GameServers`, it seems that to be able to provide arbitrary count values and/or lists of values that are tied to GameServers, much like Player Tracking values are right now, is very useful for a wide variety of use cases.

This feature design’s contention is to replace Player Tracking with a generic way to track general counts as well as lists against a `GameServer` by an user provided key, as well as with integrated allocation, Fleet scheduling and SDK support, such that it can support the use case of player tracking as it currently stands, but also use cases like multi-tenant room server counting, or any other game specific value that could be utilised for a custom integration.

An added benefit would be that simple gauge data as metrics would be exposed as well, although we may not want to advocate this as a blessed path for _only_ exporting metrics, if not taking advantage of other functionality.

This feature would be built behind the `GameServerCountsAndLists` feature gate and should be on-par with `PlayerTracking` before the `PlayerTracking` functionality is removed.

# Requirements

*   Define on a `GameServer` a set of attached lists and/or counters attached to an arbitrary, user supplied key
*   Out of scope: The ability to add/edit or delete `GameServer` keys for counters and list at runtime. Keys should be explicitly predefined with the `GameServer` definition  to put some limits on what can be stored against fuetcd and ideally avoid overloading the Kubernetes API control plane (although we will need strong documentation about this, as this will definitely put extra load on the control plane).

## Counters

*   Counters can have an initial value (0 is the default).
*   Counters can have a set capacity (maximum value), but by default are 0 (max of int64). 
    *   We deliberately are using the term “capacity” across both list and counters to be consistent between the two pieces of functionality..
*   Incrementation / decrementation below 0 or above set capacity will be a no-op. I.e. No operations to increment/decrement a counter will error. 
*   Counters must be >= 0
*   SDK capability to atomically get, increment, decrement and set a counter local value, which is then set to the backing `GameServer` CRD status.
    *   Note: There are race conditions we can’t avoid between SDK updates and Allocation or external updates.
*   SDK capability to change the maximum value for a counter.
*   The ability to atomically increment or decrement counts on allocation
    *   If a user wants to ensure there is room for the increment or decrement, that should be explicitly included in the filter options (i.e. decrement by one, but filter for counts that are > 0 so that there is something to decrement).
[...]

PR #2946, opened 2023-02-02
Arbitrary Counts and Lists Feature/CRD

igooch:
**What type of PR is this?**

/kind feature

Creates an alpha feature gate for the Arbitrary Counts and Lists feature that will eventually replace player tracking.
Adds Counts and Lists to the gamerserver, gamerserverset, and fleet CRDs.

Working on #2716

### DIFF
--- a/examples/gameserver.yaml
+++ b/examples/gameserver.yaml
@@ -86,6 +86,23 @@ spec:
   #   # set this GameServer's initial player capacity
   #   initialCapacity: 10
   #
+  # [Stage:Alpha]
+  # [FeatureFlag:CountsAndLists]
+  # Lists provides the configuration for generic (player, room, session, etc.) tracking features.
+  # Commented out since Alpha, and disabled by default
+  # lists:
+  #   - name: playerlist
+  #     count: 0
+  #     capacity: 1000
+  #     values:
+  #   - name: roomlist
+  #     count: 3
+  #     capacity: 333
+  #     values:
+  #       - room1
+  #       - room2
+  #       - room3
+  #
   # Pod template configuration
   template:
     # pod metadata. Name & Namespace is overwritten
--- a/install/helm/agones/templates/crds/_gameserverspecschema.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecschema.yaml
@@ -164,7 +164,7 @@ properties:
           properties:
             name: # counter name
               type: string
-            count: # current count
+            count: # initial count
               type: integer
             capacity: # max capacity of the counter
               type: integer
@@ -178,14 +178,14 @@ properties:
           properties:
             name: # array name (players, sessions, rooms, etc.)
               type: string
-            count: # number of items in the values array
+            count: # initial number of items in the values array
               type: integer
-            capacity: # maximum capacity of the array (can be less than or equal to value of maxItems)
+            capacity: # max capacity of the array (can be less than or equal to value of maxItems)
               type: integer
             values:
               type: array
               x-kubernetes-list-type: set # Requires items in the array to be unique
-              maxItems: 1000 # maximum possible size of the value array (cannot be updated)
+              maxItems: 1000 # max possible size of the value array (cannot be updated)
               items: # name of the item (player1, session1, room1, etc.)
                 type: string
 {{- end }}
--- a/install/helm/agones/templates/crds/_gameserverspecschema.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecschema.yaml
@@ -153,6 +153,42 @@ properties:
             type: integer
             title: The initial player capacity of this Game Server
             minimum: 0
+{{- if .featureCountsAndLists }}
+      counts:
+        type: array
+        title: Array of player, room, session, etc. counters
+        nullable: true
+        maxItems: 1000
+        items:
+          type: object
+          properties:
+            name: # counter name
+              type: string
+            count: # current count
+              type: integer
+            capacity: # max capacity of the counter
+              type: integer
+      lists:
+        type: array
+        title: Array of player, room, session, etc. lists
+        nullable: true
+        maxItems: 1000
+        items:
+          type: object
+          properties:
+            name: # array name (players, sessions, rooms, etc.)
+              type: string
+            count: # number of items in the values array
+              type: integer
+            capacity: # maximum capacity of the array (can be less than or equal to value of maxItems)
+              type: integer
+            values:
+              type: array
+              x-kubernetes-list-type: set # Requires items in the array to be unique
+              maxItems: 1000 # maximum possible size of the value array (cannot be updated)
+              items: # name of the item (player1, session1, room1, etc.)
+                type: string
+{{- end }}
 {{- if .featureSafeToEvict }}
       eviction:
         type: object
--- a/install/helm/agones/templates/crds/_gameserverspecschema.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecschema.yaml
@@ -154,7 +154,7 @@ properties:
             title: The initial player capacity of this Game Server
             minimum: 0
 {{- if .featureCountsAndLists }}
-      counts:
+      counters:
         type: array
         title: Array of player, room, session, etc. counters
         nullable: true
--- a/install/helm/agones/templates/crds/_gameserverstatus.yaml
+++ b/install/helm/agones/templates/crds/_gameserverstatus.yaml
@@ -65,6 +65,40 @@ status:
           nullable: true
           items:
             type: string
+{{- if .featureCountsAndLists }}
+    counts:
+      type: array
+      nullable: true
+      maxItems: 1000
+      items:
+        type: object
+        properties:
+          name:
+            type: string
+          count:
+            type: integer
+          capacity:
+            type: integer
+    lists:
+      type: array
+      nullable: true
+      maxItems: 1000
+      items:
+        type: object
+        properties:
+          name:
+            type: string
+          count:
+            type: integer
+          capacity:
+            type: integer
+          values:
+            type: array
+            x-kubernetes-list-type: set
+            maxItems: 1000
+            items:
+              type: string
+{{- end }}
 {{- if .featureSafeToEvict }}
     eviction:
       type: object
--- a/install/helm/agones/templates/crds/_gameserverstatus.yaml
+++ b/install/helm/agones/templates/crds/_gameserverstatus.yaml
@@ -66,7 +66,7 @@ status:
           items:
             type: string
 {{- if .featureCountsAndLists }}
-    counts:
+    counters:
       type: array
       nullable: true
       maxItems: 1000
--- a/install/helm/agones/templates/crds/fleet.yaml
+++ b/install/helm/agones/templates/crds/fleet.yaml
@@ -100,7 +100,7 @@ spec:
                             - type: integer
                             - type: string
                 template:
-                  {{- $data := dict "metadata" true "podPreserveUnknownFields" .Values.gameservers.podPreserveUnknownFields "featureSafeToEvict" $featureGates.SafeToEvict }}
+                  {{- $data := dict "metadata" true "podPreserveUnknownFields" .Values.gameservers.podPreserveUnknownFields "featureSafeToEvict" $featureGates.SafeToEvict "featureCountsAndLists" $featureGates.CountsAndLists }}
                   {{- include "gameserver.schema" $data | indent 17 }}
             status:
               description: 'FleetStatus is the status of a Fleet. More info:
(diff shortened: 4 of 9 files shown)


install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
...
+        title: Array of player, room, session, etc. counters
+        nullable: true
+        maxItems: 1000
+        items:
+          type: object
+          properties:
+            name: # counter name
+              type: string
+            count: # current count
+              type: integer
```
zmerlynn, line 168: is this to avoid users of the CRD having to get the size of the array? o/w why do we need a count?

install/helm/agones/templates/crds/_gameserverstatus.yaml
```diff
@@ -65,6 +65,40 @@ status:
           nullable: true
           items:
             type: string
+{{- if .featureCountsAndLists }}
+    counts:
```
zmerlynn, line 69: maybe more of a design question, but this whole feature seems like it belongs in .Status rather then .Spec as it reflects the ongoing gameserver status. Here it's reflected in both?

zmerlynn (commented):
The feature gate LGTM, left some comments on the CRD.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
igooch: Yep, that's all it is. We'll use it for aggregating number of players/rooms/sessions etc. across gameservers, but again that's just to avoid getting the size of the array. Any opinions on if we should leave it in/out?

zmerlynn: yeah I just realized it was probably a convenience thing. Add it to the validation for sure, though.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
             type: integer
             title: The initial player capacity of this Game Server
             minimum: 0
+{{- if .featureCountsAndLists }}
+      counts:
```
gongmax, line 157: nit: the design doc uses `counter`. Personally I feel it's better since the `counter` object contains the names, capacity, besides count. But not sure whether there were other decisions made on the naming that I wasn't aware of.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
...
+        type: array
+        title: Array of player, room, session, etc. counters
+        nullable: true
+        maxItems: 1000
+        items:
+          type: object
+          properties:
+            name: # counter name
+              type: string
+            count: # current count
```
gongmax, line 167: should it have a `minimum: 0`?

igooch: Yes, planning on doing that as part of the validation PR (coming up next :))

install/helm/agones/templates/crds/_gameserverspecschema.yaml
igooch: How about "counters"?

gongmax: counters looks good to me

install/helm/agones/templates/crds/_gameserverstatus.yaml
zmerlynn: Oh, I see your point. Can a counter/list only be started in the Spec, or will we allow new counters/lists at any given time?

igooch: Ideally allow new counters/lists at any given time? Be able to start up a gameserver with a counter/list, or add one on later.

igooch: There's also the overall size limits for the CRD to consider too. Letting users have 1000 lists of 1000 values each in multiple places will leave the potential for users to go over the CRD size limits.

gongmax: From the design: 

> Out of scope: The ability to add/edit or delete GameServer keys for counters and list at runtime. 

So new counters/lists at any given time seems not allowed.

gongmax: I think it makes sense to have this in both Spec and Status. In Spec it's the initial value while in Status it's the current value. The property name and comment are a little confusing though, i.e. in line 167 of `_gameserverspecschema.yaml`, I think the comment should be `initial count` rather than `current count`.

gongmax (approved):

examples/gameserver.yaml
```diff
@@ -86,6 +86,23 @@ spec:
   #   # set this GameServer's initial player capacity
   #   initialCapacity: 10
   #
+  # [Stage:Alpha]
+  # [FeatureFlag:CountsAndLists]
+  # Lists provides the configuration for generic (player, room, session, etc.) tracking features.
+  # Commented out since Alpha, and disabled by default
+  # lists:
```
markmandel, line 93: This doesn't look right -- should be separate `counts` and `lists`.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
             type: integer
             title: The initial player capacity of this Game Server
             minimum: 0
+{{- if .featureCountsAndLists }}
+      counters:
+        type: array
+        title: Array of player, room, session, etc. counters
+        nullable: true
+        maxItems: 1000
```
markmandel, line 161: Nice touch. I can't imaging anyone will need this many, but nice to have a top limit.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
...
+        items:
+          type: object
+          properties:
```
markmandel, line 164: So in the original design we have this:

```yaml
  counters: # list of counters. Key value below is the key for each counter.
    rooms: # key for the counter (room)
      default: 1 # initial value
      capacity: 100 # maximum possible count value
```

Where the name is the key, and `default` is the default value.

In this type of setup, what is the key used for? And why the deviation from the design?

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
...
+              type: integer
+      lists:
+        type: array
+        title: Array of player, room, session, etc. lists
+        nullable: true
+        maxItems: 1000
+        items:
+          type: object
+          properties:
+            name: # array name (players, sessions, rooms, etc.)
```
markmandel, line 179: Same as above, in the original design we have:

```yaml
  counters: # list of counters. Key value below is the key for each counter.
    rooms: # key for the counter (room)
      default: 1 # initial value
      capacity: 100 # maximum possible count value
```

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
...
+              type: string
+            count: # initial number of items in the values array
+              type: integer
+            capacity: # max capacity of the array (can be less than or equal to value of maxItems)
+              type: integer
+            values:
+              type: array
+              x-kubernetes-list-type: set # Requires items in the array to be unique
```
markmandel, line 187: `x-kubernetes-list-type: set` -- fancy!

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
...
+          type: object
+          properties:
+            name: # counter name
+              type: string
+            count: # initial count
+              type: integer
+            capacity: # max capacity of the counter
+              type: integer
+      lists:
+        type: array
```
markmandel, line 172: This should be an object.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -153,6 +153,42 @@ properties:
             type: integer
             title: The initial player capacity of this Game Server
             minimum: 0
+{{- if .featureCountsAndLists }}
+      counters:
+        type: array
```
markmandel, line 158: This should be an `object`

install/helm/agones/templates/crds/_gameserverstatus.yaml
```diff
@@ -65,6 +65,40 @@ status:
           nullable: true
           items:
             type: string
+{{- if .featureCountsAndLists }}
+    counters:
+      type: array
```
markmandel, line 70: Should be an `object`.

install/helm/agones/templates/crds/_gameserverstatus.yaml
```diff
@@ -65,6 +65,40 @@ status:
...
+        type: object
+        properties:
+          name:
+            type: string
+          count:
+            type: integer
+          capacity:
+            type: integer
+    lists:
+      type: array
```
markmandel, line 83: Should also be an object (similar format to the spec)

install/helm/agones/templates/crds/fleet.yaml
```diff
@@ -129,6 +129,40 @@ spec:
                     capacity:
                       type: integer
                       minimum: 0
+{{- if $featureGates.CountsAndLists }}
+                counters:
+                  type: array
```
markmandel, line 134: This should also be an object - same as above.

install/helm/agones/templates/crds/fleet.yaml
```diff
@@ -129,6 +129,40 @@ spec:
...
+                    type: object
+                    properties:
+                      name:
+                        type: string
+                      count:
+                        type: integer
+                      capacity:
+                        type: integer
+                lists:
+                  type: array
```
markmandel, line 147: Should also be an object, and reworked to fit the design.

install/helm/agones/templates/crds/gameserverset.yaml
```diff
@@ -112,6 +112,40 @@ spec:
                     capacity:
                       type: integer
                       minimum: 0
+{{- if $featureGates.CountsAndLists }}
+                counters:
+                  type: array
```
markmandel, line 117: Same as above - all should be an object.

markmandel (commented):
Sorry for the super late review, I should have looked at this when you asked - my apologies.

examples/gameserver.yaml
igooch: I only wrote a sample for the lists here and not a sample for couters. I can add that in.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
igooch: Hopefully no one will hit the limit. It's recommended for CEL validation which we may do in the future https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/#writing-validation-rules-in-openapiv3-schemas

install/helm/agones/templates/crds/_gameserverspecschema.yaml
igooch: I haven't put in the defaults yet, so I'll add those in in one or two PRs. Along those lines, would it make sense for the default to be 0 for a counter initial value?

install/helm/agones/templates/crds/_gameserverspecschema.yaml
markmandel: A good example is `labels` in ObjectMeta:

https://github.com/googleforgames/agones/blob/921b1c1aef91d3fea08876113ecf1338cbae7fa5/install/helm/agones/templates/crds/k8s/_io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml#L68-L72

install/helm/agones/templates/crds/_gameserverspecschema.yaml
markmandel: My point is that the naming in this schema doesn't match the design.

examples/gameserver.yaml
markmandel: This also doesn't match the design. Lists don't have a count value either on them in the spec.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
markmandel: Found some more references in the Open API Spec:
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#parameter-object-examples

Particularly under _A free-form query parameter, allowing undefined parameters of a specific type:_

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#model-with-mapdictionary-properties

(Although you can't use $ref in K8s CRDs, you have to specify everything inline, hence the need for includes).

That should help!

install/helm/agones/templates/crds/_gameserverspecschema.yaml
igooch: Hopefully no one will hit the limit :) Putting in a limit was recommended for if we do CEL validation in the future.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
igooch: Fixes in new PR https://github.com/googleforgames/agones/pull/2965

Issue #2926, opened 2023-01-25
Refactor the apiHooks / featureGates global singletons


zmerlynn:
I introduced the `apiHooks` global singleton in #2868, but it has some real issues with unit tests. This is not surprising, given that it's the very definition of spooky action at a distance.

An example of this issue can be found in failures like:

```
--- FAIL: TestGameServerPodHostName (0.00s)
    gameserver_test.go:1167: 
        	Error Trace:	/go/src/agones.dev/agones/pkg/apis/agones/v1/gameserver_test.go:1167
        	Error:      	Received unexpected error:
        	            	unknown eviction.safe value ""
        	            	agones.dev/agones/pkg/apis/agones/v1.generic.SetEviction
        	            		/go/src/agones.dev/agones/pkg/apis/agones/v1/apihooks.go:87
        	            	agones.dev/agones/pkg/apis/agones/v1.(*GameServer).Pod
        	            		/go/src/agones.dev/agones/pkg/apis/agones/v1/gameserver.go:709
        	            	agones.dev/agones/pkg/apis/agones/v1.TestGameServerPodHostName
        	            		/go/src/agones.dev/agones/pkg/apis/agones/v1/gameserver_test.go:1166
        	            	testing.tRunner
        	            		/usr/local/go/src/testing/testing.go:1446
        	            	runtime.goexit
        	            		/usr/local/go/src/runtime/asm_amd64.s:1594
        	Test:       	TestGameServerPodHostName
```

The reason this test is failing is that another test has enabled the `SafeToEvict` feature gate, which turns on different validation in `generic.SetEviction`: https://github.com/googleforgames/agones/blob/45d26e8e75b9615f06937009acd8c6f7e48548fb/pkg/apis/agones/v1/apihooks.go#L60-L64 .. now, we could get around this by having any test that calls `.Pod` also lock the `FeatureTestMutex` to fence any modifications of the feature gates, but at that point you might as well just run all f the `v1` package sequentially.

I think the right thing to do here is ... rather brutal. I would propose the introduction of e.g. an "AgonesContext" or something to store what we're doing with global singletons, then storing the feature gates and the apiHooks in it. This would let us avoid `FeatureTestMutex` and any spooky action that might happen from variable apiHooks or feature gates.

PR #2948, opened 2023-02-03
Remove apiHooks singleton

zmerlynn:
Apologies for the size of this - I iterated like 3 or 4 different ways to do this and, believe it or not, this was the least painful:

Instead of using a singleton `apiHooks`, this PR changes all things that depend eventually on apiHooks to take explicit interfaces. In most cases, that's the v1.APIHooks interface, which stays, but in a couple of cases I minimized to the specific portion needed. We do this by:

* Reverting a lot of the nature of #2868, but keeping the interface "layering" with a chunk of the interface defined in agonesv1/ and the remainder outside.

* Plumbing hooks from the top as before.

* Introduces a "FakeAPIHooks" interface for unit tests to use. Crucially, this interface is not at all dependent on feature gates.
  * As a call out, the fakes don't quite last all the way to to the controller tests that create a full controller. Those use the generic provider. I may poke at this more to remove the influence of feature gates here as well.

Towards #2926, but I'm not calling it fixed with this - I think we should bust the feature gates global in the runtime package as well to make the unit tests more predictable, but this ~mostly resolves the `apiHooks` cases.


pkg/cloudproduct/generic/generic_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"agones.dev/agones/pkg/apis"
+	"agones.dev/agones/pkg/apis/agones"
+	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
+	"agones.dev/agones/pkg/util/runtime"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestSetEviction(t *testing.T) {
```
roberthbailey, line 28: Were there any modifications from the `TestSetEviction` function deleted from `pkg/apis/agones/v1/apihooks_test.go`?

pkg/cloudproduct/generic/generic_test.go
```diff
@@ -0,0 +1,221 @@
...
+			assert.NoError(t, err)
+
+			err = (&generic{}).SetEviction(tc.safeToEvict, tc.pod)
+			assert.NoError(t, err)
+			assert.Equal(t, tc.wantPod, tc.pod)
+		})
+	}
+}
+
+func TestGameServerPodAutoscalerAnnotations(t *testing.T) {
```
roberthbailey, line 127: Similarly, were there any changes from the deleted `TestGameServerPodAutoscalerAnnotations` function in `pkg/apis/agones/v1/gameserver_test.go`?

pkg/cloudproduct/generic/generic.go
```diff
@@ -16,20 +16,62 @@
...
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 )
 
 // New returns a new generic cloud product
 //
 //nolint:revive // ignore the unexported return; implements ControllerHooksInterface
-func New() (*generic, agonesv1.APIHooks, error) { return &generic{}, nil, nil }
+func New() (*generic, error) { return &generic{}, nil }
```
roberthbailey, line 33: Is there a reason for the second return value, since there is never a case where this can return an error? I don't think `New()` is implementing an interface that requires this so maybe it can be simplified?

Then test code that used this could have the call `generic.New()` inlined instead of ignoring the second return value, e.g. in `pkg/gameservers/controller_test.go`.

pkg/gameservers/health.go
```diff
@@ -59,11 +58,13 @@ type HealthController struct {
...
 // NewHealthController returns a HealthController
-func NewHealthController(health healthcheck.Handler,
+func NewHealthController(
+	health healthcheck.Handler,
 	kubeClient kubernetes.Interface,
 	agonesClient versioned.Interface,
 	kubeInformerFactory informers.SharedInformerFactory,
-	agonesInformerFactory externalversions.SharedInformerFactory) *HealthController {
+	agonesInformerFactory externalversions.SharedInformerFactory,
+	waitOnFreePorts bool) *HealthController {
```
roberthbailey, line 67: I didn't see any places in the tests where we set this to true. Are we missing test coverage for that condition?

pkg/gameservers/migration.go
```diff
@@ -46,35 +45,39 @@ import (
...
+	podSynced                cache.InformerSynced
+	podLister                corelisterv1.PodLister
+	gameServerSynced         cache.InformerSynced
+	gameServerGetter         getterv1.GameServersGetter
+	gameServerLister         listerv1.GameServerLister
+	nodeLister               corelisterv1.NodeLister
+	nodeSynced               cache.InformerSynced
+	workerqueue              *workerqueue.WorkerQueue
+	recorder                 record.EventRecorder
+	syncPodPortsToGameServer func(*agonesv1.GameServer, *corev1.Pod) error
```
roberthbailey, line 58: Similarly, this is always set to the nil function in the tests. Is this another place with missing test coverage?

roberthbailey (commented):
My only remaining question is that in the test code there is now a lot of `gs.Pod(agonesv1.FakeAPIHooks{})` boilerplate. I'm wondering if there is a way to simplify this, e.g. by adding a wrapper function that does this automatically or something? In this change it makes sense but it feels like it's going to become cargo-culted copy-paste in the future.

zmerlynn:
> My only remaining question is that in the test code there is now a lot of `gs.Pod(agonesv1.FakeAPIHooks{})` boilerplate. I'm wondering if there is a way to simplify this, e.g. by adding a wrapper function that does this automatically or something? In this change it makes sense but it feels like it's going to become cargo-culted copy-paste in the future.

I personally like the explicitness and don't mind the copypasta - it's basically just a statement of "the behavior here may be influenced by API hooks", but I'm locking them to the fake ones for unit test consistency.

pkg/gameservers/health.go
zmerlynn: This is covered in https://github.com/googleforgames/agones/blob/35e899d302f5516a9845a13ac699abd410ce7645/pkg/gameservers/health_test.go#L62-L100 (but not at a higher level).

pkg/cloudproduct/generic/generic_test.go
zmerlynn: It was modified to shift packages but is otherwise the same.

pkg/cloudproduct/generic/generic_test.go
zmerlynn: It was modified to shift packages but is otherwise the same.

pkg/cloudproduct/generic/generic.go
zmerlynn: Originally I was being a bit stodgy about all of the cloudproduct interfaces having a `New()` function that could return an error since the ones that have meaningful auto-detection might error on the k8s client, but you're right that as implemented, there's no cause. Let me fix this up.

zmerlynn: Done

pkg/gameservers/migration.go
zmerlynn: This is tested here: https://github.com/googleforgames/agones/blob/3c38876382d505244af68489211d8916dce07596/pkg/gameservers/gameservers_test.go#L107-L164

It's again slightly indirected from the hook value (these were both examples of how I tested the hook callers prior to the hooks getting factored out and maybe they could be better now?).

zmerlynn:
Thanks for the review, PTAL!

roberthbailey (approved):

PR #2949, opened 2023-02-03
Move the e2e test steps to a child Cloud Build

gongmax:
As we are expending the E2E test matrix, the `cloudbuild.yaml` will become bloated if we just simply add new e2e test steps into it. The PR extracts the common logic of the e2e test cloud build step into a separate cloudbuild config file. And in the main Cloud Build, add a step to submit child Cloud Build for each of the e2e test and wait for them to complete. In this way, the main `cloudbuild.yaml` can keep maintainable and readable. 
Noted:
- The main Cloud Build should fail if any of the child e2e test Cloud Build fail
- A clean up step was added to cancel any orphan child e2e test Cloud Build to prevent multiple sets of tests running on the same cluster (Thanks for zmerlynn@ for pointing out the issue)

One recent successful build: 
https://console.cloud.google.com/cloud-build/builds;region=global/7067e537-cc4d-4a16-a0e8-07d0bb8e1180?project=agones-images
child e2e test cloud build:
https://console.cloud.google.com/cloud-build/builds;region=global/0e1a7aad-138f-4063-8495-7fbe00c59ba0;step=2?project=agones-images
https://console.cloud.google.com/cloud-build/builds;region=global/16fe2621-2cdc-409a-ad9f-ea5eff3abc3c;step=2?project=agones-images

/kind cleanup


cloudbuild.yaml
```diff
@@ -279,67 +268,77 @@ steps:
...
-    - 'CustomFasSyncInterval=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&SafeToEvict=true&PodHostname=true&SplitControllerAndExtensions=true&Example=true'
-    - 'gke-autopilot'
-    - 'gke-autopilot-e2e-test-cluster-1-24'
-    - 'us-west1'
-    - "${_REGISTRY}"
-  id: e2e-feature-gates-gke-autopilot
-  waitFor:
-    - e2e-wait-to-become-leader
+    on_going_builds=($(gcloud builds list --ongoing --filter "tags:'e2e-test'" --format="value(id)" | xargs))
+    for on_going_build in ${on_going_builds[@]}
```
zmerlynn, line 281: It looks like `gcloud builds cancel` takes multiple IDs, any reason for the loop?

I think this could all be collapsed into:

```
gcloud builds list --ongoing --filter "tags:'e2e-test'" --format="value(id)" | xargs --no-run-if-empty gcloud builds cancel 
```

cloudbuild.yaml
```diff
@@ -279,67 +268,77 @@ steps:
...
+   - '-c'
+   - |
+      set -e
+      pids=()
+      cloudProducts=("generic" "gke-autopilot")
+      for cloudProduct in ${cloudProducts[@]}
+      do
+        if [ $cloudProduct = generic ]
+        then 
+          featureWithGate="CustomFasSyncInterval=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&SafeToEvict=true&PodHostname=true&SplitControllerAndExtensions=true&Example=true"
```
zmerlynn, line 302: I don't think we need it yet but I'm wondering if we should instead keep a config blob and extract here (like JSON file and `jq`)? I think this is fine for now but it's a little unwieldy.

cloudbuild.yaml
```diff
@@ -279,67 +268,77 @@ steps:
...
+          featureWithoutGate=""
+          testClusterLocation="us-west1-c"
+          testCluster="e2e-test-cluster"
+        else
+          featureWithGate="CustomFasSyncInterval=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&SafeToEvict=true&PodHostname=true&SplitControllerAndExtensions=true&Example=true"
+          featureWithoutGate="SafeToEvict=true&SplitControllerAndExtensions=true"
+          testClusterLocation="us-west1"
+          testCluster="gke-autopilot-e2e-test-cluster-1-24"
+        fi
+        # TODO: Add more versions in the list, and generate the test cluster name in the loop rather than hardcode once the clusters are created
```
zmerlynn, line 312: As a callout that I'm sure will be a nuisance here, we can't support Autopilot on <1.24 yet. (I need to add it to the User Guide, woops.)

cloudbuild.yaml
```diff
@@ -279,67 +268,77 @@ steps:
...
+        if wait -n; then
+          :
+        else
+          status=$?
+          echo "One of the e2e test child cloud build exited with nonzero status $status. Aborting."
+          for pid in "${pids[@]}"; do
+            # Send a termination signal to all the children, and ignore errors
+            # due to children that no longer exist.
+            kill "$pid" 2> /dev/null || :
+            echo "killed $pid"
```
zmerlynn, line 333: This doesn't cancel the build, right? It's just a fast exit?

e2e-test-cloudbuild.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+#
+# build the e2e test runner
+#
+
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-f', 'Dockerfile', '-t', 'e2e-runner', '.']
+  dir: 'build/e2e-image'
+  id: build-e2e
+  waitFor: ['-']
```
zmerlynn, line 29: Is there no way to share this image with a sub-build? Seems fine, just seems a bit redundant. :)

zmerlynn: Actually, there's a good reason to avoid this here: If you can move this step back to the main build, you can use `submit --no-source` and skip the complexity of the .gcloudignore file.

Can the name (`e2e-runner`) can be tagged and uploaded to the CI registry like the "normal" images?

gongmax: I can only think of one way to share is to push the image somewhere and pull it in the sub-build, since different cloud builds are running on different machines. But I feel it's a little overhead since this step takes only around 1 min.

cloudbuild.yaml
gongmax: It's just a fast exit and doesn't cancel the build.

cloudbuild.yaml
gongmax: I see, I will keep in mind when revise this part.

cloudbuild.yaml
gongmax: Yes, totally agreed. I will think about it in following PRs.

cloudbuild.yaml
gongmax: I was not sure about if there is any limit around the number of IDs `gcloud builds cancel` can take at once.  So just using loop for safety. I will take your suggestion If there is no concern around this, it's much more elegant.

e2e-test-cloudbuild.yaml
gongmax: It's an interesting point that it might be possible to use `submit --no-source`. Does the `e2e-runner` image contains asset under `/test/e2e`? I think those source are necessary for the e2e test?

zmerlynn: Oh, you're right. I rescind that comment - you'd probably have to build the e2e binary and stuff it in the runner, which is more restructuring.

cloudbuild.yaml
zmerlynn: The `xargs` default limit is substantial, but more importantly, worst case it just spills to N commands.

cloudbuild.yaml
```diff
@@ -279,67 +268,72 @@ steps:
...
+          featureWithGate="CustomFasSyncInterval=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&SafeToEvict=true&PodHostname=true&SplitControllerAndExtensions=true&Example=true"
+          featureWithoutGate="SafeToEvict=true&SplitControllerAndExtensions=true"
+          testClusterLocation="us-west1"
+          testCluster="gke-autopilot-e2e-test-cluster-1-24"
+        fi
+        # TODO: Add more versions in the list, and generate the test cluster name in the loop rather than hardcode once the clusters are created
+        k8sVersions=("1.24")
+        for k8sVersion in ${k8sVersions[@]}
+        do
+          gcloud builds submit . --config=e2e-test-cloudbuild.yaml \
```
zmerlynn, line 311: Hmm, I'm looking at https://pantheon.corp.google.com/cloud-build/builds;region=global/e36f952d-efc8-488d-b79d-49cee84bf285;step=23?project=agones-images and I see one issue: The output is all spliced together.

Even if I go to the top of that log and flip to https://console.cloud.google.com/cloud-build/builds/5f8581b6-cfc7-4873-9cd5-cf15ee67b30f?project=258182270954, it's not obvious which build is running at that point.

Putting yourself in the shoes of a developer looking trying to understand failed e2e, can we make sure this is less toily?

I wonder, and it's a little bit nuts, if the steps in the child build can have substitutions in them? So for example instead of `e2e-stable` they become `e2e-stable-${CLOUD_PRODUCT}-${K8S_VERSION}` or something - then even this output in the main build would be understandable.

gongmax: After some research and consultation with Cloud Build folks, it turns out it is not possible to use substitutions in the `id` field of build config file. At this point, I feel there is no way to improve the main cloud build output and developer will need to go to the failed child e2e cloud build to see details about failed test. Maybe we can do something to the go test output to include the cluster name? But I'm not sure whether that's feasible either. What's your thought?

zmerlynn: Here is my thought for something lightweight:

* change line 311 to e.g.:

```
gcloud builds submit [...] |& sed -e "s/^/${cloudProduct}-${k8sVersion}: /"
```

this should prefix every line of the output with e.g. `autopilot-1.24: `, including the line at the top of the output that lists the links off to other builds.

* I would additionally add some output to the front of `entrypoint.sh` to all the incoming args. I think even just something like `echo e2e-runner arguments: "$*"` would be good enough - something so that if you're actually looking at a sub-build, you can decipher what config it is quickly.

gongmax: Oh, I didn't realize the build log in the console is actually the `gcloud builds submit` output. I will try this out.

zmerlynn: Output looks great after [4924d3b](https://github.com/googleforgames/agones/pull/2949/commits/4924d3bdd356b80c2a951c8d634a161aef6e1c55)!

I'm actually going to rescind the second bullet. This looks clean enough I think we can just rely on it. If you happen to browse to one of the sub-builds, you can view the substitutions in the `Execution Detail` tab as well: https://screenshot.googleplex.com/n8dwbbJkVEJ8Bdh .. and since the links are the console links anyways, I think those links will only really work for us. However, the GCS logs agones-bot lists will have the interleaved logs, so it should be possible to figure out what configuration failed regardless.

zmerlynn (approved):

PR #2950, opened 2023-02-03
Graduate `SafeToEvict` to Beta

### LINKED ISSUE #2931: Graduate `SafeToEvict` to Beta
I propose graduating `SafeToEvict` to Beta: It was Alpha in 1.29, with documentation in #2924.

The reasoning for the fast promotion is discussed in #2794: 

> Graduation to Beta should be relatively rapid as the feature has broad backward compatibility. The only situation where backward compatibility is violated is if the user is negatively impacted by a PDB on the game server pods, e.g. the user is relying on the default `safe-to-evict=false` pod annotation, but _also_ is assuming that node drain (e.g. via cluster upgrade) will evict. Given that our existing documentation doesn't allow for node drains, we think this is a safe case for backwards incompatibility. To be conservative about it, though, when we graduate `SafeToEvict` to `Beta`, we will flag it a potentially breaking change in Release Notes to call attention to it. To allow this situation, the user simply needs to set `eviction: safe: OnUpgrade`. 

I recognize that this feature was relatively undocumented as of yet (hence #2924) and the contract might be not quite right, but I think the risk to graduation is minimal.

Towards #2777

zmerlynn:
Runs the checklist, changes docs to reflect default status.

Fixes #2931

### DIFF
--- a/site/content/en/docs/Advanced/controlling-disruption.md
+++ b/site/content/en/docs/Advanced/controlling-disruption.md
@@ -12,7 +12,12 @@ description: >
 
 By default, Agones assumes your game server should never be disrupted voluntarily and configures the `Pod` appropriately - but this isn't always the ideal setting. Here we discuss how Agones allows you to control the two most significant sources of voluntary `Pod` evictions, node upgrades and Cluster Autoscaler, using the `eviction` API on the `GameServer` object. 
 
+{{% feature publishVersion="1.30.0" %}}
 {{< beta title="`eviction` API" gate="SafeToEvict" >}}
+{{% /feature %}}
+{{< feature expiryVersion="1.30.0" >}}
+{{< alpha title="`eviction` API" gate="SafeToEvict" >}}
+{{% /feature %}}
 
 ## Benefits of Allowing Voluntary Disruption
 
--- a/site/content/en/docs/Advanced/controlling-disruption.md
+++ b/site/content/en/docs/Advanced/controlling-disruption.md
@@ -12,7 +12,7 @@ description: >
 
 By default, Agones assumes your game server should never be disrupted voluntarily and configures the `Pod` appropriately - but this isn't always the ideal setting. Here we discuss how Agones allows you to control the two most significant sources of voluntary `Pod` evictions, node upgrades and Cluster Autoscaler, using the `eviction` API on the `GameServer` object. 
 
-{{< alpha title="`eviction` API" gate="SafeToEvict" >}}
+{{< beta title="`eviction` API" gate="SafeToEvict" >}}
 
 ## Benefits of Allowing Voluntary Disruption
 
--- a/site/content/en/docs/Advanced/scheduling-and-autoscaling.md
+++ b/site/content/en/docs/Advanced/scheduling-and-autoscaling.md
@@ -97,17 +97,30 @@ This affects the Cluster autoscaler, Allocation Scheduling, Pod Scheduling and F
 
 #### Cluster Autoscaler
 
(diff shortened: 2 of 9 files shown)


zmerlynn:
(Marking this WIP, there's a unit test that's a little angry.)

zmerlynn:
Should be better, PTAL!

site/content/en/docs/Advanced/controlling-disruption.md
```diff
@@ -12,7 +12,7 @@ description: >
 
 By default, Agones assumes your game server should never be disrupted voluntarily and configures the `Pod` appropriately - but this isn't always the ideal setting. Here we discuss how Agones allows you to control the two most significant sources of voluntary `Pod` evictions, node upgrades and Cluster Autoscaler, using the `eviction` API on the `GameServer` object. 
 
-{{< alpha title="`eviction` API" gate="SafeToEvict" >}}
+{{< beta title="`eviction` API" gate="SafeToEvict" >}}
```
roberthbailey, line 15: This should be wrapped in a feature publishVersion tag so that it doesn't show as beta until the next release is cut.

zmerlynn: Good catch - did it elsewhere, forgot this one.

zmerlynn:
Should be good now, PTAL!

site/content/en/docs/Advanced/scheduling-and-autoscaling.md
```diff
@@ -97,17 +97,30 @@ This affects the Cluster autoscaler, Allocation Scheduling, Pod Scheduling and F
...
+
+If a gameserver can tolerate [being evicted](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works)
+(generally in combination with setting an appropriate graceful termination period on the gameserver pod) and you
+want the Cluster Autoscaler to compact your cluster by evicting game servers when it would allow the Cluster
+Autoscaler to reduce the number of nodes in the cluster, [Controlling Disruption]({{< relref "controlling-disruption.md" >}}) describes
+how to choose the `.eviction` setting appropriate for your `GameServer` or `Fleet`.
+
+{{% /feature %}}
+{{< feature expiryVersion="1.30.0" >}}
 When using the “Packed” strategy, Agones will ensure that the Cluster Autoscaler doesn't attempt to evict and move `GameServer` `Pods` onto new Nodes during
```
roberthbailey, line 112: The new page no loses this note about the default behavior. I like having the section linking to the new disruption page, but I think it should still say that the default behavior is to disable the cluster autoscaler (we don't have to say how here since we link to the other page, but keeping the description of the default behavior seems important).

roberthbailey: Maybe keep

"When using the “Packed” strategy, by default Agones will ensure that the Cluster Autoscaler doesn’t attempt to evict and move GameServer Pods onto new Nodes during gameplay"

At the top of the section, followed by the alert & new paragraph starting with "If a gameserver can tolerate [being evicted](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works) ..."

zmerlynn: Fixed. I also noticed I had a rendering issue on 1.29 - the alerts don't render well unless you put them in {{< blocks, but the markdown requires it, sooo.. I mixed them. PTAL.

roberthbailey (approved):

PR #2953, opened 2023-02-07
Create stale.yml to mark stale issues..

geetachavan1:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Dry run of Action BOT to mark stale issues. This will just apply the Stale label to issues in open state for 30 days not having any activity**Which issue(s) this PR fixes**:

Closes #

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,35 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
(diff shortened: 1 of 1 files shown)


gongmax (commented):
nit: might use .yaml instead of .yml to keep consistency.

Will this touch the existing issues or only new ones after this is enabled? There are bunch of existing issues older than 30 days. I don't think we should simply close them without any manual audit.

geetachavan1:
@gongmax , I will change the name to .yaml. This PR will not close any issues but will mark them as 'stale', + this is marked as dry run so will not label any issue but will just generate logs which can be verified.. But anyways I will close this PR at this moment and do some test, publish the results to you all and then recreate the same PR

Issue #2954, opened 2023-02-07
Node.js SDK test is flaky


zmerlynn:
[...]

PR #2955, opened 2023-02-07
Allow occasional flakes in SDK conformance tests

zmerlynn:
Most of the churn is outside the SDK conformance tests, yet they have an outmoded impact on the flakes. Let's go ahead and retry these as well.

This is a bandaid for #2954 - but I'd like to set up some flake reporting at some point as well.


markmandel (approved):
Makes sense to me.

PR #2963, opened 2023-02-10
Update the apiserver request dashboard

zmerlynn:
This is nearly unreviewable, so here's a screenshot:

image

Changes:

* The original metrics were deprecated. `apiserver_request_total` exactly replaces `apiserver_request_count`, but the old `apiserver_request_latencies_summary`, which was a gauge, was replaced by `apiserver_request_duration_seconds_bucket`, which is a historgram type. Changed the expressions to match these changes.

* Transitioned to $__rate_interval instead of 5m, per recommendation: https://grafana.com/docs/grafana/latest/datasources/prometheus/template-variables/#use-__rate_interval

* Fixed an issue where `label_values(label)` isn't supported on GMP (it seems - probably because this goes across ALL metrics), but `label_values(label, metric)` is - this is a change I'll need to make on nearly every other dashboard but I'll get there later.

* Upgraded the graph from a deprecated graph type to the new time series type.

I created this PR by using "Save", then copying the JSON blob back to the configmap. I then compensated for the new "fixed" UID datasource by applying this hack: https://github.com/grafana/grafana/discussions/45230#discussioncomment-4088203 .

### DIFF
--- a/build/grafana/dashboard-apiserver-requests.yaml
+++ b/build/grafana/dashboard-apiserver-requests.yaml
@@ -27,20 +27,29 @@ data:
         "list": [
           {
             "builtIn": 1,
-            "datasource": "-- Grafana --",
+            "datasource": {
+              "type": "datasource",
+              "uid": "grafana"
+            },
             "enable": true,
             "hide": true,
             "iconColor": "rgba(0, 211, 255, 1)",
             "name": "Annotations & Alerts",
+            "target": {
+              "limit": 100,
+              "matchAny": false,
+              "tags": [],
+              "type": "dashboard"
+            },
             "type": "dashboard"
           }
         ]
       },
       "editable": true,
-      "gnetId": null,
+      "fiscalYearStartMonth": 0,
       "graphTooltip": 0,
-      "iteration": 1551904655558,
       "links": [],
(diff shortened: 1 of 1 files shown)


build/grafana/dashboard-apiserver-requests.yaml
```diff
@@ -74,27 +85,31 @@ data:
           "linewidth": 1,
           "links": [],
           "nullPointMode": "null",
+          "options": {
+            "alertThreshold": true
+          },
           "paceLength": 10,
           "percentage": false,
+          "pluginVersion": "9.3.6",
```
gongmax, line 93: Just for my learning purpose, why the `pluginVersion` is not explicitly set previously?

gongmax (approved):

build/grafana/dashboard-apiserver-requests.yaml
zmerlynn: Like I said, this was made by "save" in the UI then copying the JSON blob. I have no idea why some of these values changed - I suspect at some point they extended the schema to include the plugin version so you could "bottle" charts like this.

markmandel (approved):

PR #2965, opened 2023-02-13
Updates counters lists schema status CRDs for gameservers fleets

igooch:
Updates CRDs per comments on Arbitrary Counts and Lists Feature/CRD #2946. 

**What type of PR is this?**

/kind feature

Updates schemas for Counts and Lists in the gamerserver, gamerserverset, and fleet CRDs.
Validates Counts and Lists in the gamerserver, gamerserverset, and fleet CRDs.

Working on #2716


pkg/apis/agones/v1/gameserver.go
```diff
@@ -438,6 +451,23 @@ func (gss *GameServerSpec) Validate(apiHooks APIHooks, devAddress string) ([]met
 		}
 	}
 
+	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
```
markmandel, line 454: To avoid the cyclic complexity, I'd suggest pulling this into it's own helper function.

examples/gameserver.yaml
```diff
@@ -88,15 +88,19 @@ spec:
   #
   # [Stage:Alpha]
   # [FeatureFlag:CountsAndLists]
-  # Lists provides the configuration for generic (player, room, session, etc.) tracking features.
+  # Counts and Lists provides the configuration for generic (player, room, session, etc.) tracking features.
   # Commented out since Alpha, and disabled by default
+  # counters:
+  #   gameCounter:
```
markmandel, line 94: Nit:

```suggestion
  #   games
```
Adding "Counter" the name seems redundant to me, since it's already under `counters`. WDYT?

Same with `playerList` below.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -187,6 +187,10 @@ type GameServerSpec struct {
 	// (Alpha, PlayerTracking feature flag) Players provides the configuration for player tracking features.
 	// +optional
 	Players *PlayersSpec `json:"players,omitempty"`
+	// (Alpha, CountsAndLists feature flag) Counters and Lists provides the configuration for generic tracking features.
+	// +optional
+	Counters *CountersSpec `json:"counters,omitempty"`
```
markmandel, line 192: Shouldn't this be?

```suggestion
	Counters map[string]*CountersSpec `json:"counters,omitempty"`
```

pkg/apis/agones/v1/gameserver.go
```diff
@@ -198,6 +202,15 @@ type PlayersSpec struct {
 	InitialCapacity int64 `json:"initialCapacity,omitempty"`
 }
 
+// CountersSpec tracks if counter specified (for giving error message if feature gate not set)
+type CountersSpec struct {
+}
```
markmandel, line 207: `Default` and `Capacity` ?

pkg/apis/agones/v1/gameserver.go
```diff
@@ -187,6 +187,10 @@ type GameServerSpec struct {
 	// (Alpha, PlayerTracking feature flag) Players provides the configuration for player tracking features.
 	// +optional
 	Players *PlayersSpec `json:"players,omitempty"`
+	// (Alpha, CountsAndLists feature flag) Counters and Lists provides the configuration for generic tracking features.
+	// +optional
+	Counters *CountersSpec `json:"counters,omitempty"`
+	Lists    *ListsSpec    `json:"lists,omitempty"`
```
markmandel, line 193:
```suggestion
	Lists    map[string]*ListsSpec    `json:"lists,omitempty"`
```
?

markmandel (commented):
Looking much better! Some questions for you.

pkg/apis/agones/v1/gameserver.go
igooch: The default for count is applied by the CRD automatically if no value for count is given, so default is not a settable but the count (which automatically overrides the default) is. 

As far as what to include is this struct -- is there a reason why capacity is needed here if the CRD already validates that it's an integer >= 0?

markmandel: If there is no values to store `Default` or `Count` user set values into, then you can't access that. The incoming and outgoing data for the CRD need somewhere to serialise and deserialise to and from to.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -198,6 +202,15 @@ type PlayersSpec struct {
 	InitialCapacity int64 `json:"initialCapacity,omitempty"`
 }
 
+// CountersSpec tracks if counter specified (for giving error message if feature gate not set)
+type CountersSpec struct {
```
gongmax, line 206: nit: should this be `CounterSpec`? since it represents one Counter. Similar for `ListsSpec`

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -155,33 +155,32 @@ properties:
...
       counters:
-        type: array
-        title: Array of player, room, session, etc. counters
+        type: object
+        title: Map of player, room, session, etc. counters
         nullable: true
-        maxItems: 1000
-        items:
+        maxProperties: 1000
+        additionalProperties:
```
markmandel, line 162: I just realised we are also missing the `default` option here (I hope that's not a keyword in the OpenAPI spec, so that users can specify where they want a count to start at (by default, the default should be 0).

For example:

```yaml
  counters: # list of counters. Key value below is the key for each counter.
    rooms: # key for the counter (room)
      default: 1 # initial value
      capacity: 100 # maximum possible count value
```

igooch: Default is a key word (and currently set to 0 for counters). The user can set a number for Count as the initial value, and if they don't specify the Count it will default to 0.

igooch: Would it make sense to switch up the name in the spec to initialCount to differentiate from status count?

markmandel: Ahhh got it, so we actually end up at:

 ```yaml
   counters: # list of counters. Key value below is the key for each counter.
>     rooms: # key for the counter (room)
       count: 1 # initial value
       capacity: 100 # maximum possible count value
 ```

That works! SGTM!

markmandel: > Would it make sense to switch up the name in the spec to initialCount to differentiate from status count?

🤔 I'm okay with it the way it is I think? WDYT?

igooch: With rooms as an object with values count and capacity, yes.
```
counters: # list of counters. Key value below is the key for each counter.
  rooms: # key for the counter (room)
    count: 1 # initial value
    capacity: 100 # maximum possible count value
```

markmandel: My tabbing was off - but that looks good to me 👍🏻

markmandel (approved):
🥳

PR #2968, opened 2023-02-14
Run e2e tests on multiple clusters with different versions

gongmax:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Closes #


cloudbuild.yaml
```diff
@@ -291,27 +291,32 @@ steps:
       set -o pipefail
       pids=()
       cloudProducts=("generic" "gke-autopilot")
+      k8sVersions=("1.23" "1.24" "1.25")
+      clusterRegions=("us-east1" "us-west1" "us-central1")
```
roberthbailey, line 295: When we add support for 1.26 and drop 1.23, if we don't shift this array to keep it aligned with the k8sVersions array then we are going to shift the region in which we are looking for the test cluster location. If these are going to be tightly coupled, then I wonder if we can stick them into the same data structure as pairs (e.g. `({version1, region1}, {version2, region2}, ...)`). WDYT?

cloudbuild.yaml
```diff
@@ -291,27 +291,32 @@ steps:
...
         do
+          if [ $cloudProduct = gke-autopilot ] && [ ${k8sVersions[$i]} = 1.23 ]
+          then
+            continue
+          fi
+          if [ $cloudProduct = generic ]
+          then
+            featureWithGate="CustomFasSyncInterval=false&SafeToEvict=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&PodHostname=true&SplitControllerAndExtensions=true&FleetAllocationOverflow=true&Example=true"
+            featureWithoutGate=""
+            testClusterLocation="${clusterRegions[$i]}-c"
```
roberthbailey, line 309: We should consider making the regular tests use regional clusters too. Maybe not as part of this PR, but something to consider since there is no cost difference between zonal & regional.

cloudbuild.yaml
gongmax: Totally makes sense. I'm making them pairs.

cloudbuild.yaml
gongmax: Sounds good. I can make the change in the other PR where the test clusters are defined, and have a following up PR to change the CI steps once the clusters are created.

cloudbuild.yaml
```diff
@@ -291,27 +291,32 @@ steps:
...
+          if [ $cloudProduct = gke-autopilot ] && [ $version = 1.23 ]
+          then
+            continue
+          fi
+          if [ $cloudProduct = generic ]
+          then
+            featureWithGate="CustomFasSyncInterval=false&SafeToEvict=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=true&PodHostname=true&SplitControllerAndExtensions=true&FleetAllocationOverflow=true&Example=true"
+            featureWithoutGate=""
+            testClusterLocation="${region}-c"
+            testCluster="gke-standard-e2e-test-cluster-${version//./-}"
```
roberthbailey, line 310: nit: Maybe something to consider in the future would be to name the clusters as `"${cloudProduct}-e2e-test-cluster-${version//./-}"`. So we'd have names like `generic-e2e-test-cluster-1-24` and `gke-autopilot-e2e-test-cluster-1-24`. 

I'm thinking that once a couple of feature gates get promoted to beta, we won't need the if/else here for feature gates based on the provider, and if we move to regional standard clusters then the only difference remaining is how we construct the test cluster names.

roberthbailey (approved):

roberthbailey:
Here we go.....

Issue #2682, opened 2022-07-26
GameServers in allocated state can respond to fleet updating


keisni:
**Is your feature request related to a problem? Please describe.**
a question about fleet update. it seems that the controller leaves GameServers in allocated state alone in the process of fleet updating, whether rolling or recreating. but some of our gameservers have long ttls (they are not specific to battles, but players can enter and leave at any time). so i`m wondering how to tell these gameservers to start the shutting down procedure (start by telling players to leave game or just kicking them out).

**Describe the solution you'd like**
We could set an annotation / label (I think an annotation would be better?) on an Allocated GameServer that is left behind after a Fleet update - then the WatchGameServer SDK could pick it up?

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

markmandel:
Super interesting problem! We were discussing this a bit internally and came up with some other user journeys. Would love feedback on the below design.

---



# Objective

In a usage of Agones in which `Allocated` `GameServers` are utilised for a long time, such as a Lobby `GameServer`, or potentially a session based `GameServer` that is being reused multiple times in a row, it would be useful if that `GameServer` process could be notified in some way if and when a Fleet has been updated such that an `Allocated` `GameServer`, may want to actively perform a graceful shutdown and release its resources such that it can be replaced by a new version.


# Requirements

There are two scenarios of which to consider in which we might want to notify a `GameServer` that a `Fleet` has been updated that will need to have a solution:


## Fleet Update (Rolling/Recreate)

This is a standard [Fleet Rolling Update](https://agones.dev/site/docs/guides/fleet-updates/#rolling-update-strategy). When a rolling update starts, there should be an ability to notify any `Allocated` `GameServers` within that `Fleet` that a rolling update has begun.

Out of scope: Notifying GameServers when a rolling update has finished.


## Canary Deploy Through Multiple Fleets

If the user is updating a Fleet through a “[Two (or more) Fleets Strategy](https://agones.dev/site/docs/guides/fleet-updates/#two-or-more-fleets-strategy)” to canary a new version there is no rolling update, but instead the current Fleet is manually scaled down, while the replacement Fleet is manually scaled up.

In this scenario, the only indication that a Fleet is being replaced, is that the Fleet replica count could drop below the number of `Allocated` `GameServers`. In this scenario, the correct number to notify would equal:

Total Allocated GameServers - Fleet Replica Count

[...]

PR #2971, opened 2023-02-15
Fix for `make gen-crd-client`

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Having `agonesv1.FakeApiHooks` be public means that the code generation toolkit will see it as a generation target, and then get all confused.

Fix is to make it private, and keep it's stubbable capabilities.

Since we also need a mock for other parts of the Agones system, I also created a no-op version for general usage as well in the `testing` package. (We can't use this one in `agonesvs`, since Go will complain about cyclic dependencies).

Blocking work on https://github.com/googleforgames/agones/issues/2682

Not sure if the stubs actually get used anywhere, but figured I would leave them in as that was the original code.


pkg/apis/agones/v1/apihooksfake.go
```diff
@@ -20,42 +20,43 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// FakeAPIHooks is a stubabble, fake implementation of APIHooks
-type FakeAPIHooks struct {
+// fakeAPIHooks is a stubabble, fake implementation of APIHooks
+// This needs to be private, so it doesn't get picked up by the DeepCopy() generation toolkit.
+type fakeAPIHooks struct {
```
roberthbailey, line 25: This struct is now duplicated here and in `pkg/testing`. I don't think there are any dependencies in `pkg/testing` that prevent it from being included into this package, so can we just move this file there (and remove it from here) so that it isn't duplicated?

markmandel: I also figured (possibly incorrectly?) that stubbing was only really necessary for `pkg/apis/agones/v1` - so a generic no-op version that was shareable seemed to make the most sense.

roberthbailey: Thanks. @zmerlynn - can you take a look when you get a chance to see if there is a better way to do this without code duplication? Until then, merging this to get the make target fixed seems appropriate.

PR #2980, opened 2023-02-17
Kubernetes Update template, Release template, and site content change for multiple k8s version support

gongmax:
…for multiple k8s version support

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Closes #


docs/governance/kubernetes_update_process.md
```diff
@@ -0,0 +1,10 @@
+
+# Kubernetes versions update policy
+
+We will support 3 releases of Kubernetes, targeting the newest version as being the version available in the GKE rapid channel (since tests run on GKE we can't support versions that GKE doesn't support). However, we will ensure that at least one of the 3 versions chosen for each Agones release is supported by each of the major cloud providers (EKS and AKS). The vendored version of client-go will be aligned with the middle of the three supported Kubernetes versions.
```
zmerlynn, line 4: Same as above, I'd link the external docs so it's obvious.

site/content/en/docs/Installation/_index.md
```diff
@@ -45,17 +43,17 @@ Each version of Agones supports a specific version of Kubernetes. When a new ver
...
-| 1.27           | 1.23               |
-| 1.26           | 1.23               |
-| 1.25           | 1.22               |
-| 1.24           | 1.22               |
-| 1.23           | 1.22               |
-| 1.22           | 1.21               |
-| 1.21           | 1.21               |
+| Agones version | Kubernetes version   |
+| -------------- | ------------------   |
+| 1.29           | {{% k8s-version %}}  |
```
zmerlynn, line 48: Nit: This just renders as `[1.23 1.24 1.25]`, do we want to do any better? Like commas or anything? I dunno, I'm not sure it's worth it.

Similarly do we want consistency through the rest of the table?

zmerlynn (approved):

site/content/en/docs/Installation/_index.md
gongmax: Good call out. Did some trick to make it looks nicer.

markmandel:
FYI: @geetachavan1 and @Kalaiselvi84

gongmax:
E2E test failed
```
time="2023-02-23 07:44:57.342" level=info msg="Creating default game server" test=TestGameServerHealthyAfterDeletingPodWhileOneExtensionsDown
    high_availability_test.go:55: 
        	Error Trace:	/go/src/agones.dev/agones/test/e2e/extensions/high_availability_test.go:55
        	            	creating {game-server [{udp-port Dynamic <nil> 7654 0 UDP}] {false 0 0 0}  { 0 0} {{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] [] [{game-server us-docker.pkg.dev/agones-images/examples/simple-game-server:0.14 [] []  [] [] [] {map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}] map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}]} [] [] nil nil nil nil   IfNotPresent nil false false false}] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] [] <nil> nil}} <nil> map[] map[] {}} GameServer instances failed (): Internal error occurred: failed calling webhook "mutations.agones.dev": failed to call webhook: Post "https://agones-controller-service.agones-system.svc:443/mutate?timeout=10s": EOF
        	Test:       	TestGameServerHealthyAfterDeletingPodWhileOneExtensionsDown
        	Messages:   	Could not get a GameServer ready
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x20f82b0]
```

.github/ISSUE_TEMPLATE/kubernetes_update.md
```diff
@@ -1,38 +1,51 @@
...
-    - [ ] Minikube
-    - [ ] Kind
-- [ ] Update the k8s image used in the helm [pre-delete-hook](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/hooks/pre_delete_hook.yaml)
-- [ ] Update client-go
-- [ ] Update CRD API reference
+- [ ] Update the Kubernetes version of the below test clusters to {version_2}
+    - [ ] Minikube (Get the patch version [here](https://github.com/kubernetes/minikube/blob/master/CHANGELOG.md))
+    - [ ] Kind (Confirm {version_2} is supported and get the patch version [here](https://github.com/kubernetes-sigs/kind/releases))
+- [ ] Update the k8s image used in the helm [pre-delete-hook](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/hooks/pre_delete_hook.yaml) to {version_2}
+- [ ] Update client-go in `go.mod` and `test/terraform/go.mod` to {version_2} and re-run `go mod tidy` and `go mod vendor`
```
roberthbailey, line 25: nit: I always have to remind myself which command to run to update client go (since we shouldn't be editing `go.mod` directly). Can you insert the command here since you've added the `go mod tidy` command?

site/content/en/docs/Installation/_index.md
```diff
@@ -22,9 +22,7 @@ description: >
 - Game Servers must have the [game server SDK]({{< ref "/docs/Guides/Client SDKs/_index.md"  >}}) integrated, to manage Game Server state, health checking, etc.
 
 {{< alert title="Warning" color="warning">}}
-Later versions of Kubernetes may work, but this project is tested against {{% k8s-version %}}, and is therefore the supported version.
-Agones will update its support to the n-1 version of what is available across the majority of major cloud providers - GKE, EKS and
-AKS, while also ensuring that all Cloud providers can support that version.
+This release has been tested against Kubernetes versions {{% k8s-version %}}. Other versions may work, but are unsupported.
```
roberthbailey, line 25: I'm wondering if above (where we say that GKE, AKS, and EKS are supported) we should give a caveat that not all cloud providers may support all versions of Kubernetes.

roberthbailey, line 25: On this line, would it make sense to say "has been tested against Kubernetes versions ... on GKE." to make it clear? @markmandel

site/content/en/docs/Installation/_index.md
```diff
@@ -45,17 +43,17 @@ Each version of Agones supports a specific version of Kubernetes. When a new ver
```
roberthbailey, line 43: "Each version support a specific version of k8s" (line 42) should be updated since now a version will support more than one.

site/content/en/docs/Installation/_index.md
roberthbailey, line 48: This needs to be behind a release guard so that it doesn't get published when your PR merges. It should be added in as part of the 1.30 release (since we aren't retroactively adding support for new versions to 1.29).

site/content/en/docs/Installation/_index.md
```diff
@@ -22,9 +22,7 @@ description: >
 - Game Servers must have the [game server SDK]({{< ref "/docs/Guides/Client SDKs/_index.md"  >}}) integrated, to manage Game Server state, health checking, etc.
 
 {{< alert title="Warning" color="warning">}}
-Later versions of Kubernetes may work, but this project is tested against {{% k8s-version %}}, and is therefore the supported version.
-Agones will update its support to the n-1 version of what is available across the majority of major cloud providers - GKE, EKS and
```
roberthbailey, line 26: You've moved the policy from the website into github. I think the website should either maintain the policy here or it should link to the policy in github. As it is, folks that only use the website will no longer be able to find our upgrade policy.

site/content/en/docs/Installation/_index.md
```diff
@@ -45,17 +43,17 @@ Each version of Agones supports a specific version of Kubernetes. When a new ver
...
-| 1.29           | 1.24               |
-| 1.28           | 1.23               |
-| 1.27           | 1.23               |
-| 1.26           | 1.23               |
-| 1.25           | 1.22               |
-| 1.24           | 1.22               |
-| 1.23           | 1.22               |
-| 1.22           | 1.21               |
-| 1.21           | 1.21               |
+| Agones version | Kubernetes version   |
```
roberthbailey, line 46: Starting with the next release we should change the column header to Kubernetes versions (plural).

.github/ISSUE_TEMPLATE/kubernetes_update.md
gongmax: Oh! I manually updated the client-go version last time! I will dig more and add the command here.

site/content/en/docs/Installation/_index.md
gongmax: I think it should be fine.
After my PR merges, {{% k8s-version %}} is still 1.24 for production environment. For 1.30 release, I explictly added the release step to modify this part:
"for the current version, replace `{{% k8s-version %}}` with hardcoded Kubernetes versions supported by the current version. And add a row for the Agones release version with `{{% k8s-version %}}` as its supported Kubernetes versions"

site/content/en/docs/Installation/_index.md
gongmax: I was following the proposal in [go/cgtn235](https://goto.google.com/cgtn235) which suggested to remove the policy from the user facing website. But I'm ok to add it back.

roberthbailey: Funny how I wrote that 18 months ago and had zero recollection of it.... Unless anyone objects that works for me.

site/content/en/docs/Installation/_index.md
roberthbailey: That makes sense.

roberthbailey (approved):

docs/governance/kubernetes_update_process.md
```diff
@@ -0,0 +1,10 @@
+
+# Kubernetes versions update policy
+
+We will support 3 releases of Kubernetes, targeting the newest version as being the [default version in the GKE Rapid channel](https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions) (since tests run on GKE we can't support versions that GKE doesn't support). However, we will ensure that at least one of the 3 versions chosen for each Agones release is supported by each of the major cloud providers (EKS and AKS). The vendored version of client-go will be aligned with the middle of the three supported Kubernetes versions.
```
markmandel, line 4: This paragraph is redundant to that which is in the docs, is it necessary to have and maintain two copies?

docs/governance/kubernetes_update_process.md
```diff
@@ -0,0 +1,10 @@
...
+
+# Kubernetes versions update process
```
markmandel, line 6: Should this be part of the contributing guide (near the release section maybe?)

Then it's not floating in a file that's not linked to anything.

gongmax: Sounds good. Moved it to the CONTRIBUTING.md and removed this file.

markmandel (approved):

PR #2982, opened 2023-02-18
Make Alpha & Beta CRD declaration consistent.

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

We realised that with the CRD schema application being hidden behind a feature flag in the helm CRD yaml, then if the feature flag was turned off, an end user wouldn't get a validation error since the values passed in would be pruned by the K8s control plane.

To make thins consistent across Alpha CRD values:

* Alpha & Beta schemas are always available within the CRD definitions.
* Validation checks are in place for the feature flag (no change needed)
* Alpha & Beta CRD values are pointers, so they can be nullable (this required a change in Eviction handling and ApiHooks interface)
* Some general cleanup of comments and descriptions in the schema.

### DIFF
--- a/install/helm/agones/templates/crds/_gameserverspecschema.yaml
+++ b/install/helm/agones/templates/crds/_gameserverspecschema.yaml
@@ -153,7 +153,6 @@ properties:
             type: integer
             title: The initial player capacity of this Game Server
             minimum: 0
-{{- if .featureCountsAndLists }}
       counters:
         type: object
         title: Map of player, room, session, etc. counters
@@ -162,11 +161,13 @@ properties:
         additionalProperties:
           type: object
           properties:
-            count: # initial count
+            count:
+              title: Initial count value
               type: integer
               default: 0
               minimum: 0
-            capacity: # max capacity of the counter
+            capacity:
+              title: Max capacity of the counter
               type: integer
               minimum: 0
       lists:
@@ -177,18 +178,18 @@ properties:
         additionalProperties:
           type: object
           properties:
-            capacity: # max capacity of the array (can be less than or equal to value of maxItems)
+            capacity:
               type: integer
+              title: Max capacity of the array (can be less than or equal to value of maxItems)
               minimum: 0
               maximum: 1000 # must be equal to values.maxItems
             values:
+              title: set of all the items in the list
               type: array
               x-kubernetes-list-type: set # Requires items in the array to be unique
               maxItems: 1000 # max possible size of the value array (cannot be updated)
               items: # name of the item (player1, session1, room1, etc.)
                 type: string
-{{- end }}
-{{- if .featureSafeToEvict }}
       eviction:
         type: object
         title: Eviction tolerance of the game server
@@ -210,5 +211,4 @@ properties:
         default: 1
         minimum: 1
         maximum: 1
-{{- end }}
 {{- end }}
--- a/pkg/apis/agones/v1/apihooksfake.go
+++ b/pkg/apis/agones/v1/apihooksfake.go
@@ -26,7 +26,7 @@ type fakeAPIHooks struct {
 	StubValidateGameServerSpec  func(*GameServerSpec) []metav1.StatusCause
 	StubValidateScheduling      func(apis.SchedulingStrategy) []metav1.StatusCause
 	StubMutateGameServerPodSpec func(*GameServerSpec, *corev1.PodSpec) error
-	StubSetEviction             func(EvictionSafe, *corev1.Pod) error
+	StubSetEviction             func(*Eviction, *corev1.Pod) error
 }
 
 var _ APIHooks = fakeAPIHooks{}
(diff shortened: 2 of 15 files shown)


pkg/apis/agones/v1/apihooksfake.go
```diff
@@ -56,9 +56,9 @@ func (f fakeAPIHooks) MutateGameServerPodSpec(gss *GameServerSpec, podSpec *core
 }
 
 // SetEviction is called by gs.Pod to enforce GameServer.Status.Eviction.
-func (f fakeAPIHooks) SetEviction(safe EvictionSafe, pod *corev1.Pod) error {
+func (f fakeAPIHooks) SetEviction(eviction *Eviction, pod *corev1.Pod) error {
```
markmandel, line 59: @zmerlynn I made this change to handle Eviction being a pointer. PTAL and let me know what you think of the approach.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -177,18 +178,18 @@ properties:
...
           type: object
           properties:
-            capacity: # max capacity of the array (can be less than or equal to value of maxItems)
+            capacity:
               type: integer
+              title: Max capacity of the array (can be less than or equal to value of maxItems)
               minimum: 0
               maximum: 1000 # must be equal to values.maxItems
             values:
+              title: set of all the items in the list
```
markmandel, line 187: @igooch PTAL at my changes here as well.

Also, not for this PR, in the original design the user couldn't set a default list value (like you can with counters) -- but it could be useful! Just wanted to highlight this to check that this was on purpose.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -293,7 +293,7 @@ type GameServerStatus struct {
 	// +optional
 	Players *PlayerStatus `json:"players"`
 	// (Alpha, SafeToEvict feature flag) Eviction specifies the eviction tolerance of the GameServer.
-	Eviction Eviction `json:"eviction,omitempty"`
+	Eviction *Eviction `json:"eviction,omitempty"`
```
markmandel, line 296: @zmerlynn just noting I made this a pointer to be consistent with all other Alpha/Beta properties added to a CRD.

(I can't find the best practice doc I originally referenced for this approach.)

install/helm/agones/templates/crds/_gameserverspecschema.yaml
igooch: Yes, they can set/populate all values to start (if the user wants).

markmandel: Neat! SGTM!

pkg/apis/agones/v1/gameserver.go
zmerlynn: Hmm, I saw that the others were but actually didn't understand why. It makes sense if they're `nullable`, but this field doesn't need to be nullable, I don't think.

markmandel: Aha, found the reference I was looking for:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

Having it be a non pointer means that if the feature flag is off, there is still some kind of data stored in etcd (probably a lot of empty strings, since that's the default). Whereas if it's a pointer, there is nothing to serialise/deserialise.

markmandel: Acctually, this is an optional field no -- I mean, the end user doesn't _have_ to set it? So we should probably mark it as such.

zmerlynn: Yes, I was just about to point out that by those conventions, we should add the `+optional` as well. But the concept LGTM - will give this a more thorough review in a bit (catching up).

markmandel: All good - no rush. I'll add in the `+optional` here as well.

zmerlynn (approved):

pkg/apis/agones/v1/apihooksfake.go
zmerlynn: LGTM

pkg/apis/agones/v1/gameserver.go
zmerlynn: I approved but would like to see this before it goes in. :)

markmandel: Done!

markmandel:
Oh no, too many concurrent builds, never became the older PR in queue.

```
Finished Step #21 - "e2e-wait-to-become-leader"
ERROR
ERROR: build step 21 "gcr.io/cloud-builders/gcloud" failed: context deadline exceeded
```

PR #2985, opened 2023-02-22
Update Agones GKE quickstart with an Autopilot cluster

shannonxtreme:
Updates the agones.dev GKE documentation with an Autopilot cluster instead of Standard.

Autopilot clusters are GKE's recommended mode for production workloads outside of specific niche needs, so updating the Agones docs to use it is a sensible change.

Changes: 

- Minor style changes like "Google Cloud Platform" to "Google Cloud"
- Word wrapping for readability
- Zone to region because Autopilot is a regional cluster
- Remove Standard create cluster and change flag explanations to match updated command (l83-108)
- Remove dedicated node pool instructions, because no node pools in Autopilot (l108-122)
- Update Windows instructions to explicitly use GKE Standard with dedicated node pools

/cc @zmerlynn 
/kind documentation

### DIFF
--- a/site/content/en/docs/Installation/Creating Cluster/gke.md
+++ b/site/content/en/docs/Installation/Creating Cluster/gke.md
@@ -11,11 +11,13 @@ description: >
 
 Take the following steps to enable the Kubernetes Engine API:
 
-1. Visit the [Kubernetes Engine][kubernetes] page in the Google Cloud Platform Console.
+1. Visit the [Kubernetes Engine][kubernetes] page in the Google Cloud console.
 1. Create or select a project.
-1. Wait for the API and related services to be enabled. This can take several minutes.
+1. Wait for the API and related services to be enabled. This can take several
+   minutes.
 1. [Enable billing][billing] for your project.
-  * If you are not an existing GCP user, you may be able to enroll for a $300 US [Free Trial][trial] credit.
+
+   If you are not an existing Google Cloud user, you may be able to enroll for a $300 US [Free Trial][trial] credit.
 
 [kubernetes]: https://console.cloud.google.com/kubernetes/list
 [billing]: https://support.google.com/cloud/answer/6293499#enable-billing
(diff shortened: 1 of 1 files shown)


markmandel (commented):
I would argue that we should have a guide for GKE standard, and a separate page for GKE autopilot.

Any reason not to?

shannonxtreme:
Yes, there are a couple: 

 - An Autopilot cluster has fewer moving parts right off the bat, so it simplifies the user journey at this stage (the "quickstart" phase)
 - We should provide a single path to success instead of branching options at this stage. Having separate pages would mean that the reader right away has to wonder which mode they should start with.

I think we can instead add, after the create cluster step, a note saying that you can do this with a GKE standard cluster as well by following the Windows instructions but using Linux images instead. WDYT? Ideally this should be as straightforward as possible though

zmerlynn:
We could also maybe change the Windows instructions to be the Standard instructions, with a separate callout for Windows. Or bump it out to an Advanced section (which is odd in a Quick Start).

zmerlynn (commented):
Whatever we do, the Autopilot pieces need to be release guarded: https://agones.dev/site/docs/contribute/#documentation-for-upcoming-features

markmandel:
> I think we can instead add, after the create cluster step, a note saying that you can do this with a GKE standard cluster as well by

Given then Autopilot support is Alpha, do we want to be making the optimal path away from the GKE standard that is the stable solution by putting Autopilot first?

Also as a note - do we need to tie the Helm installation params for 1.30.0 to this document, since they differ from the standard installation at this time?

I would also advocate that we have tradeoffs/considerations section somewhere here for cluster type selection options.

zmerlynn:
@shannonxtreme I agree with @markmandel's point here. I think for the time being we need the branch, and it should probably stay GKE Standard first since we're still relying on an Alpha Agones feature to accomplish it. After, we can flip the order and figure out how to make it clearer when you _might_ need Standard.

To that vein, maybe something like:
* Change "## Creating a Cluster" to "## Creating a GKE Standard Cluster"
* then adding a section below of "## Creating a GKE Autopilot Cluster", in which we link out to the Autopilot docs (which already discuss why you might choose Autopilot).

Later, after we promote the currently Alpha feature gate we're relying on, I'd propose we flip the order and do a little more refactoring.

shannonxtreme:
I didnt know that it was using an alpha feature gate, my bad. I think we should leave it as-is for now and maybe add a note saying "You can also deploy Agones on a GKE Autopilot cluster in Alpha. For details on why you might choose Autopilot, refer to <link>". 

And when we do promote to stable, we can use the content from this PR and change the note to "You can also deploy Agones on a GKE Standard cluster. For details on why you might choose Standard, refer to <link>". WDYT?  IDK if it's necessary to have full instructions for both modes in the quickstart (but im open to having my mind changed)

markmandel:
Just a point of concern -- we keep using the phrase "quickstart". This section isn't really a quickstart, it's the docs on how to  create a GKE cluster for Agones. There is no other documentation on this -- so in case we are thinking of this doc as the fast onboarding path, let's not do that please.

shannonxtreme:
Ack. So then is it ok to have full instructions for both modes on the page? I think maybe I can structure it like 

- Create a firewall rule
- Create a cluster
  - Autopilot instructions plus a link to the next step, installing Agones
  - Existing Standard instructions + the existing optional instructions for dedicated node pools and windows

Wdyt?

markmandel:
> with a link to why choose it

I think the link is good. We also had some great internal content on Autopilot vs GKE for Agones workloads I think we could also leverage.

shannonxtreme:
I'll close this PR but keep the branch open. When AP support is stable at some point I'll reopen with updates.

PR #2988, opened 2023-02-22
Added back metrics support for extensions

chiayi:
Add back metrics support within extensions

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature

/kind hotfix

This adds back the metrics support within `Extensions` that was removed during the trimming process.

### DIFF
--- a/cmd/extensions/main.go
+++ b/cmd/extensions/main.go
@@ -33,6 +33,7 @@ import (
 	"agones.dev/agones/pkg/gameserverallocations"
 	"agones.dev/agones/pkg/gameservers"
 	"agones.dev/agones/pkg/gameserversets"
+	"agones.dev/agones/pkg/metrics"
 	"agones.dev/agones/pkg/util/apiserver"
 	"agones.dev/agones/pkg/util/https"
 	"agones.dev/agones/pkg/util/runtime"
@@ -40,6 +41,7 @@ import (
 	"agones.dev/agones/pkg/util/webhooks"
 	"github.com/heptiolabs/healthcheck"
 	"github.com/pkg/errors"
+	prom "github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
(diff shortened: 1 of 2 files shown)


cmd/extensions/main.go
```diff
@@ -139,8 +144,44 @@ func main() {
...
+		if err != nil {
+			logger.WithError(err).Fatal("Could not register stackdriver exporter")
+		}
+		// It is imperative to invoke flush before your main function exits
+		defer sd.Flush()
+	}
+
+	// Prometheus metrics
+	if ctlConf.PrometheusMetrics {
+		logger.Info("registering prometheus metrics...")
```
zmerlynn, line 162: let's take the logging back out and make this symmetric with controller

cmd/extensions/main.go
```diff
@@ -139,8 +144,44 @@ func main() {
...
 
-	health := healthcheck.NewHandler()
+	// If we are using Prometheus only exporter we can make reporting more often,
+	// every 1 seconds, if we are using Stackdriver we would use 60 seconds reporting period,
+	// which is a requirements of Stackdriver, otherwise most of time series would be invalid for Stackdriver
+	metrics.SetReportingPeriod(ctlConf.PrometheusMetrics, ctlConf.Stackdriver)
+
+	// Add metrics controller only if we configure one of metrics exporters
+	if ctlConf.PrometheusMetrics || ctlConf.Stackdriver {
+		rs = append(rs, metrics.NewController(kubeClient, agonesClient, kubeInformerFactory, agonesInformerFactory))
```
zmerlynn, line 183: There's a difference between exporting metrics about this process, which is what line 160-174 are doing, and exporting metrics about the fleet / etc. The "metrics controller" is doing the latter, and I don't think we need two instances of it running, so I'm _pretty sure_ you can safely kill 182-184 here, which means you also don't need `rs`.

zmerlynn (changes requested):

cmd/extensions/main.go
chiayi: done.

cmd/extensions/main.go
chiayi: okay! Done!

zmerlynn (approved):

PR #2997, opened 2023-02-24
Unreal Engine SDK: Add HttpHost variable to configure host in case it's not localhost

ViktorWii:
> /kind feature

**Replace hardcoded 127.0.0.1 with variable, so it's possible to configure where Agones should connect to**:


roberthbailey:
/cc @markmandel 

A similar change was proposed in https://github.com/googleforgames/agones/issues/1183 and we did not move forward with that proposal. 

I'll ask a similar question here - can you help me understand what the scenario is for needing to connect to a remote SDK? Is it for development? Are you deploying pods without the sdk sidecar into a cluster?

ViktorWii:
Well, currently our approach is that we have cluster set up on remote machine, and in-editor session connects to specified IP to mock/replace remote agones with itself. That way we can test things both on game and backend without rebuilding server after each change and redeploying it to backend

ViktorWii:
So the answer is yes, it's just for development purposes. 
It's fare to say that we could potentially host cluster locally but there is few catches that makes it difficult and most importantly inefficient

markmandel:
Does this approach solve your issue instead?

https://agones.dev/site/docs/guides/local-game-server/

ViktorWii:
If I'm understand correctly, it doesn't solve our problems completely. 
For example it makes difficult for backend to track multiple instances of Agones (in cases when multiple developers runs local servers in the same time). Moreover with our approach it's possible both to call QA to your local server instance and to let QA launch sessions "the normal way" (request session from client build, not from UE Editor)

markmandel:
> For example it makes difficult for backend to track multiple instances of Agones (in cases when multiple developers runs local servers in the same time).

How? If it's an instance of Agones in the system just like any other, what is difficult to track?

> Moreover with our approach it's possible both to call QA to your local server instance and to let QA launch sessions "the normal way" (request session from client build, not from UE Editor)

Same question as above. The client should 

What is the flow for your current process? Reading it again:

> Well, currently our approach is that we have cluster set up on remote machine, and in-editor session connects to specified IP to mock/replace remote agones with itself

It sounds like it's exactly the same as our current `dev-address` annotation process, but possibly more complicated?

zmerlynn:
Closing due to inactivity - please reopen if you feel strongly!

PR #3000, opened 2023-02-25
Add extensions status dash and dash name change

chiayi:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Adds a separate dashboard for extensions status. It also adds `Alpha` to the Extensions resource dashboard name. 

Picture of Status dashboard:
image

### DIFF
--- a/build/grafana/dashboard-extensions-status.yaml
+++ b/build/grafana/dashboard-extensions-status.yaml
@@ -1,0 +1,843 @@
+# Copyright 2023 Google LLC All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


zmerlynn:
This isn't quite what I was thinking. Rather, I think we should fork the whole status dashboard as `(Alpha)`, adding the extensions panels (which you created here) in. That way someone can just use the Alpha status dashboard as their status dashboard, if they're running Alpha gates. Does that make sense?

zmerlynn:
> Do you mean to have another dashboard that contains all the current status along with the new Extensions panels @zmerlynn

Yes. The idea is that if you're running the feature, you use the Alpha dashboard rather than the normal one. Similarly, it means that when we graduate `SplitControllerAndExtensions`, we basically just copy it over the primary dashboard.

chiayi:
Picture of the `Alpha` status page: 
image

zmerlynn (approved):
Awesome, thanks!

PR #3002, opened 2023-02-27
Adds Counter to SDK alpha.proto

igooch:
**What type of PR is this?**

/kind feature

- Adds Counter proto message and rpc methods to SDK alpha.proto as part of feature CountsAndLists.
- Formats Counter proto message and rpc methods per Google AIP https://google.aip.dev/general guidance.
- Stubs out counter methods in the SDK server and alpha_test.
- Implements the Counter gRPC "get" and "update" in the LocalSDKServer.
- Includes Google API protos needed for following the Google AIP guidance.
- Includes generated SDK files for Golang and Swagger only.
- Includes vendor fmutils for working with FieldMasks.

Working on #2716 

For localsdk_test I found that the testify/assert.Equal did not work for comparing two protos. This is a known issue for proto equality with testify (https://github.com/stretchr/testify/issues/758), so we compare the protos with go-cmp/cmp.Diff (https://protobuf.dev/reference/go/faq/#deepequal). 

A second note:
When implementing the localsdk I noted that the field of the protos are pointers. Fields as pointers were disabled in proto3 (https://github.com/golang/protobuf/issues/15), so it appears that even thought the alpha.proto and other proto files use `syntax = "proto3";` that the protos are somehow compiled as proto2.


proto/sdk/alpha/alpha.proto
```diff
@@ -122,6 +127,24 @@ service SDK {
...
+    rpc GetCounter(GetCounterRequest) returns (Counter) {
+      option (google.api.http) = {
+        get: "/v1alpha1/{name=counters/*}"
+      };
+      option (google.api.method_signature) = "name";
+    }
+
+    // Updates a Counter. Returns INVALID_ARGUMENT if the name of the counter
+    // is non-empty and does not equal the existing name.
+    rpc UpdateCounter(UpdateCounterRequest) returns (Counter) {
```
gongmax, line 141: Look at the design doc, it defines the SDK operations as `CountIncrement`, `CountDecrement`, etc. Just want to confirm, you are grouping them together as the `UpdateCounter` to follow the AIP standard, right?

igooch: Correct yes

proto/sdk/alpha/alpha.proto
```diff
@@ -146,4 +169,48 @@ message PlayerID {
 // List of Player IDs
 message PlayerIDList {
     repeated string list = 1;
-}
\ No newline at end of file
+}
+
+// A representation of a Counter.
+message Counter {
```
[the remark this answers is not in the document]
igooch: I will update to have the Name in the Counter message per the RPC guidelines. Doesn't look like it will be an issue as we have our own custom converter for SDK -> GameServer in /agones/pkg/sdkserver/sdk.go.

proto/sdk/alpha/alpha.proto
```diff
@@ -146,4 +169,48 @@ message PlayerID {
...
+  string name = 1 [
+                  (google.api.field_behavior) = REQUIRED,
+                  (google.api.resource_reference) = {
+    type: "agones.dev/Counter"
+  }];
+}
+
+message UpdateCounterRequest {
+  // The name of the Counter to update
+  string name = 1 [
```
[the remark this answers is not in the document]
igooch: Will update.

proto/sdk/alpha/alpha.proto
```diff
@@ -122,6 +127,24 @@ service SDK {
...
+        get: "/v1alpha1/{name=counters/*}"
+      };
+      option (google.api.method_signature) = "name";
+    }
+
+    // Updates a Counter. Returns INVALID_ARGUMENT if the name of the counter
+    // is non-empty and does not equal the existing name.
+    rpc UpdateCounter(UpdateCounterRequest) returns (Counter) {
+      option (google.api.http) = {
+        patch: "/v1alpha1/{name=counters/*}"
```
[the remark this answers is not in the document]
igooch: Will update.

pkg/sdkserver/localsdk.go
```diff
@@ -585,6 +589,59 @@ func (l *LocalSDKServer) GetPlayerCapacity(_ context.Context, _ *alpha.Empty) (*
...
 
+// GetCounter returns a Counter. Returns NOT_FOUND if the counter does not exist.
+// [Stage:Alpha]
+// [FeatureFlag:CountsAndLists]
+func (l *LocalSDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest) (*alpha.Counter, error) {
+	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		return nil, errors.Errorf("%s not enabled", runtime.FeatureCountsAndLists)
+	}
+
+	l.logger.Info("Getting Counter")
```
zmerlynn, line 600: nit: log messages without data are often frustrating, consider here and below including e.g. `in.Name`.

pkg/sdkserver/localsdk.go
```diff
@@ -585,6 +589,59 @@ func (l *LocalSDKServer) GetPlayerCapacity(_ context.Context, _ *alpha.Empty) (*
...
+	l.recordRequest("getcounter")
+	l.gsMutex.RLock()
+	defer l.gsMutex.RUnlock()
+
+	if counter, ok := l.gs.Status.Counters[in.Name]; ok {
+		return &alpha.Counter{Name: in.Name, Count: &counter.Count, Capacity: &counter.Capacity}, nil
+	}
+	return nil, status.Errorf(codes.NotFound, "%s Counter NOT_FOUND", in.Name)
```
zmerlynn, line 608: It's weird to see `status` introduced where it's just not present in the API - even line 597 isn't using a status return. I'm not saying it's wrong but do we need a plan to convert over to full status, or should be keep it consistent with what's already here?

Regardless, the second half after `codes.NotFound` is for humans - just use prose instead of `NOT_FOUND`

pkg/sdkserver/localsdk_test.go
```diff
@@ -607,6 +610,130 @@ func TestLocalSDKServerPlayerConnectAndDisconnect(t *testing.T) {
...
+	assert.NoError(t, err)
+	if diff := cmp.Diff(notupdated, startingCounter, protocmp.Transform()); diff != "" {
+		t.Errorf("unexpected difference:\n%v", diff)
+	}
+
+	count := int64(9)
+	capacity := int64(999)
+	// Note: the field mask filter in UpdateCounter will remove fields from the request object that
+	// are not included in the FieldMask Paths. Here `Name` and `Capacity` will be removed from
+	// the `updated` object after we call `l.UpdateCounter()`.
```
zmerlynn, line 719: What am I missing? I was expecting to see an `expected` that only had `count`, but I see all fields present on line 722, not filtered down?

pkg/sdkserver/localsdk_test.go
```diff
@@ -607,6 +610,130 @@ func TestLocalSDKServerPlayerConnectAndDisconnect(t *testing.T) {
...
+	updated := &alpha.Counter{Name: name, Count: &count, Capacity: &capacity}
+	fm := fieldmaskpb.FieldMask{Paths: []string{"count"}}
+	expected := &alpha.Counter{Name: name, Count: &count, Capacity: &HUNDRED}
+	counterRequest := alpha.UpdateCounterRequest{Counter: updated, UpdateMask: &fm}
+	counter, err := l.UpdateCounter(context.Background(), &counterRequest)
```
zmerlynn, line 724: nit: more idiomatic would be `want` instead of `expected` and `got` instead of `counter`.

even nittier: this has more lvalues than I like, would generally inline `counterRequest` into `UpdateCounter` and I personally would end up inlining `fm` into the `UpdateCounterRequest` structure. I personally find this collection of variables a little difficult to read.

pkg/sdkserver/localsdk_test.go
```diff
@@ -607,6 +610,130 @@ func TestLocalSDKServerPlayerConnectAndDisconnect(t *testing.T) {
...
+	assert.NoError(t, err)
+	if diff := cmp.Diff(expected, counter, protocmp.Transform()); diff != "" {
+		t.Errorf("unexpected difference:\n%v", diff)
+	}
+
+	endingCounter, err := l.GetCounter(context.Background(), &alpha.GetCounterRequest{Name: "sessions"})
```
zmerlynn, line 730: nit: s/endingCounter/got/ again, then no :=

pkg/sdkserver/sdk.go
```diff
@@ -77,5 +77,16 @@ func convert(gs *agonesv1.GameServer) *sdk.GameServer {
 		}
 	}
 
+	if runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		if gs.Status.Counters != nil {
+			counters := make(map[string]*sdk.GameServer_Status_CounterStatus, len(gs.Status.Counters))
+			for key, counter := range gs.Status.Counters {
+				tmpCounter := counter.DeepCopy()
+				counters[key] = &sdk.GameServer_Status_CounterStatus{Count: tmpCounter.Count, Capacity: tmpCounter.Capacity}
```
zmerlynn, line 85: rather than counter.DeepCopy(), I'd be tempted to replace line 85 with:

```
counters[key] = &sdk.GameServer_Status_CounterStatus{Count: proto.Int32(counter.GetCount()), Capacity: proto.Int32(counter.Capacity())}
```

it's more idiomatic for proto manipulation.

pkg/sdkserver/localsdk_test.go
```diff
@@ -607,6 +610,130 @@ func TestLocalSDKServerPlayerConnectAndDisconnect(t *testing.T) {
...
+		found := false
+		l.updateObservers.Range(func(_, _ interface{}) bool {
+			found = true
+			return false
+		})
+		return found, nil
+	})
+	assert.NoError(t, err)
+
+	ONE := int64(1)
```
zmerlynn, line 706: btw, I meant to comment the other day, but the canonical way to do these is to use the proto helpers, like: https://pkg.go.dev/google.golang.org/protobuf/proto#Int64 (all throughout here)

zmerlynn (changes requested):

pkg/sdkserver/localsdk.go
igooch: I introduced it as part of the https://google.aip.dev/193 `Services must return a [google.rpc.Status](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto) message when an API error occurs, and must use the canonical error codes defined in [google.rpc.Code](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto).` 

Do you have a strong preference for keeping the errors as-is?

pkg/sdkserver/localsdk_test.go
igooch: If you were to print out the `updated` object that was originally passed into the UpdateCounterRequest there would only be the `count` field as that particular object would have been filtered down during the UpdateCounter. 

However the UpdateCounter returns the entire Counter object as it was written back to the gameserver, which is why `expected` has all fields, showing the original `name` and `capacity` values and modified `count` value. Is there a clearer way of writing that comment?

zmerlynn: Ah, now I think I understand. But if `UpdateCounter` is destructive to `updated`, that's probably a bug. In general I expect RPC calls not to destroy the things I'm passing. Maybe fix that and add a test as well?

I would expect to see `expected` (well, `got`) different than `updated` (which should probably be called `update`, the past tense is odd) given that the field mask doesn't have all the fields.

pkg/sdkserver/localsdk.go
zmerlynn: The best thing about the AIPs is at any time you can just pull up https://google.aip.dev/200 to ignore them. :)

I personally think you should just file an issue on this one and stay consistent with existing patterns. It does look like it would be straightforward to move the whole SDK, since we're not using a special error type or anything. But I think it should be done as a follow-on, possibly enforced with a helper in testing to prevent "error type leaks".

But as a more concrete point, if you do introduce them for this particular new API, I would fix line 597 at least.

zmerlynn:
I'm likely to be really slow this week but will review if I get a chance. If not, get another pass from @gongmax?

pkg/sdkserver/localsdk.go
igooch: How much consistency are we looking for? 

If we need to change all errors that are passed through (creating gameserver line 116, decoding line 715, etc.) then it seems excessive to change everything for the sake of adhering to guidelines. In this case I would just leave the error formatting as-is.

But if we're only looking to consistency in errors that are created within the SDK (i.e. lines 412, 430) then most of these are related to the PlayerTracking which will be removed at some point in the future. In this case I would use the status.Error instead of the errors.Error since we'll be removing the PlayerTracking code.

pkg/sdkserver/sdk.go
igooch: Counter doesn't have `counter.GetCount()` or `counter.Capacity()` Are those methods I should implement?

gongmax (approved):

proto/sdk/alpha/alpha.proto
```diff
@@ -122,6 +127,24 @@ service SDK {
             get: "/alpha/player/connected"
         };
     }
+
+    // Gets a Counter. Returns NOT_FOUND if the Counter does not exist.
+    rpc GetCounter(GetCounterRequest) returns (Counter) {
+      option (google.api.http) = {
+        get: "/v1alpha1/{name=counters/*}"
```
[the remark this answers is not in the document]
igooch: The consensus of the conversation starting here https://github.com/googleforgames/agones/issues/1677#issuecomment-656292055 was to use the `v1alpha1`

pkg/sdkserver/localsdk_test.go
zmerlynn: It might be, it might not be. Maybe they send the update then cache the updated value? It's an SDK, it just seems non-idiomatic to muck with the caller's structure in this case. (We don't have `const` but this would be `const` in other languages, I feel like.)

pkg/sdkserver/localsdk.go
```diff
@@ -585,6 +589,59 @@ func (l *LocalSDKServer) GetPlayerCapacity(_ context.Context, _ *alpha.Empty) (*
...
+
+	l.logger.WithField("name", in.Counter.Name).Info("Update Counter")
+	l.recordRequest("updatecounter")
+	l.gsMutex.RLock()
+	defer l.gsMutex.RUnlock()
+
+	name := in.Counter.Name
+	if counter, ok := l.gs.Status.Counters[name]; ok {
+		// Check if the UpdateMask paths are valid, return INVALID_ARGUMENT if not.
+		if in.UpdateMask.IsValid(counter.ProtoReflect().Interface()) {
```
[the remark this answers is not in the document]
igooch: Good point -- I'll add a TODO since I'll be working on the get/update for Lists next.

pkg/sdkserver/localsdk.go
zmerlynn: For me, consistency is more important than adhering to guidelines. I would rather not piecemeal adhere to guidelines, but I do think there's an opportunity to fix line 116 / 715 etc. if you want _later_. So I'm proposing either (a) ignoring aip-193 in your code, or (b) adhering to it and filing an issue for follow-on to fix the rest.

igooch: Sounds good -- I'll just remove the references to status.Error and keep consistent with error.Error.

pkg/sdkserver/localsdk_test.go
zmerlynn: I realized that on my previous review, I forgot I was reading the server-side code, and nothing the caller sends would change. Sorry, you're right, it's ok this is destroyed.

zmerlynn (approved):
Thanks for the patience.

PR #3011, opened 2023-03-02
Add Sigterm handler and readiness probe to extensions

### LINKED ISSUE #2999: Ensure `agones-extensions` is HA
While debugging #2992, I think we basically proved that `SplitControllerAndExtensions` can still result in short service flickers for the webhook. Hard to say as @chiayi was never able to repro locally (only @agones-bot repro'd).

My guess is that unless `kube-apiserver` retries webhook calls (we don't think it does), there will always be opportunities for the Agones webhooks to fail when a pod fails, suggesting calls should be retried. However, we should understand this and prove it one way or another.

chiayi:
STILL WIP. Added the original e2e extensions test that was flaky due to not being able to reach the extensions pod. 

**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Ensure that we take care of voluntary extension pod disruption by adding readiness and handler for SIGTERM. 

Closes #2999 

WIP

### DIFF
--- a/cmd/extensions/main.go
+++ b/cmd/extensions/main.go
@@ -67,10 +67,12 @@ const (
 	defaultResync                = 30 * time.Second
 	apiServerSustainedQPSFlag    = "api-server-qps"
 	apiServerBurstQPSFlag        = "api-server-qps-burst"
+	readinessShutdownDuration    = "readiness-shutdown-duration"
 )
 
 var (
-	logger = runtime.NewLoggerWithSource("main")
+	podReady bool
+	logger   = runtime.NewLoggerWithSource("main")
 )
 
 func setupLogging(logDir string, logSizeLimitMB int) {
@@ -93,7 +95,7 @@ func setupLogging(logDir string, logSizeLimitMB int) {
 
 // main initializes the extensions service for Agones
 func main() {
-	ctx := signals.NewSigKillContext()
+	ctx, cancelCtx := context.WithCancel(context.Background())
 	ctlConf := parseEnvFlags()
 
 	if ctlConf.LogDir != "" {
@@ -170,6 +172,23 @@ func main() {
 		health = healthcheck.NewHandler()
 	}
 
+	podReady = true
+	health.AddReadinessCheck("agones-extensions", func() error {
+		if !podReady {
+			return errors.New("asked to shut down, failed readiness check")
+		}
+		return nil
+	})
+
+	signals.NewSigTermHandler(func() {
+		logger.Info("Pod shutdown has been requested, failing readiness check")
+		podReady = false
+		time.Sleep(ctlConf.ReadinessShutdownDuration)
+		cancelCtx()
+		logger.Infof("Readiness shutdown duration has passed, exiting pod")
+		os.Exit(0)
+	})
+
 	// If we are using Prometheus only exporter we can make reporting more often,
 	// every 1 seconds, if we are using Stackdriver we would use 60 seconds reporting period,
 	// which is a requirements of Stackdriver, otherwise most of time series would be invalid for Stackdriver
@@ -241,6 +260,7 @@ func parseEnvFlags() config {
 	pflag.Int32(logSizeLimitMBFlag, 1000, "Log file size limit in MB")
 	pflag.String(logLevelFlag, viper.GetString(logLevelFlag), "Agones Log level")
 	pflag.Duration(allocationBatchWaitTime, viper.GetDuration(allocationBatchWaitTime), "Flag to configure the waiting period between allocations batches")
+	pflag.Duration(readinessShutdownDuration, viper.GetDuration(readinessShutdownDuration), "Time in seconds for SIGTERM handler to sleep for.")
 	cloudproduct.BindFlags()
 	runtime.FeaturesBindFlags()
 	pflag.Parse()
@@ -264,6 +284,7 @@ func parseEnvFlags() config {
 	runtime.Must(viper.BindEnv(logSizeLimitMBFlag))
 	runtime.Must(viper.BindEnv(allocationBatchWaitTime))
 	runtime.Must(viper.BindPFlags(pflag.CommandLine))
+	runtime.Must(viper.BindEnv(readinessShutdownDuration))
 	runtime.Must(cloudproduct.BindEnv())
 	runtime.Must(runtime.FeaturesBindEnv())
 	runtime.Must(runtime.ParseFeaturesFromEnv())
@@ -278,13 +299,14 @@ func parseEnvFlags() config {
 		Stackdriver:       viper.GetBool(enableStackdriverMetricsFlag),
 		StackdriverLabels: viper.GetString(stackdriverLabels),
 
-		NumWorkers:              int(viper.GetInt32(numWorkersFlag)),
-		APIServerSustainedQPS:   int(viper.GetInt32(apiServerSustainedQPSFlag)),
-		APIServerBurstQPS:       int(viper.GetInt32(apiServerBurstQPSFlag)),
-		LogDir:                  viper.GetString(logDirFlag),
-		LogLevel:                viper.GetString(logLevelFlag),
-		LogSizeLimitMB:          int(viper.GetInt32(logSizeLimitMBFlag)),
-		AllocationBatchWaitTime: viper.GetDuration(allocationBatchWaitTime),
+		NumWorkers:                int(viper.GetInt32(numWorkersFlag)),
+		APIServerSustainedQPS:     int(viper.GetInt32(apiServerSustainedQPSFlag)),
+		APIServerBurstQPS:         int(viper.GetInt32(apiServerBurstQPSFlag)),
+		LogDir:                    viper.GetString(logDirFlag),
+		LogLevel:                  viper.GetString(logLevelFlag),
+		LogSizeLimitMB:            int(viper.GetInt32(logSizeLimitMBFlag)),
+		AllocationBatchWaitTime:   viper.GetDuration(allocationBatchWaitTime),
+		ReadinessShutdownDuration: viper.GetDuration(readinessShutdownDuration),
 	}
 }
 
@@ -298,14 +320,15 @@ type config struct {
 	Stackdriver       bool
 	StackdriverLabels string
 
-	GCPProjectID            string
-	NumWorkers              int
-	APIServerSustainedQPS   int
-	APIServerBurstQPS       int
-	LogDir                  string
-	LogLevel                string
-	LogSizeLimitMB          int
-	AllocationBatchWaitTime time.Duration
+	GCPProjectID              string
+	NumWorkers                int
+	APIServerSustainedQPS     int
+	APIServerBurstQPS         int
+	LogDir                    string
+	LogLevel                  string
+	LogSizeLimitMB            int
+	AllocationBatchWaitTime   time.Duration
+	ReadinessShutdownDuration time.Duration
 }
 
 type runner interface {
--- a/install/helm/agones/templates/extensions-deployment.yaml
+++ b/install/helm/agones/templates/extensions-deployment.yaml
@@ -71,6 +71,7 @@ spec:
       priorityClassName: {{ .Values.agones.priorityClassName }}
       {{- end }}
       serviceAccountName: {{ .Values.agones.serviceaccount.controller.name }}
+      terminationGracePeriodSeconds: {{ mul .Values.agones.extensions.readiness.periodSeconds .Values.agones.extensions.readiness.failureThreshold 3 }}
       containers:
       - name: agones-extensions
         image: "{{ .Values.agones.image.registry }}/{{ .Values.agones.image.extensions.name}}:{{ default .Values.agones.image.tag .Values.agones.image.extensions.tag }}"
@@ -114,6 +115,8 @@ spec:
               fieldPath: metadata.namespace
         - name: CONTAINER_NAME
           value: "agones-extensions"
+        - name: READINESS_SHUTDOWN_DURATION
+          value: {{ mul .Values.agones.extensions.readiness.periodSeconds .Values.agones.extensions.readiness.failureThreshold 2 }}s
         ports:
         - name: webhooks
           containerPort: 8081
@@ -127,6 +130,13 @@ spec:
           periodSeconds: {{ .Values.agones.extensions.healthCheck.periodSeconds }}
           failureThreshold: {{ .Values.agones.extensions.healthCheck.failureThreshold }}
           timeoutSeconds: {{ .Values.agones.extensions.healthCheck.timeoutSeconds }}
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 8080
+          initialDelaySeconds: {{ .Values.agones.extensions.readiness.initialDelaySeconds }}
+          periodSeconds: {{ .Values.agones.extensions.readiness.periodSeconds }}
+          failureThreshold: {{ .Values.agones.extensions.readiness.failureThreshold }}
         resources:
 {{- if .Values.agones.extensions.resources }}
 {{ toYaml .Values.agones.extensions.resources | indent 10 }}
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -99,6 +99,10 @@ agones:
     pdb:
         minAvailable: 1
     replicas: 2
+    readiness:
+      initialDelaySeconds: 3
+      periodSeconds: 3
+      failureThreshold: 3
   ping:
     install: true
     pdb:
--- a/pkg/util/signals/signals.go
+++ b/pkg/util/signals/signals.go
@@ -18,6 +18,7 @@ package signals
 
 import (
 	"context"
+	"os"
 	"os/signal"
 	"syscall"
 )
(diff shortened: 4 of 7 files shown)


cmd/extensions/main.go
```diff
@@ -170,6 +171,16 @@ func main() {
...
 
+	podReady = true
+	health.AddReadinessCheck("agones-extensions", func() error {
+		if !podReady {
+			return errors.New("Pod is not in Ready state")
+		}
+		return nil
+	})
+
+	signals.NewSigTermHandler(ctx, HandleSigTerm)
```
zmerlynn, line 182: We talked about it previously, but you're going to need to deal with line 97 (the `NewSigKillContext`). What I would probably do is change line 97 to:

```
ctx, cancelCtx := context.WithCancel(context.Background())
```

then in `HandleSigTerm`, you'll need to call that cancelFunc().

Frankly, to make this easier to understand and avoid using global variables, I would consider just inlining the function here, i.e.:

```
signals.NewSigTermHandler(crd, func() {
  podReady = false
  time.Sleep(mumble mumble)
  cancelCtx()
  ???exit???
})
```

but you're on the right track.

pkg/util/signals/signals.go
```diff
@@ -18,12 +18,28 @@ package signals
...
+// NewSigTermHandler creates a channel to listen to SIGTERM and runs the handle function
+func NewSigTermHandler(ctx context.Context, handle func()) {
+	_, cancel := context.WithCancel(ctx)
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGTERM)
+
+	go func() {
+		<-c
+		handle()
+		time.Sleep(30 * time.Second)
```
zmerlynn, line 42: Oh, I see what you did. I would take the logic out of here - I don't think this function even needs to be context aware. I would especially take the sleep out of here, it's pretty bespoke.

cmd/extensions/main.go
chiayi: Ooooh, I see.

pkg/util/signals/signals.go
chiayi: Yea your suggestion looks and feels much cleaner.

cmd/extensions/main.go
```diff
@@ -170,6 +171,21 @@ func main() {
 		health = healthcheck.NewHandler()
 	}
 
+	podReady = true
+	health.AddReadinessCheck("agones-extensions", func() error {
+		if !podReady {
+			return errors.New("Pod is not in Ready state")
```
zmerlynn, line 177: I would make this error clearer and say something like "Pod shutting down, failing readiness checks". The error you have here is tautological.

cmd/extensions/main.go
```diff
@@ -170,6 +171,21 @@ func main() {
...
+		}
+		return nil
+	})
+
+	signals.NewSigTermHandler(func() {
+		podReady = false
+		time.Sleep(30 * time.Second)
```
zmerlynn, line 184: make this a configurable flag. Personally I would consider defaulting it from helm - I say that because I think it's not hard to make it something like `periodSeconds * failureThreshold * 2`, where the *2 is just a safety/fudge factor.

install/helm/agones/templates/extensions-deployment.yaml
```diff
@@ -123,6 +123,10 @@ spec:
           httpGet:
             path: /live
             port: http
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 8080
           initialDelaySeconds: {{ .Values.agones.extensions.healthCheck.initialDelaySeconds }}
```
zmerlynn, line 130: the way you've added this, the `initialDelaySeconds` etc. are applied to the `readinessProbe` but were previously configured on the liveness probe. I would add a `.readiness` with similar settings (it looks like we default to 3/3/3 in values.yaml, that SGTM for readiness as well).

test/e2e/extensions/high_availability_test.go
```diff
@@ -64,9 +64,38 @@ func TestGameServerHealthyAfterDeletingPodWhileOneExtensionsDown(t *testing.T) {
...
+		logger.Infof("Name of extensions pod %v: %v", i, list.Items[i].ObjectMeta.Name)
+	}
+	require.NoError(t, err, "Could not get list of Extension pods")
+
+	logger.Infof("Removing one of the Extensions Pods: %v", list.Items[1].ObjectMeta.Name)
+	deleteAgonesExtensionsPod(ctx, t, false)
+
+	gs := framework.DefaultGameServer(defaultNs)
+	logger.Infof("Creating game-server %s...", gs.Name)
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(t, defaultNs, gs)
```
zmerlynn, line 89: so here's a thought.. the extensions pod is an API pod. For the most part, it's not exactly involved in the "wait until ready" part and doesn't have a lot of interaction with the `GameServer` after creation. Furthermore, this only really does the one interaction, and does it right after we delete the extensions pod, when the pod should be lame-ducked. The more meaningful time for it to try is right after the readiness failure and the connection is drained.

We can't detect that timing exactly, of course, but I would proposed that a meaningful test here is just to loop for like 30s, creating and delete a gameserver. That probes the API continuously for that period and should give us a more meaningful signal.

zmerlynn (changes requested):

cmd/extensions/main.go
chiayi: Done!

install/helm/agones/templates/extensions-deployment.yaml
chiayi: I see, created a section for readinessProbe and added a section in .Values as well for readiness.

test/e2e/extensions/high_availability_test.go
chiayi: I added a loop around the create and delete step. Please take a look.

cmd/extensions/main.go
```diff
@@ -170,6 +173,21 @@ func main() {
 		health = healthcheck.NewHandler()
 	}
 
+	podReady = true
+	health.AddReadinessCheck("agones-extensions", func() error {
+		if !podReady {
+			return errors.New("extensions Pod failed Readiness check, shutting down")
```
zmerlynn, line 179: nit: This error still seems to imply that something failed, vs the reality is that we're failing readiness checks _because_ we're shutting down. Something like "asked to shut down, failing readiness check".

cmd/extensions/main.go
```diff
@@ -67,10 +67,13 @@ const (
 	defaultResync                = 30 * time.Second
 	apiServerSustainedQPSFlag    = "api-server-qps"
 	apiServerBurstQPSFlag        = "api-server-qps-burst"
+	sleepPeriod                  = "sleep-period"
+	sleepFailureThreshold        = "sleep-failure-threshold"
```
zmerlynn, line 71: Why two flags, when they're only ever used as `sleepPeriod * sleepFailureThreshold`.

I would suggest a single flag, `shutdownDurationFlag`, maybe call it `--shutdown-duration`. And make it a `pflag.Duration` - canonically anything that is a time duration should be a Duration.

cmd/extensions/main.go
```diff
@@ -241,6 +262,8 @@ func parseEnvFlags() config {
 	pflag.Int32(logSizeLimitMBFlag, 1000, "Log file size limit in MB")
 	pflag.String(logLevelFlag, viper.GetString(logLevelFlag), "Agones Log level")
 	pflag.Duration(allocationBatchWaitTime, viper.GetDuration(allocationBatchWaitTime), "Flag to configure the waiting period between allocations batches")
+	pflag.Int32(sleepPeriod, 15, "Time for SIGTERM handler to sleep for.")
```
zmerlynn, line 265: Any idea why we set defaults twice here? Shouldn't this be..

```
plfag.Int32(sleepPeriodFlag, viper.GetInt(sleepPeriodFlag), ...)
```

I see for other ints we're doing the same thing, but it doesn't make sense? (Maybe let's start a trend here and fix it for these.)

Regardless, the value here is mismatched from the value on line 246.

(But separately, as I mentioned above, this should be a duration.)

cmd/extensions/main.go
```diff
@@ -170,6 +173,21 @@ func main() {
...
 
+	podReady = true
+	health.AddReadinessCheck("agones-extensions", func() error {
+		if !podReady {
+			return errors.New("extensions Pod failed Readiness check, shutting down")
+		}
+		return nil
+	})
+
+	signals.NewSigTermHandler(func() {
```
zmerlynn, line 184: If we have a logger available, I would recommend adding a log message at the top and right before exiting. Something like "shutdown requested, failing readiness checks" at the top and "shutdown duration passed, exiting" or somesuch.

install/helm/agones/values.yaml
```diff
@@ -99,6 +99,12 @@ agones:
     pdb:
         minAvailable: 1
     replicas: 2
+    readiness:
+      initialDelaySeconds: 3
+      periodSeconds: 3
+      failureThreshold: 3
+    sleepPeriod: 15
+    sleepFailurThreshold: 2
```
zmerlynn, line 107: I think you may have gotten the two sleep variables off my previous suggestion about the variables _here_. 

I'm not sure the shutdown duration should make it to values.yaml. Instead, I think that you should remove it here, but in `extensions-deployment.yaml`, set:

```
- SHUTDOWN_DURATION
  value: {{ .Value.agones.extensions.periodSeconds * .Value.agones.extensions.failureThreshold * 2 }}s
```

That's where I was suggesting a multiplication.

The point is, we might want to refine the failure thresholds / readiness checking for some issue, but in general, we just need some safety factor beyond the readiness-failure duration (which is where the 2* is from) to drain. The user doesn't need to configure that part, I don't think.

test/e2e/extensions/high_availability_test.go
```diff
@@ -64,9 +64,40 @@ func TestGameServerHealthyAfterDeletingPodWhileOneExtensionsDown(t *testing.T) {
...
+
+	logger.Infof("Removing one of the Extensions Pods: %v", list.Items[1].ObjectMeta.Name)
+	deleteAgonesExtensionsPod(ctx, t, false)
+
+	endTime := time.Now().Add(30 * time.Second)
+	for time.Now().Before(endTime) {
+		gs := framework.DefaultGameServer(defaultNs)
+		logger.Infof("Creating game-server %s...", gs.Name)
+		readyGs, err := framework.CreateGameServerAndWaitUntilReady(t, defaultNs, gs)
+		require.NoError(t, err, "Could not get a GameServer ready")
```
zmerlynn, line 92: can we avoid waiting, too?

zmerlynn (changes requested):

install/helm/agones/values.yaml
chiayi: Ooh I see, I did not realize you could do the multiplication here.

zmerlynn: you can't in the values.yaml, but in the deployment yaml, sure.

test/e2e/extensions/high_availability_test.go
chiayi: I believe we could. Just add a separate method that will return the Gameserver right after it sends a request to through agones client.

chiayi: I added a method the method. Please take a look.

install/helm/agones/values.yaml
chiayi: done!

cmd/extensions/main.go
chiayi: I named it ReadinessShutdownDuration just to clarify that its for the readiness check.

test/e2e/framework/framework.go
```diff
@@ -247,6 +247,14 @@ func (f *Framework) CreateGameServerAndWaitUntilReady(t *testing.T, ns string, g
 	return readyGs, nil
 }
 
+// CreateGameServer Creates a GameServer.
```
zmerlynn, line 250: I would either factor this with the above so that `CreateGameServerAnd` calls this function, or I would just inline this in the test you wrote. Frankly the latter seems simple enough, since this falls in that "just a wrapper" category that Go tends to avoid.

cmd/extensions/main.go
```diff
@@ -278,13 +299,14 @@ func parseEnvFlags() config {
...
-		LogSizeLimitMB:          int(viper.GetInt32(logSizeLimitMBFlag)),
-		AllocationBatchWaitTime: viper.GetDuration(allocationBatchWaitTime),
+		NumWorkers:                int(viper.GetInt32(numWorkersFlag)),
+		APIServerSustainedQPS:     int(viper.GetInt32(apiServerSustainedQPSFlag)),
+		APIServerBurstQPS:         int(viper.GetInt32(apiServerBurstQPSFlag)),
+		LogDir:                    viper.GetString(logDirFlag),
+		LogLevel:                  viper.GetString(logLevelFlag),
+		LogSizeLimitMB:            int(viper.GetInt32(logSizeLimitMBFlag)),
+		AllocationBatchWaitTime:   viper.GetDuration(allocationBatchWaitTime),
+		ReadinessShutdownDuration: time.Duration(viper.GetDuration(readinessShutdownDuration).Seconds()),
```
zmerlynn, line 309: Why are you extracting Seconds when the type is a `time.Duration`?

install/helm/agones/templates/extensions-deployment.yaml
```diff
@@ -114,6 +114,8 @@ spec:
               fieldPath: metadata.namespace
         - name: CONTAINER_NAME
           value: "agones-extensions"
+        - name: READINESS_SHUTDOWN_DURATION
+          value: {{ .Value.agones.extensions.periodSeconds * .Value.agones.extensions.failureThreshold * 2 | quote }}
```
zmerlynn, line 118: You need the `s` I had in the example. So 

```
{{ mumble }}s
```

to indicate seconds, since it's a duration now.

cmd/extensions/main.go
zmerlynn: The readiness check isn't the cause of the shutdown. Shutdown was requested (by `SIGTERM`), so we're failing readiness checks. Does that make sense? This log message is suggesting that failed readiness checks are somehow _why_ we're shutting down, but that's inverting the cause - instead, we're shutting down, so we're failing readiness checks to drain existing connections.

install/helm/agones/templates/extensions-deployment.yaml
```diff
@@ -114,6 +114,8 @@ spec:
               fieldPath: metadata.namespace
         - name: CONTAINER_NAME
           value: "agones-extensions"
+        - name: READINESS_SHUTDOWN_DURATION
+          value: {{ mul .Value.agones.extensions.readiness.periodSeconds .Value.agones.extensions.readiness.failureThreshold 2 | quote }}s
```
zmerlynn, line 118: This looks good. Let's also go ahead and configure `terminationGracePeriodSeconds` while we're at it - that way if someone configures things such that it's longer than the 30s `kubelet` gives us by default, we have a reasonable configuration. It can be configured for e.g. `periodSeconds * failureThreshold * 3` to give another safety envelope, but if we go beyond *2, I'd be shocked.

test/e2e/extensions/high_availability_test.go
```diff
@@ -64,9 +64,40 @@ func TestGameServerHealthyAfterDeletingPodWhileOneExtensionsDown(t *testing.T) {
...
+
+	logger.Infof("Removing one of the Extensions Pods: %v", list.Items[1].ObjectMeta.Name)
+	deleteAgonesExtensionsPod(ctx, t, false)
+
+	endTime := time.Now().Add(30 * time.Second)
+	for time.Now().Before(endTime) {
+		gs := framework.DefaultGameServer(defaultNs)
+		logger.Infof("Creating game-server %s...", gs.Name)
+		newGs, err := framework.AgonesClient.AgonesV1().GameServers(defaultNs).Create(context.Background(), gs, metav1.CreateOptions{})
+		require.NoError(t, err, "Could not get a GameServer ready")
```
zmerlynn, line 92: It's no longer trying to get it ready, and line 93 should probably go, too.

zmerlynn (changes requested):

cmd/extensions/main.go
```diff
@@ -170,6 +172,23 @@ func main() {
...
+	podReady = true
+	health.AddReadinessCheck("agones-extensions", func() error {
+		if !podReady {
+			return errors.New("asked to shut down, failed readiness check")
+		}
+		return nil
+	})
+
+	signals.NewSigTermHandler(func() {
+		logger.Info("pod shutdown has been requested, failing readiness check")
```
zmerlynn, line 184: total nit, but s/pod/Pod/ for consistency with other log messages.

zmerlynn (approved):
Thanks for your patience. :)

cmd/extensions/main.go
chiayi: Done!

zmerlynn (approved):

chiayi:
Thank YOU for having the patience to go through this so many times! 😄 @zmerlynn

PR #3032, opened 2023-03-16
Fix GKE Autopilot auto-detection for 1.26

zmerlynn:
The `workload-defaulter` no longer exists as a standalone webhook on 1.26 - instead use the new `warden-mutating` webhook. I am following up internally with finding a less brittle way to do this.

### DIFF
--- a/pkg/cloudproduct/gke/gke.go
+++ b/pkg/cloudproduct/gke/gke.go
@@ -34,8 +34,6 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


gongmax (approved):

pkg/cloudproduct/gke/gke.go
gongmax, line 1: nit: do we have unit test for this logic?

gongmax (approved):

PR #3039, opened 2023-03-23
Adds List to SDK alpha.proto

igooch:
**What type of PR is this?**

/kind feature

- Adds List proto message and rpc methods to SDK alpha.proto as part of feature CountsAndLists.
- Removes "optional" from both List and Counter fields in alpha.proto per https://cloud.google.com/apis/design/design_patterns.md#optional_primitive_fields.
- Formats List proto message and rpc methods per Google AIP https://google.aip.dev/general guidance.
- Includes "add" and "remove" methods per Google AIP https://google.aip.dev/144 guidance for repeated fields.
- Stubs out List "get" "update" "add" and "remove" methods in the SDK server and alpha_test.
- Implements the List gRPC "get" "update" "add" and "remove" in the LocalSDKServer.
- Adds unit tests for the List and Counter methods in the LocalSDKServer.
- Makes minor formatting changes to Counter methods to be consistent with List.
- Includes generated SDK files for Golang and Swagger only.

Working on #2716


pkg/sdkserver/localsdk.go
```diff
@@ -620,27 +620,164 @@ func (l *LocalSDKServer) UpdateCounter(ctx context.Context, in *alpha.UpdateCoun
...
+// [Stage:Alpha]
+// [FeatureFlag:CountsAndLists]
+func (l *LocalSDKServer) UpdateList(ctx context.Context, in *alpha.UpdateListRequest) (*alpha.List, error) {
+	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		return nil, errors.Errorf("%s not enabled", runtime.FeatureCountsAndLists)
+	}
+
+	l.logger.WithField("name", in.List.Name).Info("Updating List")
+	l.recordRequest("updatelist")
+	l.gsMutex.RLock()
```
gongmax, line 682: I think we should use `Lock()` here to prevent a dirty read of `l.gs.Status.Lists`. Same for the `UpdateCounter`

proto/sdk/alpha/alpha.proto
```diff
@@ -207,3 +247,62 @@ message UpdateCounterRequest {
...
+
+  // Required. Mask (list) of fields to update.
+  // Fields are specified relative to the List
+  // (e.g. `capacity`, `values`; *not* `List.capacity` or `List.values`).
+  google.protobuf.FieldMask update_mask = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+message AddListValueRequest {
+  // The name of the List to add a value to.
+  string name = 1 [
```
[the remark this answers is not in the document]
igooch: I chose to leave it as "name" since I thought it would be confusing to have a string list, string name (GetListRequest), and a List list (UpdateListRequest).

proto/sdk/alpha/alpha.proto
```diff
@@ -138,13 +138,53 @@ service SDK {
...
+      };
+      option (google.api.method_signature) = "list,update_mask";
+    }
+
+    // Adds a value to a List and returns updated List. Returns NOT_FOUND if the List does not exist.
+    // Returns ALREADY_EXISTS if the value is already in the List.
+    // Returns OUT_OF_RANGE if the List is already at Capacity.
+    rpc AddListValue(AddListValueRequest) returns (List) {
+      option (google.api.http) = {
+        post: "/v1/{name=lists/*}:addValue"
```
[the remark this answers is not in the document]
igooch: Per the AIP yes, but I think this confuses things, since we're not actually passing a list object, we're passing the string name of the list. IMO the naming should be consistent with GetList which also passes a string name. LMK if you think we should follow the AIP in here.

proto/sdk/alpha/alpha.proto
```diff
@@ -138,13 +138,53 @@ service SDK {
...
+    // Gets a List. Returns NOT_FOUND if the List does not exist.
+    rpc GetList(GetListRequest) returns (List) {
+      option (google.api.http) = {
+        get: "/v1alpha1/{name=lists/*}"
+      };
+      option (google.api.method_signature) = "name";
+    }
+
+    // UpdateList returns the updated List. Returns NOT_FOUND if the List does not exist (name cannot be updated).
+    // **THIS WILL OVERWRITE ALL EXISTING LIST.VALUES WITH REQUEST LIST.VALUES** use AddListValue()
```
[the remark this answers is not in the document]
igooch: Correct yes, this can also change the capacity.

proto/sdk/alpha/alpha.proto
```diff
@@ -207,3 +247,62 @@ message UpdateCounterRequest {
...
+// A representation of a List.
+message List {
+  option (google.api.resource) = {
+      type: "agones.dev/List"
+      pattern: "lists/{list}"
+  };
+  // The name of the List
+  string name = 1;
+  // The maximum capacity of the List
+  optional int64 capacity = 2;
```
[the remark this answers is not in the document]
igooch: Good point -- capacity should always be set (either directly by the user or by the default value), so this should never be null.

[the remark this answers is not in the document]
gongmax: We decided to not use `optional`.

pkg/sdkserver/sdk.go
```diff
@@ -86,6 +86,14 @@ func convert(gs *agonesv1.GameServer) *sdk.GameServer {
 			}
 			result.Status.Counters = counters
 		}
+
+		if gs.Status.Lists != nil {
```
gongmax, line 90: Better to have unit test coverage.

pkg/sdkserver/localsdk.go
```diff
@@ -617,30 +620,167 @@ func (l *LocalSDKServer) UpdateCounter(ctx context.Context, in *alpha.UpdateCoun
...
+
+// UpdateList returns the updated List. Returns NOT_FOUND if the List does not exist (name cannot be updated).
+// **THIS WILL OVERWRITE ALL EXISTING LIST.VALUES WITH ANY REQUEST LIST.VALUES**
+// Use AddListValue() or RemoveListValue() for modifying the List.Values field.
+// Returns INVALID_ARGUMENT if the field mask path(s) are not field(s) of the List.
+// If a field mask path(s) is specified, but the value is not set in the request List object,
+// then the default value for the variable will be set (i.e. 0 for "capacity", empty list for "values").
+// [Stage:Alpha]
+// [FeatureFlag:CountsAndLists]
+func (l *LocalSDKServer) UpdateList(ctx context.Context, in *alpha.UpdateListRequest) (*alpha.List, error) {
```
gongmax, line 682: Do we also need to add validation logic to check the required fields in the request actually exist?

gongmax (commented):
Overall looks good. One suggestion is to add more unit test coverage. Right now the `localsdk_test` only covers part of the happy path for both counter and list. IMO, it's better to have test coverage like:
#### GetCounter
- Unhappy path:
  - Not found
- Happy path
#### UpdateCounter
- Unhappy path: 
  - Counter not found
  - `counter` in the request is missing
  - `update_mask` in the request is missing
  - `paths` in the `update_mask` is missing
  - `paths` is empty
  - `paths` has invalid value
  - `counter` out of range
  - `capacity` out of range (once the logic is implemented)
- Happy case
  - `in.counter` does not have the field specified in `paths` so default value is used (already covered)
  - `in.counter` has all the fields specified in `paths` so they are applied
  - `in.counter` has field which is not included in `paths`, so it gets filtered

Similar for `GetList` and `UpdateList`

For `AddListValue`, better to add unhappy path where 1. List not found. 2. the value already exists. 2. Out of capacity.
For `RemoveListValue`, better to add unhappy path where 1. List not found. 2. the value not found.

pkg/sdkserver/localsdk.go
igooch: At this point the fields do exist, because the default fields are added to the object at creation if not specified.

gongmax: Indeed, but aren't the default value for `List` and `UpdateMask` in the `UpdateListRequest` `nil`?

igooch: Ah gotcha, I can add a nil check on the UpdateListRequest.

gongmax (approved):

PR #3047, opened 2023-03-28
Identify and close stale issues that are inactive for the specified time

geetachavan1:
**What type of PR is this?**
> Uncomment only one ` /kind ` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

1. This PR will automate the process of identifying the stale issues which have been inactive more than 30 days
2. Wait for another 30 days before marking Obsolete
3. Wait for another 30 days. if still inactive, it will close the issue.
4. The issue carries "awaiting maintainer" label will be skipped and will remain open.
5. At every stage it will add a relevant comment.
6. To start with, it will run the action on the 1st and 15th of every month and will use the default value of 30 for operations-per-run. This will prevent it from applying stale labels on the all the issues at once.
It is using GitHub Actions CI/CD platform.

If approved and merged, for the successful run we will need to create following labels

1. stale
2. obsolete
3. close
4. awaiting maintainer


.github/workflows/obsolete.yaml
```diff
@@ -0,0 +1,36 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
```
markmandel, line 1: Apache header please.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,33 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
```
markmandel, line 1: Apache header please.

.github/workflows/obsolete.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+    steps:
+    - name: Track stale issues and check if obsolete
+      uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 29
+        stale-issue-message: >
+          This issue is marked as obsolete due to inactivity for last 60 days. 
+          To avoid being marked as "Obsolete" please add a comment or add "awaiting-maintainer" label. Thank you for your contributions
```
markmandel, line 30:
```suggestion
          To avoid being closed in 30 days please add a comment or add "awaiting-maintainer" label. Thank you for your contributions.
```

Should it be that if they want to avoid the next state (obsolete), they should do one of the above?

.github/workflows/obsolete.yaml
```diff
@@ -0,0 +1,36 @@
...
+    permissions:
+      issues: write
+    name: Track Obsolete Issues
+
+    steps:
+    - name: Track stale issues and check if obsolete
+      uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 29
```
markmandel, line 27: Curious - why 29 and not 30?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,33 @@
...
+    permissions:
+      issues: write
+    name: stale issues
+
+    steps:
+    - name: Stale issues
+      uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 31
```
markmandel, line 26: Curious - why 31 and not 30?

.github/workflows/close.yaml
```diff
@@ -0,0 +1,34 @@
...
+      issues: write
+      
+
+    steps:
+    - uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 30
+        stale-issue-message: >
+          @author, We are closing this as there is no activity in this issue for last 90 days. Please reopen if you’d like to discuss anything further.
```
markmandel, line 27: I think the better way to do this is to use `close-issue-message` and `days-before=close` on the obsolete workflow, since that is the workflow I see in the github action docs:

e.g. 
https://github.com/actions/stale/tree/main#close-issue-message:~:text=Configure%20different%20stale%20timeouts%3A

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,33 @@
...
+        stale-issue-message: >
+          'This issue is marked as Stale due to inactivity for more than 30 days. 
+          To avoid being marked as "Stale" please add "awaiting-maintainer" label or comment. Thank you for your contributions '
```
markmandel, line 29:
```suggestion
          'This issue is marked as Stale due to inactivity for more than 30 days. 
          To avoid being marked as `obsolete` please add `awaiting-maintainer` label or comment. Thank you for your contributions.'
```

.github/workflows/obsolete.yaml
geetachavan1: The github actions does not have intermediate step that we need to have. In Agones case, we want to apply stale, then wait for 'x' days, apply obsolete and wait again for 'x' then close.. You will also see that close action is part of only close.yaml and is removed from stale.yaml and obsolete.yaml. As for the days,  if I keep 30 days for obsolete.yaml as well, there is a chance of stale.yaml kicking in before obsolete.yaml due to cron disfunction. If that happens then it will remove the 'stale' label as applying label is relevant activity and we will be in loop. For obsolete.yaml and close.yaml, I have 'only-on' option but not in case of 'stale.yaml' Hope it make sense..

.github/workflows/stale.yaml
geetachavan1: Compensating the missed day in 'obsolete' so the total is 90+ days inactivity as decided earlier..

PR #3051, opened 2023-03-29
Add helm setting for leader election

chiayi:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This PR adds helm config for controller leader election


install/helm/agones/templates/_helpers.tpl
```diff
@@ -69,4 +69,24 @@ feature gates. In the process, we validate that the feature gate is known.
...
 {{- end -}}
 {{- end -}}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{/* 
+Derives the replica number for agones-controller based on current helm and feature flag settings.
+*/}}
+{{- define "agones.controller.replica" -}}
```
zmerlynn, line 77: I forgot this was landing all in `controller.yaml`, any reason we can't just inline the logic rather than adding the helper? I think if you inline it at e.g. line 16 in controller.yaml, you don't even need a `define`, it just becomes a matter of setting $controllerReplicas.

install/helm/agones/templates/controller.yaml
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 {{- $featureGates := include "agones.featureGates" . | fromYaml }}
+{{- $controllerReplica := include "agones.controller.replica" . }}
```
zmerlynn, line 16: nit: I would call this `controllerReplicas` (plural), or just `replicas`. You're in `controller.yaml`, the `controller` prefix is kind of redundant.

install/helm/agones/templates/controller.yaml
```diff
@@ -31,13 +32,13 @@ spec:
...
       heritage: {{ .Release.Service }}
-  replicas: 1
+  replicas: {{ $controllerReplica }}
   strategy:
     type: Recreate
   template:
     metadata:
       annotations:
-{{- if not $featureGates.SplitControllerAndExtensions }}
+{{- if not (gt (int $controllerReplica) 1) }}
```
zmerlynn, line 41: do you need the cast? or is it just protecting the case where they passed a non-int for `replicas`? If the latter, can we validate it above rather than casting throughout?

install/helm/agones/templates/_helpers.tpl
```diff
@@ -69,4 +69,24 @@ feature gates. In the process, we validate that the feature gate is known.
...
+{{- if $featureGates.SplitControllerAndExtensions -}}
+{{- default 2 .Values.agones.controller.replicas -}}
+{{- else -}}
+{{- default 1 .Values.agones.controller.replicas -}}
+{{- end -}}
+{{- else -}}
+{{- if not $featureGates.SplitControllerAndExtensions -}}
+{{- default 1 .Values.agones.controller.replicas -}}
+{{- else -}}
+{{- .Values.agones.controller.replicas -}}
```
zmerlynn, line 89: I like this logic better than what I specified (and update the design!), but maybe we can strip it down even further:

* if SprintControllerAndExtensions, use the replicas values passed in
* if not, replicas=1

Basically, remove the line 79-84 clause, we don't need to protect against an empty value.

As a nit, I would flip the order of clause in line 86 and 88, though: in an if/else, prefer to keep the "positive" case first.

install/helm/agones/values.yaml
```diff
@@ -94,6 +94,9 @@ agones:
       failureThreshold: 3
       timeoutSeconds: 1
     allocationBatchWaitTime: 500ms
+    replicas: 2
+    pdb:
+        minAvailable: 1
```
zmerlynn, line 99: remind that everything in values.yaml needs to be documented

zmerlynn (changes requested):

install/helm/agones/templates/controller.yaml
chiayi: Done, changed to just `replicas`

install/helm/agones/templates/controller.yaml
chiayi: there was a typing error when I was using the helper, shouldn't be an issue now.

install/helm/agones/values.yaml
chiayi: Added to `helm.md`

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -259,6 +259,9 @@ The following tables lists the configurable parameters of the Agones chart and t
...
 |                                                          |                                                                                                                                                                                                                                                            |
+| `agones.controller.replicas`                                   | The number of replicas to run in the deployment                                                                                                                                                                         | `2`                                |
+| `agones.controller.pdb.minAvailable`        | Description of the number of pods from that set that must still be available after the eviction, even in the absence of the evicted pod. Can be either an absolute number or a percentage. Mutually Exclusive with `maxUnavailable`   | `1`     |
```
zmerlynn, line 263: Suggest adding same parenthetical and link as above: (Ignored unless `SplitControllerAndExtensions` enabled.), here and below.

zmerlynn (changes requested):
Really close!

install/helm/agones/templates/controller.yaml
```diff
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 {{- $featureGates := include "agones.featureGates" . | fromYaml }}
+{{- $replicas := .Values.agones.controller.replicas }}
```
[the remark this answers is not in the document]
chiayi: Do we have to ensure that replicas is an int here? It would fail the `gt` checks if it isn't an int. 
But I guess that would mean relying on the helm check to give an typing error instead of us stating the error earlier. This also applies to validating if `replicas` is greater than 0, it will also give an typing error when comparing with a 0.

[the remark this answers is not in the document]
chiayi: I see I see. You are correct, there is a need to cast to int. This also means that we would need to check if values is 0 once it's casted. What is the behavior we want when the value is set to 0? Do we want it to default to 1?

install/helm/agones/templates/controller.yaml
```diff
@@ -13,6 +13,13 @@
 # limitations under the License.
 
 {{- $featureGates := include "agones.featureGates" . | fromYaml }}
+{{- $replicas := (int .Values.agones.controller.replicas) }}
+{{- if not $featureGates.SplitControllerAndExtensions }}
+{{- $replicas = 1 }}
+{{- end }}
+{{- if eq $replicas 0 }}
+{{- fail "replica field has an invalid value!" }}
```
zmerlynn, line 21: nit: full path of `agones.controller.replicas`, like you would set it on the command line.

install/helm/agones/templates/controller.yaml
```diff
@@ -31,13 +38,13 @@ spec:
...
       heritage: {{ .Release.Service }}
-  replicas: 1
+  replicas: {{ $replicas }}
   strategy:
     type: Recreate
   template:
     metadata:
       annotations:
-{{- if not $featureGates.SplitControllerAndExtensions }}
+{{- if not (gt $replicas 1) }}
```
zmerlynn, line 47: nit: simplify: instead of !(replicas > 1), prefer (replicas == 1) - especially true if we reject <1 above.

install/helm/agones/templates/controller.yaml
```diff
@@ -13,6 +13,13 @@
 # limitations under the License.
 
 {{- $featureGates := include "agones.featureGates" . | fromYaml }}
+{{- $replicas := (int .Values.agones.controller.replicas) }}
+{{- if not $featureGates.SplitControllerAndExtensions }}
+{{- $replicas = 1 }}
+{{- end }}
+{{- if eq $replicas 0 }}
```
zmerlynn, line 20: paranoid nit: might as well check replicas < 1 in case someone tries something silly like a negative

zmerlynn (approved):
LGTM. Fix some nits, take off the WIP and I'll get it merged.

install/helm/agones/templates/controller.yaml
chiayi: Done

install/helm/agones/templates/controller.yaml
chiayi: done

install/helm/agones/templates/controller.yaml
chiayi: done

zmerlynn:
Your code LGTM, now just get it past the @agones-bot guantlet.

chiayi:
Build failed due to a different error than prev build: 
```
Step #24 - "submit-e2e-test-cloud-build": generic-1.25: Step #1 - "e2e-feature-gates": FAIL	agones.dev/agones/test/e2e	1800.722s
```

chiayi:
Found the issue. `server`, which had the health check was moved back into the `whenLeader` function which lead it to fail the install. Moved it back out as it's own go routine.

cmd/controller/main.go
```diff
@@ -239,6 +239,12 @@ func main() {
 		rs = append(rs, gasController)
 	}
 
+	go func(r runner) {
```
zmerlynn, line 242: Three things:
1. I would just inline this as:
```
go func() {
  if err := server.Run(ctx, [...]) [...]
}()
```

OR, you can make this a helper function and use it here and below, like:

```
runRunner := func(r runner) {
}
go runRunner(server)
```
(then later use the helper again)

2. Comment here why we need to start `server` early.

cmd/controller/main.go
```diff
@@ -447,10 +453,12 @@ type httpServer struct {
 
 func whenLeader(ctx context.Context, cancel context.CancelFunc, logger *logrus.Entry, doLeaderElection bool, kubeClient *kubernetes.Clientset, start func(_ context.Context)) {
 	if !doLeaderElection {
+		logger.Info("Leader election disabled!")
```
zmerlynn, line 456: I don't think we need these logs, we dump the `ctlConf` early in logs, e.g.:

```
{"ctlConf":{"MinPort":7000,"MaxPort":8000,"SidecarImage":"gcr.io/zml-gke-dev/agones-sdk:1.31.0-d98cd1b","SidecarCPURequest":"30m","SidecarCPULimit":"0","SidecarMemoryRequest":"0","SidecarMemoryLimit":"0","SdkServiceAccount":"agones-sdk","AlwaysPullSidecar":true,"PrometheusMetrics":true,"Stackdriver":false,"StackdriverLabels":"","KeyFile":"/home/agones/certs/server.key","CertFile":"/home/agones/certs/server.crt","KubeConfig":"","GCPProjectID":"","NumWorkers":100,"APIServerSustainedQPS":400,"APIServerBurstQPS":500,"LogDir":"/home/agones/logs","LogLevel":"debug","LogSizeLimitMB":10000,"AllocationBatchWaitTime":500000000,"LeaderElection":false},"featureGates":"CountsAndLists=false\u0026CustomFasSyncInterval=true\u0026Example=false\u0026FleetAllocationOverflow=false\u0026PlayerAllocationFilter=false\u0026PlayerTracking=false\u0026PodHostname=false\u0026ResetMetricsOnDelete=false\u0026SDKGracefulTermination=true\u0026SafeToEvict=true\u0026SplitControllerAndExtensions=true\u0026StateAllocationFilter=true","message":"starting gameServer operator...","severity":"info","source":"main","time":"2023-03-31T16:39:34.126611076Z","version":"1.31.0-d98cd1b"}
```

zmerlynn (commented):
LGTM, take it out of WIP?

cmd/controller/main.go
chiayi: turned it into a helper function

cmd/controller/main.go
chiayi: done!

chiayi:
Removed `--debug` from install

PR #3062, opened 2023-03-31
Have TestPlayerConnectWithCapacityZero use framework to wait

zmerlynn:
I'm seeing e2e failures in this test, which I'm trying to track down, but it seems like when this test fails, it explodes, e.g.:

```
panic: Fail in goroutine after TestPlayerConnectWithCapacityZero has completed [...]
 agones.dev/agones/test/e2e.TestPlayerConnectWithCapacityZero.func1()
     /go/src/agones.dev/agones/test/e2e/gameserver_test.go:1111 +0x2d3
```

It then ends up wrecking status of several other tests and makes the output noisier than it actually is. I think this is largely due to using `require`, which panics. Instead, we have a framework function that does exactly what this is looping on, just use that.


gongmax (approved):

markmandel:
FYI: `require` doesn't panic, it just stops the test at that point, rather than continuing on.

test/e2e/gameserver_test.go
```diff
@@ -1103,13 +1102,10 @@ func TestPlayerConnectWithCapacityZero(t *testing.T) {
...
 	_, err = framework.SendGameServerUDP(t, gs, msg)
 	// expected error from the log.Fatalf("could not connect player: %v", err)
-	require.Error(t, err)
-	assert.Eventually(t, func() bool {
-		gs, err = framework.AgonesClient.AgonesV1().GameServers(framework.Namespace).Get(ctx, gs.ObjectMeta.Name, metav1.GetOptions{})
-		require.NoError(t, err)
-
-		return assert.Equal(t, gs.Status.State, agonesv1.GameServerStateUnhealthy)
-	}, time.Minute, time.Second)
+	if assert.Error(t, err) {
```
markmandel, line 1105: This if statement is doing exactly the same thing `require.NoError` does 😄

zmerlynn:
> FYI: `require` doesn't panic, it just stops the test at that point, rather than continuing on.

Given that stack traces, it seems to be causing some nasty interaction, especially in the `Eventually` loop. The test report is showing that it seems to leave the test with a goroutine still running and then hard panics after. The panic is very much in testify assertion code.

test/e2e/gameserver_test.go
zmerlynn: And then panicking via FailNow: https://github.com/stretchr/testify/blob/v1.8.2/require/require.go#L1257-L1265

markmandel:
Might need to file a bug on testify!

test/e2e/gameserver_test.go
markmandel: Actually - digging into FailNow, actually explains things!!!
https://pkg.go.dev/testing#T.FailNow

(emphasis mine)

> FailNow marks the function as having failed and stops its execution by calling runtime.Goexit (which then runs all deferred calls in the current goroutine). Execution will continue at the next test or benchmark. **FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines**.

zmerlynn: Right.. and testify does doc that: https://github.com/stretchr/testify/tree/v1.8.2#require-package

Basically `require` is to `assert` as `t.Fatal` is to `t.Error`, AFAICT.

markmandel: The thing that doesn't make sense... is `Eventually` shouldn't be starting a go routine? Or does it?

zmerlynn: Yeah that one I didn't look up yet, but it sure does: https://github.com/stretchr/testify/blob/v1.8.2/assert/assertions.go#L1737

How weird!

zmerlynn: I think it's so the timeout has precedence over the condition - i.e. if you set a 1m timeout, it _always_ fires.

markmandel: ...actually, it 100% does.
https://github.com/stretchr/testify/blob/f36bfe3c337aa95c86f04c721acdbafb5ffb1611/assert/assertions.go#LL1737C37-L1737C37

This seems like a bug in testify to me tbh.

markmandel: See: https://github.com/stretchr/testify/issues/902

zmerlynn: Yeah.. the framework code just leans on https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait

PR #3065, opened 2023-04-03
GameServer Allocation Filtering for Counts and Lists

igooch:
**What type of PR is this?**

/kind feature

Adds CounterSelector and ListSelector for Game Server Allocation filtering as part of feature CountsAndLists.
- Allocation filter on count value (min, max)
- Allocation filter on count available capacity (min, max)
- Allocation filter on list available capacity (min, max)
- Allocation filter on if single value is contained in a list

Working on #2716 


pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -122,7 +132,26 @@ type PlayerSelector struct {
...
+	MinCount     int64 `json:"minCount"`
+	MaxCount     int64 `json:"maxCount"`
+	MinAvailable int64 `json:"minAvailable"`
+	MaxAvailable int64 `json:"maxAvailable"`
+}
+
+// ListSelector is the filter options for a GameServer based on List available capacity and/or the
+// existence of a value in a List.
+// Undefined or 0 for MaxAvailable means unlimited maximum.
+// Undefined or "" for ValueInList means ignore field. ValueInList is case insensitive.
```
gongmax, line 147: why case insensitive? Did we called out somewhere the values in a list are case insensitive?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -174,6 +212,65 @@ func (s *GameServerSelector) Matches(gs *agonesv1.GameServer) bool {
...
+	if runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		return s.matchCountsAndLists(gs)
+	}
+
+	return true
+}
+
+func (s *GameServerSelector) matchCountsAndLists(gs *agonesv1.GameServer) bool {
+	// Match Counter Selectors
+	if s.Counters != nil && gs.Status.Counters != nil {
```
gongmax, line 224: Does GameServer has `Status.Counters` default to an empty map when it doesn't have any counter? If not, the logic here will be problematic since it will return true for such a GameServer even the GameServerSelector is asking for some counters.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -226,6 +323,68 @@ func (s *GameServerSelector) Validate(field string) ([]metav1.StatusCause, bool)
 		}
 	}
 
+	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) && (s.Counters != nil || s.Lists != nil) {
```
gongmax, line 326: I'm fine with this but just want to call out this is a behavior change for customer. I.e. for the other existing selectors, if they are not featureEnabled, we just silently ignore them when performing Match without throwing any error back.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -226,6 +323,68 @@ func (s *GameServerSelector) Validate(field string) ([]metav1.StatusCause, bool)
...
+						Message: "CounterSelector.MinAvailable must be greater than zero",
+						Field:   field,
+					})
+				}
+				if counterSelector.MaxAvailable < 0 {
+					causes = append(causes, metav1.StatusCause{
+						Type:    metav1.CauseTypeFieldValueInvalid,
+						Message: "CounterSelector.MaxAvailable must be greater than zero",
+						Field:   field,
+					})
```
gongmax, line 363: Do we need to validate `Min* <= Max*`?

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: That's a good point. I recommend always defaulting to a more explicit system than explicit. If people want case insensitivity, then they can lowercase all their content before adding/checking/removing it 👍🏻

pkg/apis/allocation/v1/gameserverallocation.go
igooch: Sure, I can add that check in.

pkg/apis/allocation/v1/gameserverallocation.go
igooch: Makes sense, I'll change it to do a direct equality check.

pkg/apis/allocation/v1/gameserverallocation.go
igooch: Yes, it should default to nil, good catch on the logic.

pkg/apis/allocation/v1/gameserverallocation.go
igooch: I'm personally for more verbose help from the code, since it's easier to track down bugs than with silent failures. Do we have a pattern for logging warnings even if we don't error?

pkg/apis/allocation/v1/gameserverallocation.go
igooch: I'm now running into lint error on build "pkg/apis/allocation/v1/gameserverallocation.go:282:1: cyclomatic complexity 31 of func `(*GameServerSelector).Validate` is high (> 30) (gocyclo)". Any suggestions for reducing cyclomatic complexity besides taking out checks? Do you see any disadvantage to changing CounterSelector and ListSelector fields to uint so that I don't need to do any non-negative checks?

gongmax: So looks like the check is on method level? I'm wondering whether having a helper func that takes a single `counterSelector` to do the validation (line 341 to 382) and calling this helper func in the for loop (line 340) would help.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -122,7 +131,26 @@ type PlayerSelector struct {
...
+}
+
+// ListSelector is the filter options for a GameServer based on List available capacity and/or the
+// existence of a value in a List.
+// Undefined or 0 for MaxAvailable means unlimited maximum.
+// Undefined or "" for ValueInList means ignore field.
+type ListSelector struct {
+	MinAvailable int64  `json:"minAvailable"`
+	MaxAvailable int64  `json:"maxAvailable"`
+	ValueInList  string `json:"valueInList"`
```
markmandel, line 150: On the design we had `contains`, but I like how expressive this is. 🤔 WDYT of `containsValue` ? Since List is probably redundant since it's stored within a `lists` parent element anyway?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -114,6 +114,15 @@ type GameServerSelector struct {
 	// Players provides a filter on minimum and maximum values for player capacity when retrieving a GameServer
 	// through Allocation. Defaults to no limits.
 	Players *PlayerSelector `json:"players,omitempty"`
+	// (Alpha, CountsAndLists feature flag) Counters provides filters on minimum and maximum values
```
markmandel, line 117: Non blocking, but it might be nice to make the format match the ones above - but I don't actually mind 🤷🏻‍♂️ it was a format I 100% made up.

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: Actually, I think the GSA is being inconsistent with Fleets and GameServers -- on those we do actively throw a validation error if you attempt to use a field that is behind a feature flag.

So I would argue, let's set a consistency here that we should validate feature flagged fields and continue doing it forward. 

Sound good?

gongmax: SGTM

gongmax: SGT,

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -122,7 +131,26 @@ type PlayerSelector struct {
 	MaxAvailable int64 `json:"maxAvailable,omitempty"`
 }
 
-// ApplyDefaults applies default values to the PlayerSelector
+// CounterSelector is the filter options for a GameServer based on the count and/or available capacity.
+// Undefined or 0 for MaxCount or MaxAvailable means unlimited maximum.
+type CounterSelector struct {
+	MinCount     int64 `json:"minCount"`
```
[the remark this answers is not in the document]
igooch: "Count" makes sense to me since it corresponds to the GamerServerStatus CounterStatus.Count. I wouldn't use Value here to represent an integer since we use Value in ListSelector to denote a string in a list of strings (that corresponds to the GamerServerStatus ListStatus.Values). If you prefer less redundancy in naming should I change it to min?

pkg/apis/allocation/v1/gameserverallocation.go
igooch: Works for me

pkg/apis/allocation/v1/gameserverallocation.go
igooch: I copied the format from pkg/apis/agones/v1/gameserver.go. I can't remember off the top of my head, but I believe @zmerlynn switched to the new comment format for some reason.

markmandel: No biggie. Keep it as is 👍🏻

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -174,9 +212,78 @@ func (s *GameServerSelector) Matches(gs *agonesv1.GameServer) bool {
...
+	}
+
 	return true
 }
 
+func (s *GameServerSelector) matchCountsAndLists(gs *agonesv1.GameServer) bool {
+	match := false
+
+	// Match Counter Selectors
+	if s.Counters != nil && gs.Status.Counters != nil {
```
gongmax, line 231: Thinking about the case that `s.Counters` is not nil but `gs.Status.Counters` is nil, the match should be false. But in the current logic, match will be set to true by line 281 in this case if the required Lists got matched. Similar for the case where `s.Lists` is not nil but `gs.Status.Lists` is nil. Also better to have unit test coverage for these cases.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -174,9 +212,78 @@ func (s *GameServerSelector) Matches(gs *agonesv1.GameServer) bool {
...
+			if available < listSelector.MinAvailable ||
+				(listSelector.MaxAvailable != 0 && available > listSelector.MaxAvailable) {
+				return false
+			}
+			// Check if List contains ContainsValue (if a value has been specified)
+			if listSelector.ContainsValue != "" {
+				valueExists := false
+				for _, value := range listStatus.Values {
+					if value == listSelector.ContainsValue {
+						valueExists = true
```
gongmax, line 273: can break here

pkg/apis/allocation/v1/gameserverallocation.go
igooch: Oh yep, good catch. I'll update.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -174,9 +212,84 @@ func (s *GameServerSelector) Matches(gs *agonesv1.GameServer) bool {
...
+			}
+			// 0 means undefined (unlimited) for MaxCount.
+			if counterStatus.Count < counterSelector.MinCount ||
+				(counterSelector.MaxCount != 0 && counterStatus.Count > counterSelector.MaxCount) {
+				return false
+			}
+		}
+		return true
+	}
+	return false
```
gongmax, line 255: nitpick: To improve readability, this blob can be rewritten like:
~~~
if gs.Status.Counters == nil {
    return false
}

for counter, counterSelector : range s.Counters {
....
}

return true
~~~

No need to do `if s.Counters != nil` again since it's already checked.

gongmax (approved):

gongmax (approved):

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -122,7 +132,26 @@ type PlayerSelector struct {
...
+	MinCount     int64 `json:"minCount"`
+	MaxCount     int64 `json:"maxCount"`
+	MinAvailable int64 `json:"minAvailable"`
+	MaxAvailable int64 `json:"maxAvailable"`
+}
+
+// ListSelector is the filter options for a GameServer based on List available capacity and/or the
+// existence of a value in a List.
+// Undefined or 0 for MaxAvailable means unlimited maximum.
+// Undefined or "" for ContainsValue means ignore field.
```
markmandel, line 147: Same thing - what's "undefined" in this context?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -226,9 +339,109 @@ func (s *GameServerSelector) Validate(field string) ([]metav1.StatusCause, bool)
...
+		}
+		if s.Lists != nil {
+			causes = append(causes, s.validateLists(field)...)
+		}
+	}
+
 	return causes, len(causes) == 0
 }
 
+// validateCounters validates that the selection field has valid values for CounterSelectors
```
markmandel, line 362: That's a lot of tests 😄

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -122,7 +132,26 @@ type PlayerSelector struct {
 	MaxAvailable int64 `json:"maxAvailable,omitempty"`
 }
 
-// ApplyDefaults applies default values to the PlayerSelector
+// CounterSelector is the filter options for a GameServer based on the count and/or available capacity.
+// Undefined or 0 for MaxCount or MaxAvailable means unlimited maximum.
```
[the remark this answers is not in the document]
igooch: Undefined is when they don't include a value, so if they only pass in MinCount and MaxCount, then MinAvailable and MaxAvailable would be undefined and would default to 0. Is there a clearer (and not as verbose as this comment) way of saying that?

pkg/apis/allocation/v1/gameserverallocation.go
igooch: Same as above, when a field is not included in the request it will default to either 0 for an integer, or an empty string for a string.

pkg/apis/allocation/v1/gameserverallocation.go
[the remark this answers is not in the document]
markmandel: I would suggest:

```suggestion
// 0 for MaxCount or MaxAvailable means unlimited maximum. Default: 0
```

"Undefined" is a term we're not using anywhere else, so it sounds thing that's unique to this, so it gives me a "wait, is that something special? what don't I know?" moment as a reader.

pkg/apis/allocation/v1/gameserverallocation.go
igooch: Same tests as for player tracking, just a bunch more fields.

markmandel (approved):

PR #3067, opened 2023-04-04
DO NOT SUBMIT, running e2es with debug SDK logging

zmerlynn:
I'm tracking an issue where the SDK `/healthz` seems to be failing, want a whole e2e run with the SDK at debug (also trying locally).


zmerlynn:
Figured out how to repro locally, closing.

PR #3068, opened 2023-04-04
Revert "Rework game server health initial delay handling (#3046)"

zmerlynn:
Seeing a high incidence of SDK sidecar failures on Autopilot e2es, looking at the logs. Reverting for now as it seems related to this PR.

This reverts commit 60ed8cdac3def2b41f77adc2962c7ac65279bff9 / #3046.


gongmax:
This revert also disable tests on AP 1.26. Is that intended?

gongmax (approved):

zmerlynn:
> This revert also disable tests on AP 1.26. Is that intended?

Yup. It is definitely the case that without #3046, autopilot 1.26 is struggling. However, it seems like _with_ #3046, we have some other problem.

gongmax (approved):

PR #3070, opened 2023-04-05
Add timeout to SDK k8s client

zmerlynn:
This seems to help with (many of?) the flakes we're seeing in CI by forcing the informer to retry lists, rather than the SDK dying after 30s of hanging.


markmandel (approved):
Seems like a good change no matter what 👍🏻

PR #3072, opened 2023-04-05
Rework game server health initial delay handling

### LINKED ISSUE #2966: Agones health check shouldn't fail during game server container image pull
**What happened**:

The gameserver-sidecar health check fails while the game server container image is being pulled.

The health check should not fail at this time, allowing image pull to complete without terminating the game server.

Given the following example config, the health check will fail 105 seconds after start if the game server isn't healthy.

```
health:
    failureThreshold: 3
    initialDelaySeconds: 60
    periodSeconds: 15
```

Unfortunately, this doesn't account for the amount of time that it takes for the container image to be pulled, which can exceed 105 seconds in some cases.  For example, if an image pull takes 3 minutes, the health check will fail and attempt to terminate the pod after 105 seconds. 

This behavior traces back to Agones v1.19, when https://github.com/googleforgames/agones/issues/2355 was merged which altered the order of container startup so that Agones starts first and the game server starts second.  This causes the delay timer to start earlier than previously.

A workaround is to increase the `initialDelaySeconds` to a larger value, the longest we expect image pull to take.  This prevents the health check from failing and terminating the game server, but configuring the delay that large introduces a blind spot in monitoring.  If image pulls take less than the delay, such as in the case where the image already exists locally, then there is no health monitoring during the remaining duration until the `initialDelaySeconds` expires.

**Environment**:
- Agones version: v1.28
- Kubernetes version (use `kubectl version`): v1.23
- Cloud provider or hardware configuration: AWS
- Others:

zmerlynn:
This is a redrive of #3046, which was reverted in #3068

Rework health check handling of InitialDelaySeconds. See https://github.com/googleforgames/agones/issues/2966#issuecomment-1492346086:

* We remove any knowledge in the SDK of InitialDelaySeconds

* We remove the runHealth goroutine from main and shift this responsibility to the /gshealthz handler

Along the way:

*  I noted that the FailureThreshold doesn't need to be enforced on both the kubelet and SDK side, so in the injected liveness probe, I dropped that to 1. Previously we were waiting more probes than we needed to. In practice this is not terribly relevant since the SDK pushes it into Unhealthy.

* Close race if enqueueState is called rapidly before update can succeed

* Re-add Autopilot 1.26 to test matrix (removed in #3059)

Fixes #2966


pkg/gameservers/controller.go
```diff
@@ -860,6 +867,10 @@ func (c *Controller) syncGameServerRequestReadyState(ctx context.Context, gs *ag
 			break
 		}
 	}
+	// Verify that we found the game server container - we may have a stale cache where pod is missing ContainerStatuses.
+	if _, ok := gsCopy.ObjectMeta.Annotations[agonesv1.GameServerReadyContainerIDAnnotation]; !ok {
```
markmandel, line 871: 👍🏻

markmandel:
I don't see any issues - code looks good. I'll wait on CI before approving, see what we managed to fix!

markmandel (approved):
Let's do this

PR #3074, opened 2023-04-06
Add tags for cluster, location, commit to e2e-test builds

zmerlynn:
Tags will allow us to slice test failures more easily in Logs Explorer / BigQuery

Adding VERBOSE: allows us to split up the primary and verbose output in the BQ sink more easily.

Together, we should be able to extract test failures more easily, in theory.

Looks similar to this, but I took out the double dashes:

image


gongmax (approved):
Love it!

PR #3075, opened 2023-04-06
GitHub action for stale issues

geetachavan1:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
 /kind feature
> /kind hotfix

1. This PR will automate the process of identifying the stale issues which have been inactive more than 30 days
2. Wait for another 30 days before marking Obsolete
3. Wait for another 30 days. if still inactive, it will close the issue.
4. The issue carries "awaiting maintainer" label will be skipped and will remain open.
5. At every stage it will add a relevant comment.
6. To start with, it will run the action on the 1st and 15th of every month and will use the default value of 30 for operations-per-run. This will prevent it from applying stale labels on the all the issues at once.
7. It is using GitHub Actions CI/CD platform.

Closes #

If approved and merged, for the successful run we will need to create following labels

1. stale
2. obsolete
3. close
4. awaiting maintainer


.github/workflows/close.yaml
```diff
@@ -0,0 +1,34 @@
+# This workflow closes issues that have had no activity for 90 days.
```
markmandel, line 1: Same feedback as on #3047 - apache header on all files please.

.github/workflows/close.yaml
```diff
@@ -0,0 +1,34 @@
...
+    permissions:
+      issues: write
+      
+
+    steps:
+    - uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 30
+        close-issue-message: >
```
markmandel, line 26: Copy across feedback from #3047 

I think the better way to do this is to use `close-issue-message` and `days-before=close` on the obsolete workflow, since that is the workflow I see in the github action docs:

e.g. 
https://github.com/actions/stale/tree/main#close-issue-message:~:text=Configure%20different%20stale%20timeouts%3A

It seems like we could do this in two yaml files, rather than three.

.github/workflows/close.yaml
```diff
@@ -0,0 +1,34 @@
...
+          @author, We are closing this as there was no activity in this issue for last 90 days. Please reopen if you’d like to discuss anything further.
+        only-labels: 'obsolete'
+        exempt-issue-labels: awaiting-maintainer
+        ascending: true
+        days-before-close: 0
+        close-issue-reason: "not_planned"
+        stale-issue-label: 'closed'
```
markmandel, line 33: Can we add `wontfix` instead (https://github.com/googleforgames/agones/labels/wontfix) on close, since that is already ignored by the changelist, and aligns with what we already have?

.github/workflows/close.yaml
markmandel: We talked about this offline, ignore this concern.

.github/workflows/close.yaml
geetachavan1: Done

.github/workflows/close.yaml
geetachavan1: Done

markmandel (approved):

gongmax:
Will this workflow close the existing open issues which do not have activity for more than 90 days? Looking at the issue queue, most of them do not have activity for 90 days, my concern is the workflow should not close them silently without us triaging first.

geetachavan1:
> Will this workflow close the existing open issues which do not have activity for more than 90 days? Looking at the issue queue, most of them do not have activity for 90 days, my concern is the workflow should not close them silently without us triaging first.

@gongmax The workflow will close only the ones that carry 'obsolete' label and  to gain 'obsolete' label first it needs to get the 'stale' label. SO the short answer is NO.. No issues will be closed silently

gongmax:
> > Will this workflow close the existing open issues which do not have activity for more than 90 days? Looking at the issue queue, most of them do not have activity for 90 days, my concern is the workflow should not close them silently without us triaging first.
> 
> @gongmax The workflow will close only the ones that carry 'obsolete' label and to gain 'obsolete' label first it needs to get the 'stale' label. SO the short answer is NO.. No issues will be closed silently

Great!

PR #3076, opened 2023-04-06
Add e2e test for leader election

chiayi:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This is the e2e test that will test the controller leader election feature 

### DIFF
--- a/test/e2e/controller/crash_test.go
+++ b/test/e2e/controller/crash_test.go
@@ -20,8 +20,10 @@ import (
 	"time"
 
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
+	"agones.dev/agones/pkg/util/runtime"
 	e2eframework "agones.dev/agones/test/e2e/framework"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


test/e2e/controller/crash_test.go
```diff
@@ -58,6 +59,68 @@ func TestGameServerUnhealthyAfterDeletingPodWhileControllerDown(t *testing.T) {
...
+	replication := len(list.Items)
+
+	// Deleting one of the pods which cause the other one to most definitely be leader
+	willBeLeader := list.Items[0]
+	for i := range list.Items[1:] {
+		err = deleteAgonesControllerPod(ctx, list.Items[i])
+		require.NoError(t, err, "Could not delete controller pod")
+	}
+
+	err = deleteAgonesControllerPod(ctx, willBeLeader)
```
zmerlynn, line 88: It looks like you are:
* killing all replicas except the first
* then killing the first
* the waiting for everything to stabilize

Why? The way you've done it here, `willBeLeader` is pretty much irrelevant - you could have just killed all the pods.

The point of deleing the other pods is to shepherd the leader to `willBeLeader`. But you can't just kill it after doing that, otherwise you might as well have just restarted the whole replicaset.

What we talked about offoine was close to this, but not exactly:
* delete all but one as you do in line 82-86
* wait for all running (separately, just add a parameter to waitForAgonesControllerRunning if you want to ensure the count - you can add a wantReplicas parameter and then only check it if it's non-negative or something)
* then delete `willBeLeader`
* then proceed to verify the controller till works, without waiting again

the point is that shepherding it into one replica, waiting for the others to come back, then killing that one replica, is a pretty solid guarantee that you're killing the leader. The way you have it here also guarantees you killed the leader but it then waits for everything to come back up - we ideally want to show that right after the leader dies, we can still function. Make sense?

chiayi: Ooh I see I see. I think I have a better understanding now. I will try this.

zmerlynn (approved):
I think you can take it out of WIP, LGTM!

chiayi:
Seemed to have timed out waiting for condition during test: `TestGameServerAllocationDuringGameServerDeletion/rolling_update`. Going to re-run to see if it's a flake.

PR #3088, opened 2023-04-12
Update docs for Stable Network ID

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

Attempting to alleviate some confusion about Stable Network ID, as it read as if the feature would enable a feature like the `ExternalDNS` provider -- whereas it's actually only for use within the cluster.

N/A

N/A


rcreasey (approved):
This helped clarify the situation for our use case. 👍

gongmax (approved):

PR #3091, opened 2023-04-12
Sort By Counters or Lists during GameServerAllocation 2716

igooch:
**What type of PR is this?**
/kind feature

- Adds Priorities "Counter" and "List" that allow Game Servers to be sorted by any Counters or Lists for allocation in a Descending or Ascending manner, where Descending (3, 2, 1, 0, nil) is the default. 
- The Counter or List sort sorts by the first Priority in the Priorities list. Any subsequent Priority(s) is only used as a tie-breaker in the order in which they are in Priorities list.
- Sorting by Counters or Lists at Allocation happens only as a tie-breaker for equal game servers after preferring game servers for state Allocated, being a node that is most allocated, and PlayerTracking if that feature is enabled.
- Updates the final tie-breaker for sorting to sort by Game Server name instead of NodeName.

 
Working on #2716


pkg/gameserverallocations/allocator.go
```diff
@@ -509,11 +510,23 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
 				requestCount = 0
 			}
 
+			// FeatureCountsAndLists: Sorting based on GameServerAllocation Priority only for strategy Distributed.
```
markmandel, line 513: So another suggestion that may make this easier to manage.

Duplicate `ListenAndAllocate` into a `ListenAndAllocateWithListsAndCounter` (or duplicate the old version, whichever way works).

Then the feature branch when disabled can be to run `ListenAndAllocate` and the feature branch when enabled could be to run `ListenAndAllocateWithListsAndCounter`.

Then you don't have to run as many intricate if/else statements -- and you don't have to worry about breaking existing systems either.

Then you can also add new functions to `allocation_cache.go` that are probably almost exact copies of `ListSortedGameServers` etc, but again, less if/else branching and probably easier to reconcile.

How does that sound?

pkg/gameserverallocations/find.go
```diff
@@ -40,6 +40,7 @@ func findGameServerForAllocation(gsa *allocationv1.GameServerAllocation, list []
 	var loop func(list []*agonesv1.GameServer, f func(i int, gs *agonesv1.GameServer))
 
 	// packed is forward looping, distributed is random looping
+	// TODO: Discuss if we change findGameServerForAllocation so all strategies use forward looping for the the pre-sorted list for each strategy once FeatureCountsAndLists gate is graduated.
```
markmandel, line 43: For the new version, I think we can throw most of this complexity away, and do simple search and filter -- we don't need this fancy looping construct anymore, since it will be pre-sorted / pre-randomised.

(Also can then also do a duplicated function as well for counts and lists)

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -484,6 +491,41 @@ func (mp *MetaPatch) Validate() ([]metav1.StatusCause, bool) {
...
 }
 
+// Priority is a sorting option for GameServers with Counters or Lists based on the count or
+// number of items in a List.
+// Sort: Sort by a "counter" or a "list".
+// Key: The name of the Counter or List. If not found on the GameServer, has no impact.
+// Order: Sort by "ascending" or "descending". Default is "descending" so bigger count is preferred.
+// "ascending" would be smaller count is preferred.
+type Priority struct {
+	Sort  string `json:"sort"`
```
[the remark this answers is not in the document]
igooch: We can't (or shouldn't) use "Type" because it's a key word in Golang. Maybe "SortType" or "Order" or "OrderType"?

pkg/gameserverallocations/allocation_cache.go
```diff
@@ -231,13 +232,83 @@ func (c *AllocationCache) ListSortedGameServers() []*agonesv1.GameServer {
...
+	list := c.getGameServers()
+	if list == nil {
+		return []*agonesv1.GameServer{}
+	}
+
+	sort.Slice(list, func(i, j int) bool {
+		gs1 := list[i]
+		gs2 := list[j]
+
+		for _, priority := range gsa.Spec.Priorities {
```
gongmax, line 255: Looks like we are literally only sorting GameServers according to the the first `Priority` of the `gsa.Spec.Priorities` list? So why the input of Allocation has a list of `Priority`? Do we want customer to specify a list of `Priority` that should be considered during allocation? If so, should we treated the list weighted? i.e. sorting GameServers based on the Priority at the beginning of the list, and only looking at the subsequent Priorities to break tie?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -484,6 +491,41 @@ func (mp *MetaPatch) Validate() ([]metav1.StatusCause, bool) {
...
+	}
+
+	if !(p.Order == "ascending" || p.Order == "descending" || p.Order == "") {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "Priority.Order must be either `ascending` or `descending`",
+			Field:   field,
+		})
+	}
+
```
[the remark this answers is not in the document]
igooch: I don't think the Key can be nil?

pkg/gameserverallocations/allocation_cache.go
```diff
@@ -231,13 +232,83 @@ func (c *AllocationCache) ListSortedGameServers() []*agonesv1.GameServer {
...
+					return false
+				}
+				// If neither gameserver has the Priority, contiue on to the next Priority.
+			}
+		}
+
+		// TODO: determine what other checks to include (Status, Node, etc.) and if they should be checked before or after CountsAndList Priority checks
+
+		// TODO: determine default if neither game server has any Priority
+		return true
```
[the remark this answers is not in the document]
igooch: This method sorts for the Distributed case when CountsAndLists is enabled. The logic for routing currently lives in pkg/gameserverallocations/allocator.go, although per Mark's comment in allocator.go I'll probably rework that logic flow. 

CountsAndLists uses the existing ListSortedGameServers() for the Packed case.

pkg/gameserverallocations/allocation_cache.go
igooch: The first Priority in the list is considered, and only in the case of a tie is a subsequent priority(s) considered. I'm currently pulling out this logic into a separate comparator.

pkg/apis/allocation/v1/gameserverallocation.go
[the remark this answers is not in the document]
gongmax: Right, I meant an empty string. Is it a valid key? I forgot whether the key of a Counter/List can be empty.

pkg/gameserverallocations/allocation_cache.go
gongmax: SG. We should explicitly called out in doc and API reference that the Priorities are weighted.

pkg/apis/allocation/v1/gameserverallocation.go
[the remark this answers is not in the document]
igooch: An empty string is technically a valid key.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -484,6 +491,53 @@ func (mp *MetaPatch) Validate() ([]metav1.StatusCause, bool) {
...
+	if !(p.SortType == Counter || p.SortType == List) {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "Priority.Sort must be either `counter` or `list`",
+			Field:   field,
+		})
+	}
+
+	if !(p.Order == Ascending || p.Order == Descending || p.Order == "") {
```
markmandel, line 530: I'm assuming Order will be defaulted, so checking for "" is probably not something we need to do? (or want to if we have defaulting)

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -484,6 +491,53 @@ func (mp *MetaPatch) Validate() ([]metav1.StatusCause, bool) {
 	return causes, len(causes) == 0
 }
 
+// Counter is a Priority SortType
+const Counter string = "counter"
```
markmandel, line 495:
```suggestion
const GameServerAllocationCounter string = "Counter"
```

Nit: Have all const values at the top of the file.

Since the package space is shared.

Also just realised a inconsistency in the design - values are always capitalised. "Packed", "Rolling", "Ready" etc. We should be consistent with the rest of Agones here too.

(Same for below)

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -484,6 +491,53 @@ func (mp *MetaPatch) Validate() ([]metav1.StatusCause, bool) {
...
+}
+
+// Validate returns if the Priority is valid.
+func (p *Priority) validate(field string) ([]metav1.StatusCause, bool) {
+	var causes []metav1.StatusCause
+
+	if !(p.SortType == Counter || p.SortType == List) {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "Priority.Sort must be either `counter` or `list`",
```
markmandel, line 525:
```suggestion
			Message: "Priority.Sort must be either `Counter` or `List`",
```

That's my bad, I wrote a bad design doc.

pkg/gameserverallocations/allocation_cache.go
```diff
@@ -231,13 +232,111 @@ func (c *AllocationCache) ListSortedGameServers() []*agonesv1.GameServer {
...
+
 		// finally sort lexicographically, so we have a stable order
 		return gs1.Status.NodeName < gs2.Status.NodeName
 	})
 
 	return list
 }
 
+// compareGameServers compares two game servers based on a CountsAndLists Priority
+// Returns -1 if gs1 < gs2; 1 if gs1 > gs2; 0 if gs1 == gs2; 0 if neither gamer server has the Priority.
```
markmandel, line 245: Nice - that's a better way of doing it!

pkg/gameserverallocations/allocation_cache.go
markmandel: I like the trick - if `compareGameServers` returns 0, then it passes through to the next one. I like it 👍🏻

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -484,6 +491,53 @@ func (mp *MetaPatch) Validate() ([]metav1.StatusCause, bool) {
...
+const Descending string = "descending"
+
+// Priority is a sorting option for GameServers with Counters or Lists based on the count or
+// number of items in a List.
+// SortType: Sort by a "counter" or a "list".
+// Key: The name of the Counter or List. If not found on the GameServer, has no impact.
+// Order: Sort by "ascending" or "descending". Default is "descending" so bigger count is preferred.
+// "ascending" would be smaller count is preferred.
+type Priority struct {
+	SortType string `json:"sortType"`
```
markmandel, line 513: > We can't (or shouldn't) use "Type" because it's a key word in Golang. Maybe "SortType" or "Order" or "OrderType"?

That is an exceptional point. 

`PriorityType` - since it's part of priority? Seems like an easy enough keyword escaping.

```suggestion
	PriorityType string `json:"type"`
```

Which does solve it for mapping - but maybe is just waiting for a footgun to go off ("type" as a json mapping)?

or

```suggestion
	PriorityType string `json:"priorityType"`
```

Either works for me.

I could work with `OrderType` as - since then it aligns with `Order` below it.

So I'd say either align with the parent `Priorities` or the neighbour element `Order`. Either works for me 👍🏻

pkg/apis/allocation/v1/gameserverallocation.go
igooch: I haven't done any defaulting. Where in the code should I do defaulting for this?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -514,6 +568,13 @@ func (gsa *GameServerAllocation) ApplyDefaults() {
 			gsa.Spec.Selectors[i].ApplyDefaults()
 		}
 	}
+
+	if runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		// TODO: Do we need to make an empty slice of Priorities here if gsa.Spec.Priorities == nil?
```
igooch, line 573: @markmandel Any idea if I need to create an empty slice of Priorities here if gsa.Spec.Priorities == nil? That seems to be a common pattern, but not sure if it's necessary here?

pkg/gameserverallocations/find.go
igooch: Right now we have three paths A) Sort by Allocated and then loop through list. B) Sort by Allocated, then randomly shuffle list, and then loop through randomized list. C) Sort by CountsAndLists and then loop through list. 

Where/when should I be randomizing? Should I make a 4th path D) Sort by CountsAndLists, then randomly shuffle list, and then loop through randomized list? Or are you suggesting a different set of paths?

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: If you `range` over a nil array I'm pretty sure it's fine (and `append` works fine too) - so probably not.

Time will tell however.

pkg/gameserverallocations/find.go
markmandel: My thought is this - for this part of the code -- the list should already be appropriately sorted when it is passed to it. So it should be a `for .. range` operation.

Then for each strategy:

_Distributed_
1. Randomise first (so that will give us a random distribution within each sorted Allocated/Ready section of the list we have)
2. Sort by:
  3. Allocated
  4. Count/List as appropriate

Question on the above: Does K8s List already give us a generally random order of GameServers, such that we don't need to bother shuffling it? I'm actually not sure! May be worth checking.

_Packed_
Sort by:
1. Allocated First
2. Packed Strategy (basically what we have now)
3. Tie break with counts and lists

I hope that makes things clearer?

pkg/apis/allocation/v1/gameserverallocation.go
igooch: OK, I'll delete these lines in that case.

pkg/gameserverallocations/allocation_cache.go
```diff
@@ -231,8 +232,129 @@ func (c *AllocationCache) ListSortedGameServers() []*agonesv1.GameServer {
...
+					if res == 1 {
+						return true
+					}
+				}
+			}
+		}
+
 		// finally sort lexicographically, so we have a stable order
-		return gs1.Status.NodeName < gs2.Status.NodeName
+		return gs1.GetObjectMeta().GetName() < gs2.GetObjectMeta().GetName()
```
igooch, line 259: @markmandel I changed this to the ObjectMeta.Name instead of the Status.NodeName because the NodeName is not unique, so this didn't create a stable order. (I discovered this during test TestAllocationCacheListSortedGameServers.) This change assumes that both 1) all Game Servers have ObjectMeta.Name, and 2) All ObjectMeta.Name are unique. Are these correct assumptions? If those assumptions are not correct, then this change would also not create a stable order. Let me know if this change would cause more issues than improvements. (Or if we should just leave it out for the sake of not making changes that are not behind a feature gate.)

igooch: @markmandel I changed this to the ObjectMeta.Name instead of the Status.NodeName because the NodeName is not unique, so this didn't create a stable order. (I discovered this during test TestAllocationCacheListSortedGameServers as gs2 and gs4 are the same when comparing by NodeName.) This change assumes that both 1) all Game Servers have ObjectMeta.Name, and 2) All ObjectMeta.Name are unique. Are these correct assumptions? If those assumptions are not correct, then this change would also not create a stable order. Would it be better to use the ObjectMeta.UID? Or would it be better to leave things as-is? Or should we leave it out for the sake of not making changes that are not behind a feature gate?

pkg/gameserverallocations/find.go
igooch: Why randomize before sorting? I thought we had the entire list of Ready/Allocated game servers here, and not a subset?

If we're keeping our current definitions of Packed and Distributed, then it seems to make more sense to more or less keep the flow as-is:
1.0 Sort by Allocated
1.1 Tie break sort by number of Ready servers on the Node
1.2 Tie break sort by Counts and Lists
1.3 Tie break sort by Game Server Name <- Separate discussion on this as it currently tie breaks on NodeName
2.0 If Distributed, randomize list
2.1 Loop through list to find matching Game Servers (original sorted list for Packed, randomized list for Distributed) 

A possible alternative would to first check for matches before sorting the Game Servers, and only sort for the Packed case:
1. Filter List of Game Servers for Matches
2.a Return unsorted list of Matches for Distributed (completely random, no sorting) <- I'd need to ask an expert in Go if this is dependably random, as this depends on how c.getGameServers() ranges through the cache map of Game Servers
2.b Sort List (same as above) and return sorted list of Matches for Packed

pkg/gameserverallocations/allocator.go
```diff
@@ -510,10 +510,13 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
...
 
 			if list == nil {
-				list = c.allocationCache.ListSortedGameServers()
+				list = c.allocationCache.ListSortedGameServers(req.gsa)
 			}
 
-			gs, index, err := findGameServerForAllocation(req.gsa, list)
+			var gs *agonesv1.GameServer
+			var index int
+			var err error
```
gongmax, line 518: Nit: why make the change?

gongmax (commented):
Can we add unit test coverage for the new validate logic added in `gameserverallocation.go`?

pkg/gameserverallocations/allocator.go
igooch: Forgot to change it back :)

pkg/gameserverallocations/find.go
gongmax: With the above being said, I think the current `findGameServerForAllocation` totally works fine. We can leave all the refactoring outside this PR.

gongmax (approved):

pkg/gameserverallocations/allocation_cache.go
markmandel: Those assumptions are correct - so this sounds like a good change 👍🏻

markmandel (approved):
Looks like a good set of initial changes 👍🏻 LGTM!

PR #3092, opened 2023-04-13
Drop log level of worker queue to Trace

zmerlynn:
The "Enqueueing" logs in particular are pretty spammy, and the conflict log is a normal operation.

Keeps "Processing", which is also a bit spammy but does help delineate when the controller acted on a change.


markmandel (approved):

zmerlynn:
Approvers: Leave this PR out of date. We're using it as a canary for @agones-bot auto-updating - next merge to `main` should tickle it. (This is a low priority PR.)

PR #3103, opened 2023-04-14
Refactor: Modify logger implementation and log level

aimuz:
admission webhook "validations.agones.dev" denied the request: FleetAutoscaler is invalid
I encountered this error in my use and did not return any details to me, I would like to see him in the agones log

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind cleanup

Closes #

### DIFF
--- a/pkg/fleetautoscalers/controller.go
+++ b/pkg/fleetautoscalers/controller.go
@@ -193,16 +193,24 @@ func (c *Controller) Run(ctx context.Context, workers int) error {
 	return nil
 }
 
-func (c *Controller) loggerForFleetAutoscalerKey(key string) *logrus.Entry {
-	return logfields.AugmentLogEntry(c.baseLogger, logfields.FleetAutoscalerKey, key)
+func loggerForFleetAutoscalerKey(key string, logger *logrus.Entry) *logrus.Entry {
+	return logfields.AugmentLogEntry(logger, logfields.FleetAutoscalerKey, key)
 }
 
-func (c *Controller) loggerForFleetAutoscaler(fas *autoscalingv1.FleetAutoscaler) *logrus.Entry {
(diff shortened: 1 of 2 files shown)


aimuz:
Thanks for the reply, I am trying to integrate agones with our internal platform, during the integration I created FleetAutoscaler and Fleet using agones' client-go, via err.Error() I get "Invalid FleetAutoscaler " error, at this point I don't know exactly what field is causing the error.

https://github.com/googleforgames/agones/blob/main/pkg/fleets/controller.go#L215

I would like to add a log where I can look at the agones log and get detailed information, just like there are logs for Fleet creation or update failure.

I'm not sure if it's agones or client-go, in other types like Deployment, or Ingress during creation or update failures, I can get the details via err.Error(). But here I did not go deeper into the investigation at the moment.

markmandel:
> I'm not sure if it's agones or client-go, in other types like Deployment, or Ingress during creation or update failures, I can get the details via err.Error(). But here I did not go deeper into the investigation at the moment.

It sounds a lot like we should improve the error message -- we pass up the `StatusDetails` with a list of causes -- so in _theory_ it should all be there.

I'll note in out controllers we `Debug` the admissionReview, but we don't `Warn` it -- that could get a little verbose.

aimuz:
```
loggerForFleet(fleet, ext.baseLogger).WithField("review", review).Warn("Invalid Fleet")
```

Does this section also need to be changed to Debug?

markmandel:
Yeah - that;s inline with what we do elsewhere.

aimuz:
Please review, I changed it to Debug

zmerlynn (approved):
LGTM - thanks @aimuz

aimuz:
I'm surprised that the build error seems to occur frequently lately, but after re-pushing it works again

aimuz:
rebase

PR #3107, opened 2023-04-17
Revert #3070, wait on networking a different way

### LINKED ISSUE #3106: Excessive WATCH load from Agones 1.31
I thought I had done this exact analysis but I didn't see this load in a sea of other data: The change in go/aggh/3070 has a flaw which results in us re-WATCHing every 3s. Which is my bad - I thought that the timeout would cover only the header portion of watches, not the entire hanging-GET.

Attached is an internal graph which shows how much the `WATCH` verb spiked after #3070 was submitted. 😦 
image

zmerlynn:
The problem addressed by #3070 is that on an indeterminate basis, we are seeing containers start without networking fully available. Once networking seems to work, it works fine.

However, the fix in #3070 introduced a downside: heavy watch traffic, because I didn't quite understand that it would also block the hanging GET of the watch. See #3106.

Instead of timing out the whole client, let's use an initial-probe approach and instead block on a successful GET (with a reasonable timeout) before we try to start informers.

Fixes #3106


pkg/sdkserver/sdkserver.go
```diff
@@ -263,6 +267,38 @@ func (s *SDKServer) Run(ctx context.Context) error {
 	return nil
 }
 
+// waitForConnection attempts a GameServer GET every 3s until the client responds.
```
markmandel, line 270: Amusingly this matches the exact pattern we use on SDK clients that don't support gRPC.

pkg/sdkserver/sdkserver.go
```diff
@@ -263,6 +267,38 @@ func (s *SDKServer) Run(ctx context.Context) error {
...
+
+		// Specifically use gameServerGetter since it's the raw client (gameServerLister is the informer).
+		// We use List here to avoid needing permission to Get().
+		_, err := s.gameServerGetter.GameServers(s.namespace).List(ctx, metav1.ListOptions{
+			FieldSelector: fields.OneTermEqualSelector("metadata.name", s.gameServerName).String(),
+		})
+		return err
+	}
+
+	for try := 0; ctx.Err() == nil; try++ {
```
markmandel, line 290: (non blocking): This might be nicer as something like https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout rather than a for loop - since it handles context cancellation etc for you.

Just a thought. Happy to approve otherwise, but figured I'd make the suggestion to see if it resonated.

zmerlynn: Done. I settled on PollImmediateInfiniteWithContext, which will be deprecated, but isn't yet in the apimachinery version we have (but we also don't have its replacement yet).

zmerlynn:
@markmandel I just realized, I thought since you left a non-blocking comment that you can approved it. If you get a sec, PTAL. Or @gongmax or @roberthbailey if you get a chance - thanks!

gongmax (approved):

PR #3114, opened 2023-04-20
Push-Chart to Helm Repo on GCS

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind feature

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -282,7 +282,7 @@ build-chart: $(ensure-build-image)
 
 # push the current chart to google cloud storage and update the index
 push-chart: $(ensure-build-image) build-chart
-	docker run --rm $(common_mounts) -w $(workdir_path) $(build_tag) bash -c \
+	docker run --network=cloudbuild --rm $(common_mounts) -w $(workdir_path) $(build_tag) bash -c \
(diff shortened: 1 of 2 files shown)


build/Makefile
```diff
@@ -282,7 +282,7 @@ build-chart: $(ensure-build-image)
 
 # push the current chart to google cloud storage and update the index
 push-chart: $(ensure-build-image) build-chart
-	docker run --rm $(common_mounts) -w $(workdir_path) $(build_tag) bash -c \
+	docker run --network=cloudbuild --rm $(common_mounts) -w $(workdir_path) $(build_tag) bash -c \
```
markmandel, line 285: Let's set this up the same as we have other targets, so have it as:

```suggestion
	docker run $(DOCKER_RUN_ARGS) --rm $(common_mounts) -w $(workdir_path) $(build_tag) bash -c \
```

And then pass in the `--network=cloudbuild` from the post_cloudbuild.yaml

Kalaiselvi84: I got it done.

gongmax (approved):

Issue #3115, opened 2023-04-20
Slow node address update results in spurious migration event


zmerlynn:
Reported [on Slack](https://agones.slack.com/archives/C9DGM5DS8/p1681897403607789):

On a GKE private cluster with autoscaling, user is seeing "mass migration events" of `GameServers`, with `Node migration occurred` in the event stream. The problem is here: https://github.com/googleforgames/agones/blob/a5af67743f1ff0dbc21b8d1d19c207a8fd505499/pkg/gameservers/migration.go#L182-L188

The `address()` derived here is identical to how it was populated in the `GameServer` in the first place: https://github.com/googleforgames/agones/blob/3c38876382d505244af68489211d8916dce07596/pkg/gameservers/gameservers.go#L37-L43

But for some reason (that I haven't reproduced yet), on the user's cluster, the `InternalDNS` entry is late-bound on the `Node` object, meaning that when the `GameServerController` populates the `GameServer` address, it sometimes populates the `InternalIP` and sometimes populates the `InternalDNS`, but then later the `address()` call in `MigrationController` picks up `InternalDNS` and thinks it's a migration.

I haven't quite found the set of conditions to repro this, but there is a relatively simple fix: In `MigrationController`, tolerate any address we find on the `Node`.

markmandel:
> But for some reason (that I haven't reproduced yet), on the user's cluster, the `InternalDNS` entry is late-bound on the `Node` object

That seems like a bigger issue - as otherwise there's no guarantee which type of address you would get on the node.

#3089 might actually be the answer?

zmerlynn:
#3089 is definitely _an_ answer, but I don't see a compelling reason we can't make the MigrationController robust to this scenario in the first place.

In my repro attempts we almost always assign an IP and DNS name together, so I haven't narrowed down where that delay is coming from.

markmandel:
From memory the migration controller checks name and address, as there no guarantee that a node _won't_ get the same name as a previously created and then deleted node - so better to check both name and address details (although addresses could be reused as well - but getting both seems very rare) -- maybe we should look at a node UID or something instead.

Also... how is a node getting new addresses? That's a neat trick.

zmerlynn:
@markmandel The MigrationController checks for the address the same way it's generated, which has a preference order. See #3116 for a proposed fix by @luckyswede. :)

markmandel:
That's a clever approach - check all the addresses 😄

PR #3116, opened 2023-04-21
Make migration controller more forgiving wrt Node/GameServer addresses

luckyswede:
**What type of PR is this?**
/kind bug

This is a workaround for issue #3115. 

#3115 

This change does not fix the root cause where node addresses sometimes include a DNS name and sometimes doesn't, it only makes sure to not initiate a migration event in the case where any node address match the gameservers address.

### DIFF
--- a/pkg/gameservers/migration.go
+++ b/pkg/gameservers/migration.go
@@ -16,6 +16,8 @@ package gameservers
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"agones.dev/agones/pkg/apis/agones"
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
@@ -30,6 +32,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	k8sv1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
@@ -179,12 +182,8 @@ func (mc *MigrationController) syncGameServer(ctx context.Context, key string) e
 	if err != nil {
 		return errors.Wrapf(err, "error retrieving node %s for Pod %s", pod.Spec.NodeName, pod.ObjectMeta.Name)
 	}
-	address, err := address(node)
-	if err != nil {
-		return err
-	}
 
-	if pod.Spec.NodeName != gs.Status.NodeName || address != gs.Status.Address {
+	if pod.Spec.NodeName != gs.Status.NodeName || !mc.anyAddressMatch(node, gs) {
 		gsCopy := gs.DeepCopy()
 
 		var eventMsg string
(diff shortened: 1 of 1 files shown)


pkg/gameservers/migration.go
```diff
@@ -211,3 +210,17 @@ func (mc *MigrationController) syncGameServer(ctx context.Context, key string) e
 
 	return nil
 }
+
+func (mc *MigrationController) anyAddressMatch(node *k8sv1.Node, gs *agonesv1.GameServer) bool {
```
markmandel, line 214: Nice! I like it.

One addition I'd like to see is a unit test for `anyAddressMatch` so we can check a few different variations of input.

pkg/gameservers/migration_test.go
```diff
@@ -137,6 +137,7 @@ func TestMigrationControllerSyncGameServer(t *testing.T) {
 					assert.Equal(t, ipChangeFixture, gs.Status.Address)
 				},
 				postTests: func(t *testing.T, m agtesting.Mocks) {
+					agtesting.AssertEventContains(t, m.FakeRecorder.Events, "Warning Scheduled GameServer/Node address mismatch: gs.Name=test: gs.Status.Address=12.12.12.12, node.Status.Addresses=99.99.99.99")
```
markmandel, line 140: Does this mean we get two events on a node migration?

Do we need an extra event, or can it all be combined?

zmerlynn: I think I would just drop this to a log message (which is I think what you had originally, but I might be mis-remembering).

zmerlynn (commented):
Thanks for the fix!

pkg/gameservers/migration.go
```diff
@@ -211,3 +210,17 @@ func (mc *MigrationController) syncGameServer(ctx context.Context, key string) e
...
+	nodeAddresses := []string{}
+	for _, a := range node.Status.Addresses {
+		if a.Address == gs.Status.Address {
+			return true
+		}
+		nodeAddresses = append(nodeAddresses, a.Address)
+	}
+	msg := fmt.Sprintf("GameServer/Node address mismatch: gs.Name=%s: gs.Status.Address=%s, node.Status.Addresses=%s",
+		gs.Name, gs.Status.Address, strings.Join(nodeAddresses, ","))
+	mc.loggerForGameServer(gs).Warn(msg)
```
zmerlynn, line 224: One minor nit: With `logrus`, we tend to use fields for all of the data pieces, e.g.:

```suggestion
	mc.loggerForGameServer(gs).WithField("gs", gs.Name).WithField("gs.Status.Address", gs.Status.Address).WithField("node.Status.Addresses", strings.Join(nodeAddresses, ",")).Warn("GameServer/Node address mismatch")
```

(note that `gs` is already a common field name, so I used that instead of calling the field `gs.Name`, but the rest I copied from your log message).

zmerlynn (approved):
Thanks for your patience, and your contribution!

zmerlynn:
@luckyswede You'll need to rebase or merge - I'm not sure what permission is missing, but I can't bring your branch up to date. (Normally we have the power to update contributors branches, but not here.)

luckyswede:
I merged latest from main now

zmerlynn:
@luckyswede Looks like you're next in e2e queue anyways, so should merge in the next ~2h.

PR #2961, opened 2023-02-08
Add documentation for SplitControllerAndExtensions

chiayi:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

**What this PR does / Why we need it**:
This is part of the documentation for the feature `SplitControllerAndExtensions`. 

**Which issue(s) this PR fixes**:
Work on https://github.com/googleforgames/agones/issues/2797

**Special notes for your reviewer**:

PR #3125, opened 2023-04-26
Added labels to the agones.allocator

### LINKED ISSUE #3093: Helm option: agones.allocator.labels
We have `agones.allocator.annotations` as helm options for the Allocator, but not `agones.allocator.labels` .

This is useful when using a cross cluster service mesh, such as Istio/Anthos Service Mesh and want to be able to use the label to route requests via a single domain.

In the [global scale demo](https://github.com/googleforgames/global-multiplayer-demo) we had to use kustomize to patch the yaml config generated from helm. It would be nice not to have to do that.

See: https://github.com/googleforgames/global-multiplayer-demo/blob/main/infrastructure/files/agones/agones-install.yaml.tpl

(Yes, that Skaffold config generated by Terraform, that will get passed to Cloud Deploy).

 `agones.allocator.labels`  that adds labels to the Deployment of the allocator.

Keep using Kustomize, but it's an extra unnecessary step.

**Additional context**

Using a cross cluster service mesh is actually a _really_ nice way to do multi-cluster allocation!

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Closes #3093


install/helm/agones/values.yaml
```diff
@@ -35,7 +35,7 @@ agones:
   serviceaccount:
     allocator:
       name: agones-allocator
-      annotations: {}
+      labels: {}
```
markmandel, line 38: We need both annotations and labels please.

Also, we'll need updates to the documentation as well.

Kalaiselvi84: I included both..

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -201,6 +201,7 @@ The following tables lists the configurable parameters of the Agones chart and t
...
 | `agones.allocator.affinity`                              | Allocator [affinity][affinity] settings for pod assignment                                                                                                                                                              | `{}`                               |
 | `agones.allocator.annotations`                           | [Annotations][annotations] added to the Agones allocator pods                                                                                                                                                           | `{}`                               |
+| `agones.allocator.labels`                           | [Labels][labels] added to the Agones allocator pods                                                                                                                                                           | `{}`                               |
```
markmandel, line 204: This table will need to added to the feature shortcoded table at the bottom so they don't show up in documentation when merged into `main` (docs get updated immeadiately)

https://github.com/googleforgames/agones/pull/2961 is a good example

markmandel (commented):
Small docs change, but otherwise good to go.

site/content/en/docs/Installation/Install Agones/helm.md
Kalaiselvi84: I've updated the helm file, please review and let me know if there's anything that needs further attention..

install/helm/agones/values.yaml
```diff
@@ -36,6 +36,7 @@ agones:
     allocator:
       name: agones-allocator
       annotations: {}
+      labels: {}
```
markmandel, line 39: I just realised something! This value hasn't been applied to the actual service Deployment! We will need to do that as well.

https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/service/allocation.yaml

markmandel (commented):
Just realised this was missing a step.

Also hope you don't mind but I edited the description to add the "closes" to the issue so it would actually close the issue.

install/helm/agones/values.yaml
Kalaiselvi84: I have added labels in this allocation.yaml file, please share your suggestions

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -160,6 +160,9 @@ spec:
 {{- end }}
 {{- if .Values.agones.allocator.annotations }}
 {{- toYaml .Values.agones.allocator.annotations | nindent 8 }}
+{{- end }}
+{{- if .Values.agones.allocator.labels }}
```
markmandel, line 164: Looks good - just checking if you had a chance to do any manual testing on this? (I don't think there's any better way to actually test this).

Easiest way might be to use `make install` and pass in some extra `HELM_ARGS` and make sure the Deployment labels are good.

Kalaiselvi84: I have validated the labels manually per your guidance

markmandel: Thanks for the validation - but I just realised after looking at it, that these values are showing up in the `annotations:` section.

We should move this code up to above line 152, so it sits under `labels:` - then it should be good to go.

Kalaiselvi84: The values have been moved to the labels section

markmandel (approved):
Looks good!

markmandel (approved):
Looks good!

PR #3137, opened 2023-05-03
Upgrade to Golang version 1.20.4

igooch:
**What type of PR is this?**

/kind cleanup

Upgrade go version to 1.20.4
Upgrade golangci-lint to the version that introduces support of Golang 1.20
Bump example image tags minor version for examples where Go was upgraded.

N/A

**Special notes for your reviewer**

Another PR after this will bump the example images version in the places where they are used. After this PR is merged, we need to build the new version of example images and upload the the registry, otherwise the second PR will not pass CI.


gongmax (approved):
LGTM

PR #3141, opened 2023-05-08
GraceTermination when GameServer get deleted

qizichao-dm:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Current SDKGracefulTermination implementation uses gsStateChannel which get updated in sendGameServerUpdate. However, when GameServer gets deleted (e.g. when fleet updates), sendGameServerUpdate will not be triggered because updateState() will skip update in this case (gs.IsBeingDeleted() returns true). Therefore SDK server will blocks even SDK.Shutdown() has been called.

This PR solves the issue by explicitly updating gsStateChannel when
- GS is being deleted
- Current SDK server state is Shutdown
- State of GS is not Shutdown.

Closes #


pkg/sdkserver/sdkserver.go
```diff
@@ -342,6 +342,16 @@ func (s *SDKServer) updateState(ctx context.Context) error {
 	// If we are currently in shutdown/being deleted, there is no escaping.
 	if gs.IsBeingDeleted() {
 		s.logger.Debug("GameServerState being shutdown. Skipping update.")
+
+		// Explicitly update gsStateChannel if current state is Shutdown since sendGameServerUpdate will not triggered.
+		if runtime.FeatureEnabled(runtime.FeatureSDKGracefulTermination) {
+			if s.gsState == agonesv1.GameServerStateShutdown && gs.Status.State != agonesv1.GameServerStateShutdown {
```
zmerlynn, line 348: LGTM, I think. Idiomatically this should all be the same `if`, i.e.:

```
if runtime.FeatureEnabled(runtime.FeatureSDKGracefulTermination) && s.gsState == agonesv1.GameServerStateShutdown && gs.Status.State != agonesv1.GameServerStateShutdown
```

zmerlynn (commented):
Thanks for the contribution!

aimuz:
I'm glad someone solved this problem

pkg/sdkserver/sdkserver.go
qizichao-dm: Done

zmerlynn (approved):

PR #3148, opened 2023-05-12
Allow setting a static NodePort for the ping service

towolf:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Similar to the allocator service, which allows setting static nodePorts, we added the same thing to the ping service, which will benefit from having a predictable static NodePort. This is for clusters which do not have access to Service Type: LoadBalancer.

### DIFF
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -183,6 +183,7 @@ agones:
       expose: true
       response: ok
       port: 80
+      nodePort: 0 # nodePort will be used if the serviceType is set to NodePort
       serviceType: LoadBalancer
       loadBalancerIP: ""
       loadBalancerSourceRanges: []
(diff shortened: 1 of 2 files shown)


install/helm/agones/values.yaml
```diff
@@ -137,6 +137,7 @@ agones:
       expose: true
       response: ok
       port: 80
+      nodePort: 0 # nodePort will be used if the serviceType is set to NodePort
```
zmerlynn, line 140: Please add documentation in https://github.com/googleforgames/agones/blob/main/site/content/en/docs/Installation/Install%20Agones/helm.md

zmerlynn: Going to get this in, then I'll handle this after this is merged.

zmerlynn (approved):

PR #3149, opened 2023-05-12
[Release Automation] Update Helm/SDK/Install Packages Version Numbers

### LINKED ISSUE #2456: Release Automation: Create a script to increment version numbers
Right now there are a number of steps in the [release process](https://github.com/googleforgames/agones/blob/main/docs/governance/templates/release_issue.md) to increment version numbers in the install files, sdks, etc:

- [ ] Ensure the [helm `tag` value][values] is correct (should be {version} if a full release, {version}-rc if release candidate)
- [ ] Ensure the [helm `Chart` version values][chart] are correct (should be {version} if a full release, {version}-rc if release candidate)
- [ ] Update SDK Package Versions
    - [ ] Update the package version in [`sdks/nodejs/package.json`][package.json] and [`sdks/nodejs/package-lock.json`][package-lock.json] by running `npm version {version}` if a full release or `npm version {version}-rc` if release candidate
    - [ ] Ensure the [`sdks/csharp/sdk/AgonesSDK.nuspec` and `sdks/csharp/sdk/csharp-sdk.csproj`][csharp] versions are correct (should be {version} if a full release, {version}-rc if release candidate)
    - [ ] Update the package version in the [`sdks/unity/package.json`][unity] package file's `Version` field to {version} if a full release, {version}-rc if release candidate
- [ ] Run `make gen-install`
- [ ] If full release, then increment the `base_version` in [`build/Makefile`][build-makefile]
- [ ] If full release, change to the `sdks/nodejs` directory and run the command `npm version {version}+1-dev` to update the package version
- [ ] If full release move the [`sdks/csharp/sdk/AgonesSDK.nuspec` and `sdks/csharp/sdk/csharp-sdk.csproj`][csharp] to {version}+1-dev
- [ ] If full release update the [`sdks/unity/package.json`][unity] package file's `Version` field to {version}+1-dev

Create a script to bump the version numbers. Consider creating a target in the Makefile to execute the script.

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind feature

Scripts to increment version numbers in helm/sdk/install packages before and after release.

Closes #2456


build/scripts/version-update/after-release/after-release.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+func main() {
```
zmerlynn, line 13: rather than the heavy duplication between `after-release` and `before-release`, I would recommend using a flag. For a script like this, no worries using the standard flag library: https://pkg.go.dev/flag

Examples in our repo include: https://github.com/googleforgames/agones/blob/main/examples/simple-game-server/main.go#L41-L49

build/scripts/version-update/after-release/after-release.go
```diff
@@ -0,0 +1,81 @@
...
+	if len(os.Args) < 2 {
+		fmt.Println("Please provide the initial version as a command-line argument")
```
zmerlynn, line 15: log.Fatalf here

build/scripts/version-update/after-release/after-release.go
```diff
@@ -0,0 +1,81 @@
...
+		"sdks/nodejs/package-lock.json",
+		"sdks/unity/package.json",
+		"sdks/csharp/sdk/AgonesSDK.nuspec",
+		"sdks/csharp/sdk/csharp-sdk.csproj",
+	}
+
+	for _, filename := range files {
+		err := UpdateVersionInFile(filename, initialVersion)
+		if err != nil {
+			fmt.Printf("Error updating file %s: %s\n", filename, err.Error())
```
zmerlynn, line 34: maybe a `log.Fatalf` instead of just a Printf?

build/scripts/version-update/after-release/after-release.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return err
+	}
+
+	content := string(fileBytes)
+
+	// Check the file extension and update version accordingly
+	ext := filepath.Ext(filename)
+	switch ext {
+	case ".yaml", ".yml", ".nuspec", ".csproj":
```
zmerlynn, line 50: I'm not sure this needs a case statement. It seems like you can just write it as

```
if ext := filepath.Ext(filename); ext == ".json" {
  updateJSONVersion()
} else {
  updateVersion()
}
```

build/scripts/version-update/after-release/after-release.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+func updateVersion(content string, initialVersion string) string {
+	re := regexp.MustCompile(regexp.QuoteMeta(initialVersion))
+	newVersion := incrementVersion(initialVersion)
+	return re.ReplaceAllString(content, newVersion+"-dev")
+}
+
+func updateJSONVersion(content string, initialVersion string) string {
```
zmerlynn, line 70: It seems like the only difference here is handling `"1.2.3"` vs `1.2.3`, is that right?

zmerlynn (changes requested):

build/scripts/version-update/after-release/after-release.go
Kalaiselvi84: I tried to update the JSON files without a switch/case statement, but the syntax was slightly different. So created a switch/case statement to ensure that the files were updated correctly.

build/scripts/version-update/after-release/after-release.go
Kalaiselvi84: yes that's correct

build/scripts/version-update/after-release/after-release.go
Kalaiselvi84: I have updated the code to make it a single script using Flag. Thanks!

build/scripts/version-update/after-release/after-release.go
Kalaiselvi84: used log instead fmt package

build/scripts/version-update/after-release/after-release.go
Kalaiselvi84: used log instead fmt

build/scripts/version-update/main.go
```diff
@@ -0,0 +1,138 @@
...
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var releaseType string
+
+func init() {
+	flag.StringVar(&releaseType, "type", "", "Specify the release type ('before' or 'after')")
```
zmerlynn, line 17: Maybe `release-stage` instead of `type` (and change variables)? `releaseType` seems like it should be what type of release it is, vs what stage the release is in.

build/scripts/version-update/main.go
```diff
@@ -0,0 +1,138 @@
...
+
+func main() {
+	flag.Parse()
+
+	if len(os.Args) < 3 {
+		log.Fatalf("Please provide the release type ('before' or 'after') and the initial version as command-line arguments")
+		return
+	}
+
+	releaseType := os.Args[1]
```
zmerlynn, line 28: The way you are parsing this right now seems to ignore the flags - instead it looks like you would execute it via `version-update before 1.2.3`, vs `version-update --type=before 1.2.3`.

Since you are using the `flag` package, you should trust that the `flag` package can parse. For non-flag arguments, you should use [flag.Args](https://pkg.go.dev/flag#Args) .. so line 23 should be `if len(flag.Args()) < 1`, for instance.

`releaseType` is already initialized in `init()` above, so you don't need this line at all, and the line below should just be `initialVersion := flag.Args()[0]`.

build/scripts/version-update/main.go
```diff
@@ -0,0 +1,138 @@
...
+			log.Fatalf("Invalid release type. Please specify 'before' or 'after'.")
+		}
+
+		if err != nil {
+			log.Fatalf("Error updating file %s: %s\n", filename, err.Error())
+		}
+	}
+}
+
+func UpdateValueInFileBeforeRelease(filename string, valueToUpdate string) error {
```
zmerlynn, line 61: There is a lot of duplication in the `BeforeRelease` and `AfterRelease` variants - you should be able to factor these together. Consider that all this really is is "replace regexp with value" and what you need is a function that takes a regexp and replaces - the regexp and replacement value are the parameters. Then you can set those parameters based on whether it's before/after and whether it's JSON, which seems to be the only differences.

build/scripts/version-update/main.go
Kalaiselvi84: Renamed type to release-stage and releaseType to releaseStage.

build/scripts/version-update/main.go
Kalaiselvi84: I have updated the code and passed the --release-stage=before flag from the command line.

build/scripts/version-update/main.go
Kalaiselvi84: I tried my best to avoid the duplication😀

Please let me know if there are any areas of the code that you think could be improved. Thanks!

build/scripts/version-update/main.go
```diff
@@ -0,0 +1,114 @@
...
+func UpdateFile(filename string, initialVersion string) error {
+	fileBytes, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+
+	content := string(fileBytes)
+
+	ext := filepath.Ext(filename)
+	validExtensions := []string{".yaml", ".yml", ".nuspec", ".csproj"}
```
zmerlynn, line 58: I would eliminate this and just change the below to `if ext == ".json"` - you're adding some amount of complexity just to validate a list that is more or less a constant, 20 lines previous to this.

Eliminating this will allow you to eliminate the logic in lines 62-72 and 86-90 as well - it seems like we're going through a lot of effort there.

zmerlynn (commented):
Getting close!

build/scripts/version-update/main.go
Kalaiselvi84: Updated the code and it looks simple now. Thanks Zach!

zmerlynn (approved):

build/scripts/version-update/main.go
```diff
@@ -0,0 +1,95 @@
+package main
+
+import (
+	"flag"
+	"io/ioutil"
```
gongmax, line 5: From linter: ` SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)`

Kalaiselvi84: replaced io/ioutil with os👍🏻

Issue #3098, opened 2023-04-13
Graduate `SplitControllerAndExtensions` to Beta


roberthbailey:
I propose that we graduate the `SplitControllerAndExtensions` feature to beta. It has been in alpha since 1.30.0. 

Context:
https://development.agones.dev/site/docs/guides/feature-stages/#feature-gates

PR #3152, opened 2023-05-15
Add Cloud Best Practices guide, add guide on Release Channels

zmerlynn:
* Adds a "Best Practices" section to the sidebar
  * Adds some generic Best Practices
  * And a GKE section that discussed Release Channels
* Move the GKE-specific "long-session" discussion out of "Controlling Disruption" and into GKE.

* Refactor the GKE getting started page considerably:
  * Add a section on choosing a release channel that calls out to the guide but otherwise suggests using `regular`.
  * Move discussion of zonal/regional to top - current instructions lean towards a zonal cluster but there is really no disadvantage to defaulting to regional.
  * Along the way, remove `gcloud config set` references. We strongly encourage multiple clusters, it's not clear to me why the instructions lean so heavily on changing your global gcloud config.
  * Assume that the `SplitControllerAndExtensions` gate is getting graduated, remove reference to Alpha support for Autopilot. (Does not remove feature flag, will leave for graduation.)

### DIFF
--- a/site/content/en/docs/Guides/Best Practices/_index.md
+++ b/site/content/en/docs/Guides/Best Practices/_index.md
@@ -1,6 +1,6 @@
 ---
-title: "Cloud Best Practices" 
-linkTitle: "Cloud Best Practices"
+title: "Best Practices" 
+linkTitle: "Best Practices"
 date: 2023-05-12T00:00:00Z
 weight: 9
 description: "Best practices for running Agones in production."
@@ -16,14 +16,16 @@ some general best practices. We also have cloud specific pages for:
 
 If you are interested in submitting best practices for your cloud prodiver / on-prem, [please contribute!]({{< relref "/Contribute" >}})
 
-## Separate game servers from other workloads
+## Separation of Agones from GameServer nodes
 
-We recommend using [taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) to ensure that your game servers are running separate from Agones controllers. There are many ways to accomplish this, here we discuss one:
+When running in production, Agones should be scheduled on a dedicated pool of nodes, distinct from where Game Servers
+are scheduled for better isolation and resiliency. By default Agones prefers to be scheduled on nodes labeled with
+`agones.dev/agones-system=true` and tolerates the node taint `agones.dev/agones-system=true:NoExecute`.
+If no dedicated nodes are available, Agones will run on regular nodes. See [taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)
+for more information about Kubernetes taints and tolerations.
 
-* Label a group of nodes `agones.dev/agones-system=true`
-* Taint the same group of nodes with `agones.dev/agones-system=true:NoExecute`
-
-If you are collecting [Metrics]({{< relref "metrics" >}}) using our standard Prometheus installation, consider isolating the `agones-metrics` namespace as well.
+If you are collecting [Metrics]({{< relref "metrics" >}}) using our standard Prometheus installation, see
+[the installation guide]({{< relref "metrics#prometheus-installation" >}}) for instructions on configuring a separate node pool for the `agones.dev/agones-metrics=true` taint.
 
 See [Creating a Cluster]({{< relref "Creating Cluster" >}}) for initial set up on your cloud provider.
 
--- a/site/content/en/docs/Guides/Best Practices/_index.md
+++ b/site/content/en/docs/Guides/Best Practices/_index.md
@@ -1,0 +1,43 @@
+---
+title: "Cloud Best Practices" 
+linkTitle: "Cloud Best Practices"
+date: 2023-05-12T00:00:00Z
+weight: 9
+description: "Best practices for running Agones in production."
+---
+
+## Overview
+
+Running Agones in production takes consideration, from planning your launch to figuring
+out the best course of action for cluster and Agones upgrades. On this page, we've collected
+some general best practices. We also have cloud specific pages for:
+
+* [Google Kubernetes Engine (GKE)]({{< relref "gke.md" >}})
+
+If you are interested in submitting best practices for your cloud prodiver / on-prem, [please contribute!]({{< relref "/Contribute" >}})
+
+## Separate game servers from other workloads
+
+We recommend using [taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) to ensure that your game servers are running separate from Agones controllers. There are many ways to accomplish this, here we discuss one:
+
+* Label a group of nodes `agones.dev/agones-system=true`
+* Taint the same group of nodes with `agones.dev/agones-system=true:NoExecute`
+
+If you are collecting [Metrics]({{< relref "metrics" >}}) using our standard Prometheus installation, consider isolating the `agones-metrics` namespace as well.
+
+See [Creating a Cluster]({{< relref "Creating Cluster" >}}) for initial set up on your cloud provider.
+
+## Redundant Clusters
+
+### Allocate Across Clusters
+
+Agones supports [Multi-cluster Allocation]({{< relref "multi-cluster-allocation" >}}), allowing you to allocate from a set of clusters, versus a single point of potential failure. There are several other options for multi-cluster allocation:
+* [Anthos Service Mesh](https://cloud.google.com/anthos/service-mesh) can be used to route allocation traffic to different clusters based on arbitrary criteria. See [Global Multiplayer Demo](https://github.com/googleforgames/global-multiplayer-demo) for an example where the match maker influences which cluster the allocation is routed to.
+* [Allocation Endpoint](https://github.com/googleforgames/agones/tree/main/examples/allocation-endpoint) can be used in Cloud Run to proxy allocation requests.
+* Or peruse the [Third Party Examples]({{< relref "../../Third Party Content/libraries-tools.md/#allocation" >}})
+
+### Spread
+
+You should consider spreading your game servers in two ways:
+* **Across geographic fault domains** ([GCP regions](https://cloud.google.com/compute/docs/regions-zones), [AWS availability zones](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html), separate datacenters, etc.): This is desirable for geographic fault isolation, but also for optimizing client latency to the game server.
+* **Within a fault domain**: Kubernetes Clusters are single points of failure. A single misconfigured RBAC rule, an overloaded Kubernetes Control Plane, etc. can prevent new game server allocations, or worse, disrupt existing sessions. Running multiple clusters within a fault domain also allows for [easier upgrades]({{< relref "Upgrading#upgrading-agones-multiple-clusters" >}}).
--- a/site/content/en/docs/Guides/Best Practices/gke.md
+++ b/site/content/en/docs/Guides/Best Practices/gke.md
@@ -1,13 +1,13 @@
 ---
-title: "GKE Best Practices"
-linkTitle: "GKE"
+title: "Google Kubernetes Engine Best Practices"
+linkTitle: "Google Cloud"
 date: 2023-05-12T00:00:00Z
-description: "Best practices for running Agones on GKE."
+description: "Best practices for running Agones on Google Kubernetes Engine (GKE)."
 ---
 
 ## Overview
 
-On this page, we've collected several GKE best practices.
+On this page, we've collected several [Google Kubernetes Engine (GKE)](https://cloud.google.com/kubernetes-engine/) best practices.
 
 ## Release Channels
 
@@ -30,7 +30,7 @@ If you need to disallow minor version upgrades for more than 6 months, consider 
 
 ### What versions are available on a given channel?
 
-You can query the versions available across different channels using `gcloud` using:
+You can query the versions available across different channels using `gcloud`:
 
 ```
 gcloud container get-server-config \
@@ -42,7 +42,7 @@ Replace the following:
 
 * **COMPUTE_REGION**: the
 [Google Cloud region](https://cloud.google.com/compute/docs/regions-zones#available)
-for the cluster.
+where you will create the cluster.
 
 ## Managing Game Server Disruption on GKE
 
--- a/site/content/en/docs/Guides/Best Practices/gke.md
+++ b/site/content/en/docs/Guides/Best Practices/gke.md
@@ -1,0 +1,60 @@
+---
+title: "GKE Best Practices"
+linkTitle: "GKE"
+date: 2023-05-12T00:00:00Z
+description: "Best practices for running Agones on GKE."
+---
+
+## Overview
+
+On this page, we've collected several GKE best practices.
+
+## Release Channels
+
+### Why?
+
+We recommned using [Release Channels](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels) for all GKE clusters. Using Release Channels has several advantages:
+* Google automatically manages the version and upgrade cadence for your Kubernetes Control Plane and its nodes.
+* Clusters on a Release Channel are allowed to use the `No minor upgrades` and `No minor or node upgrades` [scope of maintenance exclusions](https://cloud.google.com/kubernetes-engine/docs/concepts/maintenance-windows-and-exclusions#limitations-maint-exclusions) - in other words, enrolling a cluster in a Release Channel gives you _more control_ over node upgrades.
+* Clusters enrolled in `rapid` channel have access to the newest Kubernetes version first. Agones strives to [support the newest release in `rapid` channel]({{< relref "Installation#agones-and-kubernetes-supported-versions" >}}) to allow you to test the newest Kubernetes soon after it's available in GKE.
+
+{{< alert title="Note" color="info" >}}
+GKE Autopilot clusters must be on Release Channels.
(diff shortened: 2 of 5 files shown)


site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -133,33 +152,27 @@ gcloud compute firewall-rules create game-server-firewall \
...
-  --no-enable-autoupgrade \
   --enable-image-streaming \
   --machine-type=e2-standard-4
 ```
 
-{{< alert title="Note" color="info">}}
-If you're creating a cluster to run Windows game servers, you need to add the `--enable-ip-alias` flag to create the cluster with [Alias IP ranges](https://cloud.google.com/vpc/docs/alias-ip) instead of routes-based networking.
-{{< /alert >}}
-
 Flag explanations:
```
gongmax, line 171: Should we add explanations for `region` and `release-channel`? They are obvious but just want to keep consistency.

gongmax (approved):

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -133,33 +152,27 @@ gcloud compute firewall-rules create game-server-firewall \
...
-(Alpha).
+or [GKE Autopilot mode](#create-an-autopilot-mode-cluster-for-agones).
 
 ### Create a Standard mode cluster for Agones
 
+Create the cluster:
 ```bash
-gcloud container clusters create [CLUSTER_NAME] --cluster-version={{% gke-example-cluster-version %}} \
+gcloud container clusters create [CLUSTER_NAME] \
+  --region=[COMPUTE_REGION]
```
roberthbailey, line 162: You need to add a `\`at the end of these two new lines so that the command can be copy/pasted.

site/content/en/docs/Installation/Creating Cluster/gke.md
roberthbailey, line 171: Please add flag explanations for the two new flags.

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -173,12 +186,12 @@ recommended for a production deployment.
 ```bash
 gcloud container node-pools create agones-system \
   --cluster=[CLUSTER_NAME] \
-  --no-enable-autoupgrade \
+  --region=[COMPUTE_REGION] \
   --node-taints agones.dev/agones-system=true:NoExecute \
   --node-labels agones.dev/agones-system=true \
   --num-nodes=1
```
roberthbailey, line 192: I think we should add a machine-type flag here for the system node pool (`e2-standard-4`). We specify one for the metrics node pool but not here which I find strange.

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -188,17 +201,16 @@ Create a node pool for [Metrics]({{< relref "../../Guides/metrics.md" >}}) if yo
...
 ```
+where `[CLUSTER_NAME]`is the name of the cluster you created and `[COMPUTE_REGION]` is the region.
 
 Flag explanations:
 
-* cluster: The name of your existing cluster in which the node pool is created.
-* no-enable-autoupgrade: Disable automatic upgrades for nodes to reduce the likelihood of in-use games being disrupted.
 * node-taints: The Kubernetes taints to automatically apply to nodes in this node pool.
 * node-labels: The Kubernetes labels to automatically apply to nodes in this node pool.
 * num-nodes: The Agones system controllers only require a single node of capacity to run. For faster recovery time in the event of a node failure, you can increase the size to 2.
```
roberthbailey, line 216: This looks like it was copy/pasted from a section that no longer exists. If you are using a regional cluster, you get 3 nodes (so the part about 1 vs 2 is incorrect) and this is an explanation for the metrics pool not the agones system pool.

roberthbailey, line 216: There isn't a flag explanation for the machine type flag, which is strange as its the only one without one. Can you copy the flag explanation from the cluster creation section to here?

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -173,12 +186,12 @@ recommended for a production deployment.
...
 gcloud container node-pools create agones-system \
   --cluster=[CLUSTER_NAME] \
-  --no-enable-autoupgrade \
+  --region=[COMPUTE_REGION] \
   --node-taints agones.dev/agones-system=true:NoExecute \
   --node-labels agones.dev/agones-system=true \
   --num-nodes=1
 ```
-where [CLUSTER_NAME] is the name of the cluster you created.
+where `[CLUSTER_NAME]`is the name of the cluster you created and `[COMPUTE_REGION]` is the region.
```
roberthbailey, line 194: This section is missing flag explanations. It didn't seem strange until I saw that both the previous and next sections have them, so they should be added here too.

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -217,7 +229,7 @@ through [Github issues](https://github.com/googleforgames/agones/issues).
 ```bash
 gcloud container node-pools create windows \
   --cluster=[CLUSTER_NAME] \
-  --no-enable-autoupgrade \
+  --region=[COMPUTE_REGION] \
   --image-type WINDOWS_LTSC_CONTAINERD \
   --machine-type e2-standard-4 \
   --num-nodes=4
```
roberthbailey, line 235: This one should also say "where `[CLUSTER_NAME]`is the name of the cluster you created and `[COMPUTE_REGION]` is the region."

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -260,35 +252,23 @@ Agones supports Autopilot clusters in [alpha]({{<ref "/docs/Guides/feature-stage
...
-Service Level Objectives or Agreements.
-
-This is particularly true for a zonal GKE control plane, which can go down temporarily to adjust for cluster resizing,
-[automatic upgrades](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades) and
-[repairs](https://cloud.google.com/kubernetes-engine/docs/concepts/maintenance-windows-and-exclusions#repairs).
-{{< /alert >}}
-
 ## Next Steps
 
 ### GKE Standard
```
roberthbailey, line 274: In the autopilot section below it says "These installation instructions apply to Agones 1.30." I think they also apply to 1.31 and unless we close https://github.com/googleforgames/agones/issues/3098 in the next day or so they will apply to 1.32 as well. Can you adjust that to specify the range and also mention that support for GKE autopilot was added in 1.30?

zmerlynn:
(no new changes to review, merely forcing an e2e build)

site/content/en/docs/Guides/Best Practices/_index.md
```diff
@@ -0,0 +1,36 @@
...
+out the best course of action for cluster and Agones upgrades. On this page, we've collected
+some general best practices. We also have cloud specific pages for:
+
+* [Google Kubernetes Engine (GKE)]({{< relref "gke.md" >}})
+
+If you are interested in submitting best practices for your cloud prodiver / on-prem, [please contribute!]({{< relref "/Contribute" >}})
+
+## Separate game servers from other workloads
+
+We recommend using [taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) to ensure that your game servers are running separate from Agones controllers. There are many ways to accomplish this, here we discuss one: Assuming the default installation into the `agones-system` namespace:
```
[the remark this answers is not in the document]
zmerlynn: I didn't even see this section!

site/content/en/docs/Installation/Creating Cluster/gke.md
zmerlynn: Actually, I think they should be moved here and removed below. Looking at how they're explained, below we only explain the _new_ flags.

site/content/en/docs/Installation/Creating Cluster/gke.md
zmerlynn: This is getting really redundant, should we just have a tutorial on creating a node pool and stamp them out in different ways?

site/content/en/docs/Installation/Creating Cluster/gke.md
zmerlynn: Do we need to specify when the support was added?

Changed it to 1.30+.

site/content/en/docs/Installation/_index.md
```diff
@@ -57,13 +57,7 @@ The following table lists recent Agones versions and their corresponding require
 | 1.22           | 1.21                  |
 | 1.21           | 1.21                  |
 
-## Best Practices
+## Best Practices {#separation-of-agones-from-gameserver-nodes}
+{{/* keep installation/#separation-of-agones-from-gameserver-nodes permalink */}}
```
roberthbailey, line 61: This line is being rendered: https://ae79bca-dot-preview-dot-agones-images.appspot.com/site/docs/installation/#separation-of-agones-from-gameserver-nodes

site/content/en/docs/Guides/Best Practices/_index.md
```diff
@@ -0,0 +1,43 @@
...
+out the best course of action for cluster and Agones upgrades. On this page, we've collected
+some general best practices. We also have cloud specific pages for:
+
+* [Google Kubernetes Engine (GKE)]({{< relref "gke.md" >}})
+
+If you are interested in submitting best practices for your cloud prodiver / on-prem, [please contribute!]({{< relref "/Contribute" >}})
+
+## Separate game servers from other workloads
+
+We recommend using [taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) to ensure that your game servers are running separate from Agones controllers. There are many ways to accomplish this, here we discuss one:
```
roberthbailey, line 21: This section loses the information from the prior best practice that agones by default is configured to put the system components on these nodes. Also with @gongmax's PR, we could actually suggest setting the helm property to force the system components onto these nodes.

site/content/en/docs/Guides/Best Practices/_index.md
```diff
@@ -0,0 +1,43 @@
...
+
+Running Agones in production takes consideration, from planning your launch to figuring
+out the best course of action for cluster and Agones upgrades. On this page, we've collected
+some general best practices. We also have cloud specific pages for:
+
+* [Google Kubernetes Engine (GKE)]({{< relref "gke.md" >}})
+
+If you are interested in submitting best practices for your cloud prodiver / on-prem, [please contribute!]({{< relref "/Contribute" >}})
+
+## Separate game servers from other workloads
```
roberthbailey, line 19: The title is backwards from what we had before. What we are saying is to move the agones components away from game servers. 

You could also move the game servers away by using taints and tolerations for them, but the suggestion here is showing how to move the agones components so that the game servers don't need taints and tolerations. 

Consider flipping the title to be more like what we had on the other page ("Separation of Agones from GameServer nodes").

site/content/en/docs/Installation/_index.md
```diff
@@ -57,13 +57,7 @@ The following table lists recent Agones versions and their corresponding require
...
 
-### Separation of Agones from GameServer nodes
-
-When running in production, Agones should be scheduled on a dedicated pool of nodes, distinct from where Game Servers
-are scheduled for better isolation and resiliency. By default Agones prefers to be scheduled on nodes labeled with
-`agones.dev/agones-system=true` and tolerates the node taint `agones.dev/agones-system=true:NoExecute`.
-If no dedicated nodes are available, Agones will run on regular nodes.
-
-[windows]: {{% ref "/docs/Guides/windows-gameservers.md" %}}
```
roberthbailey, line 69: Removing this line breaks the link above in the supported container architectures table.

site/content/en/docs/Guides/Best Practices/_index.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Cloud Best Practices"
```
roberthbailey, line 2: Are these just "best practices"? Why the cloud prefix?

site/content/en/docs/Guides/Best Practices/gke.md
```diff
@@ -0,0 +1,60 @@
+---
+title: "GKE Best Practices"
+linkTitle: "GKE"
```
roberthbailey, line 3: Should we make the linkTitle "Google Cloud" to be consistent with how we do the cloud specific naming in the cluster creation section?

site/content/en/docs/Installation/Creating Cluster/gke.md
roberthbailey: I agree it's redundant. We could make a template but I think that might be more work to maintain than the 3 copies of the docs we currently have.

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -257,38 +281,28 @@ Agones supports Autopilot clusters in [alpha]({{<ref "/docs/Guides/feature-stage
       --release-channel=[RELEASE_CHANNEL] \
       --autoprovisioning-network-tags=game-server \
```
roberthbailey, line 282: I think we need to remove this last trailing slash.

site/content/en/docs/Guides/Best Practices/gke.md
```diff
@@ -0,0 +1,60 @@
...
+
+### What channel should I use?
+
+We recommend the `regular` channel, which offers a balance between stability and freshness. See [this guide](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels#what_channel_should_i_use) for more discussion.
+
+If you need to disallow minor version upgrades for more than 6 months, consider choosing the freshest Kubernetes version possible: Choosing the freshest version on `rapid` or `regular` will extend the amount of time before your cluster reaches [end of life](https://cloud.google.com/kubernetes-engine/docs/release-schedule#schedule-for-release-channels).
+
+### What versions are available on a given channel?
+
+You can query the versions available across different channels using `gcloud` using:
```
roberthbailey, line 33: "using gcloud using" has one too many usings in it.

site/content/en/docs/Guides/Best Practices/gke.md
```diff
@@ -0,0 +1,60 @@
...
+gcloud container get-server-config \
+  --region=[COMPUTE_REGION] \
+  --flatten="channels" \
+  --format="yaml(channels)"
+```
+Replace the following:
+
+* **COMPUTE_REGION**: the
+[Google Cloud region](https://cloud.google.com/compute/docs/regions-zones#available)
+for the cluster.
```
roberthbailey, line 45: "for the cluster" isn't quite accurate, since presumably you are doing this before creating a cluster. What about "where you will create the cluster" or something like that?

roberthbailey (commented):
Another round of comments.

site/content/en/docs/Installation/_index.md
zmerlynn: Gah, I must not have refreshed the site-server render after adding the comment. Fixed.

site/content/en/docs/Guides/Best Practices/_index.md
zmerlynn: FWIW I actually disagree with the direction we had it. I actually had read this recommendation several times due to @gongmax but I never actually parsed it fully. Why are we separating controllers from the default pool? Why are we not separating game servers and keeping the control plane, kube-system, etc, in the default pool? This sounds backwards.

I didn't intend to flip the recommendation, but the title is a reflection of the direction I think it _should_ be. Freudian slip. :)

site/content/en/docs/Installation/_index.md
zmerlynn: Ah, weird place for that. Moved it up to the section that uses it first, which seems to be convention elsewhere, and changed it to a relref, which we should probably prefer.

site/content/en/docs/Guides/Best Practices/gke.md
zmerlynn: I never noticed that dissonance but I changed it over and expanded GKE the first use.

site/content/en/docs/Guides/Best Practices/_index.md
zmerlynn: Let's make that change when #3161 goes in?

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -133,35 +152,35 @@ gcloud compute firewall-rules create game-server-firewall \
...
+  --release-channel=[RELEASE_CHANNEL] \
   --tags=game-server \
   --scopes=gke-default \
   --num-nodes=4 \
-  --no-enable-autoupgrade \
   --enable-image-streaming \
   --machine-type=e2-standard-4
 ```
 
-{{< alert title="Note" color="info">}}
```
roberthbailey, line 151: Just to be sure, is this note being removed because it's now the default?

roberthbailey (commented):
LGTM, just waiting for a render to verify the formatting on the actual site.

site/content/en/docs/Installation/Creating Cluster/gke.md
zmerlynn: Yes, exactly. It's been default since 1.22-ish (I looked it up before I removed it but can't quite remember now.)

roberthbailey (approved):

PR #3160, opened 2023-05-17
Move back to FailureThreshold failures of /gshealthz

zmerlynn:
This again reverts part of #3072. In looking at recent failures, in cases where the sidecar is slow to come up due to networking issues, `/gshealthz` fails, resulting often in the unnecessary restart of the game server.

This returns to the more generous failure threshold from prior to to just make this ~infinite, i.e. neuter kubelet liveness, since health is really owned by the controller and sidecar.

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -706,15 +706,19 @@ func (c *Controller) addGameServerHealthCheck(gs *agonesv1.GameServer, pod *core
 					},
 				},
 				// The sidecar relies on kubelet to delay by InitialDelaySeconds after the
-				// container is started (after image pull, etc), and relies on the kubelet
-				// for PeriodSeconds heartbeats to evaluate health.
+				// container is started (after image pull, etc).
 				InitialDelaySeconds: gs.Spec.Health.InitialDelaySeconds,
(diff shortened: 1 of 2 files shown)


zmerlynn:
Analysis of a recent failure.
image

pkg/gameservers/controller.go
```diff
@@ -706,15 +706,19 @@ func (c *Controller) addGameServerHealthCheck(gs *agonesv1.GameServer, pod *core
...
-				// failure is sufficient to know the game server is unhealthy.
-				FailureThreshold: 1,
+				// {FailureThreshold in a row} failed health checks, so in theory on the kubelet
+				// side, one failure is sufficient to know the game server is unhealthy. However,
+				// with only one failure, if the sidecar doesn't come up at all, we unnecessarily
+				// restart the game server. So use FailureThreshold as startup wiggle-room as well.
+				//
+				// Note that in general, FailureThreshold could also be infinite - the controller
+				// and sidecar are responsible for health management.
+				FailureThreshold: gs.Spec.Health.FailureThreshold,
```
gongmax, line 721: The default value is 3, right?

PR #3153, opened 2023-05-15
Update values yaml file for `SplitControllerAndExtensions`

chiayi:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**:
This is a step for graduating `SplitControllerAndExtensions` to Beta

**Which issue(s) this PR fixes**:
Part of graduating https://github.com/googleforgames/agones/issues/2797 

**Special notes for your reviewer**:

PR #3161, opened 2023-05-17
Add a helm flag to force Agones system components onto dedicated nodes

### LINKED ISSUE #3009: Add a helm flag to force Agones system components onto dedicated nodes
The current Agones installation attempts to put the system components onto dedicated nodes if nodes with the correct taints and labels exist. This is great for a getting started experience because it doesn't require any advanced cluster configuration but it also means that for production installations it is possible, even if dedicated nodes exist, for the agones system components to inadvertently begin running on incorrect nodes, such as those being used for the game servers. 

A helm configuration option that changes the system components from _preferring_ dedicated nodes to _requiring_ dedicated nodes. 

This is similar to what was done when creating the [allocation scenario load tests](https://github.com/googleforgames/agones/tree/main/test/load/allocation#scenario-tests):

> First, we made sure that the Agones pods would only run in the Agones node pool by changing the node affinity in the deployments for the controller, allocator service, and ping service to requiredDuringSchedulingIgnoredDuringExecution.

and the changes are captured in https://github.com/googleforgames/agones/blob/main/test/load/allocation/configure-agones.sh. 

We could change the **default** installation to require properly configured dedicated nodes. This might actually be a reasonable change to consider in the future as it  will make it harder for folks to ignore our [best practices for cluster installation](https://agones.dev/site/docs/installation/#best-practices)). But for now I'd like to propose adding it as an option and documenting the option in the best practices when installing Agones.

gongmax:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

 /kind feature

> /kind hotfix

Closes #3009

Tested with 
`helm template agones-manual --namespace agones-system  . > install-custom-1.yaml` and 
`helm template agones-manual --namespace agones-system  . --set agones.system.requireDedicatedNode=true > install-custom-2.yaml`
And then use the generated yaml file to install Agones. Verified when the flag set to true, Agones will not be installed in a cluster that doesn't have dedicated nodes for Agones system.


install/helm/agones/templates/controller.yaml
```diff
@@ -64,11 +64,21 @@ spec:
         release: {{ .Release.Name }}
         heritage: {{ .Release.Service }}
     spec:
-      {{- if .Values.agones.controller.affinity }}
```
roberthbailey, line 67: Removing this conditional makes it only possible to set the two types of affinity that we pre-define here. I don't know if anyone is using/overriding this parameter, but if so this would be a breaking change for them where their values.yaml would not work with the next release. 

Instead, I would suggest leaving this in (so that its still possible to set the affinity manually) but removing the config from `values.yaml` (as you have done). If affinity is explicitly set, then we use that. If not, then we do what you are doing here and move the default affinity into the template. 

We could also add a check that you don't both set a manual affinity block _and_ the new `requireDedicatedNode` to true at the same time, since that would not be a valid configuration. 

WDYT?

gongmax: Agreed. Fixed in the new revision.

roberthbailey: This is great, thanks!

install/helm/agones/values.yaml
```diff
@@ -51,21 +51,14 @@ agones:
       # requests:
       #   cpu: 1
       #   memory: 256Mi
+    requireDedicatedNode: false
```
roberthbailey, line 54: I'm trying to think of a use case where someone would set one of these to true but not all of them. While having fine-grained settings can be nice, it also means you have to remember to set all 3 variables instead of just a single one when you want the required affinity. 

WDYT about having a single variable vs. 3 variables?

@zmerlynn - do you have any preference on this?

gongmax: Now I'm leaning more to have a single flag as start to avoid overengineering. We can always provide more config later.

roberthbailey:
You'll need to rebase on #3153 (and fix the new section's affinity rules too).

roberthbailey (approved):

install/helm/agones/values.yaml
```diff
@@ -46,6 +46,8 @@ agones:
   createPriorityClass: true
   priorityClassName: agones-system
   cloudProduct: "auto"
+  system:
+    requireDedicatedNode: false
```
zmerlynn, line 50: have you tried `true` on Autopilot?

zmerlynn (approved):
This LGTM and I like the single flag better.

install/helm/agones/values.yaml
gongmax: Nice catch. It won't work for AP. Add the clarification in helm.md

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -269,7 +269,7 @@ The following tables lists the configurable parameters of the Agones chart and t
...
 | `agones.allocator.readiness.failureThreshold`         | Number of times before giving up (in seconds)                                                                                                                                                                           | `3`                                |
-| `agones.system.requireDedicatedNode`                  | Forces Agones system components to be scheduled on dedicated nodes                                                                                                                                                                           |  `false`                           |
+| `agones.system.requireDedicatedNode`                  | Forces Agones system components to be scheduled on dedicated nodes, only applies to the GKE Standard mode cluster                                                                                                                                                                          |  `false`                           |
```
zmerlynn, line 272: Why is that true? It should be doable: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-separation

gongmax: On that page, it's not clear to me whether we can use same mechanism as today (node affinity) on AP. The error I got when apply the install.yaml is 
~~~
node affinity is not allowed on labels with keys: 'agones.dev/agones-system'; Autopilot only allows labels with keys: 'cloud.google.com/compute-class,cloud.google.com/gke-spot,topology.kubernetes.io/region,topology.kubernetes.io/zone,failure-domain.beta.kubernetes.io/region,failure-domain.beta.kubernetes.io/zone,cloud.google.com/gke-os-distribution,kubernetes.io/os,kubernetes.io/arch,cloud.google.com/private-node,sandbox.gke.io/runtime,cloud.google.com/gke-accelerator,cloud.google.com/gke-accelerator-count,iam.gke.io/gke-metadata-server-enabled'."
~~~

So my guess is, on AP, we can only use `spec.nodeSelector`.

gongmax: We may need to add a statement somewhere to call out using `agones.*.nodeSelector` for separating agones system on AP.

roberthbailey: That's unfortunate. What we have right now is the recommended way to do it for GKE clusters w/o node autoprovisining: https://cloud.google.com/kubernetes-engine/docs/how-to/isolate-workloads-dedicated-nodes

It looks like on autopilot you use the node selector _instead_ of affinity. And it seems like even with a node selector, if we specify affinity the spec will be rejected. That's concerning, because with this change as-is, we now always have affinity specified (before you could override it to turn it off).

zmerlynn: Should we back off on this change? I feel like we could rework this into something that would work for both Autopilot and Standard.

gongmax: Agreed. I'm looking into nodeSelector

gongmax: There is no common way to do workload isolation on Autopilot and Standard without auto-provisioning. I'm going to merge this for now. The instructions on isolate workload on Autopilot should be in separate PR.

zmerlynn: I think the documentation change is incorrect then? It still says it only applies to Standard.

gongmax: I changed it to "...only applies to the GKE Standard without node auto-provisioning" in last revision.

PR #3165, opened 2023-05-19
Promote SplitControllerAndExtensions to beta

chiayi:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

**What this PR does / Why we need it**:
This PR promotes `SplitControllerAndExtensions` to beta. 

**Which issue(s) this PR fixes**:
Part of https://github.com/googleforgames/agones/issues/2797
Fixes #3098

**Special notes for your reviewer**:

PR #3176, opened 2023-05-22
Consolidate cloud build between generic and autopilot

chiayi:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

There isn't a difference between generic and autopilot, so we clean up the cloud build yaml conditional.

Part of https://github.com/googleforgames/agones/pull/3165#discussion_r1199563868

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -320,10 +320,9 @@ steps:
             featureWithGate="CustomFasSyncInterval=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=false&PodHostname=false&SplitControllerAndExtensions=true&FleetAllocationOverflow=true&Example=true"
             testCluster="gke-autopilot-e2e-test-cluster-${version//./-}"
           fi
-          featureWithoutGate=""
           testClusterLocation="${region}"
           { stdbuf -oL -eL gcloud builds submit . --suppress-logs --config=./ci/e2e-test-cloudbuild.yaml \
-            --substitutions _FEATURE_WITH_GATE=$featureWithGate,_FEATURE_WITHOUT_GATE=$featureWithoutGate,_CLOUD_PRODUCT=$cloudProduct,_TEST_CLUSTER_NAME=$testCluster,_TEST_CLUSTER_LOCATION=$testClusterLocation,_REGISTRY=${_REGISTRY},_PARENT_COMMIT_SHA=${COMMIT_SHA},_PARENT_BUILD_ID=${BUILD_ID} \
+            --substitutions _FEATURE_WITH_GATE=$featureWithGate,_CLOUD_PRODUCT=$cloudProduct,_TEST_CLUSTER_NAME=$testCluster,_TEST_CLUSTER_LOCATION=$testClusterLocation,_REGISTRY=${_REGISTRY},_PARENT_COMMIT_SHA=${COMMIT_SHA},_PARENT_BUILD_ID=${BUILD_ID} \
             |& stdbuf -i0 -oL -eL grep -v " tarball " \
             |& stdbuf -i0 -oL -eL sed "s/^/${cloudProduct}-${version}: /"; } &
           pids+=($!)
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -312,18 +312,10 @@ steps:
         for version in "${!versionsAndRegions[@]}"
         do
           region=${versionsAndRegions[$version]}
-          if [ $cloudProduct = generic ]
-          then
-            featureWithGate="CustomFasSyncInterval=false&SDKGracefulTermination=false&StateAllocationFilter=false&PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=false&PodHostname=false&SplitControllerAndExtensions=false&FleetAllocationOverflow=true&Example=true"
-            featureWithoutGate=""
-            testClusterLocation="${region}-c"
```
gongmax, line 319: We cannot do this before move the Standard cluster to regional.

chiayi: Ooooooh I see! Okay I completely missed that. Thanks!

chiayi:
This will be in draft until the Standard clusters will be in regional. If for some reason, the standard clusters will never be in regional, this pr can then be deleted.

cloudbuild.yaml
roberthbailey: I think it's just the `featureWithGate` and `featureWithoutGate` variables that get collapsed.

chiayi: I see, so that means there isn't a need for a `featureWithoutGate` anymore. Just one variable `featurewithGate`

gongmax (approved):

gongmax (dismissed):

cloudbuild.yaml
roberthbailey: Correct. You should be able to clean up the usages of `featureWithoutGate` and (unless I'm misreading it) collapse the two places where `featurewithGate` is set into one.

cloudbuild.yaml
```diff
@@ -323,7 +323,7 @@ steps:
           featureWithoutGate=""
           testClusterLocation="${region}"
           { stdbuf -oL -eL gcloud builds submit . --suppress-logs --config=./ci/e2e-test-cloudbuild.yaml \
-            --substitutions _FEATURE_WITH_GATE=$featureWithGate,_FEATURE_WITHOUT_GATE=$featureWithoutGate,_CLOUD_PRODUCT=$cloudProduct,_TEST_CLUSTER_NAME=$testCluster,_TEST_CLUSTER_LOCATION=$testClusterLocation,_REGISTRY=${_REGISTRY},_PARENT_COMMIT_SHA=${COMMIT_SHA},_PARENT_BUILD_ID=${BUILD_ID} \
+            --substitutions _FEATURE_WITH_GATE=$featureWithGate,_CLOUD_PRODUCT=$cloudProduct,_TEST_CLUSTER_NAME=$testCluster,_TEST_CLUSTER_LOCATION=$testClusterLocation,_REGISTRY=${_REGISTRY},_PARENT_COMMIT_SHA=${COMMIT_SHA},_PARENT_BUILD_ID=${BUILD_ID} \
```
roberthbailey, line 326: If it are no longer used, can you remove the `featureWithoutGate` variable on lines 323 above?

roberthbailey, line 326: Unless I'm misreading them, the two places we set `featureWithGate` are now the same and can be consolidated into a single variable (e.g. moved outside of the if statement). 

The only thing that would remain inside of the `if [ $cloudProduct = generic ]` block is how we pick the name of the test cluster.

gongmax: I think `featureWithGate` are still different. For Generic, `SplitControllerAndExtensions` is set to false while in Autopilot it's set to true

chiayi: I think we can set both to be true now that its in beta?

gongmax: oh that makes sense. We can set both to true.

roberthbailey: Looking at https://github.com/googleforgames/agones/pull/3165 again, I think we may need to wait until we hit stable to consolidate the variables. That PR flipped the `featureWithGate` for non-autopilot to false whereas on autopilot it needs to stay true.

roberthbailey:
Thanks for confirming. I'm going to close this PR and we will consolidate once the feature graduates to stable.

PR #3180, opened 2023-05-23
Counts and Lists Aggregate Values for Fleet Status and GameServerSet Status

igooch:
**What type of PR is this?**

/kind feature

Aggregates the Counter (Count/Capacity) and List (Capacity/Values) across GamerServerSets and Fleet.

Counters and Lists are only aggregated on GamerServerSets when GameServer.Status is Allocated, Ready, or Reserved.

Updates the Fleet and GameServerSet CRDs to aggregate the List values to a total Count rather than listing all values in an array, which could get extremely long. This change is in keeping with the sample Fleet Status outlined in #2716.

Working on #2716 


pkg/gameserversets/controller.go
```diff
@@ -618,6 +619,14 @@ func computeStatus(list []*agonesv1.GameServer) agonesv1.GameServerSetStatus {
 		case agonesv1.GameServerStateReserved:
 			status.ReservedReplicas++
 		}
+
+		// TODO: Aggregates all Counters and Lists for all GameServer States that are not Deleting. This
+		// means that unlike player tracking below, this will aggregate values during GameServerStateCreating.
+		// Should I mimic player tracking below and only aggregate for cases Ready, Reserved, and Allocated?
```
gongmax, line 625: I think it depends on the use case. If I understand correctly, the major use case of the aggregated counters/lists is for Fleet autoscaling. So the values of the game servers that are not in ready reserved or allocated status ***should not*** be included. For example, if the autoscaler want to ensure there are at least 10 capacity for key A in the fleet, there is only one unhealthy game server in the fleet and it has 10 capacity for key A. Autoscaler should still scale up accordingly in this case.

pkg/fleets/controller.go
```diff
@@ -693,3 +701,47 @@ func (c *Controller) filterGameServerSetByActive(fleet *agonesv1.Fleet, list []*
...
+
+// mergeLists adds the contents of l2 into l1.
+func mergeLists(l1, l2 map[string]agonesv1.AggregatedListStatus) map[string]agonesv1.AggregatedListStatus {
+	if l1 == nil {
+		l1 = make(map[string]agonesv1.AggregatedListStatus)
+	}
+
+	for key, val := range l2 {
+		// If the List exists in both maps, aggregate the values.
+		// TODO: Will this cause issues with install/helm/agones/templates/crds/fleet.yaml because in
```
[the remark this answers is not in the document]
igooch: This is part of the Status, not the Spec of the Fleet, so they don't create the Counts and Lists as part of the fleet (they create Counts and Lists as part of a GameServer). The purpose of aggregating the values held in Counts and Lists in each gameserver is to allow the user to autoscale the fleet based on a Counts and List value (we haven't implemented this yet, that's next).

pkg/fleets/controller.go
```diff
@@ -693,3 +701,47 @@ func (c *Controller) filterGameServerSetByActive(fleet *agonesv1.Fleet, list []*
...
+		// the CRD we define a maximum capacity and maximum number of values as 1000? (Also a possible
+		// issue with there being a max of 1000 Lists in a Fleet per the CRD?) So far it passes unit tests...
+		if list, ok := l1[key]; ok {
+			list.Capacity += val.Capacity
+			// We do not remove duplicates here.
```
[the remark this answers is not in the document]
igooch: The purpose of aggregating the values held in Counts and Lists in each gameserver is to allow the user to autoscale the fleet based on a Counts and List value (we haven't implemented this yet, that's next).

pkg/gameserversets/controller.go
igooch: That makes sense, I'll update to only include status ready reserved or allocated.

pkg/fleets/controller.go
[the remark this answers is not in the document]
gongmax: Understood. My question was actually about the counters/lists in the Spec of the Fleet, since that's where the 1000 limitation was applied, right? If users don't create the Counts and Lists as part of the fleet, why we included them in the Spec of the Fleet?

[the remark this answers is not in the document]
igooch: It's under Status in the fleet.yaml, so I'm thinking that's where I should remove the limit?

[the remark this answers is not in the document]
gongmax: oh sorry my bad, I thought they are on Spec. Let's keep the limit for now. We can call out that the limitations are 1000 on both game server and fleet level.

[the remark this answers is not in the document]
igooch: Thinking about this more I think we need to get rid of the limit (and defaults) at the List/Counter level. I'm leaving in place the "maxProperties" limit of 1000 which limits the total number of Lists/Counters.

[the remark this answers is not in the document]
igooch: After our conversation there is no limit on the number of GameServers in a GameServerSet or Fleet, so the theoretical limit at the List/Counter level is infinite if there is no limit in place. We'll need further discussion on what this limit should be. In the meantime I'm removing CRD changes from this PR.

gongmax (approved):

pkg/gameserversets/controller.go
```diff
@@ -640,3 +650,44 @@ func computeStatus(list []*agonesv1.GameServer) agonesv1.GameServerSetStatus {
...
+	}
+
+	for key, val := range counterStatus {
+		// If the Counter exists in both maps, aggregate the values.
+		if counter, ok := aggCounterStatus[key]; ok {
+			counter.Count += val.Count
+			counter.Capacity += val.Capacity
+			aggCounterStatus[key] = counter
+		} else {
+			aggCounterStatus[key] = agonesv1.AggregatedCounterStatus(*val.DeepCopy())
```
gongmax, line 667: nit: `agonesv1.AggregatedCounterStatus(*val.DeepCopy())` - this is different than how you handle it in line 718, what's the difference?

igooch: We're casting type CounterStatus to type AggregatedCounterStatus, in 718 the inputs are both AggregatedCounterStatus, so we don't need to cast.

markmandel:
Looks like that's a pretty consistent failure:

```
--- FAIL: TestControllerUpdateFleetPlayerStatus (0.13s)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x22c5eb9]

goroutine 9 [running]:
testing.tRunner.func1.2({0x24b8740, 0x37be820})
	/usr/local/go/src/testing/testing.go:1526 +0x372
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1529 +0x650
panic({0x24b8740, 0x37be820})
	/usr/local/go/src/runtime/panic.go:890 +0x263
agones.dev/agones/pkg/fleets.TestControllerUpdateFleetPlayerStatus.func2({0x2a6f6b8?, 0xc00021cc40})
	/go/src/agones.dev/agones/pkg/fleets/controller_test.go:675 +0x199
k8s.io/client-go/testing.(*SimpleReactor).React(0xc00081e720, {0x2a6f6b8, 0xc00021cc40})
	/go/src/agones.dev/agones/vendor/k8s.io/client-go/testing/fixture.go:530 +0x5d
k8s.io/client-go/testing.(*Fake).Invokes(0xc0001248f0, {0x2a6f6b8, 0xc00021c9a0}, {0x2a56638, 0xc0001df200})
	/go/src/agones.dev/agones/vendor/k8s.io/client-go/testing/fake.go:145 +0x34b
agones.dev/agones/pkg/client/clientset/versioned/typed/agones/v1/fake.(*FakeFleets).UpdateStatus(0xc000d64108, {0x4adb65?, 0x5bc951?}, 0xc0001dec00, {{{0x0, 0x0}, {0x0, 0x0}}, {0x0, 0x0, ...}, ...})
	/go/src/agones.dev/agones/pkg/client/clientset/versioned/typed/agones/v1/fake/fake_fleet.go:110 +0x24a
agones.dev/agones/pkg/fleets.(*Controller).updateFleetStatus(0xc00039fe80, {0x2a6c7a8, 0xc00081e750}, 0xc0006bac00)
	/go/src/agones.dev/agones/pkg/fleets/controller.go:683 +0x9d3
agones.dev/agones/pkg/fleets.TestControllerUpdateFleetPlayerStatus(0xc000102680)
	/go/src/agones.dev/agones/pkg/fleets/controller_test.go:684 +0xc3e
testing.tRunner(0xc000102680, 0x27f1210)
	/usr/local/go/src/testing/testing.go:1576 +0x217
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1629 +0x806
```

pkg/fleets/controller_test.go
```diff
@@ -685,6 +685,130 @@ func TestControllerUpdateFleetPlayerStatus(t *testing.T) {
 	assert.True(t, updated)
 }
 
+// nolint // Linter errors on lines are duplicate of TestControllerUpdateFleetListStatus
```
markmandel, line 688: (non blocking) nolint at all, or did you want to only nolint:duplicate (or whatever the duplicate lint tag is?)

pkg/gameserversets/controller.go
```diff
@@ -618,6 +619,15 @@ func computeStatus(list []*agonesv1.GameServer) agonesv1.GameServerSetStatus {
 		case agonesv1.GameServerStateReserved:
 			status.ReservedReplicas++
 		}
+
+		// Aggregates all Counters and Lists only for GameServer states Ready, Reserved, or Allocated.
+		if runtime.FeatureEnabled(runtime.FeatureCountsAndLists) &&
+			(gs.Status.State == agonesv1.GameServerStateReady ||
```
markmandel, line 625: (non blocking): Could also make this a function much like this one:

https://github.com/googleforgames/agones/blob/adfb30845d1b06cf3a04d84f2bbd19f128056254/pkg/apis/agones/v1/gameserver.go#L686

i feel like we might have this kind of check in a few places 🤔 ?

markmandel (approved):
Approved!

Left a few non-blocking questions for you in case they made sense.

markmandel (approved):
Approved!

Left a few non-blocking questions for you in case they made sense.

pkg/fleets/controller_test.go
igooch: Good call, I believe it's the dubl tag.

pkg/gameserversets/controller.go
igooch: I added in an "IsReady" function to gameserver.go. Let me know if you can think of a better name for this func.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -700,6 +700,20 @@ func (gs *GameServer) IsBeforeReady() bool {
 	return false
 }
 
+// IsReady returns true if the GameServer status is Ready, Reserved, or Allocated state.
+func (gs *GameServer) IsReady() bool {
```
markmandel, line 704: Late to the party:

Maybe `IsActive` ? Since `Ready` is the Ready state, this might get confusing?

Not sure how much I like the idea of `Active` either, but I figure `Ready` is more confusing. WDYT?

igooch: I like it. Since this is already merged I'll make the change in the next PR.

PR #3188, opened 2023-05-29
Create an agones-dev user in the build docker images and map the user/group id to the host linux system

roberthbailey:
This fixes permission issues for files that are shared between the host and container, such as all of the common_mounts that are defined in the Makefile.

**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Closes #

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -21,8 +21,7 @@ ARG GROUP_ID
 RUN groupadd -g ${GROUP_ID:-1338} agones-dev && \
     useradd -l -u ${USER_ID:-1337} -g agones-dev -s /bin/bash agones-dev && \
     install -d -m 0755 -o agones-dev -g agones-dev /home/agones-dev && \
-    mkdir /home/agones-dev/.cache && mkdir /home/agones-dev/.config && \
-    mkdir /go/pkg
+    mkdir /home/agones-dev/.cache && mkdir /home/agones-dev/.config
 
 RUN apt-get update && \
     apt-get install -y build-essential gnupg curl git wget psmisc rsync make python bash-completion \
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -16,6 +16,14 @@
 
 FROM debian:bullseye
 
+ARG USER_ID
+ARG GROUP_ID
+RUN groupadd -g ${GROUP_ID:-1338} agones-dev && \
+    useradd -l -u ${USER_ID:-1337} -g agones-dev -s /bin/bash agones-dev && \
+    install -d -m 0755 -o agones-dev -g agones-dev /home/agones-dev && \
+    mkdir /home/agones-dev/.cache && mkdir /home/agones-dev/.config && \
+    mkdir /go/pkg
+
 RUN apt-get update && \
     apt-get install -y build-essential gnupg curl git wget psmisc rsync make python bash-completion \
     zip nano jq graphviz gettext-base plantuml software-properties-common && \
@@ -35,7 +43,7 @@ WORKDIR /opt
 # credits https://cloud.google.com/sdk/docs/install#deb
 RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - && apt-get update -y && \
     apt-get install google-cloud-cli google-cloud-cli-gke-gcloud-auth-plugin google-cloud-cli-app-engine-go -y && \
-    echo "source /usr/share/google-cloud-sdk/completion.bash.inc" >> /root/.bashrc
+    echo "source /usr/share/google-cloud-sdk/completion.bash.inc" >> /home/agones-dev/.bashrc
 
 # update the path for go
 ENV PATH /usr/local/go/bin:/go/bin:$PATH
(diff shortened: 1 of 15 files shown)


roberthbailey:
The error is:

```
docker build --tag=agones-build:6efc8db6ef /workspace/build//build-image \
	--build-arg USER_ID="0" \
	--build-arg GROUP_ID="0" \
...
 ---> Running in 6e465ad8ad5b
groupadd: GID '0' already exists
```

So my trick of making the new user use the same group & user id as the system may not work on cloudbuild because it seems to be running the scripts as the root user.

build/build-image/Dockerfile
```diff
@@ -16,6 +16,14 @@
 
 FROM debian:bullseye
 
+ARG USER_ID
+ARG GROUP_ID
+RUN groupadd -g ${GROUP_ID:-1338} agones-dev && \
```
markmandel, line 21: On that's very interesting - you set the userid at build time. I always did it at runtime, but this makes FAR more sense.

build/includes/build-image.mk
```diff
@@ -23,15 +23,18 @@ build_remote_tag = $(REGISTRY)/$(build_tag)
 
 # Creates the build docker image
 build-build-image:
-	docker build --tag=$(build_tag) $(build_path)/build-image $(DOCKER_BUILD_ARGS)
+	docker build --tag=$(build_tag) $(build_path)/build-image \
+		--build-arg USER_ID="$(shell id -u ${USER})" \
```
markmandel, line 27: Would it work such that if `$(shell id -u ${USER})` was 0, then skip this step (and use the defaults)?

Could even do it as a top level Makefile items, such as:

```makefile
ifneq ($(shell id -u ${USER}), 0)
  docker_build_user := -build-arg USER_ID="$(shell id -u ${USER})" --build-arg GROUP_ID="$(shell id -g ${USER})"
endif
```

And then whenever building a docker image, use ${docker_build_user} in space of the details, and if it's root, it'll get ignored, which will be fine.

WDYT?

markmandel (commented):
Just reviewing this idea - I had some thoughts on how we could work around the Cloud build issue.

build/includes/build-image.mk
roberthbailey: That's a good idea (I was thinking something along those lines and doing it once in the Makefile way would make a lot of sense). 

One other idea that @zmerlynn suggested was to see if we could have cloud build run as a non-root user. But I think your suggestion might be simpler to implement (and test locally).

markmandel: > One other idea that @zmerlynn suggested was to see if we could have cloud build run as a non-root user. But I think your suggestion might be simpler to implement (and test locally).

I had a look to see if you could easily get cloud build to do that, and I couldn't find anything.

roberthbailey: I found one place that suggested that instead of using `gcr.io/cloud-builders/docker` you could create a new build container based off of that container and add a user to it. But this solution doesn't seem great because we'd have to keep updating a custom build image. 

The solution you proposed also adds a user (so the build runs as non-root), but doesn't require us to maintain our own base build container.

roberthbailey: Pushed a new commit with this suggestion. Let's see how it works. 🤞

build/build-sdk-images/cpp/Dockerfile
```diff
@@ -14,6 +14,12 @@
 ARG BASE_IMAGE=agones-build-sdk-base:latest
 FROM $BASE_IMAGE
 
+ARG USER_ID
```
markmandel, line 17: 🤔 do you need to repeat this, given that it's already in the base?

build/build-sdk-images/rust/Dockerfile
```diff
@@ -14,6 +14,12 @@
 ARG BASE_IMAGE=agones-build-sdk-base:latest
 FROM $BASE_IMAGE
 
+ARG USER_ID
+ARG GROUP_ID
+RUN userdel -f agones-dev && if getent group agones-dev ; then groupdel agones-dev; fi &&\
```
markmandel, line 19: I saw this in a few templates, but I'm not sure I get what it's doing? 🤔

build/build-image/Dockerfile
```diff
@@ -16,6 +16,13 @@
...
+    useradd -l -u ${USER_ID:-1337} -g agones-dev -s /bin/bash agones-dev && \
+    install -d -m 0755 -o agones-dev -g agones-dev /home/agones-dev && \
+    mkdir /home/agones-dev/.cache && mkdir /home/agones-dev/.config
```
markmandel, line 24: Curious why you need to create these directories? I take it they don't get created by default by the tools that need them?

build/build-image/Dockerfile
```diff
@@ -16,6 +16,13 @@
 
 FROM debian:bullseye
 
+ARG USER_ID
+ARG GROUP_ID
```
markmandel, line 20:
```suggestion
ARG USER_ID=1337
ARG GROUP_ID=1338
```

https://docs.docker.com/engine/reference/builder/#default-values

Then you don't have to set them separately everywhere 😄

build/build-sdk-images/cpp/Dockerfile
```diff
@@ -27,5 +33,9 @@ RUN cmake --version
 WORKDIR /go/src/agones.dev/agones
 
 # code generation scripts
-COPY *.sh /root/
-RUN chmod +x /root/*.sh
+COPY *.sh /home/agones-dev/
+RUN chmod +x /home/agones-dev/*.sh
+
+RUN chown --silent --no-dereference --recursive ${USER_ID:-1337}:${GROUP_ID:-1338} /home/agones-dev /go
```
markmandel, line 39: may need to allow "other" to write in these dirs? Root should have privileges though, right?

zmerlynn:
Let's go ahead and close this for now, I'll fish out the changes if we need it.

PR #3187, opened 2023-05-26
build: add ltsc2022 target for windows builds

davidedmondsMPG:
**What type of PR is this?**

/kind cleanup

**What this PR does / Why we need it**:

Creates Windows container images for Agones on the ltsc2022 Windows base image, which is required to run on Windows Server 2022.

**Which issue(s) this PR fixes**:

Contributes to https://github.com/googleforgames/agones/issues/54 but does not completely close.

**Special notes for your reviewer**:

I don't have access to GCP-hosted registries, so pushed these to AWS ECR instead. Had issues with the docker manifest not being able to be generated but was not certain that this was not an ECR limitation. Am happy to make further changes if the issues also appear on the registry Agones is using.

PR #3189, opened 2023-05-30
windows build for ltsc2022

ashutosji:
**What type of PR is this?** This is the fix of this PR#[3187](https://github.com/googleforgames/agones/pull/3187)
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

**What this PR does / Why we need it**: We will get ltsc2022 target for windows builds 

**Which issue(s) this PR fixes**: It won't fix any issue. We do have ltsc2019 taget for windows build, same we are trying to get for ltsc2022

Closes #


gongmax:
> Hey - looks like this is very similar to my submitted PR #3187, but replacing 2019 with 2022 rather than building both. I suspect we'd like to maintain both versions in the future (Windows containers being more sensitive to base images) for some period of time, but am not a maintainer so can't make that call myself :)

Agreed. Let's keep 2019 for now. Closing this accordingly. Thanks for contribution!

PR #3196, opened 2023-06-01
Add make gen-crd-clients to the CI suite

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Closes #


build/Makefile
```diff
@@ -842,3 +842,10 @@ update-go-deps:
 	$(DOCKER_RUN) go get -u agones.dev/agones
 	$(DOCKER_RUN) go mod tidy
 	$(DOCKER_RUN) go mod vendor
+
+test-gen-crd-client:
+    mkdir -p build/tmp
+    mv pkg/client build/tmp
+    make gen-crd-client
+    diff -r build/tmp/client pkg/client || true
```
gongmax, line 850: The indent looks wrong. Each command should be after a `Tab`.

build/Makefile
```diff
@@ -842,3 +842,10 @@ update-go-deps:
 	$(DOCKER_RUN) go get -u agones.dev/agones
 	$(DOCKER_RUN) go mod tidy
 	$(DOCKER_RUN) go mod vendor
+
+test-gen-crd-client:
+	mkdir -p build/tmp
+	mv ../pkg/client build/tmp
+	make gen-crd-client
+	diff -r build/tmp/client ../pkg/client || true
```
markmandel, line 850: We do want this to fail , so let's remove the || true

For example: https://github.com/googleforgames/agones/blob/02365d2e0ea3b7605824a11f4980aec1a616c153/build/includes/website.mk#L111

(Although we don't need to sort, I don't think, but if this consistently fails, let's see what other approach)

markmandel: As another though, is it worth deleting `build/tmp` if the diff passes?

cloudbuild.yaml
```diff
@@ -115,6 +115,17 @@ steps:
...
   args: ["ensure-build-sdk-image-base"]
 
+#
+# Preventing Broken PR Merges in CI
+#
+- name: "make-docker"
+  id: test-gen-crd-client
+  waitFor:
+    - pull-build-image
+    - pull-build-sdk-base-image
```
markmandel, line 125:
```suggestion
```

I don't think we need the sdk-base-image at this stage, so we can likely drop this.

build/Makefile
Kalaiselvi84: The code has been modified as per your suggestion.

cloudbuild.yaml
Kalaiselvi84: I've removed this line.

markmandel (approved):

PR #3197, opened 2023-06-01
Add labels and annotations to allocation response

### LINKED ISSUE #2396: Return labels and annotations with GameServerAllocation results.
I'm using the multi-cluster setup to allocate new GameServer. 
Because my nodes do not have any external IP's, I must create an ISTIO routing to communicate with my GameServer. This happens with an operator. Works perfect. My GameServer can receive the traffic if I know the endpoint.

The problem is, that I'm unable to determine which cluster has create the new GameServer because the response of the AllocationService do not provide this information:
```json
{
  "gameServerName": "demo-server-77q8b-s6wb9",
  "ports": [
    {
      "name": "default",
      "port": 7534
    }
  ],
  "address": "ip-10-250-9-35.eu-west-1.compute.internal",
  "nodeName": "ip-10-250-9-35.eu-west-1.compute.internal"
}
```

As mention before, I can't use the nodeName because this is just the k8s internal hostname which has always the same pattern for all clusters. To implement my mapping I need the cluster name or the `GameServerAllocationPolicy` name which has affected.

I can name my Fleet in that way, that I can recognize the cluster by the name of the GameServer - but I find this solution really ugly. 

Question: is it possible to add the cluster name to the AllocationService response?

austin-space:
**What type of PR is this?**

/kind feature

*This PR is not complete*. It almost certainly needs more tests and I haven't touched the documentation. I'm mostly looking for a quick gut check to see if I'm missing anything glaring.

This PR returns the labels and annotations of an allocated game server at allocation time with the allocation response or Allocation CR status. This opens up a few very interesting scenarios:
* the ability to get some information from the game server that was generated by the game server(additional connection info)
* the ability to game server state information(such as map or hardware class)
* the ability to get insight into how allocations are getting processed(for example, we can get an idea of how far we're falling through our allocation selector chain)

Closes #2396 

~~I tried to break out the changes I've made thus far into two commits: the first being my changes, the second being the result of running the various `gen` make commands. Hopefully this makes this PR a little more readable.~~ I undid most of the autogen stuff since the sdk autogen stuff appears to be broken(related to the counts and lists changes it seems).

Also worth noting that I originally tried to put this behind a feature flag, but I don't think that actually makes much sense since the fields will be in the protobuf no matter whether we put something in there or not and it could potentially be confusing for users since there is no alpha allocation protobuf. If you want I can add the feature flag back.


gongmax:
linter failure
```
pkg/apis/allocation/v1/gameserverallocation.go:614:6: exported: exported type GameServerMetadata should have comment or be unexported (revive)
type GameServerMetadata struct {
     ^
```

austin-space:
I removed the sdk generation and a change I had made to get that working for node js. Just as a heads up, it seems like that may have been broken since the last time someone made changes to the protobuf a few months ago.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -607,7 +607,14 @@ type GameServerAllocationStatus struct {
 	NodeName       string                          `json:"nodeName,omitempty"`
 	// If the allocation is from a remote cluster, Source is the endpoint of the remote agones-allocator.
 	// Otherwise, Source is "local"
-	Source string `json:"source"`
+	Source   string              `json:"source"`
+	Metadata *GameServerMetadata `json:"metadata,omitempty"`
+}
+
+// GameServerMetadata is the metadata from the allocated game server at allocation time
+type GameServerMetadata struct {
```
markmandel, line 615: 🤔  this is exactly the same at `MetaPatch` - I know the naming is not great, but should we just reuse the type?

proto/allocation/allocation.proto
```diff
@@ -91,12 +91,18 @@ message AllocationResponse {
...
   string source = 6;
+  optional GameServerMetadata metadata = 7;
 
   // The gameserver port info that is allocated.
   message GameServerStatusPort {
     string name = 1;
     int32 port = 2;
   }
+
+  message GameServerMetadata {
```
markmandel, line 102: Again, same structure as `MetaPatch` - should we reuse it?

Or do we have the concern the we can't expand if we need to later?

markmandel (commented):
Thanks for doing all this work! Added a few questions inline.

Only other things I can think of to add:

1. Updates to the documentation, at the very least the [GameServerAllocation Specification](https://agones.dev/site/docs/reference/gameserverallocation/) 
2. Update to the example (which is basically the same as above) https://github.com/googleforgames/agones/blob/main/examples/gameserverallocation.yaml
3. End to end tests , could probably extend what we already have, primarily:

https://github.com/googleforgames/agones/blob/9c33780753002ba414f77e393507bc4907f30cd4/test/e2e/gameserverallocation_test.go#L39

and 

https://github.com/googleforgames/agones/blob/9c33780753002ba414f77e393507bc4907f30cd4/test/e2e/allocator_test.go#L151

Will suffice.

But really excited to have this in Agones!

@gongmax @zmerlynn - do we see any reason to put this behind a feature gate? I think it's likely fine, since I can't think of anything we'd want to change in this, but figured I would ask.

proto/allocation/allocation.proto
austin-space: That's the primary concern. I did think about doing exactly that, but doing so would cause issues with back/forwards compatibility if someone wanted to change either of them. They're semantically different enough that it seemed risky.

markmandel: That makes sense. I rescind my question then 👍🏻

zmerlynn:
> @gongmax @zmerlynn - do we see any reason to put this behind a feature gate? I think it's likely fine, since I can't think of anything we'd want to change in this, but figured I would ask.

Since it's a new return from the API, I can't imagine needing a feature gate - rather, I think the question is whether it should be introduced straight to the `v1` surface or go through the beta API surface first. I think the answer to that is still no, though - I'm ok with this going straight to `v1`.

gongmax:
> > @gongmax @zmerlynn - do we see any reason to put this behind a feature gate? I think it's likely fine, since I can't think of anything we'd want to change in this, but figured I would ask.
> 
> Since it's a new return from the API, I can't imagine needing a feature gate - rather, I think the question is whether it should be introduced straight to the `v1` surface or go through the beta API surface first. I think the answer to that is still no, though - I'm ok with this going straight to `v1`.

I agreed. Previously we introduced the `Source` field in allocation response, which was not behind a feature gate. I cannot think of any drawback to not follow the same pattern.

markmandel:
SGTM!

in which case one we have some docs, examples and e2e tests, this should be good to merge 👍🏻 🤸🏻

site/content/en/docs/Reference/gameserverallocation.md
```diff
@@ -148,6 +148,18 @@ The `spec` field is the actual `GameServerAllocation` specification, and it is c
 - `metadata` is an optional list of custom labels and/or annotations that will be used to patch
   the game server's metadata in the moment of allocation. This can be used to tell the server necessary session data
 
+Once created the `GameServerAllocation` will have a `status` field consisting of the following:
```
markmandel, line 151: Oh good point! We never documented this here!

site/content/en/docs/Reference/gameserverallocation.md
```diff
@@ -148,6 +148,18 @@ The `spec` field is the actual `GameServerAllocation` specification, and it is c
...
+
+- `state` is the current state of a GameServerAllocation, e.g. `Allocated`, or `UnAllocated`
```
markmandel, line 153: Minor nit

```suggestion
- `State` is the current state of a GameServerAllocation, e.g. `Allocated`, or `UnAllocated`
```

test/e2e/gameserverallocation_test.go
```diff
@@ -608,6 +608,52 @@ func TestGameServerAllocationPreferredSelection(t *testing.T) {
...
+	flt := defaultFleet(framework.Namespace)
+	flt.Spec.Replicas = 1
+	flt.Spec.Template.ObjectMeta.Labels = label
+	flt.Spec.Template.ObjectMeta.Annotations = annotations
+	flt, err := fleets.Create(ctx, flt, metav1.CreateOptions{})
+	if assert.Nil(t, err) {
+		defer fleets.Delete(ctx, flt.ObjectMeta.Name, metav1.DeleteOptions{}) // nolint:errcheck
+	} else {
+		assert.FailNow(t, "could not create first fleet")
+	}
```
markmandel, line 632: Small code cleanup:

```suggestion
  defer fleets.Delete(ctx, flt.ObjectMeta.Name, metav1.DeleteOptions{}) // nolint:errcheck
  require.NoError(t, err)
```

test/e2e/gameserverallocation_test.go
```diff
@@ -608,6 +608,52 @@ func TestGameServerAllocationPreferredSelection(t *testing.T) {
...
+		assert.Equal(t, allocationv1.GameServerAllocationAllocated, gsa.Status.State)
+		assert.Equal(t, t.Name(), gsa.Status.Metadata.Labels[role])
+		assert.Equal(t, flt.ObjectMeta.Name, gsa.Status.Metadata.Labels[agonesv1.FleetNameLabel])
+		assert.Equal(t, annotationValue, gsa.Status.Metadata.Annotations[annotationKey])
+		gs, err := gameServers.Get(ctx, gsa.Status.GameServerName, metav1.GetOptions{})
+		assert.Nil(t, err)
+		assert.Equal(t, flt.ObjectMeta.Name, gs.ObjectMeta.Labels[agonesv1.FleetNameLabel])
+	} else {
+		assert.FailNow(t, "could not completed gsa1 allocation")
+	}
```
markmandel, line 654: Small code cleanup

```suggestion
  require.NoError(t, err)

  assert.Equal(t, allocationv1.GameServerAllocationAllocated, gsa.Status.State)
  assert.Equal(t, t.Name(), gsa.Status.Metadata.Labels[role])
  assert.Equal(t, flt.ObjectMeta.Name, gsa.Status.Metadata.Labels[agonesv1.FleetNameLabel])
  assert.Equal(t, annotationValue, gsa.Status.Metadata.Annotations[annotationKey])
  gs, err := gameServers.Get(ctx, gsa.Status.GameServerName, metav1.GetOptions{})
  require.NoError(t, err)
  assert.Equal(t, flt.ObjectMeta.Name, gs.ObjectMeta.Labels[agonesv1.FleetNameLabel])
```

pkg/sdk/sdk_grpc.pb.go
```diff
@@ -16,7 +16,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.2.0
-// - protoc             v3.21.6
+// - protoc             v3.21.12
```
markmandel, line 19: 🤔  wondering why this is here? (and in the other sdk files)

markmandel (commented):
Looks good! Added some notes on some code cleanup.

One extra e2e test to add and I think we're good to go:
https://github.com/googleforgames/agones/blob/main/test/e2e/allocator_test.go

Just to e2e test the conversion to grpc and rest work as expected - you could even just add labels and annotations to the existing tests rather than spin up a whole new one if that's easier (also avoids waiting for the tls certs to refresh any more).

pkg/sdk/sdk_grpc.pb.go
austin-space: That's a good question, and one I can't answer. I don't have a standalone install of protoc on my machine. Something I did notice is that some of the protobuf seems to have not been generated for some client libraries(if you run it on main, it fails) so this may be an artifact of that?

markmandel: Ooh I see - this is an artifact of running the `make gen-allocation-grpc`? (I would have thought it would have only happened for `make gen-sdk-grpc` 🤔  weird.

I mean it's a change in a comment - not the end of the world 🤷🏻 weird though.

test/e2e/allocator_test.go
```diff
@@ -208,6 +208,7 @@ func TestAllocatorWithSelectors(t *testing.T) {
 			require.NoError(t, err)
 			require.Equal(t, response.GameServerName, allocatedResponse.GameServerName)
 			helper.ValidateAllocatorResponse(t, allocatedResponse)
+			assert.Equal(t, flt.ObjectMeta.Name, allocatedResponse.Metadata.Labels[agonesv1.FleetNameLabel])
```
markmandel, line 211: Clever!

markmandel (approved):
Looks good! Sorry about the delay in final approval!

PR #3198, opened 2023-06-01
Adds Counters and Lists to FleetAutoScaler CRD

igooch:
**What type of PR is this?**

/kind feature

Adds Counters and Lists to FleetAutoScaler Custom Resource Definition.
Adds Counters and Lists to the full Fleet example.
Adds Counters and Lists FleetAutoScaler examples.
Adds Counters and Lists autoscaling policies to the FleetAutoScaler and provides validation of the policies.

Working on #2716 

This is just the CRD and CRD validation change, and autoscaling by Counter or List does not yet work as I still need to make changes to agones/pkg/fleetautoscalers/fleetautoscalers.go which will be coming up in the next PR.


examples/counterfleetautoscaler.yaml
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+ name: fleet-autoscaler-counter
+spec:
+ fleetName: fleet-example
+ policy:
+   type: Counter # Counter based autoscaling
+   counter:
+     # The name of the Counter.
+     key: players
+     # Size of a buffer of counted items that are available in the Fleet.
```
gongmax, line 37: Is this size of a buffer based on the counter capacity that is available over the current aggregate counter capacity in the Fleet?

examples/counterfleetautoscaler.yaml
```diff
@@ -0,0 +1,45 @@
...
+     # it can be specified either in absolute (i.e. 5) or percentage format (i.e. 5%)
+     bufferCount: 5
```
gongmax, line 39: Is the `bufferCount` required? A policy without `bufferCount` seems meaningless.

igooch: True, this should be required. I also can't decide whether or not to make "Key" required as well (maybe a discussion for later) as it would be difficult to scale based on all possible counters in the fleet.

examples/counterfleetautoscaler.yaml
igooch: We're overloading the word "Count" here, it really means "Available Capacity" which is Capacity - Count. Should we rename this something besides Count or Capacity? Perhaps minAvailability, maxAvailability, bufferAvailability (or something else)?

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -328,6 +400,130 @@ func (b *BufferPolicy) ValidateBufferPolicy(causes []metav1.StatusCause) []metav
 	return causes
 }
 
+// ValidateCounterPolicy validates the FleetAutoscaler Counter policy settings
+// #TODO: Does NOT validate if a Counter with name CounterPolicy.Key is present in the fleet.
+// nolint:dupl  // Linter errors on lines are duplicate of ValidateListPolicy
+func (c *CounterPolicy) ValidateCounterPolicy(causes []metav1.StatusCause) []metav1.StatusCause {
```
gongmax, line 406: I think Key should be required because I can't think of an use case where user wants to scale based on all counters. WDYT?

igooch: Agreed it should be required -- I didn't see a way of validating here that the key actually exists in the fleet since the fleet isn't passed through into the fleetautoscaler.go.

gongmax: Make sense. I think we can make it a no-op in controller if the fleet doesn't have the key.

examples/listfleetautoscaler.yaml
```diff
@@ -31,15 +31,18 @@ spec:
...
+      # Size of a buffer based on the list capacity that is available over the current aggregate
+      # list length in the Fleet (available capacity).
+      # bufferLength can be specified either in absolute (i.e. 5) or percentage format (i.e. 5%)
+      # Required.
       bufferLength: 5
-      # minimum aggregate list capacity that can be provided by this FleetAutoscaler.
-      # if not specified, the actual minimum capacity will be bufferLength
+      # Minimum aggregate list capacity that can be provided by this FleetAutoscaler.
+      # If minLength is not specified, the actual minimum capacity will be bufferLength.
       minLength: 10
```
gongmax, line 45: Do you think if we should change min/maxLength to min/maxCapacity too to keep consistent with the Counter?

examples/counterfleetautoscaler.yaml
igooch: Redid the naming to clarify it's min/maxCapacity = total Capacity (not available capacity as stated above).

examples/listfleetautoscaler.yaml
igooch: Makes sense, I'll make that change.

gongmax (approved):

examples/counterfleetautoscaler.yaml
```diff
@@ -0,0 +1,47 @@
...
+ policy:
+   type: Counter # Counter based autoscaling
+   counter:
+     # The name of the Counter.
+     # Required.
+     key: players
+     # Size of a buffer of counted items that are available in the Fleet (available capacity).
+     # bufferCount can be specified either in absolute (i.e. 5) or percentage format (i.e. 5%).
+     # Required.
+     bufferCount: 5
```
markmandel, line 41: I'm wondering 🤔  Should this be:

```suggestion
     bufferSize: 5
```

To match the `Buffer` auto scaling strategy, or should it be:

```suggestion
     bufferCapacity: 5
```

To match `min/maxCapacity` ?

Now that we changed `min/max` this feels out of place, and that way we could also be the same across `Count` and `List` based scaling (and we could consolidate Go types as well)

WDYT?

igooch: Makes sense, I made the change to bufferSize.

Issue #3208, opened 2023-06-09
Hardcoded namespace breaks `agones-controller` deployment in non-default namespaces


maxcorbin:
# Expected
Upgrading to `>=1.31.0` should be backwards compatible with existing deployments.

# Actual
Upgrading to `>=1.31.0` causes the controller to be unresponsive to API calls in deployments outside of namespaces named `agones-system`. This prevents fleets from autoscaling or from replacing shutdown gameservers.

# Description
In `main.go` for the `agones-controller` deployment, there is a hard-coded namespace argument that is set to `agones-system`. While I know that this is the recommended setup, the helmchart does accept other namespaces and will deploy just fine into `default`, for example. Such deployments are broken on `>=1.31.0` due to the new `SplitControllerAndExtensions` work. [I believe the problem is here.](https://github.com/googleforgames/agones/blob/9840265f3855f7db143e28ab3c14ad5431fe6572/cmd/controller/main.go#L465)

# Diagnosis
The issue can be diagnosed by the presence of this in the controller logs:

```
leaderelection.go:334] error initially creating leader election record: namespaces "agones-system" not found
```

# Attempted Solutions
I tried scaling down to 1 replica and I also tried turning the `SplitControllerAndExtensions` flag off in the hopes that it would avoid the codepath where the hardcoded value is, but this doesn't work. I believe that the namespace value is not gated behind any control flow or other logic, and that it will affect all deployments that use a non-default namespace for versions 1.31.0 and on, even when leader election is not needed.

# Workaround
We resolved this in the short term by rolling back to `1.30.0`. Obviously we would prefer not to stay on the older version for too long, but we are unable to upgrade until this is fixed. I don't see an easy way to refer to the namespace that is provided to the deployment in the Helm chart, or else I would be happy to provide a PR to fix the problem.

Please let me know if there's additional information needed, or if I am missing something simple that would help me write a PR.

markmandel:
That's a legit bug. Thanks for filing the issue!

zmerlynn:
Fixed by #3209 - thanks for the report!

PR #3209, opened 2023-06-09
Have leader election use namespace from env var

chiayi:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Currently leader election uses hardcoded namespace, this changes so that it uses the namespace from the env vars

Towards #3208 

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -84,6 +84,7 @@ const (
 	kubeconfigFlag               = "kubeconfig"
 	allocationBatchWaitTime      = "allocation-batch-wait-time"
 	defaultResync                = 30 * time.Second
+	podNamespace                 = "pod-namespace"
 	leaderElectionFlag           = "leader-election"
 )
 
@@ -249,7 +250,7 @@ func main() {
 	// This allows the controller to not fail health check during install when there is replication
 	go runRunner(server)
 
-	whenLeader(ctx, cancel, logger, ctlConf.LeaderElection, kubeClient, func(ctx context.Context) {
+	whenLeader(ctx, cancel, logger, ctlConf.LeaderElection, kubeClient, ctlConf.PodNamespace, func(ctx context.Context) {
 		kubeInformerFactory.Start(ctx.Done())
 		agonesInformerFactory.Start(ctx.Done())
 
@@ -282,6 +283,7 @@ func parseEnvFlags() config {
 	viper.SetDefault(enableStackdriverMetricsFlag, false)
 	viper.SetDefault(stackdriverLabels, "")
 	viper.SetDefault(allocationBatchWaitTime, 500*time.Millisecond)
+	viper.SetDefault(podNamespace, "agones-system")
 	viper.SetDefault(leaderElectionFlag, false)
 
 	viper.SetDefault(projectIDFlag, "")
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -84,6 +84,7 @@ const (
 	kubeconfigFlag               = "kubeconfig"
 	allocationBatchWaitTime      = "allocation-batch-wait-time"
 	defaultResync                = 30 * time.Second
+	podNamespace                 = "pod-namespace"
```
zmerlynn, line 87: I would just call this `lease-namespace` and make it clear it's the namespace to use for the controller lease. The helm chart can make sure it's aligned appropriately.

Strictly speaking, I don't think this needs to be in the same namespace as the pod - I think the controller just needs rights to it and for it to exist, which was the issue, AIUI.

chiayi: Would this be confusing since the namespace is the same one as the controller pods? Or should there be another env variable for leader election?

markmandel: As a note, this doesn't (and probably shouldn't?) come from helm. We can use the downward API to get this information from K8s directly:

https://kubernetes.io/docs/concepts/workloads/pods/downward-api/

markmandel: Just thinking, someone could do a yaml install to a different namespace, and expect it to work. If we use the downward API, this handles that scenario.

chiayi: It should be already using the downward api: https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/controller.yaml#L159-L162 is there a big difference between exposing it as a env variable vs a downward-api volume?

markmandel: Oh sweet - I didn't realise it was already there.

If it's already there as an env var, then you are good to go!

zmerlynn: If this is already plumbed into the deployment then I have no problem with the existing name - AIUI you're basically consuming an env variable that we already set here, so I guess my `lease-namespace` comment is ignorable.

cmd/controller/main.go
```diff
@@ -282,6 +283,7 @@ func parseEnvFlags() config {
 	viper.SetDefault(enableStackdriverMetricsFlag, false)
 	viper.SetDefault(stackdriverLabels, "")
 	viper.SetDefault(allocationBatchWaitTime, 500*time.Millisecond)
+	viper.SetDefault(podNamespace, "agones-systems")
```
markmandel, line 286:
```suggestion
	viper.SetDefault(podNamespace, "agones-system")
```

markmandel (commented):
Saw a small typo below, but other than that, no issues on my end here. @zmerlynn any on your end?

Other than the typo, this look good to merge?

zmerlynn:
> Saw a small typo below, but other than that, no issues on my end here. @zmerlynn any on your end?

No, this LGTM other than the typo - wasn't sure if @chiayi was ready.

cmd/controller/main.go
chiayi: Fixed!

markmandel (approved):
Noice!

PR #3206, opened 2023-06-08
Add dev to prod data replication script

Kalaiselvi84:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind feature

**What this PR does / Why we need it**:

**Which issue(s) this PR fixes**:

Closes #

**Special notes for your reviewer**:

PR #3210, opened 2023-06-10
Remove Feature Expiry Version Shortcodes

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind feature

Closes #


build/scripts/remove-feature-expiryVersion/main.go
```diff
@@ -0,0 +1,129 @@
...
+	var sb strings.Builder
+	expiryBlock := false
+	publishBlock := false
+	preserveLines := true
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		// Check if the line contains the starting of the expiryVersion shortcode with the specified version
+		if strings.Contains(line, "{{% feature expiryVersion=\""+version+"\" %}}") {
```
markmandel, line 96: Just so you know, you can have both `{{% .... %}}` as shortcode, but also `{{< ... >}}` shortcodes as well, so we should support both.

https://gohugo.io/content-management/shortcodes/

markmandel (commented):
This is definitely looking good! Going to remove so much hand editing!

And I think once #3215 lands, we could probably also get rid of that "extra configuration" table at the bottom that makes this trickier (in fact, I'll update that PR now so that that is now gone).

Also, as per #3206 - this should be a `make` target as well please.

build/scripts/remove-feature-expiryVersion/main.go
Kalaiselvi84: I have added both shortcodes in the script and created a `feature-shortcode-update` target in website.mk

build/scripts/feature-shortcode-update/main.go
```diff
@@ -0,0 +1,158 @@
...
+	"bufio"
+	"flag"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func main() {
+	dirPath := "site/content/en/docs"
```
Kalaiselvi84, line 28: Based on my examination of the previous releases, it appears that the shortcodes are only present in the site/content/en/docs directory. Kindly inform me if there are any other directories that require the inclusion of these shortcodes.

markmandel: Nope, that's it - you got it!

markmandel (approved):

PR #3211, opened 2023-06-12
Adds Counts and Lists AutoScale Policies

igooch:
**What type of PR is this?**

/kind feature

Adds autoscaling for Counters and Lists.

- Computes the desired number of replicas for the Fleet (scales up or down) based on either a CounterPolicy or ListPolicy with an Integer Buffer or a Percent Buffer.

Adds Priorities for sorting Counters and Lists to GameServerSet and Fleet CRDs.

- GameServerSet Priorities are populated from the Priorities in the Fleet CRD.
- Priority is for Game Server deletion on Fleet scale down.
- Priority is considered for both Distributed and Packed strategies. (Only implemented for Packed strategy in this PR. To be implemented for strategy Distributed in a future PR).
- Priority is only considered after sorting by Node in the Packed strategy.
- The Priority first sorts by Capacity and then by Count (or number of Values in a List) if Capacity is equal.
- If two game servers are on the same Node, the game server with the Counter or List denoted in the BufferPolicy will be marked for deletion first before a game server without the specified Counter or List.
- This scale down logic is held in pkg/gameserversets/gameserversets.go as it could not go in common.go due to cyclical dependencies.
- In order for the FleetAutoScaler to use this logic the FleetAutoScaler Controller pkg/fleetautoscalers/controller.go now has a PerNodeCounter and GameServerLister.

Working on #2716 

I moved the `type Priority struct` and its basic validation to common.go from gameserverallocation.go as this struct is used by Fleet, GameServerSet, and GameServerAllocation.


pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,425 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+	}
+	return 0, false, errors.Errorf("unable to apply ListPolicy %v", l)
+}
+
+// getSortedGameServers returns the list of Game Servers for the Fleet in the order in which the
+// Game Servers would be deleted.
+func getSortedGameServers(f *agonesv1.Fleet, gameServerLister listeragonesv1.GameServerLister,
+	nodeCounts map[string]gameservers.NodeCount) ([]*agonesv1.GameServer, error) {
+	// TODO: Should we handle this differently for strategy Distributed?
+	gss := f.GetGameServerSpec()
```
igooch, line 389: Should we handle this differently for strategy Distributed?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,425 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+	}
+
+	return replicas, limited, nil
+}
+
+// scaleDownByPercent
+// TODO: This death-spirals down to min capacity, so the customer would need to have something
+// in place to prevent gameservers from being evicted elsewhere if they have Count / Values on
+// them. I'm assuming this is the behavior we intend, and it just needs good documentation?
+func scaleDownByPercent(f *agonesv1.Fleet, gameServerLister listeragonesv1.GameServerLister,
```
igooch, line 602: // TODO: This death-spirals down to min capacity, so the customer would need to have something
// in place to prevent gameservers from being evicted elsewhere if they have Count / Values on
// them. I'm assuming this is the behavior we intend, and it just needs good documentation?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,425 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+		case false:
+			if list, ok := gs.Status.Lists[key]; ok {
+				aggCount -= int64(len(list.Values))
+				aggCapacity -= list.Capacity
+			}
+		}
+		replicas--
+		// Check if we've reached MinCapacity
+		if aggCapacity == minCapacity {
+			limited = true // TODO: Should we have this return true or false when scaling down to 0 (MinCapacity == 0)?
```
igooch, line 643: Should we have this return true or false when scaling down to 0 (MinCapacity == 0)?

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -429,7 +429,7 @@ func (c *CounterPolicy) ValidateCounterPolicy(causes []metav1.StatusCause) []met
 			causes = append(causes, metav1.StatusCause{
 				Type:    metav1.CauseTypeFieldValueInvalid,
 				Field:   "bufferSize",
-				Message: "bufferSize must be bigger than 0",
+				Message: "bufferSize must be bigger than 0", // TODO: Should we allow bufferSize to be Zero (in case of Scaling down to 0)? In case of BufferPercentage MaxCapacity can be set to 0, in case of BufferInt Max Capacity must be >= 2.
```
igooch, line 432: Should we allow bufferSize to be Zero (in case of Scaling down to 0)? In case of BufferPercentage MaxCapacity can be set to 0, in case of BufferInt Max Capacity must be >= 2.

pkg/gameserversets/controller.go
```diff
@@ -441,7 +441,7 @@ func computeReconciliationAction(strategy apis.SchedulingStrategy, list []*agone
 			handleGameServerUp(gs)
 		case agonesv1.GameServerStateReady:
 			handleGameServerUp(gs)
-		case agonesv1.GameServerStateReserved:
+		case agonesv1.GameServerStateReserved: // TODO: Isn't this already handled above in pass 1?
```
igooch, line 444: Isn't this already handled above in pass 1? (On line 409)

markmandel (commented):
Started on the review, but didn't get too far into the meat. Will come back around shortly.

install/helm/agones/templates/crds/fleet.yaml
```diff
@@ -110,6 +110,22 @@ spec:
...
                             - type: integer
                             - type: string
+                priorities: # which gameservers in the Fleet are most important to keep around - impacts scale down logic. Must be identical to priorities in gameserverset.yaml.
+                  type: array
+                  title: Configuration of Counters and Lists scale down logic
+                  nullable: true
+                  items:
+                    type: object
+                    properties:
+                      priorityType:
```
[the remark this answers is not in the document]
igooch: We decided to name it priorityType because `type` is a key word as in `type: object`.

I can't remember why we didn't do an enum, maybe for verbose failure? If it's better to do an enum I can switch those over. Would we still use string for `order` so that we can do defaulting? Or would that be better as an enum as well and require the user to specify the order?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,425 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+	var aggCapacity int64 // The Aggregate Capacity of the specified Counter or List of all GameServers across the GameServerSet in the Fleet
+	var minCapacity int64 // The Minimum Aggregate Capacity
+	var maxCapacity int64 // The Maximum Aggregate Capacity
+
+	var bufferSize intstr.IntOrString
+	// TODO: Our aggregate counts in include Allocated, Ready, and Reserved replicas, however
+	// I'm not 100% sure if f.Status.Replicas includes all three states or just Allocated and Ready?
+	// If it's the latter, then we would need to add in f.Status.ReservedReplicas.
+	// There may be some additional nuance here since "Reserved instances won't be deleted on scale down, but won't cause an autoscaler to scale up"
+	replicas := f.Status.Replicas
```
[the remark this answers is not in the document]
igooch: Hmm, should we update AggregateCounts and AggregateCapacity to just include Allocated + Ready? And maybe have a second field for reserved? Or three separate fields for AllocatedAggregateCount, ReadyAggregateCount, ReservedAggregateCount?

install/helm/agones/templates/crds/fleet.yaml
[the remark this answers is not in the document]
markmandel: So on `type` - I'm looking at this and it looks totally inconsistent to me, so I think we should change it. Sorry about the backtrack.

Example:
https://github.com/googleforgames/agones/blob/cef40425076e5e9a766952eb2f499d9b447b703a/install/helm/agones/templates/crds/fleetautoscaler.yaml#L59

I think it's better to move validation into the CRD spec as much as possible, as long term we'll want to not have webhooks as much as we can (move to CEL etc), as it's easier to scale and should be more performant as well.

Also we use an enum and default it within a CRD if it's blank in the Fleet autoscaler as well, so it should be fine (we can also make it `nullable` if it becomes tricky I think).

For example:

https://github.com/googleforgames/agones/blob/cef40425076e5e9a766952eb2f499d9b447b703a/install/helm/agones/templates/crds/fleetautoscaler.yaml#L62-L69

pkg/apis/autoscaling/v1/fleetautoscaler.go
markmandel: It's not something we can support in autoscaling. If it's zero and they make an allocation request, there is nowhere for it to go, since everything is zero.

This is why we have webhooks, because then the end user can use other signals - login services, matchmakers etc to let them know they need to scale up GameServers. As Agones, we don't get any of that information.

install/helm/agones/templates/crds/fleet.yaml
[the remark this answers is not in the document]
igooch: Does `priorityType` really need to be named `type`? I would argue that line 63 should have been named `policyType` rather than `type`. Yes, it's technically possible, but it makes the CRD confusing to read since we're overloading a field name with specific meaning in the OpenAPISchema rather than being precise (or less confusingly generic). Same thing with Go, which has `type` as a reserved word. We can (and did) use `Type` as a field name, but that feels about as off as using `Func` as a function name.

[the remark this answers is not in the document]
markmandel: So I'm not fussed about the Go struct side (in fact k8s also keeps a Go struct name that isn't `type`.

But `type` in the CRD is what K8s use, and what we use in a variety of places, so this is inconsistent with the rest of the API surface.

Here's the K8s example in code:

https://github.com/googleforgames/agones/blob/cef40425076e5e9a766952eb2f499d9b447b703a/vendor/k8s.io/api/apps/v1/types.go#L396-L408

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy

So yeah, I think it does to be consistent with both K8s and the rest of our CRD api surface.

pkg/fleetautoscalers/fleetautoscalers.go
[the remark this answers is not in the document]
gongmax: I'm leaning to the later to keep consistent with the replica fields, and be more flexible.

pkg/apis/agones/v1/common.go
```diff
@@ -212,3 +220,37 @@ func (ao *AllocationOverflow) Apply(gs *GameServer) {
 		}
 	}
 }
+
+// Priority is a sorting option for GameServers with Counters or Lists based on the Capacity.
+// PriorityType: Sort by a "Counter" or a "List".
+// Key: The name of the Counter or List. If not found on the GameServer, has no impact.
+// Order: Sort by "Ascending" or "Descending". Default is "Descending" so bigger Capacity is preferred.
```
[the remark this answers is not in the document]
igooch: Oh yep, I copy/pasted the comment and forgot to change it. Good catch!

pkg/gameserversets/controller.go
markmandel: Is a good question, but this a piece of code I'm not touching 😄

pkg/gameserversets/gameserversets.go
```diff
@@ -115,3 +151,78 @@ func ListGameServersByGameServerSetOwner(gameServerLister listerv1.GameServerLis
...
+			}
+		}
+		gs1ok = ok1
+		gs2ok = ok2
+	}
+	// If only one game server has the Priority, prefer that server. I.e. nil < gsX when Order is
+	// Descending (3, 2, 1, 0, nil), and nil > gsX when Order is Ascending (0, 1, 2, 3, nil).
+	// No Order specified "" is the same as Ascending.
+	if (gs1ok && p.Order == agonesv1.GameServerPriorityDescending) ||
+		(gs2ok && p.Order == "") ||
```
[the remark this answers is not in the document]
markmandel: I would agree with this - at this point, the default should be set, so we can assume a value.

pkg/gameserversets/gameserversets.go
```diff
@@ -75,12 +78,45 @@ func sortGameServersByLeastFullNodes(list []*agonesv1.GameServer, count map[stri
...
+		if acTotal > bcTotal {
+			return false
+		}
+
+		// if the Nodes have the same count and CountsAndLists flag is enabled, sort based on CountsAndLists Priorities.
+		if runtime.FeatureEnabled(runtime.FeatureCountsAndLists) && (priorities != nil) {
+			for _, priority := range priorities {
+				res := compareGameServerCapacity(&priority, a, b)
+				switch priority.Order {
+				case agonesv1.GameServerPriorityAscending, "":
```
markmandel, line 96: I would say at this point, we can assume there is a default, and if there isn't, that's a bug.

pkg/gameserversets/gameserversets.go
```diff
@@ -65,6 +67,7 @@ func sortGameServersByLeastFullNodes(list []*agonesv1.GameServer, count map[stri
 			return false
 		}
 
+		// TODO: What if they're not on the same node -- do we want to delete the gameserver of "equal" Nodes where one Node has creating gameservers and the other has Ready gameservers?
 		// if both are in the same node, make sure to delete pre-Ready GameServers first
 		if a.Status.NodeName == b.Status.NodeName {
```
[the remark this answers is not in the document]
markmandel: (Assuming packed), if they are not in the same node, then we delete as per packed node rules (so least full first) -- which is what we have now.

pkg/fleets/controller.go
```diff
@@ -364,6 +365,20 @@ func (c *Controller) upsertGameServerSet(ctx context.Context, fleet *agonesv1.Fl
 			"Scaling active GameServerSet %s from %d to %d", gsSetCopy.ObjectMeta.Name, active.Spec.Replicas, gsSetCopy.Spec.Replicas)
 	}
 
+	// Update GameServerSet Counts and Lists Priorities if not equal to the Priorities on the Fleet
```
markmandel, line 368: There's an interesting question in here which I hadn't considered.

If someone changes the priorities in a Fleet, what should happen?

I think the answer you've gone for is that it updates the active `GameServerSet`. I think that makes a lot of sense 👍🏻  - but I'm just writing it out to make sure we both agree.

pkg/fleetautoscalers/fleetautoscalers.go
markmandel: This is what we worked out on our doc. Allocated GameServers never go away, so their list/counts are counted towards the amount that is needed - so it shouldn't death spiral to min, it should do as expected (least I hope so - that exercise made my head hurt 🤯 )

pkg/fleetautoscalers/fleetautoscalers.go
markmandel: If I use what currently exists as a reference:

https://github.com/googleforgames/agones/blob/cef40425076e5e9a766952eb2f499d9b447b703a/pkg/fleetautoscalers/fleetautoscalers.go#L216-L223

I think that means it should return false to be consistent. Do you read it the same? (extra eyes good).

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,425 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+	if replicaCapacity == 0 {
+		return 0, false, errors.Errorf("cannot scale up as adding additional replicas does not increase Capacity")
+	}
+
+	additionalReplicas := int64(math.Ceil((float64(buffer) - float64(availableCapacity)) / float64(replicaCapacity)))
+
+	// Check if we've hit MaxCapacity
+	if (additionalReplicas*capacity)+aggCapacity > maxCapacity {
+		// TODO: This should do Floor division automatically, or should I switch the types to float64 then do Math.Floor to make it explicit?
+		additionalReplicas = (maxCapacity - aggCapacity) / capacity
```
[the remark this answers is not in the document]
markmandel: I will trust your testing 😄 No strong opinions from me on this one.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -30,6 +30,11 @@ import (
 
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	autoscalingv1 "agones.dev/agones/pkg/apis/autoscaling/v1"
+	listeragonesv1 "agones.dev/agones/pkg/client/listers/agones/v1"
+	"agones.dev/agones/pkg/fleets"
+	"agones.dev/agones/pkg/gameservers"
+	gsSets "agones.dev/agones/pkg/gameserversets"
```
markmandel, line 36: Can we use `gameserversets` or `gssets` since package names are all lower case, and otherwise I end up looking for a `gsSets` object 😄 

```suggestion
	"agones.dev/agones/pkg/gameserversets"
```

pkg/fleetautoscalers/fleetautoscalers.go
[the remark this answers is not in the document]
markmandel: To counter my own argument - when I did player tracking, I only looked at Ready/Reserved/Allocated

https://github.com/googleforgames/agones/blob/cef40425076e5e9a766952eb2f499d9b447b703a/pkg/gameserversets/controller.go#L630-L648

But I also never got to doing autoscaling based on player tracking (which for us is what the aggregates are most used for), so it's possible that would have changed over time, since you may well want to count the capacity that is incoming form a yet-to-be-Ready-GameServer, so you don't overshoot on resources and then have to scale it back (which is why I would recommend counting across all GameServers in the gsSet + Fleet) 🤔  I do wonder why I did that.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,425 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+// Game Servers would be deleted.
+func getSortedGameServers(f *agonesv1.Fleet, gameServerLister listeragonesv1.GameServerLister,
+	nodeCounts map[string]gameservers.NodeCount) ([]*agonesv1.GameServer, error) {
+	// TODO: Should we handle this differently for strategy Distributed?
+	gss := f.GetGameServerSpec()
+	gsList, err := fleets.ListGameServersByFleetOwner(gameServerLister, f)
+	if err != nil {
+		return nil, err
+	}
+	gameServers := gsSets.SortGameServersByStrategy(gss.Scheduling, gsList, nodeCounts, f.Spec.Priorities)
```
markmandel, line 394: Does the Fleet have the scheduling - do you need the GameServerSet here?

I think we chatted about this - should `SortGameServersByStrategy` move into some kind of common place? (Maybe we need a `/pkg/sorting`  or something like that?) Currently packages that are controllers don't tend to bleed into each other. I'm wondering if we want to maintain that seperation.

pkg/fleetautoscalers/fleetautoscalers.go
markmandel: Distributed sorting should be across all the nodes, rather than having counts/lists be the tie-breaker within a node, as it is for Packed -- so I think the answer is yes?

markmandel (commented):
I'll keep going through this once a day or so, because it's a big one, and might take some time 😄 

But good job tackling some really tricky problems!

pkg/fleetautoscalers/fleetautoscalers.go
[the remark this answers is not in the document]
markmandel: I'll write this another way as well -- I assume that when we get into autoscaling, there may be extra status values above and beyond what we had in https://github.com/googleforgames/agones/issues/2716#:~:text=1000%20item%20capacity-,Status,-status%3A%0A%20%23 that are useful to us. I figure we can also add them then, so we don't have to get this 100% right in this round.

install/helm/agones/templates/crds/fleet.yaml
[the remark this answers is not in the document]
igooch: Ah naming, I'll change it over to Type.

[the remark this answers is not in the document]
markmandel: 😆 I know, right?

pkg/apis/autoscaling/v1/fleetautoscaler.go
igooch: Does your comment "If it's zero and they make an allocation request, there is nowhere for it to go" refer to the buffer? Or the desired number of replicas? Just confirming because we can reach a desired number of replicas == 0.

pkg/fleetautoscalers/fleetautoscalers.go
igooch: I tried to move the SortGameServersByStrategy to common.go but could not because of cyclical dependencies. Is it best to put it in a separate package? Would we still need the gameServerLister and nodeCounter passed into the Fleetautoscaler, or would that be kept entirely separate?

pkg/fleetautoscalers/fleetautoscalers.go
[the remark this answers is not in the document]
igooch: It seems like if we are trying to abide by the "Reserved instances won't be deleted on scale down, but won't cause an autoscaler to scale up" then the aggregate values should not include reserved gameservers, as it stands this autoscaling logic includes reserved counts. If we're not abiding by that logic for Counts and Lists then we should probably make that clear? (And fwiw, we're technically tie-breaking on packed node sorting for any gameserver that is not in an "IsBeforeReady" state in sortGameServersByLeastFullNodes.)

pkg/fleetautoscalers/fleetautoscalers.go
igooch: It's not quite the same -- the bufferPolicy is able to calculate the desiredReplicas and then determine if it's limited or not. 

We're checking after removing each unique gameserver one by one if we've reached the limit. If we've reached our desired state on line 624 then we exit before checking if it's limited. So in our case we would almost always hit aggCapacity == minCapacity if we have not reached the desired capacity, yet we've hit our minimum. I think the only case where this line might return "false" is when remove every single game server (so we don't go back up into the loop to check desiredCapacity again), but we've hit minCapacity because we've removed all the gameservers.

Alternatively I could add an emtpy gameserver at the end of the gameserver list to get it to loop one last time (after all gameservers have been removed). Would that be a reasonable thing to do?

pkg/fleetautoscalers/fleetautoscalers.go
igooch: It will death spiral to minCapacity -- all of the bufferPolicy, etc. calculate the "desired replicas" and it's not until the computeDesiredFleetSize returns back to the controller that the controller determines how many replicas can actually be deleted.

pkg/fleetautoscalers/fleetautoscalers.go
igooch: I guess I should have asked a different question: This function calls SortGameServersByStrategy which sorts by either Packed or Distributed strategies. In the case of Packed I am fairly confident that the list of gameservers in order of deletion will be the same as when the SortGameServersByStrategy is later called if there are actually gameservers that can be deleted. I'm not confident that we'll get back the same list of game servers when the controller later calls SortGameServersByStrategy when the stragegy is Distributed. Which would mean that the "desiredReplicas" we calculate here would be a different number than if we'd calculated it based on a different ordered list of gameservers.

The general issue is that the FleetAutoscaler Controller is expected "desired replicas" and calls SortGameServersByStrategy later. This isn't the ideal flow for Counts and Lists -- in an ideal scenario we would have the list of game servers in order of deletion first. I'm not sure we want to muck about with the controller that much though?

pkg/fleets/controller.go
igooch: Yes, agreed that updates to the Fleet Priorities should update the GameServerSet Priorities.

pkg/gameserversets/controller.go
igooch: Fine by me

pkg/gameserversets/gameserversets.go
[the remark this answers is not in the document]
igooch: I'm wondering about states like:

a.Status.NodeName != b.Status.NodeName
acTotal == bcTotal && a.IsBeforeReady()?

Also wondering about:
a.IsBeforeReady() && b.Status.State == agonesv1.GameServerStateAllocated

pkg/fleetautoscalers/fleetautoscalers.go
igooch: Ah, didn't realize they were all lowercase. Made the change.

pkg/fleetautoscalers/fleetautoscalers.go
markmandel: I would apply the same logic here, as we do elsewhere: if the _desired_ capacity is less than the current aggregate allocated capacity - we are limited. 

So I think we should be looking at `desiredCapacity` here, not aggCapacity (Aggregate capacity?)

pkg/gameserverallocations/allocation_cache.go
```diff
@@ -237,14 +237,14 @@ func (c *AllocationCache) ListSortedGameServers(gsa *allocationv1.GameServerAllo
 			for _, priority := range gsa.Spec.Priorities {
 				res := compareGameServers(&priority, gs1, gs2)
 				switch priority.Order {
-				case allocationv1.GameServerAllocationAscending:
+				case agonesv1.GameServerPriorityAscending:
```
markmandel, line 240: Suggestion: Pull the renaming out as a separate PR, just to remove some changes from this one.

pkg/gameserversets/gameserversets.go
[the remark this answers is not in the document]
markmandel: > a.Status.NodeName != b.Status.NodeName

That's handled by the node count sorting below.

> acTotal == bcTotal && a.IsBeforeReady()?

I don't think this happens, because state checks happen before node count checks.

> a.IsBeforeReady() && b.Status.State == agonesv1.GameServerStateAllocated

Not a concern for this function. Allocated state gets filtered later on.

pkg/fleetautoscalers/fleetautoscalers.go
markmandel: > I tried to move the SortGameServersByStrategy to common.go but could not because of cyclical dependencies.

Yeah, that makes sense.

> Is it best to put it in a separate package?

That's sorta what I'm thinking - move both `SortGameServersByStrategy` and probably the `PerNodeCounter` into there (since they are so tied together), and we can have a central place for sorting strategies. Otherwise we seem to be arbitrarily importing from `pkg/gameservers` and `pkg/gameserversets` which feels a little messy to me.

(This refactor could be a separate PR against the current code base if you wanted to split the work out, and carve some chunks off this one).

markmandel (commented):
Hope this review helps answer some questions 🤯

pkg/fleetautoscalers/controller_test.go
```diff
@@ -300,6 +302,12 @@ func TestControllerSyncFleetAutoscaler(t *testing.T) {
 			Time: time.Now(),
 		}
 
+		fakeWatch := watch.NewFake()
```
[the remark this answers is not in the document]
igooch: The build was failing with `race detected during execution of test` on TestControllerSyncFleetAutoscaler. I think adding the node counter and game server lister to the fleetautoscaler controller tipped this test into failing, but it was also complaining about not having Fleet and FleetAutoScaler watches, so I added in fake watches for those too.

--- FAIL: TestControllerSyncFleetAutoscaler (0.00s)
{"message":"agones.dev/agones/pkg/client/informers/externalversions/factory.go:119: Failed to watch *v1.Fleet: unhandled watch: ...
{"message":"agones.dev/agones/pkg/client/informers/externalversions/factory.go:119: Failed to watch *v1.GameServer: unhandled watch: ...
{"message":"agones.dev/agones/pkg/client/informers/externalversions/factory.go:119: Failed to watch *v1.FleetAutoscaler: unhandled watch: ...
{"message":"k8s.io/client-go/informers/factory.go:134: Failed to watch *v1.Node: unhandled watch: ...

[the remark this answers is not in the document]
markmandel: I'm _fairly_ sure those are just warnings, and if there isn't a watcher, it doesn't actually matter. (I feel like I've seen those warnings a lot over the years and just ignored them.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,425 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+// Game Servers would be deleted.
+func getSortedGameServers(f *agonesv1.Fleet, gameServerLister listeragonesv1.GameServerLister,
+	nodeCounts map[string]gameservers.NodeCount) ([]*agonesv1.GameServer, error) {
+	// TODO: Should we handle this differently for strategy Distributed?
+	gss := f.GetGameServerSpec()
+	gsList, err := fleets.ListGameServersByFleetOwner(gameServerLister, f)
+	if err != nil {
+		return nil, err
+	}
+	gameServers := gssets.SortGameServersByStrategy(gss.Scheduling, gsList, nodeCounts, f.Spec.Priorities)
```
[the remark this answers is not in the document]
igooch: I think it would be f.Spec.Template.Spec.Scheduling?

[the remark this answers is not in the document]
markmandel: Checking https://agones.dev/site/docs/reference/agones_crd_api_reference/#agones.dev/v1.Fleet

I'm reasonably sure it's Fleet.Spec.Scheduling?

https://github.com/googleforgames/agones/blob/88c38faa01921adaf9bf0a043b6e8c79ee015a28/pkg/apis/agones/v1/fleet.go#L60-L75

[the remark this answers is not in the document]
igooch: Oh yep, you are correct. The gsSet.Spec is the made from the f.Spec.Scheduling in fleet.go, so those should be the same thing.

```
func (f *Fleet) GameServerSet() *GameServerSet {
	gsSet := &GameServerSet{
		ObjectMeta: *f.Spec.Template.ObjectMeta.DeepCopy(),
		Spec: GameServerSetSpec{
			Template:   f.Spec.Template,
			Scheduling: f.Spec.Scheduling,
		},
	}
  ```

File: pkg/apis/autoscaling/v1/fleetautoscaler.go (hunk 4 above)
Comment (igooch) [reply pkg/apis/autoscaling/v1/fleetautoscaler.go:432]: Per discussion we do not allow autoscaling to Zero (unless by custom Webhook).

File: pkg/fleetautoscalers/fleetautoscalers.go (hunk 15 above)
Comment (igooch) [reply pkg/fleetautoscalers/fleetautoscalers.go]: OK, I will make changes to sorting by Priorities for Strategy distributed.

File: pkg/fleetautoscalers/fleetautoscalers.go (hunk 7 above)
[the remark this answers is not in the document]
Comment (igooch) [reply pkg/fleetautoscalers/fleetautoscalers.go:260]: Per conversation:

We change from current calculation of AggregateCount and Capacity for Ready, Reserved, or Allocated only to AggregateCounts and AggregateCapacity is across all statuses (like Replicas). This will not include game servers that are being shut down (as is the current state for status.Replicas).

We add an AllocatedAggregateCounts and AllocatedAggregateCapacity that we use to calculate the appropriate buffer at the buffer should be determined based on allocated available capacity.

File: pkg/gameserverallocations/allocation_cache.go (hunk 16 above)
Comment (igooch) [reply pkg/gameserverallocations/allocation_cache.go:240]: I'm going to leave this in since I changes the location of the code to agonesv1 this variable would have changed regardless.

File: pkg/fleetautoscalers/controller_test.go (hunk 17 above)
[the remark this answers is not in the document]
Comment (markmandel) [reply pkg/fleetautoscalers/controller_test.go:298]: Just noting that now that #3260 is merged, you can remove all these fake watchers 😃

File: pkg/fleetautoscalers/fleetautoscalers.go (hunk 2 above)
Comment (igooch) [reply pkg/fleetautoscalers/fleetautoscalers.go:599]: Resolved with refactor.

File: pkg/fleetautoscalers/fleetautoscalers.go (hunk 1 above)
Comment (igooch) [reply pkg/fleetautoscalers/fleetautoscalers.go]: I'll add this in a separate PR just for the sake of keeping this PR from turning into a XXXL.

**Review by markmandel — COMMENTED** [review]
This looks good to me! @gongmax since this is such a big PR, can you also a do a review pass as well before we merge this? Your extra eyes would be appreciated!

File: pkg/fleetautoscalers/fleetautoscalers.go [hunk 19]
```diff
@@ -224,3 +234,298 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+	return replicas + additionalReplicas, limited, nil
+}
+
+// scaleDown scales down for either Integer or Percentage Buffer.
+func scaleDown(f *agonesv1.Fleet, gameServerLister listeragonesv1.GameServerLister,
+	nodeCounts map[string]gameservers.NodeCount, key string, isCounter bool, replicas int32,
+	availableCapacity, aggCount, aggCapacity, minCapacity, buffer int64) (int32, bool, error) {
+	// Exit early if we're already at MinCapacity or DesiredCapacity to avoid calling
+	// getSortedGameServers if unnecessary
+	if aggCapacity == minCapacity {
```
gongmax, line 472: Is this necessary, since the validation is done at line 303 and 343

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,298 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+		return replicas, true, nil
+	}
+
+	if availableCapacity == buffer {
```
gongmax, line 476: Is this necessary, since the validation is done at line 342

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -224,3 +234,298 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
+		// Check if we're Limited (Below MinCapacity)
+		if aggCapacity < minCapacity {
+			return replicas + 1, true, nil
+		}
+		// Check if we're at our desired Buffer
+		if availableCapacity == buffer {
+			return replicas, false, nil
+		}
+		// Check if we're at Limited
+		if aggCapacity == minCapacity {
```
gongmax, line 517: I think this check should happen before line 513, to indicate it's scaleLimited in the case where `aggCapacity == minCapacity` and `availableCapacity == buffer` are true at the same time.

pkg/gameserversets/gameserversets.go
```diff
@@ -75,12 +76,45 @@ func sortGameServersByLeastFullNodes(list []*agonesv1.GameServer, count map[stri
...
+				}
+			}
+		}
+
+		// if the gameservers are on the same node, then sort lexigraphically for a stable sort
+		if a.Status.NodeName == b.Status.NodeName {
+			return a.GetObjectMeta().GetName() < b.GetObjectMeta().GetName()
+		}
+		// if both Nodes have the same count, one node is emptied first (packed scheduling behavior)
+		return a.Status.NodeName < b.Status.NodeName
```
gongmax, line 117: Should we just rely on `GetObjectMeta().GetName()` and skip the checking of `Status.NodeName`? I.e. simplify this to:
~~~
return a.GetObjectMeta().GetName() < b.GetObjectMeta().GetName()
~~~

gongmax (approved):
Some minor comments. But overall LGTM.

pkg/gameserversets/gameserversets.go
markmandel: No because that wouldn't give us a per-node deterministic ordering.

If two nodes have the same count, we need to always pick one over the other in a deterministic manner. This fixes a bug where concurrent processes could bounce between scaling down nodes of similar counts rather than just picking one and sticking with it.

gongmax: Gotcha, make sense.

pkg/fleetautoscalers/fleetautoscalers.go
igooch: I interpret "limited" to mean that we could not reach our desired buffer because we hit min or max capacity, which is why I first check for the buffer. A different interpretation would be that "limited" means that we are at min or max capacity (regardless of the desired capacity).

Based on the comment `// scalingLimited indicates that the calculated scale would be above or below the range defined by MinReplicas and MaxReplicas` in /pkg/fleetautoscalers/controller.go#L395 I'm leaning towards the first interpretation.

gongmax: Make sense!

pkg/fleetautoscalers/fleetautoscalers.go
igooch: The validation at 303 and 343 that checks if this is below minCapacity, it doesn't check if we're at min Capacity.

pkg/fleetautoscalers/fleetautoscalers.go
igooch: Good point, removing this.

markmandel (approved):

roberthbailey:
🎉

PR #3234, opened 2023-06-26
GKE Autopilot: Add primary container annotation for game server container

zmerlynn:
`autopilot.gke.io/primary-container` is a new annotation that will be supported by Autopilot ~soon that indicates which container resource defaulting should affect. Prior to this if I were to create a GameServer with no resource requests, Autopilot would default-up the sidecar container, as it occurs first.

The annotation will be supported in 1.27 ~soon, but this change is safe on all versions (just not useful, necessarily).

### DIFF
--- a/pkg/apis/agones/v1/apihooks.go
+++ b/pkg/apis/agones/v1/apihooks.go
@@ -28,7 +28,7 @@ type APIHooks interface {
 	// ValidateScheduling is called by Fleet and GameServerSet Validate() to allow for product specific validation of scheduling strategy.
 	ValidateScheduling(apis.SchedulingStrategy) []metav1.StatusCause
 
-	// MutateGameServerPodc is called by createGameServerPod to allow for product specific pod mutation.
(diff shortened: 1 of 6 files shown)


pkg/apis/agones/v1/apihooks.go
```diff
@@ -28,8 +28,8 @@ type APIHooks interface {
 	// ValidateScheduling is called by Fleet and GameServerSet Validate() to allow for product specific validation of scheduling strategy.
 	ValidateScheduling(apis.SchedulingStrategy) []metav1.StatusCause
 
-	// MutateGameServerPodSpec is called by createGameServerPod to allow for product specific pod mutation.
-	MutateGameServerPodSpec(*GameServerSpec, *corev1.PodSpec) error
+	// MutateGameServerPodc is called by createGameServerPod to allow for product specific pod mutation.
```
gongmax, line 31: typo

zmerlynn: fixed, oops

gongmax (approved):

roberthbailey (approved):

Issue #3096, opened 2023-04-13
Graduate `StateAllocationFilter` to Stable


roberthbailey:
I propose that we graduate the `StateAllocationFilter` feature to stable. It has been in beta since 1.26.0. 

Context:
https://development.agones.dev/site/docs/guides/feature-stages/#feature-gates

markmandel:
I concur!

markmandel:
Just noting that this was done in #3308

PR #3238, opened 2023-06-28
move StateAllocationFilter to stable

ashutosji:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix

Closes #https://github.com/googleforgames/agones/issues/3096


markmandel:
I see a couple of failing tests - `TestFindGameServerForAllocationPacked` and `TestFindGameServerForAllocationPacked/one_label` in the unit tests. 

Are they also failing for you locally?

pkg/gameserverallocations/allocation_cache.go
```diff
@@ -67,9 +67,8 @@ func NewAllocationCache(informer informerv1.GameServerInformer, counter *gameser
 	}
 
 	// if we can do state filtering, then cache both Ready and Allocated GameServers
-	if runtime.FeatureEnabled(runtime.FeatureStateAllocationFilter) {
 		c.matcher = readyOrAllocatedGameServerMatcher
-	}
+
```
markmandel, line 71: We could clean up up more, and replace the matcher above on line 66 directly with the readyOrAllocatedGameServerMatcher -- and we could delete the `readyGameServerMatcher` function entirely.

(There's possibly a good reason to get rid of `matcher` as a data structure entirely, since we don't need to switch it out at runtime anymore).

pkg/gameserverallocations/allocation_cache_test.go
```diff
@@ -545,13 +531,12 @@ func TestAllocationCacheCompareGameServers(t *testing.T) {
 	}
 }
 
-func TestAllocatorRunCacheSyncFeatureStateAllocationFilter(t *testing.T) {
+func TestAllocatorRunCacheSync(t *testing.T) {
 	t.Parallel()
 
 	// TODO(markmandel): When this feature gets promoted to stable, replace test TestAllocatorRunCacheSync below with this test.
 	runtime.FeatureTestMutex.Lock()
 	defer runtime.FeatureTestMutex.Unlock()
```
markmandel, line 539:
```suggestion

```

Since we have no feature flags set, we don't need the mutex anymore.

We do this because Go tests run in parallel, but feature flags are stored globally, so this allows us to have tests that don't collide. No feature flags here anymore means we don't need the mutex anymore!

pkg/gameserverallocations/find_test.go
```diff
@@ -94,7 +94,7 @@ func TestFindGameServerForAllocationPacked(t *testing.T) {
 				{ObjectMeta: metav1.ObjectMeta{Name: "gs6", Namespace: "does-not-apply", Labels: oneLabel}, Status: agonesv1.GameServerStatus{NodeName: "node1", State: agonesv1.GameServerStateReady}},
 			},
 			test: func(t *testing.T, list []*agonesv1.GameServer) {
-				assert.Len(t, list, 3)
+				assert.Len(t, list, 5)
```
markmandel, line 97: Just checking why the change here?

site/content/en/docs/Integration Patterns/high-density-gameservers.md
```diff
@@ -6,8 +6,7 @@ weight: 70
 description: >
   How to run multiple concurrent game sessions in a single GameServer process.
 ---
-{{< beta title="Allocation State Filter" gate="StateAllocationFilter" >}}
-
+{{% feature publishVersion="1.32.0" %}}
```
markmandel, line 9: What we would want to do here instead is use `feature expiryVersion` to wrap the `beta` shortcode - since we want to remove it up until the next release.

The way it is currently would mean that as soon as we pushed this live, the whole block of docs would be removed.

site/content/en/docs/Integration Patterns/player-capacity.md
```diff
@@ -10,7 +10,6 @@ description: >
 {{< alpha
 title="Player Tracking and Allocation Player Filter"
 gate="PlayerTracking,PlayerAllocationFilter" >}}
-{{< beta title="Allocation State Filter" gate="StateAllocationFilter" >}}
```
markmandel, line 13: Same as above - need to wrap this in a feature shortcode to expire it on the next version, so it doesn't get removed immediately (docs get published immediately on merge to `main`).

site/content/en/docs/Reference/gameserverallocation.md
```diff
@@ -36,8 +36,6 @@ spec:
         game: my-game
       matchExpressions:
         - {key: tier, operator: In, values: [cache]}
-      # [Stage:Beta]
-      # [FeatureFlag:StateAllocationFilter]
```
markmandel, line 40: All this will need to be feature code wrapped.

markmandel (commented):
Good start! Gave you some direction on how to tidy it up.

Haven't looked into why the tests are failing, but let us know what you find and we'll narrow it down if you are struggling with it.

pkg/gameserverallocations/allocation_cache.go
ashutosji: > 

Yeah, I forgot it. Thanks for reminding i will correct that.

pkg/gameserverallocations/find_test.go
ashutosji: The reason of changing 3 to 5 was the actual and expected outcome was differing. Correct me if i am wrong in assert.Len() function we are passing length as gameserver with state in GameServerStateReady?

ashutosji:
> I see a couple of failing tests - `TestFindGameServerForAllocationPacked` and `TestFindGameServerForAllocationPacked/one_label` in the unit tests.
> 
> Are they also failing for you locally? Are you able to reproduce the issues?



> I see a couple of failing tests - `TestFindGameServerForAllocationPacked` and `TestFindGameServerForAllocationPacked/one_label` in the unit tests.
> 
> Are they also failing for you locally? Are you able to reproduce the issues?

Yeah, It's still passing!
Screenshot 2023-07-03 at 3 16 52 PM

markmandel:
Just touching base to see if you need a hand here. If not no worries, but figured I would check.

Kalaiselvi84:
Closing this PR now, and will handle this issue in a different PR - https://github.com/googleforgames/agones/pull/3308

PR #3241, opened 2023-06-29
fix: Label PR Warning

aimuz:

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Closes #

```
Warning: Unexpected input(s) 'repo-token', valid inputs are ['script', 'github-token', 'debug', 'user-agent', 'previews', 'result-encoding']
```


markmandel:
@Kalaiselvi84 you are the expert on this one - this look good to you?

Kalaiselvi84:
> @Kalaiselvi84 you are the expert on this one - this look good to you?

I have tested it in my fork without including `repo-token: ${{ secrets.GITHUB_TOKEN }}`, and it is working fine. According to the documentation at https://docs.github.com/en/actions/security-guides/automatic-token-authentication, the secrets.GITHUB_TOKEN is automatically available to all workflows running in the repository, and there is no need to explicitly define or store it in the repository's secrets. However, it is important to consider the security implications, as using this token could potentially grant unauthorized access to the repository. Nevertheless, since it is an open-source project, it should be acceptable to omit its usage. WDYT?

aimuz:
In https://github.com/actions/github-script, the repo-token is not valid

https://github.com/actions/github-script#using-a-separate-github-token

markmandel (approved):
Sounds like this is a valid fix then.

PR #3231, opened 2023-06-22
Move SDKGracefulTermination To Stable

Kalaiselvi84:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

**What this PR does / Why we need it**:

**Which issue(s) this PR fixes**:

Closes #3094

**Special notes for your reviewer**:

PR #3252, opened 2023-07-08
More Local Dev Server Support

CauhxMilloy:
**What type of PR is this?**
/kind feature
(but also documentation)

This PR adds better support for integration between Dev Servers (called [Local Game Server](https://agones.dev/site/docs/guides/local-game-server/) in guides) and [Local Development](https://agones.dev/site/docs/guides/client-sdks/local/).
Note that these changes do not affect prod flows, only developer flows when running manually with extra (new) command line arguments.

* Updating SDK Server (`sdk-server/main.go`) for "out of cluster" mode.
  * Adding `--kubeconfig` cli flag to pass in a suitable k8s config file.
    * Makes use of `clientcmd.BuildConfigFromFlags()` to connect to the k8s cluster.
    * If the flag is not provided (as is the case for typical prod use), the `BuildConfigFromFlags()` call will internally fall back to the `InClusterConfig()`.
    * This is similar to the `--kubeconfig` cli arg for `cmd/controller/main.go`.
  * Adding `--no-graceful-termination` cli flag.
    * If set (default unset), the context `ctx` object will not be wrapped by the `NewSDKServerContext()` call. This means that exiting the program will not wait (block) for the `GameServer` state to update to `Shutdown`.
    * This flag to disable graceful shutdown is added as a convenient dev flow option, to quickly terminate and restart the SDK Server when testing locally.
    * Graceful termination was first added as an optional feature in https://github.com/googleforgames/agones/pull/2205.
      * Fully rolled into stable (no longer optional feature) in https://github.com/googleforgames/agones/pull/3231.
    * This flag does not affect normal (prod) flows (graceful termination still enabled by default).
* Updating gameservers/controller.go to handle `RequestReady` to `Ready` transition (for dev server).
  * `RequestReady` is set via SDK Server (from SDK calls). Need to transition to `Ready` to be allocatable.
* Updated "Local Development" guide page (`docs/Guides/Client SDKs/local.md`) with information with running "out of cluster" mode.
  * Also added information about running from source code instead of prebuilt binaries -- which many devs would prefer.
* Added links between "Local Development" guide page (`docs/Guides/Client SDKs/local.md`) and "Local Game Server" guide page (`docs/Guides/local-game-server.md`) about how they can be used together.

N/A? I didn't see any issues that exactly related to what this solves.

Agones is cool. 😎 Thanks for the FOSS.


CauhxMilloy:
Cleaned up to try and pass CI as much as I could. `run-sdk-conformance-test-rust` is currently failing. Seems like its deep in some unfamiliar make files and I don't have access to the logs at https://console.cloud.google.com/cloud-build/builds/b2cb1adf-17fb-42c6-b0b3-b19b7ed0c7d1?project=258182270954. I guess I'll leave as is for now. Sorry if this is something silly in the PR. Let me know what I can do to address it.

Look forward to the review. Thanks!

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -13,25 +13,41 @@ namespace as it - usually referred to in Kubernetes terms as a "sidecar".
...
 
-macOS
-* sdk-server.darwin.amd64
-* sdk-server.darwin.arm64
+### Running from source code
 
-Linux
-* sdk-server.linux.amd64
-* sdk-server.linux.arm64
+Instead of downloading and running executable binaries from the internet, you can simply run from source code.
```
markmandel, line 37: 🤔 not sure we want to be advocating people compile from source. I mean you _can_, not sure if you _should_ ? It feels like it complicates things quite a bit, and there is no real reason to  - we provide binaries (and docker images too)?

Unless there is a strong reason to that I'm missing? (I can't think of another project that does this).

Removing the documentation on compilation would also simplify everything considerably.

cmd/sdk-server/main.go
```diff
@@ -53,15 +54,17 @@ const (
...
-	delayFlag       = "delay"
-	timeoutFlag     = "timeout"
-	grpcPortFlag    = "grpc-port"
-	httpPortFlag    = "http-port"
+	localFlag                 = "local"
+	fileFlag                  = "file"
+	testFlag                  = "test"
+	testSdkNameFlag           = "sdk-name"
+	kubeconfigFlag            = "kubeconfig"
+	noGracefulTerminationFlag = "no-graceful-termination"
```
markmandel, line 62: 🤔 do we need this? if it's a fully local experience, is it better to not shutdown until the GameServer moves to `Shutdown` state? (it can also be ctrl+c'd I think?)

pkg/gameservers/controller.go
```diff
@@ -543,23 +543,29 @@ func (c *Controller) syncDevelopmentGameServer(ctx context.Context, gs *agonesv1
 		return gs, nil
 	}
 
-	// Only move from Creating -> Ready. Other manual state changes are up to the end user.
-	// We also don't want to move from Allocated -> Ready every time someone allocates a GameServer.
-	if gs.Status.State != agonesv1.GameServerStateCreating {
+	// Only move from Creating -> Ready or RequestReady -> Ready.
```
markmandel, line 546: Can you explain the flow here?

Messing with GameServer state worries me a little 😄 , and I'm not entirely sure how this ends up redirecting traffic to the local gameserver + sdk instance? (or maybe it doesn't?)

I'm wondering if there's a way we could hijack the existing `agones.dev/dev-address` annotation to keep this as a special case flow? 🤔 

https://agones.dev/site/docs/guides/local-game-server/

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -149,3 +172,52 @@ wget https://raw.githubusercontent.com/googleforgames/agones/{{< release-branch
 chmod o+r gameserver.yaml
 docker run --network=host --rm -v $(pwd)/gameserver.yaml:/tmp/gameserver.yaml us-docker.pkg.dev/agones-images/release/agones-sdk:{{<release-version>}} --local -f /tmp/gameserver.yaml
 ```
+
+## Running locally using "out of cluster" mode
```
markmandel, line 176: Just a heads up, our docs publish on merge, so if you have new features that aren't in the current release, please wrap them in a `feature` shortcode like so: https://agones.dev/site/docs/contribute/

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -149,3 +172,52 @@ wget https://raw.githubusercontent.com/googleforgames/agones/{{< release-branch
...
+
+An alternative to running completely isolated from a cluster is to run in "out of cluster" mode.
+This allows you to run locally but interact with the controllers within a cluster.
+This workflow works well when running a [Local Game Server](https://agones.dev/site/docs/guides/local-game-server/) in your cluster and want to run the server locally.
```
markmandel, line 180: Rather than use links to the website, please either the `ref` or `relref` shortcodes.

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -149,3 +172,52 @@ wget https://raw.githubusercontent.com/googleforgames/agones/{{< release-branch
...
+This means being able to allocate your game in its normal flow (much more prod-like environment) and be able to debug (e.g. breakpoint) your server code.
+This can also be done with [running Minikube locally](https://agones.dev/site/docs/installation/creating-cluster/minikube/), which is great for early prototyping.
+
+The name "out of cluster" mode is to contrast [InClusterConfig](https://pkg.go.dev/k8s.io/client-go/tools/clientcmd#InClusterConfig) which is used in the internal golang kubeconfig API.
+
+To run in "out of cluster" mode, instead of passing `--local` or `-f ./gameserver.yaml`, you'd use `--kubeconfig` to connect into your cluster.
+However, there are a number of commands that are necessary/useful to run when running in "out of cluster" mode.
+Here is a sample with each piece discussed after.
+```bash
+GAMESERVER_NAME='my-local-server' \
```
markmandel, line 190: I feel like to do this cleanly, we should update the sdkserver to also allow command line args for the GameServer name and Pod Namespace -- just to be consistent, and make life easier for end users. Then we don't have to mix args and env vars.

Also would prefer to see these docs using a binary, not running from source.

site/content/en/docs/Guides/local-game-server.md
```diff
@@ -9,6 +9,8 @@ description: >
 
 You can register a local game server with Agones. This means you can run an experimental build of your game server in the Agones environment without the need of packaging and deploying it to a fleet. This allows you to quickly iterate on your game server code while still being able to plugin to your Agones environment.
 
+This can be used in combination with a [local SDK Server]({{< ref "./Client SDKs/local.md" >}}) in ["out of cluster" mode]({{< ref "./Client SDKs/local.md#running-locally-using-out-of-cluster-mode" >}}).
```
markmandel, line 12: So when reading the "out of cluster mode" - I automatically assumed I could connect to my local game server binary.... but reading this, it doesn't seem to be the case.

I'm wondering if there's a way we can combine it into one whole process? 🤔 not entirely sure how, but it feels like it's worth exploring. WDYT?

markmandel (commented):
Really interesting idea 👍🏻 

Dropped some comments on cleanup and tidying and some approach things, but let me know what you think of my questions, and also let me know if I misunderstood anything in there.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: Compiling from source is a very common occurrence for handling dependencies. For example, Bazel makes use of pulling source tar balls from dependency libs. Gentoo replaces its entire package manager with downloading and compiling source code. Many code bases also offer this as an alternative to downloading prebuilt binaries.

Downloading unknown binaries, especially with no SHA256, is decently frowned upon in many places. I definitely was apprehensive about "here, run this exe". FWIW, interpreted code (straight from source) or docker images don't really have this issue because you see exactly what is being run.

Honestly, building from source is really helpful -- especially with interpreted languages (or languages that "feel interpreted", https://go.dev/doc/). It's really easy to test, debug, iterate, etc. I understand that for most binaries in Agones, this doesn't really make sense -- one can't run from source for a controller in a cluster. But for the SDK Server, when run locally, this actually makes a lot of sense.

cmd/sdk-server/main.go
CauhxMilloy: >  if it's a fully local experience, is it better to not shutdown until the GameServer moves to Shutdown state?

If you're developing locally (including running your server in a debugger, setting breakpoints, exiting, re-running, hitting exceptions, crashing, etc) then you probably won't be completing the normal server lifetime. It would be really clunky to _require_ a full server lifetime (progressing through all states) in order to do the most basic debugging -- especially in the context of quickly iterating and you want to run your game server to just try something real quick.

> (it can also be ctrl+c'd I think?)

Unfortunately, no. The open connections checking server state will persist the binary. Spamming `^C` will not work, requiring a `kill -9` to get it to exit.

Preventing the process from exiting was the point of "graceful termination". See https://github.com/googleforgames/agones/pull/2205. As written in that PR, it prevents termination unless `SIGKILL` is received.

In case you want to try this yourself: do note that running the SDK Server then pressing `^C` will properly exit. However, after any state updates, like a call to `agonesSdk.ReadyAsync()` will then get state monitoring to keep the processes alive. However, passing `--no-graceful-termination` will always allow for exiting with `^C` at any point.

pkg/gameservers/controller.go
CauhxMilloy: > Can you explain the flow here?

Note that this function only applies for dev game servers. Note this snippet called early in `syncDevelopmentGameServer` (this function):
```golang
if !isDevGs {
    return gs, nil
}
```
So this function ONLY applies to dev game servers.

Previously, this function ONLY progressed the dev game server from `Creating` to `Ready`. This change adds in the progression of `RequestReady` to `Ready`.

Why this is important: When working with the SDK, a call to `agonesSdk.ReadyAsync()` will put the `GameServer` into the `RequestReady` state. The `GameServer` must be in the `Ready` state for the allocation service (or alternative) to allocate the `GameServer` when requested. Progression from `RequestReady` to `Ready` is always done by the controller (this file), except for dev game servers (see `syncGameServerRequestReadyState()`). This change simply makes it always the case that `RequestReady` to `Ready` will properly be progressed.

Yes, it is true that a manual REST call could put the `GameServer` into `Ready`. Something like:
```
curl -X PUT -H "Content-Type: application/json" --data "$(curl -s http://localhost:8082/apis/agones.dev/v1/namespaces/default/gameservers/my-local-server | jq '.status.state="Ready"')" http://localhost:8082/apis/agones.dev/v1/namespaces/default/gameservers/my-local-server
```
But doing this as a separate step when developing is **_EXTREMELEY TEDIOUS_**. This also isn't what happens normally. If the normal (prod) flow takes care of this, then the controller should still handle it for dev servers as well.

> I'm wondering if there's a way we could hijack the existing `agones.dev/dev-address` annotation to keep this as a special case flow?

I do actually have thoughts about using annotations, similar to `agones.dev/dev-address`, for handling how state progression should work. I actually think that special cases should be minimized. IMHO, dev workflows should accomplish 2 things: be easy to do and representative of prod. Making controller-originating `GameServer` progression as similar between dev workflows and prod is important to make sure that realistic behavior is encountered during development/testing.

This all said, this change is very small -- simply adding in proper progression from `RequestReady` to `Ready` for dev servers. The changes that I have in mind that would try to bring more consistency and making use of annotations; that I'm saving for another PR (to keep this one more straight forward).

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: Good call. Sorry. Will update.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: > Then we don't have to mix args and env vars.

I agree that this would be nicer. I can add in the extra flags.

> Also would prefer to see these docs using a binary, not running from source.

As mentioned in the other comment, I think running from source is nicer than downloaded binaries. But sure, for the sake of documentation consistency, I can update this.

site/content/en/docs/Guides/local-game-server.md
CauhxMilloy: >  I automatically assumed I could connect to my local game server binary

That is the case. This is what I do for local testing. I have my custom game server running in Visual Studio, I run the SDK Server on a terminal, and I have it connect to my cluster. This allows me to allocate the game server via normal prod flows (e.g. external calls onto the allocator service) which are then fully integrated with my breakpoint-able game server in VS.

> I'm wondering if there's a way we can combine it into one whole process?

I mean, the steps to do so are mentioned in documentation. One would need to call `kubectl apply -f dev-gameserver.yaml`, then run `./sdk-server.linux.amd64 ...`, then run their custom game server binary. Everything is then connected.

CauhxMilloy:
Replied to explain a few points. Will commit more changes on relevant topics. Will re-merge main back in. Probably will get back to this tomorrow night.

cmd/sdk-server/main.go
markmandel: That's fair, and makes sense. My only thought here then would be to make the flag a positive name with a default of `true` rather than a negative with a default of `false`.

So `--graceful-termination=false` would probably be easier to reason with. WDYT?

pkg/gameservers/controller.go
markmandel: aaaah I see I see! I didn't expand up and see that this was specific to dev servers 👍🏻 That makes total sense. Thanks for the context 👍🏻 

We should add some unit and/or e2e tests for this then as well.

https://github.com/googleforgames/agones/blob/c431327653809997d193722788b66b1c0b55354d/pkg/gameservers/controller_test.go#L171

https://github.com/googleforgames/agones/blob/c431327653809997d193722788b66b1c0b55354d/test/e2e/gameserver_test.go#L475

site/content/en/docs/Guides/Client SDKs/local.md
markmandel: This says two things to me:

1. We should file a ticket to have SHA's with our binaries - that's a good point to make.
2. Maybe we need to improve our documentation on using the container based approach?

This particular documentation's focus is on local testing - not on compiling from source (maybe that should be elsewhere - would like more eyes on that before going down that path), so conflating the two makes things unnecessarily complicated.

I'd say remove it. If you know how to compile source, you know how to go down that path -- if not, it's teaching git, and go and everything else, all to do local testing, and that's a lot.

I don't disagree on the incentives to compile from source - but I think removing the documentation for it in this section is worth the better onboarding and developer documentation experience.

Calling in some extra eyes to see if I'm off base here: @zmerlynn @igooch @gongmax - WDYT?

site/content/en/docs/Guides/local-game-server.md
markmandel: A few thoughts on this:

1. From reading the documentation, the combination of the two approaches wasn't immediately obvious to me. I'm _almost_ wondering if this would be better as a part of https://agones.dev/site/docs/guides/local-game-server/ than here 🤔  (at the very least, they should probably link to each other though).
2. We should be super clear that you need both to actually connect. With the current docs it's easy to assume you can and you may not realise until the very end that you can't (or even that the `agones.dev/dev-address` GameSever approach exists -- which is kinda why I'm thinking we may want this in https://agones.dev/site/docs/guides/local-game-server/ as well, since one really does lead into the other.
3. Possibly a wacky idea 😄 since we can pass a gameserver.yaml to the local game server for state -- could we make the "out of cluster mode" server create the `agones.dev/dev-address` for you? It's using local kubectl permissions anyway -- save you a `kubectl apply`? Just trying to streamline development even more.

markmandel (commented):
I'm excited where this going 👍🏻 

Thanks for taking the time, and put some more thoughts and questions down.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: > We should file a ticket to have SHA's with our binaries - that's a good point to make.

That definitely sounds like a good idea.

> Maybe we need to improve our documentation on using the container based approach?

I think container based approaches for dev workflows have pros and cons. The major pros are that all you need is docker installed; after that, everything is just a few commands to get running. The downside is that its more levels of stuff to run/debug. If one is wanting to delve into Agones (like me 😄) or just wanting to have a simpler dev environment, I don't think forcing people into larger constructs (like containerization) is the best.

> This particular documentation's focus is on local testing - not on compiling from source

I don't particularly see the diff here. But, as I said in one of my other comments, I'm fine with changing bash samples over to use the downloaded binaries (for consistency).

> I'd say remove it. If you know how to compile source, you know how to go down that path

I still think that it's nice to at least nudge people in the right direction. When I read through this page, I just wanted to run from source. I had to then dig around the source code to find the file to run. If the documentation just said "you want `cmd/sdk-server/main.go`", that would have been amazingly helpful on its own.

> if not, it's teaching git, and go and everything else, all to do local testing, and that's a lot.

I don't think its the documentation's job to ensure the reader understands prerequisite info. Just like other pages aren't responsible for teaching the reader what REST is or how `curl` works. 🤷 Running a game server service on k8s is not trivial. If the reader cant run from source code, doesn't know what git is (on GitHub), etc, they're in a bit of a rut.

cmd/sdk-server/main.go
CauhxMilloy: I think `--blah=false` (with `--blah=true` as default) vs `--no-blah` is the same to me. I've definitely seen various binaries use either approach. Happy to change this around, if that's your preference.

site/content/en/docs/Guides/local-game-server.md
CauhxMilloy: > I'm _almost_ wondering if this would be better as a part of https://agones.dev/site/docs/guides/local-game-server/ than here 🤔

> I'm thinking we may want this in https://agones.dev/site/docs/guides/local-game-server/ as well

Ummm, this is https://agones.dev/site/docs/guides/local-game-server/ (`site/content/en/docs/Guides/local-game-server.md`). Maybe I'm misunderstanding you? 😅

> (at the very least, they should probably link to each other though).

I added links to both pages, this is the one from https://agones.dev/site/docs/guides/local-game-server/ to https://agones.dev/site/docs/guides/client-sdks/local/. In `site/content/en/docs/Guides/Client SDKs/local.md` (line 180) I have the reverse link.

>  could we make the "out of cluster mode" server create the agones.dev/dev-address for you?

That seems like a large expansion of the behavior of the SDK Server. While it obviously isn't my call to say what's best for the codebase, it just seems like that change would be larger and more out of place.

cmd/sdk-server/main.go
CauhxMilloy: Updated. Renamed to `--graceful-termination` and flipped logic.

site/content/en/docs/Guides/local-game-server.md
markmandel: > Ummm, this is https://agones.dev/site/docs/guides/local-game-server/ (site/content/en/docs/Guides/local-game-server.md). Maybe I'm misunderstanding you? 

Sorry, I 100% wasn't clear. I was wondering it the whole new documentation set in `local.md` should be in this documentation - since my gut says that they are likely always going to be used together. WDYT? That also then solves any confusion around connection etc.

Also I was providing feedback generally to the docs, not just these specific pages -- understandably confusing.

> > could we make the "out of cluster mode" server create the agones.dev/dev-address for you?

> That seems like a large expansion of the behavior of the SDK Server. While it obviously isn't my call to say what's best for the codebase, it just seems like that change would be larger and more out of place.

What do you think of the idea though? Can always be a secondary issue filed for future work, just thought it might be handy.

site/content/en/docs/Guides/Client SDKs/local.md
markmandel: One extra thought I had as a probable good middle ground (and in combination with the other comment I added to `local-game-server.md`):

1. Move most of the documentation for how to run "out of cluster" into `local-game-server.md`, but do the docs with running as a binary (assuming you agree they go better together as a development flow), as that is the general explanation of "how do I do an out of cluster setup"
2. On this page, we already have a "Running Local Mode in a Container", so why don't we add a section right below that of "Running Local Mode from source code" - where we _can_ explain how to run from source (and if need be explain how to git clone the right release branch etc).

That way we aren't conflating running from source with how to do other things, but if you want the source path, then you can definitely go down that path, and the docs for how to use the local sdk can just use the binary for simplicity.

WDYT?

zmerlynn: I'm generally in agreement with @markmandel that running from source is not the direction I would prefer to support on the public docs. The main reason is just that: support. Documenting a path comes with a certain "warranty", implied or not, and I can easily see someone filing a bug because e.g. `go run` didn't work for them due to differing Go versions. I I don't personally want to be in the business of supporting every user's Go environ and understanding how it differs from our build image.

I'm not against bumping this out to a section with a disclaimer, though.

pkg/gameservers/controller.go
CauhxMilloy: Added the related tests in the suggested locations.

site/content/en/docs/Guides/local-game-server.md
CauhxMilloy: > I was wondering it the whole new documentation set in `local.md` should be in this documentation - since my gut says that they are likely always going to be used together.

I suppose it _could_ go in either place. Personally, I feel like this has _more_ to do with running the local SDK Server (as you need to change the commands used when running it). I definitely think "out of cluster" mode should be mentioned in both spots.

> What do you think of the idea though? Can always be a secondary issue filed for future work, just thought it might be handy.

If you're just asking for my opinion, I don't particularly see any use for it (for me). I suppose some might like omitting a `kubectl apply`, but that's not really what I'd be optimizing for. What I'd be looking for is "breakpoint locally but in the most prod-y way possible". As the expected order of lifetimes/creation on prod would be: `GameServer` resource, SDK Server binary, custom game server binary, I would want to keep it in that order. This means that the custom game binary, the allocation system, etc, can all be run in a consistent state (code abstracting away from this flow vs prod). Swapping around the order or operations, or changing what creates what, for dev testing (in such an integrated context) doesn't really make sense to me, as it isn't representative of how things will be working at the end.

Speaking of "other things that could be done for handy-ness":
As I alluded to in https://github.com/googleforgames/agones/pull/3252#discussion_r1259155061, I have other desires for potential future PRs. Namely that dev servers wouldn't automatically go from `Creating` to `Ready`. Instead, they would do a normal `Creating` to `Scheduled`, and would only progress to `Ready` after the SDK call was made (+ controller state progression) -- all of this would match https://agones.dev/site/docs/reference/gameserver/#gameserver-state-diagram. An `annotation` could be used to "auto-ready" (e.g. `agones.dev/auto-ready: true`) to keep the current functionality. Also an `annotation` for `agones.dev/recycle-gs: true` would be nice to progress from `Shutdown` to `Creating`/`Scheduled` (so when the binary exits, calling SDK Shutdown, the `GameServer` would just be reset instead of deleted). Again, all of this is a bit different/extra -- this PR is just trying to add in the finishing touches for a very useful flow that's mostly there already. Lol, I hesitated to really go into all this extra stuff as I don't want it to distract from this (seemingly more straightforward) PR. Hopefully this helps with the perspective that I'm looking at things from.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: > Move most of the documentation for how to run "out of cluster" into `local-game-server.md`

Maybe this is a bit of a duplicated topic, I also wrote in https://github.com/googleforgames/agones/pull/3252#discussion_r1264965321 that I think it _could_ go in either place. But here made more sense as this is where the user is changing what is running (different cli args).

> On this page, we already have a "Running Local Mode in a Container", so why don't we add a section right below that of "Running Local Mode from source code" - where we can explain how to run from source (and if need be explain how to git clone the right release branch etc).

That sounds good to me. I can re-arrange the sections to make it more linear/separated out.

> I don't personally want to be in the business of supporting every user's Go environ and understanding how it differs from our build image.
> I'm not against bumping this out to a section with a disclaimer, though.

I like disclaimers. 😄 I think this is a very understandable stance to take, I'd be happy to add in a "this is here to document that it can be done; you're on your own for ensuring your golang environment is setup properly".

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -13,25 +13,41 @@ namespace as it - usually referred to in Kubernetes terms as a "sidecar".
...
-* sdk-server.darwin.arm64
+### Running from source code
 
-Linux
-* sdk-server.linux.amd64
-* sdk-server.linux.arm64
+Instead of downloading and running executable binaries from the internet, you can simply run from source code.
+First clone the [Agones GitHub repo](https://github.com/googleforgames/agones).
+You can switch to a specific release's branch/tag or run from main.
+Running from source code will require having golang installed, which can be done by following the instructions [here](https://go.dev/doc/install).
```
markmandel, line 40:
```suggestion
If you wish to run from source rather than pre-built binaries,  you will need [Go installed](https://go.dev/doc/install).

**Disclaimer:** Agones is run and tested with the version of Go specified by the `GO_VERSION` variable in the project's [build Dockerfile](https://github.com/googleforgames/agones/blob/main/build/build-image/Dockerfile). Other versions are not supported, but may still work.

You will need to clone the [Agones GitHub repo](https://github.com/googleforgames/agones) and switch to your installed specific release's branch/tag. For example:

```bash
git clone https://github.com/googleforgames/agones.git
cd agones
git checkout release-{{< release-version >}}
```

```

Adding an example and simplifying the language (although I think my suggestion formatting in here got a bit weird).

File: site/content/en/docs/Guides/Client SDKs/local.md [hunk 9]
```diff
@@ -13,25 +13,41 @@ namespace as it - usually referred to in Kubernetes terms as a "sidecar".
...
 
-Windows
-* sdk-server.windows.amd64.exe
+With golang installed and the Agones repository cloned, the SDK Server can easily be run with the following command (from the agones clone directory):
+```bash
+go run cmd/sdk-server/main.go
+```
+Note: This command does not contain any of the necessary command line flags used in the following sections.
+It simply serves as an example of how to run from source code instead of running the prebuilt binary.
+Passing commandline flags (e.g. `--local`) will work in the same way.
```
markmandel, line 48:
```suggestion
With Go installed and the Agones repository cloned, the SDK Server can be run with the following command (from the Agones clone directory):
```bash
go run cmd/sdk-server/main.go --local
/```

Commandline flags (e.g. `--local`) are exactly the same as command line flags when utilising a pre-built binary.
```

Tech writer nit: Some things aren't "easily" or "simple" for some people, so better to remove the adjective.

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -40,6 +56,11 @@ For example:
 ```bash
 ./sdk-server.linux.amd64 --local
 ```
+or
+```bash
+go run cmd/sdk-server/main.go --local
+```
```
markmandel, line 62:
```suggestion
```

Since we have the section above now.

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -13,25 +13,41 @@ namespace as it - usually referred to in Kubernetes terms as a "sidecar".
...
+To run the SDK Server, you will need a copy of the binary.
+This can either be done by downloading a prebuilt binary or running from source code.
+
+### Getting the prebuilt binary
+
+To run the prebuilt binary, for the latest release, you will need to download {{% ghrelease file_prefix="agonessdk-server" %}}, and unzip it.
 You will find the executables for the SDK server, for each type of operating system.
+`sdk-server.darwin.amd64` and `sdk-server.darwin.arm64` are for <u>MacOS</u>, `sdk-server.linux.amd64` and `sdk-server.linux.arm64` are for <u>Linux</u>, and `sdk-server.windows.amd64.exe` is for <u>Windows</u>.
 
-macOS
```
markmandel, line 25: I'd prefer to keep this section in. Not everyone knows what OS matches to what code names, etc. Better to be explicit.

site/content/en/docs/Guides/local-game-server.md
markmandel: These are cool ideas! I'd love you to write up some issues and designs for these and we can discuss them further.

Thinking about docs organisation some more, https://agones.dev/site/docs/guides/client-sdks/local/ is about a purely local experience (and the simplest to get started with).

This file is about integrating with your online system, likely tied to how you like to run things in production (also it's further down in the docs, so it's assumed by this point, you've managed to work out how the SDK works, done some integration etc). 

Jumping from "how do I integrate the SDK" to "now I'm working in a cluster" within the same page as it is now seems to me like a big leap concept with for new user onboarding to me.

So let's move the out-of-cluster documentation into a new section here, and add a link to this doc at the bottom (I'll add a suggestion to `local.md` on what I think will work there), since I think the annotations and the out-of-cluster mode is all interlinked (I don't see one being done without the other), without changing the learning path drastically.

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -149,3 +172,52 @@ wget https://raw.githubusercontent.com/googleforgames/agones/{{< release-branch
 chmod o+r gameserver.yaml
 docker run --network=host --rm -v $(pwd)/gameserver.yaml:/tmp/gameserver.yaml us-docker.pkg.dev/agones-images/release/agones-sdk:{{<release-version>}} --local -f /tmp/gameserver.yaml
 ```
+
```
markmandel, line 175: The same as our other docs: https://github.com/googleforgames/agones/blob/3d57eaa4fe18590634c17137e369302c981b91b9/site/content/en/docs/Integration%20Patterns/allocation-from-fleet.md?plain=1#L28

```suggestion

## Next Steps:
* Learn how to connect your local development game server binary into a running Agones Kubernetes cluster for even more live [development options](({{< ref "/docs/Guides/local-game-server.md" >}})).

```

(This would be at the bottom of the page -- see note on moving below to `local-game-server.md`).

markmandel (commented):
Code looks good! Just nailing down the documentation! 👍🏻 

Since this is so tied to that onboarding and development loop, want to make sure it's as seamless as possible to learn.

CauhxMilloy:
Just to touch base here, I know I have more changes to make here. Will have time by next Monday (during the weekend, if not later this week). Sorry for the hold up.

markmandel:
> Just to touch base here, I know I have more changes to make here. Will have time by next Monday (during the weekend, if not later this week). Sorry for the hold up.

No worries at all - we all have lives 😄  appreciate your contributions!

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: Added `publishDate` to new page. Wrapped links with `{{% feature publishVersion="1.34.0" %}}`.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: Just to follow up here, this work has been done. `--gameserver-name` and `--pod-namespace` can be used as cli args, and `GAMESERVER_NAME` and `POD_NAMESPACE` still work as env vars (which are used in k8s). Also updated this to be consistent and use pre-built binary commands.

site/content/en/docs/Guides/local-game-server.md
CauhxMilloy: >  I'd love you to write up some issues and designs for these and we can discuss them further.

I went ahead and filed https://github.com/googleforgames/agones/issues/3284.

>  let's move the out-of-cluster documentation into a new section

I've now moved all the "out of cluster" documentation into `site/content/en/docs/Advanced/out-of-cluster-dev-server.md`. This actually allowed me to go much further into various details. I've updated `site/content/en/docs/Guides/Client SDKs/local.md` and `site/content/en/docs/Guides/local-game-server.md` with `Next Steps` that link to out of cluster configuration.

pkg/gameservers/controller_test.go
```diff
@@ -224,6 +224,41 @@ func TestControllerSyncGameServerWithDevIP(t *testing.T) {
...
+			ua := action.(k8stesting.UpdateAction)
+			gs := ua.GetObject().(*agonesv1.GameServer)
+			updateCount++
+
+			assert.Equal(t, agonesv1.GameServerStateReady, gs.Status.State)
+
+			return true, gs, nil
+		})
+
+		ctx, cancel := agtesting.StartInformers(mocks, c.podSynced)
```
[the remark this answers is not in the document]
CauhxMilloy: Thanks, missed this when putting this test together.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: I've moved "running from source" down to the last section (just above `Next Step`). Reworded to be a mix of new and old. Feel free to suggest further edits.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: Cleaned up any `go run` outside of the "run from source" section.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: Cleaned up.

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: This section was still here. It's right above this. I just changed this to be a formatted sentence, rather than half-bullets. I went ahead and changed this mostly back (although, now its a single set of bullets).

site/content/en/docs/Guides/Client SDKs/local.md
```diff
@@ -13,25 +13,41 @@ namespace as it - usually referred to in Kubernetes terms as a "sidecar".
...
+### Getting the prebuilt binary
+
+To run the prebuilt binary, for the latest release, you will need to download {{% ghrelease file_prefix="agonessdk-server" %}}, and unzip it.
 You will find the executables for the SDK server, for each type of operating system.
+`sdk-server.darwin.amd64` and `sdk-server.darwin.arm64` are for <u>MacOS</u>, `sdk-server.linux.amd64` and `sdk-server.linux.arm64` are for <u>Linux</u>, and `sdk-server.windows.amd64.exe` is for <u>Windows</u>.
 
-macOS
-* sdk-server.darwin.amd64
-* sdk-server.darwin.arm64
+### Running from source code
```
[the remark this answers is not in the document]
CauhxMilloy: Moved to last section (before `Next Steps`).

site/content/en/docs/Guides/Client SDKs/local.md
CauhxMilloy: Added/updated `Next Steps` on `site/content/en/docs/Guides/Client SDKs/local.md` and `site/content/en/docs/Guides/local-game-server.md` to link to the new `site/content/en/docs/Advanced/out-of-cluster-dev-server.md` page.

site/content/en/docs/Advanced/out-of-cluster-dev-server.md
```diff
@@ -0,0 +1,176 @@
+---
+title: "Out of Cluster Dev Server"
+linkTitle: "Out of Cluster Dev Server"
+date: 2023-07-22T17:21:25Z
+publishDate: 2023-08-15T07:00:00Z
```
markmandel, line 5: Nice pickup on the date 😄

markmandel (approved):
LGTM! 

I checked the documentation rendering too, and it looks good 👍🏻  nice!

Holding off merging until @zmerlynn takes a look just in case you had any final thoughts.

markmandel:
gke-autopilot-1.25 flaked again. Weird.

markmandel:
Zack's away at the moment, so I'm going to go forward and merge this - we can always clean up later if we decide we need to.

Thank you very much for your contribution and continued persistence!

PR #3253, opened 2023-07-10
Use maintenance exclusion to prevent auto-upgrade, add 1.27 test clusters

gongmax:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

The Kubernetes version of the test cluster should stay at the same minor version as when it was created. Use maintenance exclusion to prevent auto-upgrade.

Also include the 1.27 test clusters in this PR.
The clusters have already been updated: https://pantheon.corp.google.com/kubernetes/list/overview?e=13803378&hl=en&jsmode=o&mods=logs_tg_prod&project=agones-images

Closes #


zmerlynn:
Hmm, can we look at what it would look like to do an [exclusion window](https://cloud.google.com/kubernetes-engine/docs/concepts/maintenance-windows-and-exclusions) instead, since it's what we now recommend in the docs?

install/terraform/modules/gke-autopilot/cluster.tf
```diff
@@ -63,6 +65,21 @@ resource "google_container_cluster" "primary" {
...
+    # When exclusions and maintenance windows overlap, exclusions have precedence.
+    daily_maintenance_window {
+      start_time = "03:00"
+    } 
+    maintenance_exclusion{
+      exclusion_name = format("%s-%s", local.name, "exclusion")
+      start_time = local.maintenanceExclusionStartTime
+      end_time = local.maintenanceExclusionEndTime
+      exclusion_options {
+        scope = "NO_MINOR_OR_NODE_UPGRADES"
```
zmerlynn, line 78: I think we're ok with just `NO_MINOR_UPGRADES`. On GKE the nodes will never get upgraded past the minor version of the master, so `NO_MINOR_UPGRADES` locks the nodes, too - but will allow the version to be updated _within_ the minor.

zmerlynn (approved):

Issue #909, opened 2019-07-15
Should we rebase images from alpine to gcr.io/distroless/static ?


roberthbailey:
Kubernetes has moved to using gcr.io/distroless/static for most base images (see https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190316-rebase-images-to-distroless.md#implementation-history). We should investigate whether it makes sense to move from alpine to gcr.io/distroless/static for Agones.

markmandel:
Going through old issues :smile: 

https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md

Looks like we should be able to switch straight to `gcr.io/distroless/static` without any issue.

markmandel:
I'd like to return to this one day. Distroless is very much the "google way" - so at some point we should probably move over.

roberthbailey:
@igooch - FYI

markmandel:
I'm shifting this to "awaiting-maintainer" - I think we should do it.

aimuz:
I have time for this.

markmandel:
Starting on this right now - Alphin 3.14 stopped getting security updates in May, so seemed like a good time to make the change across the board.

PR #3270, opened 2023-07-17
Alpine ➡️ Distroless as Agones base image

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix
> /kind release

Alpine 3.14 stopped getting security updates back in May of this year, so we needed to update things anyway. Seemed like a good time to move over to the Google supported base container image.

Also since we're always building off distroless' latest image, we don't need to update version numbers to make sure releases always have the latest version, so we always get ongoing security updates.

Work on #909

Next step to close out #909 would be to update the examples as well (particularly simple-game-server).

### DIFF
--- a/cmd/allocator/Dockerfile
+++ b/cmd/allocator/Dockerfile
@@ -12,14 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.14
+FROM gcr.io/distroless/static-debian11:nonroot
 
 ARG ARCH=amd64
-RUN apk --update add ca-certificates && \
-    adduser -D -u 1000 agones
+WORKDIR /
 
-COPY --chown=agones:agones ./bin/allocator.linux.$ARCH /home/agones/allocator
-COPY --chown=agones:agones ./bin/LICENSES ./bin/dependencies-src.tgz /home/agones/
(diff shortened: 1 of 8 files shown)


cmd/allocator/Dockerfile
```diff
@@ -12,14 +12,14 @@
...
 
-COPY --chown=agones:agones ./bin/allocator.linux.$ARCH /home/agones/allocator
-COPY --chown=agones:agones ./bin/LICENSES ./bin/dependencies-src.tgz /home/agones/
+COPY ./bin/allocator.linux.$ARCH /allocator
+COPY ./bin/LICENSES .
+COPY ./bin/dependencies-src.tgz .
 
-USER 1000
-ENTRYPOINT ["/home/agones/allocator"]
+USER nonroot:nonroot
```
gongmax, line 24: A dump question, is this needed if the nonroot image is used?

markmandel: You know, I always assumed so, but the distroless docs are... not the best 😄 so it's a good question to ask.

We can try and see.

gongmax (approved):

aimuz (approved):

PR #3271, opened 2023-07-18
SDK WatchGameServer logs error on shutdown

### LINKED ISSUE #3245: Golang SDK `WatchGameServer` logs error on shutdown
**What happened**:

I have a GameServer that uses the Golang SDK. For this GameServer I register a callback using `WatchGameServer`.

When the GameServer shuts down, for example on Fleet scale down, I often get error logs that I believe are not really errors.

More specifically here https://github.com/googleforgames/agones/blob/main/sdks/go/sdk.go#L142C3-L142C3
and here https://github.com/googleforgames/agones/blob/main/sdks/go/sdk.go#L145

1. In the first case, when `err == io.EOF` I think the log should not be an error at all, as `EOF` signals that the stream was closed gracefully.

2. In the second case, the full error is `error watching GameServer: rpc error: code = Unavailable desc = error reading from server: EOF`. I am not sure why the error is wrapped here and not directly `io.EOF` but in this case, as the GameServer is shutting down, it make sense to get an EOF and maybe it should not be considered an error. 
Maybe we could record the GameServer state in the callback loop, and if the last state was shutdown, then do not log this message as error?

3. This leads me to another question: messages here are logged directly to `os.Stderr`. It would be great if some logger was used instead so it is more configurable. I've seen logrus used in agones implementation. Would it make sense to use it here too? Maybe passing a `*logrus.Entry` as input parameter to `NewSDK()`?

Write a GameServer using Golang SDK and register a callback using `WatchGameServer`. Trigger the GameServer shutdown and check the GameServer logs.

**Environment**:
- Agones version: 1.32.0
- Install method (yaml/helm):
- Others:

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Replaced `fmt` statements with `log` statements and added a condition to prevent logging when `gs.ObjectMeta.DeletionTimestamp` is not equal to zero

Closes #3245


aimuz:
@clairesc for review

sdks/go/sdk.go
```diff
@@ -139,14 +140,18 @@ func (s *SDK) WatchGameServer(f GameServerCallback) error {
...
 					return
 				}
-				_, _ = fmt.Fprintf(os.Stderr, "error watching GameServer: %s\n", err.Error())
+				log.Printf("error watching GameServer: %s\n", err.Error())
 				// This is to wait for the reconnection, and not peg the CPU at 100%.
 				time.Sleep(time.Second)
 				continue
 			}
+
+			if gs.ObjectMeta.DeletionTimestamp != 0 {
```
markmandel, line 152: 🤔 

So I would handle this a bit differently. There are two problems we are trying to resolve:

1. Not output errors here when shutting down
2. Have a pluggable log system.

So, I would actually make the second part, part of the first - because we still want to handle `io.EOF` as above, but not output the log.

So I would have a pluggable log as per:
https://github.com/googleforgames/agones/issues/3245#issuecomment-1623999369

But have a function that is called here where the log functions are, that checks if `gs.ObjectMeta.DeletionTimestamp.isZero()` (I think that's the right function)` and if not, then runs the pluggable log function.

Does that make sense?

Kalaiselvi84: I've included the Logger function to log the error message when `gs.ObjectMeta.DeletionTimestamp` is not zero and removed log when  "err == io.EOF" and any other error occurs while watching the GameServer.

`DeletionTimestamp` is of type int64, so I added the following condition:
 `if reflect.ValueOf(gs.ObjectMeta.DeletionTimestamp).IsZero()`  to check if it is zero or not. 
 [reference](https://stackoverflow.com/questions/33115946/how-to-know-if-a-variable-of-arbitrary-type-is-zero-in-golang)

sdks/go/sdk.go
```diff
@@ -42,6 +43,14 @@ type SDK struct {
 	alpha  *Alpha
 }
 
+// ErrorLog is a function to log the error.
+type ErrorLog func(error, string)
+
+// Logger is a pluggable function that outputs the error message to standard error.
+var Logger ErrorLog = func(err error, msg string) {
+	fmt.Fprintf(os.Stderr, "msg: %s\n", err)
```
markmandel, line 51:
```suggestion
	fmt.Fprintf(os.Stderr, "%s: %s\n", msg, err)
```

Since the message isn't in the log.

sdks/go/sdk.go
```diff
@@ -136,17 +145,21 @@ func (s *SDK) WatchGameServer(f GameServerCallback) error {
...
-					_, _ = fmt.Fprintln(os.Stderr, "gameserver event stream EOF received")
 					return
 				}
-				_, _ = fmt.Fprintf(os.Stderr, "error watching GameServer: %s\n", err.Error())
 				// This is to wait for the reconnection, and not peg the CPU at 100%.
 				time.Sleep(time.Second)
 				continue
 			}
+
+			if !reflect.ValueOf(gs.ObjectMeta.DeletionTimestamp).IsZero() {
```
markmandel, line 159: Two things:

1. https://pkg.go.dev/time#Time.IsZero exists - no need to reflect.
2. We still need to have at least similar log output as before - right now we're reversing what is required.

My suggestion on a function previously could even be within `WatchGameServer`, something like:

```go
var log = function(gs *GameServer, msg string, err error) {
  if gs.ObjectMeta.DeletionTimestamp.IsZero() {
    return
  }
  ErrorLog(err, msg)
}
```
Then each of the log statements to stderr could become:

```go
log(gs, nil, "gameserver event stream EOF received")
```

and 

```go
logf(gs, err, "error watching GameServer")
```

respectively.

make sense?

sdks/go/sdk.go
```diff
@@ -42,6 +43,14 @@ type SDK struct {
 	alpha  *Alpha
 }
 
+// ErrorLog is a function to log the error.
+type ErrorLog func(error, string)
+
+// Logger is a pluggable function that outputs the error message to standard error.
+var Logger ErrorLog = func(err error, msg string) {
```
markmandel, line 50: I feel like error should come second - like it would in a printf statement. WDYT?

```suggestion
var Logger ErrorLog = func(msg string, err error) {
```

Kalaiselvi84: Swapped the positions

sdks/go/sdk.go
Kalaiselvi84: 1. Removed reflect and added `time.Unix(gs.ObjectMeta.DeletionTimestamp, 0).IsZero()`
2. The code defines a function called 'log' that checks if the 'DeletionTimestamp' is zero. If it is, the function returns nothing, otherwise the function calls the 'Logger' function with a message and an error parameter.
       Included log statement for the condition "err == io.EOF" and other errors.

sdks/go/sdk.go
Kalaiselvi84: done!

sdks/go/sdk.go
```diff
@@ -132,17 +140,22 @@ func (s *SDK) WatchGameServer(f GameServerCallback) error {
 	if err != nil {
 		return errors.Wrap(err, "could not watch gameserver")
 	}
-
+	log := func(gs *sdk.GameServer, msg string, err error) {
+		if time.Unix(gs.ObjectMeta.DeletionTimestamp, 0).IsZero() {
```
markmandel, line 144:
```suggestion
		if gs.ObjectMeta.DeletionTimestamp.IsZero() {
```
This _should_ work 🤔 , here is an example - we do this a bunch of places.

https://github.com/googleforgames/agones/blob/80f74c4e0f257683cbe5318ae222368cf2442537/pkg/gameservers/controller.go#L428

More examples: https://github.com/search?q=repo%3Agoogleforgames%2Fagones%20IsZero&type=code

markmandel: Okay, big slice for humble 🥧 for me. I didn't realise this was the protobuf version, not the K8s API version 🤦🏻 

So, we can simplify this to:

```go
gs.ObjectMeta.DeletionTimestamp == 0
```

And that will make everything happy.

Sorry about that!

markmandel (commented):
LGTM!

@clairesc can you take a quick look, make sure we didn't miss anything?

clairesc:
I wanted to say LGTM but I gave it a quick go and got this exception when the GameServer is shutdown (via Fleet autoscaling)
```
goroutine 23 [running]:
agones.dev/agones/sdks/go.(*SDK).WatchGameServer.func1(0xc00036f5b0?, {0x146da6b?, 0xc00008ff20?}, {0x0?, 0x0?})
        /go/src/agones/sdks/go/sdk.go:144 +0x23
agones.dev/agones/sdks/go.(*SDK).WatchGameServer.func2()
        /go/src/agones/sdks/go/sdk.go:155 +0x119
created by agones.dev/agones/sdks/go.(*SDK).WatchGameServer
        /go/src/agones/sdks/go/sdk.go:149 +0x14a
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x11e0003]
```

I guess `gs` is `nil` when `stream.Recv` returns an error? (which would makes sense)

sdks/go/sdk.go
```diff
@@ -42,6 +42,14 @@ type SDK struct {
 	alpha  *Alpha
 }
 
+// ErrorLog is a function to log the error.
+type ErrorLog func(string, error)
+
+// Logger is a pluggable function that outputs the error message to standard error.
+var Logger ErrorLog = func(msg string, err error) {
+	fmt.Fprintf(os.Stderr, "%s: %s\n", msg, err)
```
markmandel, line 50:
```suggestion
  if gs == nil {
    return
  }
	fmt.Fprintf(os.Stderr, "%s: %s\n", msg, err)
```

Since @clairesc 's comment is very good point!

Kalaiselvi84: can I add the `gs == nil` check here? WDYT?
log := func(gs *sdk.GameServer, msg string, err error) {
    if gs == nil || gs.ObjectMeta.DeletionTimestamp == 0 {
        return
    }
    Logger(msg, err)
}

markmandel: Yeah that's way better.

markmandel:
I think I fixed this flake. Updating.

markmandel:
@clairesc look good now?

clairesc:
> @clairesc look good now?

Yes LGTM

Kalaiselvi84:
@markmandel, It looks like the PR didn't get merged automatically after the successful build. Could you kindly perform the merge for this PR?🙂

markmandel (approved):
Looks like I forgot to approve. Done!

PR #3277, opened 2023-07-19
Distroless base image for crd-client

### LINKED ISSUE #909: Should we rebase images from alpine to gcr.io/distroless/static ?

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #909


examples/crd-client/Makefile
```diff
@@ -25,7 +25,7 @@
 
 REPOSITORY ?= us-docker.pkg.dev/agones-images/examples
 
-server_tag = $(REPOSITORY)/crd-client:0.9
+server_tag = $(REPOSITORY)/crd-client:1.0
```
roberthbailey, line 28: @markmandel - how do you feel about going from 0.9 to 1.0 vs. 0.9 to 0.10? Are we happy calling this image a 1.0 image?

examples/crd-client/Dockerfile
```diff
@@ -14,21 +14,18 @@
 
 # Gather dependencies and build the executable
 FROM golang:1.20.4 as builder
-WORKDIR /go/src/crd-client
+WORKDIR /
```
roberthbailey, line 17: I'm curious why you changed the paths in the build container. That shouldn't affect the location in the final container.

examples/crd-client/Dockerfile
```diff
@@ -14,21 +14,18 @@
...
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o client .
 
 # Create the final image that will run the webhook server for FleetAutoscaler webhook policy
-FROM alpine:3.13.1
-RUN adduser -D -u 1000 client
+FROM gcr.io/distroless/static-debian11:nonroot
 
-COPY --from=builder /go/src/crd-client \
-                    /home/client
+COPY --from=builder /client /
```
roberthbailey, line 26: I do like just copying the binary over; that seems like an improvement from the previous commands. 👍

examples/crd-client/Dockerfile
Kalaiselvi84: Mark suggested moving the `WORKDIR` path to the root directory ('/'). 
Example PR: https://github.com/googleforgames/agones/pull/3270/files

examples/crd-client/Makefile
markmandel: That's a good point - I figure let's go to 0.10 -- examples can always change I guess (or at least we should be consistent with our examples).

examples/crd-client/Dockerfile
markmandel: I was being consistent with everything else I've seen with distroless / done with distroless.

There's no $HOME folders in distroless, so it seems like everyone drops everything in /.

markmandel: Oh sorry, I should have looked at this properly - this is the build step.

🤷🏻 actually for this, I have no strong opinion.

markmandel:
For testing for this -- take as standard release/dev Agones install, create this as an image and apply the yaml - it's shouldn't fail, and you can see the output in it's logs.

You can see a step through here: 
https://agones.dev/site/docs/guides/access-api/#example

examples/crd-client/Dockerfile
Kalaiselvi84: Based on testing, the change to the root directory ('/') for WORKDIR in the build container didn't work as expected, so I've reverted it back.

examples/crd-client/Makefile
Kalaiselvi84: I've changed it to 0.10

markmandel (approved):
Nice!

PR #3279, opened 2023-07-20
Move simple-game-server to Distroless base

markmandel:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**What this PR does / Why we need it**:

This includes:
* Replacing Alpine with Distroless as the base image
* Tweaked the creation of the mulit-arch image manifest to use a more recent command (something changed in Docker 24.0.4 that broke the current setup, also this is less steps).
* Updated all the reference to simple-game-server:0.16 to simple-game-server:0.17.

**Which issue(s) this PR fixes**:

Work on #909

**Special notes for your reviewer**:

First run of this will fail, as I haven't pushed up the image. Wanted to get a set of eyes on it first before pushing up the new version.

Local testing has this passing the e2e tests I manually ran (not the full suite), so I'm reasonably confident 😄 .

PR #3280, opened 2023-07-20
Flaky: TestAllocatorAllocatePriority

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Fix a flake that would see a feature flag flip _just_ between `ApplyDefaults()` and `Validate()` -- which is more impressive than anything else.

```
--- FAIL: TestAllocatorAllocatePriority (1.18s)
    allocator_test.go:169:
        	Error Trace:	/go/src/agones.dev/agones/pkg/gameserverallocations/allocator_test.go:169
        	            				/go/src/agones.dev/agones/pkg/gameserverallocations/allocator_test.go:176
        	Error:      	"[spec.selectors[0].counters: Forbidden: Feature CountsAndLists must be enabled spec.selectors[0].lists: Forbidden: Feature CountsAndLists must be enabled]" should have 0 item(s), but has 2
        	Test:       	TestAllocatorAllocatePriority
```

N/A

### DIFF
--- a/pkg/gameserverallocations/allocator_test.go
+++ b/pkg/gameserverallocations/allocator_test.go
@@ -131,6 +131,10 @@ func TestAllocatorAllocate(t *testing.T) {
 func TestAllocatorAllocatePriority(t *testing.T) {
 	t.Parallel()
 
+	// TODO: remove when `CountsAndLists` feature flag is moved to stable.
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
+
 	run := func(t *testing.T, name string, test func(t *testing.T, a *Allocator, gas *allocationv1.GameServerAllocation)) {
 		f, gsList := defaultFixtures(4)
 		a, m := newFakeAllocator()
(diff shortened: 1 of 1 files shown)


gongmax (approved):
Interesting finding!

markmandel:
Just happened to happen to me on PR #3279!

igooch (commented):
This wasn't the test that flaked, it was `func TestAllocatorAllocatePriority`. Does this lock need to be added to all the tests?

Or should this be added to `func TestAllocatorApplyAllocationToGameServerCountsListsActions` after `t.Parallel` the instead of processing for each test case?

```
	runtime.FeatureTestMutex.Lock()
	assert.NoError(t, runtime.ParseFeatures(string(runtime.FeatureCountsAndLists)+"=true"))
```

markmandel:
> This wasn't the test that flaked, it was `func TestAllocatorAllocatePriority`. Does this lock need to be added to all the tests?

You raise a really good point! That's what I get for moving too fast.

I'll add it to the other tests this could possibly affect in this file.

PR #3282, opened 2023-07-21
GameServerAllocation to sort Priorities by Allocated Capacity

igooch:
**What type of PR is this?**

 /kind feature

Updates GameServerAllocation to sort by Allocated Capacity when sorting by Priorities.

Adds in a SortKey for GameServerAllocation Priorities  to differentiate between requests with different priorities and / or scheduling during  ListenAndAllocate.

Removes a duplicate import of k8s.io/apimachinery/pkg/api/validation.

Working on #2716

### DIFF
--- a/pkg/apis/allocation/v1/gameserverallocation.go
+++ b/pkg/apis/allocation/v1/gameserverallocation.go
@@ -21,7 +21,7 @@ import (
 	"agones.dev/agones/pkg/apis"
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	"agones.dev/agones/pkg/util/runtime"
-	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
+	hashstructure "github.com/mitchellh/hashstructure/v2"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
@@ -69,12 +69,12 @@ type GameServerAllocationList struct {
 type GameServerAllocationSpec struct {
 	// MultiClusterPolicySelector if specified, multi-cluster policies are applied.
 	// Otherwise, allocation will happen locally.
-	MultiClusterSetting MultiClusterSetting `json:"multiClusterSetting,omitempty"`
+	MultiClusterSetting MultiClusterSetting `json:"multiClusterSetting,omitempty" hash:"ignore"`
 
 	// Deprecated: use field Selectors instead. If Selectors is set, this field is ignored.
 	// Required is the GameServer selector from which to choose GameServers from.
 	// Defaults to all GameServers.
-	Required GameServerSelector `json:"required,omitempty"`
+	Required GameServerSelector `json:"required,omitempty" hash:"ignore"`
 
 	// Deprecated: use field Selectors instead. If Selectors is set, this field is ignored.
 	// Preferred is an ordered list of preferred GameServer selectors
@@ -82,12 +82,12 @@ type GameServerAllocationSpec struct {
 	// If the first selector is not matched, the selection attempts the second selector, and so on.
 	// If any of the preferred selectors are matched, the required selector is not considered.
 	// This is useful for things like smoke testing of new game servers.
-	Preferred []GameServerSelector `json:"preferred,omitempty"`
+	Preferred []GameServerSelector `json:"preferred,omitempty" hash:"ignore"`
 
 	// (Alpha, CountsAndLists feature flag) The first Priority on the array of Priorities is the most
 	// important for sorting. The allocator will use the first priority for sorting GameServers by
-	// available Capacity in the Selector set and acts as a tie-breaker after sorting the game servers
-	// by State and Strategy. Impacts which GameServer is checked first.
+	// available Capacity in the Selector set. Acts as a tie-breaker after sorting the game servers
+	// by State and Strategy Packed. Impacts which GameServer is checked first.
 	// +optional
 	Priorities []agonesv1.Priority `json:"priorities,omitempty"`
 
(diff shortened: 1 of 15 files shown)


pkg/gameserverallocations/allocation_cache.go
```diff
@@ -275,14 +276,16 @@ func compareGameServers(p *agonesv1.Priority, gs1, gs2 *agonesv1.GameServer) int
...
 				return -1
 			}
-			if counter1.Count > counter2.Count {
+			if availCapacity1 > availCapacity2 {
 				return 1
 			}
-			if counter1.Count == counter2.Count {
-				return 0
+			if availCapacity1 == availCapacity2 {
+				return 0 // TODO: Should we tie-break on Capacity and/or Count?
```
igooch, line 288: Should we tie-break on Capacity and/or Count before returning that two game servers with the same available capacity are equal?

markmandel: I was digging through the implementation, and I think it's fine we don't tie break here. 

We end up tie breaking anyway in 

https://github.com/googleforgames/agones/blob/2f8a583ec6430a093944746e83d68f2a6ed0e6be/pkg/gameserverallocations/allocation_cache.go#L176-L176

So I think it makes sense to keep that tie breaking in a single place.

pkg/gameserverallocations/allocator.go
```diff
@@ -496,6 +496,7 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
 	// continued.
 
 	var list []*agonesv1.GameServer
+	var sortKey *uint64 // Pointer so that we can check for nil (and set this value on the first req)
```
markmandel, line 499: 🤔 any reason the default value of 0 wouldn't suffice (rather than `nil`)? Save you managing pointers?

pkg/gameserverallocations/allocator.go
```diff
@@ -508,6 +509,26 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
...
+				newSortKey, err := req.gsa.SortKey()
+				if err != nil {
+					c.baseLogger.Warn("error getting sortKey for GameServerAllocationSpecPriorities", err)
+				}
+
+				if sortKey == nil {
+					sortKey = &newSortKey
+				}
+
+				if newSortKey != *sortKey {
```
markmandel, line 525: This looks great!

markmandel (commented):
Looks good - just had a couple of questions.

pkg/gameserverallocations/allocator.go
igooch: The SortKey() returns 0 if there's an error. An alternative would be to return 1 on error for SortKey(), and then sortKey as a non-pointer uint64, wdyt?

pkg/gameserverallocations/allocation_cache.go
igooch: Ah, that's right.

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -378,11 +378,11 @@ func (s *GameServerSelector) Validate(fldPath *field.Path) field.ErrorList {
 
 	if runtime.FeatureEnabled(runtime.FeaturePlayerAllocationFilter) && s.Players != nil {
 		if s.Players.MinAvailable < 0 {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("players").Child("minAvailable"), s.Players.MinAvailable, apimachineryvalidation.IsNegativeErrorMsg))
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("players").Child("minAvailable"), s.Players.MinAvailable, apivalidation.IsNegativeErrorMsg))
```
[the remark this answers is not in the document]
igooch: Wasn't my addition, but yes it is quite nice :).

pkg/gameserverallocations/allocator.go
markmandel: That does ask an interesting question - if the sortKey fails (admittedly very unlikely), should we reset sorting? In which case a 0 would be fine. WDYT?

igooch: Not quite following -- are you saying we should not reset the sortKey when a gameserverallocation generate sortKey fails?

markmandel: I think we're saying the same thing. That the `list` (and everything else) should reset if there was an error generating the `SortKey()`.

In which case, if `var sortKey *uint64` became `var sortKey uint64` - the 0 value would work the same as it does now - in which case, there's no real reason for the pointer.

In the current code, if `SortKey()` errors, it returns 0, which will likely reset the `list` (etc) values (assuming the last `sortKey` > 0).

Did that make more sense?

igooch: The pointer doesn't work exactly the same as 0 though.

If we got rid of lines 521-523, then the code as written would reset requestCount to 0 on the first request which is incorrect. 

If we changed line 521 to `sortKey == 0` then in the case of an error it would override the existing sortKey to be equal to 0 and continue on with the current batch, and then on the next (passing) request the newSortKey would not be equal to 0 and would reset requestCount and list. This slits the batch into two, with the error being attached to the first batch of requests.

As the code is currently written the error (and any directly adjacent errors) would be in their own batch, and passing batches would be grouped separately.

So maybe we have three options:

1. Totally ignore any errors (besides logging)
2. Split on error and attach the error to the previous batch
3. Split on error where any error(s) is in its own batch

pkg/gameserverallocations/allocator.go
```diff
@@ -508,6 +509,26 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
 				requestCount = 0
 			}
 
+			if runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+				// SortKey returns the sorting values (list of Priorities) as a determinstic key.
+				// In case gsa.Spec.Priorities is nil this will still return a sortKey.
+				// In case of error this will return 0 for the sortKey.
+				newSortKey, err := req.gsa.SortKey()
+				if err != nil {
+					c.baseLogger.Warn("error getting sortKey for GameServerAllocationSpecPriorities", err)
```
markmandel, line 518:
```suggestion
					c.baseLogger.WithError(err).Warn("error getting sortKey for GameServerAllocationSpecPriorities")
```

pkg/gameserverallocations/allocator.go
markmandel: > If we got rid of lines 521-523, then the code as written would reset requestCount to 0 on the first request which is incorrect.

Oooh, I see that `requestCount++` on line 506 - goooot it!

Okay yep - that makes sense then, let's leave it as `nil` 👍🏻 thanks for the explanation.

I think as you have it now makes a lot of sense.

igooch: Actually, would it make more sense to move the `requestCount++` to line 531? As is, requestCount = 0 doesn't make much sense when we're in the middle of processing a request.

markmandel: It's a slight change to the existing logic, but I actually have no strong opinions either way tbh. Your call 😄

igooch: If you have no strong feelings then I'm for it :)

markmandel (approved):
🎉

PR #3288, opened 2023-07-27
Add note about which namespace is used for game serves deployed from fleets

mikeseese:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix
> /kind release

This PR provides additional context about where `GameServer` resources are deployed from when instantiated by `Fleet`'s. Specifically, this addition attempts to remove confusion for those thinking that the `helm` chart value `gameservers.namespaces` changes where game servers are deployed by explicitly stating they deploy where the fleet is deployed. In the event that someone deploys a `Fleet` on a namespace that isn't included in the `helm` chart value `gameservers.namespaces`, they will end up receiving the error:

> error looking up service account agones-system/agones-sdk: serviceaccount "agones-sdk" not found

similar to #146 and #1370 (references included for others in the same situation as me finding those posts via Google).

This PR attempts to prevent users from getting into this scenario in the first place.

N/A

This PR was the result of this [Slack thread](https://agones.slack.com/archives/C9DGM5DS8/p1690336338842599)


markmandel (approved):
Sorry for the delay, was just dealing with some gnarly bugs.

Really good addition! Thanks for the contribution!

PR #3294, opened 2023-07-28
Bug/test allocator after delete replica

markmandel:
Just a draft with better debugging to see how it fails on GkE 1.27 standard.


markmandel:
No longer needed. Fix incoming.

PR #3306, opened 2023-08-01
Flaky: TestAllocatorAllocateOnGameServerUpdateError

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

No, but for-realsies, I think I got it this time.

So this is what I think the flake is:

On each retry in `Allocate` the attempted GameServer is removed from the list of batched GameServers.

We can retry up to 5 times, but because Jitter and machine speed, often it's less than that -- so the test passes with only 4 GamServers in the list.

But if the unit test is feeling speedy, it can work through all the potential `GameServers` and then come to the conclusion - nope, there are none left, so I'll say I can't find one.

The answer: Provide more GameServers than the allocation retries limit.

N/A

N/A


pkg/gameserverallocations/allocator_test.go
```diff
@@ -474,8 +477,8 @@ func TestAllocatorAllocateOnGameServerUpdateError(t *testing.T) {
 	require.NoError(t, a.Run(ctx))
 	// wait for 4 gameservers in the cache
```
gongmax, line 478: nit: better to update the comments, similar for line 503

markmandel: Good catch. Updating.

gongmax (approved):

PR #3299, opened 2023-08-01
Add Node.Status.Address to GameServer.Status in CRD and SDK

zmerlynn:
Adds `addresses` to `GameServer.Status` and SDK reflection of same. Example:

```
  status:
    address: 34.69.11.29
    addresses:
    - address: 10.128.0.82
      type: InternalIP
    - address: 34.69.11.29
      type: ExternalIP
    - address: gk3-autogones-pool-2-b43ef98c-shdq
      type: Hostname
```

Towards #3089 (next step is to add to allocation APIs)

PR #3307, opened 2023-08-02
Add GameServer addresses to the allocation APIs

### LINKED ISSUE #3089: Return all advertised Gameserver Addresses
The change suggested in https://github.com/googleforgames/agones/issues/1921 and addressed in  https://github.com/googleforgames/agones/pull/1928 introduced the following order of preference when selecting an address for advertisement on a gameserver resource:
https://github.com/googleforgames/agones/blob/02e57babccab6489b18d956286152174fe9a1eaa/pkg/gameservers/gameservers.go#L43-L68

However, there are use cases where `NodeInternalIP` might be a more preferable selection for a gameserver's advertised address.

Instead of a hard coded order for address selection, I would rather this logic be configurable to fit the specific situation.  This doesn't have to be something that is handled on a pod-by-pod basis, but could just be something that's configured for all gameservers on a cluster.

It seems overkill, but it might be cool to have this configuration option be part of the GameserverSpec, or PodSpec.

### LINKED ISSUE #3134: Support IPv6 for Game Server addresses
Game server addresses can currently only be IPv4 or DNS. DNS _can_ return an IPv6 address, but this requires some external integration to get right (see #2767, #1921). We could also support native IPv6 addresses for game servers rather than just DNS, though.

@chiayi tested on a [dual-stack GKE cluster](https://cloud.google.com/kubernetes-engine/docs/concepts/alias-ips#dual_stack_network) and things ~work - except during the `GameServer` address selection ends up picking the IPv4 `ExternalIP`. 

To properly support IPv6, I think it's just a matter of:

- [ ] Allow configuration of `GameServer` address selection: #3089 
- [ ] Extend e2e matrix to include IPv6

zmerlynn:
Follow on from #3299:

* Add `addresses` to allocation proto and `GameServerAllocation` response
* Document `address` and `addresses` on GameServer reference page

Note to reviewer: Commits starting with `make` are auto-generated bits, you can skip them (probably)

Closes #3089
Next steps: I'm going to test this with IPv6 to verify, then likely write an integration guide (done in #3311), which should be enough to complete #3134


roberthbailey (commented):
Did a quick pass and lgtm. Will defer to others for a more thorough review.

gongmax (approved):

PR #3322, opened 2023-08-11
Implements GetCounter and UpdateCounter on the SDK Server

igooch:
**What type of PR is this?**

/kind feature

Implements GetCounter and UpdateCounter on the SDK Server for feature Counts and Lists.
Removes the LocalSDKServer code for UpdateCounter as this followed the previous version of the alpha.proto.

Working on #2716


pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +784,196 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
-	return nil, errors.Errorf("Unimplemented -- GetCounter coming soon")
+
+	s.logger.WithField("name", in.Name).Info("Getting Counter")
+	s.gsUpdateMutex.RLock()
+	defer s.gsUpdateMutex.RUnlock()
+
+	// If we have made updates then return the SDK view of the Counter, otherwise get the counter from the gameserver
+	if counterUpdated, ok := s.gsCounterUpdates[in.Name]; ok {
+		s.logger.WithField("Get Counter from SDK Server", counterUpdated).Debugf("Got Counter %s", in.Name)
+		return &alpha.Counter{Name: in.Name, Count: counterUpdated.count, Capacity: counterUpdated.capacity}, nil
```
markmandel, line 795: I'm not sure how this applies to the gRPC standard, but at this point,  I feel like we should (assuming there is a diff) get the latest gameserver and apply any diff that exists/set overlap over the top of the existing counts, or possible values that are still in the batch pipeline.

The other thing we can do (with the suggestions above), is check the `Generation` stored with each processed batch request, and if `Generation` is greater than the one from `s.gameserver().ObjectMeta.Generation` we know we'll want to apply it to `gs.gameServer()`, since our local cache has yet to be populated with our batched changes.

(I hope that makes sense).

pkg/sdkserver/sdkserver.go
```diff
@@ -69,6 +71,23 @@ var (
...
+	// Tracks if there have been any CountSet requests
+	countSet bool
+	// Tracks if there have been any CapacitySet requests
+	capacitySet bool
+	// Capacity of the Counter (as retreived during first request) or overwritten by capacitySet.
+	capacity int64
+	// Count of the Counter (as retreived during first request) or overwritten by countSet.
+	count int64
+	// Tracks the sum of CountIncrement, CountDecrement, and/or CountSet requests from the client SDK.
+	diff int64
```
markmandel, line 88:
```suggestion
	diff int64
	// which generation from https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta where this is applied.
	Generation *int64
```

If we track which generation (it's only ever increases in value), then we can know if the current version we get back from `s.gameServer()` has the applied new batch in it's state or not, so we know whether to apply the changes stored in this batch (and if we should apply it again).

Other option to be inline in above, could also add a `batchApplied: bool` (or similar name).

pkg/sdkserver/sdkserver.go
```diff
@@ -105,6 +124,7 @@ type SDKServer struct {
 	gsReserveDuration   *time.Duration
 	gsPlayerCapacity    int64
 	gsConnectedPlayers  []string
+	gsCounterUpdates    map[string]counterUpdateRequest
```
markmandel, line 127:
```suggestion
	gsCounterUpdates    map[string][]counterUpdateRequest
```

See below on tracking generation against each request, so we know if it's been applied against the GameServer resource we know about at any given moment during the workflow.

Somewhere we would need to do some cleanup, but this is the general gist.

Digging into some of the later code, maybe it's easier to do something more like:

```suggestion
	gsActiveCounterUpdates    map[string]counterUpdateRequest
	gsProcessedCounterUpdates    map[string][]counterUpdateRequest
```

So it's easier to know what's yes to be processed, and what may end up being needed when doing `get` requests from the SDK.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +784,196 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+			tmpReq.capacity = in.CounterUpdateRequest.Capacity
+			// TODO: Do we need to check here that any previous changes to Count are still < Capacity?
+		default:
+			return nil, errors.Errorf("INVALID_ARGUMENT. Malformed CounterUpdateRequest: %v", in.CounterUpdateRequest)
+		}
+		s.gsCounterUpdates[name] = tmpReq
+	}
+
+	// Queue up the Update for later batch processing by updateCounters.
+	s.workerqueue.Enqueue(cache.ExplicitKey(string(updateCounters)))
```
markmandel, line 911: 👍🏻

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +784,196 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+
+	// Check if we already have a batch request started for this Counter. If not, add new request to
+	// the gsCounterUpdates map.
+	name := in.CounterUpdateRequest.Name
+	batchCounter, batchOk := s.gsCounterUpdates[name]
+
+	if !batchOk {
+		gs, err := s.gameServer()
```
markmandel, line 835: I think to simplify things, we should be able to use `GetCounter()` to get what should be the latest information rather than rebuilding it here (although there is some trickiness with the locking - we may need a private `getCount()` function that doesn't have the lock, and make the calling function invoke it locked.)

Then you know that from `GetCounter()` you will get whatever the latest view is of the counts and lists, it should hopefully be easier to do validation without having to go check as much of the batching state.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +784,196 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+					}
+				}
+			}
+		}
+		gsCopy.Status.Counters[name] = counter
+		gs, err = s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+		s.recorder.Event(gs, corev1.EventTypeNormal, "UpdateCounterBatchRequest",
```
markmandel, line 970: Nit:
```suggestion
		s.recorder.Event(gs, corev1.EventTypeNormal, "UpdateCounter",
```

Expect you were saving it for later,  but we could also do an additional Warning event if the count gets floored or ceiling-ed at 0 or `Capacity` .

markmandel (commented):
Looks good! Dropped a bunch of comments mostly around how we can avoid some race conditions as part of the process of batching and keeping an in memory cache, and hopefully some help with some simplification.

pkg/sdkserver/sdkserver.go
```diff
@@ -69,6 +71,23 @@ var (
...
 
+type counterUpdateRequest struct {
+	// Tracks CountIncrement and / or CountDecrement Requests
+	diffList []int64
+	// Counter (as retreived during first request)
+	counter agonesv1.CounterStatus
+	// Tracks if there have been any CountSet requests
+	countSet bool
+	// Tracks if there have been any CapacitySet requests
+	capacitySet bool
```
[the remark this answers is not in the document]
igooch: Ha no, the gRPC standard when wildly out the door with this change. (It would work if we were doing webhook validation :)) This was mainly from a comment months ago on not doing nilable pointers, but having nilable pointers does make sense in the context.

pkg/sdkserver/sdkserver.go
```diff
@@ -69,6 +71,23 @@ var (
 	_ beta.SDKServer  = &SDKServer{}
 )
 
+type counterUpdateRequest struct {
+	// Tracks CountIncrement and / or CountDecrement Requests
+	diffList []int64
```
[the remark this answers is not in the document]
igooch: Ah yep, not sure what I was thinking here, a single value makes way more sense.

pkg/sdkserver/sdkserver.go
igooch: And what if the diff can't be applied to the newest version of the gameserver? It implies we'd have to do updateCounter validation on every Get request.

pkg/sdkserver/sdkserver.go
igooch: Right now I'm using `counter agonesv1.CounterStatus` as the tracker of the Counter state. The Generation will change for any change to the Gameserver, regardless if it applies to a particular Counter we want to update. If we track the original Counter we can do a cmp.Equal to see if there any relevant changes to a Counter even if the Generation has changed.

pkg/sdkserver/sdkserver.go
igooch: We'll have to update this code regardless given that the counterUpdateRequest will have possible nil values for count and capacity (since with the code changes those fields will represent countSet and capacitySet only).

I'm leaning towards not grabbing changes from the SDK server at all and just getting the GameServer, otherwise we'd need to do update validation in the GetCounter request.

pkg/sdkserver/sdkserver.go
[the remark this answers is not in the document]
markmandel: We don't need to reject individual operations - all we need to do is Floor against `0`, and Ceiling against `Capacity`.

So if the single diff goes over in either direction at update time (or at GetCounter()` time, it's a check against these top of bottom values.

[Link to design](https://github.com/googleforgames/agones/issues/2716#:~:text=Note%3A%20A%20potential%20race%20condition%20here%20is%20that%20if%20count%20values%20are%20set%20from%20both%20the%20SDK%20and%20through%20the%20K8s%20API%20(Allocation%20or%20otherwise)%2C%20since%20the%20SDK%20append%20operation%20back%20to%20the%20CRD%20value%20is%20batched%20asynchronous%20any%20value%20incremented%20past%20the%20capacity%20will%20get%20silently%20truncated.)

pkg/sdkserver/sdkserver.go
markmandel: So the race condition we currently have, that I want to mitigate is:

1. We have a batch, and then we apply it.
2. We clear `gsCounterUpdates`
3. s.gameServer() is still returning the old version (it hasn't yet come back up to the Informer cache in the sdkserver).
4. Someone does a `GetCounter()`, which will return the old values not including the pending batch.

As far as I'm aware, the Informer cache doesn't immediately update if it sees a local update (although I could be wrong, but I'd be surprised) -- so I wanted to make sure we had something that would handle that period of time where it could be in flux - that's why I was suggesting https://github.com/googleforgames/agones/pull/3322/files#r1302253933 with something like `gsProcessedCounterUpdates`.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +792,190 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
 	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
 		return nil, errors.Errorf("%s not enabled", runtime.FeatureCountsAndLists)
 	}
-	// TODO(#2716): Implement me
-	return nil, errors.Errorf("Unimplemented -- GetCounter coming soon")
+
+	s.logger.WithField("name", in.Name).Info("Getting Counter")
+
+	gs, err := s.gameServer()
```
markmandel, line 798: From conversations we had today, this is the logic I would propose for `GetCounter()`:

1. Grab `s.gameServer()` and all it's Count information.
2. Check a local copy of what was last successfully updated to K8s for count intformation. If it's `ObjectMeta.Generation` value is greater than that of `s.gameServer()` use that information instead.
3. If there is any batch data waiting, then apply that batch data to what you got from (1) and (2).
4. We now should have the most up to date count information, and capacity information we could possibly get.
5. If the count value is less than 0, make it zero
6. If the count value is greater than capacity, make it the capacity.
7. Return the result.

Finally, this logic (maybe in a separate private function without locking, so each public function can apply it's own locking) can then be reused in `UpdateCounter()` to determine if we'll be letting the request the user has sent into the batch queue or not.

igooch: For step 2 I added this in the s.gameServer() logic. Are you OK with putting it there, or prefer not to change that method?

markmandel: That's a _really_ good question. I'm fairly sure that that is actually going to be fine - since we have the `generation` check.

it would also mean that if we want any other SDK feature to be update-to-date immediately on `s.gameServerGetter.GameServers.(..).Update(ctx, gsCopy, ...)`, it's super easy to do (update the local version we have), and those that we don't care about (which is most other functions), if you don't update the local version, it's fine.

So let's run with it, I think it will work, and the abstraction could be useful down the line.

pkg/sdkserver/sdkserver.go
igooch: Using gsCopy instead (see conversation in GetCounter).

pkg/sdkserver/sdkserver.go
igooch: Using gsCopy to track Generation changes (see GetCounter conversation).

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +790,202 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+			newCnt := counter.Count + ctrReq.diff
+			if newCnt < 0 {
+				// If Count < 0 truncate requests by setting the Count to 0.
+				newCnt = 0
+				s.logger.Debug("truncating Count in Update Counter request to 0")
+			}
+			// TODO: This means there may be increment requests in the collapsed Diff that shouldn't have gone through. We may want a diffList for this case?
+			// If Count > Capacity and the Diff is decreasing the Count, reduce the Count by the Diff.
+			// Otherwise if Count > Capacity and the Diff is increasing the Count, truncate at Capacity.
+			if newCnt > counter.Capacity && ctrReq.diff > 0 {
```
igooch, line 969: This means there may be increment requests in the collapsed Diff that shouldn't have gone through. We may want a diffList for this case?

pkg/sdkserver/sdkserver_test.go
```diff
@@ -1084,6 +1086,377 @@ func TestSDKServerReserveTimeout(t *testing.T) {
...
+
+			go func() {
+				err = sc.Run(ctx)
+				assert.NoError(t, err)
+				wg.Done()
+			}()
+
+			// check initial value comes through
+			// async, so check after a period
+			err = wait.Poll(time.Second, 10*time.Second, func() (bool, error) {
```
igooch, line 1276: Any suggestions for speeding up this test? wait.Poll is slowing the test down.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +790,202 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+	s.logger.WithField("name", in.Name).Info("Getting Counter")
+
+	gs, err := s.gameServer()
+	if err != nil {
+		return nil, err
+	}
+
+	counter, ok := gs.Status.Counters[in.Name]
+	if !ok {
+		return nil, errors.Errorf("NOT_FOUND. %s Counter not found", in.Name)
```
markmandel, line 803: This error formatting (and the rest in this file) isn't the Google style guide for go. See:
https://google.github.io/styleguide/go/decisions.html#error-strings

```suggestion
		return nil, errors.Errorf("count not found: %s", in.Name)
```

The errors package includes stacktracking, so it does report where errors come from.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +790,202 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+			// If Count > Capacity and the Diff is decreasing the Count, reduce the Count by the Diff.
+			// Otherwise if Count > Capacity and the Diff is increasing the Count, truncate at Capacity.
+			if newCnt > counter.Capacity && ctrReq.diff > 0 {
+				newCnt = counter.Capacity
+				s.logger.Debug("truncating Count in Update Counter request to Capacity")
+			}
+			counter.Count = newCnt
+		}
+		gsCopy.Status.Counters[name] = counter
+		gs, err = s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})
```
markmandel, line 976: If I'm reading this right, are we doing an update per count key that is batched? 

If so, we should instead apply all batch requests into a single update - we shouldn't be doing _n_ updates, where _n_ is the number of keys waiting to be batched.

I think we can still do multiple events after the fact, but we should consolidate updates into a single `s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})` operation.

(Unless I'm reading the formatting of this wrong).

Then we can clear the entire `s.gsCounterUpdates` on completion, which should also mean you can remove the entire `if cmp.Equal(counter, ctrReq.counter)` section, since the key will either be there or not.

pkg/sdkserver/sdkserver_test.go
markmandel: could instead use `require.Eventually()` (which we've been slowly moving over to), which I think won't wait an initial period either.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +790,202 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+		// We didn't find the Counter named key in the gameserver.
+		if !ok {
+			return nil, errors.Errorf("NOT_FOUND. %s Counter not found", name)
+		}
+
+		batchCounter.counter = *counter.DeepCopy()
+	}
+
+	switch {
+	case in.CounterUpdateRequest.CountDiff != 0: // Update based on if Client call is CountIncrement or CountDecrement
```
markmandel, line 870: Stepped through this logic in my head 😄 looks good!

pkg/sdkserver/sdkserver.go
markmandel, line 969: I think this can be simpler: If the count ends up below 0, then make it 0, if the count ends up above capacity, then make it capacity - same as the logic we have in `SetCounter` 👍🏻

markmandel (commented):
Some cleanup, but this is looking good 👍🏻

pkg/sdkserver/sdkserver.go
gongmax: > I think we can still do multiple events after the fact

What's the downside to do only one event for the single consolidated update?

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +790,202 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+
+	// Queue up the Update for later batch processing by updateCounters.
+	s.workerqueue.Enqueue(cache.ExplicitKey(updateCounters))
+	return &alpha.Counter{}, nil
+}
+
+// updateCounter updates the Counters in the GameServer's Status with the batched update requests.
+func (s *SDKServer) updateCounter(ctx context.Context) error {
+	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		return errors.Errorf("%s not enabled", runtime.FeatureCountsAndLists)
```
gongmax, line 919: A general question about the error handling: will the workerqueue retry on this kind of error? I think it should not because it is not possible to retry successfully, but I'm not sure what's the current behavior.

markmandel: That is a good point - yes it will. So we should be sure that when we throw an error here it's because we want to retry.

pkg/sdkserver/sdkserver.go
markmandel: It's more of a readability issue when looking at the event stream. If we have 5 keys in a batch, it's probably easier to read an event per key after the fact, than try and smoosh all the info from 5 keys into a single event.

(To also be clear, I was thinking one event per key).

pkg/sdkserver/sdkserver.go
igooch: I can probably take this out -- the SDK should reject any of these errors in UpdateCounter before it gets put in the worker queue.

markmandel: Either that or log the error, and return nil, so it gets dropped immediately.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +790,202 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
 	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
 		return nil, errors.Errorf("%s not enabled", runtime.FeatureCountsAndLists)
 	}
-	// TODO(#2716): Implement me
-	return nil, errors.Errorf("Unimplemented -- GetCounter coming soon")
+
+	s.logger.WithField("name", in.Name).Info("Getting Counter")
```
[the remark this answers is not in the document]
igooch: Yep, I'll change that.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +790,202 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+		}
+		if counterUpdate.countSet != nil {
+			protoCounter.Count = *counterUpdate.countSet
+		}
+		protoCounter.Count += counterUpdate.diff
+		if protoCounter.Count < 0 {
+			protoCounter.Count = 0
+			s.logger.Debug("truncating Count in Get Counter request to 0")
+		}
+		if (counterUpdate.capacitySet == nil) && (protoCounter.Count > protoCounter.Capacity) {
```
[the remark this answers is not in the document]
igooch: This is for the case where the customer is increasing the Count, and they have not changed Capacity, so the request to increase Count should be truncated.
This is to differentiate from requests where the customer sets the Capacity below where the Count is currently, so we do not change the Count even though it is now higher than Capacity.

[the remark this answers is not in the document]
markmandel: > This is to differentiate from requests where the customer sets the Capacity below where the Count is currently, so we do not change the Count even though it is now higher than Capacity.

My feeling on this is that we should keep things as simple as possible: So, we truncate on capacity always, regardless - much easier to document, and much easier to reconcile by the end user.

WDYT?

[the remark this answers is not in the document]
igooch: Maybe? Depends on how customers want to use it. Truncating in all cases makes sense for Lists though, so I suppose we could have that same behavior here too to keep it consistent between Counts / Lists.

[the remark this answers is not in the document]
markmandel: > Truncating in all cases makes sense for Lists though

Actually that's a really good point, and makes me think we should definitely be consistent truncating across both.

Also the code is simpler 😄

pkg/sdkserver/sdkserver.go
igooch: Yes, there's an update for each key (each Counter), not an update for the whole Counters map, the disadvantage of that being obvious. The advantage is that you get an accurate per-Counter updating and event log, and updating one Counter doesn't depend on another.

I can move to updating the whole Counters map at once instead of per Counter. I would just need to make sure the event recording still happened after the update even if it's per key, so I'd probably pull that out into a separate method.

pkg/sdkserver/sdkserver.go
igooch: This one's following the https://developers.google.com/actions-center/reference/grpc-api/status_codes :) I'll change it to match the rest of the file.

markmandel: I am wondering of this might give you what you want: https://pkg.go.dev/google.golang.org/grpc/status

pkg/sdkserver/sdkserver.go
markmandel: If you do one Update for all counters - as long as you apply the Events after the fact, it's pretty much fire and forget, so I don't expect it to be an issue.

pkg/sdkserver/sdkserver.go
igooch: Yes? Although I vaguely remember a conversation about not using a different error package (can't find where we had that though). It would be `status.Errorf(codes.NotFound, "%s Counter not found", in.Name)` instead of the `errors.Errorf`. Do we have any preference on keeping the current errors package and format, or moving the CountsAndLists to the grpc status errors?

markmandel: You ask a very good question - we've not done gRPC status values anywhere else in the codebase.

Although maybe should for gRPC endpoints? And this could be the beginning of the efforts. I feel I'm the worst at sticking to API standards 😄 , so WDYT? @zmerlynn @gongmax thoughts?

pkg/sdkserver/sdkserver.go
```diff
@@ -409,7 +430,13 @@ func (s *SDKServer) gameServer() (*agonesv1.GameServer, error) {
 	// they will block here until it's ready
 	s.gsWaitForSync.Wait()
 	gs, err := s.gameServerLister.GameServers(s.namespace).Get(s.gameServerName)
-	return gs, errors.Wrapf(err, "could not retrieve GameServer %s/%s", s.namespace, s.gameServerName)
+	if err != nil {
+		return gs, errors.Wrapf(err, "could not retrieve GameServer %s/%s", s.namespace, s.gameServerName)
+	}
+	if s.gsCopy != nil && gs.ObjectMeta.Generation < s.gsCopy.Generation {
```
markmandel, line 436: We probably need a read lock here as well, yeah? Since we're reading `s.gsCopy`

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +789,190 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+	s.logger.WithField("name", in.CounterUpdateRequest.Name).Debug("Update Counter Request")
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+
+	// Check if we already have a batch request started for this Counter. If not, add new request to
+	// the gsCounterUpdates map.
+	name := in.CounterUpdateRequest.Name
+	batchCounter, batchOk := s.gsCounterUpdates[name]
+
+	if !batchOk {
```
markmandel, line 856:
```suggestion
```

I'm thinking we should get rid of this outer if statement and always get the GameServer, and update the underlying `counter` values. That way _if_ there is a value that comes through from an allocation while a batch is yet to be processed, it has a chance to get captured, whereas this way it will get ignored, and if it's the same as always then it's a no-op (and it's an in-memory cache, so it doesn't matter).

I think that will remove a potential race condition.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +789,190 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+
+	s.gsCounterUpdates[name] = batchCounter
+
+	// Queue up the Update for later batch processing by updateCounters.
+	s.workerqueue.Enqueue(cache.ExplicitKey(updateCounters))
+	return &alpha.Counter{}, nil
+}
+
+// updateCounter updates the Counters in the GameServer's Status with the batched update requests.
+func (s *SDKServer) updateCounter(ctx context.Context) error {
```
markmandel, line 918: Logic on this looks good to me!

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +789,190 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+
+	gs, err = s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Record an event per update Counter
+	for _, name := range names {
+		s.recorder.Event(gs, corev1.EventTypeNormal, "UpdateCounter",
+			fmt.Sprintf("Counter %s updated", name))
```
markmandel, line 967: I really liked what you had with the details of what was updated for each counter in the event - it is possible to still capture that here?

pkg/sdkserver/sdkserver.go
zmerlynn: @igooch I think you and I previously had a review back-and-forth over just staying consistent. I think consistency of error return types from the API wins over API guidelines. (But I haven't revewed this exact one yet, just stating my general feeling.)

pkg/sdkserver/sdkserver.go
igooch: Yep, I can add that back in.

pkg/sdkserver/sdkserver.go
igooch: Sure, we can do that.

pkg/sdkserver/sdkserver.go
igooch: Oh yep good point, I'll add that in.

pkg/sdkserver/sdkserver.go
igooch: So are we leaning towards Mark's original suggestion of `return nil, errors.Errorf("count not found: %s", in.Name)`?

markmandel: Honestly, I don't have strong opinions either way (original with lower case or use gRPC errors).

I've not seen any major downside to using `return nil, errors.Errorf("count not found: %s", in.Name)?` with gRPC though - but very happy to hear dissenting opinions.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +792,190 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+	}
+
+	return &protoCounter, nil
 }
 
-// UpdateCounter returns the updated Counter. Returns NOT_FOUND if the Counter does not exist.
-// Returns INVALID_ARGUMENT if the FieldMask paths are invalid.
+// UpdateCounter collapses all UpdateCounterRequests for a given Counter into a single request.
+// Returns NOT_FOUND if the Counter does not exist (name cannot be updated).
+// Returns OUT_OF_RANGE if the Count is out of range [0,Capacity].
```
markmandel, line 838: Small nit: need to fixup the errors in the docs.

pkg/sdkserver/sdkserver.go
```diff
@@ -762,20 +792,190 @@ func (s *SDKServer) GetCounter(ctx context.Context, in *alpha.GetCounterRequest)
...
+		names = append(names, name)
+	}
+
+	gs, err = s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Record an event per update Counter
+	for _, name := range names {
```
markmandel, line 967: Love it!

markmandel (approved):
Saw one nit in the method comments, but approving so you can merge when ready. Nice work!

markmandel:
_sigh_

```
   gameserver_test.go:953: 
        	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:953
        	Error:      	Not equal: 
        	            	expected: "spec.template.spec.containers[0].resources.requests[cpu]"
        	            	actual  : "spec.template.spec.containers[0].resources.requests"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-spec.template.spec.containers[0].resources.requests[cpu]
        	            	+spec.template.spec.containers[0].resources.requests
        	Test:       	TestGameServerResourceValidation
```

markmandel:
Sorry - should have been clear - the _sigh_ was because the flake is showing up. This PR is great 😄 

* https://github.com/googleforgames/agones/issues/3362

PR #3330, opened 2023-08-17
Added TF DNS config options to prevent Autopilot destroy / create on existing cluster

abmarcum:
**What type of PR is this?**
/kind bug

Prevents re-creating Autopilot clusters if they already exist. 


markmandel:
Deferring to @zmerlynn and @gongmax who have been doing more of the autopilot work than I have.

zmerlynn:
@abmarcum Thanks for the contribution! Can you describe the problem in more detail? Is our TF config recreating DNS settings? This seems like a flaw in the Google TF module, if so - if not, could you explain why not? Thanks!

zmerlynn:
I see the issue on my desktop now as well. LGTM.

zmerlynn (approved):

PR #3310, opened 2023-08-02
Verify gen-all-sdk-grpc

Kalaiselvi84:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**What this PR does / Why we need it**:

**Which issue(s) this PR fixes**:

Closes #3303

**Special notes for your reviewer**:

PR #3337, opened 2023-08-21
Catch up C++ SDK to `make gen-all-sdk-grpc`

zmerlynn:
We need to keep an eye on what changed here so we can understand if we need to make this hermetic, but with this diff fixed we should be able to submit #3310.


markmandel (approved):
Assuming it passes 😃🤞🏻

PR #3340, opened 2023-08-22
SDK Conformance: Use -test consistently instead of -no-build

zmerlynn:
Not clear why these were using -no-build. Using -test gives us retries and nice reporting.

Helping #3310


markmandel (approved):
Compared times with another build, and both are around 4m. Let's do it.

PR #3343, opened 2023-08-22
Do not refresh cache if no update

gongmax:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

In the UpdateFunc, check whether there are actual change for the GameServer by checking the resourceVersion, and skip the allocation cache refresh for the GameServer if so. This prevents the race condition where a GS is Allocated and removed from the cache but the UpdateFunc adds it back with a stale state.

Closes #


pkg/gameserverallocations/allocation_cache.go
```diff
@@ -70,6 +70,9 @@ func NewAllocationCache(informer informerv1.GameServerInformer, counter *gameser
 			if !ok {
 				return
 			}
+			if oldGs.ObjectMeta.ResourceVersion == newGs.ObjectMeta.ResourceVersion {
```
markmandel, line 73: I like it! Good find!

Worth a unit test, or nah?

gongmax: Oh, I will add an unit test.

markmandel (approved):

PR #3345, opened 2023-08-24
chore: k8s.io/kubernetes/.generated_files

aimuz:

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Ignore generation of files

Closes #


aimuz:
[google-oss-prow](https://github.com/apps/google-oss-prow) Won't read the .generated_files ?

markmandel (commented):
Haven't seen this before - but one thought, not sure how much we want to be scrubbing OSS licence files from dependencies we're vendoring. Seems like bad form?

aimuz:
Actually, what I'm trying to achieve is that the size tag calculation ignores some of the generated files. I'm not trying to delete the vendor directory.

For example, 3342 doesn't really change much code, but it has a size/XXL tag.

If the vendor directory is ignored, then the label might be size/S

This is supported by prow on k8s.

markmandel:
OIC! @roberthbailey you know PROW - any thoughts?

.gitattributes
```diff
@@ -14,3 +14,7 @@
 
 # See https://github.com/golang/go/issues/9281
 * -text
+
+vendor/** linguist-generated
```
roberthbailey, line 18: This line and line 20 seem redundant. Should this one be removed?

.gitattributes
```diff
@@ -14,3 +14,7 @@
...
+vendor/ linguist-generated
```
roberthbailey, line 19: I think this needs to have `linguist-generated=true` according to https://github.com/kubernetes/test-infra/blob/master/prow/gitattributes/gitattributes.go#L89

roberthbailey (commented):
I'm not sure I understand how all of the file deletions in vendor are related to the change to have prow generate change sizes differently. 

If that is cleaning up dependencies, can we separate it into a different PR?

aimuz:
It looks like setting linguist-generated=true won't work either，Are you currently using the prow version of k8s, or is there another version within your Google?

zmerlynn:
FWIW I don't think we should take any prow config changes as I've been kind of arguing that we should just drop it. We're not using any of the features and at this point it's kind of noisy.

markmandel:
> FWIW I don't think we should take any prow config changes as I've been kind of arguing that we should just drop it. We're not using any of the features and at this point it's kind of noisy.

+1 from me.

zmerlynn:
@aimuz Thanks for the submission. We decided to eliminate Prow since it wasn't doing much for us - the webhooks are now disabled, https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2131 removes the config. @markmandel put up https://github.com/googleforgames/agones/pull/3532 to implement the size labeler (using https://github.com/CodelyTV/pr-size-labeler), borrowed from Quilkin.

I'm going to close out this PR, but we may need to add a `files_to_ignore` - let's handle that later.

aimuz:
If you just use prow's size function, it's really not necessary, but you always want a retest function, so that you can quickly retry if the test fails.

PR #3348, opened 2023-08-25
Move the restapi folder GRPC_PORT off of the same port as the go folder tests

zmerlynn:
This may be causing some sort of conflict in the tests in #3310. Not sure why only #3310 is seeing it, but it's not hard to make this consistent.

(In case it's not clear on review, the problem is the second invocation where we set `GRPC_PORT` to 9001, which is the same as the go folder.)


markmandel (approved):
Good catch!

gongmax (approved):

zmerlynn:
I'm going to let this ride in with #3349 to cut down on merges.

PR #3358, opened 2023-08-31
Yaml linter

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
 
/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #


igooch (approved):

gongmax (approved):

test/load/allocation/scenario-values.yaml
```diff
@@ -20,31 +20,31 @@ agones:
         memory: 4000Mi
     affinity:
       nodeAffinity:
-        preferredDuringSchedulingIgnoredDuringExecution: null
+        preferredDuringSchedulingIgnoredDuringExecution:
```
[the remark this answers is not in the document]
zmerlynn: I believe we explicitly need to `null` it here. cc @gongmax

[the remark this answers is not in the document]
Kalaiselvi84: Included `null`

.yamllint
```diff
@@ -0,0 +1,7 @@
+extends: default
+
+rules:
+  truthy: disable
+  line-length:
+    max: 1500
+  trailing-spaces: disable
```
Kalaiselvi84, line 7: This is to ignore the `trailing-spaces` error

zmerlynn: Trailing spaces seems like something we should flag?

Kalaiselvi84: The yamllint configuration will cause it to ignore spaces at the end of lines. Can we have it?

zmerlynn: Right, I'm saying: I think we should lint that there are no trailing spaces (unless it is prohibitive for some reason).

Kalaiselvi84: excluded it from this config file

.yamllint
```diff
@@ -4,4 +4,3 @@ rules:
   truthy: disable
```
zmerlynn, line 4: Do we have a lot of violations of this?

Kalaiselvi84: All of our GitHub Actions workflow scripts use the `on` keyword, and we cannot replace it with `true`. Therefore, I've disabled that specific check.

zmerlynn: Because the GitHub Actions parser is looking for `on` and not `true`? Huh.

Kalaiselvi84: That's right..

build/Makefile
```diff
@@ -489,7 +489,7 @@ build-extensions-binary-linux-arm64: $(ensure-build-image)
 lint: LINT_TIMEOUT ?= 15m
 lint: $(ensure-build-image)
 	docker run -t -e "TERM=xterm-256color" -e "$(gomod_on)" --rm $(common_mounts) -w $(workdir_path) $(DOCKER_RUN_ARGS) $(build_tag) bash -c \
-		"golangci-lint run ./examples/... && golangci-lint run --timeout $(LINT_TIMEOUT) ./..."
+		"golangci-lint run ./examples/... && golangci-lint run --timeout $(LINT_TIMEOUT) ./... && find ../ -name '*.yml' -o -name '*.yaml' | grep -v './install/' | xargs yamllint"
```
zmerlynn: two things here:
1. Are you sure about `..`? I think you're at the top of the repo here
2. I think we should prefer exclusions using the yamllint config itself rather than `find`. You can just use `yamllint .` and then [ignore paths in a `.gitignore` style way](https://yamllint.readthedocs.io/en/stable/configuration.html#ignoring-paths). That allows people to just run `yamllint` outside docker or the Makefile and keep the config the same. I think it would just look roughly like:
```
ignore:
  - /install/
```

I wouldn't bother configuring the yaml suffixes, it looks like the default is appropriate.

PR #3370, opened 2023-09-10
Outputting the private IP of GKE cluster.

canuc:
**What type of PR is this?**
>
/kind feature

Adding private host to the output, necessary for people to communicate with the k8s api from within the same VPC on GCP.

Closes #


canuc:
Closing as I found an issue with public clusters that causes an error.

PR #3375, opened 2023-09-13
handle static port policy

### LINKED ISSUE #2314: TCPUDP not supported

ashutosji:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**What this PR does / Why we need it**: This PR is a fix for static portPolicy.

**Which issue(s) this PR fixes**: https://github.com/googleforgames/agones/issues/2314

Closes #2314 


pkg/apis/agones/v1/gameserver.go
```diff
@@ -715,18 +715,33 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
...
+				Protocol:      p.Protocol,
+			}
+			err := gs.ApplyToPodContainer(pod, *p.Container, func(c corev1.Container) corev1.Container {
+				c.Ports = append(c.Ports, cp)
+
+				return c
+			})
+			if err != nil {
+				return nil, err
+			}
```
markmandel, line 744: So  my suggestion would be to create a unit test to make it easier to test if the output of this function matches what you expect.

Here is an example: https://github.com/googleforgames/agones/blob/30f247dd3e276fe14fe87ed43efcc639ed839902/pkg/apis/agones/v1/gameserver_test.go#L1240

ashutosji: Thanks! I added the test cases.

markmandel (commented):
This looks like a good start!

Next step, let's create an e2e test for it! You can likely copy some existing examples for a lot of it - let me know if you would like me to point to some.

ashutosji:
> This looks like a good start!
> 
> Next step, let's create an e2e test for it! You can likely copy some existing examples for a lot of it - let me know if you would like me to point to some.

Sure! I will create e2e test for it. Correct me if I am wrong, I think you are talking about this existing examples https://github.com/googleforgames/agones/blob/main/test/e2e/framework/framework.go#L523. Am I on right track?

markmandel:
Yes, but I'm talking about using something like: 
https://github.com/googleforgames/agones/blob/18163e820cc567c13fd71ae444e63b6986042c53/test/e2e/gameserver_test.go#L862

And doing a version with a Static port.

One thing to note - you will probably have to edit your firewall configurations to allow TCP connections to have the test pass for you locally (by default it was off).

test/e2e/gameserver_test.go
```diff
@@ -907,6 +907,107 @@ func TestGameServerTcpUdpProtocol(t *testing.T) {
...
+	gs.Spec.Ports[0].Name = "gameserver"
+	gs.Spec.Template.Spec.Containers[0].Env = []corev1.EnvVar{{Name: "TCP", Value: "TRUE"}}
+
+	errs := gs.Validate(agtesting.FakeAPIHooks{})
+	require.Len(t, errs, 0)
+
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(t, framework.Namespace, gs)
+	if err != nil {
+		assert.FailNow(t, "Could not get a GameServer ready", err.Error())
+	}
```
markmandel, line 927: Nit:

```suggestion
  require.NoError(t, err)
```

Same for below.

markmandel:
Bringing offline conversation back to PR.

I see why you are getting the error message for `ready GameServer instance has 1 port(s), want 2` on this check here:

https://github.com/googleforgames/agones/blob/98bd2c5af7f0de512f0939c178f7867217096058/test/e2e/framework/framework.go#L232-L243

For the Dynamic Port Allocation strategy, it actually edits the backing `GameServer`'s ports, which I had forgotten it actually did!

https://github.com/googleforgames/agones/blob/cd85fa0af8f4e91b35fc0d6af618dcdb7300812e/pkg/portallocator/portallocator.go#L186-L198

Whereas what we are doing with this implementation is returning the appropriate `Pod` declaration.

So we have two options:

1) Rather than editing `GameServer.Pod()`, we instead update the GameServer record itself in the `gameserver/controller`, to match the type of change we make in the `PortAllocator` for `Dyanamic` and `Passthrough` port configurations.

The appropriate place does seem to be here:
https://github.com/googleforgames/agones/blob/cd85fa0af8f4e91b35fc0d6af618dcdb7300812e/pkg/gameservers/controller.go#L493-L504

Before creating the Pod, and then when it Updates the GameServer those new port details will get updated along with them.

2) Edit the `CreateGameServerAndWaitUntilReady` so that if the Port has a Static policy, we don't increment `expectedPortCount`, so we ignore the check for this particular use case.

I'm leaning towards No. 1, since that brings this implementation inline with what the PortAllocator does, and lowers the number of testing paths we need to manage (and special cases in the tests) - but what are your thoughts?

ashutosji:
> Bringing offline conversation back to PR.
> 
> I see why you are getting the error message for `ready GameServer instance has 1 port(s), want 2` on this check here:
> 
> https://github.com/googleforgames/agones/blob/98bd2c5af7f0de512f0939c178f7867217096058/test/e2e/framework/framework.go#L232-L243
> 
> For the Dynamic Port Allocation strategy, it actually edits the backing `GameServer`'s ports, which I had forgotten it actually did!
> 
> https://github.com/googleforgames/agones/blob/cd85fa0af8f4e91b35fc0d6af618dcdb7300812e/pkg/portallocator/portallocator.go#L186-L198
> 
> Whereas what we are doing with this implementation is returning the appropriate `Pod` declaration.
> 
> So we have two options:
> 
> 1. Rather than editing `GameServer.Pod()`, we instead update the GameServer record itself in the `gameserver/controller`, to match the type of change we make in the `PortAllocator` for `Dyanamic` and `Passthrough` port configurations.
> 
> The appropriate place does seem to be here:
> 
> https://github.com/googleforgames/agones/blob/cd85fa0af8f4e91b35fc0d6af618dcdb7300812e/pkg/gameservers/controller.go#L493-L504
> 
> Before creating the Pod, and then when it Updates the GameServer those new port details will get updated along with them.
> 
> 2. Edit the `CreateGameServerAndWaitUntilReady` so that if the Port has a Static policy, we don't increment `expectedPortCount`, so we ignore the check for this particular use case.
> 
> I'm leaning towards No. 1, since that brings this implementation inline with what the PortAllocator does, and lowers the number of testing paths we need to manage (and special cases in the tests) - but what are your thoughts?

Yes, You are right! Rather than writing and handling number of test cases we should proceed with the case1.
```
if gs.Spec.Ports[0].PortPolicy == agonesv1.Static && gs.Spec.Ports[0].Protocol == agonesv1.ProtocolTCPUDP {
			gs.Spec.Ports = []agonesv1.GameServerPort{
				{
					Name:          "tcp-port",
					ContainerPort: gs.Spec.Ports[0].ContainerPort,
					HostPort:      gs.Spec.Ports[0].HostPort,
					Protocol:      corev1.ProtocolTCP,
				},
				{
					Name:          "udp-port",
					ContainerPort: gs.Spec.Ports[0].ContainerPort,
					HostPort:      gs.Spec.Ports[0].HostPort,
					Protocol:      corev1.ProtocolUDP,
				},
			}

		}
		logrus.WithField("After", gs).Debug("Made Static PortPolicy changes")
          gs, err = c.createGameServerPod(ctx, gs)
```
However, I tried and added the above piece of code before Creating the pod here: https://github.com/googleforgames/agones/blob/cd85fa0af8f4e91b35fc0d6af618dcdb7300812e/pkg/gameservers/controller.go#L506 but somehow it's throwing error `Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)`. And it is throwing the error after this log `logrus.WithField("After", gs).Debug("Made Static PortPolicy changes")`. I have not pushed the code because I am still working on it.

markmandel:
> but somehow it's throwing error Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)

Hard to debug without seeing the code 😃 so if you get really stuck, please upload the full log and the code.

markmandel:
Now we have the nodejs fix in place, updating so we can get a clean build and test 👍🏻

ashutosji:
> Now we have the nodejs fix in place, updating so we can get a clean build and test 👍🏻

I am not getting why Static test cases are failing in CI/CD build. They are working in local as expected.
Also, the other test cases that i haven't touched are also failing in CI/CD.

FYI, This test case https://github.com/googleforgames/agones/blob/main/test/e2e/gameserver_test.go#L978 I have not touched and it's passing in local but failed in CI/CD.
https://screenshot.googleplex.com/4Cnvg55Pueq3WFF

Could you help me with this?

markmandel (commented):
Added some notes below. I agree some of these test failures are weird.

It's potentially possible that since you hardcoded `gs.Spec.Ports[0]` that may cause a panic in places where GameServers may not have ports in some capacity - which could cause various components to restart  -hence the cascading failures.

I updated this branch to `main` but also see my recommendation on using a `range` rather than hardcoding.

pkg/gameservers/controller.go
```diff
@@ -503,6 +503,23 @@ func (c *Controller) syncGameServerCreatingState(ctx context.Context, gs *agones
 	// Maybe something went wrong, and the pod was created, but the state was never moved to Starting, so let's check
 	_, err := c.gameServerPod(gs)
 	if k8serrors.IsNotFound(err) {
+		name := gs.Spec.Ports[0].Name
```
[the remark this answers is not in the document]
ashutosji: Previously, I had done it in that way only. The reason I did it in this way:
`gs.Spec.Ports[0].Name = "default"`
For TCP, I added `"-tcp"` and it become `"default-tcp"`.
So, When I was doing for `"-udp"` it was becoming `"default-tcp-udp"`.
No issue, I will correct it. Thanks!

pkg/gameservers/controller.go
```diff
@@ -503,6 +503,23 @@ func (c *Controller) syncGameServerCreatingState(ctx context.Context, gs *agones
...
+
+		if gs.Spec.Ports[0].PortPolicy == agonesv1.Static && gs.Spec.Ports[0].Protocol == agonesv1.ProtocolTCPUDP {
```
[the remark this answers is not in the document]
ashutosji: Yes, You are right! Thank you so much for guidance. I will correct it.

pkg/gameservers/controller.go
```diff
@@ -503,6 +503,24 @@ func (c *Controller) syncGameServerCreatingState(ctx context.Context, gs *agones
 	// Maybe something went wrong, and the pod was created, but the state was never moved to Starting, so let's check
 	_, err := c.gameServerPod(gs)
 	if k8serrors.IsNotFound(err) {
+
+		for i := range gs.Spec.Ports {
+			if gs.Spec.Ports[0].PortPolicy == agonesv1.Static && gs.Spec.Ports[0].Protocol == agonesv1.ProtocolTCPUDP {
```
markmandel, line 508:
```suggestion
			if gs.Spec.Ports[i].PortPolicy == agonesv1.Static && gs.Spec.Ports[i].Protocol == agonesv1.ProtocolTCPUDP {
```

markmandel (commented):
Small fix, but otherwise looks good 👍🏻

pkg/gameservers/controller.go
ashutosji: Yes, I missed it. Thanks for pointing it.

test/e2e/gameserver_test.go
markmandel: Bumping this final nit (and below), then this is good to merge.

markmandel: I think I accidentally resolved this yesterday when I didn't mean to.

if we can convert all the 

```
	if err != nil {
		assert.FailNow(t, ...)
	}
```

Statements into `require.NoError(t, err)` that would be the last thing needed before this is good to merge.

markmandel:
Bumping this comment: https://github.com/googleforgames/agones/pull/3375/files#r1379210777 as the only thing still required to get this PR completed.

ashutosji:
> Bumping this comment: https://github.com/googleforgames/agones/pull/3375/files#r1379210777 as the only thing still required to get this PR completed.

I did the changes. But, Somehow build is failing at step-20. I am not getting why it's failing because it's not related to my code changes. Could you please look into this?

markmandel (approved):
Looks good. #3505 should fix the flake I believe, so once that merges, this will go in next.

Issue #2842, opened 2022-12-02
Should we add `govulncheck` to CI?


markmandel:
Though I guess that would make it non-hermetic.

_Originally posted by @zmerlynn in https://github.com/googleforgames/agones/pull/2841#issuecomment-1335296321_

Looks kind this came out in September: https://go.dev/blog/vuln, which is awesome - I was looking for a tool like this for ages.

I would be pro incorporating it into CI. I personally like that security issues break CI. They make it a high priority to fix it immediately, before work can continue.

CI is already non-hermetic, since we also check things like 404 links, etc.

markmandel:
One area of concern is, reading https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck

> There is no support for silencing vulnerability findings.

Sometimes there are vulnerabilities that are valid, but there are no fixes for at this time. We may want to wait until there is a mechanism for exceptions in reporting (unless we roll or own?)

aimuz:
Nice note to add govulncheck, should I implement it inside a PR?

Maybe govulncheck could scan it as a daily task, at the right time, and be able to create issues automatically

markmandel:
> Nice note to add govulncheck, should I implement it inside a PR?

As long as we can do exceptions if we need to - that would be my opinion.

One thought I had, we could do `govulncheck -json` to provide output - I wonder if we use something like `jq` to be able to provide exceptions to specific security advisories?

aimuz:
https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-supply-chain-security#what-are-dependabot-updates

markmandel:
I wrote this, and I think we don't need to do this, since we have dependabot - so closing as `wontfix`.

PR #3376, opened 2023-09-13
ci: enable dependabot

aimuz:

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Ref: https://github.com/googleforgames/agones/issues/2842

Checking for updates with dependabot

https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"


aimuz:
Needs to be activated by the repository administrator at this link https://github.com/googleforgames/agones/settings/security_analysis

markmandel:
We've had depdendabot installed and running for a while now - not sure if this gives us anything extra? (or actually makes security scanning less often than the default?)

aimuz:
I'd like him to be able to upgrade dependencies automatically, thus avoiding the need to manually upgrade dependencies in most cases

roberthbailey:
> I'd like him to be able to upgrade dependencies automatically, thus avoiding the need to manually upgrade dependencies in most cases

In the past when I've asked dependabot to create a patch it has often been unsuccessful in doing so. 

Also, many of the security issues it finds are in the other language SDKs (e.g. nodejs) so we would need rules for the various directories in https://github.com/googleforgames/agones/tree/main/sdks to make this fully automated. 

We don't get alerts very often, so I'm wondering how much this will help our current patching process.

aimuz:
> Also, many of the security issues it finds are in the other language SDKs (e.g. nodejs) so we would need rules for the various directories in https://github.com/googleforgames/agones/tree/main/sdks to make this fully automated.

Currently only gomod and github action are configured, which should not affect sdk.

For sdk, it does require additional rules to work better

markmandel:
Just coming back to this...

> I'd like him to be able to upgrade dependencies automatically, thus avoiding the need to manually upgrade dependencies in most cases

Given we don't actually want automatic updates, does this give us anything above what GitHub provides us by default?

aimuz:
Rather than providing automatic updates for all dependencies by default, this configuration will only turn on a few of the included ones.

If there is no need for auto-updates, then this one I think will turn that off 😀

PR #3382, opened 2023-09-15
Upgrade Docsy

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #


Kalaiselvi84:
@markmandel
 I've tried three ways to upgrade Docsy,
1. Copied latest version of Docsy to site/themes, but I encountered a rendering issue - https://gist.github.com/Kalaiselvi84/86cbb64e00aa4bf3657817a09bf9867b
2. Error  received when followed the migrating to Hugo modules approach - https://gist.github.com/Kalaiselvi84/69383165776fc8d7cd450cbc770f6853
3. Successfully updated Docsy using Hugo modules, and now the `site-server, site-static, hugo server, and hugo-test` targets are passing.

Could you please share your thoughts?

markmandel:
Let's see what the preview looks like!

site/go.mod
```diff
@@ -3,3 +3,5 @@ module github.com/agones/agones/site
 go 1.20
 
 require gopkg.in/yaml.v2 v2.2.8
```
markmandel, line 5: Shouldn't there also be a change in the config.toml and a hugo.toml? - looking at https://www.docsy.dev/docs/updating/convert-site-to-module/

I feel like this is just using our existing docsy theme in the `docsy` folder?

Kalaiselvi84: Yes, I've followed the same instructions from https://www.docsy.dev/docs/updating/convert-site-to-module/
and received this error -> https://gist.github.com/Kalaiselvi84/69383165776fc8d7cd450cbc770f6853

markmandel: > Shouldn't there also be a change in the config.toml and a hugo.toml? - looking at https://www.docsy.dev/docs/updating/convert-site-to-module/
> 
> I feel like this is just using our existing docsy theme in the `docsy` folder?

Yes, but you would need to migrate first:

So you have to do https://www.docsy.dev/docs/updating/convert-site-to-module/

markmandel:
> * Copied latest version of Docsy to site/themes, but I encountered a rendering issue - https://gist.github.com/Kalaiselvi84/86cbb64e00aa4bf3657817a09bf9867b

With this approach - did you attempt to resolve the issues within the .md files? If so, where did you end up?

Kalaiselvi84:
Followed the [Migrate to hugo modules](https://www.docsy.dev/docs/updating/convert-site-to-module/) instructions. To resolve the below error, deleted the `vendor` folder from the `site` directory

```
go: can't compute 'all' using the vendor directory
	(Use -mod=mod or -mod=readonly to bypass.)
Error: failed to load modules: failed to list modules: failed to execute 'go [list -m -json all]': failed to execute binary "go" with args [list -m -json all]: go: can't compute 'all' using the vendor directory
	(Use -mod=mod or -mod=readonly to bypass.)
 *errors.errorString

```

After resolving the error, all the targets passed successfully. However, the Agones site's front page doesn't appear as expected, 
https://screenshot.googleplex.com/3bQFi6d4JsugDgX

site/go.mod
```diff
@@ -3,3 +3,5 @@ module github.com/agones/agones/site
 go 1.20
```
markmandel, line 3: Apache header please.

site/hugo.toml
```diff
@@ -17,7 +17,7 @@ title = "Agones"
 enableRobotsTXT = true
 
 # Hugo allows theme composition (and inheritance). The precedence is from left to right.
-theme = ["docsy"]
+theme = ["github.com/google/docsy", "github.com/google/docsy/dependencies"]
```
markmandel, line 20: Should we delete the `layouts/docsy` folder as well?

.github/ISSUE_TEMPLATE/kubernetes_update.md
```diff
@@ -30,7 +30,7 @@ List of items to do for upgrading to {version_1} {version_2} {version_3}
 - [ ] Regenerate Kubernetes resource includes (e.g. ObjectMeta, PodTemplateSpec)
     - [ ] Start a cluster with `make gcloud-test-cluster` (this cluster will use Kubernetes {version_2}), uninstall agones using `helm uninstall agones -n agones-system`, and then run  `make gen-embedded-openapi` and `make gen-install`
 - [ ] Update documentation for creating clusters and k8s API references to align with the above clusters versions and the k8s API version
-    - [ ] `site/config.toml`
+    - [ ] `site/hugo.toml`
```
markmandel, line 33: Is this value in the hugo.toml? Looks like it's still in config.toml?

markmandel (commented):
Seems like some CSS changes in our custom landing page.

How's you CSS? 😃

Kalaiselvi84:
> Seems like some CSS changes in our custom landing page.
> 
> How's you CSS? 😃

Please let me know the required changes. Will do it🙂

markmandel:
Sounds good - will check out the preview and see what the differences are!

markmandel:
That's a weird one:

```
	helm upgrade --install --atomic --wait --namespace=agones-system \
	--create-namespace \
	--set agones.image.tag=1.35.0-dev-8c189bb,agones.image.registry="us-docker.pkg.dev/agones-images/ci" \
	--set agones.image.controller.pullPolicy="Always",agones.image.controller.pullSecret= \
	--set agones.image.extensions.pullPolicy="Always",agones.image.allocator.pullPolicy="Always" \
	--set agones.image.ping.pullPolicy="Always",agones.image.sdk.alwaysPull=true \
	--set agones.allocator.service.serviceType="LoadBalancer" \
	--set agones.controller.logLevel="debug" \
	--set agones.featureGates="PlayerAllocationFilter=true&PlayerTracking=true&ResetMetricsOnDelete=false&PodHostname=false&SplitControllerAndExtensions=false&FleetAllocationOverflow=true&CountsAndLists=true&Example=true" \
	--set agones.allocator.service.loadBalancerIP=34.139.170.120 \
	--set agones.metrics.serviceMonitor.enabled=false \
	 \
Release "agones" does not exist. Installing it now.
Error: release agones failed, and has been uninstalled due to atomic being set: services "agones-allocator" not found
```

Not seen that before as a flake. Happened on generic-1.27

Kalaiselvi84:
I've aligned the words in the "cover.html" file. The dropdown menu for "Release" and the search box are not functioning properly. I'm working on fixing this and the alignment of all the tabs.

https://screenshot.googleplex.com/4Wbbcanx6FZZsTU

markmandel:
I'm not seeing anything in the general documentation that's broken, really it's just that header and the landing page.

https://github.com/googleforgames/agones/blob/main/site/assets/scss/_variables_project.scss is the place where you can edit and add css rules - it's quite possible those have changed between versions of Docsy - hence the issue, since the [_index.html](https://github.com/googleforgames/agones/blob/main/site/content/en/_index.html) page is totally custom.

It might be worth commenting out all the custom rules and seeing where things are, and adding them back in one by one to see if it breaks / fixes anything.

Kalaiselvi84:
> I'm not seeing anything in the general documentation that's broken, really it's just that header and the landing page.

The site-server target, which uses to check the agones site locally, is causing some confusion for me. On the local version, both the `Release` tab and the search option appear to be broken.


>https://github.com/googleforgames/agones/blob/main/site/assets/scss/_variables_project.scss is the place where you can edit and add css rules - it's quite possible those have changed between versions of Docsy - hence the issue, since the [_index.html](https://github.com/googleforgames/agones/blob/main/site/content/en/_index.html) page is totally custom.

>It might be worth commenting out all the custom rules and seeing where things are, and adding them back in one by one to see if it breaks / fixes anything.

Thank you for pointing that out. I will check and update the`_variables_project.scss` and `_index.html` files  and see how it goes.

markmandel:
> The site-server target, which uses to check the agones site locally, is causing some confusion for me. On the local version, both the Release tab and the search option appear to be broken.

Yeah, I see that too on the preview linked above (that's what I means for the "header")

Have a play, see what you can work out 👍🏻

Kalaiselvi84:
Found a bug while testing the alignment of the landing page
 https://screenshot.googleplex.com/4TczBjVhyGvehqX

markmandel:
Definitely getting closer!

Kalaiselvi84:
@markmandel 
I made changes to the cover.html file to adjust the title alignment on the landing page. However, the hyperlinks in the "Release" dropdown menu still seem to use the old format of the cover.html, and the title alignment appears fine. I'm uncertain about the source of these old HTML formats since the file has been modified.

We've listed release versions up to 1.17.0 in the "Release" dropdown menu. However, when I click on any of the older versions from the list, all the versions preceding 1.17.0 reappear in the dropdown menu. I'm unsure how these older versions are being included since they were supposed to be limited to 1.17.0.

I'm currently stuck and unsure how to address these issues.. Your insights and assistance would be greatly appreciated.🙏

markmandel:
> I made changes to the cover.html file to adjust the title alignment on the landing page. However, the hyperlinks in the "Release" dropdown menu still seem to use the old format of the cover.html, and the title alignment appears fine. 

The release dropdown looks good to me? What difference are you seeing?

The title is a bit weirdly aligned, not sure what is happening there either.

> I'm uncertain about the source of these old HTML formats since the file has been modified, and I'm also curious as to why the landing page is not functioning properly with the old format.

I expect some versions got updated along the way. Docsy uses https://getbootstrap.com/ to do a lot of the layout work, and it's possible there were breaking changes between versions.

It's also possible therefore that some of the CSS rules in https://github.com/googleforgames/agones/blob/main/site/assets/scss/_variables_project.scss break the current version of Docsy. If you remove any of those rules (it should refresh when trying with `make site-server`) does it fix any of the issues?

> We've listed release versions up to 1.17.0 in the "Release" dropdown menu. However, when I click on any of the older versions from the list, all the versions preceding 1.17.0 reappear in the dropdown menu. I'm unsure how these older versions are being included since they were supposed to be limited to 1.17.0.

That's just snapshots of older versions of the website you are viewing - which have dropdowns of the older versions, since that was what they had at the time. Keep an eye on the URL when you click thorugh.

Kalaiselvi84:
> The release dropdown looks good to me? What difference are you seeing?

In the Dev tool, the alignment is fine with the existing 'cover.html' for all versions, but it's causing an issue on the landing page.


> I expect some versions got updated along the way. Docsy uses https://getbootstrap.com/ to do a lot of the layout work, and it's possible there were breaking changes between versions.

While updating docsy, the dependency for the bootstrap is added in this `site/go.sum` https://github.com/googleforgames/agones/pull/3382/files#diff-7d8af8e6d28115e8d9161298a0e71ef7f5fe71e60a8381099e1a63abb7f0a992. Will it be fine?

> That's just snapshots of older versions of the website you are viewing - which have dropdowns of the older versions, since that was what they had at the time. Keep an eye on the URL when you click thorugh.

Understood.. a separate 'site' is being built for each version, and the corresponding versions are accessible via dropdowns on the website. Thanks!

markmandel:
This is getting really close!

Kalaiselvi84:
The current change looks good in the smaller screen. However, if we increase the gap between the tools, the bigger screen will look good, but the smaller screen will look messy. Share your thoughts please..

Kalaiselvi84:
Are we planning to keep the `docsy` folder to avoid the issue in the release process?

markmandel:
> Are we planning to keep the docsy folder to avoid the issue in the release process?

I'm not convinced that was the issue specifically.

I'd still like to keep the docsy folder as it allows us to make patches to docsy without having to first get them pushed upstream.

markmandel:
> Are we planning to keep the docsy folder to avoid the issue in the release process?

Sorry I should have asked a clarifying question: In this PR we _using_ the theme `site/themes/docsy` - I just saw it go back in, so I'm assuming that we aren't? And we're integrating it another way?

It looks like we migrated to using Hugo modules? If that's the case, then yeah - delete the docsy folder if we're not using it anymore. Sorry, I should have asked that first.

Kalaiselvi84:
> It looks like we migrated to using Hugo modules? If that's the case, then yeah - delete the docsy folder if we're not using it anymore.

Yes, migrated to Hugo modules: https://www.docsy.dev/docs/updating/convert-site-to-module/. However, the `site-static` target requires the `agones/site/themes/docsy/assets/vendor/Font-Awesome/scss` folder. Is it good to keep the `docsy` folder?

markmandel:
> However, the site-static target requires the agones/site/themes/docsy/assets/vendor/Font-Awesome/scss folder

Why does it need it? Seems like it shouldn't be required anymore?

Kalaiselvi84:
> Why does it need it? Seems like it shouldn't be required anymore?

Will delete it. This is not necessary as we are initially cleaning up the working directory in build script?

markmandel:
> Will delete it. This is not necessary as we are initially cleaning up the working directory in build script?

Well if we're not using that folder for anything for generating the website, then we don't need the folder 👍🏻 

The [build  script](https://github.com/googleforgames/agones/blob/main/build/release/pre_cloudbuild.yaml) doesn't affect any of this - the updated version uses whatever is latest from `main` as the code to run. So if the tool runs on Ci and local, it will work there.

Kalaiselvi84:
I've attempted various alignment methods for the Agones title, description, and asciinema-player on the landing page. While the alignment appears satisfactory on a laptop screen, it doesn't translate as well to larger screens.

For reference, here are the different alignment methods I've tried with `@media`: https://gist.github.com/Kalaiselvi84/3bde60bf8158b9100508095c8f0c3ab7

Based on the latest preview available at https://c28c58b-dot-preview-dot-agones-images.appspot.com/, the alignment seems acceptable on both screens.

Should we proceed with this current configuration? WDYT?

markmandel:
> Should we proceed with this current configuration? WDYT?

I think I might need to download this and see what playing around with the CSS rules will do 😃

A couple of things of note if you have some good ideas on (and so I don't forget), and are probably the biggest issues I see:

### In the header
image

The "AGONES" should be bold. This is what the current website looks like:

image

Should be a case of setting the font weight appropriately.

#### Front page

That big gap between the console and the "Host, Run and Scale dedicated game servers on Kubernetes"

image

There are some other layout differences, but nothing I think we need to be huge blockers on.

Only other thought - the Community Page ([preview](https://c28c58b-dot-preview-dot-agones-images.appspot.com/site/community/)) seems to have the wide layout. We could check the source code there to see how they are doing that (if we want to go wide that is).

Kalaiselvi84:
I believe I might be overlooking a detail. I attempted to use the command `hugo mod init github.com/me-at-github/my-existing-site`. Specifically, I tried variations such as `hugo mod init github.com/googleforgames/agones/tree/main/site` and `hugo mod init github.com/kalaiselvi84/agones/tree/main/site`.

However, I received the following error:

```
go: /usr/local/google/home/kalaiselvim/Desktop/agones/site/go.mod already exists
Error: failed to init modules: failed to execute 'go [mod init github.com/kalaiselvi84/agones/tree/main/site]': failed to execute binary "go" with args [mod init github.com/kalaiselvi84/agones/tree/main/site]: go: /usr/local/google/home/kalaiselvim/Desktop/agones/site/go.mod already exists
 *errors.errorString
```

I've ignored this error because the error message indicates that `go.mod` already exists. I am unsure if this could be a problem?

markmandel:
> I believe I might be overlooking a detail. I attempted to use the command `hugo mod init github.com/me-at-github/my-existing-site`. Specifically, I tried variations such as `hugo mod init github.com/googleforgames/agones/tree/main/site` and `hugo mod init github.com/kalaiselvi84/agones/tree/main/site`. Reference-https://www.docsy.dev/docs/updating/convert-site-to-module/

Curious - why? 😃

Kalaiselvi84:
> > I believe I might be overlooking a detail. I attempted to use the command `hugo mod init github.com/me-at-github/my-existing-site`. Specifically, I tried variations such as `hugo mod init github.com/googleforgames/agones/tree/main/site` and `hugo mod init github.com/kalaiselvi84/agones/tree/main/site`. Reference-https://www.docsy.dev/docs/updating/convert-site-to-module/
> 
> Curious - why? 😃

I've corrected the error and included the appropriate go.mod and go.sum files in the latest PR. Apologies for the oversight..

markmandel:
Filed https://github.com/Kalaiselvi84/agones/pull/325 against your branch, that I believe fixes all the issues.

PTAL and let me know if you see anything missing!

Kalaiselvi84:
Closing this PR, since the task has been completed in another PR: https://github.com/googleforgames/agones/pull/3417

PR #3394, opened 2023-09-21
Reduce controller memory footprint considerably

### LINKED ISSUE #3380: Agones Controller Memory leak to 460M
**What happened**:
Agones controller memory usage leak to 460M

Agones controller memory usage stays at normal level, not spike unreasonably

Unable to provide specific steps to reproduce currently. This spike happened during normal usage.

Unable to get agones_gameservers_count from metrics at 460M memory usage
Memory usage recovers to normal 40M+ after Agones controller has OOMKilled

**Environment**:
- Agones version:  1.34.0
- Kubernetes version (use `kubectl version`): 1.22
- Cloud provider or hardware configuration: Tencent Cloud
- Others:
image
image

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

In the past it wasn't deemed necessary to give the sub-controllers of the GameServer controller (Health, Missing and Migration) more than 1 worker for their queues, as speed of processing wasn't hugely important for their particular workloads.

An unfortunate consequence of that is that their queues can back up quite significantly, since they can't process events very fast at all. causing big jumps in memory usage.

This change brings those controllers in line with all the others, such that they have the same number of workers as every other controller - and now we have a much more stable memory footprint for the controller since they can process their incoming events in a timely manner.

Closes #3380

Added a couple of extra checks to queuing as well.

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -336,21 +336,21 @@ func (c *Controller) Run(ctx context.Context, workers int) error {
 
 	// Run the Health Controller
 	go func() {
-		if err := c.healthController.Run(ctx); err != nil {
+		if err := c.healthController.Run(ctx, workers); err != nil {
 			c.baseLogger.WithError(err).Error("error running health controller")
 		}
 	}()
 
 	// Run the Migration Controller
 	go func() {
-		if err := c.migrationController.Run(ctx); err != nil {
+		if err := c.migrationController.Run(ctx, workers); err != nil {
 			c.baseLogger.WithError(err).Error("error running migration controller")
 		}
 	}()
 
 	// Run the Missing Pod Controller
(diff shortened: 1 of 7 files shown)


gongmax (approved):
I guess the only possible downside is the CPU usage could go up since more processes?

markmandel:
> I guess the only possible downside is the CPU usage could go up since more processes?

You ask an interesting question, so I went back and looked at the metrics!

Interestingly, it's actually a little less! I wonder if having all that memory backing up makes things difficult for the CPU and/or also garbage collection.

image

markmandel:
@Kalaiselvi84 this would be a good one to have in the release notes for sure 👍🏻

PR #3397, opened 2023-09-21
Change to use grpc-dotnet instead of Grpc.Core in C# SDK

yoshd:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Grpc.Core used in the C# SDK is currently deprecated.
https://grpc.io/blog/grpc-csharp-future/

Changed to use [grpc-dotnet](https://github.com/grpc/grpc-dotnet) instead.

Accordingly, code generation for .proto files has been modified to occur at build time, similar to the [Rust SDK](https://github.com/googleforgames/agones/tree/main/sdks/rust/proto).
To do this, C#-specific rewritten .proto file was output to the `sdks/csharp/sdk` folder.

Closes #


sdks/csharp/sdk/AgonesSDK.cs
```diff
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 using Agones.Dev.Sdk;
-using Grpc.Core;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Grpc.Core;
```
yoshd, line 19: The namespace `Grpc.Core` remains, but it is included in grpc-dotnet.

build/build-sdk-images/csharp/gen.sh
```diff
@@ -16,52 +16,37 @@
 
 set -ex
 
-header() {
-    cat /go/src/agones.dev/agones/build/boilerplate.go.txt "$1" | sponge "$1"
-}
+proto=/go/src/agones.dev/agones/proto
+sdk=${proto}/sdk
+googleapis=${proto}/googleapis
+csharp_proto_file_output_dir=/go/src/agones.dev/agones/sdks/csharp/sdk/proto
```
markmandel, line 22: Curious - does it have to be in `sdk/proto`? Also wondering if we should update the .gitignore so it doesn't ignore that folder anymore, since the csharp code is now generated at build time.

yoshd: It doesn't necessarily need to be in `sdk/proto`, but since only the C# SDK uses this specially rewritten `.proto` files, I decided to output it there.
Currently it doesn't seem to be ignored by .gitignore, but is there likely to be a problem?

markmandel: Sorry, I wasn't clear, lemme explain this another way:

Shall we put the proto files in `/go/src/agones.dev/agones/sdks/csharp/proto` (the `sdk` seems redundant), which is the same as how we do the same thing with rust https://github.com/googleforgames/agones/tree/main/sdks/rust/proto

yoshd: I see, I understand!
I did it that way at first, but the actual C# SDK implementation was in `sdks/csharp/sdk`.
For this reason, although it is redundant, I thought it would be better to put them in the same hierarchy.
WDYT?
https://github.com/googleforgames/agones/tree/main/sdks/csharp/sdk

markmandel: oooooh! So it is!

I guess what's the usual convention for csharp? Are the protos usually kept with the source code (i.e. `sdk`) or kept separately (i.e. one level up)?

We should probably just do whatever the regular convention is - I don't actually mind either way.

yoshd: Normally, the proto file is often shared between the server and client, so I think it should be placed in a separate folder from the source code.
(However, this time it seems to be different in that a proto file exclusively for the C# SDK is generated.)
That said, I agree that it would be better to have a structure similar to most projects.
Now let's modify it to output to `sdks/csharp` .

markmandel: > Now let's modify it to output to sdks/csharp .

Sorry, wasn't sure I 100% understood.

I _think_ we're saying to leave it at `/go/src/agones.dev/agones/sdks/csharp/sdk/proto` yes? or no?

yoshd: Sorry, that's no.
I modify it to `csharp/sdk`.
And I just committed it.

markmandel: Got it! That works for me!

markmandel (approved):

PR #3400, opened 2023-09-25
Adds GameServerAllocation e2e tests for Counters

igooch:
**What type of PR is this?**

/kind feature

* Adds GameServerAllocation filtering e2e test for Counters
* Adds GameServerAllocation sorting e2e test for Counters
* Adds GameServerAllocation CounterActions e2e test for Counters
* Updates Fleet and GameServer e2e tests to delete the fleet after test run
* Adds Counters to GameServerAllocation example
* Changes GameServer UpdateCount logic to truncate requests rather than a no-op for increment or decrement requests that go above capacity or below zero
* Changes UpdateCounterCapacity in gameserver.go to update Count to equal Capacity when Capacity is set to a value less than Count

Working on #2716

### DIFF
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -879,16 +879,16 @@ func (gs *GameServer) UpdateCount(name string, action string, amount int64) erro
 		cnt := counter.Count
 		if action == GameServerPriorityIncrement {
 			cnt += amount
-			// only check for Count > Capacity when incrementing
-			if cnt > counter.Capacity {
-				return errors.Errorf("unable to UpdateCount with Name %s, Action %s, Amount %d. Incremented Count %d is greater than available capacity %d", name, action, amount, cnt, counter.Capacity)
-			}
 		} else {
 			cnt -= amount
-			// only check for Count < 0 when decrementing
-			if cnt < 0 {
-				return errors.Errorf("unable to UpdateCount with Name %s, Action %s, Amount %d. Decremented Count %d is less than 0", name, action, amount, cnt)
-			}
+		}
+		// Truncate to Capacity if Count > Capacity
+		if cnt > counter.Capacity {
+			cnt = counter.Capacity
+		}
+		// Truncate to Zero if Count is negative
+		if cnt < 0 {
+			cnt = 0
 		}
 		counter.Count = cnt
 		gs.Status.Counters[name] = counter
@@ -904,6 +904,7 @@ func (gs *GameServer) UpdateCounterCapacity(name string, capacity int64) error {
 	}
 	if counter, ok := gs.Status.Counters[name]; ok {
 		counter.Capacity = capacity
+		// TODO: If Capacity is now less than Count, do we want to reset Count here to equal Capacity?
 		gs.Status.Counters[name] = counter
 		return nil
 	}
@@ -930,6 +931,7 @@ func (gs *GameServer) AppendListValues(name string, values []string) error {
 	}
 	if list, ok := gs.Status.Lists[name]; ok {
 		mergedList := mergeRemoveDuplicates(list.Values, values)
+		// TODO: Truncate and apply up to cutoff
 		if len(mergedList) > int(list.Capacity) {
 			return errors.Errorf("unable to AppendListValues: Name %s, Values %s. Appended list length %d exceeds list capacity %d", name, values, len(mergedList), list.Capacity)
 		}
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -904,7 +904,10 @@ func (gs *GameServer) UpdateCounterCapacity(name string, capacity int64) error {
 	}
 	if counter, ok := gs.Status.Counters[name]; ok {
 		counter.Capacity = capacity
-		// TODO: If Capacity is now less than Count, do we want to reset Count here to equal Capacity?
+		// If Capacity is now less than Count, reset Count here to equal Capacity
+		if counter.Count > counter.Capacity {
+			counter.Count = counter.Capacity
+		}
 		gs.Status.Counters[name] = counter
 		return nil
 	}
--- a/test/e2e/gameserverallocation_test.go
+++ b/test/e2e/gameserverallocation_test.go
@@ -260,6 +260,84 @@ func TestCreateFleetAndGameServerPlayerCapacityAllocation(t *testing.T) {
 	require.NotEqual(t, gs1.ObjectMeta.Annotations["agones.dev/last-allocated"], gs2.ObjectMeta.Annotations["agones.dev/last-allocated"])
 }
 
+func TestCounterGameServerAllocation(t *testing.T) {
+	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		t.SkipNow()
+	}
+	t.Parallel()
+	ctx := context.Background()
+	client := framework.AgonesClient.AgonesV1()
+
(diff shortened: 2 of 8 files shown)


test/e2e/gameserverallocation_test.go
```diff
@@ -260,6 +263,542 @@ func TestCreateFleetAndGameServerPlayerCapacityAllocation(t *testing.T) {
...
+	unallocated := allocationv1.GameServerAllocationUnAllocated
+
+	testCases := map[string]struct {
+		gsa           allocationv1.GameServerAllocation
+		wantGsaErr    bool                                   // For invalid GSA
+		wantAllocated allocationv1.GameServerAllocationState // For a valid GSA: "allocated" if you expect the GSA to succed in allocating a GameServer, "unallocated" if not
+		wantState     agonesv1.GameServerState
+	}{
+		// TODO: These tests assume packed strategy. Should I also test for distributed? Or is that behavior already sufficiently tested elsewhere in this e2e?
+		"Allocate to same GameServer MinAvailable (available capacity)": {
```
igooch, line 300: These tests assume packed strategy. Should I also test for distributed? Or is that behavior already sufficiently tested elsewhere in this e2e?

test/e2e/gameserverallocation_test.go
```diff
@@ -260,6 +263,542 @@ func TestCreateFleetAndGameServerPlayerCapacityAllocation(t *testing.T) {
...
+						"games": {
+							Capacity: &zero,
+						}}}},
+			wantGsaErr:   false,
+			wantCount:    &five,
+			wantCapacity: &zero,
+		},
+		// TODO: These allocated a gameserver (although the Counter action is not performed) Is this what we want?
+		// Do we want to validate negative CounterActions in the GSA?
+		"decrement past zero": {
```
igooch, line 695: These allocated a gameserver (although the Counter action is not performed) Is this what we want? Do we want to validate negative CounterActions in the GSA?

test/e2e/gameserverallocation_test.go
```diff
@@ -260,6 +263,542 @@ func TestCreateFleetAndGameServerPlayerCapacityAllocation(t *testing.T) {
...
+					Counters: map[string]allocationv1.CounterAction{
+						"games": {
+							Action: &decrement,
+							Amount: &five,
+						}}}},
+			wantGsaErr: false,
+			wantCount:  &zero,
+		},
+		// TODO: IIRC the SDK with zero out Count if Capacity is set to zero. Do we want this same behavior for GameServerAllocation?
+		"change capacity": {
```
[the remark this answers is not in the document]
igooch: Note: this set of "failing" tests aren't doing any validation / assertions currently.

test/e2e/gameserverallocation_test.go
markmandel: (Something weird is with the formatting on this file??? So a bit hard to read 🤔 I wonder what happened there)

But for just manipulating Counters, i think only testing packed is fine (I'm assuming we test packed/distributed in priority ordering tests in another place).

test/e2e/gameserverallocation_test.go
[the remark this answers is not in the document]
markmandel: Yep, same rules everywhere - if you go below zero, it's still zero.

Or as you asking if we should validate it?

If validation: No, if you want to ensure there is room for a subtraction to occur, then you should add that to the filter. ([design note](https://github.com/googleforgames/agones/issues/2716#:~:text=If%20a%20user%20wants%20to%20ensure%20there%20is%20room%20for%20the%20increment%20or%20decrement%2C%20that%20should%20be%20explicitly%20included%20in%20the%20filter%20options%20(i.e.%20decrement%20by%20one%2C%20but%20filter%20for%20counts%20that%20are%20%3E%200%20so%20that%20there%20is%20something%20to%20decrement).))

test/e2e/gameserverallocation_test.go
markmandel: i think the answer to this is the same as above.

test/e2e/gameserverallocation_test.go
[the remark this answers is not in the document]
igooch: Asking if we set Capacity to 0, should we also set Count to 0? We do this in the SDK, but we don't have that pattern implemented in GameServerAllocation (Count stays as-is regardless if Capacity is set to less than Count).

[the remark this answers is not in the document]
igooch: Also, for CountActions is a action can't be done we drop the action. I.e. if the Count is 5 and the action is to decrement by 6, the CounterAction will simply not happen. Are you saying we should truncate? For SDK we truncate batched requests, however we reject individual requests for `out of range. Count must be within range [0,Capacity]`.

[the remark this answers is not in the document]
markmandel: > . Are you saying we should truncate

Yes, truncate - since the allocation can't check (or the filter is the check if the end user wants it).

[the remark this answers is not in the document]
igooch: My only problem with truncating is that's it's inconsistent with the behavior in the SDK. The SDK will reject any request that tries to increment beyond capacity, or decrement past zero. I don't have any opinion on truncating or not truncating, just that whatever we choose should be consistent across the GSA and SDK.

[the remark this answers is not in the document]
markmandel: > My only problem with truncating is that's it's inconsistent with the behavior in the SDK

We can't be consistent though, since we don't know at allocation time which GameServer the allocation will be applied to. The SDK is applying an operation to a specific GameServer known ahead of time. I'd argue they are different enough that consistency isn't as strong a requirement.

If the end user wants to ensure that no matter what, there is enough room for the operation to be applied, they can do so with a filter - but the current design doesn't force them to do that if it's not required for their particular use case where truncation is acceptable.

Or to put it another way, if you want it to be consistent (ish?), use a filter expression on your count/list, but we don't require it.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -904,6 +904,7 @@ func (gs *GameServer) UpdateCounterCapacity(name string, capacity int64) error {
 	}
 	if counter, ok := gs.Status.Counters[name]; ok {
 		counter.Capacity = capacity
+		// TODO: If Capacity is now less than Count, do we want to reset Count here to equal Capacity?
```
igooch, line 907: If Capacity is now less than Count, do we want to reset Count here (in UpdateCounterCapacity) to equal Capacity?

markmandel: This is a really good question. I think the answer should be "yes", since it's the most consistent experience. This might also be a "let's say 'yes' now, and see what the feedback is".

igooch: One sticky point if we have the CounterAction:

Counter: {
  Count 11
  Capacity 20
}

CounterAction: {
  Decrement Count by 10
  Set Capacity to 10
}

Capacity gets set to 10, which then sets Count to also 10.
The Count then gets decremented by 10, which sets the count to 0.

(We could update Count first and then Capacity, but then we'd have issues on scale up when a customer wants to add Count and Capacity.)

Should we go ahead and do then, and just document that behavior?

markmandel: Yeah I think you are right - let's just pick an order and let's document it, and we can adjust later on if need be.

markmandel:
🙌🏻  lemme know when you would like a review 👍🏻

markmandel (approved):

markmandel:
Looks great!

PR #3407, opened 2023-09-29
Adds CounterActions and ListActions to Allocation.proto

igooch:
**What type of PR is this?**

/kind feature

* Adds CounterActions and ListActions to Allocation.proto to be consistent with the GameServerAllocationSpec
* Adds CounterActions and ListActions to the converter for converting between a GameServerAllocation proto and the GameServerAllocationSpec

Working on #2716 


pkg/allocation/converters/converter.go
```diff
@@ -469,3 +486,100 @@ func convertGSAPrioritiesToAllocationPriorities(in []agonesv1.Priority) []*pb.Pr
...
+		if v.Amount != nil {
+			amount := v.Amount.GetValue()
+			ca.Amount = &amount
+		}
+		if v.Capacity != nil {
+			capacity := v.Capacity.GetValue()
+			ca.Capacity = &capacity
+		}
+
+		// TODO: Action and Amount must be used together. Do we want to check that both Action & Amount are not nil,
```
gongmax, line 510: Should this logic be in https://github.com/googleforgames/agones/blob/68698827d35faaea5659926bc6312c4e8257b9bd/pkg/apis/allocation/v1/gameserverallocation.go#L513?

markmandel: Agreed this should be checkedin the Validate() (so we can generally assume that the request is valid)

So in this case, to be defensive, we should check both are not nil, as that is the expected incoming request, and we don't want to accidentally have a panic if a bug happens upstream.

proto/allocation/allocation.proto
```diff
@@ -196,3 +202,21 @@ message Priority {
...
 }
+
+// CounterAction is an optional action that can be performed on a Counter at allocation.
+// Action: "Increment" or "Decrement" the Counter's Count (optional). Must also define the Amount.
+// Amount: The amount to increment or decrement the Count (optional). Must be a positive integer.
+// Capacity: Update the maximum capacity of the Counter to this number (optional). Min 0, Max int64.
+message CounterAction {
+  google.protobuf.StringValue action = 1;
+	google.protobuf.Int64Value amount = 2;
+	google.protobuf.Int64Value capacity = 3;
```
markmandel, line 213: We've used `int64` in the other proto defintinions in this file - but we're using `google.protobuf.Int64Value` here?

I do see we also use it here:

https://github.com/googleforgames/agones/blob/d2189ca7765e8089b1664242dae3a821f5b96a1d/proto/sdk/alpha/alpha.proto#L231-L245

But only noticed the difference now, so curious what the difference is?

Does this provide a nullable value on the generated code side?

igooch: We went with pointers so that the field could be nil. Otherwise int64 defaults to 0 which in this case is a valid input value.

pkg/allocation/converters/converter.go
igooch: Added code in Validate.

proto/allocation/allocation.proto
markmandel: Gotcha! Make sense - thanks for confirming!

markmandel (approved):
Looks great!

PR #3413, opened 2023-10-04
Convert shell script to Go

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #


build/export-openapi.sh
```diff
@@ -84,7 +84,7 @@ for f in *.json; do
   contents=$(cat "$f" | jq 'del(.description)' | sed 's/\\n/\\\\n/g' | sed '$d' | sed '1d' | sed ':a;N;$!ba;s/\n/\\n/g' | sed 's/\$/\\$/g' | sed 's/\&/\\&/g' | sed 's@\"@\\"@g')
   ref=$(basename "$f" .json)
 
-  find -maxdepth 1 -name '*.json' | xargs sed -i 's@"$ref": "#/definitions/'"$ref"'"@'"$contents"'@g'
+  find -maxdepth 1 -name '*.json' | awk -F, -v contents="$contents" '{ gsub("\"\\$ref\": \"#/definitions/" ref "\"", contents, $0)} {print}'
```
markmandel, line 87: What was wrong with sed 😃?

Kalaiselvi84: sed complains, "Argument list too long"

markmandel: Now we're getting "./export-openapi.sh: line 87: /usr/bin/xargs: Argument list too long" 😬 so at least a different issue!

gongmax: In the `awk` command, why do we need `-F,`? We probably don't want print either.

Kalaiselvi84: `-F` is to specify the field separator. Will remove `-F` and print.

gongmax: I understood what `-F` does but do we need it?

Kalaiselvi84: Below two files modified using `make gen-embedded-openapi`  with the command `find -maxdepth 1 -name '*.json' | awk -v contents="$contents" '{ gsub("\"\\$ref\": \"#/definitions/" ref "\"", contents, $0)}'`

1) _io.k8s.api.core.v1.PodTemplateSpec.yaml - https://gist.github.com/Kalaiselvi84/400398322a50efcd1a0c00d2c44c4c19

2) _io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml -https://gist.github.com/Kalaiselvi84/0c5f582223e036eb864b9c453b7341d7

Kalaiselvi84: > I understood what -F does but do we need it?
You're right, it's not necessary. Excluded it🙂

markmandel:
@gongmax I'll let you decide when we've got what we need from this 👍🏻

build/export-openapi.sh
```diff
@@ -84,7 +84,8 @@ for f in *.json; do
   contents=$(cat "$f" | jq 'del(.description)' | sed 's/\\n/\\\\n/g' | sed '$d' | sed '1d' | sed ':a;N;$!ba;s/\n/\\n/g' | sed 's/\$/\\$/g' | sed 's/\&/\\&/g' | sed 's@\"@\\"@g')
   ref=$(basename "$f" .json)
 
-  find -maxdepth 1 -name '*.json' | xargs sed -i 's@"$ref": "#/definitions/'"$ref"'"@'"$contents"'@g'
+  find -maxdepth 1 -name '*.json' | xargs -I {} bash -c 'sed -i "s@\"$1\": \"#/definitions/$1\"@$2@g" {}' _ "$ref" "$contents"
```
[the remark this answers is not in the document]
Kalaiselvi84: Yes, makes sense. Thanks for pointing this out. Will look into it👍🏻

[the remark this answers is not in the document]
Kalaiselvi84: The command using `xargs sed` without the `-i` option runs successfully, but it does not replace the contents as intended:

```
find -maxdepth 1 -name '*.json' | xargs sed 's@"$ref": "#/definitions/'"$ref"'"@'"$contents"'@g'
```

Similarly, the following commands also execute without errors but fail to perform the intended replacements:

```
1) find -maxdepth 1 -name '*.json' | xargs -I 'sed -i "s@"$ref": "#/definitions/'"$ref"'"@'"$contents"'@g"'

2) find -maxdepth 1 -name '*.json' | xargs -n1 -I 'sed -i "s@"$ref": "#/definitions/'"$ref"'"@'"$contents"'@g"'

3) find -maxdepth 1 -name '*.json' | xargs -I {} bash -c 'sed -i "s@\"$ref\": \"#/definitions/$1\"@$2@g" {}' _ "$ref" "$contents"
```

I also attempted using a hardcoded value:
```
find -maxdepth 1 -name '*.json' | xargs sed 's/"$ref"/AGONES/g'
```
same here, no error and no replacement with "AGONES".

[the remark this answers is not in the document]
Kalaiselvi84: I tested the following command with a sample JSON file locally, and it worked, but it's not working for our use case:

```
find -maxdepth 1 -name 1.json | xargs sed -i -e '/"$ref": "#\/definitions\/'"$ref"'"/s/'"$ref"'/'"$contents"'/g'
```
I would appreciate your insights and suggestions on how to proceed further to address this issue effectively...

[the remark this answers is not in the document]
gongmax: Confirmed the above command still have `Argument list too long` error.

markmandel:
Meta question: Is it time to convert this into a little Go program?

build/export-openapi.sh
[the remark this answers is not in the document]
markmandel: I'm fairly sure that the issue is the `xargs` not the `sed` command, so a few options that may work/help:

1. Use `-exec` on `find`, rather than `xargs` (
1. Use `parrallel` https://tecadmin.net/solutions-to-the-argument-list-too-long-error/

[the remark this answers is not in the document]
Kalaiselvi84: yes, I have tried with the below awk command and it is working fine with a simple file.

 `find -maxdepth 1 -name 1.json | xargs awk -i inplace -v ref='"\\$ref": "#/definitions/'"$ref"'"' -v contents="$contents" '{gsub(ref, contents)}; {print}'`

```
$ cat 1.json 
{
  "name": "Spring",
  "age": 300,
  "$ref": "#/definitions/k8s",
  "hobbies": ["AAA", "BBB", "CCC"],
  "address": {
    "street": "123 M st",
    "city": "SF",
    "state": "CA",
    "zip": "123"
  }
}

$ echo $ref
k8s

$ echo $contents
support version update

$ find -maxdepth 1 -name 1.json | xargs awk -i inplace -v ref='"\\$ref": "#/definitions/'"$ref"'"' -v contents="$contents" '{gsub(ref, contents)}; {print}'

$ cat 1.json 
{
  "name": "Spring",
  "age": 300,
  support version update,
  "hobbies": ["AAA", "BBB", "CCC"],
  "address": {
    "street": "123 M st",
    "city": "SF",
    "state": "CA",
    "zip": "123"
  }
}
```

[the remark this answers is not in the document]
gongmax: I think Kalai tried getting rid of `xargs` and using a loop instead, which still throw the Arguments too long error?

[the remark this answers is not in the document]
markmandel: Absolutely have not tried it, so may need some tweaking, but does this work:

```
find -maxdepth 1 -name '*.json' -exec sed -i 's@"$ref": "#/definitions/'"$ref"'"@'"$contents"'@g' {}
```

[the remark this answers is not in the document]
Kalaiselvi84: >Use -exec on find, rather than xargs (


find . -maxdepth 1 -name '*.json' -exec bash -c 'sed -i "s@\"\$ref\": \"#/definitions/$1\"@$2@g" "$0"' {} "$ref" "$contents" \;

received the same error - ./export-openapi.sh: line 87: /usr/bin/find: Argument list too long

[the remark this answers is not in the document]
Kalaiselvi84: I have included all the sed/awk commands I've tried so far, along with any associated errors - https://gist.github.com/Kalaiselvi84/9e94d55cb84c93d80d12818f5f347a51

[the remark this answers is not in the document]
markmandel: I wonder if the contents being replaced it just too large for a single bash command (looking at you Pod Spec) - in which case, probably time to convert this to a Go program.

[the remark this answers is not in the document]
Kalaiselvi84: working on it👍🏻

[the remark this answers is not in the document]
gongmax: `awk: not an option: -i` is because the make command is using `mawk`

build/replaceRef.go
```diff
@@ -0,0 +1,56 @@
...
+	if len(os.Args) != 4 {
+		log.Println("Usage: replaceRef <jsonFilePath> <ref> <contents>")
+		return
+	}
+
+	jsonFilePath := os.Args[1]
+	ref := os.Args[2]
+	contents := os.Args[3]
+
+	processJSON(jsonFilePath, ref, contents)
```
markmandel, line 34: I totally realise how what I said would be confusing, but my comment was more to replace the entire bash file with a Go program, not just this bit 😃

But maybe this is easier?

Kalaiselvi84: got it👍🏻, will covert the bash file with Go 🙂

build/replaceRef.go
```diff
@@ -0,0 +1,56 @@
...
+func processJSON(filename, ref, contents string) {
+	log.Println("Processing JSON file:", filename)
+
+	file, err := os.ReadFile(filename)
+	if err != nil {
+		log.Println("Error reading:", filename, err)
+		return
+	}
+
+	searchPattern := "\"" + ref + "\": \"#/definitions/" + ref + "\""
```
gongmax, line 46: This doesn't look right to me. Should it be `searchPattern := "\"\$ref\": \"#/definitions/" + ref`? I didn't test it though.

Kalaiselvi84:
@markmandel, I really need your help to convert the bash script into a Go script. It looks like the initial steps in the bash script to set the shell options (errexit, nounset, and pipefail) aren't required in Go. Please correct me if I'm wrong. Could you please kindly split the tasks and share the steps?

build/replaceRef.go
Kalaiselvi84: I've completed a Go script to convert from json to yaml file - https://gist.github.com/Kalaiselvi84/c0487628acdfda6f8bb62047b8e88036

markmandel:
>  really need your help to convert the bash script into a Go script. It looks like the initial steps in the bash script to set the shell options (errexit, nounset, and pipefail) aren't required in Go. Please correct me if I'm wrong.

You are correct! They are just some best practices in bash, basically to exit immediately on failure (bash doesn't by default), etc - so you can ignore that line.

The biggest thing this script does is pull [Swagger](https://swagger.io/specification/v2/) specifications out of `http://127.0.0.1:8001/openapi/v2` (when running `kube proxy` you can also see it), for two particular objects: "io.k8s.api.core.v1.PodTemplateSpec" and "io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta" (which you can see at the bottom of the file so that we can use them include them into our own custom resource definitions (CRDs).

A few interesting wrinkles are:
* The Swagger API has some properties that are incompatible with what Kubernetes wants in CRDs, so we delete some of them
* We need to add some properties to a few definitions as well that don't already have them.
* Within the original Swagger spec that comes from K8s, has a `$ref` attribute, which is a reference to another definition. Unfortunately - we aren't allowed to use `$ref` attributes - so a chunk of the work (the `sed` parts you were just looking at) is replacing those $refs with the original definitions found in the Swagger file.

So here are some ideas for converting this entire script to Go:

* `kubectl proxy` - probably just run that as an `os.exec` in  a go routine, and then `os.Exit(0)` when the script is done.
* I think we can keep the JSON as a json string and manipulate it as a string, and it should probably work.
* https://github.com/itchyny/gojq could be used to replace they [jq](https://jqlang.github.io/jq/) parts, 
* https://github.com/itchyny/json2yaml could be used to replace the end result json to YAML (rather than yq)
 * A lot of the other `sed` commands are formatting to allow things to work in Bash scripts, and may be possible to ignore (escaping `\\` for example)


The bash script in this instance uses the local file system to do the work it does, which is easier for bash. For a Go program, in theory, you could do this in memory, but it might be easier to debug on disk. No strong opinion here - whatever is easier. 

Some background on each of the bash functions:

`do_expand` 

Starting with a specific definition (e.g. `do_expand "io.k8s.api.core.v1.PodTemplateSpec"` it pulls that definition out of the large `openapi.json` and makes it's own file. It then looks through that file for and $ref values and does the same thing to them recursiveley.

`json_2_helm_yaml`

That takes the JSON file, converts it to YAMl and applies the appropriate headers and footers.

`for f in *.json; do`

Not a bash function, but the final loop in the process that's worth highlighting.

This part at the end goes through each json file and replaces the $ref values with the contents of each of the others - thereby expanding the contents of all of them, before converting the two items we want into their final yaml formats in https://github.com/googleforgames/agones/tree/main/install/helm/agones/templates/crds/k8s

That's hopefully a good start - lemme know if you have questions!

Kalaiselvi84:
Thanks! This helps!

I've completed the replacing and convert [json_2_yaml](https://gist.github.com/Kalaiselvi84/c0487628acdfda6f8bb62047b8e88036) tasks, I believe. Will work on the remaining steps.

build/replaceRef
aimuz, line 1: Do we need to commit the binary to the repository?

aimuz: Let's change .gitignore.

Kalaiselvi84: [build/replaceRef](https://github.com/googleforgames/agones/pull/3413/files/e17aa7d8ee0b771db9f6ff0c255f50c789d709cf#diff-c5644979ffcc09cd2b20b312729340269f410b4693454f1e6f67c0e9daf3b301) is a piece of Bash script conversion into Go code, but our focus is on converting the whole file to Go code.

Kalaiselvi84:
@markmandel, I've completed the steps below. Please review them, and I will continue working on the remaining steps https://github.com/googleforgames/agones/blob/93a4c4ab663d9a19f65959920d0a71408cabe457/build/export-openapi.sh#L72-L92

- Check, Remove and Create Temporary Directory.
- Start Kubectl Proxy.
- HTTP Request and OpenAPI JSON Handling.
- Make an HTTP GET request to http://127.0.0.1:8001/openapi/v2.
- Store the response in a file named openapi.json in the tmp directory.
- Remove specific fields from the openapi.json file using the removeFields function.
- Read the modified openapi.json file.
- Extract and save definitions from the OpenAPI JSON.
- Recursively search for $ref values in the definitions and expand them.
- Modify specific JSON files (io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.json and io.k8s.apimachinery.pkg.util.intstr.IntOrString.json) by updating required fields.
- Attempt to escape double backslashes in a JSON file.

build/replaceRef.go
```diff
@@ -0,0 +1,56 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package main implements a program that replaces ref with contents in the given json files
+package main
+
+import (
+	"log"
```
markmandel, line 19: Any reason not to pull all this up into the above go file?

build/export-openapi.sh
```diff
@@ -73,6 +73,12 @@ mkdir orig
 cp *.json orig
 rm openapi.json
 
+echo "Compiling replaceRef.go..."
```
markmandel, line 76: I know you aren't there yet, but once we have a comprehensive go script, we can delete export-openapi.sh

markmandel (commented):
On first pass, this is looking like it's heading in the right direction! We'll know once we see the final output, but this is looking good.

Only thought (which I mention below), probably makes sense to make it all one big go file.

build/replaceRef.go
Kalaiselvi84: I'll consolidate everything into the above Go file👍🏻

build/exp-openapi/exp-openapi.go
```diff
@@ -0,0 +1,284 @@
...
+	"log"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func main() {
+	tmpDir := "../tmp"
```
roberthbailey, line 30: Is there a reason we can't use `ioutil.TempFile`

Kalaiselvi84: If I am correct `lint` isn't happy with ioutil package

build/export-openapi.go
markmandel, line 1: Should this live in `build/scripts` like our other Go scripts?

I'm wondering if the go.mod in this particular directory is causing issues 🤔

Kalaiselvi84: Will move it to `build/scripts` and remove the `build/go.mod` and `build/go.sum`.

Kalaiselvi84: Moved it to the `scripts` directory. Created go.mod and go.sum in the `k8s-export-openapi` directory. Did `go mod tidy` and didn't use `go mod vendor`.

I am trying to create a target in `build/Makefile`, getting below error,
```
gen-embedded-openapi: $(ensure-build-image)
    docker run --rm $(common_mounts) -w $(workdir_path)/build $(build_tag) go run scripts/k8s-export-openapi/export-openapi.go
    
Error:
go: finding module for package github.com/itchyny/json2yaml
scripts/k8s-export-openapi/export-openapi.go:29:2: cannot query module due to -mod=vendor
        (Go version in go.mod is at least 1.14 and vendor directory exists.)
make: *** [Makefile:729: gen-embedded-openapi] Error 1
```

Please share your thoughts..

markmandel: This doesn't look the same as to how you run the other scripts:

https://github.com/googleforgames/agones/blob/7080208f0dd0c820e97404fd6826201994545347/build/includes/website.mk#L121-L123

Is there a reason? 🤔

markmandel: Also, should probably be `main.go` as well 😄

Kalaiselvi84: I tried this way too

```
gen-embedded-openapi: ensure-build-image
	docker run --rm $(common_mounts) --workdir=$(mount_path) $(DOCKER_RUN_ARGS) $(build_tag) \
		go run build/scripts/k8s-export-openapi/export-openapi.go
		

$ make gen-embedded-openapi 
mkdir -p ~/.kube/
mkdir -p /usr/local/google/home/kalaiselvim/Desktop/agones/build//.gocache
mkdir -p /usr/local/google/home/kalaiselvim/Desktop/agones/build//.config/gcloud
mkdir -p ~/.config/helm
mkdir -p ~/.cache/helm
docker run --privileged --rm tonistiigi/binfmt:qemu-v6.2.0 --install arm64
installing: arm64 qemu-aarch64 already registered
{
  "supported": [
    "linux/amd64",
    "linux/arm64",
    "linux/386"
  ],
  "emulators": [
    "jar",
    "ksh",
    "llvm-14-runtime.binfmt",
    "python3.11",
    "qemu-aarch64"
  ]
}
make ensure-image IMAGE_TAG=agones-build:d4663e28ac BUILD_TARGET=build-build-image
make[1]: Entering directory '/usr/local/google/home/kalaiselvim/Desktop/agones/build'
make[1]: Leaving directory '/usr/local/google/home/kalaiselvim/Desktop/agones/build'
docker run --rm -v /usr/local/google/home/kalaiselvim/Desktop/agones/build//.config/gcloud:/root/.config/gcloud -v ~/.kube/:/root/.kube -v ~/.config/helm:/root/.config/helm -v ~/.cache/helm:/root/.cache/helm -v /usr/local/google/home/kalaiselvim/Desktop/agones:/go/src/agones.dev/agones -v /usr/local/google/home/kalaiselvim/Desktop/agones/build//.gomod:/go/pkg/mod -v /usr/local/google/home/kalaiselvim/Desktop/agones/build//.gocache:/root/.cache/go-build --workdir=/go/src/agones.dev/agones  agones-build:d4663e28ac \
        go run build/scripts/k8s-export-openapi/export-openapi.go
go: finding module for package github.com/itchyny/json2yaml
build/scripts/k8s-export-openapi/export-openapi.go:29:2: cannot query module due to -mod=vendor
        (Go version in go.mod is at least 1.14 and vendor directory exists.)
make: *** [includes/website.mk:143: gen-embedded-openapi] Error 1
```

markmandel: `export-openapi.go` -> `main.go` like the other scripts please.

build/scripts/k8s-export-openapi/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/agones/agones/build/scripts/k8s-export-openapi
+
+go 1.22
```
markmandel, line 3:
```suggestion
go 1.20
```

Since we aren't yet running 1.22 in the build image.

build/export-openapi.go
Kalaiselvi84: Renamed it to `main.go`

build/scripts/k8s-export-openapi/go.mod
Kalaiselvi84: changed it to 1.20, but we have 1.21 in [build/scripts/site-config-update-version/go.mod](https://github.com/googleforgames/agones/blob/05c97b31964eb51772a21ae3d81f8c62330144ca/build/scripts/site-config-update-version/go.mod#L3)

build/export-openapi.go
Kalaiselvi84: If the script looks good, shall we proceed with the Kubernetes update process first? We can address the target issue afterward?

build/scripts/k8s-export-openapi/go.mod
markmandel: Yeah, it's weird it doesn't complain, given:

https://github.com/googleforgames/agones/blob/05c97b31964eb51772a21ae3d81f8c62330144ca/build/build-image/Dockerfile#L26

But everything else is `1.20` so might as well standardise.

build/export-openapi.go
markmandel: Working now?

Kalaiselvi84: No luck yet, same issue with main.go and 1.20 version.
 Does the script looks good to you?

 Below files are modified from the script :
_io.k8s.api.core.v1.PodTemplateSpec.yaml ->
 Log -> https://gist.github.com/Kalaiselvi84/57093c9369e504caab255119ed04ec22
_io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml -> Log 
-> https://gist.github.com/Kalaiselvi84/4282f0a3a408cf70d30dbc3537ac4b8b

Kalaiselvi84: > Working now?

The script works fine,  if I run with binary file `./main` file or with `go run main.go`

markmandel: Awesome! On first glance, it looks pretty good.

For next steps, let's clean this PR up, and include the generated `_io.k8s.api.core.v1.PodTemplateSpec.yaml` and `_io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml` so it's easier to see if there are any relevant differences in the output.

So (I think this covers it):

* We can delete the `export-openapi.sh`
* Update the `kubernetes_update.md` checlist to use your new make command
* Replace the `_io.k8s.api.core.v1.PodTemplateSpec.yaml` and `_io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml` with your new generated version.
* And any other cleanup that might need to be done.

And I think that should keep us moving forward!

Kalaiselvi84: Completed:
- Deleted the `export-openapi.sh`
- Replaced the _io.k8s.api.core.v1.PodTemplateSpec.yaml and _io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml with new generated version.


Pending: In progress
- Update the kubernetes_update.md checlist to use your new make command


I tried manually running the go script inside the docker container: I am getting `27.0.0.1:8001: connect: connection refused`

```
$ docker run -it --rm -v /usr/local/google/home/kalaiselvim/Desktop/agones:/go/src/agones.dev/agones agones-build:d4663e28ac /bin/sh
# cd src/agones.dev/agones/build/scripts/k8s-export-openapi
# ls
export-openapi  go.mod  go.sum  main  main.go  vendor
# go run main.go
2023/10/19 23:11:32 Get "http://127.0.0.1:8001/openapi/v2": dial tcp 127.0.0.1:8001: connect: connection refused
exit status 1
```
I haven't faced this issue while running the script locally

markmandel: oooh that's tricky. The Build Image does have `kubectl` installed in it... but if it fails for some reason, that likely will happen.

🤔 this might sound weird, but run `kubectl` locally before running the new `make` command that runs the script in the docker file - does that solve it? (it should refresh the authentication token against Kubernetes).

Given that we used to run `export-openapi.sh` in the docker images it should work.

https://github.com/googleforgames/agones/blob/05c97b31964eb51772a21ae3d81f8c62330144ca/build/Makefile#L728-L729

install/helm/agones/templates/crds/k8s/_io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml
```diff
@@ -16,37 +17,27 @@
...
+    type: object
   creationTimestamp:
-    description: |-
-      CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC.
-
-      Populated by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
-    type: string
     format: date-time
-    nullable: true
+    type: string
```
markmandel, line 28: Looks like there might be a bug in `makeCreationTimestampNullable` - since this field isn't nullable.

build/export-openapi.go
Kalaiselvi84: $ kubectl proxy
Starting to serve on 127.0.0.1:8001

Reproduced the same error with `-mod=mod` locally
```
gen-embedd: ensure-build-image
	docker run --rm $(common_mounts) --workdir=$(mount_path) $(DOCKER_RUN_ARGS) $(build_tag) \
		go run -mod=mod build/scripts/k8s-export-openapi/main.go

Error:		
       go run -mod=mod build/scripts/k8s-export-openapi/main.go
go: finding module for package github.com/itchyny/json2yaml
2023/10/19 23:40:31 Get "http://127.0.0.1:8001/openapi/v2": dial tcp 127.0.0.1:8001: connect: connection refused
exit status 1
make: *** [includes/website.mk:143: gen-embedd] Error 1
```

install/helm/agones/templates/crds/k8s/_io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml
Kalaiselvi84: `makeCreationTimestampNullable` code modifies the "creationTimestamp" field under "properties" nullable to true in `io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.json`

```
    "creationTimestamp": {
      "$ref": "#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.Time",
      "description": "CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC.\n\nPopulated by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata",
      "nullable": true
    },
```

markmandel: But it's not in this file that I see here. It should be in here (unless tests pass?)

markmandel:
```
101a102,104
```

Will also need a `make gen-install` 😄  I forgot that step.

build/export-openapi.go
markmandel: Not running `kubectl proxy` locally, was just suggesting something like `kubectl get pods` just to read some information.

build/scripts/k8s-export-openapi/main.go
```diff
@@ -0,0 +1,470 @@
...
+		}
+	}
+
+	err := os.MkdirAll(tmpDir, os.ModePerm)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Start the kubectl proxy.
+	cmd := exec.Command("kubectl", "proxy")
```
markmandel, line 48: So we're running `kubectl proxy` here... so it should be started.

We might want to print/log the output from the command, to make sure it's doing what we expect.

(this might need to be put in a Go routine, since calling any of the output commands such as: https://pkg.go.dev/os/exec#Cmd.Output will block).

build/scripts/k8s-export-openapi/main.go
```diff
@@ -0,0 +1,470 @@
...
+	err = cmd.Start()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Make an HTTP request to fetch the OpenAPI JSON.
+	resp, err := http.Get("http://127.0.0.1:8001/openapi/v2")
```
markmandel, line 55: We may want to throw a few seconds of Sleep above this, just to make sure the proxy is ready before we query it.

install/helm/agones/templates/crds/k8s/_io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml
Kalaiselvi84: `makeCreationTimestampNullable` didn't have bug I guess, All the changes are happing only in `tmp` dir. But the nullable=true is present in the `orig` dir if I am correct. Let me check..

build/scripts/k8s-export-openapi/main.go
Kalaiselvi84: Included -> 5 seconds sleep time, to stop the proxy if already is running at the beginning, and graceful termination at the end.

build/scripts/k8s-export-openapi/main.go
Kalaiselvi84: Done.

Kalaiselvi84:
> ```
> make[1]: Leaving directory '/workspace/build'
> sort /workspace/install/yaml/install.yaml > /tmp/agones-install/install.current.yaml.sorted
> diff /tmp/agones-install/install.yaml.sorted /tmp/agones-install/install.current.yaml.sorted
> 101a102,104
> ```
> 
> Will also need a `make gen-install` 😄 I forgot that step.

I am not able to execute the code locally now, don't know the reason. I have deleted the `tmp` folder already.🙈
Will run `make gen-install` once the issue is fixed..

build/scripts/k8s-export-openapi/main.go
aimuz: Can he be implemented using client-go here?

Kalaiselvi84: Goroutine encountered an error: `Failed to start kubectl proxy: signal: interrupt`. I've excluded it.

I used the same target, `gen-embedded-openapi`, and it successfully executed using the Go script with the `-mod=mod` option. No modifications were needed in the Kubernetes checklist.

Pending task:
Check `nullable=true` missing in yaml file

go.mod
```diff
@@ -74,6 +74,7 @@ require (
 	github.com/googleapis/gax-go/v2 v2.12.0 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
 	github.com/imdario/mergo v0.3.6 // indirect
+	github.com/itchyny/json2yaml v0.1.4 // indirect
```
markmandel, line 77: This should all be reset - there should be changes to the go.mod here or the go.sum I don't think.

vendor/modules.txt
```diff
@@ -239,6 +239,8 @@ github.com/heptiolabs/healthcheck
 # github.com/imdario/mergo v0.3.6
 ## explicit
 github.com/imdario/mergo
+# github.com/itchyny/json2yaml v0.1.4
```
markmandel, line 242: I don't think this should be here either.

Kalaiselvi84: CI build complains🤔

go: inconsistent vendoring in /go/src/agones.dev/agones:
	github.com/itchyny/json2yaml@v0.1.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt

	To ignore the vendor directory, use -mod=readonly or -mod=mod.
	To sync the vendor directory, run:
		go mod vendor

Kalaiselvi84: Alright, if you say so, there must be a reason behind it. I've reverted the changes in go.sum, go.mod, and vendor/modules.txt

Kalaiselvi84: Oh no, If I run the code the go.sum and go.mod files are automatically updated. I guess this is because of `-mod=mod`?
```
gen-embedded-openapi: ensure-build-image
	docker run --rm $(common_mounts) --workdir=$(mount_path) $(DOCKER_RUN_ARGS) $(build_tag) \
		go run -mod=mod build/scripts/k8s-export-openapi/main.go
```

However, the code is failing without `-mod=mod`. What could be the solution?

markmandel: Maybe a silly question - what happens if we do:

```
gen-embedded-openapi: ensure-build-image
	docker run --rm $(common_mounts) --workdir=$(mount_path)/build/scripts/k8s-export-openapi $(DOCKER_RUN_ARGS) $(build_tag) \
		go run -mod=mod ./main.go
```

(Might not need the `-mod=mod` either at that point?)

Kalaiselvi84: I believe that if the `workdir` changes, the paths to other files in the code will also need to be modified.

markmandel: Try it? I expect the script will fail, but we'll see if it updates the other go.mod and go.sum files -- and if it doesn't, then we know making the path changes worth it.

markmandel: The good news is - with the agones/go.mod and agones/go.sum changes reverted - CI is passing again - so that's at least a step in the right direction!

Kalaiselvi84: Script passed with the below modifications:

tmpDir := "../../tmp"
boilerplateContent, err := os.ReadFile("../../boilerplate.yaml.txt")
outputPath := filepath.Join("../../../install/helm/agones/templates/crds/k8s", "_"+filename+".yaml")

markmandel: Love it 😄

Kalaiselvi84:
CI build came out green. Does everything look good?

markmandel (approved):
This looks good to me! @gongmax did you want to also give it a once over before I merge, to make sure I didn't miss anything?

gongmax:
> This looks good to me! @gongmax did you want to also give it a once over before I merge, to make sure I didn't miss anything?

Taking a look...

gongmax (approved):

markmandel:
YAY!

THE BASH MONSTROSITY I SHOULD NEVER HAVE WRITTEN IN THE FIRST PLACE IS NOW GONE!

PR #3418, opened 2023-10-11
Adds Counter FleetAutoScaler e2e Test

igooch:
**What type of PR is this?**

>/kind feature

- Adds FleetAutoScaler e2e Test for Counter
- Updates fleetautoscalers applyCounterOrListPolicy to never return less than one desired replica when scaling down the fleet, as the fleet cannot scale back up from zero
- Adds validation that the autoscaler policy MinCapacity for both Counters and LIsts is greater than zero when buffer is a percentage

Working on #2716 

### DIFF
--- a/pkg/fleetautoscalers/fleetautoscalers_test.go
+++ b/pkg/fleetautoscalers/fleetautoscalers_test.go
@@ -969,6 +969,33 @@ func TestApplyCounterPolicy(t *testing.T) {
 				wantErr:  false,
 			},
 		},
+		"scale up integer": {
+			fleet: modifiedFleet(func(f *agonesv1.Fleet) {
+				f.Spec.Template.Spec.Counters = make(map[string]agonesv1.CounterStatus)
+				f.Spec.Template.Spec.Counters["rooms"] = agonesv1.CounterStatus{
+					Count:    7,
+					Capacity: 10}
+				f.Status.Replicas = 3
+				f.Status.ReadyReplicas = 3
+				f.Status.AllocatedReplicas = 0
+				f.Status.Counters = make(map[string]agonesv1.AggregatedCounterStatus)
+				f.Status.Counters["rooms"] = agonesv1.AggregatedCounterStatus{
+					Count:    21,
+					Capacity: 30}
+			}),
+			featureFlags: string(utilruntime.FeatureCountsAndLists) + "=true",
+			cp: &autoscalingv1.CounterPolicy{
+				Key:         "rooms",
+				MaxCapacity: 100,
+				MinCapacity: 10,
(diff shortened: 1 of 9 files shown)


pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -1162,6 +1189,7 @@ func TestApplyCounterPolicy(t *testing.T) {
 					Capacity: 7}
 			}),
 			featureFlags: string(utilruntime.FeatureCountsAndLists) + "=true",
+			// TODO: Is this what we want (replica = 0 if MaxCapcity < per Counter Capacity)? If buffer = 1 then must be at least replica = 1? Or Max Capacity takes precedence?
```
igooch, line 1192: Is this what we want (replica = 0 if MaxCapcity < per Counter Capacity)? If buffer = 1 then must be at least replica = 1? Or Max Capacity takes precedence?

Test case: https://github.com/googleforgames/agones/blob/2003e674cd3c8a3fceac5ae2f2fb5c3a5f5fc48f/pkg/fleetautoscalers/fleetautoscalers_test.go#L1149-L1187

markmandel: Yeah, a requested buffer shouldn't ever be less than 0 (for regular buffer it's a min value of 1), so there must always be at least one replica.

Scaling to zero is way too hard 😃 since we can't control game server spinup time.

igooch: The requested buffer here is 1, but it's the MaxCapacity that's taking precedence, and since MaxCapacity < per Counter Capacity it scales to 0. 

Just to clarify, in general, you're saying we should never scale to 0?

markmandel: There has to be some number of gameserver(s) to have some allocatable capacity to go towards the buffer - so there should always be at least one, as otherwise that would go against what the fleet autoscaler is requesting.

igooch: Sounds good -- I'll add in a check to make sure we never return less than 1 desired replica.

test/e2e/README.md
```diff
@@ -10,14 +10,21 @@ Prerequisites:
...
 To run e2e tests on your kubectl configured cluster:
 
 ```
 make test-e2e
 ```
 
+To run a single test on your kubectl configured cluster you can optionally include any flags listed
+in e2e test args in the agones/build/Makefile such as `FEATURE_GATES="CountsAndLists=true"`:
+
+```
```
markmandel, line 24: Nice!

markmandel (approved):

PR #3429, opened 2023-10-13
Add Cloudbuild step to run performance test by using the scenario test framework. 

gongmax:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #

Test cluster has not been created yet, it's blocked by the limit increase approval.


build/perf-test-image/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM gcr.io/cloud-builders/gcloud
```
markmandel, line 1: Doesn't the build image already have all these dependencies? Just questioning if we need another image to the set?

ci/perf-test-cloudbuild.yaml
```diff
@@ -0,0 +1,65 @@
+---
```
markmandel, line 1: Apache licence please.

build/perf-test-image/Dockerfile
gongmax: Yes build image has the necessary dependencies but it also have many dependencies which is not required for this task. These dependencies add around 5 more minutes to the total build time. That's the only downside I see now. I wrote theses steps in this way to aline with what we did for e2e test, and I don't have strong objection to switch to build image though.

markmandel: You don't need to build the image though - you could pull the prebuilt image we build and cache, and run that instead.

Copy this section to do it:

https://github.com/googleforgames/agones/blob/93a4c4ab663d9a19f65959920d0a71408cabe457/cloudbuild.yaml#L76-L83

ci/perf-test-cloudbuild.yaml
```diff
@@ -0,0 +1,105 @@
...
+    id: pull-build-image
+    dir: build
+    env: ['REGISTRY=${_REGISTRY}', 'CUSTOM_LOCAL_TAG=performance-test']
+    args: [pull-build-image-with-tag]
+
+  #
+  # Run the performance tests with default feature gates setting
+  #
+  - name: performance-test
+    script: |
```
zmerlynn, line 47: This "feels" long enough that we should maybe bump it out of YAML and into its own thing. WDYT?

zmerlynn (approved):

PR #3436, opened 2023-10-18
simple-game-server: Adds a graceful termination delay

zmerlynn:
* Adds `gracefulTerminationDelaySec` to imitate a gameserver delaying after SIGTERM (eviction).

* Adds test/eviction/evictpod.go, which I used to test the above.

This should allow easier testing in autoscaling scenarios (including eventually possibly e2e tests with eviction: safe: Always).


test/eviction/evictpod.go
```diff
@@ -0,0 +1,66 @@
...
+	policy "k8s.io/api/policy/v1beta1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+)
+
+// Borrowed from https://stackoverflow.com/questions/62803041/how-to-evict-or-delete-pods-from-kubernetes-using-golang-client
+func evictPod(client *kubernetes.Clientset, name, namespace string) error {
+	return client.PolicyV1beta1().Evictions(namespace).Evict(context.TODO(), &policy.Eviction{
```
roberthbailey, line 34: Is there a V1 version of this API (my quick look at godoc makes me think there is)? I think throughout most of our code base we no longer rely on beta APIs, which makes it so that we aren't at risk when APIs get removed in future k8s releases.

test/eviction/evictpod.go
```diff
@@ -0,0 +1,66 @@
...
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		}})
+}
+
+func main() {
+	kubeconfig := flag.String("kubeconfig", filepath.Join(homedir.HomeDir(), ".kube", "config"), "(optional) absolute path to the kubeconfig file")
+	namespace := flag.String("namespace", "default", "Namespace (required)")
```
roberthbailey, line 43: namespace isn't required, since there is a default value.

test/eviction/evictpod.go
```diff
@@ -0,0 +1,66 @@
...
+	pod := flag.String("pod", "", "Pod name (required)")
+	flag.Parse()
+
+	if *pod == "" || *namespace == "" {
```
roberthbailey, line 47: The check for namespace isn't necessary since there is a default value.

test/eviction/evictpod.go
```diff
@@ -0,0 +1,66 @@
...
+		log.Fatal("--pod and --namespace must be non-empty")
+	}
+
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		log.Fatalf("Could not create in cluster config: %v", err)
```
roberthbailey, line 53: nit: I think "in cluster config" refers to a configuration for something running inside of a cluster and trying to reach the k8s API. This is a tool meant to run outside of a cluster so the error is that it can't create a valid client configuration.

test/eviction/evictpod.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	var kubeClient *kubernetes.Clientset
```
roberthbailey, line 56: style nit: Why not just use `:=` on the line below?

test/eviction/evictpod.go
```diff
@@ -0,0 +1,66 @@
...
+	kubeClient, err = kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("Could not create the kubernetes clientset: %v", err)
+	}
+
+	err = evictPod(kubeClient, *pod, *namespace)
```
roberthbailey, line 62: style nit: Cuddle the `err` variable and the `if` statement.

test/eviction/evictpod.go
zmerlynn: Done

test/eviction/evictpod.go
zmerlynn: I went ahead and took out the `(required)`

test/eviction/evictpod.go
zmerlynn: I copied this from the client examples, so any style here was there as well, but I'll fix it.

test/eviction/evictpod.go
zmerlynn: The error from k8s is clear enough here, so falling back to the server error message for this weird case:

```
$ go run test/eviction/evictpod.go --pod simple-game-server-f8lft-56fhg --namespace=""
2023/10/18 16:08:32 Pod eviction failed: an empty namespace may not be set during creation
exit status 1
```

test/eviction/evictpod.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		log.Fatalf("Could not build config: %v", err)
+	}
+
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("Could not create the kubernetes clientset: %v", err)
+	}
+
+	if err = evictPod(kubeClient, *pod, *namespace); err != nil {
```
roberthbailey, line 61: nit: `if err := evictPod(...)`

test/eviction/evictpod.go
```diff
@@ -0,0 +1,64 @@
...
+
+	policy "k8s.io/api/policy/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+)
+
+// Borrowed from https://stackoverflow.com/questions/62803041/how-to-evict-or-delete-pods-from-kubernetes-using-golang-client
+func evictPod(client *kubernetes.Clientset, name, namespace string) error {
```
roberthbailey, line 33: nit: I know it doesn't follow the example, but proper style is to pass in the context here and use context.Background in main().

roberthbailey (commented):
Left a few nits. I also verified that passing a negative time to sleep is ok and doesn't need any additional error checking.

zmerlynn:
> Once this merges and we push the new image version, we need to update all of the references across the code base. Can you do that in a follow up PR?

Yup, that was the plan!

PR #3442, opened 2023-10-19
add opt-out ttlSecondsAfterFinished setting for the pre-delete hook

### LINKED ISSUE #3432: Add ability to automatically cleanup the `delete-agones-resources` jobs after completion
When uninstalling an Agones release via helm, it [creates a Job](https://github.com/googleforgames/agones/blob/main/install/helm/agones/templates/hooks/pre_delete_hook.yaml#L19) to force deleting other Agones resources if `agones.crds.cleanupOnDelete` is set to `false`.

Jobs, and their associated Pods, in Kubernetes inherently do not get deleted so that developers can review the outcome of the job. However, as this job is fairly autonomous it ends up feeling like more clutter than value to have the pods lingering until K8s garbage collection deletes them. This is exacerbated if you are bringing Agones releases up and down multiple times during development.

Implementing the [TTL mechanism for finished jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs) would allow these jobs to be deleted after some period of time that could be configured as a chart value.

Users could create their own tooling/scripts to manually delete the jobs, or they could set `agones.crds.cleanupOnDelete=false`, but it would be ideal if Agones provided an interface to handle deleting completed jobs.

**Additional context**

Slack conversation [here](https://agones.slack.com/archives/C9DGM5DS8/p1697490791912969)

mikeseese:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix
> /kind release

This PR adds the `agones.crds.cleanupJobTTL` Helm value to set the [Kubernetes `ttlSecondsAfterFinished`](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) spec value for the `pre_delete_hook` Job. This will, by default, cleanup the Job and associated Pods once they complete after 60 seconds, regardless if successful or not. Users can opt out of this functionality by setting `agones.crds.cleanupJobTTL` to `<= 0`. The ref doc just states 0 for simplicity, but the Helm template is more forgiving if they want to use null. Also if they put a float or falsey non-integer in for the value (e.g. `3.5` or `null` or `false`), it will consider it as `0` in an `int` conversion and not add the ttl value.

By having this feature, the Jobs, which are likely to succeed, won't pile up and clutter the cluster/dashboards after multiple iterations (particularly on a development cluster).

Closes #3432


site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -100,6 +100,7 @@ The following tables lists the configurable parameters of the Agones chart and t
...
 | `agones.crds.cleanupOnDelete`                            | Run the pre-delete hook to delete all GameServers and their backing Pods when deleting the helm chart, so that all CRDs can be removed on chart deletion                                                                                                                                                         | `true`                                    |
+| `agones.crds.cleanupJobTTL`                              | The number of seconds for Kubernetes to delete the associated Job and Pods of the pre-delete hook after it completes, regardless if the Job is successful or not. Set to `0` to disable cleaning up the Job or the associated Pods.                                                                              | `60`                                      |
```
markmandel, line 103: This looks great - quick thing, can you make a copy of the table add `feature` codes around the new and old version?

See: https://agones.dev/site/docs/contribute/#within-a-page

We publish the docs on merge with `main`, so we have a way to keep old features hidden until they are ready in the next release.

Lemme know if you have any questions.

mikeseese: I think I implemented this properly; let me know if I missed something!

markmandel:
Looks like a flake, although a funny one:


```
--- FAIL: TestGameServerAllocationValidate (0.12s)
    gameserverallocation_test.go:1185: 
        	Error Trace:	/go/src/agones.dev/agones/pkg/apis/allocation/v1/gameserverallocation_test.go:1185
        	Error:      	"[spec.scheduling: Unsupported value: "FLERG": supported values: "Packed", "Distributed" spec.required.counters: Forbidden: Feature CountsAndLists must be enabled spec.required.lists: Forbidden: Feature CountsAndLists must be enabled]" should have 1 item(s), but has 3
        	Test:       	TestGameServerAllocationValidate
FAIL
FAIL	
```

site/content/en/docs/Installation/Install Agones/helm.md
markmandel: Looks good! Will see the preview when the build finishes! 👍🏻

markmandel (approved):
Nice!

mikeseese:
Ah, I pressed the "update branch" button, and that seems to have removed the lgtm label. I wasn't sure if that being updated with `main` was a requirement.

markmandel:
> Ah, I pressed the "update branch" button, and that seems to have removed the lgtm label. I wasn't sure if that being updated with `main` was a requirement.

Doesn't matter - it'll also get automatically updated and merged from here out.

PR #3445, opened 2023-10-20
Implements UpdateList, AddListValue, and RemoveListValue in the SDK Server

igooch:
**What type of PR is this?**

/kind feature

- Implements the List Update (Capacity), Add Value to List, and Remove Value from List in the SDK Server.

Working on #2716 


pkg/sdkserver/sdkserver.go
```diff
@@ -985,41 +999,238 @@ func (s *SDKServer) GetList(ctx context.Context, in *alpha.GetListRequest) (*alp
...
 // [FeatureFlag:CountsAndLists]
 func (s *SDKServer) UpdateList(ctx context.Context, in *alpha.UpdateListRequest) (*alpha.List, error) {
 	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
 		return nil, errors.Errorf("%s not enabled", runtime.FeatureCountsAndLists)
 	}
-	// TODO(#2716): Implement Me
-	return nil, errors.Errorf("Unimplemented -- UpdateList coming soon")
+
+	if in.List == nil || in.UpdateMask == nil {
+		return nil, errors.Errorf("INVALID_ARGUMENT. List: %v and UpdateMask %v cannot be nil", in.List, in.UpdateMask)
```
markmandel, line 1050: Nit: I'm guessing this code is from before we moved to standard Go errors, but just in case - something like:

```suggestion
		return nil, errors.Errorf("invaid argument. List: %v and UpdateMask %v cannot be nil", in.List, in.UpdateMask)
```

pkg/sdkserver/sdkserver.go
```diff
@@ -985,41 +999,238 @@ func (s *SDKServer) GetList(ctx context.Context, in *alpha.GetListRequest) (*alp
...
+			return nil, errors.Errorf("OUT_OF_RANGE. No available capacity. Current Capacity: %d, List Size: %d", list.Capacity, len(list.Values))
+		}
+		// Verify value does not already exist in the list
+		// TODO: This does not check batched but not yet applied append / remove values. Should we do this?
+		// (Easy to check not yet applied values, hard to check removed and re-added values.) I'm
+		// thinking this would be better / easier to do as part of the batch apply update to list, and
+		// not verify here.
+		for _, val := range list.Values {
+			if in.Value == val {
+				return nil, errors.Errorf("ALREADY_EXISTS. Value: %s already in List: %s", in.Value, in.Name)
```
markmandel, line 1119: So from the design for `ListAppend` (assuming we're still calling it that) [we have](https://github.com/googleforgames/agones/issues/2716#:~:text=Returns%20false%2C%20if%20the%20value%20already%20exists%20in%20the%20list%2C%20or%20if%20the%20list%20is%20already%20at%20capacity%20(to%20the%20latest%20knowledge%20of%20the%20SDK).):

> Returns false, if the value already exists in the list, or if the list is already at capacity (to the latest knowledge of the SDK).

(or at least on a best effort basis).

So we might need to modify the return type on this function?

pkg/sdkserver/sdkserver.go
```diff
@@ -985,41 +999,238 @@ func (s *SDKServer) GetList(ctx context.Context, in *alpha.GetListRequest) (*alp
...
+		if batchList.capacitySet != nil {
+			capacity = *batchList.capacitySet
+		} else {
+			capacity = int64(len(list.Values) + len(batchList.listAppend) - len(batchList.listDelete))
+		}
+		if list.Capacity <= capacity {
+			return nil, errors.Errorf("OUT_OF_RANGE. No available capacity. Current Capacity: %d, List Size: %d", list.Capacity, len(list.Values))
+		}
+		// Verify value does not already exist in the list
+		// TODO: This does not check batched but not yet applied append / remove values. Should we do this?
```
markmandel, line 1113: Could we get the up to date list from `GetList(...)`, and then check that list of values?

pkg/sdkserver/sdkserver.go
```diff
@@ -985,41 +999,238 @@ func (s *SDKServer) GetList(ctx context.Context, in *alpha.GetListRequest) (*alp
...
+			}
+			// Add value to remove to gsListUpdates map.
+			batchList := s.gsListUpdates[in.Name]
+			batchList.listDelete = append(batchList.listDelete, in.Value)
+			s.gsListUpdates[in.Name] = batchList
+			// Queue up the Update for later batch processing by updateLists.
+			s.workerqueue.Enqueue(cache.ExplicitKey(updateLists))
+			return &alpha.List{}, nil
+		}
+		return nil, errors.Errorf("NOT_FOUND. Value: %s not found in List: %s", in.Value, in.Name)
```
markmandel, line 1165: From the [design](https://github.com/googleforgames/agones/issues/2716#:~:text=Returns%20false%20if%20the%20value%20is%20not%20found%20in%20the%20list%20(to%20the%20latest%20knowledge%20of%20the%20SDK)%2C), similar to append

> Returns false if the value is not found in the list (to the latest knowledge of the SDK),

So potentially a tweak on return types.

pkg/sdkserver/sdkserver.go
```diff
@@ -985,41 +999,238 @@ func (s *SDKServer) GetList(ctx context.Context, in *alpha.GetListRequest) (*alp
...
+	}
+	gsCopy := gs.DeepCopy()
+
+	s.logger.WithField("batchListUpdates", s.gsListUpdates).Debug("Batch updating List(s)")
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+
+	names := []string{}
+
+	for name, listReq := range s.gsListUpdates {
```
markmandel, line 1185: As a potetially interesting thought - I would figure `GetList()` would do this work as well / already.

You could port that into there, and call `GetList(...)` from here as well to get what should be the source of truth and then update?

pkg/sdkserver/sdkserver.go
```diff
@@ -985,41 +999,238 @@ func (s *SDKServer) GetList(ctx context.Context, in *alpha.GetListRequest) (*alp
...
+		list, ok := gsCopy.Status.Lists[name]
+		if !ok {
+			continue
+		}
+		if listReq.capacitySet != nil {
+			list.Capacity = *listReq.capacitySet
+		}
+		if len(listReq.listDelete) != 0 {
+			// TODO: Is there a quicker way of doing this while still keeping the ordering in tact?
```
markmandel, line 1194: Not that I can think of 🤔 The only potential thought I have that might be faster, is keep the list of deleted items in a Map, since the order of deleted item's shouldn't matter (I don't think?) - that would mean you could likely remove all values in a single loop, rather than a loop of loops.

sdks/go/alpha.go
```diff
@@ -172,3 +173,37 @@ func (a *Alpha) SetCounterCapacity(key string, amount int64) (bool, error) {
...
+func (a *Alpha) GetListCapacity(key string) (int64, error) {
+	list, err := a.client.GetList(context.Background(), &alpha.GetListRequest{Name: key})
+	if err != nil {
+		return -1, errors.Wrapf(err, "could not get List %s capacity", key)
+	}
+	return list.Capacity, nil
+}
+
+// SetListCapacity sets the capacity for a given list. Capacity must be between 1 and 1000.
+// TODO: Should capacity be between 0 and 1000?
```
markmandel, line 191: [Yep](https://github.com/googleforgames/agones/issues/2716#:~:text=Capacity%20can%20be%20no%20longer%20than%201000%20items.%20This%20could%20possibly%20be%20expanded%20in%20the%20future%20depending%20on%20use%20cases%20and/or%20performance.)!

> Capacity can be no longer than 1000 items. This could possibly be expanded in the future depending on use cases and/or performance.

markmandel (commented):
Generally looks like a good approach - just added some thoughts!

pkg/sdkserver/sdkserver.go
```diff
@@ -985,41 +999,238 @@ func (s *SDKServer) GetList(ctx context.Context, in *alpha.GetListRequest) (*alp
...
+		// TODO: This does not check batched but not yet applied append / remove values. Should we do this?
+		// (Easy to check not yet applied values, hard to check removed and re-added values.) I'm
+		// thinking this would be better / easier to do as part of the batch apply update to list, and
+		// not verify here.
+		for _, val := range list.Values {
+			if in.Value == val {
+				return nil, errors.Errorf("ALREADY_EXISTS. Value: %s already in List: %s", in.Value, in.Name)
+			}
+		}
+		batchList.listAppend = append(batchList.listAppend, in.Value)
```
markmandel, line 1122: Should we also check `listDelete` to see if the added value is in there, and if so, remove it from `listDelete` ?

sdks/go/alpha.go
igooch: Ok, in the [#2716](https://github.com/googleforgames/agones/issues/2716) it mentioned 1:1000, so wanted to double check it's 0:1000.

pkg/sdkserver/sdkserver.go
igooch: It returns false from the client SDK.

pkg/sdkserver/sdkserver.go
igooch: I think I tried that with Counter and it was complicated with the s.gsUpdateMutex.Lock(). Any idea how to do that?

pkg/sdkserver/sdkserver.go
igooch: In the batched updateList we first delete values, and then add values. So, in effect any adds happen after deletes. In general, is this the behavior we want?

pkg/sdkserver/sdkserver.go
igooch: I mucked around with it a bit, and if we delete values of the list as we're going through it it will mess with the index, so that `list.Values = append(list.Values[:i], list.Values[i+1:]...)` would no longer delete the correct index,or the index may no longer exist. We could create a new slice and write the non-deleted values to that new slice, although I don't know if that will make much of a difference in terms of actual execution time?

pkg/sdkserver/sdkserver.go
markmandel: The way I would do it is move the implementation of `GetList` into a non locked private function, and make sure to document that calls to that function must have appropriate locking.

Then `GetList` and all the other functions can call that private function to get the implementation, but can have control over the over arching mutex lock before the invocation of the non-locking private function.

pkg/sdkserver/sdkserver.go
markmandel: > We could create a new slice and write the non-deleted values to that new slice, although I don't know if that will make much of a difference in terms of actual execution time?

Yeah, that's how I'd probably do it - create a new slice, loop through the existing one, check the map and if the current value doesn't exist in the deletion map, add it to the new slice, and that new slice is the source of truth.

Uses more space (but not really a lot), but should be O(n) in time -- and I think it's a simpler implementation?

pkg/sdkserver/sdkserver.go
```diff
@@ -978,48 +992,258 @@ func (s *SDKServer) updateCounter(ctx context.Context) error {
...
+			s.logger.Debugf("truncating Values in Update List request to List Capacity %d", list.Capacity)
+			list.Values = append([]string{}, list.Values[:list.Capacity]...)
+		}
+		gsCopy.Status.Lists[name] = list
+		names = append(names, name)
+	}
+
+	gs, err = s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})
+	if err != nil {
+		return err
```
igooch, line 1220: Should we add a log line here? `s.recorder.Event(gs, corev1.EventTypeWarning, "UpdateList", fmt.Sprintf("Unable to update List(s) with batched change requests: %v", s.gsListUpdates))` Or will this automatically retry? The changes aren't lost -- they're in s.gsCounterUpdates, but will it get added back to the queue?

pkg/sdkserver/sdkserver.go
igooch: Looking at this again, I'm not sure we want to use GetList here. Here we get the gameserver (and lock on it), and update the values directly in the gameserver copy. If we use "Get List" we'd have to do that for each list, which would be calling s.gameServer() for each list.

pkg/sdkserver/sdkserver.go
markmandel: Any return of an error means the worker queue will retry (all our workerqueues are like this), so you don't have to do anything extra, and I don't think a log is required. It's expected to fail when out of sync.

markmandel (commented):
Getting close I think! Nice stuff!

pkg/apis/agones/v1/gameserver.go
```diff
@@ -947,16 +946,20 @@ func (gs *GameServer) AppendListValues(name string, values []string) error {
 		}
 		// If only some values are duplicates, those duplicate values are silently dropped.
 		list.Values = mergedList
+		// Truncate values if more than capacity
+		if len(list.Values) > int(list.Capacity) {
```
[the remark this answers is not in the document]
igooch: Good catch, line 940 if from before the truncation. I'll remove that.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -947,16 +946,20 @@ func (gs *GameServer) AppendListValues(name string, values []string) error {
 		}
 		// If only some values are duplicates, those duplicate values are silently dropped.
```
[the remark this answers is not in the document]
igooch: Good point, I'll remove.

pkg/sdkserver/sdkserver.go
```diff
@@ -978,48 +992,246 @@ func (s *SDKServer) updateCounter(ctx context.Context) error {
...
+	s.logger.WithField("name", in.Name).Debug("Add List Value")
+
+	list, err := s.GetList(ctx, &alpha.GetListRequest{Name: in.Name})
+	if err != nil {
+		return nil, err
+	}
+
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+
```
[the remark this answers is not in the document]
igooch: Even if we make getList a private function it calls s.gameServer() which has a lock, so it will always have a lock regardless.

pkg/sdkserver/sdkserver.go
```diff
@@ -978,48 +992,246 @@ func (s *SDKServer) updateCounter(ctx context.Context) error {
...
+	// Verify room to add another value
+	if int(list.Capacity) <= len(list.Values) {
+		return nil, errors.Errorf("out of range. No available capacity. Current Capacity: %d, List Size: %d", list.Capacity, len(list.Values))
+	}
+	// Verify value does not already exist in the list
+	for _, val := range list.Values {
+		if in.Value == val {
+			return nil, errors.Errorf("already exists. Value: %s already in List: %s", in.Value, in.Name)
```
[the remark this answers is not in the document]
igooch: We throw an error here, which the client SDK then returns to its caller as false.

pkg/sdkserver/sdkserver.go
```diff
@@ -978,48 +992,246 @@ func (s *SDKServer) updateCounter(ctx context.Context) error {
...
+
+	gs, err := s.gameServer()
+	if err != nil {
+		return nil, err
+	}
+
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+
+	// TODO: Pull in variable Max Capacity from CRD instead of hard-coded number here.
```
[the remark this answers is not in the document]
igooch: I'm not sure if it's possible to generate a variable based on the CRD min / max, so partly putting this here as a "look into this later" and partly as a warning that if the CRD changes, that this code does not change. I don't have a const anywhere for this, so I should probably add that in.

pkg/sdkserver/sdkserver.go
igooch: They're still somewhat different. This method updates agonesv1.ListStatus and the Getlist updates alpha.List.

pkg/sdkserver/sdkserver.go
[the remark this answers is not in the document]
markmandel: > I'm not sure if it's possible to generate a variable based on the CRD min / max

That's a good question... I also don't think that's possible, but I get the impetus. Const does seem like the win, although it's duplicate declaration.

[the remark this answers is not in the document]
markmandel: ..and resolving, since I saw you did this 👍🏻 😄

pkg/sdkserver/sdkserver.go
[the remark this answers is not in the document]
markmandel: Oh yeah, that's 100% true. And it's also batching the operation for a later date - so I think this is reasonable. Removing suggestion!

pkg/sdkserver/sdkserver.go
```diff
@@ -978,48 +992,246 @@ func (s *SDKServer) updateCounter(ctx context.Context) error {
...
+	s.logger.WithField("name", in.Name).Debug("Remove List Value")
+
+	list, err := s.GetList(ctx, &alpha.GetListRequest{Name: in.Name})
+	if err != nil {
+		return nil, err
+	}
+
+	s.gsUpdateMutex.Lock()
+	defer s.gsUpdateMutex.Unlock()
+
```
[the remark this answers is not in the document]
markmandel: Removing as well.

markmandel (approved):
Awesome!

PR #3448, opened 2023-10-21
Update fleet autoscaling limited signification(#2828)

### LINKED ISSUE #2828: FleetAutoScaler scalingLimited is reporting the incorrect state
**What happened**:
It seems as though the scalingLimited attribute of an autoscaler is incorrectly reporting it's state.

I expect scalingLimited = false when the number of gameservers < the maxReplicas 

1.  Configure an autoscaler with: max - 50, min - 10, buffer - 5
2. Allocate 1 or 2  GameServers
3. Observer how scalingLimited is set to true even though we're below our max replica threshold
image (3)

**Environment**:
- Agones version: 1.26
- Kubernetes version (use `kubectl version`): 1.24
- Others: Original Slack Thread: https://agones.slack.com/archives/C9DGM5DS8/p1668793997712099

atgane:
**What type of PR is this?**:

/kind documentation

image

- On #2828 , we found that user can easily misunderstand the meaning of `ScalingLimited`. 

image

- It's actually `ScalingUpOrDownLimited` in [applyBufferPolicy()](https://github.com/googleforgames/agones/blob/main/pkg/fleetautoscalers/fleetautoscalers.go#plain=1#L203). It works like above image:

- ✅ So I want to change grafana metric title `ScalingLimited` to `Scaling Down or Up Limited` and enhance code readability

- Closes #2828

**Special notes for your reviewer**: N/A


pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -221,18 +221,19 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
 	}
 
-	limited := false
+	scalingDownLimited := false
+	scalingUpLimited := false
 
 	if replicas < b.MinReplicas {
 		replicas = b.MinReplicas
-		limited = true
+		scalingDownLimited = true
```
gongmax, line 229: Sorry for the back and force but better to change this and line 233, 236 too for consistency

PR #3450, opened 2023-10-23
Update Supported Kubernetes to 1.26, 1.27, 1.28

Kalaiselvi84:
**What type of PR is this?**
>

/kind breaking

> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Work on #3371 


pkg/gameserverallocations/allocator_test.go
```diff
@@ -81,6 +81,7 @@ func TestAllocatorAllocate(t *testing.T) {
 
 	require.NoError(t, a.Run(ctx))
 	// wait for it to be up and running
+	// nolint:staticcheck
 	err := wait.PollImmediate(time.Second, 10*time.Second, func() (done bool, err error) {
```
gongmax, line 85: I saw you fixed this lint warning with `PollUntilContextTimeout` which looks correct. Do you mind fix these too? We can also do this in a separate PR. Up to you.

pkg/gameserverallocations/allocator.go
```diff
@@ -692,6 +692,7 @@ func Retry(backoff wait.Backoff, fn func() error) error {
 			return false, nil
 		}
 	})
+	// nolint:staticcheck
```
gongmax, line 695: Let's create an issue to track the fix.

gongmax (commented):
Did you perform step "Regenerate Kubernetes resource includes (e.g. ObjectMeta, PodTemplateSpec)"? I didn't see any change for ObjectMeta and PodTemplateSpec.

Kalaiselvi84:
> Regenerate Kubernetes resource includes

Yes, I've executed this step, and within this PR, neither _io.k8s.api.core.v1.PodTemplateSpec.yaml nor _io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.yaml were modified.
These two files were previously altered by the PR related to the [Go script conversion](https://github.com/googleforgames/agones/pull/3413/files#diff-b40ff7394c6e4e3d7105988c2dae58b41a6b4872b636b153d4c9e9ed99fd3757). For reference, I set up the 1.27 cluster before this conversion took place.

pkg/gameserverallocations/allocator_test.go
Kalaiselvi84: sure, will replace all `PollImmediate` with `PollUntilContextTimeout`.

Kalaiselvi84: I've replaced `PollImmediate` with `PollUntilContextTimeout`.
Is it fine to replace `wait.Poll(time.Second, 10*time.Second, func() (bool, error)` with `wait.PollUntilContextTimeout(ctx, time.Second, 10*time.Second, true, func(ctx context.Context) (bool, error))`?

gongmax: Yes, `wait.Poll` can be replaced with `wait.PollUntilContextTimeout`. Again, up to you whether do that in this PR or a following up PR.

pkg/sdkserver/sdkserver.go
```diff
@@ -295,6 +295,7 @@ func (s *SDKServer) WaitForConnection(ctx context.Context) error {
 	}
 
 	try := 0
+	// nolint:staticcheck
 	return wait.PollImmediateInfiniteWithContext(ctx, 4*time.Second, func(ctx context.Context) (bool, error) {
```
gongmax, line 299: This can also be replaced with `PollUntilContextCancel`

test/e2e/framework/framework.go
```diff
@@ -292,6 +292,7 @@ func (f *Framework) WaitForGameServerState(t *testing.T, gs *agonesv1.GameServer
 // Each Allocated GameServer gets deleted allocDuration after it was Allocated.
 // GameServers will continue to be Allocated until a message is passed to the done channel.
 func (f *Framework) CycleAllocations(ctx context.Context, t *testing.T, flt *agonesv1.Fleet, period time.Duration, allocDuration time.Duration) {
+	// nolint:staticcheck
 	err := wait.PollImmediateUntil(period, func() (bool, error) {
```
gongmax, line 296: This can also be replaced with [PollUntilContextCancel](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediateUntil)

gongmax (approved):
There are some deprecated methods can be replaced. But I'm totally fine to fix it in a following PR if you prefer.

Kalaiselvi84:
> There are some deprecated methods can be replaced. But I'm totally fine to fix it in a following PR if you prefer.

Will create a separate PR for this fix😅

PR #3468, opened 2023-10-27
Disable deletion protection for Autopilot test clusters

gongmax:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Seems deletion protection is enabled by default for AP clusters, and we saw cluster couldn't be deleted when we updated k8s version for test clusters. This PR disable it explicitly. 

Closes #


roberthbailey:
This is changing the value is the terraform module, which changes it for both the test clusters but also for anyone that is using our example module: https://github.com/googleforgames/agones/blob/main/examples/terraform-submodules/gke-autopilot/module.tf

I think it would make more sense to make it a variable, set it to true by default (to keep the current behavior) and set it to false explicitly in the test configs.

Kalaiselvi84:
Are we good to merge this PR?

zmerlynn (approved):

zmerlynn (changes requested):
Wait. One thing I realized for both this and maintenance exclusions is that we injected it into the `install/terraform` path. I don't think that's right - I think maintenance exclusions should be optional in the base install and we need this in the e2e specific one. Unless I'm mis-remembering the structure here?

gongmax:
> Wait. One thing I realized for both this and maintenance exclusions is that we injected it into the `install/terraform` path. I don't think that's right - I think maintenance exclusions should be optional in the base install and we need this in the e2e specific one. Unless I'm mis-remembering the structure here?

Make sense. Updated accordingly.

install/terraform/modules/gke/variables.tf
```diff
@@ -20,21 +20,23 @@ variable "cluster" {
...
+    "releaseChannel"                = "UNSPECIFIED"
+    "kubernetesVersion"             = "1.27"
+    "windowsInitialNodeCount"       = "0"
+    "windowsMachineType"            = "e2-standard-4"
+    "autoscale"                     = false
+    "workloadIdentity"              = false
+    "minNodeCount"                  = "1"
+    "maxNodeCount"                  = "5"
+    "maintenanceExclusionStartTime" = null
+    "maintenanceExclusionEndTime"   = null
```
zmerlynn, line 39: did you mean to add `deletionProtection` in here as well?

gongmax: We didn't see any issue with standard cluster last time when we recreate test clusters. So I'd prefer not add it here unless we see the issue next time.

zmerlynn (approved):

PR #3484, opened 2023-11-01
Adds Go SDK Client List Functions

igooch:
**What type of PR is this?**

/kind feature

Adds Go client functions for List

Working on #2716 

### DIFF
--- a/sdks/go/alpha.go
+++ b/sdks/go/alpha.go
@@ -20,6 +20,7 @@ import (
 	"agones.dev/agones/pkg/sdk/alpha"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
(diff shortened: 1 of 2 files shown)


sdks/go/alpha.go
```diff
@@ -172,3 +173,88 @@ func (a *Alpha) SetCounterCapacity(key string, amount int64) (bool, error) {
...
+// ListContains returns if a string exists in a List's values list, given the List's key (name)
+// and the string value. Search is case-sensitive.
+// Will error if the key was not predefined in the GameServer resource on creation.
+func (a *Alpha) ListContains(key, value string) (bool, error) {
+	list, err := a.client.GetList(context.Background(), &alpha.GetListRequest{Name: key})
+	if err != nil {
+		return false, errors.Wrapf(err, "could not get List %s", key)
+	}
+	// TODO: Should I move this logic to the SDK Server, or is it OK here?
+	for _, val := range list.Values {
```
igooch, line 212: Should I move this logic to the SDK Server, or is it OK here? Alpha proto & SDK have Add / Remove / Update / Get, but no "contains".

sdks/go/alpha.go
```diff
@@ -172,3 +173,87 @@ func (a *Alpha) SetCounterCapacity(key string, amount int64) (bool, error) {
...
+	return list.Values, nil
+}
+
+// AppendListValue appends a string to a List's values list, given the List's key (name)
+// and the string value. Will error if the string already exists in the list.
+// Will error if the key was not predefined in the GameServer resource on creation.
+func (a *Alpha) AppendListValue(key, value string) (bool, error) {
+	_, err := a.client.AddListValue(context.Background(), &alpha.AddListValueRequest{Name: key, Value: value})
+	if err != nil {
+		return false, errors.Wrapf(err, "could not get List %s", key)
```
gongmax, line 245: The error message here doesn't line up with what `AddListValue` returns https://github.com/googleforgames/agones/blob/main/proto/sdk/alpha/alpha.proto#L173-L174. Similar for `DeleteListValue`.

igooch: We're not really following the gRPC at this point. If we were the SDK Server would return <alpha.List nil>, status.Errorf(<grpc error code>, <error message>) and the SDK client would have something like list, err := client.AddListValue(...) statusCode := status.FromError(err).

gongmax (approved):

PR #3488, opened 2023-11-03
Adds Counter conformance test

igooch:
**What type of PR is this?**

 /kind feature

Working on #2716 

### DIFF
--- a/pkg/sdkserver/localsdk.go
+++ b/pkg/sdkserver/localsdk.go
@@ -144,6 +144,12 @@ func NewLocalSDKServer(filePath string, testSdkName string) (*LocalSDKServer, er
 	if runtime.FeatureEnabled(runtime.FeaturePlayerTracking) && l.gs.Status.Players == nil {
 		l.gs.Status.Players = &sdk.GameServer_Status_PlayerStatus{}
 	}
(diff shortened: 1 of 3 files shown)


pkg/sdkserver/localsdk.go
```diff
@@ -629,7 +635,7 @@ func (l *LocalSDKServer) UpdateCounter(ctx context.Context, in *alpha.UpdateCoun
 	name := in.CounterUpdateRequest.Name
 
 	l.logger.WithField("name", name).Info("Updating Counter")
-	l.recordRequest("updateCounter")
+	l.recordRequest("updatecounter")
```
markmandel, line 638: This looks good - the only thought I have, is if there might be value to breaking this out into seperate types of updates - i.e. if capacity is set, do a `l.recordRequest("countercapacity")`, if there's a value set `l.recordRequest("counterset")`, if there's addition, `l.recordRequest("counterincrement")` ?

That way you're not just testing "Did a counter action happen", but which type?

This is also covered by the unit test getting expected data, but wanted float the idea, see if it resonated.

But otherwise, this look good 👍🏻

markmandel (approved):
Like it!

PR #3489, opened 2023-11-06
controller refresh certificate

### LINKED ISSUE #3133: Refreshing controller cert if its changed
**What happened**: 
I'm using self-signed certificate managed by cert-manager. After re-creating/renewing certificate and secret, pod controller starts spamming errors in log and I'm not able create fleet. Only after re-creating pod controller its feel good again.
`http: TLS handshake error from xx:xx remote error: tls: bad certificate`

There is no errors in logs and controller functions good

Create certificate and secret for controller. Re-create or renew cert using cert-manager or manually. Check controller logs.

Probably there is need the same fix as was done for allocator [here](https://github.com/googleforgames/agones/pull/1145)

**Environment**:
- Agones version: 1.30.0
- Kubernetes version: client (1.21) and server (1.23)

ashutosji:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
 /kind feature
> /kind hotfix
> /kind release

**Which issue(s) this PR fixes**:https://github.com/googleforgames/agones/issues/3133

Closes #3133 

### DIFF
--- a/pkg/util/https/server.go
+++ b/pkg/util/https/server.go
@@ -43,11 +43,11 @@ type tls interface {
 type Server struct {
 	logger   *logrus.Entry
 	Mux      *http.ServeMux
+	Certs    *cryptotls.Certificate
+	CertMu   sync.Mutex
 	tls      tls
 	certFile string
 	keyFile  string
-	CertMu   sync.Mutex
-	Certs    *cryptotls.Certificate
 }
 
 // NewServer returns a Server instance.
--- a/pkg/util/https/server.go
+++ b/pkg/util/https/server.go
@@ -31,6 +31,8 @@ const (
 	tlsDir = "/certs/"
 )
 
+var closeCh = make(chan struct{})
+
 // tls is a http server interface to enable easier testing
 type tls interface {
 	Close() error
@@ -59,7 +61,22 @@ func NewServer(certFile, keyFile string) *Server {
 		certFile: certFile,
 		keyFile:  keyFile,
 	}
-	wh.setupServer()
+
+	tlsCert, err := readTLSCert()
+	if err != nil {
+		logrus.WithError(err).Fatal("could not load TLS certs.")
+	}
+	wh.CertMu.Lock()
+	wh.Certs = tlsCert
+	wh.CertMu.Unlock()
+
+	wh.tls = &http.Server{
+		Addr:    ":8081",
+		Handler: wh.Mux,
+	}
+
+	// Start a goroutine to watch for certificate changes
+	go watchForCertificateChanges(wh)
 
 	wh.Mux.HandleFunc("/", wh.defaultHandler)
 	wh.logger = runtime.NewLoggerWithType(wh)
@@ -67,28 +84,17 @@ func NewServer(certFile, keyFile string) *Server {
 	return wh
 }
 
-func (s *Server) setupServer() {
-	s.tls = &http.Server{
-		Addr:    ":8081",
-		Handler: s.Mux,
-		TLSConfig: &cryptotls.Config{
-			GetCertificate: s.getCertificate,
-		},
+// It will load the key pair certificate
+func readTLSCert() (*cryptotls.Certificate, error) {
+	tlsCert, err := cryptotls.LoadX509KeyPair(tlsDir+"server.crt", tlsDir+"server.key")
+	if err != nil {
+		return nil, err
 	}
-
-	// Start a goroutine to watch for certificate changes
-	go s.watchForCertificateChanges()
-}
-
-// getCertificate returns the current TLS certificate
-func (s *Server) getCertificate(hello *cryptotls.ClientHelloInfo) (*cryptotls.Certificate, error) {
-	s.CertMu.Lock()
-	defer s.CertMu.Unlock()
-	return s.Certs, nil
+	return &tlsCert, nil
 }
 
 // watchForCertificateChanges watches for changes in the certificate files
-func (s *Server) watchForCertificateChanges() {
+func watchForCertificateChanges(s *Server) {
 	// Watch for changes in the tlsDir
 	cancelTLS, err := fswatch.Watch(s.logger, tlsDir, time.Second, func() {
 		// Load the new TLS certificate
@@ -107,7 +113,9 @@ func (s *Server) watchForCertificateChanges() {
 		s.logger.WithError(err).Fatal("could not create watcher for TLS certs")
 	}
 
-	defer cancelTLS()
+	// Wait for the signal to close
+	<-closeCh
+	cancelTLS()
 }
 
 // Run runs the webhook server, starting a https listener.
@@ -116,6 +124,7 @@ func (s *Server) Run(ctx context.Context, _ int) error {
 	go func() {
 		<-ctx.Done()
 		s.tls.Close() // nolint: errcheck,gosec
+		close(closeCh)
 	}()
 
 	s.logger.WithField("server", s).Infof("https server started")
--- a/pkg/util/https/server.go
+++ b/pkg/util/https/server.go
@@ -19,11 +19,19 @@ import (
 	"crypto/tls"
 	"net/http"
 	"sync"
+	"time"
 
+	"agones.dev/agones/pkg/util/fswatch"
 	"agones.dev/agones/pkg/util/runtime"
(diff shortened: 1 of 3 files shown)


cmd/controller/main.go
```diff
@@ -262,6 +269,48 @@ func main() {
...
+		newHTTPServer.TLSConfig = &tls.Config{
+			Certificates: []tls.Certificate{*newCert},
+		}
+
+		// Update the TLS configuration
+		go func() {
+			if err := newHTTPServer.ListenAndServeTLS("", ""); err != nil {
+				logger.WithError(err).Error("Failed to update TLS configuration")
+			}
+		}()
```
ashutosji, line 296: Hi @markmandel and @zmerlynn,
The moto is to update reloaded certificates in the existing server. But, however i did not find a way to update `httpsServer.TLSConfig`. I thought of creating a new server and updating the reloaded certificates. But it doesn't seems correct to me and technically it is wrong. Since, we are using custom server by utilizing this package https://github.com/googleforgames/agones/blob/main/pkg/util/https/server.go. Can i get right approach or suggestions on this?

markmandel: I found a great StackOverflow post on this, with links to several examples:
https://stackoverflow.com/questions/37473201/is-there-a-way-to-update-the-tls-certificates-in-a-net-http-server-without-any-d

TL;DR - https://pkg.go.dev/crypto/tls#Config exposes a `GetCertificate(...)`  function that will be called when a certificate is required - returning whatever is the current certificate from that is the appropriate way to go.

ashutosji: Sure, Thanks I will check this.

markmandel: This is the package we should update:
https://github.com/googleforgames/agones/blob/1708fe9a6efdad4061c7a53ec8835dc92d4f4fbe/pkg/util/https/server.go

gongmax (commented):
I still think we should implement it like the pattern mentioned in https://stackoverflow.com/questions/37473201/is-there-a-way-to-update-the-tls-certificates-in-a-net-http-server-without-any-d. I.e. have a structure to keep the certs, which get refreshed bya goroutine to watch for certificate changes (i.e. watch on the tlsDir). Then the tls.TLSConfig.GetCertificate can be set to a func that simply return whatever is in the structure rather than load the cert again.
The major concern I have for the current implementation is that I'm not quite sure when the GetCertificate is called and whether it could miss some corner case, which would not be an issue in the option that using goroutine to watch for certificate changes.

ashutosji:
> I still think we should implement it like the pattern mentioned in https://stackoverflow.com/questions/37473201/is-there-a-way-to-update-the-tls-certificates-in-a-net-http-server-without-any-d. I.e. have a structure to keep the certs, which get refreshed bya goroutine to watch for certificate changes (i.e. watch on the tlsDir). Then the tls.TLSConfig.GetCertificate can be set to a func that simply return whatever is in the structure rather than load the cert again. The major concern I have for the current implementation is that I'm not quite sure when the GetCertificate is called and whether it could miss some corner case, which would not be an issue in the option that using goroutine to watch for certificate changes.

You are right! Rather than loading the file every time, I can store the current certificate in the structure and have GetCertificate function which will simple return the certificate.

But however, There are two issue that I am facing:

1. Again started getting the same `Error from server (InternalError): error when creating "https://raw.githubusercontent.com/googleforgames/agones/release-1.37.0/examples/simple-game-server/gameserver.yaml": Internal error occurred: failed calling webhook "mutations.agones.dev": failed to call webhook: Post "https://agones-controller-service.agones-system.svc:443/mutate?timeout=10s": tls: failed to verify certificate: x509: certificate signed by unknown authority`
2.  Facing the issue while handling watch certificate in the unit test here: https://github.com/googleforgames/agones/blob/main/pkg/util/https/server_test.go#L41

gongmax (commented):
I think line 123 should be changed to `err := s.tls.ListenAndServeTLS("", "")`, similar as https://github.com/googleforgames/agones/blob/main/cmd/allocator/main.go#L351

pkg/util/https/server.go
```diff
@@ -43,23 +53,63 @@ type Server struct {
...
+	defer s.CertMu.Unlock()
+	return s.Certs, nil
+}
+
+// watchForCertificateChanges watches for changes in the certificate files
+func (s *Server) watchForCertificateChanges() {
+	// Watch for changes in the tlsDir
+	cancelTLS, err := fswatch.Watch(s.logger, tlsDir, time.Second, func() {
+		// Load the new TLS certificate
+		tlsCert, err := cryptotls.LoadX509KeyPair(tlsDir+"server.crt", tlsDir+"server.key")
```
gongmax, line 95: I think the cert need to be fetched once outside the watcher so it can capture the initial value, similar as: https://github.com/googleforgames/agones/blob/main/cmd/allocator/main.go#L412

ashutosji: Yes, You are right. I had added few logs and checked getCertificate() function is getting called but it has null value.

pkg/util/https/server.go
```diff
@@ -43,23 +53,63 @@ type Server struct {
...
+	s.tls = &http.Server{
+		Addr:    ":8081",
+		Handler: s.Mux,
+		TLSConfig: &cryptotls.Config{
+			GetCertificate: s.getCertificate,
+		},
+	}
+
+	// Start a goroutine to watch for certificate changes
+	go s.watchForCertificateChanges()
```
gongmax, line 80: This doesn't need to be a gorountine, just `s.watchForCertificateChanges()` should be fine

pkg/util/https/server.go
```diff
@@ -43,23 +53,63 @@ type Server struct {
...
+		defer s.CertMu.Unlock()
+		// Update the Certs structure with the new certificate
+		s.Certs = &tlsCert
+		s.logger.Info("TLS certs updated")
+	})
+	if err != nil {
+		s.logger.WithError(err).Fatal("could not create watcher for TLS certs")
+	}
+
+	defer cancelTLS()
```
gongmax, line 110: This will close the watcher after this func returns. I think you need to return this cancelTLS() function to the caller (actually up to the main.go of controller/extension) and defer it there, so the watcher can keep watching the events.

pkg/util/https/server.go
kevlar-ashu: Sure, I made `WatchForCertificateChanges(`) function public so that it should be accessible from extensions/controller.

pkg/util/https/server.go
kevlar-ashu: Thanks you so much.
Yes, watcher was getting close after func return. I have modified the `WatchForCertificateChanges()` function and called it in extensions. It's working!

pkg/util/https/server.go
ashutosji: Sure, I made `WatchForCertificateChanges()` function public so that it should be accessible from extensions/controller.

pkg/util/https/server.go
ashutosji: Thanks you so much.
Yes, watcher was getting close after func return. I have modified the `WatchForCertificateChanges()` function and called it in extensions. It's working!

pkg/util/https/server.go
```diff
@@ -16,50 +16,111 @@ package https
...
 // tls is a http server interface to enable easier testing
 type tls interface {
 	Close() error
 	ListenAndServeTLS(certFile, keyFile string) error
 }
 
+// CertServer holds the Server certificate
+type CertServer struct {
+	Certs  *cryptotls.Certificate
+	CertMu sync.Mutex
```
gongmax, line 43: nit: these fields doesn't have to exported (beginning with upper case) since they are only used in this package, right? Better to change them to unexported for stricter restriction. Similar for the `CertServer` in `Server`.

gongmax (commented):
LGTM, only have one minor comment.

pkg/util/https/server.go
kevlar-ashu: Yeah! Thanks for pointing. It should be unexported type.

ashutosji: Yeah! Thanks for pointing. It should be unexported type.

gongmax (approved):

PR #3505, opened 2023-11-14
Fixes occasional data race flake with TestSDKServerAddListValue

igooch:
**What type of PR is this?**

/kind bug

Fixes occasional data race flake with TestSDKServerAddListValue by putting the s.gsListUpdates call in the logger after the mutex lock.

NA

### DIFF
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -1175,9 +1175,10 @@ func (s *SDKServer) updateList(ctx context.Context) error {
 	}
 	gsCopy := gs.DeepCopy()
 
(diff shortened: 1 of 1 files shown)


pkg/sdkserver/sdkserver.go
```diff
@@ -1175,10 +1175,11 @@ func (s *SDKServer) updateList(ctx context.Context) error {
 	}
 	gsCopy := gs.DeepCopy()
 
-	s.logger.WithField("batchListUpdates", s.gsListUpdates).Debug("Batch updating List(s)")
```
markmandel, line 1178: That would do it! Nice catch.

markmandel (approved):

PR #3516, opened 2023-11-20
Adds gameserverallocation e2e tests for Lists

igooch:
**What type of PR is this?**

/kind feature

Adds GameServerAllocation e2e tests for Lists.

Changes GameServerAllocation action to truncate the List to Capacity if only the capacity is changed, to be consistent with the behavior with Counters.

Working on #2716 

### DIFF
--- a/test/e2e/gameserverallocation_test.go
+++ b/test/e2e/gameserverallocation_test.go
@@ -261,7 +261,7 @@ func TestCreateFleetAndGameServerPlayerCapacityAllocation(t *testing.T) {
 	require.NotEqual(t, gs1.ObjectMeta.Annotations["agones.dev/last-allocated"], gs2.ObjectMeta.Annotations["agones.dev/last-allocated"])
 }
 
-func TestCounterGameServerAllocation(t *testing.T) {
+func TestCounterAndListGameServerAllocation(t *testing.T) {
 	if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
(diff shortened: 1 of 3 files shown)


test/e2e/gameserverallocation_test.go
```diff
@@ -720,6 +845,110 @@ func TestCounterGameServerAllocationActions(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.Equal(t, allocated, gs1.Status.State)
+			assert.NotNil(t, gs1.ObjectMeta.Annotations["agones.dev/last-allocated"])
+
+			list, ok := gs1.Status.Lists["players"]
+			assert.True(t, ok)
+			if testCase.wantCapacity != nil {
+				assert.Equal(t, *testCase.wantCapacity, list.Capacity)
+			}
+			if diff := cmp.Diff(testCase.wantValues, list.Values); diff != "" {
```
markmandel, line 930: Any reason not `assert.Equals` or `require.Equals` ? Then you get a nice diff between the two values as well if they aren't Equal.

markmandel (commented):
Looks good - just one question, then I think good to merge.

test/e2e/gameserverallocation_test.go
igooch: I was thinking that assert.Equal couldn't handle slices, but apparently it works with string slices. Changing to assert.Equal.

markmandel (approved):

PR #3522, opened 2023-11-29
Ensure ci/save_cache and ci/restore_cache images don't get deleted by cleanup policy

### LINKED ISSUE #3476: CI: Ensure `ci/save_cache` and `ci/restore_cache` images don't get deleted by cleanup policy
We implemented a 40 day cleanup policy on the `/ci` container registry, which accidentally deleted our images that we use as Cloud Build Steps to store and retrieve cache files in Google Cloud Storage.

We need a solution so that (a) the images don't get deleted and (b) the image should get updated every so often.

So a few initial thoughts/questions:

* We should probably rebuild the images on a schedule (once a week?) to keep security updates etc up to date.
* The cache image is built on top of `gcr.io/cloud-builders/gcloud-slim` which should get regular updates, so if we rebuild the caching images on a schedule, that should keep the dependency up to date.
* https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/master/cache/cloudbuild.yaml puts the image in gcr, rather than Artifact Registry. Maybe we want to vendor this dependency, with our own version?
* If we do schedule rebuilds on a schedule, we could keep this image in `/ci` with the 40 day cleanup policy, since it would get replaced every 40 days (I assume it reset the created date on recreation?).

Context:

* https://cloud.google.com/artifact-registry/docs/repositories/cleanup-policy
* https://cloud.google.com/build/docs/configuring-builds/use-community-and-custom-builders
* Source: https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/cache

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #3476 


Kalaiselvi84:
@markmandel, Could you please review this PR? I am unsure if I am in the right direction.

markmandel (commented):
I assume this is a vendoring of the code from the community repo? But looks like you updated the tags to use Artifact Registry?  👍🏻 

This should sit somewhere in `/build` -- maybe under `/build/build-image/cache` ?

The only other thing that looks like something we should do is ensure each file has the Apache 2 licence. It's weird that the original builder files don't.... but we should definitely make sure we do.

Ooh - also, let's throw a README.md in the directory explaining where this originally came from and what changes were made on import in case we ever want to update it.

Sound good?

Kalaiselvi84:
> I assume this is a vendoring of the code from the community repo? But looks like you updated the tags to use Artifact Registry? 👍🏻

Yes, that's correct. However, I am fine with the second approach. Looking forward to hearing your preference.

> This should sit somewhere in `/build` -- maybe under `/build/build-image/cache` ?

Moved files to `/build/build-image/cache`.

> The only other thing that looks like something we should do is ensure each file has the Apache 2 licence. It's weird that the original builder files don't.... but we should definitely make sure we do.

Done!

> Ooh - also, let's throw a README.md in the directory explaining where this originally came from and what changes were made on import in case we ever want to update it.

Included the origin and the changes made in our repo.

Kalaiselvi84:
The cloudbuild.yaml script is failing; it looks like we need to modify the steps based on our repo.🤔

markmandel:
> The cloudbuild.yaml script is failing; it looks like we need to modify the steps based on our repo.🤔

What's the output when it fails?

Kalaiselvi84:
> What's the output when it fails?

Log indicates: Step #1 - "build_save_cache": Unknown option: build
Here's the full log - https://gist.github.com/Kalaiselvi84/13b9286ad4ab6e1e0477d79eeb06c96c

build/build-image/cache/cloudbuild.yaml
```diff
@@ -0,0 +1,146 @@
...
+  - name: 'gcr.io/cloud-builders/docker'
+    id: build_base_image
+    args:
+      - 'build'
+      - '--tag=us-docker.pkg.dev/$PROJECT_ID/ci/cache:${_VERSION}'
+      - '--tag=us-docker.pkg.dev/$PROJECT_ID/ci/cache:latest'
+      - '--file=Dockerfile-base'
+      - '.'
+
+  - name: 'us-docker.pkg.dev/$PROJECT_ID/ci/save_cache'
```
markmandel, line 27:
```suggestion
  - name: 'gcr.io/cloud-builders/docker'
```

Since this is building the docker container, so it needs to use docker as the build step container.

build/build-image/cache/cloudbuild.yaml
```diff
@@ -0,0 +1,146 @@
...
+    args:
+      - 'build'
+      - '--tag=us-docker.pkg.dev/$PROJECT_ID/ci/save_cache:${_VERSION}'
+      - '--tag=us-docker.pkg.dev/$PROJECT_ID/ci/save_cache:latest'
+      - '--file=Dockerfile-save'
+      - '--build-arg=project_id=$PROJECT_ID'
+      - '--cache-from=us-docker.pkg.dev/$PROJECT_ID/ci/cache:latest'
+      - '.'
+
+  - name: 'us-docker.pkg.dev/$PROJECT_ID/ci/restore_cache'
```
markmandel, line 38:
```suggestion
  - name: 'gcr.io/cloud-builders/docker'
```

markmandel (commented):
Try that 👍🏻

build/build-image/cache/cloudbuild.yaml
Kalaiselvi84: Success Log: https://gist.github.com/Kalaiselvi84/5ccadb47f469435391ce91a3148cfeae

Whole time I was searching for the alternate options.😬

build/includes/release.mk
```diff
@@ -87,3 +87,8 @@ pre-build-release:
 post-build-release:
 	docker run --rm $(common_mounts) -w $(workdir_path)/build/release $(build_tag) \
 		gcloud builds submit . --substitutions _VERSION=$(base_version) --config=./post_cloudbuild.yaml $(ARGS)
+
+# Ensure that ci/save_cache and ci/restore_cache are not deleted by the cleanup policy
+cache-save-restore:
```
markmandel, line 92: 🤔 I don't need this for the setup of the Cloud Build scheduled triggers, but if we want to keep it for testing purposes (which is useful), I think it might be better to be part of https://github.com/googleforgames/agones/blob/main/build/includes/build-image.mk since it's not a target needed for the release process.

Open to suggestion on a better place, but that's the best idea I had.

Kalaiselvi84: Understood. I'll go ahead and remove it since it's not required for the setup of the Cloud Build scheduled triggers.

markmandel (approved):

PR #3526, opened 2023-12-01
Add Headless Service for agones-allocator

govargo:
**What type of PR is this?**

/kind feature

I'd like to use agones-allocator with [MultiClusterService](https://cloud.google.com/kubernetes-engine/docs/how-to/multi-cluster-services)(MCS) in multiple Kubernetes clusters.

In order to allocate with multiple clusters environment, currently there are some ways.
 1.  External [Multi Cluster Gateway API](https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-multi-cluster-gateways), however, Internal Multi Cluster Gateway API(gke-l7-rilb-mc) [doesn't support cross region](https://cloud.google.com/kubernetes-engine/docs/how-to/enabling-multi-cluster-gateways#restrictions_and_limitations)
 2. Standalone Internal Application LoadBalancer (Internal Application loadbalancer support [cross region](https://cloud.google.com/load-balancing/docs/l7-internal/setting-up-l7-cross-reg-internal)) outside of GKE
 3. Anthos Service Mesh or other L7 layer proxy. e.g. [global-multiplayer-demo](https://github.com/googleforgames/global-multiplayer-demo/blob/main/platform/open-match/base/agones-allocator-vs.yaml) uses ASM

I'd like to add another solution, 4. MultiClusterService + client-side load balancing.
This diagram is what I'd like to deploy.

image

In Kubernetes env, L7 proxy or client-side loadbalancing is needed for gRPC's load balancing.
MultiClusterService is the L4 layer service, so we need client-side loadbalancing, if we'd like to load balancing agones-allocator's gRPC.

```
e.g.  client-side loadbalancing by golang

conn, err := grpc.DialContext(ctx, allocatorEndpoint, grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`)
```

To communite by client-side loadbalancing, DNS name resolusion for each agones-allocator pod is needed.
DNS name resolusion for each pod can be done by [Headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) in Kubernetes.
However, current helm chart for agones-allocator doesn't support Headless Service.
So this is why I sent this PR.

Thank you.

None


install/helm/agones/templates/service/allocation.yaml
```diff
@@ -72,6 +72,9 @@ spec:
       protocol: TCP
 {{- end }}
   type: {{ .Values.agones.allocator.service.serviceType }}
+{{- if and (eq .Values.agones.allocator.service.serviceType "ClusterIP") (.Values.agones.allocator.service.headless) }}
```
markmandel, line 75: Rather than a special case for "headless", any reason not to do a `agones.allocator.service.clusterIP` - which can be set to whatever you want, and provides greater flexibility?

zmerlynn:
@gongmax FYI

install/helm/agones/templates/service/allocation.yaml
govargo: Thank you for your comment!
Surely. `clusterIP` field which users can set any value has greater flexibility and can also be used in cases other than headless.

I changed my commit while referring to `loadBalanceIP`.
`{{- $useLoadBalancerIP := and (ne .Values.agones.allocator.service.loadBalancerIP "") (eq .Values.agones.allocator.service.serviceType "LoadBalancer") }}`

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -72,6 +73,9 @@ spec:
       protocol: TCP
 {{- end }}
   type: {{ .Values.agones.allocator.service.serviceType }}
+{{- if $clusterIP }}
```
markmandel, line 76:
```suggestion
{{- if (ne .Values.agones.allocator.service.clusterIP "") }}
```

I write this because this change sent me lookinga thte K8s [reference docs](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#:~:text=to%20type%20ExternalName.-,clusterIP%20(string),-clusterIP%20is%20the), and looking at `Service.spec.clusterIP`, and I saw it say:

> Only applies to types ClusterIP, NodePort, and LoadBalancer

So it's not just the `ClusterIP` value for `Service.spec.type` -- so I'd recommend leaving it open, and let the end user decide when it's appropriate to set the value, rather than constrain them?

govargo:
Retest with minor documentation clean-up https://github.com/googleforgames/agones/pull/3526/commits/5c0610983c6e0d54d26c435981df318b2e139711 and passed CI

markmandel (approved):
Looks good!

PR #3535, opened 2023-12-07
CloudBuild: Fix for cache image rebuild

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Somewhere `substitutions` got lost, adding it back in. I think that will solve my issue with creating a scheduled trigger. Will need to merge before I can test it.

N/A

N/A

### DIFF
--- a/build/build-image/cache/cloudbuild.yaml
+++ b/build/build-image/cache/cloudbuild.yaml
@@ -137,6 +137,9 @@ steps:
       - name: original
         path: /original
 
+substitutions:
+  _VERSION: '1.2'
+
 images:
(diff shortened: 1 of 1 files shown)


build/build-image/cache/cloudbuild.yaml
```diff
@@ -137,6 +137,9 @@ steps:
       - name: original
         path: /original
 
+substitutions:
```
Kalaiselvi84, line 140: ohh right! I was initially adding this substitution in the target, but we later decided not to use any target

PR #3536, opened 2023-12-09
Go CRD Comment Updates for Counters and Lists

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix
> /kind release

Updated some comments on the Counter and Lists with an aim to make the generated reference document more readable and provide more context of the operations therein.

Work on #2716

Noted some things that stood out as small quality of life improvements in comments. Non blocking for this release, but if we like them, let's create issues / track them in #2716 to be updated in the next one.

### DIFF
--- a/pkg/apis/allocation/v1/gameserverallocation.go
+++ b/pkg/apis/allocation/v1/gameserverallocation.go
@@ -21,13 +21,13 @@ import (
 	"agones.dev/agones/pkg/apis"
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	"agones.dev/agones/pkg/util/runtime"
-	hashstructure "github.com/mitchellh/hashstructure/v2"
+	"github.com/mitchellh/hashstructure/v2"
 	corev1 "k8s.io/api/core/v1"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	"k8s.io/apimachinery/pkg/labels"
-	field "k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 const (
@@ -105,11 +105,12 @@ type GameServerAllocationSpec struct {
 	// You can use this to tell the server necessary session data
 	MetaPatch MetaPatch `json:"metadata,omitempty" hash:"ignore"`
 
-	// (Alpha, CountsAndLists feature flag) Counters and Lists provide a set of actions to perform
-	// on Counters and Lists during allocation.
+	// (Alpha, CountsAndLists feature flag) Counter actions to perform during allocation.
 	// +optional
 	Counters map[string]CounterAction `json:"counters,omitempty" hash:"ignore"`
-	Lists    map[string]ListAction    `json:"lists,omitempty" hash:"ignore"`
+	// (Alpha, CountsAndLists feature flag) List actions to perform during allocation.
+	// +optional
+	Lists map[string]ListAction `json:"lists,omitempty" hash:"ignore"`
 }
 
 // GameServerSelector contains all the filter options for selecting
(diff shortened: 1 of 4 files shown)


pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -146,40 +147,56 @@ type PlayerSelector struct {
...
 
 // CounterSelector is the filter options for a GameServer based on the count and/or available capacity.
-// 0 for MaxCount or MaxAvailable means unlimited maximum. Default for all fields: 0
 type CounterSelector struct {
-	MinCount     int64 `json:"minCount"`
-	MaxCount     int64 `json:"maxCount"`
+	// MinCount is the minimum current value. Defaults to 0.
+	// +optional
+	MinCount int64 `json:"minCount"`
+	// MaxCount is the maximum current value. Defaults to 0, which translates as max(in64).
```
markmandel, line 154: (not for this release) This 0 as `max(int64)` feels a bit confusing to me. Should we make a default for `MaxCount` (and the others in this pattern) to be `max(int64)`, so it shows up in the CRD values as that at inspection time -- which I would argue is more intuitive to the end user?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -146,40 +147,56 @@ type PlayerSelector struct {
...
-// Action: "Increment" or "Decrement" the Counter's Count (optional). Must also define the Amount.
-// Amount: The amount to increment or decrement the Count (optional). Must be a positive integer.
-// Capacity: Update the maximum capacity of the Counter to this number (optional). Min 0, Max int64.
 type CounterAction struct {
-	Action   *string `json:"action,omitempty"`
-	Amount   *int64  `json:"amount,omitempty"`
-	Capacity *int64  `json:"capacity,omitempty"`
+	// Action must to either "Increment" or "Decrement" the Counter's Count. Must also define the Amount.
+	// +optional
+	Action *string `json:"action,omitempty"`
```
markmandel, line 183: (not for this release) default to increment as a quality of life improvement?

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -146,40 +147,56 @@ type PlayerSelector struct {
...
+	// Amount is the amount to increment or decrement the Count. Must be a positive integer.
+	// +optional
+	Amount *int64 `json:"amount,omitempty"`
```
markmandel, line 186: (not for this release) should this default to 1 ? I feel like it's going to be the most common value?

pkg/apis/allocation/v1/gameserverallocation.go
igooch: max(int64) = 9223372036854775807 which also seems non-intuitive. We could probably default to a round number, although I don't know what would make the most sense.

pkg/apis/allocation/v1/gameserverallocation.go
igooch: If amount != nil default to decrement? We could do that.

pkg/apis/allocation/v1/gameserverallocation.go
igooch: If Action != nil default to 1? We can do that. We don't want to default in general because the CounterAction could be to update the Capacity only, and not increment or decrement by an Amount.

igooch (approved):

gongmax (approved):

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: Yeah, that's a good point - that massive a number is ... massive 😄 okay, let's leave as is and see what feedback ends up being.

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: Ah yes good point! It could be a pure `Capacity` set. Ignore this suggestion 👍🏻

pkg/apis/allocation/v1/gameserverallocation.go
markmandel: 🤔 I also think ignore this suggestion as well - given the rest!

PR #3549, opened 2023-12-18
Delete the performance test fleet and fleetautoscaler after each test

gongmax:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

After the performance test finishes, for some reason some gameserver could be stuck in Allocated state which makes the next execution of the test will timeout because the number of Ready gameserver can not reach expected state. In the PR, we clean up the fleet and fleetautoscaler after each test finish so the next one can start with a clean state.

Closes #

### DIFF
--- a/build/performance-test.sh
+++ b/build/performance-test.sh
@@ -70,6 +70,9 @@ cat performance-test-variable.txt
 
 printf "\nStart testing."
 ./runScenario.sh performance-test-variable.txt
+
+kubectl delete -f performance-test-fleet.yaml
+kubectl delete -f performance-test-autoscaler.yaml
 printf "\nFinish testing."
 
 rm performance-test-fleet.yaml performance-test-autoscaler.yaml performance-test-variable.txt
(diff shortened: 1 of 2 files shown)


build/performance-test.sh
```diff
@@ -70,6 +70,9 @@ cat performance-test-variable.txt
 
 printf "\nStart testing."
 ./runScenario.sh performance-test-variable.txt
+
+kubectl delete -f performance-test-fleet.yaml
```
igooch, line 74: Is there a difference between this command and `kubectl delete fleet performance-test-fleet`?

igooch (approved):

build/performance-test.sh
gongmax: No difference :)

roberthbailey: `-f` means to read from a file so it will delete whatever resources are in the file.

roberthbailey (approved):

PR #3559, opened 2023-12-26
Increase performance test cluster size

gongmax:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature

/kind hotfix

> /kind release

Performance test timed out because game servers cannot be created due to lack of CPU. Increasing the test cluster size to fix it.

Closes #

### DIFF
--- a/build/terraform/performance/module.tf
+++ b/build/terraform/performance/module.tf
@@ -53,8 +53,8 @@ module "gke_standard_cluster" {
   location = each.value[0]
   overrideName = format("standard-performance-test-cluster-%s", replace(each.key, ".", "-"))
   releaseChannel = each.value[1]
-  machineType = "n2-standard-2"
-  initialNodeCount = 300
+  machineType = "e2-standard-4"
+  initialNodeCount = 310
 }
 
 resource "google_compute_firewall" "udp" {


roberthbailey (approved):

build/terraform/performance/module.tf
```diff
@@ -53,8 +53,8 @@ module "gke_standard_cluster" {
   location = each.value[0]
   overrideName = format("standard-performance-test-cluster-%s", replace(each.key, ".", "-"))
   releaseChannel = each.value[1]
-  machineType = "n2-standard-2"
-  initialNodeCount = 300
+  machineType = "e2-standard-4"
```
roberthbailey, line 56: This is going to double the number of cores in the test cluster. Do you have autoscaling enabled to scale the number of nodes down when it isn't in use?

build/terraform/performance/module.tf
```diff
@@ -53,8 +53,8 @@ module "gke_standard_cluster" {
...
+  initialNodeCount = 310
```
roberthbailey, line 57: To piggyback on the above point, if we enable node autoscaling, then we should adjust the max number of allowed nodes as the initial node count wouldn't make any difference.

roberthbailey:
As a hotfix this seems reasonable, but I'd like to make sure we don't have a bunch of idle nodes sitting around when the tests aren't running.

build/terraform/performance/module.tf
gongmax: Sorry I didn't make it quite clear in the PR description. The current test cluster is already using type "e2-standard-4" actually. The code here drifted from the fact by some mistake. So the only actual change in this PR is to increase the size to 310. 
Good point to enable autoscaling. I will test and make a quick follow up PR for it.

PR #3581, opened 2024-01-09
Adds Counters and Lists to CSharp SDK

igooch:
**What type of PR is this?**

/kind feature

Adds the C# SDK for Counters and Lists as part of the CountsAndLists feature. 

Methods Implemented are based on the Go SDK methods in https://github.com/googleforgames/agones/blob/main/sdks/go/alpha.go:

-  `func (a *Alpha) GetCounterCount(key string) (int64, error)` -> `Task<long> GetCounterCountAsync(string key);`
-  `func (a *Alpha) IncrementCounter(key string, amount int64) (error)` -> `Task IncrementCounterAsync(string key, long amount);`
-  `func (a *Alpha) DecrementCounter(key string, amount int64) (error)` ->  `Task DecrementCounterAsync(string key, long amount);`
-  `func (a *Alpha) SetCounterCount(key string, amount int64) (error)` -> `Task SetCounterCountAsync(string key, long amount);`
-  `func (a *Alpha) GetCounterCapacity(key string) (int64, error)` -> `Task<long> GetCounterCapacityAsync(string key);`
-  `func (a *Alpha) SetCounterCapacity(key string, amount int64) (error)` -> `Task SetCounterCapacityAsync(string key, long amount);`

-  `func (a *Alpha) GetListCapacity(key string) (int64, error)` -> `Task<long> GetListCapacityAsync(string key);`
-  `func (a *Alpha) SetListCapacity(key string, amount int64) (error)` -> `Task SetListCapacityAsync(string key, long amount);`
-  `func (a *Alpha) ListContains(key, value string) (bool, error)` -> `Task<bool> ListContainsAsync(string key, string value);`
-  `func (a *Alpha) GetListLength(key string) (int, error)` -> `Task<int> GetListLengthAsync(string key);`
-  `func (a *Alpha) GetListValues(key string) ([]string, error)` -> `Task<IList<string>> GetListValuesAsync(string key);`
-  `func (a *Alpha) AppendListValue(key, value string) (error)` -> `Task AppendListValueAsync(string key, string value);`
-  `func (a *Alpha) DeleteListValue(key, value string) (error)` -> `Task DeleteListValueAsync(string key, string value);`

Adds C# unit tests for the new methods. C# units tests can be run by navigating to `agones/build` and running `make sdk-shell-csharp`. Once in the C# shell navigate to the test folder `/go/src/agones.dev/agones/sdks/csharp/test`. From there use standard dotnet commands such as `dotnet test --filter Counter` to run the unit tests.

Adds C# conformance tests for the new methods. C# conformance tests can be run by navigating to `agones/build` and running `make run-sdk-conformance-test-csharp`.

Working on #2716 

### DIFF
--- a/sdks/csharp/sdk/Alpha.cs
+++ b/sdks/csharp/sdk/Alpha.cs
@@ -467,13 +467,10 @@ namespace Agones
                 };
                 var list = await client.GetListAsync(request,
                   deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
-                foreach (string val in list.Values)
+                if (list.Values.Contains(value))
                 {
-                    if (val == value)
-                    {
-                        return true;
-                    };
-                }
+                    return true;
+                };
                 return false;
             }
             catch (RpcException ex)
--- a/sdks/csharp/sdk/Alpha.cs
+++ b/sdks/csharp/sdk/Alpha.cs
@@ -20,6 +20,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Grpc.Net.Client;
+using gProto = Google.Protobuf.WellKnownTypes;
 
 [assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Agones.Test")]
 namespace Agones
@@ -387,6 +388,199 @@ namespace Agones
             catch (RpcException ex)
             {
                 LogError(ex, $"Unable to invoke SetCounterCapacity({key}, {amount}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// GetListCapacityAsync returns the Capacity for a List, given the List's key (name).
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// </summary>
+        /// <returns>The List's capacity</returns>
+        public async Task<long> GetListCapacityAsync(string key)
+        {
+            try
+            {
+                var request = new GetListRequest()
+                {
+                    Name = key,
+                };
+                var list = await client.GetListAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                return list.Capacity;
+            }
+            catch (RpcException ex)
+            {
+                LogError(ex, $"Unable to invoke GetListCapacity({key}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// SetListCapacityAsync sets the capacity for a given list. Capacity must be between 0 and 1000.
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// </summary>
+        /// <returns>True if the set List capacity request was successful.</returns>
+        public async Task<bool> SetListCapacityAsync(string key, long amount)
+        {
+            try
+            {
+                var list = new List()
+                {
+                    Name = key,
+                    Capacity = amount,
+                };
+                // FieldMask to update the capacity field only
+                var updateMask = new gProto.FieldMask()
+                {
+                    Paths = { "capacity" },
+                };
+                var request = new UpdateListRequest()
+                {
+                    List = list,
+                    UpdateMask = updateMask,
+                };
+                var response = await client.UpdateListAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                return true;
+            }
+            catch (RpcException ex)
+            {
+                LogError(ex, $"Unable to invoke SetListCapacity({key}, {amount}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// ListContainsAsync returns if a string exists in a List's values list, given the List's key
+        /// and the string value. Search is case-sensitive.
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// </summary>
+        /// <returns>True if the value is found in the List</returns>
+        public async Task<bool> ListContainsAsync(string key, string value)
+        {
+            try
+            {
+                var request = new GetListRequest()
+                {
+                    Name = key,
+                };
+                var list = await client.GetListAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                foreach (string val in list.Values)
+                {
+                    if (val == value)
+                    {
+                        return true;
+                    };
+                }
+                return false;
+            }
+            catch (RpcException ex)
+            {
+                LogError(ex, $"Unable to invoke ListContains({key}, {value}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// GetListLengthAsync returns the length of the Values list for a List, given the List's key.
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// </summary>
+        /// <returns>The length of List's values array</returns>
+        public async Task<int> GetListLengthAsync(string key)
+        {
+            try
+            {
+                var request = new GetListRequest()
+                {
+                    Name = key,
+                };
+                var list = await client.GetListAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                return list.Values.Count;
+            }
+            catch (RpcException ex)
+            {
+                LogError(ex, $"Unable to invoke GetListLength({key}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// GetListValuesAsync returns the Values for a List, given the List's key (name).
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// </summary>
+        /// <returns>The List's values array</returns>
+        public async Task<List<string>> GetListValuesAsync(string key)
+        {
+            try
+            {
+                var request = new GetListRequest()
+                {
+                    Name = key,
+                };
+                var list = await client.GetListAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                var values = new string[list.Values.Count];
+                list.Values.CopyTo(values, 0);
+                return values.ToList();
+            }
+            catch (RpcException ex)
+            {
+                LogError(ex, $"Unable to invoke GetListValues({key}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// AppendListValueAsync appends a string to a List's values list, given the List's key (name)
+        /// and the string value. Will error if the string already exists in the list.
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// </summary>
+        /// <returns>True if the append List value request was successful.</returns>
+        public async Task<bool> AppendListValueAsync(string key, string value)
+        {
+            try
+            {
+                var request = new AddListValueRequest()
+                {
+                    Name = key,
+                    Value = value,
+                };
+                var response = await client.AddListValueAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                return true;
+            }
+            catch (RpcException ex)
+            {
+                LogError(ex, $"Unable to invoke AppendListValue({key}, {value}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// DeleteListValueAsync removes a string from a List's values list, given the List's key
+        /// and the string value. Will error if the string does not exist in the list.
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// </summary>
+        /// <returns>True if the delete List value request was successful.</returns>
+        public async Task<bool> DeleteListValueAsync(string key, string value)
+        {
+            try
+            {
+                var request = new RemoveListValueRequest()
+                {
+                    Name = key,
+                    Value = value,
+                };
+                var response = await client.RemoveListValueAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                return true;
+            }
+            catch (RpcException ex)
+            {
+                LogError(ex, $"Unable to invoke DeleteListValue({key}, {value}).");
                 throw;
             }
         }
--- a/sdks/csharp/sdk/Alpha.cs
+++ b/sdks/csharp/sdk/Alpha.cs
@@ -213,7 +213,7 @@ namespace Agones
 
         /// <summary>
         /// GetCounterCountAsync returns the Count for a Counter, given the Counter's key (name).
-        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// Throws error if the key was not predefined in the GameServer resource on creation.
         /// </summary>
         /// <returns>The Counter's Count</returns>
         public async Task<long> GetCounterCountAsync(string key)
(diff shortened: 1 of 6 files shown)


sdks/csharp/sdk/Alpha.cs
```diff
@@ -210,6 +211,380 @@ public async Task<List<string>> GetConnectedPlayersAsync()
...
+            {
+                LogError(ex, $"Unable to invoke IncrementCounter({key}, {amount}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// DecrementCounterAsync decreases the current count by the given nonnegative integer amount.
+        /// The Counter Will not go below 0. Will execute the decrement operation against the current CRD value.
+        /// Returns false if the count is at 0 (to the latest knowledge of the SDK), and no decrement will occur.
```
Jensaarai, line 282: `Returns false if the count is at 0`

This method always returns true, should it be returning a value from the response object?

sdks/csharp/sdk/Alpha.cs
```diff
@@ -210,6 +211,380 @@ public async Task<List<string>> GetConnectedPlayersAsync()
...
+                LogError(ex, $"Unable to invoke GetCounterCount({key}).");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// IncrementCounterAsync increases a counter by the given nonnegative integer amount.
+        /// Will execute the increment operation against the current CRD value. Will max at max(int64).
+        /// Will error if the key was not predefined in the GameServer resource on creation.
+        /// Returns false if the count is at the current capacity (to the latest knowledge of the SDK),
```
Jensaarai, line 242: Here too, this method always returns true at the moment

sdks/csharp/sdk/Alpha.cs
Jensaarai: This seems to be the case for other bool-returning methods in this file, they may all need updates?

sdks/csharp/sdk/Alpha.cs
```diff
@@ -210,6 +211,380 @@ public async Task<List<string>> GetConnectedPlayersAsync()
...
+            {
+                var request = new GetListRequest()
+                {
+                    Name = key,
+                };
+                var list = await client.GetListAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                var values = new string[list.Values.Count];
+                list.Values.CopyTo(values, 0);
+                return values.ToList();
```
Jensaarai, line 527: Since `Values` is already a collection (a repeated field), you should be able to use `.ToList()` on it
```suggestion
                return list.Values.ToList();

```

igooch: Thanks! I'm new to C#, so this is helpful.

sdks/csharp/sdk/Alpha.cs
igooch: This is part of the weirdness of the API design #2716 expecting a `bool` return value, which doesn't match with the standard gRPC calls https://github.com/googleforgames/agones/blob/929a3e71c68d3992ba50a96f251c3047fe523b18/sdks/csharp/proto/sdk/alpha/alpha.proto#L138-L146 that take in a `UpdateCounterRequest` and return a `Counter` (similar pattern for `List` as well). The SDK server follows the same pattern of taking in an update request and return the updated object https://github.com/googleforgames/agones/blob/929a3e71c68d3992ba50a96f251c3047fe523b18/pkg/sdkserver/sdkserver.go#L849. Returning `true` if the SDK returns an object, and throwing an exception if the SDK returns an error, is how I tried to work around this. Do you have a suggestion for a better way to handle this?

Jensaarai: Interesting...honestly it doesn't seem like this SDK is that useful, for C# at least. I notice that design doc says `The SDK will batch operations every 1 second for performance reasons, but changes made through the SDK will be atomically accurate through the SDK.` but unless gRPC is doing this internally, this SDK doesn't seem to do anything of the sort.

It's just a super thin wrapper on top of a gRPC client, and at this point it seems more useful to just use the generated gRPC client itself? I suppose the SDK is helping to craft various requests here, but other than that (which already shouldn't be difficult), it's not doing much.

If this design is kept, I would expect the documentation here is at least updated to reflect the actual implementation, or the implementation updated to match the documentation, whichever is appropriate. It seems like the design doc is expecting the SDK to be doing a lot more than it currently is.

markmandel: > It's just a super thin wrapper on top of a gRPC client

all the SDKs are thin wrappers over a gRPC client. Or thin wrappers over a HTTP client. The meat of the logic is stored in the sdk server. Hence we can have multiple language SDKs without the overhead or repeating a lot of logic.

> but unless gRPC is doing this internally, this SDK doesn't seem to do anything of the sort.

That's exactly what the SDK server does. The SDK Server plus the language library are "the SDK" as a whole.

>If this design is kept, I would expect the documentation here is at least updated to reflect the actual implementation, or the implementation updated to match the documentation, whichever is appropriate. It seems like the design doc is expecting the SDK to be doing a lot more than it currently is.

Not sure I follow here. The documentation and design follow what the functionality of the SDK does - how the implementation works should be fairly immaterial to the end user.

> It seems like the design doc is expecting the SDK to be doing a lot more than it currently is.

I wrote the design doc 😄 this is doing exactly what it is supposed to.

sdks/csharp/sdk/Alpha.cs
```diff
@@ -210,6 +211,378 @@ public async Task<List<string>> GetConnectedPlayersAsync()
...
+        {
+            try
+            {
+                var request = new GetListRequest()
+                {
+                    Name = key,
+                };
+                var list = await client.GetListAsync(request,
+                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
+                foreach (string val in list.Values)
```
castaneai, line 470: NITS: `list.Values.Contains(...)` can be used.

sdks/csharp/sdk/Alpha.cs
Jensaarai: My bad, I was asked to take a quick look at this and being new to the project, I obviously missed some core architectural pieces 😅 

I suppose my feedback boils down to this: when/how should this function return false? If it never does, perhaps it should change to a void (`Task`) function?

sdks/csharp/sdk/Alpha.cs
```diff
@@ -210,6 +211,378 @@ public async Task<List<string>> GetConnectedPlayersAsync()
...
+            if (amount < 0)
+            {
+                throw new ArgumentOutOfRangeException($"CountIncrement amount must be a positive number, found {amount}");
+            }
+            try
+            {
+                var request = new CounterUpdateRequest()
+                {
+                    Name = key,
+                    CountDiff = amount,
```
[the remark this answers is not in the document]
igooch: Yes, for IncrementCounter and DecrementCounter the amount 0 is meaningless.

sdks/csharp/sdk/Alpha.cs
igooch: Would it make sense to, instead of throwing an error, to log the error and return false? Or is the void (Task) function more idiomatic in C#? (And for my new-to-C# understanding, are you recommending `public async Task<void> DecrementCounterAsync(string key, long amount)`?)

castaneai (approved):

sdks/csharp/sdk/Alpha.cs
markmandel: > I suppose my feedback boils down to this: when/how should this function return false? If it never does, perhaps it should change to a void (`Task`) function?

"Returns false if the count is at 0 (to the latest knowledge of the SDK), and no decrement will occur."

Basically a false is a "hey, this is a no-op, just letting you know"- so to this point, the SDK shouldn't be throwing an exception here.

markmandel: Digging into this a bit more - the original design of this function was such that if the counter was at capacity (0,capacity), then it should return false, not that it should error -- whereas now it throws an exception.

@igooch and I were talking about this offline, and would love your feedback on what the appropriate response here should be - @Jensaarai and @castaneai (since we have you both here), what would you prefer to happen here if you try to decrement below 0, or increment above capacity?

castaneai: My feeling is that throwing an exception (more specifically, InvalidOperationException) is more appropriate as C# API.

markmandel: @igooch I believe you were also advocating for an exception return as well? @Jensaarai how do you feel?

I was arguing  for the boolean, but happy to be out-voted if y'all are in agreement 😃 - if we are  drawing consensus to throwing an exception here, then I would also agree to returning `void` from the function, and drop the boolean entirely (and discuss probably doing this for all other SDKs as well).

Jensaarai: I'm not really familiar with Go, but it appears that its `error` is roughly equivalent to an exception in C#? https://go.dev/blog/error-handling-and-go

In that case, it seems like these functions shouldn't return a bool, and in Go a non-nil error would denote failure, while in C# it would be an exception.

What I'm not sure about is the behavior of the Go gRPC client vs. C#. For instance, here in Go, there's a call to `UpdateCounter` that can return `err`, and if it does, it gets wrapped and returned:
```Go
// DecrementCounter decreases the current count by the given nonnegative integer amount.
// The Counter Will not go below 0. Will execute the decrement operation against the current CRD value.
// Returns false if the count is at 0 (to the latest knowledge of the SDK), and no decrement will occur.
func (a *Alpha) DecrementCounter(key string, amount int64) (bool, error) {
	if amount < 0 {
		return false, errors.Errorf("CountDecrement amount must be a positive int64, found %d", amount)
	}
	_, err := a.client.UpdateCounter(context.Background(), &alpha.UpdateCounterRequest{
		CounterUpdateRequest: &alpha.CounterUpdateRequest{
			Name:      key,
			CountDiff: amount * -1,
		}})
	if err != nil {
		return false, errors.Wrapf(err, "could not decrement Counter %s by amount %d", key, amount)
	}
	return true, err
}
```

In C#, we have:
```csharp
        public async Task<bool> DecrementCounterAsync(string key, long amount)
        {
            if (amount < 0)
            {
                throw new ArgumentOutOfRangeException($"DecrementCounter amount must be a positive number, found {amount}");
            }
            try
            {
                var request = new CounterUpdateRequest()
                {
                    Name = key,
                    CountDiff = amount * -1,
                };
                var updateRequest = new UpdateCounterRequest()
                {
                    CounterUpdateRequest = request,
                };
                var response = await client.UpdateCounterAsync(updateRequest,
                  deadline: DateTime.UtcNow.AddSeconds(RequestTimeoutSec), cancellationToken: ctoken);
                return true;
            }
            catch (RpcException ex)
            {
                LogError(ex, $"Unable to invoke DecrementCounter({key}, {amount}).");
                throw;
            }
        }
```

I imagine that here, the `UpdateCounterAsync` gRPC method would throw an `RpcException` in the same `err` case that Go is handling? In which case, I think the re-throw is already appropriate, and thus the boolean is not that useful.

Additionally, I think the `ArgumentOutOfRangeException` for `amount < 0` is the appropriate thing for C# to do :)

sdks/csharp/sdk/Alpha.cs
Jensaarai: Alternatively, if this method was changed to return `Task<IReadOnlyList<string>>`, you could just `return list.Values;` because `Values` is a `repeated` field https://github.com/googleforgames/agones/blob/main/proto/sdk/alpha/alpha.proto#L276 and in the C# protobuf library, `RepeatedField<T>` implements `IReadOnlyList<T>` https://github.com/protocolbuffers/protobuf/blob/main/csharp/src/Google.Protobuf/Collections/RepeatedField.cs#L31

sdks/csharp/sdk/Alpha.cs
igooch: Is the recommended action here to update `public async Task<bool> DecrementCounterAsync(string key, long amount)` to `public async Task DecrementCounterAsync(string key, long amount)`? If it is, @Jensaarai @castaneai do you have recommendations for how to update the tests? (I'm running into issues like ` error CS0029: Cannot implicitly convert type 'void' to 'System.Threading.Tasks.Task'`.)

castaneai: Yes. Changing to `Task` is correct. In the test, you don't need to use response, only await.

markmandel: @igooch would you like me to write up a ticket to update the SDK methods globally to remove the boolean from count increment/decrement operations?

igooch: Yes, @markmandel please write up a ticket to update the bool from the operations that don't use it. I removed it from the IncrementCounterAsync, DecrementCounterAsync, SetCounterCountAsync, SetCounterCapacityAsync, SetListCapacityAsync, AppendListValueAsync, and DeleteListValueAsync.

igooch: @castaneai @Jensaarai the change from `public async Task<bool>` to `public async Task` is ready for review when you have the time.

Jensaarai (approved):
Changes look good to me!

sdks/csharp/sdk/Alpha.cs
igooch: @castaneai did you want to review one last time before we merge the PR?

markmandel (commented):
Just realised - https://agones.dev/site/docs/guides/client-sdks/csharp/ needs some docs please 😄

site/content/en/docs/Guides/Client SDKs/csharp.md
```diff
@@ -14,35 +14,35 @@ Check the [Client SDK Documentation]({{< relref "_index.md" >}}) for more detail
...
-| Lists           | GetListLength       | ❌         |
-| Lists           | GetListValues       | ❌         |
-| Player Tracking | GetConnectedPlayers | ✔️          |
-| Player Tracking | GetPlayerCapacity   | ✔️          |
-| Player Tracking | GetPlayerCount      | ✔️          |
-| Player Tracking | IsPlayerConnected   | ✔️          |
-| Player Tracking | PlayerConnect       | ✔️          |
-| Player Tracking | PlayerDisconnect    | ✔️          |
-| Player Tracking | SetPlayerCapacity   | ✔️          |
+| Lifecycle       | Ready               | ✔️           |
```
markmandel, line 17: `feature` shortcodes please! (may be easiest to duplicate the table).

site/content/en/docs/Guides/Client SDKs/csharp.md
```diff
@@ -201,6 +201,155 @@ var playerId = "player1";
 bool isConnected = await agones.Alpha().IsPlayerConnectedAsync(playerId);
 ```
 
+## Counters
```
markmandel, line 204: This section should be `feature` shortcoded as well please.

site/content/en/docs/Guides/Client SDKs/csharp.md
```diff
@@ -201,6 +201,155 @@ var playerId = "player1";
...
+
+```csharp
+string key = "rooms";
+long count = await agones.Alpha().GetCounterCapacityAsync(key);
+```
+## Lists
+
+### Alpha: AppendListValue
+
+Appends a string to a List's values list, given the List's key (name) and the string value. Will
```
markmandel, line 279: Will also error if it goes over capacity?

markmandel (commented):
Just need `feature` shortcodes in the docs, then I think this is good to merge.

site/content/en/docs/Guides/Client SDKs/csharp.md
igooch: Yep, yes it does.

site/content/en/docs/Guides/Client SDKs/csharp.md
igooch: Added that in

markmandel:
Really need to fix this flake:

```
E: time="2024-04-02 00:48:53.552" level=info msg="Finished Allocation."
VERBOSE:     gameserverallocation_test.go:1442: 
VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserverallocation_test.go:1442
VERBOSE:         	Error:      	Not equal: 
VERBOSE:         	            	expected: 100
VERBOSE:         	            	actual  : 99
VERBOSE:         	Test:       	TestGameServerAllocationDuringMultipleAllocationClients
VERBOSE: --- FAIL: TestGameServerAllocationDuringMultipleAllocationClients (60.57s)
```

markmandel:
@zmerlynn can you do a final approval on this? I just pushed up some changes to Ivy's branch and felt weird approving my own changes 😄

zmerlynn (approved):
I approve your last commit. :)

PR #3594, opened 2024-01-17
Counters: Update Default Capacity

### LINKED ISSUE #3579: Counters: Default capacity should be 0 (max), not 1000
**What happened**:

Creating a Fleet with the following yaml:

```yaml
apiVersion: agones.dev/v1
kind: Fleet
metadata:
  name: simple-game-server
spec:
  replicas: 2
  template:
    spec:
      ports:
        - name: default
          containerPort: 7654
      counters:
        rooms:
          count: 0
      lists:
        players:
          values: []
      template:
        spec:
          containers:
            - name: simple-game-server
              image: us-docker.pkg.dev/agones-images/examples/simple-game-server:0.23
              resources:
                requests:
                  memory: 64Mi
                  cpu: 20m
                limits:
                  memory: 64Mi
                  cpu: 20m
```

When I look at the resulting yaml output, I see that the default capacity is 1000 in the counter status:

```yaml
status:
  address: XX.XXX.XXX.XXX
  addresses:
  - address: XX.XXX.XXX.XXX
    type: InternalIP
  - address: XX.XXX.XXX.XXX
    type: ExternalIP
  - address: gke-test-cluster-default-393bd66b-swtm
    type: Hostname
  counters:
    rooms:
      capacity: 1000 # ⬅️ is the issue
      count: 0
  eviction:
    safe: Never
  immutableReplicas: 1
  lists:
    players:
      capacity: 1000
      values: []
  nodeName: gke-test-cluster-default-393bd66b-swtm
  ports:
  - name: default
    port: 7889
  reservedUntil: null
  state: Ready
```

[Link to relevant design](https://github.com/googleforgames/agones/issues/2716#:~:text=Counters%20can%20have%20a%20set%20capacity%20(maximum%20value)%2C%20but%20by%20default%20are%200%20(max%20of%20int64))

`capacity` should be 0, since it's a counter, and we don't need to limit its size for performance reasons.

Apply yaml above, then `kubectl get gs <name> -o yaml` to see the result.

N/A

**Environment**:
- Agones version:   1.37.0
- Kubernetes version (use `kubectl version`): v1.27.7-gke.1056000
- Others: N/A

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #3579 


igooch:
Try running `make gen-install` from within the agones/build directory.

Kalaiselvi84:
shall I revert the change in `install/yaml/install.yaml` and try `make gen-install`?

Kalaiselvi84:
I have tested with Counters default capacity with `9223372036854775807` and `0`.

The Counters info are missing in the status with [9223372036854775807](https://gist.github.com/Kalaiselvi84/f164f6501bccfbe453234fbb2075445f)

Counters are present with [0](https://gist.github.com/Kalaiselvi84/f7c140d0bf5e26c5e52c66cc1753dd84)

igooch:
The TL;DR is currently our max capacity per game server = max(int64) / replicas.

The fleet aggregates the count and capacity of all the game servers counters in the fleet. So if game server counter capacity = max(int64) and there are 3 replicas in the fleet, then the fleet aggregate counter capacity = 3 * max(int64) which errors. When I put in max(int64) / 3 as the max capacity for a game server, then the fleet and game servers started up just fine. 

We'll need to change the code, and documentation, so that if the aggregate capacity or count for the fleet goes above max(int64), then the aggregate would be max(int64), so that we don't get an error on aggregation. We could also update fleet to aggregate up to uint64 to give the fleet a bit more headroom. We should probably do this as a separate PR before updating the defaults.

igooch:
Opened #3604 to address the aggregating past int64 issue.

markmandel (commented):
Just holding off on this one until we decide if we should make count capacity `nullable` and the Go model *capacity to indicate essentially unlimited capacity.

@igooch @zmerlynn WDYT? (from chat discussion)

Kalaiselvi84:
[This link](https://stackoverflow.com/questions/30833314/golang-overflows-int64) shows the example using `big.NewInt` to handle the larger number.

markmandel:
Just checking - but @igooch @zmerlynn we can mere this as is now?

If I really want the max capacity to be represented by `nil` once I've looked at the developer experience, I'll roll up my sleeves and make the chage.

Sound good?

Issue #3593, opened 2024-01-16
SDK server does not remove items from lists ( new 1.37 feature )


jlory:
**What happened**:

I'm currently heavily testing the new counters and lists feature from Agones 1.37, I noticed that removing an item from a list does not remove the item on the GameServer CRD.

**What you expected to happen**:

I assume that it should be be possible to remove an item from a list and that it will be reflected in the CRD status object.

**How to reproduce it (as minimally and precisely as possible)**:

Create a fleet with a list, add an item to that list, then remove it ( from an sdk so it goes to the sdk server ), describe the GameServer object with kubectl, the item is still in the list.

**Anything else we need to know?**:

I digged into the sdk server and I believe I found the issue, the sdk server holds a list of updates that need to go out https://github.com/googleforgames/agones/blob/v1.37.0/pkg/sdkserver/sdkserver.go#L132 later on we have the function that update the CRD to the k8s api, if it's a success we then clean the counters update buffer but not the list one: https://github.com/googleforgames/agones/blob/v1.37.0/pkg/sdkserver/sdkserver.go#L1233

I did test locally and it's a one liner change minus unit tests:

```go
// Clear the gsCounterUpdates and gsListUpdates
s.gsCounterUpdates = map[string]counterUpdateRequest{}
s.gsListUpdates = map[string]listUpdateRequest{}
```

If the issues is ack by you, I will submit a PR very soon.

**Environment**:
- Agones version: 1.37
- Kubernetes v1.26.11-eks-8cb36c9
- Cloud provider or hardware configuration: AWS
- Install method (yaml/helm): Helm
- Troubleshooting guide log(s):

jlory:
Will re-create that issue~~

igooch:
Thanks for trying it out and catching that! @markmandel can review the PR.

jlory:
Closing this issue since it was fixed by: https://github.com/googleforgames/agones/pull/3606

PR #3606, opened 2024-01-23
SDK server not clearing lists on update

jlory:
**What type of PR is this?**
>
> /kind breaking
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

/kind bug

Fixes the issue where items in list could not be removed.

https://github.com/googleforgames/agones/issues/3593

Had to recreate the PR because of CLA issues.


pkg/sdkserver/sdkserver.go
```diff
@@ -1230,8 +1230,8 @@ func (s *SDKServer) updateList(ctx context.Context) error {
 
 	// Cache a copy of the successfully updated gameserver
 	s.gsCopy = gs
-	// Clear the gsCounterUpdates
-	s.gsCounterUpdates = map[string]counterUpdateRequest{}
+	// Clear the gsListUpdates
+	s.gsListUpdates = map[string]listUpdateRequest{}
```
markmandel, line 1234: Do we need an update to a unit test to ensure this doesn't happen again?

jlory: I'll have a look in the unit test file to see what's doable.

markmandel: Appreciate you taking a look.

No rush, but just a heads up - we're cutting a release next Tuesday, so if you want this in the next release, we should get some unit tests in.

@igooch any advice on a good spot to write a unit test?

jlory: I have something actually but it's at work, so tomorrow I should send an update to this MR, idea is:

- add a function that get the len() of the outgoing lists
- happy path = len() == 0
- network failure / issues updating = original len

igooch: Looks good! The only thing I'd add is checking that the `gsListUpdates` has the update, i.e. is not zero after the update/add/remove call and before the get list call. Something like

```
if testCase.updated {
  updates := sc.gsListUpdatesLen()
  assert.Equal(t, 1, updates)
}
```

igooch: ^Scratch that above suggestion, it would introduce a race condition. Code looks good as it stands.

igooch (approved):
LGTM

PR #3608, opened 2024-01-24
WIP: Hacking on webhook int64-ness

zmerlynn:
WIP - just putting it up for debugging

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -16,7 +16,6 @@ package gameservers
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strconv"
 	"sync"
@@ -48,6 +47,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	runtimeschema "k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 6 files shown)


pkg/gameservers/controller.go
```diff
@@ -253,24 +253,26 @@ func (ext *Extensions) creationMutationHandler(review admissionv1.AdmissionRevie
...
+	ext.baseLogger.Infof("ZZZML AFTER Status.Counters: %v", gs.Status.Counters)
 
 	newGS, err := json.Marshal(gs)
 	if err != nil {
 		return review, errors.Wrapf(err, "error marshalling default applied GameServer %s to json", gs.ObjectMeta.Name)
 	}
+	ext.baseLogger.Infof("ZZZML AFTER JSON: %s", string(newGS))
 
-	patch, err := jsonpatch.CreatePatch(obj.Raw, newGS)
+	jsonPatch, err := jsonpatch.CreateMergePatch(obj.Raw, newGS)
```
markmandel, line 271: So the issue here I think, is that this type of patch isn't supported.

See:
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#:~:text=When%20allowing%20a,JSON%20patch%20operations.

This creates a https://datatracker.ietf.org/doc/html/rfc7396 style merge patch, whereas AdmissionReviews can only handle https://jsonpatch.com/ types of JSON Patches.

See also: https://kubernetes.io/docs/reference/config-api/apiserver-admission.v1/#admission-k8s-io-v1-AdmissionResponse

markmandel: So I did a bit more digging, and I wondered "what does `controller-runtime` use, because they have the same problem!

https://github.com/kubernetes-sigs/controller-runtime/blob/b88ed7a3602b85b6cfd0acfe9c25033b978bdb83/go.mod#L23

Which is: https://github.com/gomodules/jsonpatch

So in theory, should be a drop in replacement without any code changes other than go.mod

markmandel:
Just checking in if you were going to finish this up?

Or should we turn this into an issue and close out the PR to be tackled at a later date?

zmerlynn:
I think we should abandon it. I'm not even sure it's worth an issue, and here's why: JSON support for high int64 values seems fussy at best. If I had this much trouble in golang, there are probably other languages where it wouldn't translate well, either, which is concerning. :/

PR #3612, opened 2024-01-25
Nodepool upgrades on GKE Terraform apply

### LINKED ISSUE #3339: [GKE Standard/Terraform] Cluster nodes aren't upgrading with the control plane
**What happened**:

`terraform apply` the [GKE Terraform modules](https://github.com/googleforgames/agones/tree/main/install/terraform/modules/gke) repeatedly over multiple Agones versions, I am noticing that while my Kubernetes control plane version is updating, the nodes aren't being updated.

Noticed this in the Global Scale Game demo: https://github.com/googleforgames/global-multiplayer-demo/issues/190 my Control Plane was `1.27.3-gke.100` but my nodes, created a while ago remained on `1.24.9-gke.3200`.

This is the source: https://github.com/googleforgames/global-multiplayer-demo/blob/main/infrastructure/agones-gke.tf

This doesn't seem to be an issue with Autopilot.

Nodes to be upgraded with the Control Plane.

Create a cluster with the Terraform module with an older version set of Kuberntes, then change the Kubernetes version and apply again. The node version won't change.

**Environment**:
- Agones version:  1.33.0
- Kubernetes version (use `kubectl version`): 1.27.x
- Others: N/A

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Per the docs:
"...node pools defined inside a cluster can't be changed (or added/removed) after cluster creation without deleting and recreating the entire cluster."

Which is not great - since you can end up with out-of-sync K8s versions between the control plane and nodes, an inability to change nodepool sizes and just a general lack of flexibility.

Moving the node pool definitions out of the cluster definition solves this issue!

Closes #3339

This is the same terraform we use for our dev clusters on GKE.


markmandel:
I dropped a `breaking` on this as it changes the structure of the Terraform, so we should call this out in the release notes.

I don't expect it to be an issue since the older version wouldn't let you update nodes anyway, but better to be sure.

gongmax (approved):

Issue #3570, opened 2024-01-08
Upgrade from Debian Bullseye to Bookworm


markmandel:
Just noticing we're using Debian 11 everywhere. It's currently being kept up to date with security fixes, but it would be good to be proactive and update to Debian 12.

This can definitely be done as a series of smaller PR's.

For:
https://github.com/search?q=repo%3Agoogleforgames%2Fagones%20debian%3Abullseye&type=code
Switch from `debian:bullseye` to `debian:bookworm`

For:
https://github.com/search?q=repo%3Agoogleforgames%2Fagones%20debian11&type=code
Switch from `gcr.io/distroless/static-debian11:nonroot` to `gcr.io/distroless/static-debian12:nonroot`

markmandel:
@Kalaiselvi84 this would likely be good to for you to work on if you have run out of tasks 👍🏻

Kalaiselvi84:
@markmandel, I've a few things to discuss:
1. I have updated most of the example images. Will push the new image to the production once the PRs are merged and then create a separate PR to update the rest of the files with new tag.

2. The [build](https://paste.googleplex.com/create/) is failing with `bookworm` for `cpp-simple` image. Log is too large to fit in gist.github.com, so I used paste instead.

3. Could you please let me know how to test the changes in the `/build` and `/cmd` directories?

markmandel:
> Could you please let me know how to test the changes in the /build and /cmd directories?

The standard CI process will test those -- or you can do a `make build-images && make push && make install` locally and see if Agones is up and running.

Kalaiselvi84:
The command `make build-images && make push && make install` is failing in both directories!

markmandel:
Should be run from `/build` - I'll check the draft PR's you wrote, see what I can see 👍🏻

markmandel:
> 2\. The [build](https://paste.googleplex.com/create/) is failing with `bookworm` for `cpp-simple` image. Log is too large to fit in gist.github.com, so I used paste instead.

Doesn't look like I got a gist / paste. Would likely need to see a draft PR and the output to help debug.

markmandel:
Just a reminder that once #3576 is merged and published, will need to replace all the `simple-game-server:0.23` instances with `simple-game-server:0.24` with a follow up PR (e2e tests, docs, etc)

markmandel:
[...]

PR #3621, opened 2024-01-31
Breaking: Remove Cmake gRPC install when not found

markmandel:
**What type of PR is this?**
>

/kind breaking

> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

gRPC upgrades with gRPC where very brittle, and attempts to upgrade dependencies and/or switch to a `FetchContent_Declare` type workflow were becoming increasingly time-consuming - and not resulting with a working solution.

Speaking with the gRPC team, they recommended not installing dependencies within CMake (and how they do it admittedly an antipattern). So this change removes the ability for the CMakeLists.txt to install gRPC and makes finding the gRPC dependency a hard requirement, and updates the documentation accordingly.

This change isn't onerous to the end user, as our SDK has only the one dependency - gRPC (which does install all its dependencies). At some point gRPC will have official support for vcpkg, at which point we should probably move with them - but in the meantime, the cpp-simple example has been updated with this change, and it's an extra ~4 lines.

This update also include a lot of other fixes:
* Upgrade from Debian Bullseye to Bookworm for all cpp related containers
* Builds now actually happen in parallel, reducing build time to minutes not hours.
* Bump to gRPC 1.57.1
* Fix bug with gRPC dependency utf8_range.
* Move to C++17, because it's required for gRPC.

Work on #3570

This was over a week of my life 😭


.github/ISSUE_TEMPLATE/kubernetes_update.md
```diff
@@ -39,10 +39,9 @@ List of items to do for upgrading to {version_1} {version_2} {version_3}
         - [ ] `dev_minikube_example_cluster_version`, which is {version_2} with the supported patch version
 - [ ] If client-go pulled in a new version of gRPC, then also
     - [ ] Update the `grpc_release_tag` in the SDK [base image grpc version](https://github.com/googleforgames/agones/blob/main/build/includes/sdk.mk).
-    - [ ] Update the gRPC version number in C++ CMake scripts, `AGONES_GRPC_VERSION` [here](https://github.com/googleforgames/agones/blob/main/sdks/cpp/CMakeLists.txt)
-          and `gRPC_GIT_TAG` [here](https://github.com/googleforgames/agones/blob/main/sdks/cpp/cmake/prerequisites.cmake)
+    - [ ] Update the gRPC version number in C++ gRPC Dependency documentation [here](https://github.com/googleforgames/agones/blob/release-1.38.0/site/content/en/docs/Guides/Client%20SDKs/cpp.md).
```
roberthbailey, line 42: The link should point at the main branch instead of a release branch.

build/build-sdk-images/restapi/Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN wget -q https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
 
 RUN echo '§' && apt-get -qy update
 
-RUN apt-get install -qq -y openjdk-11-jre > /dev/null
+RUN apt-get install -qq -y openjdk-17-jre > /dev/null
```
roberthbailey, line 31: I know it's just one line, but is this related to the cmake changes or should it be a separate review?

.github/ISSUE_TEMPLATE/kubernetes_update.md
roberthbailey, line 42: The version number is hardcoded in `examples/cpp-simple/Dockerfile` as well now. Maybe make that an arg in the docker file near the top to make it easy to update via the checklist?

build/build-sdk-images/restapi/Dockerfile
markmandel: Good call out - So it has be done with the upgrade to bookworm at the same time, since there is no `openjdk-11-jre` on bookworm - so everything fails.

.github/ISSUE_TEMPLATE/kubernetes_update.md
markmandel: Gah, thanks! Updating.

markmandel: I thought about this - but then you will end up with another hardcoded value in the `Makefile` - so either way, you end up with a hardcoded value. I erred on the side of readability for the end user when looking at the example, rather than the user having to parse an `ARG` in the Dockerfile and then winding back to wherever that was specified.

markmandel: Blarg. I swore I did that! On it.

zmerlynn (approved):

roberthbailey (approved):

PR #3626, opened 2024-02-01
Script to bump example images

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #


build/scripts/bump-image/main.go
```diff
@@ -0,0 +1,106 @@
...
+		"install":  true,
+		"pkg":      true,
+		"site":     true,
+		"test":     true,
+	}
+)
+
+func init() {
+	flag.StringVar(&imageName, "imageName", "", "Image name to update")
+	flag.StringVar(&version, "version", "", "Version to update to")
```
gongmax, line 45: If I understood correctly, the `version` here is the version we are going to update from, instead of update to, right?

Kalaiselvi84: This flag specifies the version that we want to update to the new tag. However, I am open to changing its description to either `version to update from` or simply `version to update` if that makes more sense.

gongmax (approved):
LGTM

PR #3627, opened 2024-02-01
Flake: List Add/Delete Unit Tests

### LINKED ISSUE #3622: Flaky: unit test TestSDKServerAddListValue
**What happened**:
```
Step #20 - "tests": --- FAIL: TestSDKServerAddListValue (31.98s)
Step #20 - "tests":     --- FAIL: TestSDKServerAddListValue/Add_multiple_values_past_capacity (1.12s)
Step #20 - "tests":         sdkserver_test.go:1477: 
Step #20 - "tests":             	Error Trace:	/go/src/agones.dev/agones/pkg/sdkserver/sdkserver_test.go:1477
Step #20 - "tests":             	Error:      	Not equal: 
Step #20 - "tests":             	            	expected: v1.ListStatus{Capacity:10, Values:[]string{"one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten"}}
Step #20 - "tests":             	            	actual  : v1.ListStatus{Capacity:10, Values:[]string{"one", "two", "three", "four", "five", "six", "seven", "eight", "nine"}}
Step #20 - "tests":             	            	
Step #20 - "tests":             	            	Diff:
Step #20 - "tests":             	            	--- Expected
Step #20 - "tests":             	            	+++ Actual
Step #20 - "tests":             	            	@@ -2,3 +2,3 @@
Step #20 - "tests":             	            	  Capacity: (int64) 10,
Step #20 - "tests":             	            	- Values: ([]string) (len=10) {
Step #20 - "tests":             	            	+ Values: ([]string) (len=9) {
Step #20 - "tests":             	            	   (string) (len=3) "one",
Step #20 - "tests":             	            	@@ -11,4 +11,3 @@
Step #20 - "tests":             	            	   (string) (len=5) "eight",
Step #20 - "tests":             	            	-  (string) (len=4) "nine",
Step #20 - "tests":             	            	-  (string) (len=3) "ten"
Step #20 - "tests":             	            	+  (string) (len=4) "nine"
Step #20 - "tests":             	            	  }
Step #20 - "tests":             	Test:       	TestSDKServerAddListValue/Add_multiple_values_past_capacity
```

Tests should pass.

Not entirely sure. Seems to mostly happen in CI, have yet to try and reproduce locally.

**Environment**:
- Agones version: dev
- Kubernetes version (use `kubectl version`): varies
- Troubleshooting guide log(s): see above what happened section
- Others: N/A

### LINKED ISSUE #3625: Flaky: TestSDKServerRemoveListValue
**What happened**:

```
--- FAIL: TestSDKServerRemoveListValue (16.32s)
    --- FAIL: TestSDKServerRemoveListValue/Remove_all_values (1.12s)
        sdkserver_test.go:1626: 
            	Error Trace:	/go/src/agones.dev/agones/pkg/sdkserver/sdkserver_test.go:1626
            	Error:      	Not equal: 
            	            	expected: v1.ListStatus{Capacity:100, Values:[]string{}}
            	            	actual  : v1.ListStatus{Capacity:100, Values:[]string{"one"}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,3 +2,4 @@
            	            	  Capacity: (int64) 100,
            	            	- Values: ([]string) {
            	            	+ Values: ([]string) (len=1) {
            	            	+  (string) (len=3) "one"
            	            	  }
            	Test:       	TestSDKServerRemoveListValue/Remove_all_values
        sdkserver_test.go:1636: 
            	Error Trace:	/go/src/agones.dev/agones/pkg/sdkserver/sdkserver_test.go:1636
            	Error:      	Not equal: 
            	            	expected: 0
            	            	actual  : 1
            	Test:       	TestSDKServerRemoveListValue/Remove_all_values
```

Tests to pass

Not entirely sure. Seems to mostly happen in CI, have yet to try and reproduce locally.

**Environment**:
- Agones version: dev
- Kubernetes version (use `kubectl version`): varies
- Troubleshooting guide log(s): Log included in `What happend` section
- Others: N/A

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Both unit tests had the same issue - under high load, the workqueue could process the current data set that was batched for updates, resulting in non-deterministic test state.

Moving the `sc.Run(...)` operation until only the moment where update checks are required resolves this issue.

Fixes #3625
Fixes #3622

### DIFF
--- a/pkg/sdkserver/sdkserver_test.go
+++ b/pkg/sdkserver/sdkserver_test.go
@@ -1437,16 +1437,8 @@ func TestSDKServerAddListValue(t *testing.T) {
 			require.NoError(t, err)
 			assert.NoError(t, sc.WaitForConnection(ctx))
 			sc.informerFactory.Start(ctx.Done())
-			assert.True(t, cache.WaitForCacheSync(ctx.Done(), sc.gameServerSynced))
-
-			wg := sync.WaitGroup{}
-			wg.Add(1)
-
-			go func() {
-				err = sc.Run(ctx)
-				assert.NoError(t, err)
-				wg.Done()
-			}()
+			require.True(t, cache.WaitForCacheSync(ctx.Done(), sc.gameServerSynced))
+			sc.gsWaitForSync.Done()
 
 			// check initial value comes through
 			require.Eventually(t, func() bool {
@@ -1469,6 +1461,14 @@ func TestSDKServerAddListValue(t *testing.T) {
 			assert.Equal(t, testCase.want.Values, got.Values)
 			assert.Equal(t, testCase.want.Capacity, got.Capacity)
 
+			// start workerqueue processing at this point, so there is no chance of processing the above updates
+			// earlier.
+			sc.gsWaitForSync.Add(1)
+			go func() {
+				err = sc.Run(ctx)
+				assert.NoError(t, err)
+			}()
+
 			// on an update, confirm that the update hits the K8s api
 			if testCase.updated {
 				select {
@@ -1487,7 +1487,6 @@ func TestSDKServerAddListValue(t *testing.T) {
 			assert.Equal(t, testCase.expectedUpdatesQueueLen, glu)
 
 			cancel()
-			wg.Wait()
 		})
 	}
 }
@@ -1586,16 +1585,8 @@ func TestSDKServerRemoveListValue(t *testing.T) {
 			require.NoError(t, err)
 			assert.NoError(t, sc.WaitForConnection(ctx))
 			sc.informerFactory.Start(ctx.Done())
-			assert.True(t, cache.WaitForCacheSync(ctx.Done(), sc.gameServerSynced))
-
-			wg := sync.WaitGroup{}
-			wg.Add(1)
-
-			go func() {
-				err = sc.Run(ctx)
-				assert.NoError(t, err)
-				wg.Done()
-			}()
+			require.True(t, cache.WaitForCacheSync(ctx.Done(), sc.gameServerSynced))
+			sc.gsWaitForSync.Done()
 
 			// check initial value comes through
 			require.Eventually(t, func() bool {
@@ -1618,6 +1609,14 @@ func TestSDKServerRemoveListValue(t *testing.T) {
 			assert.Equal(t, testCase.want.Values, got.Values)
 			assert.Equal(t, testCase.want.Capacity, got.Capacity)
 
+			// start workerqueue processing at this point, so there is no chance of processing the above updates
+			// earlier.
+			sc.gsWaitForSync.Add(1)
+			go func() {
+				err = sc.Run(ctx)
+				assert.NoError(t, err)
+			}()
+
 			// on an update, confirm that the update hits the K8s api
 			if testCase.updated {
 				select {
@@ -1636,7 +1635,6 @@ func TestSDKServerRemoveListValue(t *testing.T) {
 			assert.Equal(t, testCase.expectedUpdatesQueueLen, glu)
 
 			cancel()
-			wg.Wait()
 		})
 	}
 }


roberthbailey (approved):

pkg/sdkserver/sdkserver_test.go
```diff
@@ -1469,6 +1461,14 @@ func TestSDKServerAddListValue(t *testing.T) {
 			assert.Equal(t, testCase.want.Values, got.Values)
 			assert.Equal(t, testCase.want.Capacity, got.Capacity)
 
+			// start workerqueue processing at this point, so there is no chance of processing the above updates
+			// earlier.
+			sc.gsWaitForSync.Add(1)
```
zmerlynn, line 1466: Given that the go routine was moved down, do we need the `Done()` and `Add(1)`? You already moved the `sc.Run`.

markmandel: @igooch asked me offline the same question.

So in this test we're faking things a bit, since we have this WaitGroup in the sdkserver:

https://github.com/googleforgames/agones/blob/5f58fce4c00f97ef012db8f8a503898a4725d11b/pkg/sdkserver/sdkserver.go#L242

This WaitGroup exists to pause any requests for `sdk.GameServer()` data within the SDK until the SDK has received this information from the Informer.

https://github.com/googleforgames/agones/blob/5f58fce4c00f97ef012db8f8a503898a4725d11b/pkg/sdkserver/sdkserver.go#L452

In this test, we manually start the pieces of the client-go machinery that we need (which is usually done in `sc.Run(...)` ~ line 1440, but ignore the pieces we don't want (primarily starting the workerqueue).

```
 			sc.informerFactory.Start(ctx.Done())
			require.True(t, cache.WaitForCacheSync(ctx.Done(), sc.gameServerSynced))
			sc.gsWaitForSync.Done()
```

So we're calling `sc.gsWaitForSync.Done()` because we know that the `InformerFactory` has started, we've waited for `WaitForCacheSync`, so the information is there, and `sc.gameServer()` 

But this does mean, that when we get to calling `sc.Run(...)` this will call `sc.gsWaitForSync.Done()` (as above), which will panic, as there is nothing for it to decrement on on. So we `sc.gsWaitForSync.Add(1)` to give the WaitGroup something to decrement on with `.Done()`, even though most of the work being done in `sc.Run(...)` is redundant at this point.

Does that answer the question?

zmerlynn: Not exactly. I understand where the existing sync points are and that there's a `gsWaitForSync.Done()` called in `sc.Run()`. But here on new line 1441 you are calling `Done()`- but there's no one that would be waiting here. You are then undoing that on new line 1466, and not starting `sc.Run()` until after that.

I'm arguing that I _think_ you can remove 1441 and 1466 and have the same guarantee - you got that guarantee by moving the `sc.Run()` lower. Unless there is something async running in new lines 1443-1462 that I'm missing?

markmandel: > but there's no one that would be waiting here

There's a .Add(1) in the inital construction of the object:

https://github.com/markmandel/agones/blob/cbbe83f7351bb9c11d83c1c98e0924bd8b41da22/pkg/sdkserver/sdkserver.go#L218

And then in this code:

https://github.com/markmandel/agones/blob/cbbe83f7351bb9c11d83c1c98e0924bd8b41da22/pkg/sdkserver/sdkserver.go#L1011-L1020

This call to `s.gameServer()` would hang without the call to `gsWaitForSync.Done()` since it has his this in it:

https://github.com/markmandel/agones/blob/cbbe83f7351bb9c11d83c1c98e0924bd8b41da22/pkg/sdkserver/sdkserver.go#L451

zmerlynn: Oh, I missed that `GetList` was calling `s.gameServer()` under the hood - that clarifies it. (I was misreading it as something else. :/)

Sounds good, understood!

PR #3628, opened 2024-02-02
Adds simple game server for gen AI

igooch:
**What type of PR is this?**

/kind documentation

Creates a simple game server that can connect to a gen AI endpoint.


markmandel:
🤔  should this be in this repo, or something like https://github.com/googleforgames/GenAI-quickstart which is more tied to Google Cloud products?

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,288 @@
...
+
+	<-sigCtx.Done()
+	os.Exit(0)
+}
+
+func initClient(endpoint string) (*connection, error) {
+	// TODO: create option for a client certificate
+	client := &http.Client{}
+	conn := &connection{client: client, endpoint: endpoint}
+	return conn, nil
```
roberthbailey, line 95: This always returns a nil error. Unless there is a chance of producing an error in this function it can be simplified to just return the connection.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,288 @@
...
+type connection struct {
+	client   *http.Client
+	endpoint string // full base URL for API requests
+	// TODO: create options for routes off the base URL
+}
+
+type GenAIRequest struct {
+	MaxOutputTokens int     `json:"max_output_tokens"`
+	Prompt          string  `json:"prompt"`
+	Temperature     float64 `json:"temperature"`
```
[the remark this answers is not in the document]
igooch: This structure in general will probably only work with Vertex as there is no standard API structure.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,288 @@
...
+	"agones.dev/agones/pkg/util/signals"
+	sdk "agones.dev/agones/sdks/go"
+)
+
+// main starts a UDP or TCP server
+func main() {
+	sigCtx, _ := signals.NewSigKillContext()
+
+	port := flag.String("port", "7654", "The port to listen to traffic on")
+	udp := flag.Bool("udp", true, "Server will listen on UDP")
```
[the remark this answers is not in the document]
igooch: Are you OK with just having TCP in that case? UDP will truncate the longer responses.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,288 @@
...
+
+	log.Print("Creating SDK instance")
+	s, err := sdk.NewSDK()
+	if err != nil {
+		log.Fatalf("Could not connect to sdk: %v", err)
+	}
+
+
+	log.Print("Starting Health Ping")
+	// TODO: Is "cancel" still being used even if no method explicitly uses it?
```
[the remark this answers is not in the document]
igooch: The question is more around if we actually need the variable "cancel". It's passed through to the UDP and TCP listener functions that were copied over from examples/simple-game-server/main.go, but those functions don't ever use the parameter. Would it make more sense to do `ctx, _ := signals.NewSigKillContext()`?

examples/simple-genai-server/main.go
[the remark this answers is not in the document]
roberthbailey: If you are pointing this at the api from https://github.com/googleforgames/GenAI-quickstart how does the request work if it's directed to an LLM running on k8s instead of Vertex? Do known fields in the json get dropped?

Maybe it's too early and we need to experiment, but do we expect this to end up as a union of all possible prompt parameters?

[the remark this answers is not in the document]
igooch: It depends on if their endpoint will accept extraneous values in the json request? We could go the opposite direction and remove all but the prompt from the struct. Although, that still assumes that the endpoint variable name is "prompt". The most "size fits all" way would to have the customer pass in the whole json request body. But if we do that then we lose the ability easily modify each request's prompt. If you have a better way of doing this please let me know.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,288 @@
...
+	for scanner.Scan() {
+		tcpHandleCommand(conn, scanner.Text(), s, cancel, clientConn)
+	}
+	log.Printf("TCP Client %s disconnected", conn.RemoteAddr().String())
+}
+
+func tcpHandleCommand(conn net.Conn, txt string, s *sdk.SDK, cancel context.CancelFunc, clientConn *connection) {
+	log.Printf("TCP txt: %v", txt)
+
+	if txt == "EXIT" {
```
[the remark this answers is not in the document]
igooch: It's mostly just a quicker way of deleting the game server than kubectl during dev. I can take it out.

examples/simple-genai-server/main.go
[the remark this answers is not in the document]
roberthbailey: In the simple game server, the cancel function is used when you want to force stop health checking but that code has been removed here. I don't think we have any need to cancel manually, so I think there is no need for the cancel function in this program (or this comment).

Also, right now you call `signals.NewSigKillContext` twice. Once on line 38 at the top of main and again on line 90 for the health ping. I think if you remove the second invocation and pass `sigCtx` into `doHealth` then when the process in interrupted or terminated, the health checking will stop.

examples/simple-genai-server/main.go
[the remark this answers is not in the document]
roberthbailey: Yes. Just using TCP sounds fine with me.

examples/simple-genai-server/main.go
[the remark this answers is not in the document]
roberthbailey: I'm ok leaving this for now but I think we will end up iterating on the request structure as we test with different backends.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	if gac := os.Getenv("GenAiContext"); gac != "" {
+		genAiContext = &gac
+	}
+	if p := os.Getenv("Prompt"); p != "" {
+		prompt = &p
+	}
+	if se := os.Getenv("SimEndpoint"); se != "" {
+		simEndpoint = &se
+		log.Printf("Creating Client at endpoint %s", *simEndpoint)
```
roberthbailey, line 66: I think creating the client should be done outside of checking env variables. Right now if I pass a flag (rather than an environment variable) then the client won't ever be initialized.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+		go tcpListener(port, s, compConn)
+	} else {
+		go handleChat(*prompt, compConn, simConn, *numChats)
+	}
+
+	<-sigCtx.Done()
+	os.Exit(0)
+}
+
+func initClient(endpoint string, context string) (*connection) {
```
roberthbailey, line 109: nit: For a single return variable you don't need to use parenthesis.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+
+type connection struct {
+	client   *http.Client
+	endpoint string // full base URL for API requests
+	context  string
+	// TODO: create options for routes off the base URL
+}
+
+type GenAIRequest struct {
+	Context         string  `json:"context,omitempty"`
```
roberthbailey, line 124: nit: gofmt on this? (it should remove spaces to move the types closer to the variable names)

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+		Prompt: prompt,
+	}
+	jsonStr, err := json.Marshal(jsonRequest)
+	if err != nil {
+		return "", fmt.Errorf("unable to marshal json request: %v", err)
+	}
+
+	req, err := http.NewRequest("POST", clientConn.endpoint, bytes.NewBuffer(jsonStr))
+	if err != nil {
+		return "unable create http POST request", err
```
roberthbailey, line 140: nit: Use the `return "", fmt.Errorf(...)` pattern here to be consistent with the rest of the function.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+	log.Print("Marking this server as ready")
+	if err := s.Ready(); err != nil {
+		log.Fatalf("Could not send ready message")
+	}
+
+	// Start up TCP listener so the user can interact with the GenAI endpoint manually
+	if simConn == nil {
+		go tcpListener(port, s, compConn)
+	} else {
+		go handleChat(*prompt, compConn, simConn, *numChats)
```
roberthbailey, line 102: nit re: Naming - maybe something like "autonomousChat" here?

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+	go doHealth(s, ctx)
+
+	log.Print("Marking this server as ready")
+	if err := s.Ready(); err != nil {
+		log.Fatalf("Could not send ready message")
+	}
+
+	// Start up TCP listener so the user can interact with the GenAI endpoint manually
+	if simConn == nil {
+		go tcpListener(port, s, compConn)
```
roberthbailey, line 100: This is selecting between "interactive" mode and "autonomous" mode. Can the function names / comment make that clearer?

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	response, err := handleGenAIRequest(prompt, conn1)
+	if err != nil {
+		log.Fatalf("could not send request: %v", err)
+	} else {
+		log.Printf("%d PROMPT: %s\nRESPONSE: %s\n", numChats, prompt, response)
+	}
+
+	numChats -= 1
+	handleChat(response, conn2, conn1, numChats)
```
roberthbailey, line 176: I don't have anything against recursion, but a loop might be clearer, e.g.

`for (i = numChats; i--; i >0) { handleGenAIRequest() }`

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func tcpListener(port *string, s *sdk.SDK, clientConn *connection) {
```
roberthbailey, line 179: clientConn here is the genAI client?

roberthbailey, line 179: I don't think we need to pass the `sdk.SDK` down through the interactive part. It's done in the simple game server so that you can interactively control the game server's calls into the SDK. But here the SDK should just be in the background and not part of the main program flow.

roberthbailey, line 179: This was probably copied from the simple game server, but why is port a pointer to a string instead of just a string?

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+		conn, err := ln.Accept()
+		if err != nil {
+			log.Printf("Unable to accept incoming TCP connection: %v", err)
+		}
+		go tcpHandleConnection(conn, s, clientConn)
+	}
+}
+
+// handleConnection services a single tcp connection to the server
+func tcpHandleConnection(conn net.Conn, s *sdk.SDK, clientConn *connection) {
```
roberthbailey, line 197: I'm not sure this structure of 4 nested functions for handling the interactive TCP endpoint makes the most sense in this program. In the simple game server, it's done so that some code can be shared between TCP / UDP and also because the message processing is really long. Can you streamline / inline some of this or think of another way to simplify it?

examples/simple-genai-server/main.go
igooch: I mostly did it for flipping between sending the reponses to the different endpoints. I can switch it to iterative though, that's easy enough.

examples/simple-genai-server/main.go
igooch: Yep, it was just copied over. I can reduce it.

examples/simple-genai-server/main.go
roberthbailey: Ah, I missed that the arguments kept flipping. This may be cleaner than doing it iteratively in that case (maybe add a comment so it's clear?).

examples/simple-genai-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+	Prompt          string  `json:"prompt"`
+}
+```
+
+If you need a different request structure, you will need to clone or fork this repository and
+modify this in `main.go`. Update `REPOSITORY` in the `agones/examples/simple-genai-server/Makefile`
+to your own container registry and run `make build && make push` from within the
+`agones/examples/simple-genai-server` path. Then modify the `gameserver.yaml` to pull the image from
+your container registry. If you are making a series of changes you may also want to add
+`imagePullPolicy: Always` to the container image in `gameserver.yaml`.
```
roberthbailey, line 25: It might be nice to add this into the yaml file (commented out like we do [here](https://github.com/googleforgames/agones/blob/main/examples/cpp-simple/fleet.yaml#L34)).

examples/simple-genai-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```
+type GenAIRequest struct {
+	Context         string  `json:"context,omitempty"`
+	Prompt          string  `json:"prompt"`
+}
+```
+
+If you need a different request structure, you will need to clone or fork this repository and
+modify this in `main.go`. Update `REPOSITORY` in the `agones/examples/simple-genai-server/Makefile`
```
roberthbailey, line 21: Can you turn `main.go` and `agones/examples/simple-genai-server/Makefile` into links to the files?

examples/simple-genai-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+to your own container registry and run `make build && make push` from within the
+`agones/examples/simple-genai-server` path. Then modify the `gameserver.yaml` to pull the image from
+your container registry. If you are making a series of changes you may also want to add
+`imagePullPolicy: Always` to the container image in `gameserver.yaml`.
+
+## Setting up the Game Server
+
+This example uses two separate clusters for the GenAI server and the game server. For the game
```
roberthbailey, line 29: I'm wondering why we don't suggest running 1 autopilot cluster with everything in it. After setting up the GenAI cluster, you just install agones into it and launch the game server. WDYT?

examples/simple-genai-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+server follow the instructions for [creating a cluster](https://agones.dev/site/docs/installation/creating-cluster/)
+and [installing Agones](https://agones.dev/site/docs/installation/install-agones/).
+
+Modify the `gameserver.yaml` `GenAiEndpoint` value to your inference server's endpoint. If you want
```
roberthbailey, line 33: Would it be better to have 2 yaml files? One for interactive mode and one for autonomous mode? That would simplify the instructions here a bit (pick the right file) and let this section focus on describing the configuration options.

examples/simple-genai-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+responses for the chat. The `GenAiContext` is sent to the `GenAiEndpoint` with each request, and the
+`SimContext` is sent to the `SimEndpoint` with each request as part of the GenAIRequest structure.
+The default values for `GenAiContext` and `SimContext` are empty strings. The `Prompt` is the first
+sent to prompt send to the GenAI endpoint to start the chat. The default values for the prompt is an
+empty string. `NumChats` is the number of requests made to the `SimEndpoint` and `GenAiEndpoint`.
+The default value for is `NumChats` is `1`.
+
+## Running the Game Server
+
+Once you have modified the game server, apply to your Agones cluster with `kubectl -f gameserver.yaml`.
```
roberthbailey, line 49: Most people running this probably won't modify the game server (edit go code, build and push to their own repo). Do you mean configure the game server for your own GenAI endpoint?

examples/simple-genai-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+If you did not set up `SimEndpoint` you can manually send requests to the GenAI endpoint. Retreive
+the IP address and port:
+
+```bash
+kubectl get gs gen-ai-server -o jsonpath='{.status.address}:{.status.ports[0].port}'
+```
+
+You can now send requests to the GenAI endpoint:
+
+> [!NOTE]
```
roberthbailey, line 68: I haven't seen this before, but it renders nicely in github!

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,243 @@
...
+
+	port := flag.String("port", "7654", "The port to listen to traffic on")
+	genAiEndpoint := flag.String("GenAiEndpoint", "", "The full base URL to send API requests to to simulate computer (NPC) responses to user input")
+	genAiContext := flag.String("GenAiContext", "", "Context for the GenAI endpoint")
+	prompt := flag.String("Prompt", "", "The first prompt for the GenAI endpoint")
+	simEndpoint := flag.String("SimEndpoint", "", "The full base URL to send API requests to to simulate user input")
+	simContext := flag.String("SimContext", "", "Context for the Sim endpoint")
+	numChats := flag.Int("NumChats", 1, "Number of back and forth chats between the sim and genAI")
+
+	var simConn *connection
```
roberthbailey, line 48: nit: Move the variable declaration below the flag parsing closer to where the variables are initially used.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,243 @@
...
+	s, err := sdk.NewSDK()
+	if err != nil {
+		log.Fatalf("Could not connect to sdk: %v", err)
+	}
+
+	log.Print("Starting Health Ping")
+	go doHealth(s, sigCtx)
+
+	log.Print("Marking this server as ready")
+	if err := s.Ready(); err != nil {
```
roberthbailey, line 88: I'd suggest marking the server as ready after initializing the connections. Otherwise you could be ready and then immediately crash (e.g. if no genaiendpoint is specified).

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,243 @@
...
+	ln, err := net.Listen("tcp", ":"+port)
+	if err != nil {
+		log.Fatalf("Could not start TCP server: %v", err)
+	}
+	defer ln.Close() // nolint: errcheck
+
+	for {
+		conn, err := ln.Accept()
+		if err != nil {
+			log.Printf("Unable to accept incoming TCP connection: %v", err)
```
roberthbailey, line 199: I think you need to add a `continue` after logging the error (or use if/else). Otherwise you still pass the connection to the `tcpHandleConnection` function.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,243 @@
...
+
+// Two AIs (connection endpoints) talking to each other
+func autonomousChat(prompt string, conn1 *connection, conn2 *connection, numChats int) {
+	if numChats <= 0 {
+		return
+	}
+	response, err := handleGenAIRequest(prompt, conn1)
+	if err != nil {
+		log.Fatalf("could not send request: %v", err)
+	} else {
```
roberthbailey, line 178: Since the if clause causes the program to exit, this doesn't need to be in an else clause and can be un-indented, e.g.

```
if err != nil {
	log.Fatalf("could not send request: %v", err)
}
log.Printf("...")
```

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,243 @@
...
+		log.Printf("%d %s RESPONSE: %s\n", numChats, conn1.name, response)
+	}
+
+	numChats -= 1
+	// Flip between the connection that the response is sent to
```
roberthbailey, line 183: nit: end comment with a period

https://google.github.io/styleguide/go/decisions#comment-sentences

examples/simple-genai-server/README.md
igooch: I mean modify the gameserver.yaml to use their endpoint. I'll update the wording.

examples/simple-genai-server/README.md
igooch: Good idea, I'll make separate example files.

examples/simple-genai-server/README.md
igooch: It's because the GenAI Quickstart is a zonal (Standard) cluster, so they'd need to modify the GenAI Quickstart cluster terraform. I can make instructions for setting up Agones within an existing cluster. Separately I can do a PR to get GenAI Quickstart onto an Autopilot cluster.

examples/simple-genai-server/README.md
igooch: Oh yep, looks like GitHub can do relative links.

examples/simple-genai-server/main.go
igooch: Ah, I meant to make this fatal.

examples/simple-genai-server/main.go
```diff
@@ -0,0 +1,240 @@
...
+	if err := s.Ready(); err != nil {
+		log.Fatalf("Could not send ready message")
+	}
+
+	// Start up TCP listener so the user can interact with the GenAI endpoint manually
+	if simConn == nil {
+		go tcpListener(*port, genAiConn)
+	} else {
+		log.Printf("Starting autonomous chat with Prompt: %s", *prompt)
+		go autonomousChat(*prompt, genAiConn, simConn, *numChats)
```
roberthbailey, line 106: Do you think that is autonomous mode the gameserver should exit once the conversation is done (maybe with an optional sleep time in case you want to grab the logs)? When running it I find it a bit strange that the gameserver just keeps running in the ready state once it's done working but continues to send health pings.

examples/simple-genai-server/main.go
```diff
@@ -103,7 +104,16 @@ func main() {
 		go tcpListener(*port, genAiConn)
 	} else {
 		log.Printf("Starting autonomous chat with Prompt: %s", *prompt)
-		go autonomousChat(*prompt, genAiConn, simConn, *numChats)
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go autonomousChat(*prompt, genAiConn, simConn, *numChats, &wg)
+		wg.Wait()
+
+		log.Printf("Shutting down the Game Server.")
```
roberthbailey, line 112: I think it would be simpler if you just put these lines into the `autonomousChat` function when you hit the base case:

```
	if numChats <= 0 {
		log.Printf("Shutting down the Game Server.")
		shutdownErr := s.Shutdown()
		if shutdownErr != nil {
			log.Printf("Could not shutdown")
		}
		return
	}
```

I also think that having the wait group here means that the `<-sigCtx.Done()` line below never gets executed until after the autonomous chats complete. So an interrupt wouldn't be caught.

roberthbailey:
Just the one last comment on the new code that you added. Then this should be ready to merge!

examples/simple-genai-server/main.go
igooch: I changed up the code to pass through the sigCtx to the autonomousChat.

We get an error `"error":"http: Server closed"` after the request to shutdown, but it looks like that's coming from the SDK Server httpEndpoint.

```
{"gsKey":"default/gen-ai-server-auto","message":"SDK server shutdown requested, waiting for game server shutdown","severity":"info","source":"*sdkserver.SDKServer","state":"RequestReady","time":"2024-02-23T23:25:18.158333903Z"}
{"message":"Shutting down SDK server","severity":"info","source":"main","time":"2024-02-23T23:25:18.196158733Z"}
{"error":"http: Server closed","message":"http server closed","severity":"info","source":"main","time":"2024-02-23T23:25:18.196415148Z"}
{"gsKey":"default/gen-ai-server-auto","message":"...shutting down workers","queue":"agones.dev.default.gen-ai-server-auto","severity":"info","source":"*sdkserver.SDKServer","time":"2024-02-23T23:25:18.196206956Z"}
{"error":"http: Server closed","gsKey":"default/gen-ai-server-auto","message":"Health check: http server closed","severity":"error","source":"*sdkserver.SDKServer","time":"2024-02-23T23:25:18.196492334Z"}
```

roberthbailey (approved):

PR #3631, opened 2024-02-02
Define SDKServer LogLevel early

### LINKED ISSUE #3629: Some info logs logged from sidecar when sdkServer.logLevel set to Error
**What happened**: 
With `sdkServer.logLevel` set to `Error` sidecar logs are still visible such as
```
Starting sdk sidecar
```
And
```
Starting SDKServer grpc service...
```

To see only error logs from the sidecar

1. Set `sdkServer.logLevel` to `Error` via manifest entry
2. Start a gs
3. View logs and note that some info level logs are present

One log originates here https://github.com/googleforgames/agones/blob/main/cmd/sdk-server/main.go#L231
Another log originates here https://github.com/googleforgames/agones/blob/main/cmd/sdk-server/main.go#L75

These are triggered before logging is configured.

@markmandel posted that logging is configured here https://github.com/googleforgames/agones/blob/5f58fce4c00f97ef012db8f8a503898a4725d11b/pkg/sdkserver/sdkserver.go#L249C1-L261C3

Suggestion from Mark
```
Off the top of my head, I can't see any reason we couldn't make the log level an environment / arg on the sidecar pod instead, and have it be global
```

**Environment**:
- Agones version: 1.32
- Kubernetes version (use `kubectl version`): 1.26
- Others:

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #3629


Kalaiselvi84:
I've defined the sdkServer logLevel at the beginning of the sdk-server/main.go file. Please let me know if this is the correct approach and what are the other necessary changes required?

markmandel:
> I've defined the sdkServer logLevel at the beginning of the sdk-server/main.go file. Please let me know if this is the correct approach and what are the other necessary changes required?

To add the environment variable to the sidecar, you will need to edit this function to add an environment variable to the set that already exists:
https://github.com/googleforgames/agones/blob/5f58fce4c00f97ef012db8f8a503898a4725d11b/pkg/gameservers/controller.go#L641

Kalaiselvi84:
> To add the environment variable to the sidecar, you will need to edit this function to add an environment variable to the set that already exists:

I've added an environment variable to the `sidecar` method in the `pkg/gameservers/controller.go` file and updated the `Run` method to reflect this capture in the `pkg/sdkserver/sdkserver.go` file. Does this change seem correct to you? Please let me know if anything else needs to be modified.

pkg/sdkserver/sdkserver.go
```diff
@@ -246,17 +247,17 @@ func (s *SDKServer) Run(ctx context.Context) error {
 		return err
 	}
 
-	logLevel := agonesv1.SdkServerLogLevelInfo
-	// grab configuration details
-	if gs.Spec.SdkServer.LogLevel != "" {
-		logLevel = gs.Spec.SdkServer.LogLevel
+	var logLevel string
```
markmandel, line 250: So the one that really needs changing is here:

https://github.com/googleforgames/agones/blob/8b2c44256954cd449e6a39ad68f1da767c80e927/cmd/sdk-server/main.go#L72-L75

Which is some of the Info level output that was happening when logging was set to Warning/Error only.

What I'm thinking we should do here, is take a `logrus.Level` as part of `NewSDKServer` (which will mean refactoring some unit tests, and some error handling code) and passing it in after parsing in main.go.

Then we can set the `s.logger.Logger.SetLevel(level)` right under:
https://github.com/googleforgames/agones/pull/3631/files#diff-2b47120b02bc51219ccd6d97b8fe68f9d5043997eba480c0655babcafd771e4eR185 (line 185), so that is also covers all logging in this file.

Hopefully that all was clear!

Kalaiselvi84: Following your guidance, I made below modifications:
1. Removed the hardcoded log level in `controller.go` and replaced it with `gs.Spec.SdkServer.LogLevel`.
2. Added log level as a parameter in the `NewSDKServer` function to accept the log level.
3. Added log level parsing in `main.go` and passed it to `NewSDKServer` function.
4. Set `s.logger.Logger.SetLevel(logLevel)` right after its initialization  in the `NewSDKServer` function.
5. Modified the `controller_test.go` and `sdkserver_test.go` files based on the lint suggestion.

Please let me know if anything missing

cmd/sdk-server/main.go
```diff
@@ -70,6 +71,14 @@ var (
 )
 
 func main() {
+	logLevelEnv := os.Getenv("LOG_LEVEL")
```
markmandel, line 74: Rather than use `os.Getenv` - please make this part of the `parseEnvFlags()` function and object that it returns - then we have all argument and environment variable parsing in a single place.

pkg/sdkserver/sdkserver.go
```diff
@@ -246,17 +248,17 @@ func (s *SDKServer) Run(ctx context.Context) error {
...
+	logLevelEnv := os.Getenv("LOG_LEVEL")
```
markmandel, line 252: This section can be deleted since we set the log level in the constructor.

markmandel (commented):
Couple of small things, but overall looks good.

Will need a manual test!

To test, take a single `GameSever` such as: https://github.com/googleforgames/agones/blob/main/examples/simple-game-server/gameserver.yaml and set different log levels (Info, warn, error, debug). See how it changes the log output for the sidecar.

cmd/sdk-server/main.go
Kalaiselvi84: Please ignore the above question. I've tested the latest code changes with the `Info`, `Error`, `Debug`, and `Warn` log levels. Please review the logs at https://gist.github.com/Kalaiselvi84/74fe77e6cce7ebc55485efbeb1e0d79d and share your feedback.

cmd/sdk-server/main.go
```diff
@@ -25,8 +25,10 @@ import (
 	"strings"
 	"time"
 
+	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	gwruntime "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
igooch, line 31: @markmandel it looks like this is the same package that we use in our other controllers. Is there a reason we use this one over https://pkg.go.dev/log/slog?

markmandel: Slog didn't exist when we started this project 😄

pkg/sdkserver/sdkserver.go
```diff
@@ -246,19 +247,8 @@ func (s *SDKServer) Run(ctx context.Context) error {
...
-	s.logger.WithField("logLevel", logLevel).Debug("Setting LogLevel configuration")
-	level, err := logrus.ParseLevel(strings.ToLower(string(logLevel)))
-	if err == nil {
-		s.logger.Logger.SetLevel(level)
-	} else {
-		s.logger.WithError(err).Warn("Specified wrong Logging.SdkServer. Setting default loglevel - Info")
-		s.logger.Logger.SetLevel(logrus.InfoLevel)
-	}
+	currentLogLevel := s.logger.Logger.GetLevel()
+	s.logger.WithField("logLevel", currentLogLevel).Debug("Setting LogLevel configuration")
```
markmandel, line 251:
```suggestion
```

Don't think we need these lines anymore.

markmandel (commented):
Looks great - a couple of lines to remove, and I think this is good to go 👍🏻

pkg/sdkserver/sdkserver.go
Kalaiselvi84: I've taken out these two lines.

cmd/sdk-server/main.go
Kalaiselvi84: Could you please confirm if the manual testing looks fine? Is this how we should be testing it?

markmandel: Yep! Looks perfect! 👍🏻

markmandel (approved):

markmandel:
Just a note - this is approved, you can take it out of draft 😄

PR #3672, opened 2024-02-26
by default disable agones-metrics nodepools

### LINKED ISSUE #3669: [ Terraform | GKE ] Add option to not create `agones-metrics` nodepools with cluster
In our Terraform for GKE we have no way to turn off the `agones-metrics` nodepool, which seems like an oversight, as unless you are running Prometheus or Graphana, you don't need the nodes.

https://github.com/googleforgames/agones/blob/4b0f7255300675a574bf2f9a8a764bc67586942e/install/terraform/modules/gke/cluster.tf#L211-L247

We should add a variable to the Terraform config to disable the creation of these nodes.

As part of this work, looking at our standard e2e clusters:

image

I don't think we're installing prometheus or graphana on these clusters, so these nodes aren't doing anything - we should use this feature to disable these nodes for the e2e clusters, since they aren't required.

ashutosji:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

**What this PR does / Why we need it**: Since we're not installing Prometheus or Graphana on these clusters. By default, `agones-metrics` nodepools would be disabled as nodes aren't doing anything. However, we can enable it when needed by passing `-var enableAgonesMetricsNodepool=true` on the `terraform apply `command.

**Which issue(s) this PR fixes**: https://github.com/googleforgames/agones/issues/3669

Closes #3669 


markmandel:
Thanks for sending this through!

To compete this enhancement, can we also add documentation here for the new parameters please?
https://agones.dev/site/docs/installation/terraform/gke/#installation

You can see details on writing docs here: https://agones.dev/site/docs/contribute/ - the new parameters will need to go behind a `feature` shortcode.

Also, we'll want a variable here: https://github.com/googleforgames/agones/blob/main/build/includes/google-cloud.mk with it enabled to be on, so that the default test cluster we use for development still has the metric nodes.

ashutosji:
> Thanks for sending this through!
> 
> To compete this enhancement, can we also add documentation here for the new parameters please? https://agones.dev/site/docs/installation/terraform/gke/#installation
> 
> You can see details on writing docs here: https://agones.dev/site/docs/contribute/ - the new parameters will need to go behind a `feature` shortcode.
> 
> Also, we'll want a variable here: https://github.com/googleforgames/agones/blob/main/build/includes/google-cloud.mk with it enabled to be on, so that the default test cluster we use for development still has the metric nodes.

@markmandel,
Thanks for the review.
I have updated the changes as suggested by you.
Just a quick question, Shall I add next upcoming version in the shortcode or current version(1.38.0) is ok?

examples/terraform-submodules/gke/module.tf
```diff
@@ -16,6 +16,9 @@
 // Run:
 //  terraform apply -var project="<YOUR_GCP_ProjectID>" [-var agones_version="1.17.0"]
 
+// To enable agones-metrics add below command in `terraform apply`:
+// -var enable_agones_metrics_nodepool=true
+
```
markmandel, line 21:
```suggestion
```

I think this can be rolled back - we don't have any other variable specific docs in a comment here.

build/includes/google-cloud.mk
```diff
@@ -32,6 +32,7 @@ gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_MIN_NODECOUNT ?= 1
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_MAX_NODECOUNT ?= 5
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_WINDOWSINITIALNODECOUNT ?= 0
 gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_WINDOWSMACHINETYPE ?= e2-standard-4
+gcloud-test-cluster: GCP_CLUSTER_NODEPOOL_ENABLE_AGONES_METRICS_NODEPOOL ?= true
```
markmandel, line 35: 👍🏻

markmandel (commented):
Small revert, but otherwise, LGTM 👍🏻

examples/terraform-submodules/gke/module.tf
ashutosji: > I think this can be rolled back - we don't have any other variable specific docs in a comment here.

Thanks for the review and suggestions.
Done!

markmandel (approved):

PR #3679, opened 2024-03-04
Adds Chat Message History and Connects to the NPC Chat API

igooch:
**What type of PR is this?**
/kind documentation

Adds chat message history to the GenAI Request. This will make for better responses from VertexAI.

Adds the request type NPC which allows for making requests to the [Google for Games GenAI NPC API](https://github.com/googleforgames/GenAI-quickstart/genai/api/npc_chat_api).

N/A

This works with Vertex, so in https://github.com/googleforgames/GenAI-quickstart/blob/205ba78aa65782f00a8654e31c88efdc4fdaded6/genai/api/npc_chat_api/config/config.toml#L20-L21 comment out "GKEGenAI", and uncomment "VertexAI".

You will need to first start up and run the GenAI Quickstart cluster, then start and run the NPC chat server.

### DIFF
--- a/examples/simple-genai-server/gameserver_npcchat.yaml
+++ b/examples/simple-genai-server/gameserver_npcchat.yaml
@@ -29,26 +29,26 @@ spec:
           image: us-docker.pkg.dev/agones-images/examples/simple-genai-game-server:0.1
           # imagePullPolicy: Always  # add for development
           env:
-            - name: GenAiEndpoint
+            - name: GEN_AI_ENDPOINT
             # Use the service endpoint address when running in the same cluster as the inference server.
               # TODO (igooch): Change this to the `/genai/npc-chat` endpoint when it's properly plumbed in the inference server
               value: "http://npc-chat-api.genai.svc.cluster.local:80"
             # GenAiContext is not passed to the npc-chat-api endpoint.
-            - name: GenAiNpc # False by default. Use GenAiNpc "true" when using the npc-chat-api as the GenAiEndpoint.
+            - name: GEN_AI_NPC # False by default. Use GEN_AI_NPC "true" when using the npc-chat-api as the GEN_AI_ENDPOINT.
               value: "true"
-            - name: FromID # Default is "2".
+            - name: FROM_ID # Default is "2".
               value: "2"
-            - name: ToID # Default is "1".
+            - name: TO_ID # Default is "1".
               value: "1"
-            - name: SimEndpoint
+            - name: SIM_ENDPOINT
               value: "http://192.1.1.2/genai/chat"
-            - name: SimContext
+            - name: SIM_CONTEXT
               value: "Ask questions about one of the following: What happened here? Where were you during the earthquake? Do you have supplies?"
-            - name: SimNpc
-              value: "false" # False by default. Use SimNpc "true" when using the npc-chat-api as the SimEndpoint.
-            - name: Prompt
+            - name: SIM_NPC
+              value: "false" # False by default. Use SIM_NPC "true" when using the npc-chat-api as the SIM_ENDPOINT.
+            - name: PROMPT
               value: "Hello"
-            - name: NumChats
+            - name: NUM_CHATS
               value: "50"
           resources:
             requests:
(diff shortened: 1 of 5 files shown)


examples/simple-genai-server/gameserver_npcchat.yaml
```diff
@@ -0,0 +1,61 @@
...
+  template:
+    spec:
+      containers:
+        - name: simple-genai-game-server
+          image: us-docker.pkg.dev/agones-images/examples/simple-genai-game-server:0.1
+          # imagePullPolicy: Always  # add for development
+          env:
+            - name: GenAiEndpoint
+            # Use the service endpoint address when running in the same cluster as the inference server.
+              value: "http://npc-chat-api.genai.svc.cluster.local:80"
```
zmerlynn, line 34: This should be plumbed to the npc chat endpoint of `/genai` eventually. The idea is for that to be the front-end to all of it. You can add here:

```
TODO: Change this to the `/genai/npc-chat` endpoint when it's properly plumbed. 
```

examples/simple-genai-server/gameserver_npcchat.yaml
```diff
@@ -0,0 +1,61 @@
...
+            # GenAiContext is not passed to the npc-chat-api endpoint.
+            - name: GenAiNpc # False by default. Use GenAiNpc "true" when using the npc-chat-api as the GenAiEndpoint.
+              value: "true"
+            - name: SimEndpoint
+              value: "http://vertex-chat-api.genai.svc.cluster.local:80"
```
zmerlynn, line 43: Here use the `/genai/chat` endpoint instead. If you want to handle this as a follow-on and leave a TODO that's fine, though.

examples/simple-genai-server/gameserver_npcchat.yaml
```diff
@@ -0,0 +1,61 @@
...
+              value: "50"
+          resources:
+            requests:
+              memory: 64Mi
+              cpu: 20m
+            limits:
+              memory: 64Mi
+              cpu: 20m
+      # Schedule onto the game server node pool when running in the same cluster as the inference server.
+      tolerations:
```
zmerlynn, line 60: In the other examples you commented the node selector hunk out, probably do it here as well.

examples/simple-genai-server/main.go
```diff
@@ -88,6 +90,20 @@ func main() {
 		}
 		simNpc = &snpc
 	}
+	if fid := os.Getenv("FromID"); fid != "" {
```
zmerlynn, line 93: nit: It looks like this pattern is prevalent here already but it's really strange to see environment variables that are not in ALL_UPPER_CASE_WITH_UNDERSCORES. I think we can fix this in a follow-on if you want to leave a TODO, though.

zmerlynn (changes requested):

examples/simple-genai-server/gameserver_npcchat.yaml
igooch: Done. We'll probably also take this out once the inference server is moved to Autopilot too.

examples/simple-genai-server/main.go
igooch: QQ: should the actual env variables also be ALL_UPPER_CASE_? I.e. should it be `PORT := flag.String("port"...` or `PORT := flag.String("PORT"...`

zmerlynn: * The flag name *should* be lower-case-with-dashes: https://github.com/googleforgames/agones/blob/main/cmd/controller/main.go#L58-L83
  * _That said_, you're following a nearer convention: https://github.com/googleforgames/agones/blob/main/examples/simple-game-server/main.go#L41-L50 .. so honestly just keep the flag names. Let's fight that battle another day.
* The Go variable name should follow Go style, which is mixedCase: https://go.dev/doc/effective_go#mixed-caps
* The environment variable should be in UPPER_CASE_WITH_UNDERSCORES: https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

examples/simple-genai-server/gameserver_npcchat.yaml
```diff
@@ -0,0 +1,66 @@
...
+      containerPort: 7654
+      protocol: TCP
+  template:
+    spec:
+      containers:
+        - name: simple-genai-game-server
+          image: us-docker.pkg.dev/agones-images/examples/simple-genai-game-server:0.1
+          # imagePullPolicy: Always  # add for development
+          env:
+            - name: GenAiEndpoint
```
zmerlynn, line 32: these are environment variables, so you'll need to change them to match the name upper case names.

Note that you could have just used args here instead, up to you.

igooch: Thanks - missed that. Updated.

zmerlynn (approved):

PR #3682, opened 2024-03-06
fix: cannot load extensions image on minikube node

katsew:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

When testing on minikube using locally built images, use `minikube-push` to load the built images on minikube.  
At this time only the extensions image was not loaded.  
This PR has been modified to load the extensions images that are not loaded.

NA

NA


markmandel (approved):
Nice pickup - thanks! Been a while since I've used minikube for dev!

Issue #3486, opened 2023-11-02
Agones Controller has OOMKilled in version 1.35.0


alvin-7:
What happened:
Agones controller memory usage leak to 480M and Exceeding the Kubernetes memory limit caused being OOMKilled

What you expected to happen:
Agones controller memory usage stays at normal level, not spike unreasonably

How to reproduce it (as minimally and precisely as possible):
Unable to provide specific steps to reproduce currently. This spike happened during normal usage.

Anything else we need to know?:
Agones controller has OOMKilled

Environment:

Agones version: 1.35.0
Kubernetes version (use kubectl version): 1.22.5
Cloud provider or hardware configuration: Tencent Cloud
Install method (yaml/helm): helm

![image](https://github.com/googleforgames/agones/assets/38536867/2f876546-2a59-425c-9228-46b74520f779)
After upgrading Agones to version 1.35.0, I still encounter memory leakage issues. In my scenario, the fleet is constantly being deleted and added, with the total number remaining the same but the fleet names changing.

alvin-7:
The previous issue is here:
https://github.com/googleforgames/agones/issues/3380

markmandel:
If I look at the installed controller that I've installed via Helm (this is the dev version), we don't set any cpu or memory limits by default.

```
  agones-controller:
    Container ID:   containerd://d9507676a30507a6bbfb5f0e1ddbe24e089b59874e0b78193eb3b98dff2e9426
    Image:          us-docker.pkg.dev/agones-mark-dev/images/agones-controller:1.36.0-dev-b148d3b
    Image ID:       us-docker.pkg.dev/agones-mark-dev/images/agones-controller@sha256:16182889ecf89777799f2801ea73b183915044281a6765aeadc36482d8a8bc2b
    Ports:          8081/TCP, 8080/TCP
    Host Ports:     0/TCP, 0/TCP
    State:          Running
      Started:      Thu, 19 Oct 2023 13:19:23 -0700
    Ready:          True
    Restart Count:  0
    Limits:
      ephemeral-storage:  10100Mi
    Requests:
      ephemeral-storage:  10100Mi
```

It looks like you've set the memory limit to 500Mi ? This may just be too low a memory limit for the Agones controller - it can be quite memory hungry at times.

What happens if you disable the memory limit?

markmandel:
Closing!

PR #3692, opened 2024-03-10
fix: #3607 Metrics data loss in K8S controller

### LINKED ISSUE #3607: Metrics data loss in K8S controller
**What happened**:
After restarting the K8S controller, the "agones_gameservers_total" metric is no longer being collected for the "shipping-mode1-map1-3568" battle server. However, the "agones_gameservers_count" metric is still being collected.

I expected both the "agones_gameservers_total" and "agones_gameservers_count" metrics to continue being collected consistently, even after the controller restart.
- max by(type) (agones_gameservers_total{fleet_name="shipping-mode1-map1-3568"})
image
- avg by(type) (agones_gameservers_count{fleet_name="shipping-mode1-map1-3568"})
image

1. Start the K8S controller - Agones Controller.
2. Create "shipping-mode1-map1-3568" fleet
3. Check the metrics for the "shipping-mode1-map1-3568" battle server.
4. Observe that the "agones_gameservers_total" metric is being collected, but "agones_gameservers_count" metric is not.
5. Restart the controller.
6. Check the metrics for the "shipping-mode1-map1-3568" battle server after the restart.
7. Notice that the "agones_gameservers_total" metric is no longer being collected, while the "agones_gameservers_count" metric is still being collected.

1. In the cluster, there are a total of 10 fleets, and there is a continuous process of deleting existing fleets and creating new fleets. This dynamic fleet activity might have an impact on the metrics data collection.

**Environment**:
- Agones version:  1.35.0
Client Version: v1.27.2
Kustomize Version: v5.0.1
Server Version: v1.22.5-tke.19

- Others:

alvin-7:
This PR addresses two issues.

The first issue occurs when a Fleet is being deleted, which triggers two events: an Update (where the DeletionTimestamp is not nil) and a Delete. However, when the Update event is triggered, because the DeletionTimestamp is not nil, it eventually leads to resyncFleets. Inside resyncFleets, the Update function is called for each Fleet, resulting in a deadlock. The consequence of this is that metrics will not be recorded, as the lock cannot be acquired.

The second issue arises after a Fleet is deleted, and only then does the deletion of GameServers commence. When a GameServer's status is modified, it attempts to write the Fleet's status, as recorded in memory, into Metrics, which includes Fleets that have already been deleted. To address this issue, I made modifications so that only the metrics corresponding to the GameServer's current Fleet are recorded during this process.

Fixes #3607


markmandel:
Thanks for taking a pass at this! Looks like some unit tests are failing. Let us know if you would like to some help debugging them.

As part of the PR description and commit, can you provide a description of what your fix does and why its required?

It would definitely help with review, as otherwise we have to reverse engineer the fix from the bug 😄

alvin-7:
> Thanks for taking a pass at this! Looks like some unit tests are failing. Let us know if you would like to some help debugging them.
> 
> As part of the PR description and commit, can you provide a description of what your fix does and why its required?
> 
> It would definitely help with review, as otherwise we have to reverse engineer the fix from the bug 😄


This PR addresses two issues. 

The first issue occurs when a Fleet is being deleted, which triggers two events: an Update (where the DeletionTimestamp is not nil) and a Delete. However, when the Update event is triggered, because the DeletionTimestamp is not nil, it eventually leads to resyncFleets. Inside resyncFleets, the Update function is called for each Fleet, resulting in a deadlock. The consequence of this is that metrics will not be recorded, as the lock cannot be acquired.

The second issue arises after a Fleet is deleted, and only then does the deletion of GameServers commence. When a GameServer's status is modified, it attempts to write the Fleet's status, as recorded in memory, into Metrics, which includes Fleets that have already been deleted. To address this issue, I made modifications so that only the metrics corresponding to the GameServer's current Fleet are recorded during this process.

alvin-7:
```mermaid
sequenceDiagram
    actor e as event
    participant c as controller
    participant l as locker

    e->>c: on fleet be deleted
    c->>+c: recordFleetChanges
    c->>+c: recordFleetDeletion
    c->>+c: resyncFleets
    c->>+l: lock
    alt In fleet.list, the deleted Fleet no longer exists.
        c ->>- e: end function
    else the deleted Fleet exists, Dead Lock
        c->>+c: recordFleetChanges
        c->>+c: recordFleetDeletion
        c->>+c: resyncFleets
        c->>+l: lock [dead lock]
    end
```

markmandel:
TIL you can create mermaid diagrams in GitHub issues!

markmandel:
Is it also possible to write a unit test, to ensure this issue doesn't show up again?

alvin-7:
> Is it also possible to write a unit test, to ensure this issue doesn't show up again?

I attempted to write or modify test cases to cover this scenario, but the challenge is that it would require special modifications to the existing fakeFleetLister.List in order to reproduce this issue, otherwise it cannot be recreated.

markmandel:
> I attempted to write or modify test cases to cover this scenario, but the challenge is that it would require special modifications to the existing fakeFleetLister.List in order to reproduce this issue, otherwise it cannot be recreated.

What would you need to do to replicate it? I'm sure we could probably work it out.

alvin-7:
> > I attempted to write or modify test cases to cover this scenario, but the challenge is that it would require special modifications to the existing fakeFleetLister.List in order to reproduce this issue, otherwise it cannot be recreated.
> 
> What would you need to do to replicate it? I'm sure we could probably work it out.

1.Add a Standard Fleet: First, create or add a standard Fleet instance within the system.

2. Modify the Deletion Timestamp: Update the metadata.DeletionTimestamp attribute of the Fleet instance to the current time. This simulates the Fleet instance being in a state of deletion.

3. Adjust the fakeFleetLister.List Return Value: Ensure that the fakeFleetLister.List method returns this Fleet instance, which has been marked for deletion (by modifying the deletion timestamp). This step is crucial for the correct execution of the function.

I trust these steps are clear and provide a solid framework for us to replicate the Fleet deletion scenario accurately. Should you have any questions, require further clarifications, or if there's anything more I can assist you with, please do not hesitate to reach out.

markmandel:
> 1. Add a Standard Fleet: First, create or add a standard Fleet instance within the system.
> 2. Modify the Deletion Timestamp: Update the metadata.DeletionTimestamp attribute of the Fleet instance to the current time. This simulates the Fleet instance being in a state of deletion.
> 3. Adjust the fakeFleetLister.List Return Value: Ensure that the fakeFleetLister.List method returns this Fleet instance, which has been marked for deletion (by modifying the deletion timestamp). This step is crucial for the correct execution of the function.

You could absolutely do this with a new Unit Test for sure. The Fake clients for Kubernetes don't have any smarts about them, and basically only do what they are told.

**Option No. 1**
You could do all the manipulation via fake Watch events, which are propagated to List operations, such as what we do here:
https://github.com/googleforgames/agones/blob/130c99abdfc4c67cc05c8e540ca47a8fc86b96ac/pkg/metrics/controller_test.go#L118

**Option No. 2**
Provide the exact list you want to be returned at any given point and time through `AddReactor`:

https://github.com/googleforgames/agones/blob/130c99abdfc4c67cc05c8e540ca47a8fc86b96ac/pkg/gameserversets/controller_test.go#L573-L575

And that can provide you the state you need.

It still will require a new test I expect, but it would be good to test things to ensure that this doesn't regress, as right now there is no guarantee -- you found a tricky bug 😄 , so want to make sure we don't have to call you back again!

alvin-7:
> > 1. Add a Standard Fleet: First, create or add a standard Fleet instance within the system.
> > 2. Modify the Deletion Timestamp: Update the metadata.DeletionTimestamp attribute of the Fleet instance to the current time. This simulates the Fleet instance being in a state of deletion.
> > 3. Adjust the fakeFleetLister.List Return Value: Ensure that the fakeFleetLister.List method returns this Fleet instance, which has been marked for deletion (by modifying the deletion timestamp). This step is crucial for the correct execution of the function.
> 
> You could absolutely do this with a new Unit Test for sure. The Fake clients for Kubernetes don't have any smarts about them, and basically only do what they are told.
> 
> **Option No. 1** You could do all the manipulation via fake Watch events, which are propagated to List operations, such as what we do here:
> 
> https://github.com/googleforgames/agones/blob/130c99abdfc4c67cc05c8e540ca47a8fc86b96ac/pkg/metrics/controller_test.go#L118
> 
> **Option No. 2** Provide the exact list you want to be returned at any given point and time through `AddReactor`:
> 
> https://github.com/googleforgames/agones/blob/130c99abdfc4c67cc05c8e540ca47a8fc86b96ac/pkg/gameserversets/controller_test.go#L573-L575
> 
> And that can provide you the state you need.
> 
> It still will require a new test I expect, but it would be good to test things to ensure that this doesn't regress, as right now there is no guarantee -- you found a tricky bug 😄 , so want to make sure we don't have to call you back again!

Thank you very much for your guidance. I have completed writing the test cases and have submitted them. Could you please review them?

markmandel:
Heading to GDC, but will check when I can 👍🏻

markmandel:
Just letting you know I haven't forgotten about you! I'm still recovering from GDC but this is at the top of my queue.

alvin-7:
> Just letting you know I haven't forgotten about you! I'm still recovering from GDC but this is at the top of my queue.

Thank you so much for the update! I hope you had a great time at GDC and are recovering well. Please take all the time you need, and let me know if there's anything I can do to assist with the review process. Looking forward to your feedback when you're ready.

pkg/metrics/controller_test.go
```diff
@@ -289,6 +289,68 @@ func TestControllerGameServersTotal(t *testing.T) {
 	})
 }
 
+func TestControllerFleet_OnDeleting(t *testing.T) {
```
markmandel, line 292:
```suggestion
func TestControllerFleetOnDeleting(t *testing.T) {
```

I know some people use `_` we don't tend to here, so just consistency.

Also just noticed some snuck in here - oopsie. We should clean that up at a later date.

pkg/metrics/controller_test.go
```diff
@@ -289,6 +289,68 @@ func TestControllerGameServersTotal(t *testing.T) {
...
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
```
markmandel, line 294:
```suggestion
```

Don't need this since we're not interacting with any feature flags 👍🏻

pkg/metrics/controller_test.go
```diff
@@ -339,6 +401,63 @@ func TestControllerFleetReplicasCount_ResetMetricsOnDelete(t *testing.T) {
 	})
 }
 
+func TestControllerFleetAutoScaler_OnDeleting(t *testing.T) {
```
markmandel, line 404:
```suggestion
func TestControllerFleetAutoScalerOnDeleting(t *testing.T) {
```

pkg/metrics/controller_test.go
```diff
@@ -339,6 +401,63 @@ func TestControllerFleetReplicasCount_ResetMetricsOnDelete(t *testing.T) {
...
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
```
markmandel, line 406:
```suggestion

```

markmandel (commented):
Just a couple of small tweaks, and this is good to merge 👍🏻  nice job!

alvin-7:
> Just a couple of small tweaks, and this is good to merge 👍🏻 nice job!

Thank you very much for your feedback and suggestions! I have made the requested adjustments and ensured that all test cases have passed.

markmandel (approved):

markmandel:
Thanks! Approved and ready to go.

alvin-7:
> Thanks! Approved and ready to go.

Received，and  issue #3486 has already been fixed by this PR.

PR #3693, opened 2024-03-10
simple-genai-server 0.2: Make autonomous mode effective

zmerlynn:
This change does a number of things:

* Adds a flag for concurrent player "slots" to imitate more than a single player in a game server.

* Each player slot now runs forever instead of exiting, so ditched the waitgroup.

* Instead of strictly going for a specific number of turns, added a "stop phrase". With prompt engineering, we can instruct the Sim to say a phrase when its done - we then key on this and go one more turn in response and end that player.

* Did some careful prompt engineering to make Sim kind of interesting.

* Cleaned up logging. Added inference "rate" in b/s.

I've already pushed :0.2 if you want to test with this branch.


igooch (approved):
LGTM -- Only comment is to update the README https://github.com/googleforgames/agones/blob/340c39aabd4bd5760e2b8173d6df4e08b6ddbcac/examples/simple-genai-server/README.md?plain=1#L118 with the new behavior.

PR #3716, opened 2024-03-19
Be prescriptive about rotating regions when updating Kubernetes versions

zmerlynn:
* I ensured quota and added alerts in `us-east1`, `us-west1`, `europe-west1`, and `asia-east1` to ensure we have 4 regions outside of `us-central1` that we can use for e2es. (`us-central1` is where we're running perf tests, so I'd rather isolate it.)

* In this PR, I added instructions on how to cycle through the e2e regions as we bump versions, ensuring we stay in these regions and don't duplicate regions in 1.{N-3} .. 1.{N}.


build/terraform/e2e/module.tf
```diff
@@ -41,6 +41,24 @@ variable "kubernetes_versions" {
...
+    // After merge: After the Kubernetes update PR merges, and all active PRs are updated:
+    //
+    // * Move the 1.{N-3} line to the bottom and comment it out
+    // * Change the (commented out) 1.{N-3} to 1.{N+1}
+    // * You should now have 3 versions uncommented (versions 1.{N-2} .. 1.{N}),
+    //   and 1.{N+1} commented out for the next update. The new, commented out 1.{N+1}
+    //   should be using the region of the previous 1.{N-3} - this region will become
+    //   unused.
+    //
+    // Rationale: We cycle the regions us-east1 -> us-west1 -> europe-west1 -> asia-east1 -> us-east1
```
Kalaiselvi84, line 59: Are we not considering the `us-central1` region in this cycle?

Kalaiselvi84:
LGTM!

build/terraform/e2e/module.tf
zmerlynn: We talked offline, and I’d prefer to keep us-central1 for performance alone so we know they’re not interfering.

PR #3722, opened 2024-03-20
Migrating from generate-groups.sh to kube_codegen.sh

### LINKED ISSUE #3659: Migrate away from using the generate-groups script from k8s.io

The `generate-groups` script was deprecated in Kubernetes 1.28 and will no longer exist with Kuberetes 1.30 ([source](https://groups.google.com/a/kubernetes.io/g/dev/c/Ix-ACY9DhEs)). We need to update our tooling to no longer depend upon it. 

The replacement script is `kube_codegen.sh`. We should switch over to the new script.

We have to keep up with Kubernetes so there really isn't an alternative. 

**Additional context**
https://github.com/googleforgames/agones/blob/main/build/build-image/gen-crd-client.sh#L19

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #3659 


Kalaiselvi84:
The `make gen-crd-client command` did not result in any changes during this migration.

zmerlynn:
Provisional LGTM, will approve when I see that `test-gen-crd-client` is actually using it this time. :)

zmerlynn:
```
diff: ../pkg/client: No such file or directory
```

build/build-image/gen-crd-client.sh
```diff
@@ -14,9 +14,63 @@
...
+
+pushd /go/src/k8s.io/code-generator
+
+echo "Fetching the latest tags from remote..."
+git fetch --tags
+
+echo "Checking if the tag v0.30.0-beta.0 exists..."
+if git show-ref --tags | grep -q "refs/tags/v0.30.0-beta.0"; then
+    echo "Tag found. Checking out v0.30.0-beta.0..."
+    git checkout v0.30.0-beta.0
```
zmerlynn, line 36: instead of doing all this, just do this:
* change line 63 of `build/build-image/Dockerfile`:
  * copy and paste the old line so we have it around for when we revert this, comment out the copy
  * Instead of `-b kubernetes-${KUBERNETES_VER}`, just do `-b v0.30.0-beta.0`
  * Leave a note at line 53 that whenever we bump k8s versions next, consider bumping the pinned version. For the 1.29 bump we may just move code-gen to the GA'd v1.30 tag, but once we bump to 1.30, we should stop pinning it.

zmerlynn (changes requested):
Mostly a nit about which layer the tag is at but LGTM otherwise. The diff below is tiny.

build/build-image/gen-crd-client.sh
Kalaiselvi84: I've made the necessary changes to the `build/build-image/Dockerfile` and refined the `gen-crd-client.sh` script. Also, I did not bump the ForceUpdate in the Dockerfile.

zmerlynn (approved):

PR #3726, opened 2024-03-24
Nodejs counters and lists

### LINKED ISSUE #3645: Node.js CountsAndLists SDK implementation and conformance tests
Arbitrary Counters and Lists for GameServers, SDKs and Allocation #2716 has been released in Alpha, and is looking for a contributor with Node.js experience to implement the Node.js SDK. The SDK methods to implement are listed in #2716, and have been implemented in Go https://github.com/googleforgames/agones/blob/main/sdks/go/alpha.go:

- func (a *Alpha) GetCounterCount(key string) (int64, error)
- func (a *Alpha) IncrementCounter(key string, amount int64) (error)
- func (a *Alpha) DecrementCounter(key string, amount int64) (error)
- func (a *Alpha) SetCounterCount(key string, amount int64) (error)
- func (a *Alpha) GetCounterCapacity(key string) (int64, error)
- func (a *Alpha) SetCounterCapacity(key string, amount int64) (error)
- func (a *Alpha) GetListCapacity(key string) (int64, error)
- func (a *Alpha) SetListCapacity(key string, amount int64) (error)
- func (a *Alpha) ListContains(key, value string) (bool, error)
- func (a *Alpha) GetListLength(key string) (int, error)
- func (a *Alpha) GetListValues(key string) ([]string, error)
- func (a *Alpha) AppendListValue(key, value string) (error)
- func (a *Alpha) DeleteListValue(key, value string) (error)

The Node.js SDK should be implemented in the https://github.com/googleforgames/agones/tree/main/sdks/nodejs directory, and should also contain unit tests. Navigating to `agones/build` and running `make sdk-shell-node` may be helpful when running unit tests.

The protocol buffer definitions for these SDK methods are in https://github.com/googleforgames/agones/tree/main/sdks/nodejs/lib/alpha. Note that the proto methods do not exactly match the above listed SDK methods. This is due to the difference between the design in #2716 and the Google AIP https://google.aip.dev/general  guidance that the proto follows as closely as possible.

Conformance tests should be implemented in the https://github.com/googleforgames/agones/tree/main/test/sdk/nodejs directory. Note that the conformance tests use the Local SDK Server. Node.js conformance tests can be run by navigating to `agones/build` and running `make run-sdk-conformance-test-node`. The conformance test https://github.com/googleforgames/agones/blob/main/build/includes/sdk.mk will need to be updated to take in the `ALPHA_FEATURE_GATES`, and the `COUNTS_AND_LISTS_TESTS`.

steven-supersolid:
**What type of PR is this?**
/kind feature

* Counts SDK
* Lists SDK
* Unit tests

Closes #3645 

* The example would require a local game server created with a specifically named counter so unsure whether to continue with that
* Conformance tests are not done yet although there seems to be overlap with the Node.js example, however the example does a lot more


steven-supersolid:
@igooch Would you be able to shed any light on this?

igooch:
> I'm wondering why we don't use int64 instead as we use this in other places in the SDK

The google.protobuf.Int64Value is there so that it will be a nullable object rather than a primitive type that has a default value (0 in the case of Golang). In the case of updating the Count or Capacity 0 is a meaningful value, so we don't want 0 to be the default. 0 for the "diff" isn't meaningful, since that's just adding the diff to the Count.

steven-supersolid:
Thank you. Reading around, it also seems that `google.protobuf.Int64Value` is a recommended way to use a nullable int, so I wonder if there's a better way to import wrappers_pb in JavaScript. It may just be an issue with the library not exporting this useful type the correct way.

For my curiosity I was wondering if we could use a wrapper type e.g. the `Count` that we use in `SetPlayerCapacity`?

```
// Store a count variable.
message Count {
    int64 count = 1;
}
```

Also did we consider using -1 as a special value for count and capacity in the update to mean unspecified, as these would be invalid values? I don't know if these could be set by default to -1, so not setting them would leave them at -1.

markmandel:
So we use these standard wrappers all over the place - I asked the gRPC team what the canonical way to import the standard wrappers is and will get back to you.

markmandel:
Feedback I got was basically "if `const jspbWrappers = require('google-protobuf/google/protobuf/wrappers_pb');` works, then that is the correct solution, and why that is there.

So I think we should move forward with this approach 👍🏻

steven-supersolid:
Thank you and so it seems the library is missing some code as we should be able to do the following so a library user does not need to know the internal path

```
const jspbWrappers = require('google-protobuf').jspbWrappers
```

Or in the more modern format

```
const {jspbWrappers} = require('google-protobuf')
```

But yes, can go ahead with the current solution

steven-supersolid:
https://github.com/protocolbuffers/protobuf-javascript/issues/197

markmandel:
That's a new one

```
ERROR: (gcloud.container.clusters.get-credentials) ResponseError: code=404, message=Not found: projects/agones-images/locations/asia-east1/clusters/standard-e2e-test-cluster-1-26.
```

You can ignore that.

Restarting test.

markmandel:
Oooh, you will need to rebase against `main` - there is no 1.26 cluster anymore.

igooch:
@steven-supersolid we're updating the SDK, so that most of the methods that return (bool, error) are now simply returning (error). I updated the issue #3645. (Discussion: https://github.com/googleforgames/agones/pull/n #3581#discussion_r1505135182)

steven-supersolid:
Great, I think that is closer to the rest of the API. Will update

igooch:
@steven-supersolid we moved Counts and Lists from Alpha to Beta #3806. This is the last major expected change (until it goes into stable, but I imagine that will be several more releases away).

examples/nodejs-simple/gameserver-counts-and-lists.yaml
```diff
@@ -0,0 +1,46 @@
...
+apiVersion: agones.dev/v1
+kind: GameServer
+metadata:
+  generateName: simple-game-server-
+spec:
+  ports:
+    - name: default
+      portPolicy: Dynamic
+      containerPort: 7654
+  counters:
```
steven-supersolid, line 24: I made this to test with the local server but neither the counter or list can be found, so perhaps the structure is incorrect?
I started with
```
sdk-server.windows.amd64.exe --local --feature-gates CountsAndLists=true -f gameserver-counts-and-lists.yaml 
```

examples/nodejs-simple/src/index.js
```diff
@@ -56,32 +56,37 @@ const connect = async (timeout, enableAlpha) => {
 			process.exit(0);
 		});
 
-		await setTimeout(10000);
+		await setTimeout(5000);
```
steven-supersolid, line 59: Shortening the delay as there as so many calls now

sdks/nodejs/package.json
```diff
@@ -4,13 +4,13 @@
         "url": "https://github.com/googleforgames/agones/issues"
     },
     "dependencies": {
-        "@grpc/grpc-js": "1.7.3",
+        "@grpc/grpc-js": "1.10.7",
```
steven-supersolid, line 7: Updating dependencies and transitive dependencies to keep ahead of security issues

sdks/nodejs/src/alpha.js
```diff
@@ -83,7 +85,7 @@ class Alpha {
 		});
 	}
 
-	getPlayerCount() {
+	async getPlayerCount() {
```
steven-supersolid, line 88: Making async for consistency

test/sdk/nodejs/testSDKClient.js
```diff
@@ -66,4 +66,95 @@ const connect = async () => {
 	}
 };
 
+const testCounts = async(sdk) {
```
steven-supersolid, line 69: This is a port of the go test in `sdk-client-test.go`, however it is currently not executed as not sure yet how to do the equivalent of
```
if runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
```
If there is an environment variable set during the test then that could be used

steven-supersolid:
Tests are failing due new spys failing on the alpha client since merging in main and these also fail locally so perhaps client methods got renames

steven-supersolid:
It looks like it is due to removing Counts and Lists from alpha #3806 to beta. There is no beta protobuf file in JavaScript yet so need to generate one and update the sdk and tests

igooch:
There's a `make gen-sdk-grpc` command in the build directory, but it looks like it's ineffective for nodejs. We can file a separate bug that the autogeneration isn't working.

https://github.com/googleforgames/agones/blob/1b06e945ec086d14ceffdb02ef818b768e206b6a/build/includes/sdk.mk#L64-L69

steven-supersolid:
Once #3825 is done I'll merge main in and tests should pass
I need to update the example code to use the new beta SDK

steven-supersolid:
Before merging I need to look at one more change to see if we can get the new conformance tests to actually execute. Suggested approach
1. Set an environment variable in the build script when the test is run
2. In the testSDKClient check if this is set and execute the new tests

igooch:
Most of the conformance tests have `TESTS=$(DEFAULT_CONFORMANCE_TESTS),$(COUNTS_AND_LISTS_TESTS)` as part of the command. This checks that the local SDK server records each variable at some point during the test. https://github.com/googleforgames/agones/blob/bff72b1161492d7386e0bf39ca0753802c4b3e3a/build/includes/sdk.mk#L37
https://github.com/googleforgames/agones/blob/bff72b1161492d7386e0bf39ca0753802c4b3e3a/pkg/sdkserver/localsdk.go#L262
It looks like Node doesn't have this check -- can that be added in?

A side note that in order to get `make run-sdk-conformance-test-node` to run I first had to run `npm install` from the agones/sdks/nodejs directory since the `make build-images` did not install the node modules.

steven-supersolid:
I'm having a problem running the makefile and wonder if it's due to WSL?

```
make run-sdk-conformance-test-node
...
ERROR: invalid tag "/agones-sdk:1.41.0-dev-a786e2a-linux-amd64": invalid reference format
make[3]: *** [Makefile:612: build-agones-sdk-image-amd64] Error 1
make[3]: Leaving directory '/mnt/c/supersolid/project/agones/build'
make[2]: *** [includes/build-image.mk:63: ensure-image] Error 2
make[2]: Leaving directory '/mnt/c/supersolid/project/agones/build'
make[1]: *** [includes/build-image.mk:78: ensure-agones-sdk-image] Error 2
make[1]: Leaving directory '/mnt/c/supersolid/project/agones/build'
make: *** [includes/sdk.mk:168: run-sdk-conformance-test-node] Error 2
```

igooch:
> invalid reference format

You may need to set the $(REGISTRY) variable in ~/agones/build/Makefile. I'm not sure what your setup is for testing, for GKE there's instructions under agones/build/readme#running-a-test-google-kubernetes-engine-cluster for creating a registry and exporting the path to the registry.

steven-supersolid:
@igooch this should now be ready

igooch (commented):
A caveat that I'm not familiar with nodejs, so if you know of someone with more language specific knowledge that might have more insight, feel free to tag them in the comments.

Overall looks good, just a few clarifying comments.

Some of the language SDKs have unit tests. Do you think this would be beneficial to add, or would it not add anything beyond what's already being tested in test/sdk/nodejs/testSDKClient.js?

steven-supersolid:
> Some of the language SDKs have unit tests. Do you think this would be beneficial to add, or would it not add anything beyond what's already being tested in test/sdk/nodejs/testSDKClient.js?

The unit tests for the SDK can be found in sdks/nodejs/spec/ and all the new functions were covered with unit tests in this file sdks/nodejs/spec/betaAgonesSDK.spec.js in the PR. These are the must have tests and can be run by using `npm run test` in the sdk folder (assuming `npm install` has been run first). Code coverage is also set up and can be checked with `npm run cover`, the results are
```
78 specs, 0 failures
Finished in 0.198 seconds
--------------|---------|----------|---------|---------|-------------------
File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
--------------|---------|----------|---------|---------|-------------------
All files     |     100 |      100 |   98.96 |     100 | 
 agonesSDK.js |     100 |      100 |   97.14 |     100 | 
 alpha.js     |     100 |      100 |     100 |     100 | 
 beta.js      |     100 |      100 |     100 |     100 | 
--------------|---------|----------|---------|---------|-------------------
```
There is an uncovered inner function in an error handler on line 103 of agonesSDK.js which may not be possible to test. It's unrelated to counts and lists.

test/sdk/nodejs/testSDKClient.js
```diff
@@ -44,26 +46,118 @@ const connect = async () => {
...
 	} catch (error) {
 		console.error(error);
 	}
 };
 
+const testCounts = async(sdk) => {
+	// LocalSDKServer starting "rooms": {Count: 1, Capacity: 10}
+	const counter = "rooms";
+
+	try {
```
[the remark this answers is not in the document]
steven-supersolid: I've not used the Assert class much because it is usual to rely on a test runner framework such as jasmine or mocha instead.
In runtime code it is standard to rely on try/catch when dealing with async functions. In test code it is convenient to do the same.

I think in this case as we want to also check the return value it is easier to write the code with try/catch. We end up nesting in a try/catch block vs. nesting in a function call, which is preferable.

It may be worth investigating in another PR and also investigate if we could swap out jasmine for the new node.js built in test runner, which relies on assert.

examples/nodejs-simple/gameserver-counts-and-lists.yaml
steven-supersolid: Thank you, that's what I needed, so can perhaps remove this file and update the example

build/includes/sdk.mk
```diff
@@ -165,7 +165,8 @@ run-sdk-conformance-test-cpp:
 	$(MAKE) run-sdk-conformance-test SDK_FOLDER=cpp GRPC_PORT=9003 HTTP_PORT=9103
 
 run-sdk-conformance-test-node:
-	$(MAKE) run-sdk-conformance-test SDK_FOLDER=node GRPC_PORT=9002 HTTP_PORT=9102
+	# run with on-by-default (Beta) feature flags enabled
+	$(MAKE) run-sdk-conformance-test SDK_FOLDER=node GRPC_PORT=9002 HTTP_PORT=9102 TESTS=$(DEFAULT_CONFORMANCE_TESTS),$(COUNTS_AND_LISTS_TESTS)
```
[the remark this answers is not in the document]
steven-supersolid: We would need to execute `npm install` in the sdk folder, e.g. in build-sdk-test.sh the minimum required is
```
cd /go/src/agones.dev/agones/test/sdk/nodejs
npm install --quiet
```
Could we add something like that here? I don't know enough about these makefiles

build/includes/sdk.mk
```diff
@@ -165,7 +165,8 @@ run-sdk-conformance-test-cpp:
 	$(MAKE) run-sdk-conformance-test SDK_FOLDER=cpp GRPC_PORT=9003 HTTP_PORT=9103
 
 run-sdk-conformance-test-node:
-	$(MAKE) run-sdk-conformance-test SDK_FOLDER=node GRPC_PORT=9002 HTTP_PORT=9102
+	# run with on-by-default (Beta) feature flags enabled
```
igooch, line 168:
```suggestion
	# run with on-by-default (Beta) feature flags enabled. If running locally first run `SDK_FOLDER=node make test-sdk` to build dependencies.
```

build/includes/sdk.mk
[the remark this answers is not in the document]
igooch: Probably not worth modifying the make command, since the test gets build properly during CI / CD, so this only applies to local testing. I added in a suggested comment change above.

igooch (approved):
LGTM. Thanks for all your work on this!

steven-supersolid:
HTML tests failing after merging in main
```
Html Test: Attempt 20
htmltest started at 08:58:03 on /tmp/website
site/docs/guides/client-sdks/unreal/index.html
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/setting-up-dedicated-servers-in-unreal-engine/
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/game-mode-and-game-state-in-unreal-engine/
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/API/Runtime/Engine/GameFramework/AGameMode/
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/API/Runtime/Engine/GameFramework/AGameSession/
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/build-operations-cooking-packaging-deploying-and-running-projects-in-unreal-engine/
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/API/Runtime/Engine/GameFramework/AGameMode/index.html
  Non-OK status: 403 --- site/docs/guides/client-sdks/unreal/index.html --> https://docs.unrealengine.com/en-US/API/Runtime/Engine/GameFramework/AGameSession/index.html
✘✘✘ failed in 2.372015068s
8 errors in 198 documents
make: *** [includes/website.mk:76: site-test] Error 2
```

igooch:
> HTML tests failing after merging in main

Oh yep. Fix in #3847. This PR should be good to merge once the fix goes through.

markmandel:
@steven-supersolid we can never auto-update your PRs 😃 we're heading into the weekend, so please rebase against `main` and it'll automerge when passing CI 👍🏻

PR #3735, opened 2024-03-27
Docs: remove unused specifications in cert-manager helm example

qhyun2:
**What type of PR is this?**
/kind documentation

These lines do not have any effect and can be removed since they are not referenced in the helm chart.

https://github.com/search?q=repo%3Agoogleforgames%2Fagones+path%3A%2F%5Einstall%5C%2Fhelm%5C%2Fagones%5C%2Ftemplates%5C%2F%2F+disableSecret&type=code

https://github.com/search?q=repo%3Agoogleforgames%2Fagones+path%3A%2F%5Einstall%5C%2Fhelm%5C%2Fagones%5C%2Ftemplates%5C%2F%2F+allocationApiService&type=code

https://github.com/search?q=repo%3Agoogleforgames%2Fagones+path%3A%2F%5Einstall%5C%2Fhelm%5C%2Fagones%5C%2Ftemplates%5C%2F%2F+validatingWebhook&type=code

https://github.com/search?q=repo%3Agoogleforgames%2Fagones+path%3A%2F%5Einstall%5C%2Fhelm%5C%2Fagones%5C%2Ftemplates%5C%2F%2F+mutatingWebhook&type=code


markmandel:
Blerg - this actually I think is a bug. From memory, we were supposed to move this helm parameters over to the new values in #2797, but it looks like we didn't.

@zmerlynn @gongmax - any thoughts on what we want to do here?

Should we add an issue to deprecate the `agones.controller` versions of these fields, and actually create the newer `agones.extensions` fields?

In helm, I wonder if we can set it up so that `agones.controller` values, if set, could overwrite the `agones.extensions` values?

markmandel:
Just making a note to self (or someone), we should turn this into an issue, and make the fixes.

zmerlynn:
AIUI we're going to handle this via #3773 right?

markmandel:
> AIUI we're going to handle this via #3773 right?

Yes, good call. Thanks for bringing this to our attention @qhyun2 - but I'll close this for now, since we'll take this on at the above issue and PR.

PR #3734, opened 2024-03-27
Feat/counter and list defaulting order to ascending

lacroixthomas:
**What type of PR is this?**
/kind feature

**What this PR does / Why we need it**:
Defaulting the `order` from counter and list priorities to `Ascending`

**Which issue(s) this PR fixes**:

Closes #3683 

**Special notes for your reviewer**:

PR #3736, opened 2024-03-27
Counter and List Priorities: Set Default to Ascending for Fleet and G…

### LINKED ISSUE #3683: Counter and List Priorities: Default to Ascending.

A small quality of life improvement - it would be really nice for `Fleet.Spec.Priorities` and `GameServerAllocation.Spec.Priorities` the default for `order` as `Ascending`.

If not defined, make the order `Ascending`.

Leave it alone and let people be explicit with their ordering.

**Additional context**

We have the following in our CRD:

https://github.com/googleforgames/agones/blob/1c536c0aec680e8b013479bc15481c4ae47e3185/install/helm/agones/templates/crds/fleet.yaml#L129-L134

So at some point, someone figured we should do this 😄

Kalaiselvi84:
…ameServerAllocation

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix
> /kind release

Closes #3683

### DIFF
--- a/install/helm/agones/templates/crds/fleet.yaml
+++ b/install/helm/agones/templates/crds/fleet.yaml
@@ -132,6 +132,7 @@ spec:
                         enum:
                           - Ascending
                           - Descending
+                        default: Ascending
                 template:
                   {{- $data := dict "metadata" true "podPreserveUnknownFields" .Values.gameservers.podPreserveUnknownFields }}
                   {{- include "gameserver.schema" $data | indent 17 }}
--- a/pkg/apis/allocation/v1/gameserverallocation.go
+++ b/pkg/apis/allocation/v1/gameserverallocation.go
@@ -18,9 +18,6 @@ import (
 	"errors"
 	"fmt"
 
-	"agones.dev/agones/pkg/apis"
-	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
-	"agones.dev/agones/pkg/util/runtime"
 	"github.com/mitchellh/hashstructure/v2"
 	corev1 "k8s.io/api/core/v1"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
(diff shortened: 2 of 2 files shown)


install/helm/agones/templates/crds/fleet.yaml
```diff
@@ -132,6 +132,7 @@ spec:
                         enum:
                           - Ascending
                           - Descending
+                        default: Ascending
```
markmandel, line 135: Need to run `make gen-install` please 😄

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -564,6 +565,13 @@ type GameServerMetadata struct {
 
 // ApplyDefaults applies the default values to this GameServerAllocation
 func (gsa *GameServerAllocation) ApplyDefaults() {
+	if len(gsa.Spec.Priorities) == 0 {
```
markmandel, line 568: Please add tests to:

https://github.com/googleforgames/agones/blob/a1a1cd956a19e4e97fd67f6c0ad06da4259273ab/pkg/apis/allocation/v1/gameserverallocation_test.go#L31

pkg/apis/allocation/v1/gameserverallocation.go
```diff
@@ -564,6 +565,13 @@ type GameServerMetadata struct {
...
+		defaultPriority := agonesv1.Priority{
+			Order: "Ascending",
```
markmandel, line 570: Rather than a string, use:

https://github.com/googleforgames/agones/blob/e696b033c9ed0ffea489743ce9d3ea5c82de22a5/pkg/apis/agones/v1/common.go#L46

pkg/apis/allocation/v1/gameserverallocation.go
markmandel, line 568: This implementation isn't correct unfortunately - #3734 has it correct - need to loop around each one.

Kalaiselvi84: I haven't seen the other PR. Since the correct approach is in the above mentioned PR, I will close this PR.

Issue #2468, opened 2022-02-08
How to scale beyond the cluster?


emrys90:
I'm new to kubernetes/docker, so I apologize if I use any wrong terms here. I plan on hosting this on Linode as the costs of GCP/AWS is just too high in comparison. Linode has a limit of 100 nodes per cluster from what I am reading in their documentation.

How do you manage auto scaling beyond one cluster using Agones, or is that not possible?

Would we need to keep an eye on if a cluster is starting to get full and then install another instance of Agones on a new cluster?

markmandel:
Since multi-cluster work tends to be platform specific, you will need to implement this yourself for your specific platform.

We do have https://agones.dev/site/docs/advanced/multi-cluster-allocation/, although that doesn't answer your specific question.

markmandel:
Bumping out of stale. Area of research internally, will share some ideas for discussion when complete.

markmandel:
@gongmax did you want to keep this open?

markmandel:
Keeping this open!

PR #3744, opened 2024-04-04
New recommendation for multi-cluster allocation

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix
> /kind release

Offline we tend to recommend people use a Service Mesh for multi-cluster allocation, but I keep seeing people running into our docs and trying to implement our solution.

Rather than waiting on a full investigation and documentation, I wanted to put a note at the top making a recommendation to go in a different direction - so that people don't actively chase this path if they aren't aware there are other (better) paths for multi-cluster communications.

Work on #2468

N/A

### DIFF
--- a/site/content/en/docs/Advanced/multi-cluster-allocation.md
+++ b/site/content/en/docs/Advanced/multi-cluster-allocation.md
@@ -5,6 +5,23 @@ description: >
   In order to allow allocation from multiple clusters, Agones provides a mechanism to set redirect rules for allocation requests to the right cluster.
 ---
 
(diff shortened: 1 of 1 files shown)


roberthbailey (approved):

roberthbailey:
That's a weird one:

`ERROR: gcloud crashed (FileNotFoundError): [Errno 2] No such file or directory: 'site/public/index.xml'`

markmandel:
Yeah, I feel like I've seen that... maybe before?

Issue #3553, opened 2023-12-19
Flaky: TestGameServerAllocationDuringMultipleAllocationClients


Kalaiselvi84:
**What happened**:

The failed e2e test:  TestGameServerAllocationDuringMultipleAllocationClients

**What you expected to happen**:

Pass all the time

**How to reproduce it (as minimally and precisely as possible)**:

It fails sometimes in the CI

**Anything else we need to know?**:

```
VERBOSE:     gameserverallocation_test.go:1417: 
VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserverallocation_test.go:1417
VERBOSE:         	Error:      	Not equal: 
VERBOSE:         	            	expected: 100
VERBOSE:         	            	actual  : 99
VERBOSE:         	Test:       	TestGameServerAllocationDuringMultipleAllocationClients
VERBOSE: --- FAIL: TestGameServerAllocationDuringMultipleAllocationClients (38.74s)
```

**Environment**:
- Agones version: Dev
- Kubernetes version (use `kubectl version`): 1.28
- Cloud provider or hardware configuration: GKE
- Install method (yaml/helm): helm
- Troubleshooting guide log(s): [https://console.cloud.google.com/cloud-build/builds/8c2fe09e-b77b-4b9b-9ef5-5a5670ec7794?project=258182270954](https://pantheon.corp.google.com/cloud-build/builds/8c2fe09e-b77b-4b9b-9ef5-5a5670ec7794;step=2?project=agones-images&mods=logs_tg_prod)

markmandel:
More and more!

https://console.cloud.google.com/cloud-build/builds/6322a904-b37c-4bcc-8e04-5e29a8315c21?project=258182270954

This is killing me.

zmerlynn:
> https://console.cloud.google.com/cloud-build/builds/6322a904-b37c-4bcc-8e04-5e29a8315c21?project=258182270954

`could not completed gsa1 allocation : an error on the server ("error updating allocated gameserver: Operation cannot be fulfilled on gameservers.agones.dev \"preferred-m8hsh-mzrd8-g682d\": the object has been modified; please apply your changes to the latest version and try again`

This is the same conflict @gongmax sees in perf tests, and I assume it's because two allocators try to allocate the same GameServer but fail (and it's somehow not retried by the allocator).

zmerlynn:
This is also mentioned in the load test README:
[...]

PR #3747, opened 2024-04-08
feat: support multiple port ranges

### LINKED ISSUE #1911: Move to multiple dynamic range ports
Hi Mark,

Our game server uses 4 passthrough ports:
- 1 game port
- 3 tools ports
I would like to have the game port opened to the whole universe but not the 3 tools ports for which i want to allow only some sources

Proposal:
I imagine we can move port range (gameservers.minPort and gameservers.maxPort) to a list of port ranges and choose the port range to use in the game server ports definition.
So, we can define a different firewall rule for each range.

ie.
install values:
```
...
gameservers:
  namespaces: ["default"]
  ranges:
  - name: game
    minPort: 7000
    maxPort: 7099
  - name: tools
    minPort: 7100
    maxPort: 7199
```

and gamerserver definition:
```
apiVersion: "agones.dev/v1"
kind: GameServer
...
spec:
  # if there is more than one container, specify which one is the game server
  container: example-server
  # Array of ports that can be exposed as direct connections to the game server container
  ports:
  - name: default
    range: game
    portPolicy: Passthrough
    protocol: UDP
  - name: tool1
    range: tools
    portPolicy: Passthrough
    protocol: TCP
  - name: tool2
    range: tools
    portPolicy: Passthrough
    protocol: TCP
  - name: tool3
    range: tools
    portPolicy: Passthrough
    protocol: TCP
...
```

Workaround:
If the game port is UDP and Tools ports are TCP, you can only filter sources for TCP destinations.
But it's not possible if you want to apply different firewall rules for the same protocol

Thank you
Regards
Stéphane

nrwiersma:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

This adds support for multiple named port ranges that can be configured in the GameServer port using the `range` property. This is behind a feature flag `PortRanges`.

Closes #1911

I really dislike how the configuration has ended up, but I did not find a better way to do it. If you have a better idea, I would gladly take it.


nrwiersma:
I dont really understand what caused the e2e tests to fail. If you could give me a hit where to look, that would be great.

markmandel:
```
    gameserverallocation_test.go:1442: 
        	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserverallocation_test.go:1442
        	Error:      	Not equal: 
        	            	expected: 100
        	            	actual  : 98
        	Test:       	TestGameServerAllocationDuringMultipleAllocationClients

```


uuuh this test is killing me.

markmandel:
Sorry - this a flake #3553 - it's been driving me up the wall.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
```diff
@@ -52,6 +52,9 @@ properties:
             name:
               title: Name is the descriptive name of the port
               type: string
+            range:
```
markmandel, line 55: Should this default to some value? "default" maybe is the default port range? (I think that's what we had in the initial ticket?)

pkg/portallocator/portallocator.go
```diff
@@ -49,12 +49,76 @@ type Interface interface {
...
+		a.DeAllocate(gs)
+	}
+}
+
 // A set of port allocations for a node
 type portAllocation map[int32]bool
 
 //nolint:govet // ignore fieldalignment, singleton
-type portAllocator struct {
+type portRangeAllocator struct {
```
markmandel, line 119: I like this refactor 👍🏻

pkg/cloudproduct/cloudproduct.go
```diff
@@ -48,7 +48,7 @@ type ControllerHooksInterface interface {
 	SyncPodPortsToGameServer(*agonesv1.GameServer, *corev1.Pod) error
 
 	// NewPortAllocator creates a PortAllocator. See gameservers.NewPortAllocator for parameters.
-	NewPortAllocator(int32, int32, informers.SharedInformerFactory, externalversions.SharedInformerFactory) portallocator.Interface
+	NewPortAllocator(int32, int32, map[string]portallocator.PortRange, informers.SharedInformerFactory, externalversions.SharedInformerFactory) portallocator.Interface
```
markmandel, line 51: @zmerlynn can you take a peek here - since this is your area of expertise.

cmd/controller/main.go
```diff
@@ -282,7 +287,9 @@ func parseEnvFlags() config {
 	pflag.Bool(pullSidecarFlag, viper.GetBool(pullSidecarFlag), "For development purposes, set the sidecar image to have a ImagePullPolicy of Always. Can also use ALWAYS_PULL_SIDECAR env variable")
 	pflag.String(sdkServerAccountFlag, viper.GetString(sdkServerAccountFlag), "Overwrite what service account default for GameServer Pods. Defaults to Can also use SDK_SERVICE_ACCOUNT")
 	pflag.Int32(minPortFlag, 0, "Required. The minimum port that that a GameServer can be allocated to. Can also use MIN_PORT env variable.")
-	pflag.Int32(maxPortFlag, 0, "Required. The maximum port that that a GameServer can be allocated to. Can also use MAX_PORT env variable")
+	pflag.Int32(maxPortFlag, 0, "Required. The maximum port that that a GameServer can be allocated to. Can also use MAX_PORT env variable.")
+	pflag.StringToInt(portRangeMinPortFlag, cast.ToStringMapInt(viper.GetStringMap(portRangeMinPortFlag)), "Optional. List of port range min ports. Can Also use PORT_RANGE_MIN_PORT env variable.")
```
markmandel, line 291: Out of curiosity (or maybe also docs?) what's the format here for the list?

markmandel (commented):
Looks good - I can't think of a better way! Added some questions and tweaks.

Some extra docs we will need:

* https://agones.dev/site/docs/reference/gameserver/
* https://github.com/googleforgames/agones/blob/main/examples/gameserver.yaml

I think that's about it. I may have some extra tweaks to the docs as I reread, but not seeing any issues. Nice work 👍🏻

install/helm/agones/templates/crds/_gameserverspecschema.yaml
nrwiersma: I left it empty for the default case to make it backwards compatible for existing installations. Having an empty default also allows for it to be omitted if not being used. Happy to defer here.

cmd/controller/main.go
nrwiersma: I had originally thought from a perusal of the code it was the standard `key=value`, but in reality it is a JSON object. I don't love it.

nrwiersma: I have reworded to include the format in the description.

pkg/cloudproduct/cloudproduct.go
zmerlynn: The `gkeAutopilot` implementation of `APIHooks` has a `ValidateGameServerSpec` hook already: https://github.com/googleforgames/agones/blob/c1320eedf6ea21cb8e5e6a5e12005881ba725a64/pkg/cloudproduct/gke/gke.go#L134-L146 - you could reject `range` here the same way we reject everything but `Dynamic` today.

That said, we really need to reject this at _install_ time so people don't trip into this blindly. We don't have a way to do that in helm yet - do you have any ideas?

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -326,7 +326,10 @@ The following tables lists the configurable parameters of the Agones chart and t
...
-| `gameservers.podPreserveUnknownFields`                   | Disable [field pruning][pruning] and schema validation on the Pod template for a [GameServer][gameserver] definition                                                                                                                                                                                             | `false`                                   |
+{{% feature publishVersion="1.40.0" %}}
+ | `gameservers.portRanges`                                 | Port ranges from which to do named dynamic port allocation                                                                                                                                                                                                                                                       | `[]`                                      |
```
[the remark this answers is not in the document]
zmerlynn: I think this needs more documentation on format for sure. I would rather just a full JSON map format here if that's not what's already envisaged, I.e. something like `{"name1": [5000, 6000], "name2": [6000, 7000]}`.

install/helm/agones/templates/crds/_gameserverspecschema.yaml
zmerlynn: Unless we can find a way forward for detecting Autopilot with helm, `range` needs to be empty to know that someone is trying to use an unsupported feature. I don't love that.

site/content/en/docs/Installation/Install Agones/helm.md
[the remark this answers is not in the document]
zmerlynn: Oh, I see you added the documentation in a later commit, apologies.

cmd/controller/main.go
zmerlynn: I don't think this should be two concurrent flags. I would prefer a single flag that took a structure JSON blob as config (e.g. as I mentioned here: https://github.com/googleforgames/agones/pull/3747#discussion_r1557890594), it's a little weird to take two?

pkg/util/runtime/features.go
```diff
@@ -102,6 +105,7 @@ var (
 		FeatureGKEAutopilotExtendedDurationPods: false,
 		FeaturePlayerAllocationFilter:           false,
 		FeaturePlayerTracking:                   false,
+		FeaturePortRanges:                       false,
```
zmerlynn, line 108: There's a checklist above this on line 73 you need to follow a few more steps in, if you want to get this to Alpha:
* Add this to install/helm/agones/defaultfeaturegates.yaml
* Add to `ALPHA_FEATURE_GATES` in build/Makefile
* Add the inverse to the e2e-runner config in cloudbuild.yaml
* Add it to feature-stages.md

pkg/cloudproduct/cloudproduct.go
nrwiersma: I seem to remember some form of helm helper check that could throw an error. I would need to look for it again, but I guess that could be what we are looking for.

install/helm/agones/templates/controller.yaml
```diff
@@ -110,6 +110,18 @@ spec:
         # maximum port that can be exposed to GameServer traffic
         - name: MAX_PORT
           value: {{ .Values.gameservers.maxPort | quote }}
+        {{- if .Values.gameservers.portRanges }}
```
zmerlynn, line 113: Please reject `.Values.gameservers.portRanges` unless the gate is enabled. You can see an example of both feature gate detection and helm failure in this blob (which is no longer in the codebase - I think we've promoted all of the helm-relevant feature gates): https://github.com/googleforgames/agones/blob/3d57eaa4fe18590634c17137e369302c981b91b9/install/helm/agones/templates/controller.yaml#L15-L22

zmerlynn (changes requested):

cmd/controller/main.go
zmerlynn: I'm not necessarily saying to force the JSON all the way from the helm - doesn't gotemplate support building maps? you can rebuild the structure in the helm template, not dissimilar to what you're doing today. And what I mean is: I like the configuration in helm, but I think the helm template could translate to JSON. It's fussy but I think it's doable.

pkg/cloudproduct/cloudproduct.go
zmerlynn: The problem is the runtime check is tricky - as an example, we already do this here for the controller: https://github.com/googleforgames/agones/blob/main/pkg/cloudproduct/gke/gke.go#L78-L86 .. but doing that in helm is basically impossible (and doesn't make sense when you're just using helm as a renderer). :/ I'll keep thinking on this.

cmd/controller/main.go
nrwiersma: When I say configure from YAML or TOML, I mean config from a file, not from helm. This is the primary use of `viper` is to allow CLI, Env and File config.

nrwiersma:
A hair brained idea. Given the config for this feature is really not great, and e2e tests seem to be tricky at best, what if the config for this feature came in the form of a CRD. The port ranges could be made immutable, and still checked for overlap via validation webhooks. It does, without a massive amount of thought, seems like it would solve a couple issues here. It does open a few new and exciting routes to failure, but I think some of those are somewhat unavoidable in any case.

markmandel:
> A hair brained idea. Given the config for this feature is really not great, and e2e tests seem to be tricky at best, what if the config for this feature came in the form of a CRD. 

Given that the rest of the config for Agones is provided via Helm configuration parameters it seems weird to seperate this out.

That being said, seems like a good time to start a https://helm.sh/docs/topics/charts/#schema-files ?

markmandel:
> Oh, one last request (and thank you for your patience): Please add e2e tests. It's a little tricky because for normal e2e tests we use the same install, but here you need to install it a different way and then test. You can add "custom" tests in this area:

@zmerlynn could we add some port ranges to 
https://github.com/googleforgames/agones/blob/05d8232c937d756d9b8e848915e04e4159924b7b/build/e2e-image/e2e.sh#L25C18-L25C25

And call it a day? Does it need to be a separate install?

install/helm/agones/values.yaml
```diff
@@ -280,6 +280,10 @@ gameservers:
   - default
   minPort: 7000
   maxPort: 8000
+  portRanges: []
```
markmandel, line 283: Chatting with @zmerlynn offline - we were discussing calling this `additionalPortRanges` - since it's in addition to `minPort` and `maxPort` ? WDYT?

markmandel (commented):
I've lost track through this, but a capturing some more thoughts @zmerlynn where chatting about and wanted to get your take.

1) the current `minPort` and `maxPort` range should be called `default` - that way if people are naming ranges, they can use `default` (which would also be the default, so it's backward compatible).
2) For validation for `GameServerSpec` on Autopilot, we could then say "if you pick a port range other than `default` we will return an invalid configuration" (with appropriate documentation).

This means we don't have to validate that they are on autopilot at the helm install layer, and the end user gets a nice error message if they try and do something they shouldn't / can't on Autopilot that isn't as scary as a controller crash or similar.

WDYT?

site/content/en/docs/Reference/gameserver.md
```diff
@@ -25,6 +25,10 @@ spec:
   ports:
     # name is a descriptive name for the port
   - name: default
+    # [Stage:Alpha]
+    # [FeatureFlag:PortRanges]
+    # range is the optional port range name from which to select a port when using a 'Dynamic' or 'Passthrough' port policy.
+    range: game
```
markmandel, line 31: Should the example be `default` ? 🤔

markmandel:
Just to make life easy - here's the validation block for `GameServerSpec`:

https://github.com/googleforgames/agones/blob/4bb673186b2ee8431de00c5c564bd1daa1a356df/pkg/apis/agones/v1/gameserver.go#L511

Which feeds into:

https://github.com/googleforgames/agones/blob/4bb673186b2ee8431de00c5c564bd1daa1a356df/pkg/cloudproduct/gke/gke.go#L134

For the Autopilot specific validation.

(I don't think you implemented this yet)

nrwiersma:
> For the Autopilot specific validation.

I did this: https://github.com/googleforgames/agones/pull/3747/files#diff-6df711a53a9866de14c530a4b1f9316d142ead0bc9a8e9d668c50a2ff28673b2R142

Is that not correct, or did I missing something extra I need to do?

zmerlynn:
> > For the Autopilot specific validation.
> 
> I did this: https://github.com/googleforgames/agones/pull/3747/files#diff-6df711a53a9866de14c530a4b1f9316d142ead0bc9a8e9d668c50a2ff28673b2R142
> 
> Is that not correct, or did I missing something extra I need to do?

I think you checked it in the right place - I think @markmandel may be off here.

cmd/controller/main.go
```diff
@@ -449,6 +482,59 @@ func validateResource(request resource.Quantity, limit resource.Quantity, resour
...
+		validationErrors = append(validationErrors, errors.New("min Port and Max Port values are required"+rangeCtx))
+	}
+	if maxPort < minPort {
+		validationErrors = append(validationErrors, errors.New("max Port cannot be set less that the Min Port"+rangeCtx))
+	}
+	return validationErrors
+}
+
+func overlaps(minA, maxA, minB, maxB int32) bool {
+	return math.Max(float64(minA), float64(minB)) < math.Min(float64(maxA), float64(maxB))
```
zmerlynn, line 535: nit: Go 1.21 introduced generic https://pkg.go.dev/builtin#max, which avoids the cast and the math import.

cmd/controller/main.go
```diff
@@ -355,9 +362,15 @@ func parseEnvFlags() config {
 		logger.WithError(err).Fatalf("could not parse %s", sidecarMemoryLimitFlag)
 	}
 
+	portRanges, err := parsePortRanges(viper.GetString(additionalPortRangesFlag))
```
zmerlynn, line 365: It's a little awkward here that you're carrying the default port range and the additional port ranges separately. I'm ok with them split on the config surface for backwards compat, but in the controller, my thinking is that we should just add `default` to the map right here and change `config` to use the `portRanges` map.

install/helm/agones/templates/controller.yaml
```diff
@@ -110,6 +110,14 @@ spec:
         # maximum port that can be exposed to GameServer traffic
         - name: MAX_PORT
           value: {{ .Values.gameservers.maxPort | quote }}
+        {{- if .Values.gameservers.additionalPortRanges }}
+        {{- $featureGates := include "agones.featureGates" . | fromYaml }}
+        {{- if not $featureGates.PortRanges }}
+        {{- fail "gameservers.portRanges specified without feature gate PortRanges enabled!" }}
```
zmerlynn, line 116: s/portRanges/additionalPortRanges/

install/helm/agones/values.yaml
```diff
@@ -280,6 +280,8 @@ gameservers:
   - default
   minPort: 7000
   maxPort: 8000
+  additionalPortRanges: {}
+    # game: [9000, 10000]
```
zmerlynn, line 284: Probably call out the feature gate needed here

install/yaml/install.yaml
```diff
@@ -16335,6 +16341,9 @@ spec:
                               name:
                                 title: Name is the descriptive name of the port
                                 type: string
+                              range:
+                                title: the port range name from which to select a port when using a 'Dynamic' or 'Passthrough' port policy.
```
zmerlynn, line 16345: Add `Defaults to 'default'.`

pkg/cloudproduct/cloudproduct.go
zmerlynn: We came to a reasonable resolution here of enforcing it at runtime where we have good contact with the user.

zmerlynn (commented):
This generally LGTM - I have a few not super important nits. TBH, if you can't get to them before Tuesday, I'm also willing to merge on Monday and handle as a fast-follow PR.

markmandel:
> > > For the Autopilot specific validation.
> > 
> > I did this: https://github.com/googleforgames/agones/pull/3747/files#diff-6df711a53a9866de14c530a4b1f9316d142ead0bc9a8e9d668c50a2ff28673b2R142
> > Is that not correct, or did I missing something extra I need to do?
> 
> I think you checked it in the right place - I think @markmandel may be off here.



Ah yep - same spot. I just didn't see the change in the big PR 👍🏻

zmerlynn (commented):
This LGTM - we still need to figure out e2es but I think we can do that in a follow-on. One idea I had for that is just to introduce port ranges to the GKE Standard config and then make sure we have an e2e that exercises them.

markmandel (commented):
Sorry about this stalling a bit.

Found a couple of doc tweaks (and there are a couple of conflicts to resolve), but once they are done, this is definitely good to merge 👍🏻  (and we can close a **4 year old** issue!!!!)

zmerlynn (approved):
LGTM, thank you!

nrwiersma:
@zmerlynn It turns out there was a test flake due to assuming the order of the input map. I removed the offending assertions, as the assertions above already prove they came from the correct port ranges.

zmerlynn:
@nrwiersma Thanks for fixing it, LGTM!

PR #3749, opened 2024-04-09
Set Minimum Buffer Size to 1

### LINKED ISSUE #3715: Scale Down by Percent CounterPolicy or ListPolicy Can Lead to 0 Replicas
**What happened**:

Previously allowed behavior, scale to 0 replicas, was not updated. The percent buffer when there are no allocated Counters / Lists gets calculated as 0 https://github.com/googleforgames/agones/blob/732c075b90ae1f5b6da4810f05808496f612da3b/pkg/fleetautoscalers/fleetautoscalers.go#L318-L325 which is not allowed per https://github.com/googleforgames/agones/blob/732c075b90ae1f5b6da4810f05808496f612da3b/pkg/apis/autoscaling/v1/fleetautoscaler.go#L323-L325. 

Do not let the fleetautoscaler buffer or desired replicas go to 0.

https://github.com/googleforgames/agones/blob/732c075b90ae1f5b6da4810f05808496f612da3b/pkg/fleetautoscalers/fleetautoscalers_test.go#L1895

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #3715 


pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -321,8 +321,20 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
 		}
 		// The desired TOTAL capacity based on the Aggregated Allocated Counts (see applyBufferPolicy for explanation)
 		desiredCapacity := int64(math.Ceil(float64(aggAllocatedCount*100) / float64(100-bufferPercent)))
+
+		// Ensures desiredCapacity is at least 1
+		if desiredCapacity < 1 {
```
igooch, line 326: Since you check `if buffer < 1` you don't need to check desired capacity.

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -1950,7 +1951,7 @@ func TestApplyListPolicy(t *testing.T) {
 							}}}},
 			},
 			want: expected{
-				replicas: 0,
+				replicas: 1,
```
igooch, line 1954: @markmandel in the case where the autoscaler calculates percent buffer as 0 (when there are no allocated Counters / Lists) we manually change the desired buffer to 1. Should this be "limited", or not limited, or does it not matter?

pkg/fleetautoscalers/fleetautoscalers.go
Kalaiselvi84: I've eliminated the desiredCapacity check

markmandel:
@igooch is there any e2e tests here to update that would also need to be updated?

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -1950,7 +1951,7 @@ func TestApplyListPolicy(t *testing.T) {
...
 				limited:  false,
```
igooch, line 1955:
```suggestion
				limited:  true,
```

Per Mark's comment above this should be "true", so the logic will need to be modified in fleet_autoscalers.

Kalaiselvi84: Please review the latest commit, which ensures that the replicas never scale down to zero. It also modifies the TestApplyListPolicy in fleetautoscaler_test.go to reflect this change.👍🏻

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -345,8 +346,14 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
 			return scaleLimited(scale, f, gameServerLister, nodeCounts, key, isCounter, replicas,
 				capacity, aggCapacity, minCapacity, maxCapacity)
 		}
-		return scaleDown(f, gameServerLister, nodeCounts, key, isCounter, replicas, aggCount,
-			aggCapacity, minCapacity, buffer)
+		if replicas > 1 {
```
igooch, line 349: Why this line `if replicas > 1`?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -345,8 +346,14 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+			result, isLimited, err := scaleDown(f, gameServerLister, nodeCounts, key, isCounter, replicas, aggCount,
+				aggCapacity, minCapacity, buffer)
+			if result <= 1 {
+				return 1, true, nil
+			}
+			return result, isLimited, err
```
markmandel, line 355:
```suggestion
	  if result <= 1 {
		  return 1, true, nil
	  }

    return scaleDown(f, gameServerLister, nodeCounts, key, isCounter, replicas, aggCount,
				aggCapacity, minCapacity, buffer)
```

@igooch is that what you meant? (Also just a bit neater)

Kalaiselvi84: Thank you for the clarification. I just wanted to confirm one detail, did you mean to use replicas in the condition?

```
if replicas <= 1 {
		  return 1, true, nil
	  }
    return scaleDown(f, gameServerLister, nodeCounts, key, isCounter, replicas, aggCount,
				aggCapacity, minCapacity, buffer)
```

igooch: We may need some more time to go over this logic. The current number of replicas is not directly relevant to Counts and Lists as the desired replicas are not calculated based on current replicas. 

Also the logic if 
```
result <= 1 {
    return 1, true, nil
}
```
will also not be true in all cases. There are some cases where 1 is the desired number of replicas and it was not limited.

markmandel: oooh, I see what you saying.

It probably should be:

```go
result < 1 {
    return 1, true, nil
}
```

?

markmandel: Admitedly, I should sit down and review this code again, because I think I might have taken too cursory a glance.

Should it actually be `result < b.MinReplicas` (are we checking Mn/Max values right now I could see it?)

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -470,8 +471,8 @@ func scaleDown(f *agonesv1.Fleet, gameServerLister listeragonesv1.GameServerList
 	nodeCounts map[string]gameservers.NodeCount, key string, isCounter bool, replicas int32,
 	aggCount, aggCapacity, minCapacity, buffer int64) (int32, bool, error) {
 	// Exit early if we're already at MinCapacity to avoid calling getSortedGameServers if unnecessary
-	if aggCapacity == minCapacity {
-		return replicas, true, nil
+	if aggCapacity <= minCapacity {
+		return max(replicas, 1), true, nil
```
Kalaiselvi84, line 475: I've updated the `scaleDown` function to always maintain at least one replica. Please let me know what you think of this modification.

markmandel:
```suggestion
		return max(replicas, minCapacity), true, nil
```

Should it be this @igooch ?

Kalaiselvi84: I tried it on my local, `TestApplyListPolicy` passed and `TestApplyCounterPolicy/scale_down_at_MinCapacity`
failed:

 ```
/agones/pkg/fleetautoscalers/fleetautoscalers_test.go:1451
                Error:          Not equal:
                                expected: 10
                                actual  : 70
                Test:           TestApplyCounterPolicy/scale_down_at_MinCapacity
--- FAIL: TestApplyCounterPolicy/scale_down_at_MinCapacity (0.10s)
```

Comment (markmandel) [reply pkg/fleetautoscalers/fleetautoscalers.go:475]: Yeah, I think I missed the mark here. I _think_ you might be right, but this is making my head hurt.

I think you actually are correct in your approach here. I'd just like @igooch to confirm.

This looks right to me I think.

Comment (igooch) [reply pkg/fleetautoscalers/fleetautoscalers.go:475]: Although it's unlikely for replicas < 1, the 
```
if aggCapacity <= minCapacity {
		return max(replicas, 1), true, nil
```
is correct.

I still need to take the time to run through the rest of the logic here, and make sure the tests are robust enough.

Comment (Kalaiselvi84) [reply pkg/fleetautoscalers/fleetautoscalers.go:475]: Thank you both for your feedback. Please take your time to review everything thoroughly. I'm happy to do further adjustments to ensure everything is robust.👍🏻

**Review by igooch — COMMENTED** [review]
On about line 322 I'd recommend doing something like:

```
		// If the Aggregated Allocated Counts is 0 then desired capacity gets calculated as 0. If the
		// capacity of 1 replica is equal to or greater than minimum capacity we can exit early.
		if aggAllocatedCount <= 0 && capacity >= minCapacity {
			return 1, true, nil
		}
```

It's a bit convoluted, but it will get us the desired result. I _think_ this takes care of all the edge cases, but if you can think of any others let me know.

File: pkg/fleetautoscalers/fleetautoscalers.go [hunk 7]
```diff
@@ -486,6 +487,9 @@ func scaleDown(f *agonesv1.Fleet, gameServerLister listeragonesv1.GameServerList
 	// "Remove" one game server at a time in order of potential deletion. (Not actually removed here,
 	// that's done later, if possible, by the fleetautoscaler controller.)
 	for _, gs := range gameServers {
+		if replicas == 1 {
+			return 1, true, nil
+		}
```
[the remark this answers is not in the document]
Kalaiselvi84: I have rolled back all the changes from the `scaleDown` method and modified the code as per your suggestion. I adjusted the `TestApplyCounterPolicy/scale down` test case to not be limited: `limited: false`.

igooch (approved):
LGTM

PR #3751, opened 2024-04-10
Docs: Fold Sidebar-Menu

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #

This change will fold the sidebar menu of the Agones docs, following the instructions at https://www.docsy.dev/docs/adding-content/navigation/#section-menu-options. Please share your thoughts..

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -108,7 +108,7 @@ prism_syntax_highlighting = true
 # User interface configuration
 [params.ui]
 # Enable to show the side bar menu in its compact state.
-sidebar_menu_compact = false
+sidebar_menu_compact = true
 #  Set to true to disable breadcrumb navigation.
 breadcrumb_disable = false
 


markmandel:
So we tried this a while ago - and found it was a really bad experience for end users, since they couldn't find anything in the nav bar -- so we rolled it back. So I don't think we should be doing this.

Kalaiselvi84:
Thanks for letting me know about the previous attempt and the user experience. I'm okay to close this PR.

Kalaiselvi84:
@markmandel sorry, I am not fully convinced.. How about we merge this PR and see if the user experiences any unexpected behavior?

Kalaiselvi84:
Considering the recent upgrades to Hugo and Docsy, there is a possibility that the changes may function as expected??

markmandel:
See: https://github.com/googleforgames/agones/pull/2204

PR #3754, opened 2024-04-10
Remove unnecessary args from e2e-test-cloudbuild

zmerlynn:
This was an extension point I was going to push further into the e2es for tracking purposes, but it's inconsistent between the two builds in this file already.


igooch (approved):
LGTM

gongmax (approved):

PR #3756, opened 2024-04-10
Fix: Lint warning

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #

### DIFF
--- a/examples/gameserverallocation.yaml
+++ b/examples/gameserverallocation.yaml
@@ -85,34 +85,34 @@ spec:
       mode: deathmatch
     annotations:
       map: garden22
-  # [Stage: Alpha]
-  # [FeatureFlag:CountsAndLists]
-  # `Priorities` configuration alters the order in which `GameServers` are searched for matches to the configured `selectors`.
-  #
-  # Priority of sorting is in descending importance. I.e. The position 0 `priority` entry is checked first.
-  #
-  # For `Packed` strategy sorting, this priority list will be the tie-breaker within the least utilised infrastructure, to ensure optimal
-  # infrastructure usage while also allowing some custom prioritisation of `GameServers`.
-  #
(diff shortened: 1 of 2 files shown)


Kalaiselvi84:
I added rules explicitly to convert warnings into errors in the yamllint config file. Unfortunately, there isn't a built-in method to globally change all rules to error. You can find a list of available rules [here](https://yamllint.readthedocs.io/en/stable/rules.html). If there are any other rules you'd like me to include from the list, please let me know...

examples/gameserverallocation.yaml
```diff
@@ -85,34 +85,34 @@ spec:
       mode: deathmatch
     annotations:
       map: garden22
-  # [Stage: Alpha]
```
zmerlynn, line 88: This one doesn't look right to me. It's indenting it under .metadata.annotations (visually) but I don't think that's where the structure is. Can you disable yammlint for this block?

Kalaiselvi84: I have rolled back the modifications in this file and added it to the yamllint configuration to be ignored.

zmerlynn: I think it's better for pointwise suppressions like this to use comments to suppress it: https://yamllint.readthedocs.io/en/stable/disable_with_comments.html

Kalaiselvi84: The following options to disable yamllint are working fine, I chose the first one, please let me know what you think..

  # yamllint disable rule:comments-indentation rule:line-length
and 
  # yamllint disable

zmerlynn: LGTM

zmerlynn (approved):

Issue #3438, opened 2023-10-18
RollingUpdate `maxUnavailable` should ensure to allow new allocations


Ullaakut:
**What happened**:

The way the `maxUnavailable` parameter is currently implemented, as part of the RollingUpdate strategy, is counterintuitive and can lead to situations where:

* new allocations are made on outdated game server sets
* ready replicas in outdated game server sets are not reaped

**What you expected to happen**:

1. `maxUnavailable` should ensure to always be able to serve new allocations. The number of already allocated game servers should not influence this calculation, as they are not relevant to whether or not new allocations can be made.
2. New allocations should be prioritized on active game server sets over outdated ones.
3. Outdated game server sets should allow their ready replicas to be reaped to let the active game server set reach its desired replica count and finish the rolling update.

**How to reproduce it (as minimally and precisely as possible)**:

### Install Agones

 1. `kubectl create namespace agones-system`
 2. `kubectl apply --server-side -f https://raw.githubusercontent.com/googleforgames/agones/release-1.35.0/install/yaml/install.yaml`
 
### Create a Fleet

 1. `kubectl apply -f https://raw.githubusercontent.com/googleforgames/agones/release-1.35.0/examples/simple-game-server/fleet.yaml`
 4. Verify that the Fleet has 25% `maxSurge` and `maxUnavailable` by running `k get fleets.agones.dev/simple-game-server -o json | jq ".spec.strategy"`
     ```json
     {
       "rollingUpdate": {
         "maxSurge": "25%",
         "maxUnavailable": "25%"
       },
       "type": "RollingUpdate"
     }
     ```
 5. Set the number of replicas to `10`: `kubectl edit fleets.agones.dev/simple-game-server`
 6. Cause 3 allocations by running the following command: `kubectl create -f https://raw.githubusercontent.com/googleforgames/agones/release-1.35.0/examples/simple-game-server/gameserverallocation.yaml -o yaml`

### Trigger a RollingUpdate

 1. Run `watch kubectl get gameserversets` to make sure that the game servers are currently ready and not allocated.
     ```bash
     k get gameservers
     NAME                   SCHEDULING   DESIRED   CURRENT   ALLOCATED   READY   AGE
     simple-game-server-1   Packed       10        10        3           7       1m
     ```
 2. Run `kubectl edit fleets.agones.dev/simple-game-server` again, to change the game server version to 0.17 instead of 0.18.
[...]

PR #3762, opened 2024-04-15
fix: rollout strategy issues

### LINKED ISSUE #3688: MaxSurge and MinAvailable calculated incorrectly
**What happened**:

We see a very sharp decline in `Ready` GameServers when rolling over to a new GameServerSet, leading to having no `Ready` GameServers in the pool while the new version is stuck in `Scheduled` due to the nodes downloading images.

A smooth rollout with `Ready` GameServers always available.

When we investigated this, it seems there is a difference between what the docs say here: https://agones.dev/site/docs/guides/fleet-updates/#rolling-update-strategy

> By default, a Fleet will wait for new GameSevers to become Ready during a Rolling Update before continuing to shutdown additional GameServers, only counting GameServers that are Ready as being available when calculating the current maxUnavailable value which controls the rate at which GameServers are updated.

and what we see in the calculation for `MinAvailable`:

https://github.com/googleforgames/agones/blob/main/pkg/fleets/controller.go#L555

It also stands to reason the `MaxSurge` is calculated in relation to how `MinAvailable` is, and should probably not be based on `fleet.Spec.Replicas` either.

With some real world numbers from one of our clusters, we have:

Replicas: 3.46K
ReadyReplicas: 194
AllocatedReplicas: 3.22K
MaxSurge: 10%
MinAvailable: 10%

With the current calculation it is allowed to terminate 346 GameServers in the first round, but we only have 194 `Ready`, so all `Ready` GameServers are terminated. Were this based on `Ready` GameServers, it would be allowed to terminate 19, which is a rather large difference.

Another thing that struck me as odd in the calculation is `cleanupUnhealthyReplicas`:

https://github.com/googleforgames/agones/blob/main/pkg/fleets/controller.go#L529

This seems to consider everything that is not `Ready` as unhealthy but seems to ignore `Allocated` which strikes me as wrong.

It might also be a good idea to set `Replicas` to 0 on a non-active GameServerSet when `Allocated` == `Replicas`.

**Environment**:
- Agones version: 1.37.0
- Kubernetes version (use `kubectl version`): 1.27.3-gke.100
- Cloud provider or hardware configuration: GKE n2-custom-40-163840
- Others:

nrwiersma:
**What type of PR is this?**
>
> /kind breaking
/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

This PR fixes issues with the fleet rollout strategy that have been observed in production. It seems in clusters that are massively allocation heavy, rollout does not behave as expected or as documented. These patches have been confirmed to work in out production environment, using a custom build.

Each issue is in a separate commit, to allow them to be evaluated individually.

Closes #3688

Additional to the 2 bug fixes (first 2 commits), we observe that once an inactive GSS gets to the point of Allocated == Replicas, it is still possible for it to spawn GSes, if the new GS comes up fast enough and there is enough churn in the controller. Once this point is reached, there is no real reason for the GSS to have Spec Replicas, but rather it would be better for it to get its Spec Replicas set to 0 and be allowed to loose it allocations until it is cleaned up. This patch makes up the last commit.


pkg/fleets/controller.go
```diff
@@ -514,19 +514,21 @@ func (c *Controller) cleanupUnhealthyReplicas(ctx context.Context, rest []*agone
 	// Safely scale down all old GameServerSets with unhealthy replicas.
 	totalScaledDown := int32(0)
 	for i, gsSet := range rest {
+		healthy := gsSet.Status.ReadyReplicas + gsSet.Status.AllocatedReplicas + gsSet.Status.ReservedReplicas
```
markmandel, line 517: Thanks for digging into this so deeply!

I probably need to sit down and play with this to see it in effect, but I would definitely want this to sit behind a feature gate, so it can be widely tested without fear of breaking things.

To do that, my suggestion would be is duplicate the entirety of `cleanupUnhealthyReplicas` and `rollingUpdateRestFixedOnReady` with new names, e.g. "cleanupUnhealthyReplicasRollingUpdateFix` (or something similar).

(Maybe in a new file, if the controller is getting messy)

Then we could do something akin to the following:

```go
func (c *Controller) cleanupUnhealthyReplicas(ctx context.Context, rest []*agonesv1.GameServerSet,
	fleet *agonesv1.Fleet, maxCleanupCount int32) ([]*agonesv1.GameServerSet, int32, error) {

  if runtime.FeatureEnabled(runtime.FeatureRollingUpdateFix) {
    return cleanupUnhealthyReplicasRollingUpdateFix(ctx, rest, fleet, maxCleanupCount)
  }

// ....
}
```

And we could turn on/off feature flags and so we could have relavent unit tests as well.

How does that sound?

markmandel (commented):
I have some suggestions for how to feature gate this change.

Is there a e2e type scenario (akin to what we see in #3438) you can think of that would highlight the issues you were running into - that would also mean we don't run into it again in the future.

pkg/fleets/controller.go
nrwiersma: Sure, as you wish.

nrwiersma: I have managed to prove the primary tenants of the fix in unit tests, which is much simpler than e2e.

As a side note, `TestControllerRollingUpdateDeployment` seems to break the laws under with GSS Status Replicas exists, but I left the original math unchanged.

markmandel:
Thanks for making these changes - I'm probably going to opt to let this one slide past the next release on Tuesday, and give me (someone? :smile) some time to sit down with this, and make sure everything is good (which really means running some rolling updates and watching them happen 😄 )

nrwiersma:
We have production data of this patch if that would help.

markmandel:
> We have production data of this patch if that would help.

TBH - totally trust that you do. I just want to know that I've spent some time with this as an extra verification step, and give it time to back between releases as well.

markmandel:
Just a heads up - will take this for a spin this week 👍🏻

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -31,6 +31,7 @@ The current set of `alpha` and `beta` feature gates:
 | [Support for Extended Duration Pods on GKE Autopilot (*1.28+ only*)](https://github.com/googleforgames/agones/issues/3386)  | `GKEAutopilotExtendedDurationPods` | Disabled | `Alpha` | 1.37.0 |
 | [GameServer player capacity filtering on GameServerAllocations](https://github.com/googleforgames/agones/issues/1239)       | `PlayerAllocationFilter`           | Disabled | `Alpha` | 1.14.0 |
 | [Player Tracking]({{< ref "/docs/Guides/player-tracking.md" >}})                                                            | `PlayerTracking`                   | Disabled | `Alpha` | 1.6.0  |
+| [Rolling Update Fixes](https://github.com/googleforgames/agones/issues/3688)                                                | `RollingUpdateFix`                 | Disabled | `Alpha` | 1.41.0 |
```
markmandel, line 34: Small thing, can you make a copy of this table, one with the new feature and one without, and use the `feature` shortcodes to keep this feature flag hidden until the next release please.

See: https://agones.dev/site/docs/contribute/#within-a-page for details

markmandel (approved):
Finally took this for a spin, basically just wanted to see it in action - look great, let's go ahead with it!

(Rolling update on a Fleet of 50 gameservers)
Peek 2024-05-02 09-50

PR #3764, opened 2024-04-16
Feat(gameservers): Shared pod IPs with GameServer Addresses

### LINKED ISSUE #3763: Shared Pod IPs with GameServer Addresses
Sometimes, it's very useful to send traffic direct to the Pod address rather than through the node, for example if you are doing ipv6 without nat, so that traffic can go straight to the Pod within a network.

Append the [Pod IPs](https://github.com/googleforgames/agones/blob/45ea6670278945d4117ed399b9eb02442a4d3d2d/pkg/gameservers/gameservers.go#L88)[] array to the `GameServer.Status.Addresses[]`, under a key of `PodIP` (or similar).

Hacking Kube-cluster setup to get IPv6 in the right order.

**Additional context**

Should probably be implemented here:

https://github.com/googleforgames/agones/blob/45ea6670278945d4117ed399b9eb02442a4d3d2d/pkg/gameservers/gameservers.go#L88

We should also add some docs somewhere around:
https://agones.dev/site/docs/reference/gameserver/#primary-address-vs-addresses

lacroixthomas:
**What type of PR is this?**
/kind feature

Append the Pod IPs to the GameServer's addresses

Closes #3763 


markmandel:
Only thing I'd add is a test here:

https://github.com/googleforgames/agones/blob/b946cb3561453ab6fac0fa0238cb19bc2b3da0ab/test/e2e/gameserver_test.go#L62

Which checks if there's PodIp record(s), and that it's not empty (can also lookup the pod if we want and do a comparison?)

lacroixthomas:
> Only thing I'd add is a test here:
> 
> https://github.com/googleforgames/agones/blob/b946cb3561453ab6fac0fa0238cb19bc2b3da0ab/test/e2e/gameserver_test.go#L62
> 
> Which checks if there's PodIp record(s), and that it's not empty (can also lookup the pod if we want and do a comparison?)

Hello @markmandel 
I've added some check for the PodIPs on the addresses,
I'm not that familiar with the e2e of agones yet (about the framework / helpers etc), I'm gonna check this part to be able to do the lookup of the pod for the comparaison

daniellee:
Before I saw this PR, I had started working on it as well :-) Thanks for jumping on this so quickly - looking forward to seeing it merged soon. 

Testing this PR locally, it doesn't seem to populate the `gameserver.Status.Addresses` field.

I hacked this guard into the `syncGameServerStartingState` [function](https://github.com/googleforgames/agones/blob/a1af9afff25fcb8716cf2314ec541b6b27c2c947/pkg/gameservers/controller.go#L822) in `pkg/gameservers/controller.go`in to force it to wait until the pod.Status.PodIPs field has been populated:

```
if pod.Status.PodIPs == nil || len(pod.Status.PodIPs) == 0 {
	return gs, workerqueue.NewDebugError(errors.Errorf("pod IP not yet populated for Pod %s", pod.ObjectMeta.Name))
}
```

That's probably not the ideal way to solve it but it identifies that the problem is that the gameserver scheduled and ready states are set before the pod.Status.PodIPs field is populated and therefore the gameserver.Status.Addresses field never gets the PodIPs.

lacroixthomas:
> Before I saw this PR, I had started working on it as well :-) Thanks for jumping on this so quickly - looking forward to seeing it merged soon.
> 
> Testing this PR locally, it doesn't seem to populate the `gameserver.Status.Addresses` field.
> 
> I hacked this guard into the `syncGameServerStartingState` [function](https://github.com/googleforgames/agones/blob/a1af9afff25fcb8716cf2314ec541b6b27c2c947/pkg/gameservers/controller.go#L822) in `pkg/gameservers/controller.go`in to force it to wait until the pod.Status.PodIPs field has been populated:
> 
> ```
> if pod.Status.PodIPs == nil || len(pod.Status.PodIPs) == 0 {
> 	return gs, workerqueue.NewDebugError(errors.Errorf("pod IP not yet populated for Pod %s", pod.ObjectMeta.Name))
> }
> ```
> 
> That's probably not the ideal way to solve it but it identifies the problem. The gameserver's scheduled and ready states are set before the pod.Status.PodIPs field is populated and therefore the gameserver.Status.Addresses field never gets the PodIPs.

Hello, I didn't have time yesterday to investigate what happens, but nice catch, thanks for looking into it ! I'll update my MR with your fix :)

lacroixthomas:
I will add unit tests for the controller fix *

pkg/gameservers/gameservers.go
```diff
@@ -24,6 +24,11 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+const (
+	// NodePodIP identifies an IP address from a pod.
+	NodePodIP corev1.NodeAddressType = "PodIP"
```
zmerlynn, line 29: I think it's reasonable to put the const here: https://github.com/googleforgames/agones/blob/4bb673186b2ee8431de00c5c564bd1daa1a356df/pkg/apis/agones/v1/gameserver.go#L114-L150 .

test/e2e/gameserver_test.go
```diff
@@ -39,6 +39,7 @@ import (
 
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	allocationv1 "agones.dev/agones/pkg/apis/allocation/v1"
+	gameserversv1 "agones.dev/agones/pkg/gameservers"
```
zmerlynn, line 42: moving the const would mean it's in `agonesv1`, too, which seems less smelly.

zmerlynn (changes requested):
Thank you so much for the submission! This LGTM with a minor nit.

lacroixthomas:
> Thank you so much for the submission! This LGTM with a minor nit.

Thanks for the review ! I've updated the MR following your comments

site/content/en/docs/Reference/gameserver.md
```diff
@@ -190,20 +190,22 @@ Game Servers are created through Kubernetes API (either directly or through a [F
...
-and will always be one of the `addresses`. The `addresses` field contains every address in the [`Node.Status.addresses`][addresses],
+and will always be one of the `addresses`. The `addresses` field contains every address in the [`Node.Status.addresses`][addresses] and [`Pod.Status.podIPs`][podIPs] (to allow a direct pod access),
 representing all known ways to reach the `GameServer` over the network.
 
 To choose `address` from `addresses`, [Agones looks for the following address types][addressFunc], in highest to lowest priorty:
 * `ExternalDNS`
 * `ExternalIP`
 * `InternalDNS`
 * `InternalIP`
+* `PodIP`
```
zmerlynn, line 201: Since you didn't modify the `address` function in pkg/gameservers/gameservers.go, I don't think this line is true. However, to avoid a roundtrip across timezones and over a weekend, I may just go ahead and edit this, since the rest LGTM.

zmerlynn: (fixed in https://github.com/googleforgames/agones/pull/3764/commits/a671f1b91e511d13a5366f780a4175fbe081ed94)

markmandel (approved):
Looks great!

PR #3769, opened 2024-04-16
Update site documentation to reflect firewall prefix and default to Autopilot cluster creation for Agones

vicentefb:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
/kind documentation
> /kind feature
> /kind hotfix
> /kind release

Nit fix on site documentation so that the firewall created has a name with the prefix "gke-" so that the GCE enforcer doesn't delete it. Also, include a reference link to the firewall creation in each pre-reqs. On the other hand, switched the order of cluster creation so that it defaults to Autopilot cluster creation first. 

Closes #


site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -165,18 +165,45 @@ We need a firewall to allow UDP traffic to nodes tagged as `game-server` via por
...
-Create a GKE cluster in which you'll install Agones. You can use
+Create a GKE cluster (preferably an Autopilot mode cluster) in which you'll install Agones. You can use
 [GKE Standard mode](#create-a-standard-mode-cluster-for-agones)
 or [GKE Autopilot mode](#create-an-autopilot-mode-cluster-for-agones).
 
+### Create an Autopilot mode cluster for Agones
+
+{{<alert title="Note" color="info">}}
+These installation instructions apply to Agones 1.30+
+{{</alert>}}
```
zmerlynn, line 184: Not your fault but might as well handle it here: We had different instructions in 1.29 so the goal was to call them out. But it's been over a year - I think it's far enough past that we can remove this clause.

zmerlynn (changes requested):

site/layouts/shortcodes/gs-prerequisites.html
```diff
@@ -2,7 +2,7 @@
 {{- $test := eq (.Get "link_test") "true" | default true }}
 <p>The following prerequisites are required to create a GameServer:</p>
 <ol>
-<li>A Kubernetes cluster with the UDP port range 7000-8000 open on each node.</li>
+<li>A Kubernetes cluster with the UDP port range 7000-8000 open on each node. <a href="https://agones.dev/site/docs/installation/creating-cluster/gke/#creating-the-firewall">Creating the firewall</a></li>
```
markmandel, line 5: This should resolve to either the local or which snapshot version you are seeing - so it can't point at the domain directly.

You can use:
https://github.com/googleforgames/agones/blob/772fffbfaf8a3bdb0cfc1505b6ed51521f4e95d8/site/config.toml#L15

As the root of the link, via:

https://gohugo.io/templates/shortcode-templates/#:~:text=Params.shortcode_color).-,%24.Page.Site.Params,-refers%20to%20parameters

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -165,18 +165,41 @@ We need a firewall to allow UDP traffic to nodes tagged as `game-server` via por
 next section.
 
 ```bash
-gcloud compute firewall-rules create game-server-firewall \
+gcloud compute firewall-rules create gke-agones-game-server-firewall \
```
markmandel, line 168: 🤔  why this name change?

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -165,18 +165,41 @@ We need a firewall to allow UDP traffic to nodes tagged as `game-server` via por
...
   --allow udp:7000-8000 \
   --target-tags game-server \
   --description "Firewall to allow game server udp traffic"
 ```
 
 ## Creating the cluster
 
-Create a GKE cluster in which you'll install Agones. You can use
+Create a GKE cluster (preferably an Autopilot mode cluster) in which you'll install Agones. You can use
```
markmandel, line 176: Why is it preferable to use Autopilot?

There are tradeoffs, but why is it preferable?

This feels like we need a section on why choose one over the other? Or link to external docs?

site/content/en/docs/Installation/Creating Cluster/gke.md
zmerlynn: This is bleeding our internals a bit but it's a simple change that avoids our internal firewall enforcer.

site/content/en/docs/Installation/Creating Cluster/gke.md
vicentefb: Okok sg!

site/layouts/shortcodes/gs-prerequisites.html
vicentefb: Okok sg, i used /site/ now inside the href

site/content/en/docs/Installation/Creating Cluster/gke.md
zmerlynn: I think this is still using the same language?

vicentefb: Forgot to remove that, thanks!

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -165,12 +165,37 @@ We need a firewall to allow UDP traffic to nodes tagged as `game-server` via por
...
 
 ```bash
-gcloud compute firewall-rules create game-server-firewall \
+gcloud compute firewall-rules create gke-agones-game-server-firewall \
   --allow udp:7000-8000 \
   --target-tags game-server \
   --description "Firewall to allow game server udp traffic"
 ```
 
+### Create an Autopilot mode cluster for Agones
```
zmerlynn, line 174: You moved this up above "Creating a cluster", so now it appears as a subheading of "Creating the firewall". I think this block instead belongs at old line 180, swapping the two.
image

site/content/en/docs/Installation/Creating Cluster/gke.md
```diff
@@ -165,12 +165,37 @@ We need a firewall to allow UDP traffic to nodes tagged as `game-server` via por
...
+
+You can read more about choosing a [cluster mode above](#choosing-a-gke-cluster-mode).
```
zmerlynn, line 176: This line, however, belongs in the top-level "Creating the cluster" section, say after old line 179.

site/content/en/docs/Installation/Creating Cluster/gke.md
vicentefb: Ohh my bad, i just swapped them again

site/layouts/shortcodes/gs-prerequisites.html
zmerlynn: I think what @markmandel is saying is that even `baseURL` could be variable, so you need to use a function here to grab the config.

@markmandel any reason we can't use https://gohugo.io/methods/site/baseurl/ here? I think this just becomes `{{ .Site.BaseURL }}docs/installation/creating-cluster/gke/#creating-the-firewall`.

zmerlynn (commented):
Almost!

zmerlynn (approved):

PR #3781, opened 2024-04-18
Optimise GameServer Sub-Controller Queues

### LINKED ISSUE #3748: GameServer Sub-Controllers queue more than they should
**What happened**:

Looking at pprof profiles of the Agones Controller, we see a lot memory in the Migration controller and the Missing Controller queues.

Way less items queued - since they are setup to handle edge cases for Pods and GameServers,

Reviewing the code in:
* https://github.com/googleforgames/agones/blob/main/pkg/gameservers/migration.go
* https://github.com/googleforgames/agones/blob/main/pkg/gameservers/missing.go
* (maybe?) https://github.com/googleforgames/agones/blob/main/pkg/gameservers/health.go

We could do more checks at the time of the add/update/event, and only queue if the conditions are met, or if an error happens. That way far less items will be queued, and the Agones controller should have a much smaller memory footprint.

Right now, almost all Pods are queued -- which is way too much.

Run a load test with several thousand `GameServers`.

**Environment**:
- Agones version: dev
- Kubernetes version (use `kubectl version`): 1.27
- Others:

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Looking at pprof profiles of the Agones Controller, we see a lot memory in the Migration controller and the Missing Controller queues.

This implements improved checking at the event layer, before placing items in the workerqueue, which drops the memory footprint, and also allows for faster processing of the edge cases that Missing and Migration Controllers implement.

Also did a review of the Health Controller and made a small improvement there as well.

Looking at the memory pprof profile and load tests, we see a decrease in memory usage, and also less of a slow "step up" of memory over time.

Closes #3748

Here's some pretty pictures!

Here's the two tests run side by side - I only did 30m, but you can see a drop in memory, and less of a "step up" in memory usage you slowly see the beginnings of with the original code
two-tests

Also updated pprof, it's all basically JSON de/serialisation, which is pretty much what we would want to see.

pprof-flame-fixed

### DIFF
--- a/pkg/gameservers/migration.go
+++ b/pkg/gameservers/migration.go
@@ -94,13 +94,13 @@ func NewMigrationController(health healthcheck.Handler,
 	_, _ = podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			pod := obj.(*corev1.Pod)
-			if isActiveGameServerWithNode(pod) {
+			if _, _, ok, err := mc.isMigratingGameServerPod(pod); err != nil || ok {
 				mc.workerqueue.Enqueue(pod)
 			}
 		},
 		UpdateFunc: func(_, newObj interface{}) {
 			pod := newObj.(*corev1.Pod)
-			if isActiveGameServerWithNode(pod) {
+			if _, _, ok, err := mc.isMigratingGameServerPod(pod); err != nil || ok {
 				mc.workerqueue.Enqueue(pod)
 			}
 		},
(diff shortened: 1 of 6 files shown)


pkg/gameservers/migration.go
```diff
@@ -132,86 +132,109 @@ func (mc *MigrationController) loggerForGameServer(gs *agonesv1.GameServer) *log
 	return mc.loggerForGameServerKey(gsName).WithField("gs", gs)
 }
 
-// isActiveGameServerWithNode tests to see if a Pod is active and has a Node assigned.
-func isActiveGameServerWithNode(pod *corev1.Pod) bool {
-	return pod.Spec.NodeName != "" && pod.ObjectMeta.DeletionTimestamp.IsZero() && isGameServerPod(pod)
-}
+func (mc *MigrationController) isMigratingGameServerPod(pod *k8sv1.Pod) (*agonesv1.GameServer, *k8sv1.Node, bool, error) {
+	mc.baseLogger.Info("RUNNING isMigratingGameServerPod")
```
zmerlynn, line 136: .Debug, or delete maybe? At the very lease maybe log the pod name? (We have a lot of one-liner useless logs in the control plane already :/)

zmerlynn (approved):
total nit on a log line but o/w LGTM

pkg/gameservers/migration.go
markmandel: Ah thanks - good catch, I'll make that change.

markmandel:
Aaah, didn't have enough time to become leader!

```
23: e2e-wait-to-become-leader
```

pkg/gameservers/migration.go
zmerlynn: Thanks! LGTM

PR #3782, opened 2024-04-19
Configure Allocator Status Code

### LINKED ISSUE #3780: Configurable Allocator HTTP status codes on failure

Right now, the Allocator HTTP Service returns a HTTP `429: Too Many Requests` if there is no `GameServer` available:

https://github.com/googleforgames/agones/blob/4bb673186b2ee8431de00c5c564bd1daa1a356df/pkg/allocation/converters/converter.go#L484-L494

Which is quite reasonable!

I'm digging into various service meshes, and they each like different HTTP codes for things like circuit breaking and/or retries or rate limiting.

For example, Istio will not circuit break on 429 ([closed issue](https://github.com/istio/istio/issues/41105)), [only on 5xx errors](https://istio.io/latest/docs/tasks/traffic-management/circuit-breaking/#configuring-the-circuit-breaker).

As part of the helm config, allow the user to specify which HTTP code they would like on a non-allocation through the HTTP service. Then we are future proofed for any Service Mesh that is utilised.

Leave things alone, but that will make things harder for any service mesh people use.

There is https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRetry - and could redirect to another cluster on failure, but that means each cluster gets pinged once, rather than dropping the request to the full cluster for a time period.

**Additional context**

N/A

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #3780


zmerlynn:
Can we come up with an approach that doesn't involve flag interpretation in a base library?

Kalaiselvi84:
> Can we come up with an approach that doesn't involve flag interpretation in a base library? Flags should be plumbed down from the `cmd/` layer.

 I have moved the flag to `cmd/main.go` and updated the `convertStateV1ToError` in `converter.go` to use error code based on httpUnallocatedStatusCode and added a test in `converter_test.go`. Please review them and share your feedback..

zmerlynn:
> > Can we come up with an approach that doesn't involve flag interpretation in a base library? Flags should be plumbed down from the `cmd/` layer.
> 
> I have moved the flag to `cmd/main.go` and updated the `convertStateV1ToError` in `converter.go` to use error code based on httpUnallocatedStatusCode and added a test in `converter_test.go`. Please review them and share your feedback..

It doesn't look like the flag was moved to `cmd/main.go` - rather, it looks like the flag was added there, and you added it to the config, but didn't do anything with that config? Your goal is to get to a place where there is no `viper` import in pkg/allocation at all - if you look, there's only a couple of places we do that, and in both cases we only import viper to provide a function that can be used in cmd/main.go for the respective binary.

I wish I could find a good citation for this, but the general idea here is that flags in libraries make testing a pain because they're "spooky action at a distance". Flags should go into config structures that get plumbed down.

But, that said, is there a reason not to do this directly here? https://github.com/googleforgames/agones/blob/53a70b357a08c6ee1ddd70288a916e9941ebc344/cmd/allocator/main.go#L624

You can just look for `ResourceExhausted` and translate it to `Unavailable` right here instead.

Kalaiselvi84:
> Flags should go into config structures that get plumbed down.
> 
> But, that said, is there a reason not to do this directly here?
> 
> https://github.com/googleforgames/agones/blob/53a70b357a08c6ee1ddd70288a916e9941ebc344/cmd/allocator/main.go#L624

Along with adding a new flag in the config structures, I've also integrated it into the `newServiceHandler()` function and within the `Allocate()` method of the serviceHandler, this  statusCode is utilized when calling `ConvertGSAToAllocationResponse()` method and modified the `converter.go` file to capture this change. Please let me know if I am in the right direction and share your feedback..

pkg/allocation/converters/converter_test.go
```diff
@@ -1533,3 +1535,37 @@ func TestConvertAllocationResponseToGSA(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertStateV1ToError(t *testing.T) {
+	viper.Set("HTTP_UNALLOCATED_STATUS_CODE", "429")
```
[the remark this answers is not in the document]
Kalaiselvi84: I have updated the tests `TestConvertStateV1ToError` and `TestConvertGSAToAllocationResponse` to align with the modifications. These tests are now passing locally. Could you please review them and provide your feedback?

cmd/allocator/main.go
```diff
@@ -244,7 +251,7 @@ func main() {
 		os.Exit(0)
 	})
 
-	h := newServiceHandler(ctx, kubeClient, agonesClient, health, conf.MTLSDisabled, conf.TLSDisabled, conf.remoteAllocationTimeout, conf.totalRemoteAllocationTimeout, conf.allocationBatchWaitTime)
```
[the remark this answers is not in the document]
Kalaiselvi84: I have used `codes.Code` instead of `http` following your suggestion. [However, in the latest commit, I've commented out the `http` code usage, here I just tried to use the `http` code and converting it into gRPC using the `GrpcCodeFromHTTPStatus` converter method.]
I am looking forward to hearing your thoughts...

pkg/allocation/converters/converter.go
```diff
@@ -303,12 +304,12 @@ func convertGameServerSelectorsToInternalGameServerSelectors(in []*pb.GameServer
 }
 
 // ConvertGSAToAllocationResponse converts GameServerAllocation V1 (GSA) to AllocationResponse
-func ConvertGSAToAllocationResponse(in *allocationv1.GameServerAllocation) (*pb.AllocationResponse, error) {
+func ConvertGSAToAllocationResponse(in *allocationv1.GameServerAllocation, grpcStatusCode codes.Code) (*pb.AllocationResponse, error) {
```
zmerlynn, line 307: It's lengthy but I prefer `grpcUnallocatedStatusCode` here as a name so it's clear what this code is for.

pkg/allocation/converters/converter_test.go
```diff
@@ -846,6 +847,7 @@ func TestConvertGSAToAllocationResponse(t *testing.T) {
 		name             string
 		features         string
 		in               *allocationv1.GameServerAllocation
+		grpcStatusCode   codes.Code
```
zmerlynn, line 850: Again, so this is clearer, I would call it `grpcUnallocatedStatusCode`

pkg/allocation/converters/converter_test.go
```diff
@@ -873,6 +875,7 @@ func TestConvertGSAToAllocationResponse(t *testing.T) {
 					Source:   "local",
 				},
 			},
+			grpcStatusCode: codes.OK,
```
zmerlynn, line 878: Again, this would never be `codes.OK` - it doesn't really make sense as an error return. What I would, though, is just omit this field and let it be 0 (which is incidentally the same as `codes.OK`, and then below add defaulting logic.

pkg/allocation/converters/converter_test.go
```diff
@@ -1341,7 +1354,7 @@ func TestConvertGSAToAllocationResponse(t *testing.T) {
 			defer runtime.FeatureTestMutex.Unlock()
 			require.NoError(t, runtime.ParseFeatures(tc.features))
```
zmerlynn, line 1356: I would just default here so the existing tests work (and add one more test that something other than the default also works). So like:

```
grpcUnallocatedStatusCode := tc.grpcUnallocatedStatusCode
if grpcUnallocatedStatusCode == codes.OK {
  grpcUnallocatedStatusCode = codes.ResourceExhausted
}

out, err := Convert(tc.in, grpcUnallocatedStatusCode)
```

cmd/allocator/main.go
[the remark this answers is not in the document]
zmerlynn: I am missing the `GrpcCodeFromHTTPStatus` call?

This looks mostly like what I meant, from about here down. I think you were initially correct that the config surface (the flag, the helm variable) should be in http status, though, since most of the use of changing this is actually on the http side. So in this file (allocator/main.go), I would take a flag for unallocated http status and convert it into a codes.Code. Then the rest looks reasonable.

pkg/allocation/converters/converter.go
Kalaiselvi84: I've renamed it to grpcUnallocatedStatusCode

pkg/allocation/converters/converter_test.go
```diff
@@ -910,8 +912,9 @@ func TestConvertGSAToAllocationResponse(t *testing.T) {
 					NodeName: "node-name",
 				},
 			},
-			wantErrCode:      codes.ResourceExhausted,
-			skipConvertToGSA: true,
+			grpcUnallocatedStatusCode: codes.ResourceExhausted,
+			wantErrCode:               codes.ResourceExhausted,
+			skipConvertToGSA:          true,
 		},
```
zmerlynn, line 918: Add another table entry here that tests the non-default case:

```
		{
			name: "status field is set to unallocated, non-default unallocated",
			in: &allocationv1.GameServerAllocation{
				TypeMeta: metav1.TypeMeta{
					Kind:       "GameServerAllocation",
					APIVersion: "allocation.agones.dev/v1",
				},
				Status: allocationv1.GameServerAllocationStatus{
					State:          allocationv1.GameServerAllocationUnAllocated,
					GameServerName: "GSN",
					Ports: []agonesv1.GameServerStatusPort{
						{
							Port: 123,
						},
						{
							Name: "port-name",
						},
					},
					Address:  "address",
					NodeName: "node-name",
				},
			},
			grpcUnallocatedStatusCode: codes.Unimplemented,
			wantErrCode:               codes.Unimplemented,
			skipConvertToGSA:          true,
		},
```

This is basically just a permutation of the test right above, but with a new unallocated status. I chose `codes.Unimplemented` because it's different than existing codes we might return for other reasons, making it easier to see, if this test were to fail.

pkg/allocation/converters/converter_test.go
```diff
@@ -1533,3 +1543,51 @@ func TestConvertAllocationResponseToGSA(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertStateV1ToError(t *testing.T) {
```
zmerlynn, line 1547: Oh, I just saw this. Hmm. Honestly if you make the changes above, I don't think you really need this test - the errors are already checked above, and you can add a simple permutation of grpcUnallocatedStatusCode to check your new logic.

zmerlynn (commented):
Very close!

zmerlynn (approved):

pkg/allocation/converters/converter_test.go
Kalaiselvi84: I was trying with the following test case, it wasn't working for me.🤔

{
    name: "Test non-default error code",
    in: &allocationv1.GameServerAllocation{
        TypeMeta: metav1.TypeMeta{
            Kind:       "GameServerAllocation",
            APIVersion: "allocation.agones.dev/v1",
        },
        Status: allocationv1.GameServerAllocationStatus{
            State:          allocationv1.GameServerAllocationUnAllocated,
            GameServerName: "RestrictedAccessGSN",
            Ports: []agonesv1.GameServerStatusPort{
                {
                    Port: 456,
                    Name: "restricted-port-name",
                },
            },
            Address:  "restricted-address",
            NodeName: "restricted-node-name",
        },
    },
    grpcUnallocatedStatusCode: codes.PermissionDenied,
    want: &pb.AllocationResponse{
        GameServerName: "RestrictedAccessGSN",
        Address:        "restricted-address",
        NodeName:       "restricted-node-name",
        Ports: []*pb.AllocationResponse_GameServerStatusPort{
            {
                Port: 456,
                Name: "restricted-port-name",
            },
        },
        Metadata: &pb.AllocationResponse_GameServerMetadata{
            Labels: map[string]string{
                "error": "permission-denied",
            },
        },
    },
    wantErrCode: codes.PermissionDenied,
},

Issue #3767, opened 2024-04-16
In-place Agones Upgrades: SDK Compatability


zmerlynn:
> [!NOTE]
> Milestone of #3766, which we are seeking feedback on. We will move forward with pieces that seem non-contentious, though.

# SDK Compatibility
This milestone satisfies the requirements in "SDK proto" and "Client SDKs" of #3766  - in particular, see the "SDK Compatibility Contract" mentioned there.

As of now, we have APIs in Stable and Alpha, but no Beta SDK protos - in fact, there is no Beta surface in the Client SDKs because Player Tracking, the first Alpha Client SDK feature, never graduated off Alpha. The next Client SDK feature to graduate is Counts & Lists - at which time, we will need to create a Beta Client SDK surface across the SDKs. This will be handled as part of Counts & Lists graduation, not as part of this project.

Because of this, there is no need to do any coding work towards Client SDK Compatibility if we accept the requirements in #3766.

## Use `Patch` vs `Update` <a href="#update-vs-patch" id="update-vs-patch">#</a>
However, there is still a gotcha in how the `sdk-server` works today: We need to improve how `sdk-server` handles fields that are known by the remainder of the Agones Control Plane but are unknown by the `sdk-server`, which may happen in the event of newly lit feature gate paths. In particular, updates like this:

https://github.com/googleforgames/agones/blob/05d8232c937d756d9b8e848915e04e4159924b7b/pkg/sdkserver/sdkserver.go#L472

are likely to wipe out new API fields. We can use `Patch()` instead, by using optimistic locking in patch similar to [contoller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/c25fe2f5ca0f9e10d571f85da61c3d210d6345e7/pkg/client/patch.go#L117-L130).

---

To close this issue requires:

- [x] Documentation:
   * Document SDK proto compatibility guarantees and deprecation policies.
   * Keep a chart of each API and its stability level and different versions so users can easily reference to see if there are changes. The chart can be started as of Agones 1.40 (we don't need to back date it, we haven't provided a guarantee in the past.)
- [x] Move `sdk-server` to `Patch()` instead of `Update()`

zmerlynn:
I think so!

PR #3803, opened 2024-04-24
Changes `sdk-server` to Patch instead of Update

igooch:
**What type of PR is this?**

/kind feature

See #3767#update-vs-patch

Completes part II of #3767

I did not change PlayerTracking as the feature will be deprecated.

### DIFF
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -456,7 +456,7 @@ func (s *SDKServer) patchGameServer(ctx context.Context, gs, gsCopy *agonesv1.Ga
 	if err != nil {
 		return nil, err
 	}
-	fmt.Println("PATCH", string(patch))
+
 	return s.gameServerGetter.GameServers(s.namespace).Patch(ctx, gs.GetObjectMeta().GetName(), types.JSONPatchType, patch, metav1.PatchOptions{})
 }
 
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -29,6 +29,7 @@ import (
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -452,7 +453,7 @@ func (s *SDKServer) gameServer() (*agonesv1.GameServer, error) {
 // updateLabels updates the labels on this GameServer to the ones persisted in SDKServer,
 // i.e. SDKServer.gsLabels, with the prefix of "agones.dev/sdk-"
 func (s *SDKServer) updateLabels(ctx context.Context) error {
-	s.logger.WithField("labels", s.gsLabels).Debug("Updating label")
+	s.logger.WithField("labels", s.gsLabels).Debug("Patching label")
 	gs, err := s.gameServer()
 	if err != nil {
 		return err
@@ -469,7 +470,12 @@ func (s *SDKServer) updateLabels(ctx context.Context) error {
 	}
 	s.gsUpdateMutex.RUnlock()
 
-	_, err = s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})
+	patch, err := gs.Patch(gsCopy)
+	if err != nil {
+		return err
+	}
+
+	_, err = s.gameServerGetter.GameServers(s.namespace).Patch(ctx, gs.GetObjectMeta().GetName(), types.JSONPatchType, patch, metav1.PatchOptions{})
 	return err
 }
 
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -979,7 +979,12 @@ func (s *SDKServer) updateCounter(ctx context.Context) error {
 		names = append(names, name)
 	}
 
-	gs, err = s.gameServerGetter.GameServers(s.namespace).Update(ctx, gsCopy, metav1.UpdateOptions{})
+	patch, err := gs.Patch(gsCopy)
+	if err != nil {
+		return err
+	}
+
+	_, err = s.gameServerGetter.GameServers(s.namespace).Patch(ctx, gs.GetObjectMeta().GetName(), types.JSONPatchType, patch, metav1.PatchOptions{})
 	if err != nil {
 		return err
 	}
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -356,7 +356,6 @@ func (s *SDKServer) updateState(ctx context.Context) error {
 	}
 	s.gsUpdateMutex.RUnlock()
 
-	gameServers := s.gameServerGetter.GameServers(s.namespace)
 	gs, err := s.gameServer()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


pkg/sdkserver/sdkserver_test.go
```diff
@@ -44,6 +46,24 @@ import (
 	testclocks "k8s.io/utils/clock/testing"
 )
 
+// PatchGameServer is a helper function for the AddReactor "patch" that creates and applies a patch
+// to a gameserver. Returns a patched copy and does not modify the original game server.
+func PatchGameServer(t *testing.T, action k8stesting.Action, gs *agonesv1.GameServer) *agonesv1.GameServer {
```
zmerlynn, line 51: nit: un-export this, it's just a helper for the test

zmerlynn (approved):

zmerlynn:
cc @markmandel (but I'm happy to proceed with this, we have plenty of time to CI)

pkg/sdkserver/sdkserver.go
```diff
@@ -449,6 +449,17 @@ func (s *SDKServer) gameServer() (*agonesv1.GameServer, error) {
...
 }
 
+// patchGameServer is a helper function to create and apply a patch update, so the changes in
+// gsCopy are applied to the original gs.
+func (s *SDKServer) patchGameServer(ctx context.Context, gs, gsCopy *agonesv1.GameServer) (*agonesv1.GameServer, error) {
+	patch, err := gs.Patch(gsCopy)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Println("PATCH", string(patch))
```
markmandel, line 459:
```suggestion
```

Shouldn't leave this in.

pkg/sdkserver/sdkserver.go
```diff
@@ -449,6 +449,17 @@ func (s *SDKServer) gameServer() (*agonesv1.GameServer, error) {
 	return gs, nil
 }
 
+// patchGameServer is a helper function to create and apply a patch update, so the changes in
+// gsCopy are applied to the original gs.
+func (s *SDKServer) patchGameServer(ctx context.Context, gs, gsCopy *agonesv1.GameServer) (*agonesv1.GameServer, error) {
+	patch, err := gs.Patch(gsCopy)
```
markmandel, line 455: Does this `gs.Patch(...)` do the optimistic locking?

igooch: It does not. Client-go's .Patch (which the `s.gameServerGetter.GameServers(...).Patch(...)` is generated from) doesn't have the option as part of the PatchOptions, so it needs to be passed as part of the patch. There is an MergeFromWithOptimisticLock option in the https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#MergeFromWithOptions (Strategic Merge doesn't support CRDs)

```
patch := client.MergeFromWithOptions(gsCopy, client.MergeFromWithOptimisticLock{})
patchData, err := patch.Data(gs)
```

This is the type `"application/merge-patch+json"`, which I believe we decided not use, and to instead use type `"application/json-patch+json"`. I also haven't tested this yet in conjunction with Client-go's .Patch. The only examples I've seen use sigs.k8s's client to patch.

The other option would be to include the Resource Version as a "test" operation `jsonpatch.JsonPatchOperation{{Operation: "test", Path: "/metadata/resourceVersion", Value: gs.ObjectMeta.ResourceVersion}}` as a part of the list of patch operations in `gs.Patch(...)`. Per https://jsonpatch.com/ "Tests that the specified value is set in the document. If the test fails, then the patch as a whole should not apply." This also means that if the ResourceVersion is not set the Patch would fail. Given that the Game Server should always have the ResourceVesion set I'm leaning towards this option.

Do you have any thoughts on which one we should use, or if there's another better option?

zmerlynn: The issue actually discussed this - I think we can use the same way controller-manager does: https://github.com/kubernetes-sigs/controller-runtime/blob/c25fe2f5ca0f9e10d571f85da61c3d210d6345e7/pkg/client/patch.go#L110-L114, which just ensures the resourceVersion makes it into the patch, AIUI.

markmandel: > The issue actually discussed this - I think we can use the same way controller-manager does: https://github.com/kubernetes-sigs/controller-runtime/blob/c25fe2f5ca0f9e10d571f85da61c3d210d6345e7/pkg/client/patch.go#L110-L114, which just ensures the resourceVersion makes it into the patch, AIUI.

I'm relatively sure @igooch is right on this one - that example is a mergePatch, whereas we are using a JSONPatch - so it's different.

zmerlynn: > extend gs.Patch(gs *GameServer, ...options) to have a set of options, with the only one right now being WithOptimisticLocking (like the runtime-controller).

I think we can start without it being optional. We're moving from optimistically-locked Update, I'd rather just start with optimistically-locked Patch and gradually claw back.

markmandel: Huh, I swore we used that function in a bunch of places, but apparently not!

Yep - no need for options!

zmerlynn: And @igooch you're totally right, great work here! Looks good, reviewing the whole again now.

markmandel:
Checking different GKE versions, I'm seeing all kinds of weird e2e tests - something has gone unfortunately skewiff.

For example: https://console.cloud.google.com/cloud-build/builds/8456e5b8-41a1-4f9d-806a-5376be6f9570?project=258182270954

I see a bunch of counter issues:

```
VERBOSE: === RUN   TestCounters/IncrementCounter
VERBOSE: time="2024-05-06 18:13:00.950" level=info msg=IncrementCounter fields.msg="INCREMENT_COUNTER foo 10"
VERBOSE: time="2024-05-06 18:13:00.967" level=info msg="Sending GetCounterCount" fields.msg="GET_COUNTER_COUNT foo"
VERBOSE:     gameserver_test.go:1481: 
VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:1481
VERBOSE:         	Error:      	Not equal: 
VERBOSE:         	            	expected: "COUNTER: 20\n"
VERBOSE:         	            	actual  : "COUNTER: 10\n"
VERBOSE:         	            	
VERBOSE:         	            	Diff:
VERBOSE:         	            	--- Expected
VERBOSE:         	            	+++ Actual
VERBOSE:         	            	@@ -1,2 +1,2 @@
VERBOSE:         	            	-COUNTER: 20
VERBOSE:         	            	+COUNTER: 10
VERBOSE:         	            	 
VERBOSE:         	Test:       	TestCounters/IncrementCounter
VERBOSE: --- FAIL: TestCounters/IncrementCounter (0.04s)
VERBOSE: --- FAIL: TestCounters (3.25s)
```

igooch:
> Checking different GKE versions, I'm seeing all kinds of weird e2e tests - something has gone unfortunately skewiff.

That's what I'm working on debugging now. From what logs I've been able to glean it looks like the patch is not going through due to the object resource version having changed (WAI). I'm attempting to figure out if it's failing because it's returning the original error before retry, failing after multiple retries, or something else. Having a hard time debugging since I don't have any easy way of getting the container logs during test since only about one out of twenty game servers fail and there's so much pod churn. I.e. `kubectl logs -f -l agones.dev/role=gameserver -c agones-gameserver-sidecar -n 123456` will only get for the currently running pods, and won't watch / stream for newly created pods.

If you know how to create a kubectl command to watch and stream logs for newly created pods in a namespace, please let me know. I haven't been able to get kubectl get pods watch + xargs kubectl logs to work.

markmandel:
If you can output to the test which GameServers fail, you should be able to use Cloud Console logging to see all the logs for that Pod and it's sidecars.

markmandel (approved):
Looks great!

PR #3807, opened 2024-04-26
feat(gameserver): New DirectToGameServer PortPolicy allows direct traffic to a GameServer

### LINKED ISSUE #3804: Direct connection to a GameServer/Pod without NAT

We have a use case where we want to send traffic directly (via Quilkin) to gameserver instances. 

On a single node, we want to have multiple gameservers/pods and be able to send traffic to them via their publicly routable IP address (ipv4 or ipv6) where each gameserver has the same fixed ContainerPort (7777 is the default for Unreal) and without port forwarding or NAT.

A previous [issue](#3763) and [PR](#3765) added the pod IP addresses to the Gameserver Addresses but it is still not possible to directly communicate with a Gameserver instance on Agones. The three port policies (Dynamic, Passthrough and Static) all assume that the HostPort is used in some way and [GameServer.Status.Port](https://github.com/googleforgames/agones/blob/main/pkg/apis/agones/v1/gameserver.go#L842) always shows the HostPort.

To directly connect to a Gameserver instance with a publicy routable address and a fixed port without any NAT, the first item in Gameserver.Status.Ports should be the ContainerPort. Ideally no HostPorts would be set as they are not needed. A new PortPolicy called `DirectToPod`, `NoNAT` or similar would be one way to configure this.

The places in the code that would need to change:
- [Status function](https://github.com/googleforgames/agones/blob/main/pkg/apis/agones/v1/gameserver.go#L842)
- [Pod function](https://github.com/googleforgames/agones/blob/main/pkg/apis/agones/v1/gameserver.go#L716)

We have tried configuring Agones a few different ways to get around this issue.

**No PortPolicy and no ports configured**. This is the closest we got to get it working. No ports are defined in the Gameserver.Spec or the Gameserver.Status. But if you know that port 7777 is exposed then you can access the gameserver instance directly. For us to get this to work with Quilkin, Quilkin would have to default/fall back to port 7777 as Agones doesn't know anything about what ContainerPort is set. Also, that it is possible to work around the Port Policy this way seems like an undocumented, non-obvious feature.

**Additional context**

Here is an example of how you cannot get around HostPort if a PortPolicy is set. 

In the Gameserver Spec:
- Port Policy is set to Dynamic
- ContainerPort is set to 7654
- HostPort was not defined

Output from running `kubectl describe gameservers`:
```
Spec:
  Container:  simple-game-server
...
  Ports:
    Container:       simple-game-server
    Container Port:  7654
    Host Port:       7133
    Name:            default
    Port Policy:     Dynamic
    Protocol:        UDP
```

The Gameserver Status - returns the HostPort and not the ContainerPort.
```
Status:
  Address:  xxx
  Addresses:
    Address: xxx
    Type:     InternalIP
    Address:  xxx
    Type:     Hostname
    Address:  xxx
    Type:     PodIP
    Address:  xxx
    Type:     PodIP
...
  Ports:
    Name:          default
    Port:          7133
```

We tested with the PortPolicy options but none of them worked for us:
1. Dynamic PortPolicy with a ContainerPort configured. Agones sets a HostPort and that is the port that is set in GameServer.Status.Ports - and that breaks the communication with Quilkin.
2. Static PortPolicy with same port defined for HostPort and ContainerPort. Agones spins up one Gameserver instance/pod per node and is not possible to allocate more pods on that port per node.
3. Passthrough PortPolicy. Sets a different port for every gameserver instance so would have to do some sort of mapping to get this to work with Quilkin and our games.

daniellee:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix
> /kind release

- Directly connect to a Gameserver instance with a publicly routable address and a fixed port without any NAT
- Removes the HostPort dependency of the other three port policies and allows connecting directly using the PodIP and the ContainerPort

Closes #3804

- Gofmt made a lot of formatting changes in the test files. Let me know if I should roll those back.
- I don't know if you will accept a PR for another PortPolicy - let me know if you think there is a better way to approach this. I didn't come up with any good alternatives - see issue for details.
- The e2e test was the hardest part of this PR. There is no easy way to set up networking for a minikube test to allow direct traffic to a pod via its IP address and port. We are using [Calico](https://docs.tigera.io/calico/latest/reference/resources/ippool) and BGP to get this to work. I could add port-forwarding but then I am not testing anything relevant. Open to suggestions if there is a relevant e2e test that should be added to this PR.


daniellee:
I got stuck on the e2e test as setting up the pod so that you can send direct traffic to via a Pod IP and container port is not simple. Working on getting a decent e2e test in place.

test/e2e/gameserver_test.go
```diff
@@ -852,6 +848,28 @@ func TestGameServerPassthroughPort(t *testing.T) {
 	assert.Equal(t, "ACK: Hello World !\n", reply)
 }
 
+func TestGameServerPortPolicyNone(t *testing.T) {
+	framework.SkipOnCloudProduct(t, "gke-autopilot", "does not support None PortPolicy")
```
zmerlynn, line 852: That's not true, or implemented, AFAICT. :) (It looks like Autopilot should support this just fine, but separately, if you intended it not to be the case, you would need to modify the validation webhook.)

daniellee: Should I also change the [error message](https://github.com/daniellee/agones/blob/gameservers-pod-ip/pkg/cloudproduct/gke/gke.go#L42) "portPolicy must be Dynamic on GKE Autopilot" to something like "portPolicy must be Dynamic or None on GKE Autopilot"? Does that make sense?

zmerlynn: That's a good idea, yeah!

daniellee:
@zmerlynn let me know if there is anything else I need to change. In the meantime, I will try and keep the branch synchronized.

site/content/en/docs/Reference/gameserver.md
```diff
@@ -245,10 +246,11 @@ The `spec` field is the actual GameServer specification and it is composed as fo
 - `container` is the name of container running the GameServer in case you have more than one container defined in the [pod](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/). If you do,  this is a mandatory field. For instance this is useful if you want to run a sidecar to ship logs.
 - `ports` are an array of ports that can be exposed as direct connections to the game server container
   - `name` is an optional descriptive name for a port
-  - `portPolicy` has three options:
+  - `portPolicy` has four options:
```
markmandel, line 249: These changes will need to live behind a `feature` shortcode please (our docs publish immediately on merge).

See: https://agones.dev/site/docs/contribute/ for details.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -82,6 +82,8 @@ const (
 	// Passthrough dynamically sets the `containerPort` to the same value as the dynamically selected hostPort.
 	// This will mean that users will need to lookup what port has been opened through the server side SDK.
 	Passthrough PortPolicy = "Passthrough"
+	// None means the `hostPort` is ignored and if defined, the `containerPort` (optional) is used to set the port on the GameServer instance.
+	None PortPolicy = "None"
```
markmandel, line 86: Since we're in here, shall we also update:

https://github.com/googleforgames/agones/blob/981f79abce919cdb5496354cef934892f63bf1f0/pkg/apis/agones/v1/gameserver.go#L277-L281 ?

Will also need to run `make gen-api-docs gen-crd-client` to ensure the comment changes pass all the way through.

markmandel (commented):
Thanks for all this work, and the discussions!

No issues with implementation here 👍🏻 a couple of notes on some docs.

Only thing I have -let's put it behind a feature gate, just so we have an opportunity to get feedback on the feature once it has landed, and can change it if need be - you may well find things you want to tweak once you've got it up and running.

I expect it'll promote pretty fast, but better safe than sorry.

Details on feature flags here: 
* https://agones.dev/site/docs/guides/feature-stages/
* https://github.com/googleforgames/agones/blob/main/pkg/util/runtime/features.go (comments have implementation steps)

daniellee:
Getting a build error for [`submit-e2e-test-cloud-build`](https://console.cloud.google.com/cloud-build/builds/919b3ace-e280-41d6-a297-73ad1ad0ac53;step=25?project=agones-images) that I am not sure if is related to my changes. Re-running to see if it fails again the same way. I encountered a seemingly sporadic error with the [`deploy-site-static`](https://console.cloud.google.com/cloud-build/builds/3ed439f7-25ec-4bf7-8c96-ff04283db91b;step=22?project=agones-images) previously where it couldn't save the assets.

```
{
  "error": {
    "code": 401,
    "message": "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.",
    "status": "UNAUTHENTICATED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "CREDENTIALS_MISSING",
        "domain": "googleapis.com",
        "metadata": {
          "method": "google.devtools.cloudbuild.v1.CloudBuild.GetBuild",
          "service": "cloudbuild.googleapis.com"
        }
      }
    ]
  }
}
```

zmerlynn:
This last failure is on the new test:
```
=== CONT  TestGameServerPortPolicyNone
    gameserver_test.go:867: 
        	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:867
        	Error:      	Could not get a GameServer ready
        	Test:       	TestGameServerPortPolicyNone
        	Messages:   	creating {game-server [{  None <nil> 7777 0 }] {false 0 0 0}  { 0 0} {{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []} {[] [] [{game-server us-docker.pkg.dev/agones-images/examples/simple-game-server:0.32 [] []  [] [] [] {map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}] map[cpu:{{30 -3} {<nil>} 30m DecimalSI} memory:{{33554432 0} {<nil>}  BinarySI}] []} [] <nil> [] [] nil nil nil nil   IfNotPresent nil false false false}] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] [] <nil> nil <nil> [] []}} <nil> map[] map[] <nil>} GameServer instances failed (): admission webhook "validations.agones.dev" denied the request: GameServer.agones.dev "game-serverthjnv" is invalid: spec.ports[0].portPolicy: Invalid value: "None": portPolicy must be Dynamic or None on GKE Autopilot
--- FAIL: TestGameServerPortPolicyNone (2.18s)
FAIL test/e2e.TestGameServerPortPolicyNone (2.18s)
```

I'll leave you a review comment in a sec explaining what's up - basically you're not dealing with the feature flag correctly for e2es, which is different than unit tests.

test/e2e/gameserver_test.go
```diff
@@ -853,6 +849,32 @@ func TestGameServerPassthroughPort(t *testing.T) {
 	assert.Equal(t, "ACK: Hello World !\n", reply)
 }
 
+func TestGameServerPortPolicyNone(t *testing.T) {
+	t.Parallel()
+
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
+	require.NoError(t, runtime.ParseFeatures(string(runtime.FeaturePortPolicyNone)+"=true"))
```
zmerlynn, line 857: For e2es, we don't flip feature gates this way. What we do instead is do one run with "all flipped" and another run with "all default". You can see this as `e2e-feature-gate` and `e2e-stable` in the "sub-build" for e2es. We do this because we do a single helm install for the whole e2e run, so we can't really flip feature gates dynamically like we do in unit tests.

Instead of this, you should just skip the test if the feature gate is not enabled. You can see an example of this here: https://github.com/googleforgames/agones/blob/87ba7d2c6b426c7389ba257000f72603bf37e012/test/e2e/gameserver_test.go#L1153-L1155

zmerlynn:
> Thanks, it worked locally and I thought that my change for the e2e seemed to work this time - I must have been looking at the wrong place in the build log files.

It may have worked locally because you had the feature gate enabled in your install. :) I suspect if you reinstalled with the gate off, you'd see the failure. (I rarely flip when I'm testing locally, so completely understood.)

zmerlynn:
Unfortunately you just hit a different build failure that's happening across the board. I will retry it when we understand what's going on. :(

daniellee:
Finally got a green build.

In the latest iteration, I have:
- I have moved the docs changes behind feature shortcodes (and fixed a small indentation bug for the current list of PortPolicy options - you can see the formatting error on [gameserver docs page](https://agones.dev/site/docs/reference/gameserver/)). 
- Added feature gates in the code for policy none
- Got all the tests to work with feature gates 
- Ran `make gen-api-docs gen-crd-client` and generated a new crd api reference file

Let me know if I missed anything else around feature gates.

zmerlynn (commented):
This LGTM. There's a conflict in generated docs - you can rebase and re-run `make gen-crd-client` to resolve it quickly enough.

zmerlynn (approved):
Great job, thank you!

zmerlynn:
cc @markmandel who want to review, or I'll merge it soon

markmandel:
LGTM!

daniellee:
Thank you both for all the help!

Issue #3721, opened 2024-03-20
Support `Passthrough` on GKE Autopilot


zmerlynn:
When I implemented support for GKE Autopilot (#2777), I blocked the [`Passthrough`](https://agones.dev/site/docs/reference/gameserver/) port policy on Autopilot because it was complicated: Right now we do a handshake with Autopilot - we create the Pod with a special annotation that says ["give me a host port in range X..Y"](https://cloud.google.com/kubernetes-engine/docs/how-to/autopilot-hostport). We then backfill the `GameServer` object from that. That's fine for `Dynamic` since we can just treat Autopilot as the source of truth for the port number, but for `Passthrough` that doesn't work: On `Passthrough` we need to bind the container to the allocated port - but on Autopilot this allocation happens in an Autopilot webhook, during `Pod` creation. To support `Passthrough`, we would need a webhook on Pods _after_ Autopilot's `warden-mutating` webhook that could then propagate the port setting back to the container.

Looking at webhook options, we could do this pretty safely and filter it pretty much to this exact case:
* Let's do this all under an `AutopilotPassthroughPort` Feature Gate
* When there is a `Passthrough` port on Autopilot, add a new label, e.g. `agones.dev/port: "autopilot-passthrough"`
* In every install, add a webhook:
  * named `zzz-agones-mutation-webhook`: look, it's cheesy, but [webhooks are run in alphabetical order](https://www.reddit.com/r/kubernetes/comments/g864zf/is_there_a_way_to_change_the_triggering_order_of/) (I verified this), so pick an "end of the phone book" name. But even then, we could guarantee that it's run last with a [`reinvocationPolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy).
  * The webhook would be scoped to `Pod` `CREATE`
  * With an [`objectSelector`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector) of the label we created above - this drastically limits the scope and should increase the `extensions` load a bit, but not nearly so much as intercepting all Pod.

The webhook could then back the container port from the allocated port.

PR #3809, opened 2024-04-29
Passthrough autopilot - Adds an AutopilotPassthroughPort Feature Gate and new pod label

vicentefb:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

These changes are the first two steps towards #3721 
-Adds an AutopilotPassthroughPort Feature Gate
-Adds a new label, e.g. agones.dev/port: "autopilot-passthrough"

Closes #

none

### DIFF
--- a/pkg/cloudproduct/gke/gke.go
+++ b/pkg/cloudproduct/gke/gke.go
@@ -162,7 +162,7 @@ func (*gkeAutopilot) MutateGameServerPod(gss *agonesv1.GameServerSpec, pod *core
 }
 
 // setPassthroughLabel sets the agones.dev/port: "autopilot-passthrough" label to the game server container.
-// This will help to back the container port from the allocated port using an objectSelect of this label
+// This will help to back the container port from the allocated port using an objectSelector of this label
 // in GameServers that are using Passthrough Port Policy
 func setPassthroughLabel(gs *agonesv1.GameServerSpec, pod *corev1.Pod) {
 	if runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) && hasPortPolicy(gs, agonesv1.Passthrough) {
@@ -179,7 +179,6 @@ func setPrimaryContainer(pod *corev1.Pod, containerName string) {
 		return
 	}
 	pod.ObjectMeta.Annotations[primaryContainerAnnotation] = containerName
-
 }
 
 // podSpecSeccompUnconfined sets to seccomp profile to `Unconfined` to avoid serious performance
--- a/pkg/cloudproduct/gke/gke.go
+++ b/pkg/cloudproduct/gke/gke.go
@@ -155,9 +155,19 @@ func (*gkeAutopilot) ValidateScheduling(ss apis.SchedulingStrategy, fldPath *fie
 }
 
 func (*gkeAutopilot) MutateGameServerPod(gss *agonesv1.GameServerSpec, pod *corev1.Pod) error {
+	setPassthroughLabel(gss, pod)
 	setPrimaryContainer(pod, gss.Container)
 	podSpecSeccompUnconfined(&pod.Spec)
 	return nil
+}
+
+// setPassthroughLabel sets the agones.dev/port: "autopilot-passthrough" label to the game server container.
+// This will help to back the container port from the allocated port using an objectSelect of this label
+// in GameServers that are using Passthrough Port Policy
+func setPassthroughLabel(gs *agonesv1.GameServerSpec, pod *corev1.Pod) {
+	if runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) && hasPortPolicy(gs, agonesv1.Passthrough) {
+		pod.ObjectMeta.Labels[agonesv1.GameServerPortPolicyPodLabel] = "autopilot-passthrough"
+	}
 }
 
 // setPrimaryContainer sets the autopilot.gke.io/primary-container annotation to the game server container.
@@ -169,6 +179,7 @@ func setPrimaryContainer(pod *corev1.Pod, containerName string) {
 		return
 	}
 	pod.ObjectMeta.Annotations[primaryContainerAnnotation] = containerName
+
 }
 
 // podSpecSeccompUnconfined sets to seccomp profile to `Unconfined` to avoid serious performance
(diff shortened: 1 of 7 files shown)


pkg/gameservers/controller.go
```diff
@@ -521,6 +521,11 @@ func (c *Controller) syncGameServerCreatingState(ctx context.Context, gs *agones
 				})
 			}
 		}
+		if runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) && portPolicyOption(gs, agonesv1.Passthrough) {
+			gs.Spec.Template.Labels = make(map[string]string)
+			gs.Spec.Template.Labels[agonesv1.GameServerPortPolicyPodLabel] = "autopilot-passthrough"
```
zmerlynn, line 526: I would expect this to be in the GKE `cloudproduct` so that it's just triggered on Autopilot.

pkg/gameservers/controller.go
```diff
@@ -540,6 +545,15 @@ func (c *Controller) syncGameServerCreatingState(ctx context.Context, gs *agones
 	return gs, nil
 }
 
+func portPolicyOption(gs *agonesv1.GameServer, portPolicy agonesv1.PortPolicy) bool {
```
zmerlynn, line 548: maybe `hasPortPolicy`, since it's just hunting for a specific one?

pkg/gameservers/controller_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
 	"agones.dev/agones/pkg/cloudproduct/generic"
 	agtesting "agones.dev/agones/pkg/testing"
+	agonesruntime "agones.dev/agones/pkg/util/runtime"
```
zmerlynn, line 31: this is conventionally not aliased and kept as `runtime` - is that a problem? I don't think I've seen it aliased in our code.

pkg/util/runtime/features.go
```diff
@@ -52,6 +52,9 @@ const (
 	// FeaturePlayerTracking is a feature flag to enable/disable player tracking features.
 	FeaturePlayerTracking Feature = "PlayerTracking"
 
+	// FeatureAutopilotPassthroughPort is a feature flag that enables/disables Passthrough Port Policy.
+	FeatureAutopilotPassthroughPort Feature = "PassthroughPortPolicy"
```
zmerlynn, line 56: I would put this in `Dev features`, and follow the instructions on line 73-76. In particular, you're missing the `install/helm/agones` change, which you'll need when we add the webhook to make it conditional in helm.

After the next step (when you add the webhook), we can go ahead and move it to Alpha. See lines 79-83 for those instructions, but it's basically a couple of modifications to ensure we test with it enabled.

zmerlynn (changes requested):

pkg/gameservers/controller_test.go
vicentefb: If I do `runtime "agones.dev/agones/pkg/util/runtime"` there's a conflict with `"k8s.io/apimachinery/pkg/runtime"` which is already declared and I cannot remove that library because it'll trigger many undefined errors.

pkg/util/runtime/features.go
vicentefb: Ohhh okok i missed all this section of code, sg, thanks!

pkg/gameservers/controller_test.go
zmerlynn: Oh, I missed that conflict. It looks like elsewhere when there's a conflict we import it as `k8sruntime`: https://github.com/search?q=repo%3Agoogleforgames%2Fagones%20%22k8s.io%2Fapimachinery%2Fpkg%2Fruntime%22&type=code .. I would be inclined to switch that one instead.

I know this sounds nitpicky but I've often found it's an important readability thing, otherwise a reader conversat in our habits may get confused.

vicentefb: I see, makes sense, thanks! Nevertheless, since i'm moving the implementation to GKE `cloudproduct` I'll add a unit test inside gke_test.

pkg/gameservers/controller.go
vicentefb: Okok moved implementation to gke.go inside a method call in `MutateGameServerPod()`

pkg/cloudproduct/gke/gke.go
```diff
@@ -169,6 +179,7 @@ func setPrimaryContainer(pod *corev1.Pod, containerName string) {
 		return
 	}
 	pod.ObjectMeta.Annotations[primaryContainerAnnotation] = containerName
+
```
zmerlynn, line 182: nit: excess newline

pkg/gameservers/controller.go
```diff
@@ -521,6 +521,7 @@ func (c *Controller) syncGameServerCreatingState(ctx context.Context, gs *agones
 				})
 			}
 		}
+
```
zmerlynn, line 524: nit: while this newline is probably chill, there's no other diff in this file, so maybe minimize the impact and kill it :)

pkg/cloudproduct/gke/gke_test.go
```diff
@@ -222,6 +224,52 @@ func TestPodSeccompUnconfined(t *testing.T) {
...
+			},
+			wantPod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{},
+					Labels: map[string]string{
+						agonesv1.GameServerPortPolicyPodLabel: "autopilot-passthrough",
+					},
+				},
+			},
+		},
```
zmerlynn, line 259: add negative tests for:
* feature gate disabled
* no passthrough policy
just to verify it's not modifying anything in those cases. (I thikn it would trip other tests but might as well)

zmerlynn (commented):
LGTM modulo a couple of nits

pkg/cloudproduct/gke/gke_test.go
vicentefb: Sure, done!

zmerlynn (approved):

PR #3811, opened 2024-05-08
Test: CI Check for Changes in Examples Directory

### LINKED ISSUE #3794: Fail CI if a PR updates an example without modifying the version
If someone modifies anything in `examples/foo`, there should be a corresponding change to the `version :=` line in the Makefile. This should be easy to catch in CI. This would catch some of the common cases like #3793 (which should have been flagged), and also fail Dependabot PRs against examples/ early.

If we want to get fancy, there's technically files like `README.md` that can be ignored in every directory.

Kalaiselvi84:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #3794 


Kalaiselvi84:
@markmandel The script successfully detects changes in main.go, Dockerfile, Dockerfile.windows, go.mod, and go.sum within the examples directory on my local; however, it is not working as expected in CI. What could be the reason?

build/includes/sdk.mk
```diff
@@ -251,3 +251,33 @@ test-gen-all-sdk-grpc:
 		echo "$$diff_output_test_sdk"; \
 		exit 1; \
 	fi
+
+# Ensure examples images are bumped when there is any change in the scripts
+check-makefile-version-increment:
+	@echo "Current directory: $$(pwd)"
```
markmandel, line 257: That is some seriously impressive bash/make.

build/includes/sdk.mk
```diff
@@ -251,3 +251,33 @@ test-gen-all-sdk-grpc:
...
+	@echo "Checking for script changes in 'examples' subdirectories..."
+	@cd $(CURDIR)/..; \
+	changed_dirs=$$(git diff --name-only HEAD | grep -E "examples/.*/(main\.go|Dockerfile|Dockerfile\.windows|go\.mod|go\.sum)$$" | xargs -n1 -r dirname | sort -u); \
```
markmandel, line 260: Ow my head hurts 😄 can we get some descriptive comments in here to work out what this is all checking?

examples/allocation-endpoint/Makefile
```diff
@@ -30,7 +30,7 @@ PROD_REPO ?= us-docker.pkg.dev/agones-images/examples
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 project_path := $(dir $(mkfile_path))
 root_path = $(realpath $(project_path)/../..)
-version := 0.8
+version := 0.8  # Change this to ignore the version change presubmit: `check-makefile-version-increment`
```
markmandel, line 33: "ignore" ?

Not quite sure what this means. Or are you saying "if CI fails on `check-makefile-version-increment`` make sure to increment this value" ?

build/includes/sdk.mk
Kalaiselvi84: Sorry, I agree the script is confusing. Could you please share a better approach? I don't think this bash script will work well.

examples/allocation-endpoint/Makefile
Kalaiselvi84: I've updated the comment to say, "Bump this version if CI fails on `check-makefile-version-increment`"

build/includes/sdk.mk
Kalaiselvi84: > changed_dirs=$$(git diff --name-only HEAD | grep -E "examples/.*/(main\.go|Dockerfile|Dockerfile\.windows|go\.mod|go\.sum)$$" | xargs -n1 -r dirname | sort -u); \

This will find all the changed files under `examples` subdirectories that match the given patterns (main.go, Dockerfile, Dockerfile.windows, go.mod, go.sum), after that it will extract the directories containing these files, remove duplicates, and store them in `changed_dirs`.

markmandel:
```
ERROR: failed to solve: failed to compute cache key: mount callback failed on /tmp/containerd-mount4055613205: link /tmp/containerd-mount4055613205/Windows/INF/basicrender.inf /tmp/containerd-mount4055613205/Windows/System32/DriverStore/FileRepository/basicrender.inf_amd64_efdc64af60c69a6d/basicrender.inf: no such file or directory
make: *** [Makefile:619: build-agones-sdk-image-windows-ltsc2019] Error 1
```

FYI - this isn't your thing failing, you'll want to update against main.

build/includes/sdk.mk
markmandel: > Sorry, I agree the script is confusing. Could you please share a better approach? I don't think this bash script will work well.

Go script? You've created some really nice ones in the past. Probably easier too!

https://pkg.go.dev/github.com/go-git/go-git/v5 looks like the dominant Go Git library.

On failure, you could `log.Fatalf(...)` and CI would fail, since the exit code of the script would be non-zero.

> This will find all the changed files under examples subdirectories that match the given patterns (main.go, Dockerfile, Dockerfile.windows, go.mod, go.sum)

Since each of the examples use different languages, I'm wondering if we want to exclude things we know we don't care about (*.md, *.yaml, OWNERS, .gitignore) rather than an include list - as the number of examples and language will expand over time.

Kalaiselvi84: Thanks for your feedback. I will implement a Go script👍

Kalaiselvi84: Update: The Go script implementation is in progress. Looks like bash script is simpler than the Go script. Will push the commit once completed.

markmandel: > Looks like bash script is simpler than the Go script

I'm hoping it's easier to understand though 😄 `grep -oP '^[^#]*version\s*:=\s*\K[^ ]+'` - that scared me 😨 😀

Kalaiselvi84: I have implemented a Go script but ignoring comments from the files is the challenging part for me. https://gist.github.com/Kalaiselvi84/cadc597d7fa2c229ecf2a992ddbd4b01 

I think it is infeasible for me, giving up. I will close this PR. Thanks for your help and support.🙏

PR #3833, opened 2024-05-22
Passthrough autopilot - Adds mutating webhook

vicentefb:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
 /kind feature
> /kind hotfix
> /kind release

These changes are related to the 3rd step towards #3721 
-Adds `zzz-agones-mutation-webhook` webhook and its handler
-Missing to handle the case a ports array that's mixed with Dynamic and Passthrough ports for example.

Closes #

none

### DIFF
--- a/pkg/cloudproduct/gke/gke.go
+++ b/pkg/cloudproduct/gke/gke.go
@@ -263,13 +263,13 @@ func (apa *autopilotPortAllocator) Allocate(gs *agonesv1.GameServer) *agonesv1.G
 
 	var ports []agonesv1.GameServerPort
 	for i, p := range gs.Spec.Ports {
-		/*
-			if p.PortPolicy != agonesv1.Dynamic {
-				logger.WithField("gs", gs.Name).WithField("portPolicy", p.PortPolicy).Error(
-					"GameServer has invalid PortPolicy for Autopilot - this should have been rejected by webhooks. Refusing to assign ports.")
-				return gs
-			}
-		*/
+
+		if p.PortPolicy != agonesv1.Dynamic && !runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) {
+			logger.WithField("gs", gs.Name).WithField("portPolicy", p.PortPolicy).Error(
+				"GameServer has invalid PortPolicy for Autopilot - this should have been rejected by webhooks. Refusing to assign ports.")
+			return gs
+		}
+
 		p.HostPort = int32(i + 1) // Autopilot expects _some_ host port - use a value unique to this GameServer Port.
 
 		if p.Protocol == agonesv1.ProtocolTCPUDP {
--- a/pkg/cloudproduct/gke/gke.go
+++ b/pkg/cloudproduct/gke/gke.go
@@ -134,10 +134,18 @@ func (*gkeAutopilot) NewPortAllocator(portRanges map[string]portallocator.PortRa
 
 func (*gkeAutopilot) WaitOnFreePorts() bool { return true }
 
+func checkPassthroughPortPolicy(PortPolicy agonesv1.PortPolicy) bool {
+	// if feature is not enabled and port is Passthrough return true because that should be an invalid port
+	// if feature is not enabled and port is not Passthrough you can return false because there's no error  but check for None port
+	// if feature is enabled and port is passthrough return false because there is no error
+	// if feature is enabled and port is not passthrough return false because there is no error but check for None port
+	return (!runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) && PortPolicy == agonesv1.Passthrough) || PortPolicy == agonesv1.Static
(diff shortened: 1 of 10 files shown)


pkg/gameservers/controller.go
```diff
@@ -316,6 +320,44 @@ func (ext *Extensions) creationValidationHandler(review admissionv1.AdmissionRev
 	return review, nil
 }
 
+// creationMutationHandlerPod that mutates a GameServer pod when it is created
+// Should only be called on gameserver pod create operations.
+func (ext *Extensions) creationMutationHandlerPod(review admissionv1.AdmissionReview) (admissionv1.AdmissionReview, error) {
+	logger := runtime.NewLoggerWithSource("gke")
```
zmerlynn, line 326: The `Extensions` type already has a `baseLogger` in it. Personally since you only use it once, I would just remove this variable and on line 336, I would do `ext.baseLogger.WithField(...)`

That said, logging the whole review object is probably excessive even in Debug. I'd log the Pod name / namespace so you can track the flow as necessary, though.

pkg/gameservers/controller.go
```diff
@@ -316,6 +320,44 @@ func (ext *Extensions) creationValidationHandler(review admissionv1.AdmissionRev
...
+	pod := &corev1.Pod{}
+	err := json.Unmarshal(obj.Raw, pod)
+	if err != nil {
+		// If the JSON is invalid during mutation, fall through to validation. This allows OpenAPI schema validation
+		// to proceed, resulting in a more user friendly error message.
+		return review, nil
+	}
+
+	logger.WithField("review", review).Debug("creationMutationHandlerPod")
+	pod.Spec.Containers[1].Ports[0].ContainerPort = pod.Spec.Containers[1].Ports[0].HostPort
```
zmerlynn, line 337: Put a big TODO in here basically saying we need to deal with the case where there are multiple, mixed type ports. I think this is ok _for now_ but we'll need to deal with it before we enable the feature gate.

pkg/gameservers/controller_test.go
```diff
@@ -2225,3 +2274,9 @@ func newSingleContainerSpec() agonesv1.GameServerSpec {
 		},
 	}
 }
+
+func newPassthroughPortSingleContainerSpec() corev1.PodSpec {
+	return corev1.PodSpec{
+		Containers: []corev1.Container{{Name: "agones-gameserver-sidecar", Image: "container/image", Env: []corev1.EnvVar{{Name: passthroughPortEnvVar, Value: "TRUE"}}}, {Name: "simple-game-server", Image: "container2/image", Ports: []corev1.ContainerPort{{HostPort: 7777, ContainerPort: 555}}, Env: []corev1.EnvVar{{Name: passthroughPortEnvVar, Value: "TRUE"}}}},
```
zmerlynn, line 2280: Please add some newlines and make this prettier. :)

pkg/util/webhooks/webhooks.go
```diff
@@ -86,6 +86,7 @@ func (wh *WebHook) handle(path string, w http.ResponseWriter, r *http.Request) e
 		review.Response = &admissionv1.AdmissionResponse{Allowed: true}
 	}
 	review.Response.UID = review.Request.UID
+	wh.logger.WithField("name", review.Request.Name).WithField("path", path).WithField("kind", review.Request.Kind.Kind).WithField("group", review.Request.Kind.Group).Info("handling webhook request")
```
zmerlynn, line 89: Probably `Debug` rather than `Info`, it's a kind of hot path.

zmerlynn (changes requested):

pkg/cloudproduct/gke/gke.go
```diff
@@ -137,7 +137,7 @@ func (*gkeAutopilot) WaitOnFreePorts() bool { return true }
 func (g *gkeAutopilot) ValidateGameServerSpec(gss *agonesv1.GameServerSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := g.ValidateScheduling(gss.Scheduling, fldPath.Child("scheduling"))
 	for i, p := range gss.Ports {
-		if p.PortPolicy != agonesv1.Dynamic && (p.PortPolicy != agonesv1.None || !runtime.FeatureEnabled(runtime.FeaturePortPolicyNone)) {
+		if p.PortPolicy != agonesv1.Dynamic && (p.PortPolicy != agonesv1.None || !runtime.FeatureEnabled(runtime.FeaturePortPolicyNone)) && !runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) {
```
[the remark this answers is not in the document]
vicentefb: Okok ohh i forgot to actually check for the port, sg, done! I also added as a comment an explanation of the check.

zmerlynn (approved):
Good job!

PR #3842, opened 2024-05-30
Passthrough autopilot - added ports array case and updated unit tests

vicentefb:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

Handles case where we receive an array of mixed ports (Dynamic and Passthrough) for Autopilot clusters. 

Towards #3721

### DIFF
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -157,7 +157,7 @@ const (
 	NodePodIP corev1.NodeAddressType = "PodIP"
 
 	// PassthroughPortAssignmentAnnotation is an annotation to keep track of game server container and its Passthrough ports indices
-	PassthroughPortAssignmentAnnotation = "autopilot.gke.io/container-passthrough-port-assignment"
+	PassthroughPortAssignmentAnnotation = "agones.dev/container-passthrough-port-assignment"
 
 	// True is the string "true" to appease the goconst lint.
 	True = "true"
@@ -749,7 +749,7 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 
 	gs.podObjectMeta(pod)
 
-	passthroughContainerPortMap := make(map[string][]string)
+	passthroughContainerPortMap := make(map[string][]int)
 	for i, p := range gs.Spec.Ports {
 		var hostPort int32
 		if !runtime.FeatureEnabled(runtime.FeaturePortPolicyNone) || p.PortPolicy != None {
@@ -770,7 +770,7 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 			return nil, err
 		}
 		if runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) && p.PortPolicy == Passthrough {
-			passthroughContainerPortMap[*p.Container] = append(passthroughContainerPortMap[*p.Container], fmt.Sprint(i))
+			passthroughContainerPortMap[*p.Container] = append(passthroughContainerPortMap[*p.Container], i)
 		}
 	}
 
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -750,7 +750,8 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 	gs.podObjectMeta(pod)
 
 	passthroughContainerPortMap := make(map[string][]int)
-	for i, p := range gs.Spec.Ports {
+	portIdx := 0
+	for _, p := range gs.Spec.Ports {
 		var hostPort int32
 		if !runtime.FeatureEnabled(runtime.FeaturePortPolicyNone) || p.PortPolicy != None {
 			hostPort = p.HostPort
@@ -762,6 +763,7 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 			Protocol:      p.Protocol,
 		}
 		err := gs.ApplyToPodContainer(pod, *p.Container, func(c corev1.Container) corev1.Container {
+			portIdx = len(c.Ports)
 			c.Ports = append(c.Ports, cp)
 
 			return c
@@ -770,7 +772,7 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 			return nil, err
 		}
 		if runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) && p.PortPolicy == Passthrough {
-			passthroughContainerPortMap[*p.Container] = append(passthroughContainerPortMap[*p.Container], i)
+			passthroughContainerPortMap[*p.Container] = append(passthroughContainerPortMap[*p.Container], portIdx)
 		}
 	}
 
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -156,6 +156,8 @@ const (
 	// NodePodIP identifies an IP address from a pod.
 	NodePodIP corev1.NodeAddressType = "PodIP"
 
+	PassthroughPortAssignmentAnnotation = "autopilot.gke.io/container-passthrough-port-assignment"
+
 	// True is the string "true" to appease the goconst lint.
 	True = "true"
 	// False is the string "false" to appease the goconst lint.
@@ -745,9 +747,10 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 	}
 
 	gs.podObjectMeta(pod)
-	for _, p := range gs.Spec.Ports {
-		var hostPort int32
 
+	passthroughContainerPortMap := make(map[string][]string)
+	for i, p := range gs.Spec.Ports {
+		var hostPort int32
 		if !runtime.FeatureEnabled(runtime.FeaturePortPolicyNone) || p.PortPolicy != None {
 			hostPort = p.HostPort
 		}
@@ -765,7 +768,20 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 		if err != nil {
 			return nil, err
 		}
+		if runtime.FeatureEnabled(runtime.FeatureAutopilotPassthroughPort) && p.PortPolicy == Passthrough {
+			passthroughContainerPortMap[*p.Container] = append(passthroughContainerPortMap[*p.Container], fmt.Sprint(i))
+		}
 	}
+
+	if len(passthroughContainerPortMap) != 0 {
+		var containerToPassthroughMapJSON []byte
+		containerToPassthroughMapJSON, err := json.Marshal(passthroughContainerPortMap)
+		if err != nil {
+			return nil, err
+		}
+		pod.ObjectMeta.Annotations[PassthroughPortAssignmentAnnotation] = string(containerToPassthroughMapJSON)
+	}
+
 	// Put the sidecars at the start of the list of containers so that the kubelet starts them first.
 	containers := make([]corev1.Container, 0, len(sidecars)+len(pod.Spec.Containers))
 	containers = append(containers, sidecars...)
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -753,6 +753,7 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 	portIdx := 0
 	for _, p := range gs.Spec.Ports {
 		var hostPort int32
+
 		if !runtime.FeatureEnabled(runtime.FeaturePortPolicyNone) || p.PortPolicy != None {
 			hostPort = p.HostPort
 		}
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -156,6 +156,7 @@ const (
 	// NodePodIP identifies an IP address from a pod.
 	NodePodIP corev1.NodeAddressType = "PodIP"
 
+	// PassthroughPortAssignmentAnnotation is an annotation to keep track of game server container and its Passthrough ports indices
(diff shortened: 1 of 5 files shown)


pkg/apis/agones/v1/gameserver.go
```diff
@@ -156,6 +156,8 @@ const (
 	// NodePodIP identifies an IP address from a pod.
 	NodePodIP corev1.NodeAddressType = "PodIP"
 
+	PassthroughPortAssignmentAnnotation = "autopilot.gke.io/container-passthrough-port-assignment"
```
zmerlynn, line 160: Use `agones.dev` rather than `autopilot.gke.io`. (In general, annotations are "owned" by the domain name - so it would be surprising to find the Agones project emitting a `gke.io` annotation.)

pkg/apis/agones/v1/gameserver_test.go
```diff
@@ -1497,6 +1498,48 @@ func TestGameServerDisableServiceAccount(t *testing.T) {
...
+	require.NoError(t, runtime.ParseFeatures(string(runtime.FeatureAutopilotPassthroughPort)+"=true"))
+	gs := &GameServer{ObjectMeta: metav1.ObjectMeta{Name: "gameserver", UID: "1234"}, Spec: GameServerSpec{
+		Ports: []GameServerPort{
+			{Name: "defaultPassthrough", PortPolicy: Passthrough},
+			{Name: "defaultDynamic", PortPolicy: Dynamic, ContainerPort: 7654},
+			{Name: "defaultDynamicTwo", PortPolicy: Dynamic, ContainerPort: 7659},
+			{Name: "defaultPassthroughTwo", PortPolicy: Passthrough},
+		},
+		Template: corev1.PodTemplateSpec{
+			Spec: corev1.PodSpec{
```
zmerlynn, line 1514: Please test the multi-container case. The `Ports` array can take a `container` name as well, so `Ports` could be a mix of different container ports.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -745,9 +747,10 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 	}
 
 	gs.podObjectMeta(pod)
-	for _, p := range gs.Spec.Ports {
-		var hostPort int32
 
+	passthroughContainerPortMap := make(map[string][]string)
```
zmerlynn, line 752: why string and not int?

pkg/gameservers/controller.go
```diff
@@ -334,8 +334,27 @@ func (ext *Extensions) creationMutationHandlerPod(review admissionv1.AdmissionRe
 
 	ext.baseLogger.WithField("pod.Name", pod.Name).Debug("creationMutationHandlerPod")
 
-	// TODO: We need to deal with case of multiple and mixed type ports before enabling the feature gate.
-	pod.Spec.Containers[1].Ports[0].ContainerPort = pod.Spec.Containers[1].Ports[0].HostPort
+	annotation, ok := pod.ObjectMeta.Annotations[agonesv1.PassthroughPortAssignmentAnnotation]
+	if !ok {
+		return review, nil
```
zmerlynn, line 339: log this case? We're not expecting to see it, are we, since we have an object selector that limits it to the specific label for passthrough? So we know it has at least one port.

pkg/util/runtime/features.go
```diff
@@ -140,7 +140,7 @@ var (
 		FeaturePortPolicyNone:                   false,
 
 		// Dev features
-		FeatureAutopilotPassthroughPort: false,
```
zmerlynn: Did you mean to include this? This would be the equivalent of promoting it to Beta.

I think after this change goes in you can move it up to Alpha, though (follow the instructions above to do that).

zmerlynn (commented):
Looking good! A few comments.

pkg/apis/agones/v1/gameserver.go
vicentefb: Okok sounds good!

pkg/apis/agones/v1/gameserver.go
vicentefb: True, it can be an int and that will save some conversions.

pkg/gameservers/controller.go
vicentefb: You're right, i'll add a debug log, thanks!

pkg/apis/agones/v1/gameserver_test.go
vicentefb: Added!

pkg/gameservers/controller.go
```diff
@@ -334,8 +334,24 @@ func (ext *Extensions) creationMutationHandlerPod(review admissionv1.AdmissionRe
...
+	passthroughPortAssignmentMap := make(map[string][]int)
+	if err := json.Unmarshal([]byte(annotation), &passthroughPortAssignmentMap); err != nil {
+		return review, errors.Wrapf(err, "could not unmarshal annotation %q (value %q)", passthroughPortAssignmentMap, annotation)
+	}
+
+	for _, p := range pod.Spec.Containers {
+		if containerPassthrough, ok := passthroughPortAssignmentMap[p.Name]; ok {
+			for i := range containerPassthrough {
+				i := containerPassthrough[i]
+				p.Ports[i].ContainerPort = p.Ports[i].HostPort
```
zmerlynn, line 351: So, this is subtle but I think your logic here is off. In particular, in gameserver.go you're recording the index within the Ports array, but here you're indexing into container.Ports by the _absolute_ index, vs the index within that container. You can see this in your new unit text where the annotation records `containerFour: [5]`, but that's index 5 into the GameServer Ports array, not index 5 into the container Ports array.

To fix this, I'd recommend two things:
* In gamserver.go, instead of adding an index variable to the range over the GameServer ports, record the index before the line `c.Ports = append(c.Ports, cp)`, I.e. by `portIdx := len(c.Ports)`.
* Test the multi-container case here as well - in fact if you mirror the container/port structure in the test above you just added, it'll make for a good symmetric test.

pkg/gameservers/controller.go
```diff
@@ -334,8 +334,24 @@ func (ext *Extensions) creationMutationHandlerPod(review admissionv1.AdmissionRe
...
+	if !ok {
+		ext.baseLogger.WithField("pod.Name", pod.Name).Debug("creationMutationHandlerPod, " + agonesv1.PassthroughPortAssignmentAnnotation + ": " + pod.ObjectMeta.Annotations[agonesv1.PassthroughPortAssignmentAnnotation])
+	}
+
+	passthroughPortAssignmentMap := make(map[string][]int)
+	if err := json.Unmarshal([]byte(annotation), &passthroughPortAssignmentMap); err != nil {
+		return review, errors.Wrapf(err, "could not unmarshal annotation %q (value %q)", passthroughPortAssignmentMap, annotation)
+	}
+
+	for _, p := range pod.Spec.Containers {
```
zmerlynn, line 347: total nit but `p` is a weird name for a container. `c`? It confused me when I saw `p.Ports[i]` because I thought it was somehow an array on the pod.

pkg/gameservers/controller.go
```diff
@@ -334,8 +334,24 @@ func (ext *Extensions) creationMutationHandlerPod(review admissionv1.AdmissionRe
...
+		if containerPassthrough, ok := passthroughPortAssignmentMap[p.Name]; ok {
+			for i := range containerPassthrough {
+				i := containerPassthrough[i]
```
zmerlynn, line 350: Two things:
1) line 349/350 are a weird pairing because you're ranging over the indices, then immediately shadowing with the value. This can be shortened to just: `for _, i := range containerPassthrough`

That said, though, I would shy away from `i` here and probably make it slightly more descriptive, like `portIdx`.

2) Fun fact, `passthroughPortAssignmentMap[p.Name]` will default to `nil`, and a `nil` slice is an empty slice, so you can skip 348.

Basically I think line 348-350 can be replaced with:
```
for _, portIdx := range passthroughPortAssignmentMap[p.Name] {
}
```

zmerlynn (commented):
Closer!

pkg/gameservers/controller.go
vicentefb: Okok sg, I changed the for loop variable name and replaed lines 348-350.

pkg/gameservers/controller.go
vicentefb: Ohhhh interesting, good catch! I made the changes and now i'm using the container port index instead and updated unit test to reflect this mirror in the multi container test, thanks!!

pkg/gameservers/controller.go
```diff
@@ -334,8 +334,21 @@ func (ext *Extensions) creationMutationHandlerPod(review admissionv1.AdmissionRe
 
 	ext.baseLogger.WithField("pod.Name", pod.Name).Debug("creationMutationHandlerPod")
 
-	// TODO: We need to deal with case of multiple and mixed type ports before enabling the feature gate.
-	pod.Spec.Containers[1].Ports[0].ContainerPort = pod.Spec.Containers[1].Ports[0].HostPort
+	annotation, ok := pod.ObjectMeta.Annotations[agonesv1.PassthroughPortAssignmentAnnotation]
+	if !ok {
+		ext.baseLogger.WithField("pod.Name", pod.Name).Debug("creationMutationHandlerPod, " + agonesv1.PassthroughPortAssignmentAnnotation + ": " + pod.ObjectMeta.Annotations[agonesv1.PassthroughPortAssignmentAnnotation])
```
zmerlynn, line 339: I don't think you need to log the value here because you know it's empty. Just say that the annotation is empty and it's unexpected. Feel free to bump it to an `Info`, too, since it's unexpected (so potentially more relevant than debug), but not really that important as it's just noise.

Separately I think you still want to return here after logging, no? I presume the unmarshal will fail o/w.

pkg/gameservers/controller_test.go
```diff
@@ -2275,15 +2277,32 @@ func newSingleContainerSpec() agonesv1.GameServerSpec {
...
+	containerToPassthroughMapJSON, err := json.Marshal(passthroughContainerPortMap)
+	if err != nil {
+		return corev1.Pod{}
+	}
+	return corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{agonesv1.PassthroughPortAssignmentAnnotation: string(containerToPassthroughMapJSON)},
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
```
zmerlynn, line 2295: I thought you were going to change this test to also be multi-container?

pkg/gameservers/controller_test.go
```diff
@@ -2275,15 +2277,32 @@ func newSingleContainerSpec() agonesv1.GameServerSpec {
...
-				Image: "container2/image",
-				Ports: []corev1.ContainerPort{{HostPort: 7777, ContainerPort: 555}},
-				Env:   []corev1.EnvVar{{Name: passthroughPortEnvVar, Value: "TRUE"}}}},
+// Assume ports 0 and 2 are Passthrough ports for "example-server" container
+// The annotation would look like autopilot.gke.io/passthrough-port-assignment: '{"example-server":["0","2"]}'
+func newPassthroughPortSingleContainerSpec() corev1.Pod {
+	passthroughContainerPortMap := make(map[string][]int)
+	passthroughContainerPortMap["example-server"] = append(passthroughContainerPortMap["example-server"], 0, 2)
+	var containerToPassthroughMapJSON []byte
+	containerToPassthroughMapJSON, err := json.Marshal(passthroughContainerPortMap)
```
zmerlynn, line 2286: As much as it seems like it's more readable to test it using a map, you expose yourself to the possibility of a "symmetric error" since you're doing this same operation to construct the JSON in both tests. It would be better just to use a string here, and it also documents the format.

pkg/gameservers/controller_test.go
```diff
@@ -2275,15 +2277,32 @@ func newSingleContainerSpec() agonesv1.GameServerSpec {
...
-		Containers: []corev1.Container{
-			{Name: "agones-gameserver-sidecar",
-				Image: "container/image",
-				Env:   []corev1.EnvVar{{Name: passthroughPortEnvVar, Value: "TRUE"}}},
-			{Name: "simple-game-server",
-				Image: "container2/image",
-				Ports: []corev1.ContainerPort{{HostPort: 7777, ContainerPort: 555}},
-				Env:   []corev1.EnvVar{{Name: passthroughPortEnvVar, Value: "TRUE"}}}},
+// Assume ports 0 and 2 are Passthrough ports for "example-server" container
+// The annotation would look like autopilot.gke.io/passthrough-port-assignment: '{"example-server":["0","2"]}'
```
zmerlynn, line 2281: I think this comment is out of date (it's `{"example-server":[0,2]}`, right?)

zmerlynn (commented):
Almost!

pkg/apis/agones/v1/gameserver.go
```diff
@@ -745,9 +748,11 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
 	}
 
 	gs.podObjectMeta(pod)
+
+	passthroughContainerPortMap := make(map[string][]int)
+	portIdx := 0
```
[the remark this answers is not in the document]
vicentefb: I need `passthroughContainerPortMap` to be outside the for loop because i use it in the following if statement from line 779.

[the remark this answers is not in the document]
zmerlynn: but `portIdx` only applies inside the loop, right?

pkg/gameservers/controller_test.go
vicentefb: I think it could still be valid case unless I'm missing something (?) In this method i'm building the PodSpec and the `example-server` container has 3 ports so technically container ports 0 and 2 are still valid. I'm aware that the `HostPort` and `ContainerPort` are not the same value in this piece of code but that's intentional since at this point the webhook hasn't been triggered. Nevertheless, I included another container in this PodSpec and updated the comment.

pkg/apis/agones/v1/gameserver.go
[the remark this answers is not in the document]
vicentefb: Yeah, I did move `portIdx` to happen inside :)

zmerlynn (approved):

zmerlynn:
Great job!

PR #3843, opened 2024-05-30
agones-{extensions,allocator}: Pause after cancelling context

zmerlynn:
After #3839 went in, we noticed the flakes in TestAllocatorAfterDeleteReplica disappear, but TestGameServerCreationRightAfterDeletingOneExtensionsPod remained. I looked more closely at this and I developed a theory: My guess is that `kube-apiserver` connections to webhooks are actually "sticky" using http(s) keepalives. If the TCP connection never closes, we'd see the behavior we do, which is the `EOF` from kube-apiserver going to write on a dead socket.

I looked at how to close sockets on the server side to "finish" the drain, but realized that we actually do, by cancelling the context. The thing is that we cancel the context and immediately exit, but it leaves no time for anything to shut down.

Of course, then I looked more into this and ... I'm not sure why this is helping. But I was able to reproduce the e2e flake locally, and with this small change I'm able to see it go away, so 🤷 .

Reverts #3839

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -247,7 +247,8 @@ func main() {
 		podReady = false
 		time.Sleep(conf.ReadinessShutdownDuration)
 		cancelCtx()
-		logger.Infof("Readiness shutdown duration has passed, exiting pod")
+		logger.Infof("Readiness shutdown duration has passed, context cancelled")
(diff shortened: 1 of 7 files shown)


markmandel (approved):
I'm down to give it a shot 😁

PR #3844, opened 2024-05-30
Change the line to modify in Quickstart: Edit a Game Server

peterzhongyi:
This is my first pull request!

The tutorial doc on Quickstart: Edit a Game Server instructed to change a log line that can only been seen when we connect to the game server with tcp connection. However the tutorial later connects to the game server with udp connection, so the log changes can't actually been seen.

Changing the log in the udpReadWriteLoop function is more useful because the log changes will been in the terminal.


markmandel (approved):
👍🏻

PR #3845, opened 2024-05-31
agones-{extensions,allocator}: Make servers context aware

zmerlynn:
* adds an `httpserver` utility package to handle the `Run` function that controller/extensions use. Make that context aware using the same method as https.Run: https://github.com/googleforgames/agones/blob/dfa414e5e4da37798833bbf8c33919acb5f3c2ea/pkg/util/https/server.go#L127-L130
* also plumbs context-awareness through the allocator run{Mux,REST,GRPC} functions.
* adds a gRPC health server to the allocator, calls .Shutdown() on it during graceful termination - this seems to push the client off correctly.

Along the way, basically re-wrote the `TestAllocatorAfterDeleteReplica` test:
* moved the `GetAllocatorClient` up because it's fairly disruptive (new certs, etc.), so I wanted the pod stability checks to happen after
* changed `agones-allocator` stability checks to just grab the Deployment and wait for `Replicas != AvailableReplicas`.
* changed the test to actually disrupt the Pods in the middle of the test (vs before it just nukes the pods then starts the test.

Tested with e2e in a loop.

Towards #3853


zmerlynn:
Hmm, upon further testing this still has a minor flake. Let me play with it a bit more, moving to draft.

cmd/extensions/main.go
```diff
@@ -150,7 +150,7 @@ func main() {
 	agonesInformerFactory := externalversions.NewSharedInformerFactory(agonesClient, defaultResync)
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, defaultResync)
 
-	server := &httpServer{}
+	server := &httpserver.Server{Logger: logger}
```
markmandel, line 153:
```suggestion
	server := &http.Server{Logger: logger}
```

That wouldn't be redundant 😃

cmd/allocator/main.go
```diff
@@ -245,16 +246,14 @@ func main() {
...
-		logger.Infof("Readiness shutdown duration has passed, context cancelled")
-		time.Sleep(1 * time.Second) // allow a brief time for cleanup, but force exit if main doesn't
-		os.Exit(0)
+		cancelListenCtx()
 	})
 
 	grpcUnallocatedStatusCode := grpcCodeFromHTTPStatus(conf.httpUnallocatedStatusCode)
 
-	h := newServiceHandler(ctx, kubeClient, agonesClient, health, conf.MTLSDisabled, conf.TLSDisabled, conf.remoteAllocationTimeout, conf.totalRemoteAllocationTimeout, conf.allocationBatchWaitTime, grpcUnallocatedStatusCode)
+	h := newServiceHandler(context.Background(), kubeClient, agonesClient, health, conf.MTLSDisabled, conf.TLSDisabled, conf.remoteAllocationTimeout, conf.totalRemoteAllocationTimeout, conf.allocationBatchWaitTime, grpcUnallocatedStatusCode)
```
markmandel, line 256: passing in `ctx.Background()` seems not ideal here - it's passed through to all the worker queues and informers etc - so none of them will shutdown on termination.

Any reason we can't use `listenCtx` ?

pkg/util/httpserver/server.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package httpserver implements an http server that conforms to the
+// controller runner interface.
+package httpserver
```
markmandel, line 17:
```suggestion
package http
```

I'd be tempted to make the package http since we also have a https - not sure the server part is required.

cmd/allocator/main.go
zmerlynn: Part of the point was to separate the contexts between the context for all the listeners (so they could shutdown) and all the "forward" contexts (workers, informers, etc.). The idea is to drain the listeners, then the forward context should just have nothing to do shortly.

zmerlynn: FWIW the right / more common way to handle this is just to have a channel instead of a separate context, but that doesn't fit the runner interface that everything uses..

pkg/util/httpserver/server.go
zmerlynn: I really didn't want to shadow a super commonly important base library. Calling this `http` would result in it getting imported as `agoneshttp` or something pretty often, since I would also probably not rename base `http`.

cmd/allocator/main.go
zmerlynn: I went ahead and separated the contexts and added a TODO to make this more obvious.

cmd/allocator/main.go
```diff
@@ -380,11 +400,22 @@ func runGRPC(h *serviceHandler, grpcPort int) {
 
 	grpcServer := grpc.NewServer(h.getGRPCServerOptions()...)
 	pb.RegisterAllocationServiceServer(grpcServer, h)
+	grpc_health_v1.RegisterHealthServer(grpcServer, grpcHealth)
 
 	go func() {
+		go func() {
+			<-ctx.Done()
+			grpcServer.Stop()
```
markmandel, line 408: Should this be: https://pkg.go.dev/google.golang.org/grpc#Server.GracefulStop (or maybe used in lieu of a lot of this code?)

Which might help with your multi-context issue?

cmd/allocator/main.go
```diff
@@ -356,21 +368,29 @@ func runHTTP(h *serviceHandler, httpPort int, handler http.Handler) {
 	}
 
 	go func() {
+		go func() {
+			<-ctx.Done()
+			_ = server.Close()
```
markmandel, line 373: I think you also want https://pkg.go.dev/net/http#Server.Shutdown here.

cmd/allocator/main.go
zmerlynn: I went back and forth on this, and I have a version in a branch where I did exactly that. It made things worse.

Let me just go back to the drawing board here, but I think it may involve rewriting a lot of the handling basically everywhere.

zmerlynn: But no, it doesn't solve the two-context issue. The problem is that here (and in httpserver/https), we use a context to signal shutdown, but the service handler still needs a forward context to get work done to shut down. Let me see if I can just rework the listen handlers more.

markmandel: Yeah I see your point. I think you are right that we'll want two contexts, but what we could do is something akin to:

```go
<- listenerCtx.Done()
// graceful shutdown both the http and grpc listeners. 
grpcServer.GracefulStop()
server.Shutdown()
cancelInfrormer() // cancel the worker + Informer context, since now we know we aren't accepting any more connections.
```

This might require pulling the grpc + http listeners into a larger struct that has a `Run(ctx)` and `Stop()` operation on it, to provide a larger data structure to operate on (and then it could take a context, and manage it's own context for the workers and informers).

Did that make sense?

zmerlynn: Hmm, we could do something similar to the runner slices and have shutdown slices, perhaps?

markmandel: > Hmm, we could do something similar to the runner slices and have shutdown slices, perhaps?

Seems like you are rebuilding Contexts 😄

cmd/allocator/main.go
```diff
@@ -294,51 +295,62 @@ func main() {
...
+	// Finally listen on 8080 (http), used to serve /live and /ready handlers for Kubernetes probes.
+	healthserver := httpserver.Server{Logger: logger}
+	healthserver.Handle("/", health)
+	go func() { _ = healthserver.Run(listenCtx, 0) }()
+
+	// TODO: This is messy. Contexts are the wrong way to handle this - we should be using shutdown,
+	// and a cascading graceful shutdown instead of multiple contexts and sleeps.
+	<-listenCtx.Done()
+	logger.Infof("Listen context cancelled")
+	time.Sleep(5 * time.Second)
```
markmandel, line 318: I'm assuming this is because we don't actually know that the gRPC/HTTP servers have gracefully shut down yet, so we just wait for a bit?

zmerlynn: Yeah .. fully knowing they're done takes a bit more plumbing.

markmandel (approved):
Good enough 😁 thanks for all the new plumbing!

PR #3846, opened 2024-05-31
Add helpful note to edit-first-gameserver-go

peterzhongyi:
Add a note in case people modified the game server main.go file again and want to see their new changes reflected in new game servers.

Also added some explanation on why the imagePullPolicy needs to be changed to Always.


site/content/en/docs/Getting Started/edit-first-gameserver-go.md
```diff
@@ -87,6 +87,22 @@ kubectl create -f gameserver.yaml
...
+  ```yaml
+  spec:
+  containers:
+  - name: simple-game-server
+    image: ${REGISTRY}/simple-game-server:${TAG}
+    imagePullPolicy: Always
+  ```
+
+Alternatively, you can also manually increment the `version` field in the `Makefile` file and change the `TAG` variable accordingly.
+However this is more cumbersome, and is not encouraged.
```
markmandel, line 103: Why not encouraged?

site/content/en/docs/Getting Started/edit-first-gameserver-go.md
```diff
@@ -87,6 +87,22 @@ kubectl create -f gameserver.yaml
 kubectl apply -f gameserver.yaml
 ```
 
+{{< alert title="Note" color="info">}}
+If you made changed `main.go` again and want to apply the changes to the new game servers, then you also need 
+  to modify the `gamerserver.yaml` file's `imagePullPolicy` to be `Always`, or else the new game servers might
+  only pull a cached version of the image, which doesn't have the new changes.
```
markmandel, line 93: I think a link to https://kubernetes.io/docs/concepts/containers/images/#updating-images would be good. 

Something like "See the documentation for ImagePullPolicy" for more details.

site/content/en/docs/Getting Started/edit-first-gameserver-go.md
peterzhongyi: Oh I guess it's just less preferred during development period because it is more cumbersome. Added some more explanations and changed the wording a bit.

site/content/en/docs/Getting Started/edit-first-gameserver-go.md
peterzhongyi: Done

site/content/en/docs/Getting Started/edit-first-gameserver-go.md
```diff
@@ -87,6 +87,26 @@ kubectl create -f gameserver.yaml
...
+  containers:
+  - name: simple-game-server
+    image: ${REGISTRY}/simple-game-server:${TAG}
+    imagePullPolicy: Always
+  ```
+
+Alternatively, you can also manually increment the `version` field in the `Makefile` file and change the `TAG`
+variable accordingly. However this is more cumbersome because you need to increment the `version` field each
+time you want to see a new change reflected in `main.go`, so it is usually less preferred during code development.
+However, after finishing all the planned changes, you should instead increment the `version` field and remove
```
peterzhongyi, line 106: Can anyone confirm that this is indeed the common practice that we increment the version field after a patch to the game server code?

zmerlynn: Yes, it's common practice to change the version as part of the PR. We do this for `simple-game-server`, for example.

site/content/en/docs/Getting Started/edit-first-gameserver-go.md
```diff
@@ -87,6 +87,26 @@ kubectl create -f gameserver.yaml
 kubectl apply -f gameserver.yaml
 ```
 
+{{< alert title="Note" color="info">}}
+If you changed `main.go` again and want to apply the changes to the new game servers, then you also need 
+to modify the `gamerserver.yaml` file's `imagePullPolicy` to be `Always`, or else the new game servers might
```
zmerlynn, line 92: It's not the game servers pulling a new image, it's `kubelet`, or just say the node.

Something like: "or the node may use a cached copy of the image, which doesn't have the new changes."

site/content/en/docs/Getting Started/edit-first-gameserver-go.md
```diff
@@ -87,6 +87,26 @@ kubectl create -f gameserver.yaml
...
+only pull a cached version of the image, which doesn't have the new changes. See the documentation for
+[ImagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#updating-images) for more details.
```
zmerlynn, line 94: To shorten this a bit, I would just link `imagePullPolicy` above and anyone curious will know to go look at the hyperlink, which would let you drop this sentence.

site/content/en/docs/Getting Started/edit-first-gameserver-go.md
```diff
@@ -87,6 +87,26 @@ kubectl create -f gameserver.yaml
...
+  spec:
+  containers:
+  - name: simple-game-server
+    image: ${REGISTRY}/simple-game-server:${TAG}
+    imagePullPolicy: Always
+  ```
+
+Alternatively, you can also manually increment the `version` field in the `Makefile` file and change the `TAG`
+variable accordingly. However this is more cumbersome because you need to increment the `version` field each
+time you want to see a new change reflected in `main.go`, so it is usually less preferred during code development.
```
zmerlynn, line 105: I know you're trying to be helpful but I think line 104-107, after `accordingly.` can be dropped, I.e. you can present both and let people decide which toil they like better.

This is a common Kubernetes dev flow thing, so I think concision here is a little better - the note is a good reminder but a lot of people coming to the page will already be familiar with this gotcha.

zmerlynn (approved):

Issue #3848, opened 2024-05-31
Add runAsUser, runAsGroup, and allowPrivilegeEscalation to helm chart for Agones containers


zmerlynn:
**Is your feature request related to a problem? Please describe.**

Allow `securityContext` fields such as `runAsUser`, `runAsGroup`, and `allowPrivilegeEscalation` to be set on `agones-{allocator,extensions,controller,ping,sidecar}`

https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

**Describe the solution you'd like**
Helm chart change

**Describe alternatives you've considered**
You can do this with `kustomize`.

markmandel:
Not very familiar with these fields - but does it need to be configurable, or is this something we could set ourselves and leave on always?

On Agones we set the user and group, so that probably shouldn't be configurable? And I can't think of a situation where we'd want an Agones Pod to enable `allowPrivilegeEscalation`?

ThatDevopsGuy:
@markmandel - I can't think of a reason why a game server would require a _specific_ UID/GID. I could imagine that some clients might want to bind to a privileged port to offer a some sort of file serving in-client (e.g. downloading a missing asset), however. Perhaps it's best to let it be easily configurable with overrides, should they be needed, but the defaults comply with existing K8S security audits.

jharris-:
Should be good to add `runAsNonRoot` to avoid needing to set the UID/GID.
In addition to the initial request, could you add the `agones-gameserver-sidecar` to the scope of work?

zmerlynn:
@jharris- Sure, sounds good!

markmandel:
> @markmandel - I can't think of a reason why a game server would require a _specific_ UID/GID. 

This ticket isn't for `GameServer` pods though - you can set whatever you like on there via the `GameServer.Spec.Template` -- this is for the Agones components themselves.

I'm trying to work out why they should be configurable. It seems like they should be set to optimal values, whatever that may be.

ThatDevopsGuy:
> I'm trying to work out why they should be configurable. It seems like they should be set to optimal values, whatever that may be.

Another way to put it might be, is there any reason why `runAsNonRoot` wouldn't work? I think that might be all we need across the board, so the specific request might be better expressed as such.

markmandel:
> Another way to put it might be, is there any reason why runAsNonRoot wouldn't work? I think that might be all we need across the board, so the specific request might be better expressed as such.

I believe we are in furious agreement 😃

[...]

PR #3856, opened 2024-06-05
Add security context to Agones containers

peterzhongyi:
Added runAsNonRoot(true) and allowPrivilegeEscalation(false) into the containers field of controller, extensions, ping, and allocator.

Left out sidecar because it is generated by the controller and require another set of changes.

Towards #3848

### DIFF
--- a/install/helm/agones/templates/service/allocation.yaml
+++ b/install/helm/agones/templates/service/allocation.yaml
@@ -231,7 +231,7 @@ spec:
         imagePullPolicy: {{ .Values.agones.image.allocator.pullPolicy }}
         securityContext:
           runAsNonRoot: true
-          runAsUser: 1003
+          runAsUser: 1000
           allowPrivilegeEscalation: false
         livenessProbe:
           httpGet:
--- a/install/helm/agones/templates/service/allocation.yaml
+++ b/install/helm/agones/templates/service/allocation.yaml
@@ -229,6 +229,10 @@ spec:
       - name: agones-allocator
         image: "{{ .Values.agones.image.registry }}/{{ .Values.agones.image.allocator.name}}:{{ default .Values.agones.image.tag .Values.agones.image.allocator.tag }}"
         imagePullPolicy: {{ .Values.agones.image.allocator.pullPolicy }}
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 1003
+          allowPrivilegeEscalation: false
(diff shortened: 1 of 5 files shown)


zmerlynn:
This LGTM but I'd like to see it pass CI before I approve

install/helm/agones/templates/service/allocation.yaml
```diff
@@ -229,6 +229,10 @@ spec:
       - name: agones-allocator
         image: "{{ .Values.agones.image.registry }}/{{ .Values.agones.image.allocator.name}}:{{ default .Values.agones.image.tag .Values.agones.image.allocator.tag }}"
         imagePullPolicy: {{ .Values.agones.image.allocator.pullPolicy }}
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 1003
```
markmandel, line 234: Curious why this is a different userid on each container? I would have thought they would be the same?

markmandel: Also, do we need to specify the uid at all?  That seems brittle. or will it work with just `runAsNonRoot` ?

peterzhongyi: Ahh I thought each pod needs to have an unique ID across the cluster (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#:~:text=Every%20Kubernetes%20object%20also%20has,are%20each%20named%20myapp%2D1234%20.) but apparently it is referring to some other UID because I tried setting the runAsUser 1000 for all 4 binaries and they can turn up just fine.

Yes we need to specify the UID otherwise the pods can't turn up, and the error message would be `Error: container has runAsNonRoot and image has non-numeric user (appuser), cannot verify user is non-root`

https://stackoverflow.com/questions/49720308/kubernetes-podsecuritypolicy-set-to-runasnonroot-container-has-runasnonroot-and for reference. Actually I also added a comment on the pr to explain this:)

zmerlynn: I don't think we should pick a uid out of the blue. I was hoping this would just work with `runAsNonRoot` since the image specifies a username, but that appears to be wrong, as you discovered.

If we have to specify an arbitrary uid, I think it should be configurable by helm, in case someone has a weird identity setup. I'd just default it to e.g. 1000.

markmandel (approved):

PR #3869, opened 2024-06-13
Add Security Context to game server sidecar

### LINKED ISSUE #3848: Add runAsUser, runAsGroup, and allowPrivilegeEscalation to helm chart for Agones containers

peterzhongyi:
Add security context to game server sidecar. Hard coded default values in the template, like in #3856. We can make them configurable if we find a reason to.

Fixes #3848

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -44,7 +44,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
-	"k8s.io/utils/pointer"
 
 	"agones.dev/agones/pkg"
 	"agones.dev/agones/pkg/client/clientset/versioned"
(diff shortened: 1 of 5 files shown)


cmd/controller/main.go
```diff
@@ -59,33 +60,36 @@ import (
...
+	enableStackdriverMetricsFlag         = "stackdriver-exporter"
+	stackdriverLabels                    = "stackdriver-labels"
+	enablePrometheusMetricsFlag          = "prometheus-exporter"
+	projectIDFlag                        = "gcp-project-id"
+	sidecarImageFlag                     = "sidecar-image"
+	sidecarCPURequestFlag                = "sidecar-cpu-request"
+	sidecarCPULimitFlag                  = "sidecar-cpu-limit"
+	sidecarMemoryRequestFlag             = "sidecar-memory-request"
+	sidecarMemoryLimitFlag               = "sidecar-memory-limit"
+	sidecarRunAsNonRootFlag              = "sidecar-run-as-nonroot"
```
zmerlynn, line 72: You went all out on the flags! I don't think we need flags for `RunAsNonRoot` or `AllowPrivilegeEscalation`, those can be hardcoded. We don't want the user to change them, IMO.

cmd/controller/main.go
```diff
@@ -59,33 +60,36 @@ import (
...
+	sidecarRunAsUserFlag                 = "sidercar-run-as-user"
```
zmerlynn, line 73: typo: `sidercar`

install/helm/agones/templates/controller.yaml
```diff
@@ -134,6 +134,12 @@ spec:
           value: {{ .Values.agones.image.sdk.memoryRequest | quote }}
         - name: SIDECAR_MEMORY_LIMIT
           value: {{ .Values.agones.image.sdk.memoryLimit | quote }}
+        - name: SIDECAR_RUN_AS_NONROOT
+          value: "true"
+        - name: SIDECAR_RUN_AS_USER
+          value: "1000"
```
zmerlynn, line 140: I suspect this value is getting ignored given the typo above.

pkg/gameservers/controller.go
```diff
@@ -83,6 +83,7 @@ type Controller struct {
 	sidecarCPULimit        resource.Quantity
 	sidecarMemoryRequest   resource.Quantity
 	sidecarMemoryLimit     resource.Quantity
+	sidecarSecurityCtx     *corev1.SecurityContext
```
zmerlynn, line 86: If you just drop to the user, this will be simpler, too.

zmerlynn (changes requested):
Nice! Almost there.

cmd/controller/main.go
peterzhongyi: Agreed! I've removed these 2 flags.

cmd/controller/main.go
peterzhongyi: Done

install/helm/agones/templates/controller.yaml
peterzhongyi: Yeah true

pkg/gameservers/controller.go
peterzhongyi: True

cmd/controller/main.go
```diff
@@ -284,6 +286,7 @@ func parseEnvFlags() config {
 	pflag.String(sidecarCPURequestFlag, viper.GetString(sidecarCPURequestFlag), "Flag to overwrite the GameServer sidecar container's cpu request. Can also use SIDECAR_CPU_REQUEST env variable")
 	pflag.String(sidecarMemoryLimitFlag, viper.GetString(sidecarMemoryLimitFlag), "Flag to overwrite the GameServer sidecar container's memory limit. Can also use SIDECAR_MEMORY_LIMIT env variable")
 	pflag.String(sidecarMemoryRequestFlag, viper.GetString(sidecarMemoryRequestFlag), "Flag to overwrite the GameServer sidecar container's memory request. Can also use SIDECAR_MEMORY_REQUEST env variable")
+	pflag.Int32(sidecarRunAsUserFlag, 1000, "Flag to indicate the GameServer sidecar container's UID. Can also use SIDECAR_RUN_AS_USER env variable")
```
peterzhongyi, line 289: Actually I don't really understand how this line works, wherever a flag is an int, the existing code always puts an integer as the second argument instead of using viper.GetInt32(). I thought you need to call viper in order to retrieve the value in the template controller.yaml file?

zmerlynn: The `controller.yaml` file is specifying how the container is built. In our case it runs a container with various environment variables, including `SIDECAR_RUN_AS_USER`, which gets bound below (line 321). We actually never use the flags, just the env variables - but we have the flags here because people do deploy Agones other ways.

zmerlynn, line 289: This should be `viper.GetInt32(sidecarRunAsUserFlag)`, you already set the default on line 264 above. (You'll note the surrounding code just pulls the default out as well.)

pkg/gameservers/controller_test.go
```diff
@@ -1306,6 +1307,9 @@ func TestControllerCreateGameServerPod(t *testing.T) {
 			assert.Equal(t, "FEATURE_GATES", sidecarContainer.Env[2].Name)
 			assert.Equal(t, "LOG_LEVEL", sidecarContainer.Env[3].Name)
 			assert.Equal(t, string(fixture.Spec.SdkServer.LogLevel), sidecarContainer.Env[3].Value)
+			assert.Equal(t, *sidecarContainer.SecurityContext.AllowPrivilegeEscalation, false)
+			assert.Equal(t, *sidecarContainer.SecurityContext.RunAsNonRoot, true)
+			assert.Equal(t, *sidecarContainer.SecurityContext.RunAsUser, int64(1000))
```
zmerlynn, line 1312: you made a const for this on line 57, why not use it?

zmerlynn (commented):
more nits but really close

pkg/gameservers/controller_test.go
peterzhongyi: Done

zmerlynn (approved):
Nice job, thanks!

cmd/controller/main.go
peterzhongyi: Done. But why do all the other occurrences with integers all just have a number as the second variable? Like the minPortFlag and maxPortFlag down below, why don't they use viper.GetInt32()? Maybe because they are not supposed to be configurable?

zmerlynn: Huh, I didn't even notice that. I'm going to say there's just some inconsistencies here. We should be setting the default in one place and getting it, though.

PR #3871, opened 2024-06-14
featuregate: adds validation if PortPolicyNone is not enabled

### LINKED ISSUE #3867: Using portPolicy None results in a gameserver's port being 0
**What happened**:

I attempted to use the new feature in v1.41.0 to set my `portPolicy` to None and when my gameserver started, the gameserver's port in its status was set to 0. My gameserver code was written to take the Address and Port from the gameserver's status and use them for connections. When the port is 0, this failed.

My expectation is perhaps that when `portPolicy` is set to None, the Port field of the gameserver's status would become equal to the `containerPort` value. In other words, to connect to a gameserver Pod with a None `portPolicy`, I would use the PodIP and the `containerPort` to connect. 

Use any gameserver spec and set its `portPolicy` to None.

```
portPolicy:
  - name: game
    portPolicy: None
    protocol: UDP
    containerPort: 1234
```

Check the status once the gameserver is running:

image

**Environment**:
- Agones version: v1.41.0
- Kubernetes version (use `kubectl version`): v1.30.0-eks-036c24b
- Others:

daniellee:
When the feature gate PortPolicyNone is used, validation should fail if the PortPolicy field is set to the value None.

**What type of PR is this?**
/kind bug

Without this validation, then the Container Port is set to 0 and it silently fails.

Closes #3867

### DIFF
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -528,6 +528,14 @@ func (gss *GameServerSpec) validateFeatureGates(fldPath *field.Path) field.Error
 		}
 	}
 
+	if !runtime.FeatureEnabled(runtime.FeaturePortPolicyNone) {
+		for i, p := range gss.Ports {
+			if p.PortPolicy == None {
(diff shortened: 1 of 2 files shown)


igooch (approved):
LGTM

PR #3872, opened 2024-06-15
Buffer Unity SDK ReceiveData when watching for configuration changes

ZeroParticle:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Currently the Unity SDK watches for changes to the GameServer configuration by processing incoming data using `ReceiveData` on `DownloadHandlerScript`. This function is not guaranteed to be passed a single, complete message, so it is possible for the JSON parse to fail. Though extremely rare, I have run into this problem a few times while doing load tests.

No associated issue, but I can open one if needed.

The Kubernetes JS client does something similar to this using the `byline` package: https://github.com/kubernetes-client/javascript/blob/master/src/watch.ts#L116

In my testing it appears that creating messages based on newline characters is enough, so I implemented only that case.

Additional documentation on the Unity ReceiveData function: https://docs.unity3d.com/2021.3/Documentation/ScriptReference/Networking.DownloadHandler.ReceiveData.html

### DIFF
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -360,17 +360,37 @@ namespace Agones
         private class GameServerHandler : DownloadHandlerScript
         {
             private WatchGameServerCallback callback;
+            private StringBuilder stringBuilder;
+
             public GameServerHandler(WatchGameServerCallback callback)
             {
                 this.callback = callback;
+                this.stringBuilder = new StringBuilder();
             }
 
             protected override bool ReceiveData(byte[] data, int dataLength)
             {
-                string json = Encoding.UTF8.GetString(data);
-                var dictionary = (Dictionary<string, object>) Json.Deserialize(json);
-                var gameServer = new GameServer(dictionary["result"] as Dictionary<string, object>);
-                this.callback(gameServer);
+                string dataString = Encoding.UTF8.GetString(data);
+                this.stringBuilder.Append(dataString);
+
+                string bufferString = stringBuilder.ToString();
+                int newlineIndex;
+
+                while ((newlineIndex = bufferString.IndexOf('\n')) >= 0)
+                {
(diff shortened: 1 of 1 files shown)


igooch:
Do you have someone on your team that can help out with reviewing this code? Our regular maintainers aren't as familiar with the Unity SDK.

liamlutton (approved):

sdks/unity/AgonesSdk.cs
```diff
@@ -360,17 +360,37 @@ private void OnRequestCompleted(AsyncOperation _)
...
+
+                string bufferString = stringBuilder.ToString();
+                int newlineIndex;
+
+                while ((newlineIndex = bufferString.IndexOf('\n')) >= 0)
+                {
+                    string fullLine = bufferString.Substring(0, newlineIndex);
+                    try
+                    {
+                        var dictionary = (Dictionary<string, object>) Json.Deserialize(fullLine);
```
igooch, line 384: What does the input look like for this method? Is all the data expected to be on one line?

igooch (commented):
It would be nice to have at least a unit test for this. Although I'm not seeing any tests for the Unity SDK, which is... odd.

sdks/unity/AgonesSdk.cs
ZeroParticle: fullLine should always be a single JSON message on a single line. grpc-gateway turns streams into newline delineated JSON messages and JSON encoding will escape any newlines in the data, so the only newlines we encounter will be the ones intentionally sent to break full JSON messages.

igooch (approved):
Approving with the caveat that I don't have a way of testing this. LGTM per [Liam Lutton](https://github.com/liamlutton) and [Jack Schaffner](https://github.com/ZeroParticle).

PR #3877, opened 2024-06-18
added condition for distributed logic

### LINKED ISSUE #3862: Distributed defaults to using ListSortedGameServersPriorities when CountsAndLists is enabled
**What happened**:

Counters and Lists moved to Beta, so it's now on by default. If the scheduling type is set to `Distributed` it will execute the logic for CountersAndLists instead of the logic for `Distributed`. https://github.com/googleforgames/agones/blob/aa390f9f83564b04db9abd6dc985ca106cfa286f/pkg/gameserverallocations/find.go#L54

If there are Counters and Lists Priorities https://github.com/googleforgames/agones/blob/aa390f9f83564b04db9abd6dc985ca106cfa286f/pkg/apis/agones/v1/fleet.go#L83, the allocator should use the Counters and Lists logic. Otherwise, the Allocator should use the `Distributed` logic.

ashutosji:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**Which issue(s) this PR fixes**:https://github.com/googleforgames/agones/issues/3862

Closes #3862 

### DIFF
--- a/pkg/gameserverallocations/find.go
+++ b/pkg/gameserverallocations/find.go
@@ -40,9 +40,6 @@ func findGameServerForAllocation(gsa *allocationv1.GameServerAllocation, list []
 
 	var loop func(list []*agonesv1.GameServer, f func(i int, gs *agonesv1.GameServer))
 
-	// Determine if Priorities are set in the Fleet spec
-	prioritiesSet := len(gsa.Spec.Priorities) > 0
-
 	// packed is forward looping, distributed is random looping
 	switch gsa.Spec.Scheduling {
 	case apis.Packed:
@@ -52,6 +49,9 @@ func findGameServerForAllocation(gsa *allocationv1.GameServerAllocation, list []
 			}
 		}
 	case apis.Distributed:
+		// Determine if Priorities are set in the Fleet spec
+		prioritiesSet := len(gsa.Spec.Priorities) > 0
+
 		// randomised looping - make a list of indices, and then randomise them
 		// as we don't want to change the order of the gameserver slice
 		if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) || !prioritiesSet {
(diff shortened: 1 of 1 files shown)


pkg/gameserverallocations/find.go
```diff
@@ -40,6 +40,9 @@ func findGameServerForAllocation(gsa *allocationv1.GameServerAllocation, list []
 
 	var loop func(list []*agonesv1.GameServer, f func(i int, gs *agonesv1.GameServer))
 
+	// Determine if Priorities are set in the Fleet spec
+	prioritiesSet := len(gsa.Spec.Priorities) > 0
```
igooch, line 44: This can be moved to inside the `case apis.Distributed:`

igooch (commented):
Looks good, just one nit.

pkg/gameserverallocations/find.go
```diff
@@ -49,9 +49,12 @@ func findGameServerForAllocation(gsa *allocationv1.GameServerAllocation, list []
 			}
 		}
 	case apis.Distributed:
+		// Determine if Priorities are set in the Fleet spec
+		prioritiesSet := len(gsa.Spec.Priorities) > 0
+
 		// randomised looping - make a list of indices, and then randomise them
 		// as we don't want to change the order of the gameserver slice
-		if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) {
+		if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) || !prioritiesSet {
```
markmandel, line 57:
```suggestion
		if !runtime.FeatureEnabled(runtime.FeatureCountsAndLists) || len(gsa.Spec.Priorities) == 0 {
```

Small thing, but probably doesn't need a variable? WDYT @igooch ?

igooch: No strong feelings either way. It's probably not needed.

markmandel: @ashutosji bump on removing this variable, and just moving the check inline, and then I think this is good to merge.

ashutosji: > @ashutosji bump on removing this variable, and just moving the check inline, and then I think this is good to merge.

Thanks! Moved the check inline.

igooch (approved):
LGTM

PR #3880, opened 2024-06-24
feat: Add CRD Changes and Feature Flag for chain policy

### LINKED ISSUE #3008: Feature Request: Scheduled Autoscalers
> [!NOTE] 
> If you want to skip a bunch of discussion and see our @indexjoseph's proposed design, see https://github.com/googleforgames/agones/issues/3008#issuecomment-2174571364

During a scheduled in-game events or new version releases, we see pretty rapid spikes in usage of either an already high use fleet or a newer unused fleet. Both of these events we know the timing of, and our current options are either:
1. Prescale agressively: this works but means that unless we are building scheduling logic ourselves to undo the additional scale afterwards, we're paying for a lot of unused capacity.
2. Webhook autoscaler: this is a viable solution, but requires us to build a service to do so.

Introduce the concept of scheduled overrides that contain the following:
1. a start time(in UTC)
1. an end time(in UTC)
1. a priority int(higher the better, much like PriorityClasses)
1. a buffer autoscaler block

Then on autoscaling evaluation:
1. collect those overrides for which we are between the start and end time
a. if there are no matching overrides, just use the default autoscaling rule 
1. of those select the highest priority
1. apply that buffer autoscaling rule instead of the default

This would allow us to set special scaling windows for events or new version releases. A further extension could be to allow recurring windows to do time of day scheduling so that we could have a buffer window in the off hours and a percentage during higher usage, which could help with issues like that described in #2504 

As described at the top, we can either prescale agressively, which either results in us adjusting the autoscaler directly, or using the webhook autoscaler.

indexjoseph:
**What type of PR is this?**

/kind feature

New Chain Policy for Scheduled Autoscaling, see this comment for more info: [comment](https://github.com/googleforgames/agones/issues/3008#issuecomment-2174571364)

Towards #3008


zmerlynn:
A couple of things:
1. This should be under a feature gate, including the CRD addition (the refactor portion should be fine, but the new additions should be gated). See here for how to add a gate: https://github.com/googleforgames/agones/blob/main/pkg/util/runtime/features.go#L85 .. but also, you can gate helm values, see https://github.com/googleforgames/agones/blob/1f6b2a0cf2f64c97e6be5f5d0aa65dbcfa2734c4/install/helm/agones/templates/controller.yaml#L118-L119 for a recent example. Doing all of this under a dev feature gate will allow you to send smaller PRs without it all being ready for testing yet.
1. Split this up into smaller PRs - feel free to submit the validation separate from the CRD itself, since they're both big hunks to review.

install/helm/agones/defaultfeaturegates.yaml
```diff
@@ -27,7 +27,8 @@ RollingUpdateFix: false
 PortRanges: false
 PortPolicyNone: false
 
-# Dev features
+# Dev features`
```
vicentefb, line 30: super nit: `# Dev features`

examples/chainfleetautoscaler.yaml
```diff
@@ -0,0 +1,93 @@
+---
+# Copyright 2023 Google LLC All Rights Reserved.
```
zmerlynn, line 2: nit: It's 2024 :)

install/helm/agones/templates/crds/_fleetautoscalerpolicy.yaml
```diff
@@ -0,0 +1,237 @@
+# Copyright 2020 Google LLC All Rights Reserved.
```
zmerlynn, line 1: nit: Still 2024 :)

install/helm/agones/templates/crds/_fleetautoscalerpolicy.yaml
```diff
@@ -0,0 +1,237 @@
...
+        maxCapacity:  # Maximum aggregate list capacity that can be provided by this FleetAutoscaler. Required.
+          type: integer
+          minimum: 1
+        bufferSize:  # Size of a buffer based on the list capacity that is available over the current aggregate list length in the Fleet. It can be specified either in absolute (i.e. 5) or percentage format (i.e. 5%).
+          x-kubernetes-int-or-string: true
+          anyOf:
+            - type: integer
+            - type: string
+    {{- $featureGates := include "agones.featureGates" . | fromYaml }}
+    {{- if not $featureGates.ScheduledAutoscaler }}
```
zmerlynn, line 114: Shouldn't this be `if` and not `if not`?

install/yaml/install.yaml
```diff
@@ -5767,6 +5768,127 @@ spec:
                           anyOf:
                             - type: integer
                             - type: string
+                    chain:
```
zmerlynn, line 5771: I would not expect this to be in the `install.yaml` yet. You'll have to do your tests by enabling the feature gate - you can use the `FEATURE_GATES` variable to turn it on for `make install`: https://github.com/googleforgames/agones/blob/0f7c9436ed324bf1bf44dd92fe0eb32e3c621ddb/build/Makefile#L430

zmerlynn (changes requested):

install/helm/agones/templates/crds/_fleetautoscalerpolicy.yaml
indexjoseph: Yep, thanks.

install/helm/agones/templates/crds/_fleetautoscalerpolicy.yaml
```diff
@@ -0,0 +1,237 @@
...
+                    type: string
+                  end: # Defines when to stop evaluating the active period.
+                    type: string
+                activePeriod:
+                  type: object
+                  startCron: # Cron expression defining when to start applying the policy.
+                    type: string
+                  duration: # The length of time the policy should be applied for.
+                    type: string
+            policy: # Defines which policy to apply during the active period. Required.
```
markmandel, line 143: Rather than copy-paste, let's make it a nested include.

There was a good example in https://github.com/googleforgames/agones/pull/3718 which didn't get merged that meant it didn't get recursive forever.

install/helm/agones/templates/crds/_fleetautoscalerpolicy.yaml
```diff
@@ -0,0 +1,237 @@
...
+          minimum: 0
+        maxCapacity:  # Maximum aggregate list capacity that can be provided by this FleetAutoscaler. Required.
+          type: integer
+          minimum: 1
+        bufferSize:  # Size of a buffer based on the list capacity that is available over the current aggregate list length in the Fleet. It can be specified either in absolute (i.e. 5) or percentage format (i.e. 5%).
+          x-kubernetes-int-or-string: true
+          anyOf:
+            - type: integer
+            - type: string
+    {{- $featureGates := include "agones.featureGates" . | fromYaml }}
```
markmandel, line 113: (non blocking, just curious)  We've not actually done this before 🤔  - @zmerlynn is this for in-place Agones updates?

zmerlynn: We've done the helm-based feature gates before - do you mean we haven't done the conditional-CRD piece?

I thought it might be a good idea - then we can keep things under the feature gate until the API is "locked". Do you think it's weird?

markmandel: > - do you mean we haven't done the conditional-CRD piece?

Yep that. It's not a bad idea - can't think of anything pro/con 😄 

It does automatically mean that data is removed from the CRD if you turn off the feature flag, which is not a bad thing, and I believe is inline with what @igooch is designing, no?

install/yaml/install.yaml
indexjoseph: Since this is a dev feature, does it matter if I add the alpha feature gates variable in the Makefile (for testing purposes) ?

zmerlynn:
Don't commit the change to add it to the ALPHA_FEATURE_GATES until you're
ready for it to be tested in CI - that's the main purpose of the "dev"
gate. If you think each commit you add will be testable with the gate
on/off, you can go ahead and add it and we'll document it later, but the
point of the dev gate is to allow things to be a bit broken until you're
ready.

You can certainly add it locally and then use stash or whatever to strip it
before you push. That said, though, I would just get used to the syntax for
`make FEATURES_GATES=<blah> install`, it's a lot easier than having to
avoid committing it.

install/yaml/install.yaml
```diff
@@ -5672,7 +5672,7 @@ spec:
                   type: string
                   minLength: 1
                   maxLength: 63
-                  pattern: "^[a-z0-9]([-\\.a-z0-9]*[a-z0-9])?$"
+                  pattern: "^[a-z0-9]([-\\.a-z0-9]*[a-z0-9])?$"
```
vicentefb, line 5675: Is this change needed or is this just some extra spacing ?

zmerlynn: It's an artifact of the `{{- include` - not clear how easy it is to remove but I've seen it in other cases. (I'm happy to ignore it.)

zmerlynn (approved):
LGTM! @markmandel want a go?

install/yaml/install.yaml
vicentefb: okok sounds good!

examples/chainfleetautoscaler.yaml
```diff
@@ -0,0 +1,100 @@
...
+            end: "2024-02-23T16:04:04-05:00"
+          activePeriod:
+            # Timezone to be used for the startCron field. Optional.
+            timezone: "America/New_York"
+            # Start applying the bufferSize everyday at 1:00  AM EST.
+            # (Only eligible starting on Feb 20, 2024 at 4:04 PM.) Optional.
+            startCron: "0 1 * * 0"
+            # Only apply the bufferSize for this 5 hours
+            # Optional.
+            duration: "5h"
```
markmandel, line 53: Nit: What does `duration` default to?

indexjoseph: It will default to always/indefinite.

examples/chainfleetautoscaler.yaml
```diff
@@ -0,0 +1,100 @@
...
+kind: FleetAutoscaler
+metadata:
+  name: chain-fleet-autoscaler
+spec:
+  policy:
+    # Chain based policy for autoscaling.
+    type: Chain
+    chain:
+      # UID of chain entry. Optional.
+      - uid: "weekday"
```
markmandel, line 35:
```suggestion
      - id: "weekday"
```

it's an `id` I don't think it needs a `u`.

Or `name` if we have prior art using `name` in a list item.

examples/chainfleetautoscaler.yaml
```diff
@@ -0,0 +1,100 @@
...
+        schedule:
+          between:
+            # The policy becomes eligible for application starting on
+            # Feb 20, 2024 at 4:04 PM EST. Optional.
+            start: "2024-02-20T16:04:04-05:00"
+            # The policy becomes ineligible for application on
+            # Feb 23, 2024 at 4:04 PM EST. Optional.
+            end: "2024-02-23T16:04:04-05:00"
+          activePeriod:
+            # Timezone to be used for the startCron field. Optional.
```
zmerlynn, line 46: `Defaults to UTC.` instead of `Optional.`

zmerlynn (commented):
If there's a second or redundant comment, I'm sorry - I see I have two review comments but can only see one.

examples/chainfleetautoscaler.yaml
```diff
@@ -0,0 +1,101 @@
...
+          between:
+            # The policy becomes eligible for application starting on
+            # Feb 20, 2024 at 4:04 PM EST. Optional.
+            start: "2024-02-20T16:04:04-05:00"
+            # The policy becomes ineligible for application on
+            # Feb 23, 2024 at 4:04 PM EST. Optional.
+            end: "2024-02-23T16:04:04-05:00"
+          activePeriod:
+            # Timezone to be used for the startCron field.
+            # Optional.
```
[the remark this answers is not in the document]
zmerlynn: (Deleting, redundant)

zmerlynn (approved):
Awesome!

PR #3889, opened 2024-07-01
Added cleanup for app-engine services in cloudbuild

### LINKED ISSUE #3860: Add AppEngine cleanup to appropriate CloudBuild scripts.
Behind the scenes we do the following on a VM:

Have a crontab that has (truncated):

```cron
# m h  dom mon dow   command
  0 1  *   *   *     /home/markmandel/build/cleanup/cleanup-appengine.sh preview 30
  0 2  *   *   *     /home/markmandel/build/cleanup/cleanup-appengine.sh default 10
  0 3  *   *   *     /home/markmandel/build/cleanup/cleanup-appengine.sh development 10
```

and `cleanup-appengine.sh` has the following content:

```shell
#!/bin/bash
set -o xtrace

gcloud app versions list --service=$1 --filter="traffic_split<1" --sort-by=~last_deployed_time.datetime --format="table[no-heading](version.id)" | tail -n +$2 | xargs gcloud app versions delete --service=$1
```

Rather than doing this in a random VM, let's do this in our cloudbuild scripts for each of the AppEngine services outlined above.

Implementation thought: I'd suggest just outputting what would be deleted (dry run essentially) for a short period rather than delete. so we can review if it's working correctly before setting it to actually delete items.

* https://github.com/googleforgames/agones/blob/main/cloudbuild.yaml - `preview` service cleanup should go here.
* https://github.com/googleforgames/agones/blob/main/site/cloudbuild.yaml - `development` and `default` service cleanup should go here.

kamaljeeti:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**Which issue(s) this PR fixes**:https://github.com/googleforgames/agones/issues/3860

Closes #3860 

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -379,7 +379,7 @@ steps:
   #
   - name: gcr.io/cloud-builders/gcloud
     id: cleanup-services
-    waitFor: [htmltest-save-cache, rust-build-save-cache, htmltest-save-cache, tests]
+    waitFor: [htmltest-save-cache, rust-build-save-cache, tests, build-images]
     entrypoint: bash
     args:
       - -c
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -374,6 +374,23 @@ steps:
     dir: cmd/sdk-server/bin
     args: [-m, cp, '*.zip', gs://agones-artifacts/sdk-server]
 
+  #
+  # Cleanup services "preview", "development", "default"
+  #
+  - name: gcr.io/cloud-builders/gcloud
+    id: cleanup-services
+    waitFor: [htmltest-save-cache, rust-build-save-cache, htmltest-save-cache, tests]
+    entrypoint: bash
+    args:
+    - -c
+    - |
+      set -e
+      set -o pipefail
+      for service in preview development default; do
+       echo "Cleaning up $service service..."
+        gcloud app versions list --service=$service --filter="traffic_split<1" --sort-by=~last_deployed_time.datetime --format="table[no-heading](version.id)"
+      done
+
 substitutions:
   _CACHE_BUCKET: agones-build-cache
   _HTMLTEST_CACHE_KEY: htmltest-0.10.1
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -382,14 +382,14 @@ steps:
     waitFor: [htmltest-save-cache, rust-build-save-cache, htmltest-save-cache, tests]
     entrypoint: bash
     args:
-    - -c
-    - |
-      set -e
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -374,6 +374,23 @@ steps:
     dir: cmd/sdk-server/bin
     args: [-m, cp, '*.zip', gs://agones-artifacts/sdk-server]
 
+  #
+  # Cleanup services "preview", "development", "default"
+  #
+  - name: gcr.io/cloud-builders/gcloud
+    id: cleanup-services
+    waitFor: [htmltest-save-cache, rust-build-save-cache, tests, build-images]
```
zmerlynn, line 382: You frankly don't need to wait for anything here, it just needs to run...

cloudbuild.yaml
```diff
@@ -374,6 +374,23 @@ steps:
...
+    entrypoint: bash
+    args:
+      - -c
+      - |
+        set -e
+        set -o pipefail
+        for service in preview development default; do
+        echo "Cleaning up $service service..."
+        gcloud app versions list --service=$service --filter="traffic_split<1" --sort-by=~last_deployed_time.datetime --format="table[no-heading](version.id)"
```
zmerlynn, line 391: two things:
* please add the `tail` from the original so we can see what would be deleted. As it is now it's just outputting everything (in sorted order).
* please indent 390/391

cloudbuild.yaml
```diff
@@ -374,6 +374,23 @@ steps:
     dir: cmd/sdk-server/bin
     args: [-m, cp, '*.zip', gs://agones-artifacts/sdk-server]
 
+  #
+  # Cleanup services "preview", "development", "default"
+  #
+  - name: gcr.io/cloud-builders/gcloud
```
zmerlynn, line 380: I would add `allowFailure: true` here. On occasion we're going to have a race where two builds run and try to delete the same previous version, but it doesn't matter. Today we run this on an instance and probably never see logs - adding `allowFailure: true` will at least mean we see that a step failed, but not fail the whole build.

zmerlynn (changes requested):

cloudbuild.yaml
zmerlynn: Actually, if you wait for `e2e-wait-to-become-leader` here instead, it will mean that only one PR build will attempt to delete app versions at the same time, which might obviate my comment below.

kamaljeeti:
Hi @zmerlynn , closing this PR as the changes have been refined based on feedback. Opening a new PR#3890  for further review.

PR #3890, opened 2024-07-03
Added cleanup for app-engine services in cloudbuild script

### LINKED ISSUE #3860: Add AppEngine cleanup to appropriate CloudBuild scripts.

kamaljeeti:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**Which issue(s) this PR fixes**: https://github.com/googleforgames/agones/issues/3860

Closes #3860 


cloudbuild.yaml
```diff
@@ -374,6 +374,30 @@ steps:
...
+    args:
+      - -c
+      - |
+        set -e
+        set -o pipefail
+        for service in preview development default; do
+          echo "Cleaning up $service service..."
+          #Retrieve version ids
+          versions=$(gcloud app versions list --service=$service --filter="traffic_split<1" --sort-by=~last_deployed_time.datetime --format="table[no-heading](version.id)")
+           if [ "$service" == "preview" ]; then
```
zmerlynn, line 394: nit: this if is indented one extra space :)

cloudbuild.yaml
```diff
@@ -374,6 +374,30 @@ steps:
...
+            tail -n +20 <<< "$versions"
```
markmandel, line 395: Let's just make it all 50 versions - on review there isn't an actual reason to have different amounts.

kamaljeeti: I am removing the 'if' condition because I don't think it's necessary now. I've applied it previously in case we needed to remove a different number of versions, but if we're deleting 50 versions for all three services, this approach will also work.

kamaljeeti:
Hi @zmerlynn , could you please review it? Thanks.

zmerlynn:
Let's fix the 0 case so it's not always failing:
image

You can use the `--no-run-if-empty` option to skip execution if there's no input: https://stackoverflow.com/a/8296746

zmerlynn (approved):

PR #3893, opened 2024-07-03
feat: Add API Changes and Validation for FleetAutoscaler Chain Policy

indexjoseph:
**What type of PR is this?**

/kind feature

Followup of #3880, this change adds the new Chain Policy to the `fleetautoscaler.go` file and adds tests for validation of the chain policy.

Towards #3008


pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -423,6 +504,67 @@ func (l *ListPolicy) ValidateListPolicy(fldPath *field.Path) field.ErrorList {
...
+	for i, entry := range c.Items {
+		if entry.ID != "" {
+			for j, otherEntry := range c.Items {
+				// If the entry's id is the same as another entry's id in the chain, append an error
+				if (entry.ID == otherEntry.ID || entry.ID == strconv.Itoa(j)) && i != j {
+					allErrs = append(allErrs, field.Invalid(fldPath.Child("items"), entry.ID, "id of chain entry must be unique"))
+				}
+			}
+		}
+		if entry.Schedule.Between.Start != "" {
```
vicentefb, line 522: Should we check/validate that the start time is before end time as well ?

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -423,6 +504,67 @@ func (l *ListPolicy) ValidateListPolicy(fldPath *field.Path) field.ErrorList {
...
+func (c *ChainPolicy) ValidateChainPolicy(fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if !runtime.FeatureEnabled(runtime.FeatureScheduledAutoscaler) {
+		return append(allErrs, field.Forbidden(fldPath, "feature ScheduledAutoscaler must be enabled"))
+	}
+	for i, entry := range c.Items {
+		if entry.ID != "" {
+			for j, otherEntry := range c.Items {
+				// If the entry's id is the same as another entry's id in the chain, append an error
+				if (entry.ID == otherEntry.ID || entry.ID == strconv.Itoa(j)) && i != j {
```
vicentefb, line 517: Could we benefit from a `seenIDs := make(map[string]bool) ` map here ?

pkg/apis/autoscaling/v1/fleetautoscaler.go
indexjoseph: I was thinking about this, if the user enters an end time that's before the start time then the policy is just never applied.

vicentefb: But we don't ouput any error at this moment, right ?

indexjoseph: Same case with the end date being before the current date.

indexjoseph: No, we do not.

indexjoseph: Talked offline w/ Zach he said we should go ahead and validate it.

examples/chainfleetautoscaler.yaml
```diff
@@ -27,24 +27,24 @@ kind: FleetAutoscaler
...
   name: chain-fleet-autoscaler
 spec:
+  fleetName: "fleet-example"
   policy:
     # Chain based policy for autoscaling.
     type: Chain
     chain:
       # Id of chain entry. If not set, the Id will be defaulted to the index of the entry within the chain.
       - id: "weekday"
-        type: Schedule  # Schedule based condition.
```
markmandel, line 36: 🤔 why did type disappear? Should definitely still be there, since it's a include of `policy` within `chain`.

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -195,6 +210,61 @@ type ListPolicy struct {
 	BufferSize intstr.IntOrString `json:"bufferSize"`
 }
 
+// Between defines the time period that the policy is eligible to be applied.
+type Between struct {
+	// Start is the datetime that the policy is eligible to be applied.
+	// This field must conform to RFC3339 format. If this field not set or is in the past, the policy is eligible to be applied
+	// as soon as the fleet autoscaler is running.
+	Start string `json:"start"`
```
markmandel, line 218: I'm going to assume that we can't make this a Go datetime object such that it will automatically get parsed appropriately?

We should be able to: https://pkg.go.dev/time#Time.MarshalJSON

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -195,6 +210,61 @@ type ListPolicy struct {
...
+
+	// End is the datetime that the policy is no longer eligible to be applied.
+	// This field must conform to RFC3339 format. If not set, the policy is always eligible to be applied, after the start time above.
+	End string `json:"end"`
+}
+
+// ActivePeriod defines the time period that the policy is applied.
+type ActivePeriod struct {
+	// Timezone to be used for the startCron field. If not set, startCron is defaulted to the UTC timezone.
+	Timezone string `json:"timezone"`
```
markmandel, line 228: Same here, if the types we want to use have appropriate JSON interface arguments, we can likely use them directly.

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -195,6 +210,61 @@ type ListPolicy struct {
...
+	ActivePeriod ActivePeriod `json:"activePeriod"`
+}
+
+// ChainEntry defines a single entry in the ChainPolicy.
+type ChainEntry struct {
+	// ID is the unique identifier for a ChainEntry. If not set the identifier will be set to the index of chain entry.
+	ID string `json:"id"`
+
+	// Schedule defines when the policy is applied.
+	Schedule Schedule `json:"schedule"`
```
markmandel, line 259: This feels wrong to me. I think this is going to be an embedded `FleetAutoscalerPolicy` instead, since we don't have a top level `scheduled` within `ChainEntry`, you have a `type` first to say if the chain entry is a scheduled one or not.

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -423,6 +502,85 @@ func (l *ListPolicy) ValidateListPolicy(fldPath *field.Path) field.ErrorList {
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("schedule").Child("activePeriod").Child("timezone"), entry.Schedule.ActivePeriod.Timezone, fmt.Sprintf("invalid timezone: %s", err)))
+		}
+		if entry.Schedule.ActivePeriod.StartCron != "" {
+			// If startCron is not a valid cron expression, append an error
+			if _, err := cron.ParseStandard(entry.Schedule.ActivePeriod.StartCron); err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("schedule").Child("activePeriod").Child("startCron"), entry.Schedule.ActivePeriod.StartCron, fmt.Sprintf("invalid startCron: %s", err)))
+			}
+			// If the cron expression contains a CRON_TZ or TZ specification, append an error
+			if strings.Contains(entry.Schedule.ActivePeriod.StartCron, "TZ") {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("schedule").Child("activePeriod").Child("startCron"), entry.Schedule.ActivePeriod.StartCron, "CRON_TZ or TZ used in start is not supported, please use the .schedule.timezone field to specify a timezone"))
```
zmerlynn, line 561: I don't quite understand `used in start` - did you mean `used in activePeriod`?

zmerlynn (commented):
Generally LGTM, though agree with Mark's comments. Will review again after!

examples/chainfleetautoscaler.yaml
indexjoseph: I think the `type` you're referring to is for the denoting the policy type. However, currently the CRD doesn't have a type to denote whether or not an entry is scheduled or not, so I'll add that in the CRD and add it back to the example as well.

pkg/apis/autoscaling/v1/fleetautoscaler.go
indexjoseph: I've went ahead and separated `Schedule` from `Chain`, and it's now it's own policy.

pkg/apis/autoscaling/v1/fleetautoscaler.go
indexjoseph: I had to use the k8's `time`, since the built-in Golang `time` doesn't have a `DeepCopy` functions.

pkg/apis/autoscaling/v1/fleetautoscaler.go
indexjoseph: I could not find one for `time.Location`, so I've kept it as a string for now. However, I was able to find/use one for the `Start` and `End` fields.

examples/chainfleetautoscaler.yaml
zmerlynn: (I think this is resolved now, it's down on new line 38)

pkg/apis/autoscaling/v1/fleetautoscaler_test.go
```diff
@@ -456,6 +457,160 @@ func TestFleetAutoscalerListValidateUpdate(t *testing.T) {
...
+
+	testCases := map[string]struct {
+		fas          *FleetAutoscaler
+		featureFlags string
+		wantLength   int
+		wantField    string
+	}{
+		"end time before current time": {
+			fas: modifiedFAS(func(fap *FleetAutoscalerPolicy) {
+				fap.Schedule.Between.End = metav1.Date(2024, 0o7, 0o3, 15, 59, 59, 0, time.UTC)
```
zmerlynn, line 477: As a suggestion to make the tests more readable, just add a mustParseDate function that takes an RFC3339 field.

pkg/apis/autoscaling/v1/fleetautoscaler_test.go
```diff
@@ -456,6 +457,160 @@ func TestFleetAutoscalerListValidateUpdate(t *testing.T) {
...
+
+	modifiedFAS := func(f func(*FleetAutoscalerPolicy)) *FleetAutoscaler {
+		fas := scheduleFixture()
+		f(&fas.Spec.Policy)
+		return fas
+	}
+
+	testCases := map[string]struct {
+		fas          *FleetAutoscaler
+		featureFlags string
```
zmerlynn, line 471: Add an `=false` test, similar to the off test on line 543 below?

zmerlynn (commented):
Almost there!

pkg/apis/autoscaling/v1/fleetautoscaler_test.go
indexjoseph: Sg, for some odd reason k8's time representation enforces octal literal style.

zmerlynn (approved):

Issue #3870, opened 2024-06-13
Beta: UpdateList using Agones 1.41 rest API doesn't update the data in the selected list


Tsm012:
**What happened**:
Attempting to use the Lists features in Agones 1.41. I have a fleet deployed in agones and am attempting to use the:

`curl -d '{"capacity": "120", "values": ["player3", "player4"]}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players`

call as mentioned in the rest documentation. The list is declared on our "kind: Fleet" resource and declares a list players

```
lists:
  players:
    capacity: 20
    values:
    - player1
    - player2
    - player3
```

Once the fleet is deployed, I can run the command:

`curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players`

and will get the correct response.

`{"name":"players","capacity":"20","values":["player1","player2","player3"]}`

When I go to update that same list using the command:

`curl -d '{"capacity": "20", "values": ["player3", "player4"]}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players`

I get the response:

`{"name":"","capacity":"0","values":[]}`

I then check again to see if my list was updated:

`curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players`

but the reply seems to indicate nothing has changed.

`{"name":"players","capacity":"20","values":["player1","player2","player3"]}`

**What you expected to happen**:

I expected the list up update to reflect the data payload of 

`{"capacity": "20", "values": ["player3", "player4"]}`

**How to reproduce it (as minimally and precisely as possible)**:

Create a Fleet resource in agones and deploy it with a list declared as:

```
lists:
  players:
    capacity: 20
    values:
    - player1
    - player2
    - player3
```

shell into a game server in the fleet just deployed and verify that your initial values are in the list

```
curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players

{"name":"players","capacity":"20","values":["player1","player2","player3"]}
```

Attempt to update the list with new values

[...]

PR #3897, opened 2024-07-09
Changing update list to update list values along with the capacity.

chrisfoster121:
/kind bug 

**What this PR does / Why we need it**:
This PR updates the UpdateList functionality to update the list values along side the given capacity. This is very important because without it managing game server keys gets very complicated and often causes unwanted race conditions when sending multiple additions or removals to lists in quick succession.

**Which issue(s) this PR fixes**:
Closes #3870

PR #3899, opened 2024-07-10
Updating UpdateList to update the values on a list

chrisfoster121:
/kind bug
What this PR does / Why we need it:
This PR updates the UpdateList functionality to update the list values along side the given capacity. This is very important because without it managing game server keys gets very complicated and often causes unwanted race conditions when sending multiple additions or removals to lists in quick succession.

Which issue(s) this PR fixes:
Closes https://github.com/googleforgames/agones/issues/3870


chrisfoster121:
@igooch This is a duplicate of the previous pull request that I made and fixes the CLA issue. What is the process from here to get this approved?

igooch:
I pulled down your PR and installed on a cluster. It doesn't look like the code is working as intended.

```
me@me:~/agones/build$ curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"players","capacity":"10","values":["player1","player2","player3"]}
me@me:~/agones/build$ curl -d '{"capacity": "120", "values": ["player3", "player4"]}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
me@me:~/agones/build$ curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"players","capacity":"120","values":["player1","player2","player3"]}
me@me:~/agones/build$ curl -d '{"values": ["player5"]}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
me@me:~/agones/build$ curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"players","capacity":"0","values":[]}
me@me:~/agones/build$ k describe gs
…
Status:
  …
  Lists:
    Players:
      Capacity:  0
      Values:
…
```

The UpdateList should make use of the field masks for specifying which field(s) are being updated. More info on that in the PR that was closed https://github.com/googleforgames/agones/pull/3897#issuecomment-2219159146 .

Please also include tests for the new behavior.

chrisfoster121:
Hmm that is strange as it worked locally for me. To test this I am using the following for my gameserver.yaml:
```
apiVersion: "agones.dev/v1"
kind: Fleet
metadata:
  name: simple-game-server-fleet
spec:
  replicas: 1
  template:
    spec:
      sdkServer:
        logLevel: Debug
      lists:
        Players:
          capacity: 0
          values:
      container: simple-game-server
      template:
        spec:
          containers:
            - name: simple-game-server
              image: us-docker.pkg.dev/agones-images/examples/simple-game-server:0.31
            - name: alpine   
              image: alpine/curl 
              command: ["sleep"]         # Override the container entrypoint with a simple sleep command
              args: ["1000000"]  

```
One thing that is important to note is that when running kubectl describe gs I see the following:
```
Spec:
  Lists:
    Players:
      Capacity:  0
      Values:
    Players:
      Capacity:  3
      Values:
        player1
        player2
        player3
```
And:
```
Status:
  Lists:
    Players:
      Capacity:  0
      Values:
    Players:
      Capacity:  1
      Values:
        player5
```
I think this is due to there already being a predefined list named `players` and when describing the pod it capitalizes the output for attributes.

Anyways, to test I copied your commands into my shell and ran them to get the following output.
```
/ # curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"players","capacity":"3","values":["player1","player2","player3"]}/ #
/ #
/ # curl -d '{"capacity": "120", "values": ["player3", "player4"]}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"","capacity":"0","values":[]}/ #
/ #
/ # curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"players","capacity":"120","values":["player3","player4"]}/ #
/ #
/ # curl -d '{"values": ["player5"]}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"","capacity":"0","values":[]}/ #
/ #
/ # curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"players","capacity":"0","values":[]}/ #
/ #
/ # curl -d '{"capacity":"1", "values": ["player5"]}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"","capacity":"0","values":[]}/ #
/ #
/ # curl -H "Content-Type: application/json" -X GET http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/lists/players
{"name":"players","capacity":"1","values":["player5"]}/ #
````
@igooch Would you be able to provide more detailed steps to reproduce the issue you are seeing? If I can start replicating this on my end it will be a lot easier to diagnose the issue.
__________
> The UpdateList should make use of the field masks for specifying which field(s) are being updated. More info on that in the   > PR that was closed [#3897 (comment)](https://github.com/googleforgames/agones/pull/3897#issuecomment-2219159146) .

I am a bit confused by that. Looking at both [AddListValue and RemoveListvalue](https://github.com/googleforgames/agones/blob/31ed93ee9c0c3070a1139b145d0d6debac8b4572/pkg/sdkserver/sdkserver.go#L1102) in sdkserver.go, neither of them use the field masks. They both queue the changes for being processed later in [updateList](https://github.com/googleforgames/agones/blob/31ed93ee9c0c3070a1139b145d0d6debac8b4572/pkg/sdkserver/sdkserver.go#L1181). Are field masks specific to UpdateList? I went with that method because it matched the pattern that was already used. I can change this but I am just trying to understand why one way over the other.

**Comment by chrisfoster121** [comment]
@igooch I just wanted to follow up to see if you had a chance to check this out. I was unable to repro your issue so if you can provide a bit more detailed repro steps that would be great, and then I'll hopefully have a better idea of what is going wrong!

**Comment by igooch** [comment]
The seeing two `Players` keys for Status and Spec is bizarre -- `Lists` is a map, so the keys should be unique. The Spec spec should always remain the same as the original spec template, and the Status changes with the with the Update, Add, Remove, etc.

I'm also not sure what you mean by not being able to replicate? From what I can tell the code in the above comment is replicated exactly. The only difference being that there's an additional `curl -d '{"capacity":"1", "values": ["player5"]}' -H "Content-Type: application/json" -X PATCH` command at the end. 

The reason the existing SDK Server doesn't currently use the protobuf field mask is because it's only changing one field, the capacity. Or in the case of Add or Remove it's only changing the values field. If multiple values can be changed, then that's when the field mask comes into use. Using the field mask allows for updates that only change specific fields, and leave the other fields untouched. So the command `curl -d '{"capacity": "120"}' -H "Content-Type: application/json" -X PATCH` should only change the `capacity` field, and `curl -d '{"values": ["player5"]}' -H "Content-Type: application/json" -X PATCH` should only change the `values` field. Right now if someone passes in the command `curl -d '{"values": ["player5"]}' -H "Content-Type: application/json" -X PATCH`, then it uses the default value for the `UpdateListRequest` capacity, which is 0. That's why the `Get` request after this command returns an empty list. Does that make sense?

**Comment by chrisfoster121** [comment]
@igooch I have updated the PR to use the field mask system as requested. I have not gotten a chance to test this out yet and plan to do so in the next day or so. 

However, I wanted to see if I could get some feedback on this as I ran into some issues with the unit tests. When running `TestSDKServerUpdateList` in sdkserver_test.go I ran into issues getting the game server data to patch correctly. Specifically, when calling `s.patchGameServer(ctx, gs, gsCopy)` in `UpdateList` it appears to call `s.gameServerGetter.GameServers(s.namespace)` which seems to return a `FakeGameServer`. 

Is this expected for unit tests and if so is there something I need to change to allow the test case to pass correctly? In the mean time I added a line to make it directly update the game server to pass the tests, however, I don't think this should be submitted and would like to find the correct solution. Is this something you are familiar with?

File: pkg/sdkserver/sdkserver.go [hunk 1]
```diff
@@ -1077,18 +1085,36 @@ func (s *SDKServer) UpdateList(ctx context.Context, in *beta.UpdateListRequest)
...
 	if in.List.Capacity < 0 || in.List.Capacity > apiserver.ListMaxCapacity {
 		return nil, errors.Errorf("out of range. Capacity must be within range [0,1000]. Found Capacity: %d", in.List.Capacity)
 	}
 
-	if _, ok := gs.Status.Lists[name]; ok {
-		batchList := s.gsListUpdates[name]
-		batchList.capacitySet = &in.List.Capacity
-		s.gsListUpdates[name] = batchList
-		// Queue up the Update for later batch processing by updateLists.
-		s.workerqueue.Enqueue(cache.ExplicitKey(updateLists))
```
igooch, line 1090: This is the portion that sends the changes to the queue for later processing.

pkg/sdkserver/sdkserver.go
```diff
@@ -1077,18 +1085,36 @@ func (s *SDKServer) UpdateList(ctx context.Context, in *beta.UpdateListRequest)
...
+		proto.Merge(tmpList, in.List)
+		// Silently truncate list values if Capacity < len(Values)
+		if tmpList.Capacity < int64(len(tmpList.Values)) {
+			tmpList.Values = append([]string{}, tmpList.Values[:tmpList.Capacity]...)
+		}
+
+		list.Capacity = tmpList.Capacity
+		list.Values = tmpList.Values
+		gsCopy.Status.Lists[name] = list
+		s.patchGameServer(ctx, gs, gsCopy)
```
igooch, line 1114: We don't want to call this here, all the calls are batched, and then the actual change to the gameserver happens in `updatelist` https://github.com/googleforgames/agones/blob/e35d73fcae1c9e7f719d738f4f1638c7ce8d6c7d/pkg/sdkserver/sdkserver.go#L1218

@markmandel do you want to weigh in here? It does seem a bit odd to batch an "overwrite" function?

markmandel: +1 on batching - for a few reasons.

markmandel: +1 to batching for a few reasons: 

1. It puts work items in the workerqueue - which for an "overwrite" operation isn't necessary for conflicts, but is necessary if the KCP goes down for any reason -- so we stay self-healing.
2. The experience across the SDK stays the same regardless of operation.

pkg/sdkserver/sdkserver.go
```diff
@@ -1066,6 +1068,12 @@ func (s *SDKServer) UpdateList(ctx context.Context, in *beta.UpdateListRequest)
 	if in == nil {
 		return nil, errors.Errorf("UpdateListRequest cannot be nil")
 	}
+	if in.List == nil || in.UpdateMask == nil {
```
markmandel, line 1071: I defer to you @igooch on if this is correct for the fields -- you know more about this than I do. But wanted to check, because it _reads_ like List is required?

chrisfoster121:
@igooch I appear to be failing on the `submit-e2e-test-cloud-build` step:
```
Step #25 - "submit-e2e-test-cloud-build": gke-autopilot-1.28: BUILD FAILURE: Build step failure: build step 1 "e2e-runner" failed: step exited with non-zero status: 2
Step #25 - "submit-e2e-test-cloud-build": gke-autopilot-1.28: ERROR: (gcloud.builds.submit) build 8050e9fa-86df-44f5-8c48-2a4a68048363 completed with status "FAILURE"
Step #25 - "submit-e2e-test-cloud-build": One of the e2e test child cloud build exited with nonzero status 1. Aborting.
```

I am not sure what this step is, would you be able to clarify this for me?

igooch:
> @igooch I appear to be failing on the `submit-e2e-test-cloud-build` step:
> 
> ```
> Step #25 - "submit-e2e-test-cloud-build": gke-autopilot-1.28: BUILD FAILURE: Build step failure: build step 1 "e2e-runner" failed: step exited with non-zero status: 2
> Step #25 - "submit-e2e-test-cloud-build": gke-autopilot-1.28: ERROR: (gcloud.builds.submit) build 8050e9fa-86df-44f5-8c48-2a4a68048363 completed with status "FAILURE"
> Step #25 - "submit-e2e-test-cloud-build": One of the e2e test child cloud build exited with nonzero status 1. Aborting.
> ```
> 
> I am not sure what this step is, would you be able to clarify this for me?

Builds are currently broken -- @zmerlynn is investigating #3939. I'll try and take a look later today as well.

chrisfoster121:
> > @igooch I appear to be failing on the `submit-e2e-test-cloud-build` step:
> > ```
> > Step #25 - "submit-e2e-test-cloud-build": gke-autopilot-1.28: BUILD FAILURE: Build step failure: build step 1 "e2e-runner" failed: step exited with non-zero status: 2
> > Step #25 - "submit-e2e-test-cloud-build": gke-autopilot-1.28: ERROR: (gcloud.builds.submit) build 8050e9fa-86df-44f5-8c48-2a4a68048363 completed with status "FAILURE"
> > Step #25 - "submit-e2e-test-cloud-build": One of the e2e test child cloud build exited with nonzero status 1. Aborting.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am not sure what this step is, would you be able to clarify this for me?
> 
> Builds are currently broken -- @zmerlynn is investigating #3939. I'll try and take a look later today as well.

Ah I see, thank you for the heads up!

 In the meantime, I do have some questions about some of the codebase as it relates to the batching/queueing system.

I may be mistaken but it seems that when a request to change a list comes in (Add, Remove, and now Update), the current value in the queue gets overwritten.  `SDKServer` has a variable `gsListUpdates   map[string]listUpdateRequest` which is a one to one mapping. Then in `AddListValue`, for example, we have the following code:
```
batchList := s.gsListUpdates[in.Name]
batchList.valuesToAppend = list.Values
s.gsListUpdates[in.Name] = batchList
```
In this case that value for how to update the list is getting overwritten. My worry is that this will create a scenario where sending too many requests at the same time will cause the system to overwrite requests. In fact, this was the initial reason I started looking into using `UpdateList`. I was running into a scenario where I would make too many AddListValue requests in a short amount of time resulting in some of them not getting processed. Is this intentional and if so are there plans for making the request wait to return until the list has been updated? Alternatively, are there plans to make this no longer overwrite previous requests?

igooch:
> In the meantime, I do have some questions about some of the codebase as it relates to the batching/queueing system.
> 
> I may be mistaken but it seems that when a request to change a list comes in (Add, Remove, and now Update), the current value in the queue gets overwritten. `SDKServer` has a variable `gsListUpdates map[string]listUpdateRequest` which is a one to one mapping. Then in `AddListValue`, for example, we have the following code:
> 
> ```
> batchList := s.gsListUpdates[in.Name]
> batchList.valuesToAppend = list.Values
> s.gsListUpdates[in.Name] = batchList
> ```
> 
> In this case that value for how to update the list is getting overwritten. My worry is that this will create a scenario where sending too many requests at the same time will cause the system to overwrite requests. In fact, this was the initial reason I started looking into using `UpdateList`. 

The `AddListValue` locks on the resource, so multiple requests can't write at the same time. 

https://github.com/googleforgames/agones/blob/ff4c2220669804de7f954ad643b809a08d3fb567/pkg/sdkserver/sdkserver.go#L1116

It then appends the new value to the list, so that all values are written back.

https://github.com/googleforgames/agones/blob/ff4c2220669804de7f954ad643b809a08d3fb567/pkg/sdkserver/sdkserver.go#L1129

So batching complicates the `UpdateList` behavior. If two `UpdateLists` come in at the same time we could either 
a) overwrite everything, although any add or remove requests that come in after the update list request but before the batch would still modify the batch 
b) append all unique `UpdateList` values to existing list of values 
c) not batch update list requests (although as @markmandel mentioned this isn't preferred behavior) 
d) something else. 
Out of this list I'd lean towards the option b).

>I was running into a scenario where I would make too many AddListValue requests in a short amount of time resulting in some of them not getting processed. Is this intentional and if so are there plans for making the request wait to return until the list has been updated? Alternatively, are there plans to make this no longer overwrite previous requests?

This is interesting, have you seen any error like `error retrieving resource lock` or other debug logs like `out of range` or `already exists` that indicate what requests are dropped?

chrisfoster121:
@igooch I added a mutex lock and made minor adjustments to clean the code up a bit. Does this look correct?

igooch (commented):
Thanks for the update, one quick adjustment to the mutex. Just to confirm on this, the `UpdateList` will overwrite everything, although any add or remove or other update requests that come in after the initial update list request but before the batch would still modify the batch. Will this behavior work for your use case?

chrisfoster121:
> Thanks for the update, one quick adjustment to the mutex. Just to confirm on this, the `UpdateList` will overwrite everything, although any add or remove or other update requests that come in after the initial update list request but before the batch would still modify the batch. Will this behavior work for your use case?

Yes at least for us, we were expecting the list to be overwritten by UpdateList calls which this should so. This allows us to completely change the list in a single call if needed. 

I will update the code shortly. Thank you!

igooch (approved):
LGTM, thank for you for your work and patience on this!

PR #3913, opened 2024-07-18
Adds instructions for adding and removing field from a CRD

### LINKED ISSUE #3771: In-place Agones Upgrades: Storage Compatibility
> [!NOTE]
> Milestone of #3766, which we are seeking feedback on. We will move forward with pieces that seem non-contentious, though.

# Storage Compatibility

## Defaulting

Right now, Agones defaults values in the webhook, e.g. [GameServer defaulting](https://github.com/googleforgames/agones/blob/1eb7762e7a135c00f4d73a9250436cf01221e368/pkg/apis/agones/v1/gameserver.go#L340-L352) is the only real thing the [GameServer mutation webhook](https://github.com/googleforgames/agones/blob/1eb7762e7a135c00f4d73a9250436cf01221e368/pkg/gameservers/controller.go#L256-L258) does. Defaulting serves a couple of purposes:

* Code can rely on default values, e.g. [hook for eviction.safe](https://github.com/googleforgames/agones/blob/main/pkg/cloudproduct/eviction/eviction.go#L24-L57). It’s complicated to constantly have to verify the object has been properly defaulted.
* Users see the rendered default value in `kubectl describe`. This increases discoverability for new API elements, as `describe` is a common UI element.

The problem is that defaulting in the webhook alone is not safe across configuration changes. Example using `eviction.safe`:
* t0: Agones deployed at 1.29 with `SafeToEvict` disabled.
* t1: GameServer `foo` created. `eviction.safe` is not defaulted because the feature gate was not enabled.
* t2: Agones upgraded to 1.30, `SafeToEvict` enabled. `eviction.safe` is defaulted on new objects.
* t3: controller fails silently for foo, [logging unknown eviction safe value “”](https://github.com/googleforgames/agones/blob/main/pkg/cloudproduct/eviction/eviction.go#L38).

The reason for the failure is that [the hook](https://github.com/googleforgames/agones/blob/main/pkg/cloudproduct/eviction/eviction.go#L24-L57) blindly assumes the value was defaulted by the webhook, but the webhook never had a chance to run. Note that for this particular case, the gap here is very narrow in time - in particular t1 and t2 need to occur such that defaulting of the GameServer occurs on a 1.29 `agones-extensions` container, but a 1.30 `agones-controller` container attempts to create the Pod. That said, this condition could easily occur during rollout of 1.30, though, and cause a multi-second hiccup.

To solve this problem, I propose we:
* Change our defaulting code to be idempotent. Right now there is a [comment](https://github.com/googleforgames/agones/blob/1eb7762e7a135c00f4d73a9250436cf01221e368/pkg/apis/agones/v1/gameserver.go#L340C4-L340C92) ApplyDefaults applies default values to the GameServer if they are not already populated which suggests that this should be the case, but is immediately followed by [two non-idempotent operations](https://github.com/googleforgames/agones/blob/1eb7762e7a135c00f4d73a9250436cf01221e368/pkg/apis/agones/v1/gameserver.go#L347-L348) (at least non-idempotent over an upgrade).
* If defaulting is idempotent, then the event handlers for controller can run the defaulting before the GameServer enters the queue, e.g. we can call [ApplyDefaults](https://github.com/googleforgames/agones/blob/1eb7762e7a135c00f4d73a9250436cf01221e368/pkg/apis/agones/v1/gameserver.go#L340-L352) [here](https://github.com/googleforgames/agones/blob/4bb673186b2ee8431de00c5c564bd1daa1a356df/pkg/gameservers/controller.go#L389-L396) (and in similar places we `Get` in the controller - probably via helper function). This effectively ensures defaulting on creation and on subsequent update, but it relies on idempotent defaulting.

## Unknown or Disabled Fields

A similar problem exists for API fields that should not be present but have already been set, which typically only occurs if a feature gate has been disabled (either due to downgrade or explicit disablement). This takes a couple of forms:
* An old-version controller reads a new-version field that it has no knowledge of. Again using the example of `eviction.safe`, imagine downgrading from 1.30 (where the feature gate is Beta and therefore the API is defaulted) to 1.28 (prior to `eviction.safe` even existing).
* A controller has knowledge of a field but the feature gate is disabled. This path results in [validation errors on create](https://github.com/googleforgames/agones/blob/1eb7762e7a135c00f4d73a9250436cf01221e368/pkg/apis/agones/v1/gameserver.go#L487-L506), but in the case the user actively disabled a feature gate, we need to cover the latent/existing objects as well.

In either of these cases, we need to ensure that on "first touch", the controller drops the unknown fields, rather than preserving them. In general, this is a safer handling of latent unknown fields - otherwise when the feature gate is reenabled, a preserved field could surprise the user. (Note that the first case, where the field is not present at all in the CRD, is generally covered by [field pruning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning), so mostly this is figuring out logic for the latter case.)

## Update vs Patch for controllers
Note that controllers have a similar problem to the SDK of using `Update` vs `Patch`, [mentioned here](https://github.com/googleforgames/agones/issues/3767#update-vs-patch) - different controller versions may drop fields. However:
- skew between controller versions is not expected to last long (minutes at most), so the problem is short-lived. This is a little different than the `sdk-server`, which requires a full rollout.
- we call out in the constraints that some features may flap until rolled out: https://github.com/googleforgames/agones/issues/3766#feature-gates

igooch:
**What type of PR is this?**

/kind documentation

Adds instructions necessary for adding and removing a field from a CRD, so that the CRD will be compatible across upgrades. 

Closes #3771


install/helm/agones/README.md
```diff
@@ -3,3 +3,69 @@
 This chart installs the Agones application and defines deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
 
 See [Install Agones using Helm](https://agones.dev/site/docs/installation/install-agones/helm/) for installation and configuration instructions.
+
+## Development Work on Agones Custom Resource Definitions (CRDs)
```
zmerlynn, line 7: This feels like a really weird place - I feel like this is a little more user-centric README in general.

I'm struggling to come up with a better place, though. @markmandel ?

igooch: Would it make sense to follow the same pattern as the SDK and link to the ReadMe from the contributing page? https://github.com/googleforgames/agones/blob/46c5ef9f7433004c3fc1721b405793f26d5d224f/README.md?plain=1#L56

install/helm/agones/README.md
```diff
@@ -3,3 +3,69 @@
...
+
+Agones [extends the Kubernetes API with CustomResourceDefinitions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) for the kinds `Fleet`, `GameServerSet`, `GameServer`, `FleetAutoscaler`. (`GameServerAllocationPolicy` has been depcrecated in favor of `GameServerAllocation` which does not a have CRD.) Regardless of installation method, the definitions for the Agones Custom Resources are in the [agones/install/helm/agones/templates/crds](./templates/crds/) directory. The double braces `{{ }}` in the CRDs and elsewhere are [Helm chart template](https://helm.sh/docs/chart_template_guide/getting_started/) features.
```
markmandel, line 9:
```suggestion
(`GameServerAllocationPolicy` has been depcrecated in favor of `GameServerAllocation` which does not a have CRD.)
```

These are not the same thing , I think we can remove this line.

install/helm/agones/README.md
markmandel: This has both a style guide as well as a development guide - which is not a bad thing, but we've never quite done it like this, so it's a little trickier to put in an existing place.

> Would it make sense to follow the same pattern as the SDK and link to the ReadMe from the contributing page?

What README would you put it in? There isn't really a spot that I can think of. We could always add a `docs` folder to https://github.com/googleforgames/agones/tree/main/build though? But open to your suggestions.

It could possibly be a section in https://github.com/googleforgames/agones/blob/main/build/README.md#testing-and-building ?

Or take a stab at taking apart https://github.com/googleforgames/agones/blob/main/build/README.md into some sub-sectioned files, even it's just some basics - and it could be the beginning of making that a better experience - right now it's not terrific.

gongmax (approved):

PR #3927, opened 2024-07-30
[Performance] - Added a new metric inside the allocator to track the success retry rate inside the retry loop 

vicentefb:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

Closes #

### DIFF
--- a/pkg/gameserverallocations/allocator.go
+++ b/pkg/gameserverallocations/allocator.go
@@ -193,14 +193,8 @@ func (c *Allocator) Allocate(ctx context.Context, gsa *allocationv1.GameServerAl
 	latency := c.newMetrics(ctx)
 	defer func() {
 		if err != nil {
-			//var errorReason metav1.StatusReason
-			if status, ok := err.(k8serrors.APIStatus); ok || errors.As(err, &status) {
-				//errorReason = status.Status().Reason
-			}
-			latency.setError( /*string(errorReason)*/ )
+			latency.setError()
 		}
-		c.baseLogger.Info("[VICENTE] RECORDING")
-
 		latency.record()
 	}()
 	latency.setRequest(gsa)
@@ -239,6 +233,7 @@ func (c *Allocator) Allocate(ctx context.Context, gsa *allocationv1.GameServerAl
 		return nil, err
 	}
 	latency.setResponse(out)
+
 	return out, nil
 }
 
@@ -258,25 +253,19 @@ func (c *Allocator) loggerForGameServerAllocation(gsa *allocationv1.GameServerAl
 // Add here  a a number of times we are entering
 func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocationv1.GameServerAllocation) (*allocationv1.GameServerAllocation, error) {
 	var gs *agonesv1.GameServer
-	//latency := c.newMetrics(ctx)
+	latency := c.newMetrics(ctx)
 	retryCount := 0
 	err := Retry(allocationRetry, func() error {
 		var err error
-		//var errorReason metav1.StatusReason
 		gs, err = c.allocate(ctx, gsa)
 		retryCount = retryCount + 1
 
 		c.loggerForGameServerAllocation(gsa).WithError(err).Warn("Failed to Allocated. Retrying...")
 
-		if err != nil {
-			if status, ok := err.(k8serrors.APIStatus); ok || errors.As(err, &status) {
-				//errorReason = status.Status().Reason
-			}
-			//latency.setError(string(errorReason))
-		} else {
-			//latency.setError(string("Success ") + fmt.Sprint(retryCount))
+		if err == nil {
+			latency.setError(string("Success ") + fmt.Sprint(retryCount))
 		}
-		//latency.recordAllocationErrorRate()
+		latency.recordAllocationErrorRate()
 
 		return err
 
--- a/pkg/gameserverallocations/allocator.go
+++ b/pkg/gameserverallocations/allocator.go
@@ -263,7 +263,7 @@ func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocatio
 		if err != nil {
 			c.loggerForGameServerAllocation(gsa).WithError(err).Warn("Failed to Allocated. Retrying...")
 		} else {
-			latency.recordAllocationErrorRate(ctx, retryCount)
+			latency.recordAllocationRetrySuccess(ctx, retryCount)
 		}
 		return err
 	})
--- a/pkg/gameserverallocations/allocator.go
+++ b/pkg/gameserverallocations/allocator.go
@@ -193,12 +193,14 @@ func (c *Allocator) Allocate(ctx context.Context, gsa *allocationv1.GameServerAl
 	latency := c.newMetrics(ctx)
 	defer func() {
 		if err != nil {
-			var errorReason metav1.StatusReason
+			//var errorReason metav1.StatusReason
 			if status, ok := err.(k8serrors.APIStatus); ok || errors.As(err, &status) {
-				errorReason = status.Status().Reason
+				//errorReason = status.Status().Reason
 			}
-			latency.setError(string(errorReason))
+			latency.setError( /*string(errorReason)*/ )
 		}
+		c.baseLogger.Info("[VICENTE] RECORDING")
+
 		latency.record()
 	}()
 	latency.setRequest(gsa)
@@ -256,11 +258,11 @@ func (c *Allocator) loggerForGameServerAllocation(gsa *allocationv1.GameServerAl
 // Add here  a a number of times we are entering
 func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocationv1.GameServerAllocation) (*allocationv1.GameServerAllocation, error) {
 	var gs *agonesv1.GameServer
-	latency := c.newMetrics(ctx)
+	//latency := c.newMetrics(ctx)
 	retryCount := 0
 	err := Retry(allocationRetry, func() error {
 		var err error
-		var errorReason metav1.StatusReason
+		//var errorReason metav1.StatusReason
 		gs, err = c.allocate(ctx, gsa)
 		retryCount = retryCount + 1
 
@@ -268,13 +270,13 @@ func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocatio
 
 		if err != nil {
 			if status, ok := err.(k8serrors.APIStatus); ok || errors.As(err, &status) {
-				errorReason = status.Status().Reason
+				//errorReason = status.Status().Reason
 			}
-			latency.setError(string(errorReason))
+			//latency.setError(string(errorReason))
 		} else {
-			latency.setError(string("Success ") + fmt.Sprint(retryCount))
+			//latency.setError(string("Success ") + fmt.Sprint(retryCount))
 		}
-		latency.recordAllocationErrorRate()
+		//latency.recordAllocationErrorRate()
 
 		return err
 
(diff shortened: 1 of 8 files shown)


vicentefb:
image

pkg/gameserverallocations/metrics.go
```diff
@@ -32,23 +32,42 @@ import (
...
+	stateViews := []*view.View{
+		{
+			Name:        "gameserver_allocations_duration_seconds",
+			Measure:     gameServerAllocationsLatency,
+			Description: "The distribution of gameserver allocation requests latencies.",
+			Aggregation: view.Distribution(0, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2, 3),
+			TagKeys:     []tag.Key{keyFleetName, keyClusterName, keyMultiCluster, keyStatus, keySchedulingStrategy},
+		},
+		{
+			Name:        "gameserver_allocations_retry_rate",
```
markmandel, line 58: I like this metric!

We will need documentation here: https://agones.dev/site/docs/guides/metrics/ (with `feature` shortcodes) please.

pkg/gameserverallocations/metrics.go
```diff
@@ -32,23 +32,42 @@ import (
...
+			Measure:     gameServerAllocationsErrorRate,
+			Description: "The distribution of gameserver allocation retry success rate",
```
markmandel, line 60: This isn't a histogram, so it's not a distribution, it's a count.

pkg/gameserverallocations/metrics.go
markmandel, line 58:
```suggestion
			Name:        "gameserver_allocations_retry_total",
```

https://prometheus.io/docs/practices/naming/

pkg/gameserverallocations/metrics.go
```diff
@@ -82,8 +101,8 @@ func (r *metrics) setStatus(status string) {
 }
 
 // setError set the latency status tag as error.
-func (r *metrics) setError() {
-	r.mutate(tag.Update(keyStatus, "error"))
+func (r *metrics) setError(errorType string) {
```
markmandel, line 104: This whole function seems... odd to me.

Any reason we aren't using `stats.RecordWithTags()` everywhere instead?

We have some nice helpers too: https://github.com/googleforgames/agones/blob/96c3d26857deebd05917260501f7ac74094ba461/pkg/metrics/util.go#L45-L49

pkg/gameserverallocations/allocator.go
```diff
@@ -250,15 +250,26 @@ func (c *Allocator) loggerForGameServerAllocation(gsa *allocationv1.GameServerAl
...
 	err := Retry(allocationRetry, func() error {
 		var err error
 		gs, err = c.allocate(ctx, gsa)
+		retryCount++
+
 		if err != nil {
-			c.loggerForGameServerAllocation(gsa).WithError(err).Warn("failed to allocate. Retrying...")
+			c.loggerForGameServerAllocation(gsa).WithError(err).Warn("Failed to Allocated. Retrying...")
+		} else {
+			latency.setError(string("Success ") + fmt.Sprint(retryCount))
```
markmandel, line 266: What's happening here? Not sure I follow.

pkg/gameserverallocations/metrics.go
vicentefb: Mmmm i'm not entirely sure why we aren't using `stats.RecordWithTags()`. There's only one other metric in this file which is the latency one and it uses `stats.Record(r.ctx, gameServerAllocationsLatency.M(time.Since(r.start).Seconds()))` directly. I could update this file and change it in this PR to use `stats.RecordWithTags()`

markmandel: SGTM

pkg/gameserverallocations/allocator.go
vicentefb: So if there's no error inside this retry loop it means that we had a success allocation within the 5 steps that we have defined for `allocationRetry = wait.Backoff` object. I think the method name is misleading since it's not an Error that we are setting (I'll change that) but it was a way for me to set a label and being able to query it as such: 

image

markmandel: Yeah that's not a good idea for a variety of reasons (mostly labels are not metrics). You just remade histograms 😄 

Use a histogram, and do a linear bucket from 1 to 10 (whatever the top expected number is?), and store it in that.

https://prometheus.io/docs/concepts/metric_types/#histogram

Exactly the same as duration above, but instead you're capturing how many retries.

vicentefb: Ahhh okok my bad, I see what you mean, I changed it and now this is the results i get in Prometheus: 
image

markmandel: That looks better 👍🏻

pkg/gameserverallocations/metrics.go
vicentefb: Ohhh i think i see why, it first records the metrics and then it does a lot of manipulation with the tags for example: [latency.setRequest()](https://github.com/googleforgames/agones/blob/049de2fbbe8d548e68a79e548dd0d05bb37d8636/pkg/gameserverallocations/allocator.go#L200) mutates [tags](https://github.com/googleforgames/agones/blob/049de2fbbe8d548e68a79e548dd0d05bb37d8636/pkg/gameserverallocations/metrics.go#L90) same case with [latency.setResponse(out) ](https://github.com/googleforgames/agones/blob/049de2fbbe8d548e68a79e548dd0d05bb37d8636/pkg/gameserverallocations/allocator.go#L235C10-L235C21) I can probably merge latency.record() and latency.setRequest(gsa) but not sure about latency.setResponse(out) since that happens later in the function and the code can return before that. Wdyt ?

pkg/gameserverallocations/allocator.go
```diff
@@ -250,13 +250,20 @@ func (c *Allocator) loggerForGameServerAllocation(gsa *allocationv1.GameServerAl
 }
 
 // allocateFromLocalCluster allocates gameservers from the local cluster.
+// Add here  a a number of times we are entering
```
gongmax, line 253: Typo in this comment?

pkg/gameserversets/controller.go
```diff
@@ -751,3 +761,17 @@ func aggregateLists(aggListStatus map[string]agonesv1.AggregatedListStatus,
 
 	return aggListStatus
 }
+
+// newMetrics creates a new gsa latency recorder.
```
gongmax, line 765: typo: "a new gss latency..."

pkg/gameserversets/metrics.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2019 Google LLC All Rights Reserved.
```
gongmax, line 1: 2024

pkg/gameserversets/metrics.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type metrics struct {
+	ctx              context.Context
+	gameServerLister listerv1.GameServerLister
+	logger           *logrus.Entry
+	start            time.Time
+}
+
+// record the current allocation latency.
```
gongmax, line 77: current gameserver creation latency

pkg/gameserverallocations/metrics.go
```diff
@@ -123,3 +142,9 @@ func (r *metrics) setResponse(o k8sruntime.Object) {
 func (r *metrics) record() {
 	stats.Record(r.ctx, gameServerAllocationsLatency.M(time.Since(r.start).Seconds()))
 }
+
+// record the current allocation retry rate.
+func (r *metrics) recordAllocationErrorRate(ctx context.Context, retryCount int) {
```
gongmax, line 147: If I understand correctly the metrics is tracking the retry count when the retry succeed. So the name of the function is a little confusing since it is not error rate.

pkg/gameserverallocations/metrics.go
```diff
@@ -32,23 +32,42 @@ import (
...
+			Name:        "gameserver_allocations_duration_seconds",
+			Measure:     gameServerAllocationsLatency,
+			Description: "The distribution of gameserver allocation requests latencies.",
+			Aggregation: view.Distribution(0, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2, 3),
+			TagKeys:     []tag.Key{keyFleetName, keyClusterName, keyMultiCluster, keyStatus, keySchedulingStrategy},
+		},
+		{
+			Name:        "gameserver_allocations_retry_total",
+			Measure:     gameServerAllocationsRetryTotal,
+			Description: "The count of gameserver allocation retry success rate",
```
gongmax, line 60: Similarly, would it be more clear if the description is something like "The count of gameserver allocation retry until it succeeds"? since we are not calculating the rate.

pkg/gameserverallocations/metrics.go
vicentefb: I can change it to `recordAllocationRetryRate`

pkg/gameserverallocations/allocator.go
```diff
@@ -250,13 +250,20 @@ func (c *Allocator) loggerForGameServerAllocation(gsa *allocationv1.GameServerAl
 }
 
 // allocateFromLocalCluster allocates gameservers from the local cluster.
+// Registers number of times we retried before getting a success allocation
 func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocationv1.GameServerAllocation) (*allocationv1.GameServerAllocation, error) {
 	var gs *agonesv1.GameServer
+	latency := c.newMetrics(ctx)
```
gongmax, line 256: nit: maybe better to rename it to something like `retry` since this metrics has nothing to do with latency

site/content/en/docs/Guides/metrics.md
```diff
@@ -77,6 +78,48 @@ Follow the [Google Cloud Monitoring installation steps](#google-cloud-monitoring
...
+
+| Name                                                  | Description                                                                                                                                                                                 | Type      |
+|-------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------|
+| agones_gameservers_count                              | The number of gameservers per fleet and status                                                                                                                                              | gauge     |
+| agones_gameserver_allocations_duration_seconds        | The distribution of gameserver allocation requests latencies                                                                                                                                | histogram |
+| agones_gameserver_allocations_retry_total             | The count of gameserver allocation retry success rate | histogram |
```
gongmax, line 89: Please update this to align with the description in `gameserverallocations/metrics.go`

pkg/gameserversets/metrics.go
```diff
@@ -0,0 +1,99 @@
...
+	keyType      = mt.MustTagKey("type")
+
+	gameServerCreationDuration = stats.Float64("gameserver_creation/duration", "The duration of gameserver creation", "s")
+)
+
+func init() {
+
+	stateViews := []*view.View{
+		{
+			Name:        "game_server_creation_duration",
```
gongmax, line 46: Does this metrics has a corresponding description in the `metrics.md` file? I may miss it.

site/content/en/docs/Guides/metrics.md
vicentefb: Sounds good!

pkg/gameserversets/metrics.go
vicentefb: It's not and it's actually strange because this is the only metric that has an underscore in between for gameserver. I will change it to be `agones_gameserver_creation` to follow the same convention as the others as well and will add it to the `metrics.md`

gongmax (approved):

PR #3929, opened 2024-07-30
feat: Adds autoscaling logic for new Chain and Schedule policies

indexjoseph:
**What type of PR is this?**
/kind feature

This PR is a follow up of #3893.

- Adds controller logic to check for active schedules and apply policies under a Schedule
- Adds controller logic for new Chain policy, including fall through for webhook invocation failure and inactive schedules 

Towards #3008


pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -362,6 +367,97 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+
+	location, _ := time.LoadLocation(s.ActivePeriod.Timezone)
+	startCron, _ := cron.ParseStandard(s.ActivePeriod.StartCron)
+	nextStartTime := startCron.Next(currentTime.In(location)).Add(cronDelta)
+	duration, err := time.ParseDuration(s.ActivePeriod.Duration)
+
+	// If there's an err, then the duration field is empty, meaning duration is indefinite
+	if err != nil {
+		duration = 0 // Indefinite duration if not set
+	}
```
vicentefb, line 450: Is there a specific type of `err` that we are expecting to see here. I think it might more readable and safe to also check something like `err == ErrDurationFieldNotEmpty`

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -362,6 +367,97 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+
+	// If no startCron field is specified, then it's automatically true (duration is no longer relevant since we're always running)
+	if s.ActivePeriod.StartCron == "" {
+		return true
+	}
+
+	location, _ := time.LoadLocation(s.ActivePeriod.Timezone)
+	startCron, _ := cron.ParseStandard(s.ActivePeriod.StartCron)
+	nextStartTime := startCron.Next(currentTime.In(location)).Add(cronDelta)
```
vicentefb, line 444: All these fields are validated previously, right ?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -362,6 +367,97 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+
+	// Loop over all entries in the chain
+	for _, entry := range c {
+		switch entry.Type {
+		case autoscalingv1.SchedulePolicyType:
+			replicas, limited, err = applySchedulePolicy(entry.Schedule, f, gameServerLister, nodeCounts, currentTime)
+			// If no error was returned from the schedule policy (schedule is active), then return the values given
+			if err == nil {
+				return replicas, limited, nil
+			}
```
vicentefb, line 401: Should we log in case of an error ?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -362,6 +367,97 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+		case autoscalingv1.WebhookPolicyType:
+			replicas, limited, err := applyWebhookPolicy(entry.Webhook, f)
+			// If no error was returned from the webhook policy, then return the values given
+			if err == nil {
+				return replicas, limited, nil
+			}
```
vicentefb, line 407: Same

pkg/fleetautoscalers/fleetautoscalers.go
indexjoseph: Yeah, duration is already validated for every case except if it were empty, so the only error that can be returned here is if the duration is empty. I'll change the clause to `s.ActivePeriod.Duration == ""`  for it to be more readable.

pkg/fleetautoscalers/fleetautoscalers.go
indexjoseph: Yep, both the location and startCron fields are validated prior to this.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -362,6 +367,97 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+	nextStartTime := startCron.Next(currentTime.In(location)).Add(cronDelta)
+	duration, err := time.ParseDuration(s.ActivePeriod.Duration)
+
+	// If there's an err, then the duration field is empty, meaning duration is indefinite
+	if err != nil {
+		duration = 0 // Indefinite duration if not set
+	}
+
+	// If the current time is after the next start time, and the duration is indefinite or the current time is before the next start time + duration,
+	// then return true
```
vicentefb, line 453: This comment is great! Consider adding a small example with hard coded values, something like `currentTime = 3999-06-15T15:59:59Z` and `nextStartTime = ....` etc..

vicentefb (commented):
Nice

pkg/fleetautoscalers/fleetautoscalers.go
indexjoseph: In the case where there's an error here, it's either A. schedule is inactive or B. schedule contains a webhook policy that return an error. If there's only a schedule policy/schedule policy then the last schedule failed will be logged by the controller.

This leads me to ask if an inactive schedule should return an error or should we be checking if it's inactive here. I was trying to avoid calling `isScheduleActive` twice (one in chain policy to check if we should apply it, and another in the actual method itself in the case where scheduled is a policy by itself) by returning an error.

But to answer your question, yes I think it should be logged regardless. I can pass the controller's event logger through to the fleetautoscalers `computeDesiredFleetSize` method and record events within the `applyChainPolicy` and `applySchedulePolicy` methods.

How does that sound? @markmandel @zmerlynn

pkg/fleetautoscalers/fleetautoscalers.go
indexjoseph: Sgtm!

indexjoseph: I've done the documentation but now that I think about this, there's an issue with the way I'm checking how long the policy should be applied. Since the `nextStartTime` will always be moving forward in time, since the duration is added to the nextStartTime the actual period of time that the policy is applied will keep getting extended.

pkg/fleetautoscalers/fleetautoscalers.go
zmerlynn: I would shy away from logging "normal" events as error, especially since you'd be logging it every sync interval. IMO, logging here can be delegated to the individual `apply` functions, since they have better knowledge of what's abnormal - e.g. `applySchedulePolicy` can log if there's some hard error not related to the schedule simply being inactive, but `applyWebhookPolicy` wil probably always log the error, since we don't expect it.

FWIW, if these were functions in separate packages, I might answer differently, but it's not hard within the same package just to have logging consistently done at one level or the other.

pkg/fleetautoscalers/fleetautoscalers.go
zmerlynn: (Mentioned this case above)

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -362,6 +367,97 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+	}
+
+	// Fall off the chain
+	return replicas, limited, err
+}
+
+// isScheduleActive checks if a chain entry's is active and returns a boolean, true if active, false otherwise
+func isScheduleActive(s *autoscalingv1.SchedulePolicy, currentTime time.Time) bool {
+	// When retrieving the next available cronTime it returns the next minute e.g. if you want a policy to run every minute (all the time)
+	// If the current time is 1:00, the next start time returned will be 1:01, thus we use a cronDelta of 1 minute and 30 secpmds to
```
zmerlynn, line 421: *seconds

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -362,6 +367,97 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
...
+	// adjust to the minute delta (since the next start time will always be ahead of the current time by 1 minute).
+	cronDelta := (time.Minute * -1) + (time.Second * -30)
```
zmerlynn, line 423: I'm kind of ambivalent but I would tend to just call this 90seconds, here and the comment, just for concision. I think you can assume people can translate 90 -> 1:30.

pkg/fleetautoscalers/fleetautoscalers.go
zmerlynn: you might document this - as a code reader I stumbled as to why it was okay to ignore the error.

a couple of other places we also just catch the error and log a bit error saying e.g. "Timezone already validated, unexpected error from time.LoadLocation: %v". Adding checks like that can serve as documentation instead of comments.

zmerlynn (commented):
(Haven't reviewed test yet, doing that new).

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -2134,3 +2136,615 @@ func TestApplyListPolicy(t *testing.T) {
...
+			sp:           &autoscalingv1.SchedulePolicy{},
+			want: expected{
+				replicas: 0,
+				limited:  false,
+				wantErr:  true,
+			},
+		},
+		"no start time": {
+			featureFlags: string(utilruntime.FeatureScheduledAutoscaler) + "=true",
+			now:          mustParseTime("2021-01-00T00:00:00Z"),
```
zmerlynn, line 2173: what day is 2021-01-00? 🤔

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -2134,3 +2136,615 @@ func TestApplyListPolicy(t *testing.T) {
...
+				wantErr:  false,
+			},
+		},
+		"no start time, end time, cron time, duration": {
+			featureFlags: string(utilruntime.FeatureScheduledAutoscaler) + "=true",
+			now:          mustParseTime("2021-01-01T00:00:00Z"),
+			sp: &autoscalingv1.SchedulePolicy{
+				Policy: autoscalingv1.FleetAutoscalerPolicy{
+					Type: autoscalingv1.BufferPolicyType,
+					Buffer: &autoscalingv1.BufferPolicy{
```
zmerlynn, line 2285: I found this difficult to read for a silly reason: the numbers kept changing. If I could make a suggestion here, just create a constant `BufferPolicy` at the start of the function, then use it throughout - you don't really need to test anything about the buffer policy except whether it's applied or not. Then here in `Buffer`, use the constant value, and in `want`, you can make two new constants: `expectedWhenActive` and `expectedWhenNotActive`, or something like that. Then the reader just sees:

```
  Buffer: buffer Policy
}
want: expectedWhenActive
```

Does that make sense?

zmerlynn (commented):
All looks generally LGTM, just a few nits really.

pkg/fleetautoscalers/fleetautoscalers_test.go
indexjoseph: Yeah that makes sense.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -49,18 +52,29 @@ var client = http.Client{
 	},
 }
 
+// InactiveScheduleError denotes an error for schedules that are not currently active.
+type InactiveScheduleError struct{}
+
+func (m *InactiveScheduleError) Error() string {
```
zmerlynn, line 58: no reason to make this a pointer receiver, you can literally just do:

```
func (InactiveScheduleError) Error() string {
```

together with the `Is` it looks like: https://go.dev/play/p/u-QmYmCFfL_x

pkg/fleetautoscalers/fleetautoscalers.go
zmerlynn: you resolved this, it looks like

pkg/fleetautoscalers/controller.go
```diff
@@ -313,8 +313,10 @@ func (c *Controller) syncFleetAutoscaler(ctx context.Context, key string) error
 	}
 
 	currentReplicas := fleet.Status.Replicas
-	desiredReplicas, scalingLimited, err := computeDesiredFleetSize(fas, fleet, c.gameServerLister, c.counter.Counts())
-	if err != nil {
+	desiredReplicas, scalingLimited, err := computeDesiredFleetSize(fas.Spec.Policy, fleet, c.gameServerLister, c.counter.Counts())
+	var scheduleErr InactiveScheduleError
+	// If there err is nil and not an inactive schedule error (ignorable in this case), then record the event
+	if err != nil && !errors.Is(err, &scheduleErr) {
```
zmerlynn, line 319: kill line 317 and just initialize inline: https://go.dev/play/p/u-QmYmCFfL_x

zmerlynn (approved):
Nice job!

PR #3940, opened 2024-08-10
Add a CI check to fail on change to an example without a new version

### LINKED ISSUE #3794: Fail CI if a PR updates an example without modifying the version

wheatear-dev:
**What type of PR is this?**

/kind feature

Further to issue #3794, if any files change within an example - outside of a simple denylist - and the version in its Makefile is not updated, this new tool will fail the CI pipeline.

Closes #3794

Hi 👋 

It wasn't obvious where or how to add automated tests for this script. My _manual_ strategy was as follows:

1. Change at least one (non-**denylist**) file, in two relevant examples
2. Run the example in its directory, observing that it fails the "CI"
3. Increase the version of one of the two examples
4. Run the example again in its directory, observing that it still fails the "CI", but now only on one example.
5. Ensure the version of both examples are updated
6. Run again, and observe that it succeeds.

N.B. the above changes have to be committed each time, otherwise they are invisible to this script.


wheatear-dev:
Two _snags_ to be addressed:

1. If a new example is added, this will currently fail. Need to be able to return a `nil` version on the "remote" tree.
2. This is not - _yet_ - running in the CI. An investigation is required! 😅

build/scripts/example-version-checker/version.go
```diff
@@ -0,0 +1,104 @@
...
+
+	if err := scanner.Err(); err != nil {
+		log.Fatalf("Could not get version: %v", err)
+	}
+
+	return "", errors.New("no version could be found")
+}
+
+func lineContainsVersion(line string) bool {
+	return strings.HasPrefix(line, "version")
```
gongmax, line 63: nit: better to make it stricter i.e. `strings.HasPrefix(line, "version :=")

gongmax (approved):

build/scripts/example-version-checker/version.go
wheatear-dev: Hi. 👋 

Do you want me to action this change now?

You appear to be operating on this branch, and I have no desire to interfere. 😄

gongmax: Yeah feel free to go ahead and make the change. Thanks

PR #3950, opened 2024-08-14
Make the parameters that limits the number of GameServers to add configurable

vicentefb:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

There's some [parameters](https://github.com/googleforgames/agones/blob/205e87ea1c29930324a45d070499bbc561af1f06/pkg/gameserversets/controller.go#L62-L66) inside the gameserverset controller that can be configured by the user

Closes #


install/helm/agones/values.yaml
```diff
@@ -134,6 +138,10 @@ agones:
     numWorkers: 100
     apiServerQPS: 400
     apiServerQPSBurst: 500
+    maxCreationParalellismFlag: 16
+    maxGameServerCreationsPerBatch: 64
+    maxDeletionParallelismFlag: 64
+    maxGameServerDeletionsPerBatch: 64
```
gongmax, line 144: These are not needed since this blob is for extension

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -184,6 +184,10 @@ The following tables lists the configurable parameters of the Agones chart and t
...
+| `agones.controller.maxGameServerCreationsPerBatch`                           | Maximum number of GameServer creation calls per batch                                                                                                                                                      | `64`                                   |
+| `agones.controller.maxDeletionParallelism`                           | Maximum number of parallelizing deletion calls in GSS                                                                                                                                                      | `64`                                   |
+| `agones.controller.maxGameServerDeletionsPerBatch`                           | Maximum number of GameServer deletion calls per batch                                                                                                                                                      | `64`                                   |
```
gongmax, line 190: These should only be visible after next release, so the whole configurations need to be wrapped by 
```
{{% feature publishVersion="1.43.0" %}}
...
{{% /feature %}}
```

For example, https://github.com/googleforgames/agones/blame/26b17c389a466097353e56cd6d82f73f2fd32516/site/content/en/docs/Installation/Install%20Agones/helm.md#L151-L226

gongmax (commented):
Did you test install with helm and specify these configs and check whether they applied?

install/helm/agones/values.yaml
vicentefb: Oh that's true, just noticed that, thanks!

vicentefb:
Values used for the parameters:
image


Controller pod logs using those values:
image

gongmax (approved):

PR #3957, opened 2024-08-20
add metadata to agones webhook autoscaler request

### LINKED ISSUE #3951: Extend Webhook autoscaler to send fleet metadata with the request
We are running our fleets throughout different clusters and independently of one and other and use the webhook autoscaler to make sure that we have necessary sizes on all the fleets.
With this solution we are allocating game servers based on labels and not on fleet names, this means that on the `webhook autoscaler` endpoint we have no idea to determine the size since the request does not contain any metadata at all regarding the fleet in question.

We want to extend the `FleetAutoscaleRequest` to include the fleet metadata when reporting to the service(s).

We made changes to the URL used for the `webhook` but that means that we have to update the URL every time we add new label and/or annotations

swermin:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix
> /kind release

This PR extends the webhook autoscaler with metadata, so that the the service retrieving the request can use the lables/annotations to do the calculation

Closes  #3951

### DIFF
--- a/pkg/apis/autoscaling/v1/fleetautoscaler.go
+++ b/pkg/apis/autoscaling/v1/fleetautoscaler.go
@@ -316,8 +316,10 @@ type FleetAutoscaleRequest struct {
 	Namespace string `json:"namespace"`
 	// The Fleet's status values
 	Status agonesv1.FleetStatus `json:"status"`
-	// Standard object metadata; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata.
-	MetaData *metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Standard map labels; More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels.
+	Labels map[string]string `json:"labels,omitempty"`
+	// Standard map annotations; More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations.
+	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
 // FleetAutoscaleResponse defines the response of webhook autoscaler endpoint
--- a/pkg/apis/autoscaling/v1/fleetautoscaler.go
+++ b/pkg/apis/autoscaling/v1/fleetautoscaler.go
@@ -316,6 +316,8 @@ type FleetAutoscaleRequest struct {
 	Namespace string `json:"namespace"`
 	// The Fleet's status values
 	Status agonesv1.FleetStatus `json:"status"`
+	// Standard object metadata; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata.
+	MetaData *metav1.ObjectMeta `json:"metadata,omitempty"`
 }
 
 // FleetAutoscaleResponse defines the response of webhook autoscaler endpoint
--- a/site/content/en/docs/Guides/feature-stages.md
+++ b/site/content/en/docs/Guides/feature-stages.md
@@ -37,7 +37,8 @@ The current set of `alpha` and `beta` feature gates:
 | [Rolling Update Fixes](https://github.com/googleforgames/agones/issues/3688)                                                | `RollingUpdateFix`                 | Disabled | `Alpha` | 1.41.0 |
 | [Multiple dynamic port ranges](https://github.com/googleforgames/agones/issues/1911)                                        | `PortRanges`                       | Disabled | `Alpha` | 1.41.0 |
 | [Port Policy None](https://github.com/googleforgames/agones/issues/3804)                                                    | `PortPolicyNone`                   | Disabled | `Alpha` | 1.41.0 |
-| [Scheduled Fleet Autoscaling](https://github.com/googleforgames/agones/issues/3008)                                         | `ScheduledAutoscaler`              | Disabled | `Alpha` | 1.43.0 |
+| [Scheduled Fleet Autoscaling](https://github.com/googleforgames/agones/issues/3951)                                         | `ScheduledAutoscaler`              | Disabled | `Alpha` | 1.43.0 |
+| [Extend Webhook autoscaler to send fleet metadata with the request](https://github.com/googleforgames/agones/issues/3008)   | `FleetAutoscaleRequestMetaData`    | Disabled | `Alpha` | 1.47.0 |
 | Example Gate (not in use)                                                                                                   | `Example`                          | Disabled | None    | 0.13.0 |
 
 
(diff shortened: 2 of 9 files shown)


igooch (commented):
Thanks for the contribution. Could you please also add [unit](https://github.com/googleforgames/agones/blob/main/pkg/fleetautoscalers/fleetautoscalers_test.go) and [e2e](https://github.com/googleforgames/agones/blob/main/test/e2e/fleetautoscaler_test.go) tests?

gongmax:
@swermin kindly ping on this, could you resolve the conflicts and address the above comments?

swermin:
Yes ofc! I have unit tests ready and just left with the e2e tests. I’ll push the current changes and focus on the e2e tests today and this weekend if needed!

swermin:
@igooch do you have any good suggestions ok how to do the e2e tests for this?
Current default webhook autoscaler is using the example image to test the scaling. I feel like adding support in that image feels a bit weird but I have no good idea on how to add something

igooch:
I just got back from being out of office, so I'm still catching up.

For the e2e tests, do you need to make changes to `examples/autoscaler-webhook/main.go`? Depending on the test you may be able to simply add MetaData to the default in https://github.com/googleforgames/agones/blob/c26f7371980418b56b816b13d265c9cc9d092f32/test/e2e/fleetautoscaler_test.go#L549C14-L549C38. If you do need to make changes to the `examples/autoscaler-webhook/main.go`, then a maintainer will need to build and push a new example autoscaler webhook image with the changes. That being said, it looks like the unit test is failing.

From the most recent build:

```
        	Test:       	TestApplyWebhookPolicyWithMetadata
        	            	actual  : 5
        	            	expected: 11
        	Error:      	Not equal: 
        	Error Trace:	/go/src/agones.dev/agones/pkg/fleetautoscalers/fleetautoscalers_test.go:618
    fleetautoscalers_test.go:618: 
--- FAIL: TestApplyWebhookPolicyWithMetadata (0.01s)
```

You'll need to run `~/agones/build$ make gen-crd-code` to generate the deepcopy for `pkg/apis/autoscaling/v1/fleetautoscaler.go`, but even with that the test is still failing, so there's another issue going on.

markmandel:
Added a [note on the issue](https://github.com/googleforgames/agones/issues/3951#issuecomment-2629071367) as well, but my suggestion is that we let this PR through as is (with the above docs changes), and then follow up with e2e test PR.

swermin:
> Added a [note on the issue](https://github.com/googleforgames/agones/issues/3951#issuecomment-2629071367) as well, but my suggestion is that we let this PR through as is (with the above docs changes), and then follow up with e2e test PR.

Biggest problem with the e2e test is that I need to specify a version of the container to use. So there is no way for me to do the change without having to do two different PRs (at least what I know of)

I tried multiple different ways of getting a e2e test in without having to separate the builds but I cannot seem to find a solution :(

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -37,7 +37,8 @@ The current set of `alpha` and `beta` feature gates:
 | [Rolling Update Fixes](https://github.com/googleforgames/agones/issues/3688)                                                | `RollingUpdateFix`                 | Disabled | `Alpha` | 1.41.0 |
 | [Multiple dynamic port ranges](https://github.com/googleforgames/agones/issues/1911)                                        | `PortRanges`                       | Disabled | `Alpha` | 1.41.0 |
 | [Port Policy None](https://github.com/googleforgames/agones/issues/3804)                                                    | `PortPolicyNone`                   | Disabled | `Alpha` | 1.41.0 |
-| [Scheduled Fleet Autoscaling](https://github.com/googleforgames/agones/issues/3008)                                         | `ScheduledAutoscaler`              | Disabled | `Alpha` | 1.43.0 |
+| [Scheduled Fleet Autoscaling](https://github.com/googleforgames/agones/issues/3951)                                         | `ScheduledAutoscaler`              | Disabled | `Alpha` | 1.43.0 |
```
markmandel, line 40: feature shortcode here too please, and I think this will be good to go 👍🏻

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -316,6 +316,8 @@ type FleetAutoscaleRequest struct {
 	Namespace string `json:"namespace"`
 	// The Fleet's status values
 	Status agonesv1.FleetStatus `json:"status"`
+	// Standard object metadata; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata.
+	MetaData *metav1.ObjectMeta `json:"metadata,omitempty"`
```
markmandel, line 320: 🤔  I'm looking at this again, and wondering if this should be all the ObjectMeta - which is a lot, and we copy a chunk of it here already (Name and Namespace for example) - so we are also duplicating some effort.

Should it just be labels and annotations? Do we need everything?

Admittedly, in the ticket it says "metadata" but in my head I thought of it as "labels and annotations" and didn't clarify.

WDYT?

swermin: While the object is still called `MetaData` we only ever copy over annotations, labels, and name. Do you want it to be more explicit in that the fields should be called `Annotations` and `Labels`?
```
if runtime.FeatureEnabled(runtime.FeatureFleetAutoscaleRequestMetaData) {
	faReq.Request.MetaData = &metav1.ObjectMeta{
		Name:        f.ObjectMeta.Name,
		Labels:      f.ObjectMeta.Labels,
		Annotations: f.ObjectMeta.Annotations,
	}
}
```

swermin: I knew that I did think of the whole metadata copying and the object itself being too big

site/content/en/docs/Guides/feature-stages.md
swermin: I've added the feature gate here, how to I test it in a good way? You did mention that it may require me copying the whole table

pkg/apis/autoscaling/v1/fleetautoscaler.go
swermin: That is a fair point, I’ve updated the code to reflect this

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -38,6 +38,9 @@ The current set of `alpha` and `beta` feature gates:
 | [Multiple dynamic port ranges](https://github.com/googleforgames/agones/issues/1911)                                        | `PortRanges`                       | Disabled | `Alpha` | 1.41.0 |
 | [Port Policy None](https://github.com/googleforgames/agones/issues/3804)                                                    | `PortPolicyNone`                   | Disabled | `Alpha` | 1.41.0 |
 | [Scheduled Fleet Autoscaling](https://github.com/googleforgames/agones/issues/3008)                                         | `ScheduledAutoscaler`              | Disabled | `Alpha` | 1.43.0 |
+{{% feature publishVersion="1.48.0" %}}
```
markmandel, line 41: Mid table items tend to render badly, you may need to copy the whole table. But I'll run it and we can see what it looks like in preview.

(`make site-server` will give you a local version if you would like to try it).

markmandel:
/gcbrun

Thanks for your patience!

site/content/en/docs/Guides/feature-stages.md
markmandel: image

https://1d43bbe-dot-preview-dot-agones-images.appspot.com/site/docs/guides/feature-stages/#feature-gates

Going to have to copy the whole block, but otherwise, I think this is good to go.

swermin:
I don’t understand why the build failed 🧐

markmandel:
Looks like a flake in https://console.cloud.google.com/cloud-build/builds/2e11f722-e322-4bd7-a76f-9579725cdde2;step=26?project=agones-images `submit-upgrade-test-cloud-build`. @igooch you might want to look into that.

markmandel (approved):

PR #3933, opened 2024-08-07
Update Supported Kubernetes to 1.28, 1.29, 1.30

ashutosji:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

/kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**What this PR does / Why we need it**:

**Which issue(s) this PR fixes**:https://github.com/googleforgames/agones/issues/3919

Closes #3919 

**Special notes for your reviewer**:

PR #3959, opened 2024-08-20
remove ctx within the condition func

peterzhongyi:
The `ctx` and `cancelCtx` was added to avoid `helper.DeleteAgonesPod` from failing from deadline exceeded. However, it doesn't work as desired because `ctx` can get cancelled by `retryTimeout`, which will in turn fail the `grpc.Allocate(ctx, request)`. We actually don't even need to call `helper.DeleteAgonesPod` thanks to the namespace clean up in main_test.go,  so I removed it, along with the `ctx` and `cancelCtx`.

Tested by patching #3933 and running the test.

Towards #3919

### DIFF
--- a/test/e2e/allocator/pod_termination_test.go
+++ b/test/e2e/allocator/pod_termination_test.go
@@ -92,16 +92,10 @@ func TestAllocatorAfterDeleteReplica(t *testing.T) {
 
 	// Wait and keep making calls till we know the draining time has passed
 	_ = wait.PollUntilContextTimeout(context.Background(), retryInterval, retryTimeout, true, func(ctx context.Context) (bool, error) {
-		ctx, cancelCtx := context.WithTimeout(ctx, retryInterval*2)
-		defer cancelCtx()
-
-		response, err := grpcClient.Allocate(ctx, request)
+		response, err := grpcClient.Allocate(context.Background(), request)
 		logger.Infof("err = %v (code = %v), response = %v", err, status.Code(err), response)
-		if assert.NoError(t, err, "Failed grpc allocation request") {
-			helper.ValidateAllocatorResponse(t, response)
-			err = helper.DeleteAgonesPod(ctx, response.GameServerName, framework.Namespace, framework)
-			assert.NoError(t, err, "Failed to delete game server pod %s", response.GameServerName)
-		}
+		helper.ValidateAllocatorResponse(t, response)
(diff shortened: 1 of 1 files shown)


gongmax (commented):
So this will leave the Agones Pod running until next set of tests start to run. Can we modify the `main_test.go` so it performs the cleanup not only before the test but also after the test?

peterzhongyi:
> So this will leave the Agones Pod running until next set of tests start to run. Can we modify the `main_test.go` so it performs the cleanup not only before the test but also after the test?

I believe the main_test.go also performs a clean up after the test (it performs a lot of clean ups): https://github.com/googleforgames/agones/blob/a348312bbe5d3dfb134a4c7b1fb856267c6858c2/test/e2e/allocator/main_test.go#L58-L89

Specifically, checking on the logs, I see 

```
INFO[2024-08-20 21:39:30.925] Namespace 1724189094 is deleted              
ok  	agones.dev/agones/test/e2e/allocator	878.180s
```

at the end of the test, which means framework.DeleteNamespace was called as a deferred step: https://github.com/googleforgames/agones/blob/a348312bbe5d3dfb134a4c7b1fb856267c6858c2/test/e2e/allocator/main_test.go#L69

gongmax:
> > So this will leave the Agones Pod running until next set of tests start to run. Can we modify the `main_test.go` so it performs the cleanup not only before the test but also after the test?
> 
> I believe the main_test.go also performs a clean up after the test (it performs a lot of clean ups):
> 
> https://github.com/googleforgames/agones/blob/a348312bbe5d3dfb134a4c7b1fb856267c6858c2/test/e2e/allocator/main_test.go#L58-L89
> 
> Specifically, checking on the logs, I see
> 
> ```
> INFO[2024-08-20 21:39:30.925] Namespace 1724189094 is deleted              
> ok  	agones.dev/agones/test/e2e/allocator	878.180s
> ```
> 
> at the end of the test, which means framework.DeleteNamespace was called as a deferred step:
> 
> https://github.com/googleforgames/agones/blob/a348312bbe5d3dfb134a4c7b1fb856267c6858c2/test/e2e/allocator/main_test.go#L69
> 
> I was also checking the pods in the namespace 1724189094, and they were all deleted after the test. Screenshot 2024-08-20 at 4 58 43 PM

Yep you're right! I missed that defer part.

gongmax (approved):

PR #3962, opened 2024-08-22
change kubernetes API version to fix broken CI

peterzhongyi:
change kubernetes API version to fix broken CI

PR #3964, opened 2024-08-23
Use Markdown when use k8s-api-version variable

peterzhongyi:
#3962 was submitted to fix the broken CI: https://pantheon.corp.google.com/cloud-build/builds;region=global/ee067e2c-418a-4be3-869e-1cdc31a777fc;step=19?e=13803378&mods=allow_workbench_image_override&project=agones-images

Later #3963 was submitted to use k8s-api-version using string literal.

This PR achieves the same effect as #3963 with Markdown syntax since the files are .md files.


igooch (approved):
LGTM. Making a note that this style is in keeping with the rest of the site documentation.

PR #3965, opened 2024-08-24
Add instructions for running Agones on Minikube with the Windows Docker driver

brightestpixel:
**What type of PR is this?**

/kind documentation

This PR adds new instructions on how to connect to GameServers using Minikube's default Docker driver on Windows and without too much extra setup (no extra services per GameServer, `minikube tunnel`, etc.). This should improve the developer experience for those wanting to use Minikube on Windows to test out their GameServers.

The solution is to publish the ports that Agones uses through the `minikube start` command and let Docker forward connections to Minikube and any GameServers. This workaround has worked well on both Windows systems where I have tried it.

**Which issue(s) this PR fixes**: N/A

I'm not an expert on container networking, so if anyone has comments on this approach I'd love to hear them!


igooch (approved):
LGTM. Thank you for the additional testing and documentation.

Issue #3795, opened 2024-04-19
In-place Agones Upgrades: Testing


zmerlynn:
> [!NOTE]
> Milestone of #3766, which we are seeking feedback on. We will move forward with pieces that seem non-contentious, though.

# Testing

To vet In-place Agones Upgrades, we need a combination of unit testing (which are assumed in all PRs), and end-to-end (e2e) testing. The problem with our current e2es is that they assume a particular configuration and test against it. That's good for what they're testing, but we need a different style of test to test upgrades.

From past experience, the best way I have seen to test upgrades is to be *doing something* and upgrade the system in-place. I propose a system where we keep a cluster under fairly active load, and mutate configuration continuously. As a starting point, imagine three subsystems:

* **Wanderer**: Every ${PERIOD}, changes configurations randomly within a defined space, e.g. upgrade/downgrade within upgrade horizon, change feature flags, etc. I'm thinking every half hour, giving us a reasonable soak period in between as well, which may help with things like soak testing as well (at least for detecting bigger leaks).

* **Producer**: Generates load continuously - scale up/down Fleets at random, interact with GameServers via allocation, etc. Some way to ensure there's load on the system. The Producer should keep a continuous metric for operations it does (e.g. allocations) that we can monitor SLO, so if an upgrade causes dips, we notice.

* **Monitor**: Either configured as alerts, or as an active process monitoring the state of the system, we need to verify that e.g. Agones is healthy, Fleets are healthy, etc.

If we do this right, we could even set this up in a couple of different modes - one with a fast wanderer (e.g. 30 minutes) to make sure we cover the most possible configuration space), and one with a slow wanderer (e.g. a day, a week) to soak test more. The Producer/Monitor for each will look rather similar, just the rate of change will be different.

1804devs:
Some guidance would be helpful, but I've started coding for the Wanderer component. I'm not sure if I'm heading in the right direction.

1804devs:
// Wanderer component
package main

import (
    "bytes"
    "encoding/json"
    "log"
    "net/http"
    "time"
)

type Config struct {
    // Define configuration struct
}

func main() {
    // Define the Agones API endpoint
    agonesEndpoint := "http://agones-api.example.com/config"

    // Implement logic to trigger configuration changes
    for {
        // Generate random configuration changes
        config := generateRandomConfig()

        // Convert configuration to JSON
        configJSON, err := json.Marshal(config)
        if err != nil {
[...]

PR #3982, opened 2024-09-10
Updates upgrade test to install multiple versions of Agones on a cluster in succession

igooch:
**What type of PR is this?**

/kind feature

This creates a basic config walker that installs all versions of Agones that are valid for a cluster on a particular Kubernetes versions. One valid version of Agones is installed, and the controller waits until the Helm release (Agones installation) is ready (deployed) before installing the next version.

Working on #3795

**Special notes for your reviewer:**

The `upgrade-test-runner` Job runs a pod that is effectively a controller uses Helm to install Agones into the `agones-system` namespace on the same cluster. This Job needs to have all of the permissions of both Helm and Agones in order to use and create those resources.

Wait groups do not work for waiting on Kubernetes or Helm, so we use `wait.PollUntilContextTimeout` instead, similar to other e2e tests in Agones.

There are a number of `//TODO`s in the main file denoting pieces that are not included in this particular PR, but will need to be included in future PRs.

To run:

- Have a running Kubernetes cluster
- Have `REGISTRY` set to your artifact registry
- The `$(REGISTRY)` value in upgradeTest.yaml currently needs to be manually set to your registry
- make build && make push to your registry
- `kubectl apply -f permissions.yaml` for reading Pod and Node and installing Agones
- `kubectl apply -f versionMap.yaml` to load the configmap
- `kubectl apply -f upgradeTest.yaml` to run the test runner
- You should be able to see logs in logs explorer with filter `resource.labels.container_name="upgrade-test-controller"`
- Run `kubectl get po -n agones-system` to see the Agones pods. These are expected to crash loop a couple times when running on an Autopilot cluster as the Autopilot cluster needs to scale up to accommodate the pods. 
- Check `kubectl get jobs --watch` to see if the job is still running.
- Once the job is finished you should be able to see all of the previously installed Helm revisions by running `helm history agones -n agones-system`. Example for a Kubernetes version 1.29 cluster:

```
me@me:~/agones/test/upgrade$ helm list -aA
NAME  	NAMESPACE    	REVISION	UPDATED                                	STATUS  	CHART        	APP VERSION
agones	agones-system	4       	2024-09-10 21:24:12.099119645 +0000 UTC	deployed	agones-1.43.0	1.43.0
me@me:~/agones/test/upgrade$ helm history agones -n agones-system
REVISION	UPDATED                 	STATUS    	CHART        	APP VERSION	DESCRIPTION
1       	Tue Sep 10 21:18:53 2024	superseded	agones-1.40.0	1.40.0     	Install complete
2       	Tue Sep 10 21:21:42 2024	superseded	agones-1.41.0	1.41.0     	Upgrade complete
3       	Tue Sep 10 21:23:23 2024	superseded	agones-1.42.0	1.42.0     	Upgrade complete
4       	Tue Sep 10 21:24:12 2024	deployed  	agones-1.43.0	1.43.0     	Upgrade complete
```

### DIFF
--- a/test/upgrade/main.go
+++ b/test/upgrade/main.go
@@ -190,16 +190,6 @@ func runExecCommand(cmd string, args ...string) ([]byte, error) {
 	return out, err
 }
 
-// Install Agones to the cluster using the install.yaml file.
-// kubectl create namespace agones-system
-// kubectl apply --server-side -f https://raw.githubusercontent.com/googleforgames/agones/release-1.43.0/install/yaml/install.yaml
-func installYaml() error {
-	installArgs := []string{"apply", "--server-side", "-f", "https://raw.githubusercontent.com/googleforgames/agones/release-1.43.0/install/yaml/install.yaml"}
-
-	_, err := runExecCommand(KUBECTL_CMD, installArgs...)
-	return err
-}
-
 // Adds public Helm Agones releases to the cluster
 func addAgonesRepo() {
 	installArgs := [][]string{{"repo", "add", "agones", "https://agones.dev/chart/stable"},
--- a/test/upgrade/main.go
+++ b/test/upgrade/main.go
@@ -21,24 +21,37 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
 	"regexp"
 	"strings"
+	"time"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
-func main() {
-	validConfigs := ConfigTestSetup()
-
-	// TODO(igooch): Replace this with the deterministic config walker
-	for i, validConfig := range validConfigs {
-		log.Printf("validConfig %d: %v", i+1, validConfig)
-	}
-}
+const (
+	// HELM_CMD is the command line invocation of helm
+	HELM_CMD = "helm"
+	// KUBECTL_CMD is the command line invocation of kubectl
+	KUBECTL_CMD = "kubectl"
+	// IMAGE_PULL_POLICY sets the Agones Helm configuration to always pull the image
+	IMAGE_PULL_POLICY = "Always"
+	// SIDECAR_PULL_POLICY sets the Agones Helm configuration to always pull the SDK image
+	SIDECAR_PULL_POLICY = "true"
+	// LOG_LEVEL sets the Agones Helm configuration log level
+	LOG_LEVEL = "debug"
+	// HELM_CHART is the helm chart for the public Agones releases
+	HELM_CHART = "agones/agones"
+	// AGONES_REGISTRY is the public registry for Agones releases
+	AGONES_REGISTRY = "us-docker.pkg.dev/agones-images/release"
+)
 
 var (
+	// TODO: Get the build version of dev (i.e. 1.44.0-dev-b765f49)
 	// DEV is the current development version of Agones
 	DEV = os.Getenv("DEV")
 	// POD_NAME the name of the pod this container is running in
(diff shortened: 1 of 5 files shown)


test/upgrade/main.go
```diff
@@ -104,50 +136,164 @@ func findK8sVersion() string {
...
+			SIDECAR_PULL_POLICY, LOG_LEVEL, chart)
+		if err != nil {
+			log.Printf("installAgonesRelease err: %s", err)
 		}
+
+		// Wait for the helm release to install. Waits the same amount of time as the Helm timeout.
+		var helmStatus string
+		err = wait.PollUntilContextTimeout(ctx, 10*time.Second, 10*time.Minute, true, func(ctx context.Context) (done bool, err error) {
+			helmStatus = checkHelmStatus(config.agonesVersion)
+			if helmStatus == "deployed" {
```
gongmax, line 266: I have some trouble to understand the helm doc so I'll just ask. Does a "deployed" helm release necessarily mean the underlying k8s pods are running?

test/upgrade/upgradeTest.yaml
```diff
@@ -20,6 +20,7 @@ metadata:
   labels:
     app: upgrade-test-runner
 spec:
+  backoffLimit: 0
```
gongmax, line 23: Will it be helpful if retry the job a few times in case of transient errors of pod?

test/upgrade/main.go
```diff
@@ -104,50 +136,164 @@ func findK8sVersion() string {
 		log.Fatal("Could not create the kubernetes api clientset", err)
 	}
 
-	ctx := context.Background()
+	// Wait to get pod and node as these may take a while to start on a new Autopilot cluster.
+	var pod *v1.Pod
+	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 7*time.Minute, true, func(ctx context.Context) (done bool, err error) {
```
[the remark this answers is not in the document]
igooch: By definition `findK8sVersion` is running in a container in a pod, but somehow `pod, err = kubeClient.CoreV1().Pods(POD_NAMESPACE).Get(ctx, POD_NAME, metav1.GetOptions{})` runs into errors for ~3 min as the pod is trying to schedule itself onto a node. How this manages to run without yet being scheduled to a Node, and how it can be on a pod and not be able to gets its name I don't quite understand, but that's what the logs are showing. So the PollUntilContextTimeout waits until the kubeClient is able to successfully get the pod name, or if it can't get the pod name after 7 min it errors out.

test/upgrade/main.go
```diff
@@ -104,50 +136,164 @@ func findK8sVersion() string {
...
+
+	return err
+}
+
+func runConfigWalker(ctx context.Context, validConfigs []*configTest) {
+	for _, config := range validConfigs {
+		registry := AGONES_REGISTRY
+		chart := HELM_CHART
+		if config.agonesVersion == DEV {
+			// TODO: Update to templated value for registry and chart for Dev build
```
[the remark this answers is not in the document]
igooch: Don't we want to make sure there are no changes in a PR that break upgrades?

test/upgrade/main.go
igooch: In this case it does, because we run helm with the argument `wait`, which waits until all the underlying pods are running before marking the release as `deployed`.

test/upgrade/upgradeTest.yaml
igooch: I was running into issues (before adding the `PollUntilContextTimeout` to wait for Autopilot to scale up and schedule the pod to a new node), where there end up with multiple pods that were all trying to install Agones, and failing because there would be a previous install stuck from an old pod that had errored out. In theory this would eventually sort itself out, as the Helm command has the `atomic` flag, so it _shoud_ roll back after 10 min of failing to install, but it would seem to get stuck and not automatically roll back, and I would have to manually delete the helm release from the cluster. This might be OK to retry now, but we may still run into helm issues.

gongmax (approved):

PR #3990, opened 2024-09-20
Adds clusters for the in place upgrades tests

igooch:
**What type of PR is this?**

/kind feature

Adds GKE Autopilot and GKE Standard test cluster for use in the In Place Upgrades tests. The clusters follow the same pattern as the existing e2e test clusters. The Upgrades standard cluster do not require as much compute as the e2e clusters, so the initial node count is lower than for the e2e tests.

Working on #3795 

### DIFF
--- a/build/terraform/e2e/module.tf
+++ b/build/terraform/e2e/module.tf
@@ -62,22 +62,50 @@ variable "kubernetes_versions" {
   }
 }
 
+variable "test_names" {
+  description = "Use the same terraform templates for both e2e and upgrade tests. Includes test name and initial node counts for standard clusters."
+  type        = map(number)
+  default     = {
+    "e2e" = 10
+    "upgrade" = 4
+  }
+}
+
+// Handle nested loop in terraform. Flatten combines the two maps into a list.
+locals {
+  test_versions = distinct(flatten([
+    for name, nodes in var.test_names : [
+      for version, val in var.kubernetes_versions : {
+        test = name
+        version = version
+        location = val[0]
+        releaseChannel = val[1]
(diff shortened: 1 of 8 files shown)


build/terraform/e2e/module.tf
```diff
@@ -62,22 +62,50 @@ variable "kubernetes_versions" {
...
+  description = "Use the same terraform templates for both e2e and upgrade tests. Includes test name and initial node counts for standard clusters."
+  type        = map(number)
+  default     = {
+    "e2e" = 10
+    "upgrade" = 4
+  }
+}
+
+// Handle nested loop in terraform. Flatten combines the two maps into a list.
+locals {
```
gongmax, line 75: So now the creation of e2e test clusters and upgrade test clusters are coupled in this module file,  which sounds not convenient. For example, `make gcloud-e2e-test-cluster` will create both set of clusters. Can we move the creation of upgrade test clusters to a separate module file, which still use the same `gke-autopilot/module.tf` and `gke-standard/module.tf`? The structure in my mind is something like below, WDYT?
- build/terraform/test-clusters
  - e2e
    - module.tf
  - upgrade
    - module.tf
  - gke-standard
    - module.tf
  - gke-autopilot
    - module.tf
  - state-bucket
    - main.tf

igooch: Good point, I didn't think about people using this for local e2e creation. I'll break it out into separate modules.

gongmax (approved):
Approved. We can do some refactoring  to reduce the redundant code used for both e2e and upgrade as a follow up.

PR #3991, opened 2024-09-20
Test in place upgrades run tests

igooch:
**What type of PR is this?**

/kind feature

- Creates Game Server that run the `sdk-client-test` through Agones upgrades. The test fails if too many Game Servers end up in a CrashBackOff loop. It is expected that 0-2 Game Servers per version will not be healthy, as the first Game Server created immediately following an Agones installation may fail due to the cluster not being scaled up.
- This tests that the Game Servers continue to function during upgrade, and that Game Servers with an older binary version continue to function with newer SDK versions.
- Templates the Counters and Lists in the game server. We need these values templated because the `sdk-client-test` must have this Counter and List defined in the Game Server when CountsAndLists _is_ enabled, and the Game Server cannot be applied to a cluster if a Counter or List is defined and CountsAndLists _is not_ enabled.

Working on #3795 

The `us-docker.pkg.dev/agones-images/ci/sdk-client-test` still needs to be created, and have the `sdk-client-test` built at each release version noted in the versionMap and pushed to the new repo.


test/sdk/go/sdk-client-test.go
```diff
@@ -105,6 +105,9 @@ func main() {
 		testLists(sdk)
 	}
 
+	// Delay before shutdown to prevent Game Servers from churning too quickly on a running cluster
+	time.Sleep(8 * time.Second)
```
gongmax, line 109: How is this value decided? Just curious whether make it configurable can make it easy to use.

test/upgrade/main.go
```diff
@@ -277,7 +303,14 @@ func runConfigWalker(ctx context.Context, validConfigs []*configTest) {
 			log.Fatalf("PollUntilContextTimeout timed out while attempting upgrade to Agones version %s. Helm Status %s",
 				config.agonesVersion, helmStatus)
 		}
+
+		go createGameServers(cancelCtx, config.gameServerPath)
 	}
+	// Allow some soak time at the last version
```
gongmax, line 309: For the versions that are not the last one, should we also wait for a while after a version is installed and then upgrade to the next version?

test/upgrade/main.go
```diff
@@ -337,3 +370,96 @@ func createGameServerFile(agonesVersion string) string {
 
 	return gsPath
 }
+
+// Create a game server every ten seconds until the context is cancelled. The game server container
```
gongmax, line 374: Do you have a sense how many gameservers are created for each Agones version?

test/upgrade/main.go
```diff
@@ -337,3 +370,96 @@ func createGameServerFile(agonesVersion string) string {
...
+	})
+	if err != nil {
+		log.Fatal("Not able to create AddEventHandler", err)
+	}
+
+	go podInformer.Run(stopCh)
+	if !cache.WaitForCacheSync(stopCh, podInformer.HasSynced) {
+		log.Fatal("Timed out waiting for caches to sync")
+	}
+	<-stopCh
```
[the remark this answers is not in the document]
igooch: I forgot to remove it :)

test/upgrade/main.go
```diff
@@ -337,3 +370,96 @@ func createGameServerFile(agonesVersion string) string {
...
+
+	_, err := podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(_, newObj interface{}) {
+			newPod := newObj.(*v1.Pod)
+			for _, cs := range newPod.Status.ContainerStatuses {
+				if cs.Name == "sdk-client-test" && cs.State.Waiting != nil && cs.State.Waiting.Reason == "CrashLoopBackOff" {
+					log.Printf("cs.State.Waiting.Reason: %s for pod: %s", cs.State.Waiting.Reason, newPod.Name)
+					// Put failed pods into the channel until it's full.
+					select {
+					case stopWatch <- newPod.Name:
```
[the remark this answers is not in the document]
igooch: I didn't see that in testing, although it's possible. I can update.

test/sdk/go/sdk-client-test.go
igooch: From manually testing different values. 5 seconds led to more unhealthy game servers, and same for 10 seconds. It would be nice to have it configurable, but that might be hard to do since the `sdk-client-test` binary is built at a release version (1.38, 1.39, etc.), and not at run time.

test/upgrade/main.go
igooch: It takes ~1 min to install Agones, so ~60 sec / 10 sec per version. We can certainly up that value (I had it at 5 sec on an earlier iteration).

test/upgrade/main.go
igooch: We certainly could do. It takes ~1 min to install a new version, so the Game Servers are installing on the existing version for a certain percentage of that time until the controllers roll over to the new version, but we can increase that time.

test/upgrade/main.go
igooch: I put it back at create every 5 sec, and also added in wait time at each version to increase number game servers created.

test/upgrade/main.go
```diff
@@ -337,3 +375,96 @@ func createGameServerFile(agonesVersion string) string {
...
+					log.Fatalf("Too many Game Server pods in CrashLoopBackOff: %v", failedPods)
+				}
+			}
+		},
+	})
+	if err != nil {
+		log.Fatal("Not able to create AddEventHandler", err)
+	}
+
+	go podInformer.Run(stopCh)
```
gongmax, line 435: I still don't quite get the usage of the `stopCh` here. Seems no one is sending signal to stopCh, so this line is equal to `go podInformer.Run()`? If so when does this go routine get stopped?

igooch: It's a required paramenter for the .Run() function. I think it's used instead of context.Context to stop the informer anywhere in the code. We don't use it since nothing we want the informer to run until the main thread completes.

gongmax: Oh I see, I thought it's optional :)

gongmax (approved):

PR #4000, opened 2024-09-25
Formatting code with gofmt

cuishuang:
**What type of PR is this?**

> /kind cleanup

Formatting code with gofmt

Closes #


cmd/controller/pprof.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build profile
 // +build profile
```
gongmax, line 16: This serves the same purpose of line 15, right?

gongmax: I'm wondering whether we can remove the `// +build profile`

gongmax (approved):

PR #4016, opened 2024-10-12
Allowing list based fleet autoscaler to scale up from 0 replicas

### LINKED ISSUE #3943: Fleet autoscaler with "List" policy throws an error if configured with a fleet with no replicas
**What happened**:

If we setup a fleet without defining "replicas" and setup an autoscaler with "List" policy, the autoscaler throws the following error.
```
Error calculating desired fleet size on FleetAutoscaler simple-game-server-autoscaler. Error: cannot apply ListPolicy as List key MyCustomList does not exist in the Fleet Status
```

The fleet autoscaler should be able to determine the desired number of replicas even if there is no replicas present.

Setup a fleet with no replicas defined, and a list based autoscaler for the fleet. 
Ex.
```
---
apiVersion: agones.dev/v1
kind: Fleet
metadata:
  name: simple-game-server
spec:
  template:
    spec:
      lists:
        MyCustomList:
          capacity: 50
      ports:
        - name: default
          containerPort: 7654
      template:
        spec:
          containers:
            - name: simple-game-server
              image: us-docker.pkg.dev/agones-images/examples/simple-game-server:0.34
              resources:
                requests:
                  memory: 64Mi
                  cpu: 20m
                limits:
                  memory: 64Mi
                  cpu: 20m
---
kind: FleetAutoscaler
metadata:
  name: simple-game-server-autoscaler
spec:
  fleetName: simple-game-server
  policy:
    type: List  # List based autoscaling.
    list:
      key: MyCustomList
      bufferSize: 10%
      minCapacity: 50
      maxCapacity: 800
```

Below is the describe output of the fleet autoscaler
```
Name:         simple-game-server-autoscaler
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  autoscaling.agones.dev/v1
Kind:         FleetAutoscaler
Metadata:
  Creation Timestamp:  2024-08-12T14:35:58Z
  Generation:          1
  Resource Version:    28009
  UID:                 91ff6589-e3c4-4456-b5b4-e31767a8ac03
Spec:
  Fleet Name:  simple-game-server
  Policy:
    List:
      Buffer Size:   10%
      Key:           MyCustomList
      Max Capacity:  800
      Min Capacity:  50
    Type:            List
  Sync:
    Fixed Interval:
      Seconds:  30
    Type:       FixedInterval
Events:
  Type     Reason           Age                 From                        Message
  ----     ------           ----                ----                        -------
  Warning  FleetAutoscaler  12s (x25 over 76s)  fleetautoscaler-controller  Error calculating desired fleet size on FleetAutoscaler simple-game-server-autoscaler. Error: cannot apply ListPolicy as List key MyCustomList does not exist in the Fleet Status
  ```

This is crucial for agones to be used with "Flux CD" (https://fluxcd.io/) and probably other gitops based tools. 

**Environment**:
- Agones version: 1.42
- Kubernetes version (use `kubectl version`): 1.27.1
- Cloud provider or hardware configuration: Minikube
- Others:

### PR DESCRIPTION
**Pull request by geopaulm**
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Updated the controller to initialize the list status map with empty AggregatedListStatus from fleet spec. 
This allows scaling up if replicas is set to 0. Which is usually the case when deployed via Gitops systems ex. Flux.

Additionally updated the `scaleUp` function does limited scale to MinCapacity as well. Added this since I saw the desired capacity being set to Max first and then brought back to min! (This could be dangerous if the max is set to a high value)

I see that in most of the places we round up for calculating the number of replicas. This was missing for limited scaling using lists which is causing values to flap a bit for which I added a fix. Some of the tests were failing. Ex. to have a target max capacity of 45 with capacity of 10, the system was calculating 4 instead of 5. Let me know if that change is okay. 

Closes #3943

### DIFF
--- a/pkg/fleetautoscalers/fleetautoscalers.go
+++ b/pkg/fleetautoscalers/fleetautoscalers.go
@@ -355,7 +355,7 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
 		}
 		return replicas, false, nil
 	case availableCapacity < buffer: // Scale Up
-		if limited && scale == -1 { // Case where we want to scale up but we're already limited by MaxCapacity
+		if limited { // Case where we want to scale up but we're already limited by MaxCapacity or MinCapacity.
 			return scaleLimited(scale, f, gameServerLister, nodeCounts, key, isCounter, replicas,
 				capacity, aggCapacity, minCapacity, maxCapacity)
 		}
@@ -582,14 +582,9 @@ func scaleUp(replicas int32, capacity, count, aggCapacity, availableCapacity, ma
 	additionalReplicas := int32(math.Ceil((float64(buffer) - float64(availableCapacity)) / float64(replicaCapacity)))
 
 	// Check to make sure we're not limited (over Max Capacity)
-	limited, scale := isLimited(aggCapacity+(int64(additionalReplicas)*capacity), minCapacity, maxCapacity)
+	limited, _ := isLimited(aggCapacity+(int64(additionalReplicas)*capacity), minCapacity, maxCapacity)
 	if limited {
-		if scale == -1 {
-			additionalReplicas = int32(math.Ceil((float64(maxCapacity) - float64(aggCapacity)) / float64(capacity)))
-		} else {
-			additionalReplicas = int32(math.Ceil((float64(minCapacity) - float64(aggCapacity)) / float64(capacity)))
-		}
-
+		additionalReplicas = int32((maxCapacity - aggCapacity) / capacity)
 	}
 
 	return replicas + additionalReplicas, limited, nil
--- a/pkg/fleetautoscalers/fleetautoscalers.go
+++ b/pkg/fleetautoscalers/fleetautoscalers.go
@@ -582,9 +582,14 @@ func scaleUp(replicas int32, capacity, count, aggCapacity, availableCapacity, ma
 	additionalReplicas := int32(math.Ceil((float64(buffer) - float64(availableCapacity)) / float64(replicaCapacity)))
 
 	// Check to make sure we're not limited (over Max Capacity)
-	limited, _ := isLimited(aggCapacity+(int64(additionalReplicas)*capacity), minCapacity, maxCapacity)
+	limited, scale := isLimited(aggCapacity+(int64(additionalReplicas)*capacity), minCapacity, maxCapacity)
 	if limited {
-		additionalReplicas = int32((maxCapacity - aggCapacity) / capacity)
+		if scale == -1 {
+			additionalReplicas = int32(math.Ceil((float64(maxCapacity) - float64(aggCapacity)) / float64(capacity)))
+		} else {
+			additionalReplicas = int32(math.Ceil((float64(minCapacity) - float64(aggCapacity)) / float64(capacity)))
+		}
+
 	}
 
 	return replicas + additionalReplicas, limited, nil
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: pkg/fleetautoscalers/fleetautoscalers_test.go [hunk 1]
```diff
@@ -1171,7 +1171,7 @@ func TestApplyCounterPolicy(t *testing.T) {
 				BufferSize:  intstr.FromInt(10),
 			},
 			want: expected{
-				replicas: 14,
+				replicas: 15,
```
igooch, line 1174: This is fine by me. It's really just a question of if someone has a MaxCapacity that's not evenly divisible by the Counter or List Capacity do we want to slightly undershoot the MaxCapacity or slightly overshoot it. I imagine in most cases people will pick evenly divisible numbers.

igooch (commented):
Thanks for submitting the PR and tests!

pkg/fleets/controller.go
```diff
@@ -713,6 +713,14 @@ func (c *Controller) updateFleetStatus(ctx context.Context, fleet *agonesv1.Flee
 	return errors.Wrapf(err, "error updating status of fleet %s", fCopy.ObjectMeta.Name)
 }
 
+func (c *Controller) createInitialListStatus(fleet *agonesv1.Fleet) map[string]agonesv1.AggregatedListStatus {
```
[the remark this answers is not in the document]
geopaulm: Thanks for having a look at the review. 
Yes this sounds good. I have moved this to gameserversets.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -582,9 +582,14 @@ func scaleUp(replicas int32, capacity, count, aggCapacity, availableCapacity, ma
...
 	// Check to make sure we're not limited (over Max Capacity)
-	limited, _ := isLimited(aggCapacity+(int64(additionalReplicas)*capacity), minCapacity, maxCapacity)
+	limited, scale := isLimited(aggCapacity+(int64(additionalReplicas)*capacity), minCapacity, maxCapacity)
 	if limited {
-		additionalReplicas = int32((maxCapacity - aggCapacity) / capacity)
+		if scale == -1 {
+			additionalReplicas = int32(math.Ceil((float64(maxCapacity) - float64(aggCapacity)) / float64(capacity)))
+		} else {
+			additionalReplicas = int32(math.Ceil((float64(minCapacity) - float64(aggCapacity)) / float64(capacity)))
+		}
```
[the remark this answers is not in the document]
geopaulm: In this case limited = true, since aggCapacity = 0 is less than minCapacity. 
Since the scaleup function only does a limit for MaxCapacity, the desired capacity is set to max value. 
In the next run it will be set back to min value!. I saw this in the fleet autoscaler log. Isn't that dangerous?

pkg/fleetautoscalers/fleetautoscalers_test.go
geopaulm: Yeah you are right. But there are some places in the code where we use Ceil. Due to this I saw some flapiness in auto scaling events. 

Example : If the calculated value was say 3.4 the desired capacity will be first set to 3 and then it will be immediately set to 4 in the next run. Isn't it better to have this consistent?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -355,7 +355,7 @@ func applyCounterOrListPolicy(c *autoscalingv1.CounterPolicy, l *autoscalingv1.L
 		}
 		return replicas, false, nil
 	case availableCapacity < buffer: // Scale Up
-		if limited && scale == -1 { // Case where we want to scale up but we're already limited by MaxCapacity
+		if limited { // Case where we want to scale up but we're already limited by MaxCapacity or MinCapacity.
```
igooch, line 358: Nit:

```suggestion
		if limited { // Case where we want to scale up but we're already limited by MaxCapacity.
```

igooch (approved):
LGTM, thanks for all your work on this!

PR #4022, opened 2024-10-23
Set externalTrafficPolicy as Local for agones-allocator

### LINKED ISSUE #4019: Set externalTrafficPolicy as Local for agones-allocator
When using two node pools in an Agones cluster, one for Agones and one for GameServers, allocation requests sometimes fail during the process of reducing nodes in the GameServers’ node pool, especially when reducing many nodes. I think it shouldn't be affected by changes in the GameServers’ node pool.

Set the externalTrafficPolicy for the agones-allocator service to Local from Cluster(default).
https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer?hl=ja#health_check

**Additional context**

osterante:
**What type of PR is this?**
/kind feature

Closes #4019


gongmax:
Discussed in the issue and thanks for the patience. Can you keep the default value unchanged for backward compatibility? And add a description for it [here](https://github.com/googleforgames/agones/blob/main/site/content/en/docs/Installation/Install%20Agones/helm.md?plain=1#L232)? It will need to be in a `{{% feature publishVersion="1.46.0" %}}` block similar as [this](https://github.com/googleforgames/agones/commit/a78b18bc7d1d774dfaab9fc47ee8e78eeb7b2bd7#diff-481b1839587cfc3a359e5e82a78b375445ebe75fe39d5c070150bc4f3a2f51adR153), since we are release 1.45 today.

gongmax (approved):

PR #4037, opened 2024-11-18
Integrates upgrades tests into Cloud Build

igooch:
**What type of PR is this?**

/kind feature

Working on #3795


test/upgrade/main.go
```diff
@@ -391,39 +424,135 @@ func createGameServers(ctx context.Context, gsPath string) {
...
+
+	eventOptions := informers.WithTweakListOptions(func(opts *metav1.ListOptions) {
+		opts.FieldSelector = fieldSelector
+	})
+	kubeInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient, 5*time.Second,
+		informers.WithNamespace("default"), eventOptions)
+	eventInformer := kubeInformerFactory.Core().V1().Events().Informer()
+
+	_, err = eventInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(_, newObj interface{}) {
```
gongmax, line 542: Should it be AddFunc?

igooch: I think this would be `update` since we're watching the event stream of an existing (even if it's not yet fully functioning) resource. Should I add a watch for Add?

igooch: From adding in some logs to an `AddFunc` and the `UpdateFunc`, it looks like all the events are first captured by `AddFunc`, and then again multiple times by `UpdateFunc` each `defaultResync` which in this case is 5 seconds. So it does seem to make more sense to use `AddFunc`. On that same note, we can actually get rid of the `BackOff` code, as I had made an update to the Game Server template to change the restart policy to "never", so the game server pod will not go into a CrashBackOff loop as one pod can only live once. The container crashing will be caught by the watch for the `Unhealthy` game server.

gongmax (approved):

PR #4045, opened 2024-11-21
Correct CI check for examples and add a unit test

### LINKED ISSUE #4042: CI check `check-example-versions` doesn't work as expected.
We added a CI check to cover the scenario that "if any files change within an example - outside of a simple denylist - and the version in its Makefile is not updated, it should fail the CI pipeline". The PR added it is #3940 . The denylist contains all yaml file so any change in these files should not be considered in this CI check. But in a recent PR #4040 , the CI failed even the only change was the yaml files. Need to debug this CI step, or revert it to unblock further PRs.

wheatear-dev:
This fixes a bug within the CI check for unversioned changes to examples, involving an exclude pattern not matching relevant file paths that contain separators.

**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

My previous PR #3940 - which was merged and is now live - is causing CI pipelines to fail when any changes are made to YAML files in an example, even though these were supposed to have been excluded by a deny list.

This PR provides:

1. A unit test that reproduces the bug (on previous code)
2. A small code fix, that operates the `filepath.Match()` call on the filename (from `filepath.Split()`) and not the full path.

Closes #4042 

I am very sorry about this, and hope this hasn't left a bad impression.


gongmax:
Thank you for the quick fix! I didn't notice the case when reviewing the code so it's on me too :)

gongmax (approved):

PR #4048, opened 2024-11-25
Clarify docs on GKE Autopilot and node pools

danfairs:

/kind documentation

This change makes it clearer that the cluster administrator does not need to manually configure
node pools when using GKE Autopilot.


igooch (approved):
LGTM

PR #4050, opened 2024-11-26
Set timeout duration as optional argument

AndreasWintherMoen:
**What type of PR is this?**

/kind cleanup

Just a small change to add an optional _timeoutSeconds_ argument instead of a hard-coded value in Connect method in Unity SDK. This allows users to set a custom timeout duration if they have a very specific setup that requires it. The default timeout is 30 seconds to ensure backward compatibility.


AndreasWintherMoen:
Fixed CLA. Not sure what other failing check is caused by.

gongmax:
Hi @AndreasWintherMoen , thanks for contribution! Would you mind add some description to the doc [here](https://github.com/googleforgames/agones/blob/edde3c39eb5c57dad19a8fd51fd608996e8417b5/site/content/en/docs/Guides/Client%20SDKs/unity.md?plain=1#L84-L89) for the change? Thanks!

igooch (approved):
LGTM

site/content/en/docs/Guides/Client SDKs/unity.md
```diff
@@ -82,8 +82,8 @@ var agones = agonesGameObject.GetComponent<Agones.AgonesSdk>();
 ```
 
 To connect to the SDK server, either local or when running on Agones, run the async `Connect()` method.
-This will wait for up to 30 seconds if the SDK server has not yet started and the connection cannot be made,
-and will return `false` if there was an issue connecting.
+By default, this will wait for up to 30 seconds if the SDK server has not yet started and the connection cannot be made. You can specify an optional timeout duration argument to the method, e.g., Connect(60) for a 60-second timeout.
```
igooch, line 85: Should add feature expiry around the deleted lines, and publish version around the new lines.

```
{{% feature expiryVersion="1.46.0" %}}
{{% /feature %}}
```

```
publish='{{% feature publishVersion="1.46.0" %}}'
{{% /feature %}}
```

igooch (changes requested):

markmandel:
Super late to the party on this PR, but I'll ask the question: _why_? 😄 we've used 30s everywhere else, seems strange to deviate in one place without a pretty strong reason.

gongmax:
Closing now since no active engagement for a while.

PR #4051, opened 2024-11-26
Update SIDECAR_IMAGE tag in Helm chart variable 

### LINKED ISSUE #3979: Add Helm Chart Variable for Configuring SIDECAR_IMAGE Independently
Currently, the SIDECAR_IMAGE environment variable can be set in the Agones controller, but it cannot be easily modified when installing Agones via Helm. This limitation forces the image registry to follow the common Agones settings, and it cannot be customized through deployment commands, arguments, or flags. For instance, if there is a need to use a different registry for the sidecar image, such as in cases where the sidecar image must be pulled from a private registry due to issues like those described in [issue #3978](https://github.com/googleforgames/agones/issues/3978), the current setup does not provide an easy way to achieve this.

I would like to see the introduction of a Helm chart variable that allows the SIDECAR_IMAGE to be configured independently. This would enable users to specify a different image registry for the sidecar without affecting the main Agones configuration. It would provide more flexibility in managing image sources, especially in scenarios where different registries are required for different components.

The current alternative is to manually modify the deployment after the Helm installation or to maintain a custom fork of the Helm chart that includes the desired changes. However, these approaches are less efficient and more error-prone compared to having an official Helm variable for this purpose.

**Additional context**
This feature would be particularly useful in situations where the sidecar image needs to be pulled from a private registry, as in the case described in issue #3978. It would allow for easier maintenance and customization of Agones installations in environments with specific registry requirements.

0xaravindh:
**What type of PR is this?**

 /kind cleanup

Closes #3979


install/helm/agones/values.yaml
```diff
@@ -280,6 +280,7 @@ agones:
       pullPolicy: IfNotPresent
     sdk:
       name: agones-sdk
+      agonesSdkImage: us-docker.pkg.dev/agones-images/release/agones-sdk:1.46.0-dev
```
igooch, line 283: This image name does not exist. For example, when I create a game server from a development build the game server pod has an image name: `image: us-docker.pkg.dev/my-project/my-repo/agones-sdk:1.46.0-dev-cd9b7a1` which is derived from `sidecar_tag = $(REGISTRY)/agones-sdk:$(VERSION)` in ~/agones/build/Makefile.

igooch (commented):
I'm not sure this solves the original issue, which seems to be that they want to have different sidecar images for different game server pods. This PR doesn't change that the controller still applies only one SDK image to all game server pods. If we implement the ability for a controller to add different sidecar versions based on something (perhaps label or metadata of a fleet?) this will be a much more involved change.

gongmax:
Closing this as discussed.

PR #4054, opened 2024-12-02
Delete List Value(s) on Game Server Allocation

### LINKED ISSUE #4003: Enhance AllocationSpec List modification to include delete operation
Any plans to include the ability to delete a value from a list at allocation time? It's the only list manipulation missing from the allocationspec (but is in the SDK) and the Agones integration I'm working on would benefit from it. 

Concretely: my PoC backfill matchmaking implementation sets a backfill request identifier Label at a pre-defined key on the game server so incoming matches containing backfill players can find a server which wants those kinds of players by using a LabelSelector. This works nicely since the AllocationSpec lets me remove the backfill request identifier using a metadata patch operation at allocation time, so as servers get their backfills they are atomically removed from the list of servers requesting backfills. This approach is fine as-is, but is limited to only one concurrent backfill request per server, since I have to use that pre-defined Label key in order to make the LabelSelector.

If I could move to storing this ID in a List instead, then I could associate multiple backfill requests with the server (in order to for example, request backfills for tank players, healers, DPS concurrently), having the allocationSpec select on the presence of the ID in the List. Would be a nice-to-have feature, but to be feasible, this would require the ability to atomically remove the backfill ID from the list via the AllocationSpec, like I can with the Label value in my current implementation.

The current working PoC implementation is the forerunner for alternatives.

**Additional context**
This implementation is for example content slated to ship with the next version of Open Match.

igooch:
**What type of PR is this?**

/kind feature

Adds the ability to delete value(s) from a List during game server allocation.

Closes #4003

Commits have been broken out into the different parts of this PR. Follows the existing code for AddValues during game server allocation.


pkg/apis/agones/v1/gameserver.go
```diff
@@ -1026,6 +1026,37 @@ func (gs *GameServer) AppendListValues(name string, values []string) error {
 	return errors.Errorf("unable to AppendListValues: Name %s, Values %s. List not found in GameServer %s", name, values, gs.ObjectMeta.GetName())
 }
 
+// DeleteListValues removes values from the ListStatus Values list. Values in the DeleteListValues
+// list that are not in the ListStatus Values list are ignored.
+func (gs *GameServer) DeleteListValues(name string, values []string) error {
```
gongmax, line 1031: Can we add unit test for this function?

proto/allocation/allocation.proto
```diff
@@ -239,7 +239,9 @@ message CounterAction {
 // ListAction is an optional action that can be performed on a List at allocation.
 // AddValues: Append values to a List's Values array (optional). Any duplicate values will be ignored.
 // Capacity: Update the maximum capacity of the Counter to this number (optional). Min 0, Max 1000.
+// DeleteValues: Remove values from a List's Values array (optional). Any nonexistant values will be ignored.
 message ListAction {
 	repeated string addValues = 1;
 	google.protobuf.Int64Value capacity = 2;
+  repeated string deleteValues = 3;
```
gongmax, line 246: please fix the indention

gongmax (commented):
LGTM, some minor comments

gongmax (approved):

PR #4075, opened 2024-12-27
Unity SDK: Delay dispose of CancellationTokenSource when quitting

kaboing:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Quitting in Unity fires a number of events and callbacks. 

`MonoBehaviour.OnApplicationQuit` is called before `Application.wantsToQuit` which leads to undesired behaviour. `Application.wantsToQuit` can be used to interrupt a quit process, so you can shut down gracefully. If this is utilized, the AgonesSDK has already disposed the cancellation token and future calls will fail. 

We use this when controlling disruption https://agones.dev/site/docs/advanced/controlling-disruption/

Reference: https://docs.unity3d.com/6000.0/Documentation/ScriptReference/Application-wantsToQuit.html

Same solution as in UniRX: https://github.com/neuecc/UniRx/pull/463

Closes #

### DIFF
--- a/sdks/unity/AgonesSdk.cs
+++ b/sdks/unity/AgonesSdk.cs
@@ -71,6 +71,7 @@ namespace Agones
         {
             String port = Environment.GetEnvironmentVariable("AGONES_SDK_HTTP_PORT");
             sidecarAddress = "http://localhost:" + (port ?? "9358");
+            Application.quitting += OnApplicationQuitting;
         }
 
         private void Start()
@@ -79,9 +80,14 @@ namespace Agones
             HealthCheckAsync();
         }
 
-        private void OnApplicationQuit()
+        private void OnApplicationQuitting()
         {
             cancellationTokenSource.Dispose();
+        }
+
+        private void OnDestroy()
+        {
+	        Application.quitting -= OnApplicationQuitting;
         }
         #endregion
 


markmandel:
Calling our most recent Unity experts for review please: @ZeroParticle , @aallbrig, @kingshijie, @MaxHayman, @mollstam - does this change look good to you? 👋🏻

mollstam:
Hello all, a bit surprised by this ping (haven't used Agones since '21) but looking at the code I do have some concerns.

I think this change per se is OK, as it does indeed address undesired behaviour when application quit is aborted / handled gracefully, but I think the code would benefit from us zooming out a notch.

I would only ever create a `CancellationTokenSource` (CTS) if I would pass in its `Token` into an async method. To me its very weird to only use the CTS for checking its cancellation status and if so throwing, I don't believe its idiomatic C# and you might as well have a bool field you flip (in this case in the new `OnApplicationQuitting`) and when checking if flipped just throw or noop. Using a CTS provides no benefit here.

This leads me to another very important point: `CancellationTokenSource` does not cancel when disposed. Here's from [the .NET docs for `CancellationTokenSource.Dispose`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.dispose?view=net-9.0):

> Note that calling Dispose does not communicate a request for cancellation to consumers of the associated [Token](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.token?view=net-9.0). You can communicate a request for cancellation by calling methods such as [Cancel](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.cancel?view=net-9.0) or [CancelAfter](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.cancelafter?view=net-9.0).

So `AgonesSdk.cancellationTokenSource` is never cancelled (it's only used in this file no?) which makes that check in `SendRequestAsync` even more mysterious. I think the only effect it has, and maybe the original error that @kaboing was seeing ("causing future calls to fail") is trying to use a disposed CTS, which will make the `.Token` access throw an `System.ObjectDisposedException`.

If that is true I don't think the changes proposed in this PR are necessarily the fix called for. If it is as that comment in `SendRequestAsync` suggests: "To prevent that an async method leaks after destroying this gameObject" I think we're better of simply checking `if (this == null`) (which internally uses Unity's lifetime check logic) instead and getting rid of the CTS altogether.

Sorry for a long reponse, let me know what your thoughts are. Gotta run.

ZeroParticle:
@mollstam's comments seem reasonable to me. I do have some concern that the async request handler setup we have in that file does not have any cancellation implementation. It seems reasonable to attempt to cancel any inflight requests when the game object is destroyed.

kaboing:
@mollstam Makes sense, good observation. 
You were proposing something like this for `SendRequestAsync`?

```
        public async Task<AsyncResult> SendRequestAsync(string api, string json,
            string method = UnityWebRequest.kHttpVerbPOST)
        {
            // To prevent that an async method leaks after destroying this gameObject.
            if (this == null)
            {
                return new AsyncResult { ok = false };
            }
```

kaboing:
The HealthCheck shows signs of the ObjectDisposedException

https://github.com/googleforgames/agones/blob/main/sdks/unity/AgonesSdk.cs#L269-L284

markmandel:
This hasn't been updated since March - shall we close this?

markmandel:
It's been a week since I wrote the comment asking to close - and no response -- so making an executive decision to close.

Please feel free to reopen if you would like to take this up again.

PR #4090, opened 2025-01-20
fix: bump version of jsonpatch for lossy max int64

### LINKED ISSUE #3636: JSON Patch in GameServer Controller is Lossy on Int64
**TL;DR**:
JSON does not support Int64 well. The current recommendation is that anyone using the `count` or `capacity` values should avoid using extremely large numbers.

**Issue with setting count or capacity for Counter to max(int64)**: 

When the Counter `capacity` or `count` are set to max(int64) in the game server spec and running `kubectl create -f fleet.yaml` we get an error: 
`spec.template.spec.counters.sessions.capacity: Invalid value: "number": spec.template.spec.counters.sessions.capacity in body must be of type integer: "number"`
 
When the Counter `capacity` or `count` in the game server spec are close to, but not at, max(int64) `kubectl create` does not error. However, when checking the details of the game server with `kubectl describe` the values are truncated:
```
spec:
  template:
    spec:
      counters:
        sessions:
          count: 8611686018427387904
          capacity: 8611686018427387904
```
becomes
```
Spec:
  Template:
    Spec:
      Counters:
        Sessions:
          Capacity:  8611686018427388000
          Count:     8611686018427388000
```

**Debugging**:

@zmerlynn root-caused the issue #3608, and found that the source issue is with `jsonpatch` translating values to scientific notation. A different repo ran into the same error and has a longer description of the issue https://github.com/evanphx/json-patch/issues/189. We're unable to switch to this repo because the fixed method `CreateMergePatch` just gives back bytes, but in tests, etc. we use the jsonpatch structures. This will likely need to be an upstream fix in [gomodules/jsonpatch](https://github.com/gomodules/jsonpatch) or [kubernetes-sigs/json](https://github.com/kubernetes-sigs/json). The takeaway is that JSON does not support int64 well, and if the controller is having this many issues there are likely other issues that would crop up elsewhere in the code, and other languages would have yet another set of issues.

lacroixthomas:
**What type of PR is this?**

/kind bug

This PR bump the version of the jsonpath module, which includes a fix for the lossy int64 max values

Currently, the jsonpath transform the int64 (big values, max int64 for example) as a float64 with scientific notation, when cast back to a int64, it losses precision / the int64 is not the same.

Closes #3636 

A PR has been opened on the jsonpath repo: https://github.com/gomodules/jsonpatch/pull/40

The repo doesn't seems to be maintained, I open this draft PR to do some test by using by fork

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -124,3 +124,5 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace gomodules.xyz/jsonpatch/v2 => github.com/lacroixthomas/jsonpatch/v2 v2.0.0-20250120171921-4c723fc27712
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	golang.org/x/oauth2 v0.26.0
 	golang.org/x/time v0.9.0
 	golang.org/x/tools v0.29.0
-	gomodules.xyz/jsonpatch/v2 v2.4.0
+	gomodules.xyz/jsonpatch/v2 v2.5.0
 	google.golang.org/api v0.220.0
 	google.golang.org/genproto/googleapis/api v0.0.0-20250207221924-e9438ea467c6
 	google.golang.org/grpc v1.70.0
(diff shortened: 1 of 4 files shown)


lacroixthomas:
Did some manual, testing, it seems to be working well
Awaiting approval from the other repo: https://github.com/gomodules/jsonpatch/pull/40 
Will then update the go mod

markmandel:
https://github.com/gomodules/jsonpatch/releases/tag/v2.5.0 contains this fix, so now we can update this PR to point to that release.

markmandel (approved):
👍🏻

Issue #4093, opened 2025-01-21
JSON Schema Validation for Helm


igooch:
> [!NOTE]
> Milestone of https://github.com/googleforgames/agones/issues/3766

# Introduction
As part of Agones in-place upgrades we identified adding JSON schema validation as a step in validating an Agones upgrade https://github.com/googleforgames/agones/issues/3766. The goal of this validation is to make any breaking changes in the Helm chart immediately apparent to the user.

JSON schema validation will validate the Agones Helm chart .Values object (note that this is not the [values.yaml](https://github.com/googleforgames/agones/blob/release-1.46.0/install/helm/agones/values.yaml) file itself. Details are in the Helm [documentation](https://helm.sh/docs/topics/charts/#schema-files) on JSON schema files). Validation only applies to the `helm install` or `helm upgrade`, and not a direct `kubectl install` of the `values.yaml` file.

# Critical User Journey

1. An Agones release has a breaking change. For example `foo.bar` moves to `foo.controller.bar`.
2. User does not read the documentation about the breaking change, and applies `helm upgrade --install agones --version=$NewRelease` with their current chart configuration of `--set foo.bar=baz`.
3. Before the new release is installed by Helm, the validation of the Helm `.Values` object fails with error similar to `Error: validation failed foo.controller.bar is required`.
   - Note that [Helm](https://github.com/helm/helm/blob/a933bcc7d13733a2bef60a184c13fe3ed6703fb5/pkg/chartutil/jsonschema.go#L25) currently uses https://github.com/xeipuuv/gojsonschema as its JSON validation, so all error messages come from this repo, and are not modified by Helm or Agones.
5. User makes changes suggested by the Helm error, and successfully re-applies  helm upgrade with `--set foo.controller.bar=baz`.

# Design Overview
## Goals

- Validation only applies to Agones Helm installations or upgrades.
- Validate that all values required for an Agones installation are present.
- Validate that all values are of the correct type.
- Validate that there are no unknown given values. (Attempting to set a value at a path that does not exist.)
- The validation only applies to a specific chart version. I.e. the JSON schema validation for chart 1.46 is only applied to 1.46 and not 1.45 or 1.47.

## Non-Goals

- Validating configurations. JSON schema does basic validation of Helm chart values (required field is present, value is of the correct type, etc.), but is not sophisticated enough to handle configuration versions.
- Non-Helm Agones installations or upgrades.

## Future Goals

[...]

PR #4094, opened 2025-01-21
JSON Schema Validation for Helm

igooch:
**What type of PR is this?**

/kind feature

This PR adds the [Helm schema validation file](https://helm.sh/docs/topics/charts/#schema-files) `values.schema.json`. 

Reasoning for the addition of the validation file is described more fully in #4093: 

>  The goal of this validation is to make any breaking changes in the Helm chart immediately apparent to the user.
> 
> JSON schema validation will validate the Agones Helm chart .Values object (note that this is not the [values.yaml](https://github.com/googleforgames/agones/blob/release-1.46.0/install/helm/agones/values.yaml) file itself. Details are in the Helm [documentation](https://helm.sh/docs/topics/charts/#schema-files) on JSON schema files). Validation only applies to the helm install or helm upgrade, and not a direct kubectl install of the values.yaml file.

Working on #4093

Details on the creation of the `values.schema.json` file:
- Helm currently uses https://json-schema.org/draft-07/schema as the schema for the validation file.
- The schema could not be automatically inferred (as described in https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) from the values.yaml file nor the `helm get values agones -n agones-system -a -ojson` because this infers that all fields are required, when many of the fields are not in fact required, and can be safely set to nil.
- Note in the [Helm templating language](https://helm.sh/docs/chart_template_guide/control_structures/#ifelse) (which is based off of Go templating, but not exactly the same):
   > A pipeline is evaluated as false if the value is:
   > a boolean false
   > a numeric zero
   > an empty string
   > a nil (empty or null)
   > an empty collection (map, slice, tuple, dict, array)
   - This means that Helm template conditionals such as `{{- if .Values.agones.controller.generateTLS }}` will evaluate to `false` if `.Values.agones.controller.generateTLS` is `nil`, so `.Values.agones.controller.generateTLS` is not a required value.
- Helm template lines such as `{{- $replicas := (int .Values.agones.controller.replicas) }}` mean that `agones`, `agones.controller`, and `agones.controller.replicas` are all required values as these will throw errors if `nil`.
- Whether or not a field was required was determined by the above logic for all template fields in [install/helm/agones/templates](https://github.com/googleforgames/agones/tree/release-1.46.0/install/helm/agones/templates). 
- Deprecated fields are not included in this PR. We do not currently have the ability to warn on deprecated fields (noted in #4093). This should be included in a future PR once Helm has updated to a newer version of json schema.
- In the event this validation schema misses a valid edge case, the user can still attempt a `helm install` or `helm upgrade` with the flag `--skip-schema-validation`.

CI/CD testing and documentation will be added in future PRs. This PR can be manually tested with:

```
me@me:~/agones/install/helm/agones$ helm lint .
==> Linting .

1 chart(s) linted, 0 chart(s) failed
```

### DIFF
--- a/install/helm/agones/values.schema.json
+++ b/install/helm/agones/values.schema.json
@@ -1,0 +1,1682 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "agones": {
+      "type": "object",
+      "properties": {
+        "featureGates": {
+          "type": "string"
+        },
+        "rbacEnabled": {
+          "type": "boolean"
+        },
+        "registerApiService": {
+          "type": "boolean"
+        },
+        "registerServiceAccounts": {
+          "type": "boolean"
+        },
+        "registerWebhooks": {
+          "type": "boolean"
+        },
+        "cloudProduct": {
+          "type": "string"
+        },
+        "createPriorityClass": {
+          "type": "boolean"
+        },
+        "priorityClassName": {
+          "type": "string"
+        },
+        "requireDedicatedNodes": {
+          "type": "boolean"
+        },
+        "crds": {
+          "type": "object",
+          "properties": {
+            "install": {
+              "type": "boolean"
+            },
+            "cleanupOnDelete": {
+              "type": "boolean"
+            },
+            "cleanupJobTTL": {
+              "type": "integer",
+              "minimum": 0
+            }
+          },
+          "if": {
+            "properties": {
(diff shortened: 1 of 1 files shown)


peterzhongyi:
Tested locally, and verified it works. Please put the last 2 bullets points into the Agones documentation as well in a follow up PR, too.

peterzhongyi (approved):

PR #4098, opened 2025-01-24
fix: ensure fleet autoscaler policy are namespaced

### LINKED ISSUE #3954: FleetAutoscaler with Counter policy trying to scale Fleet to negative replicas count
**What happened**:
I had an issue with the fleet auto scaler trying to scale to -41 replicas. 
As I see, fleet auto scaler with Counter policy does not respect the game server's namespace when calculating the current game server count.
I have different namespaces like qa1, qa2, etc., and each has a fleet **my-fleet**, where 1-2 game servers are launched.
I did debug and saw that [ListGameServersByFleetOwner](https://github.com/googleforgames/agones/blob/049de2fbbe8d548e68a79e548dd0d05bb37d8636/pkg/fleetautoscalers/fleetautoscalers.go#L369C24-L369C51) returned all game servers in a cluster with fleet owner name **my-fleet**.
I'm expecting that fleet and game servers are namespaced resources.
I did the simple fix - we just added namespace name to fleet name, like **my-fleet-qa1**, **my-fleet-qa2**

Fleet is the namespaced resource. I need to be able to create a fleet with the same name in different namespaces with CountsAndLists policy.

1. Create 2 fleets with the same name in different namespaces.
2. Create 2 fleet autoscaler with Counter policy
3. Just autoscale them.

-

**Environment**:
- Agones version: 1.42
- Kubernetes version (use `kubectl version`): 1.29
- Others:

lacroixthomas:
**What type of PR is this?**
/kind bug

Ensure the fleet autoscaler policy are namespaced, to ensure it's not autoscaling on all namespaces (on fleet with same name but from different namespaces)

Closes #3954 

Did some manual tests following the steps described on the issue. 

TODO:
- [x] Manual testing
- [x] Make sure the `syncFleet()` and the `updateFleet()` are also namespaced
- [x] Fixed current unit tests


pkg/fleets/fleets.go
```diff
@@ -44,10 +44,10 @@ func ListGameServerSetsByFleetOwner(gameServerSetLister listerv1.GameServerSetLi
 
 // ListGameServersByFleetOwner lists all GameServers that belong to a fleet through the
 // GameServer -> GameServerSet -> Fleet owner chain
-func ListGameServersByFleetOwner(gameServerLister listerv1.GameServerLister,
+func ListGameServersByFleetOwner(gameServerNamespacedLister listerv1.GameServerNamespaceLister,
```
lacroixthomas, line 47: We most probably want to do the same for the `ListGameServerSetsByFleetOwner()`, it seems to be expected to work by namespace, but it's not the case (for the `syncFleet()` and `updateFleet()`

pkg/fleetautoscalers/controller.go
```diff
@@ -313,7 +313,8 @@ func (c *Controller) syncFleetAutoscaler(ctx context.Context, key string) error
 	}
 
 	currentReplicas := fleet.Status.Replicas
-	desiredReplicas, scalingLimited, err := computeDesiredFleetSize(fas.Spec.Policy, fleet, c.gameServerLister, c.counter.Counts())
+	gameServerNamespacedLister := c.gameServerLister.GameServers(fleet.Namespace)
```
lacroixthomas, line 316: Set the lister namespaced at this level as the whole controller is based per namespaces

igooch:
Would you be able to add a test that tests for the issue in the original issue? I.e. a test that ensures that two fleets with the same name in different namespaces scale up independently?

lacroixthomas:
> Would you be able to add a test that tests for the issue in the original issue? I.e. a test that ensures that two fleets with the same name in different namespaces scale up independently?

Yes, I'm gonna do that, I would imagine an e2e test ? 

Currently, it seems that the namespace is set by the `e2eframework`, it only expect a single namespace https://github.com/googleforgames/agones/blob/main/test/e2e/framework/framework.go#L168, I can have a look on how to managed multiple ones at the same time

lacroixthomas:
Actually, this could be something similar to the allocator one: https://github.com/googleforgames/agones/blob/main/test/e2e/allocator_test.go#L564C6-L564C33 

Will add the e2e test this week

lacroixthomas:
My e2e test is not valid, it's also passing on the `main` branch, will come back to it this week

markmandel (approved):
LGTM. Nice change.

PR #4101, opened 2025-02-05
Adds helm schema validation test to the test suite

### LINKED ISSUE #3766: RFC: In-place Agones Upgrades
> [!NOTE]
> Request for comments - please provide feedback by 2024/05/01!

# Overview

In this issue, I provide a roadmap to implement in-place upgrades in Agones. An in-place upgrade is the ability to use `kubectl apply` of a new configuration, on an Agones cluster with active Agones objects, crucially GameServer and Fleet objects, while supporting active allocations. [We don’t support in-place upgrades today](https://1-39-0.agones.dev/site/docs/installation/upgrading/), as noted in #1742, #2843, and probably many other issues.

We defined “upgrade” here more liberally than just version upgrades: Agones supports [Feature Gates](https://1-39-0.agones.dev/site/docs/guides/feature-stages/#feature-gates), and changing a feature gate can be analogous to a version upgrade, as changes to the gate may enable (or disable) new parts of the API. Additionally, there are other parts of the configuration, such as the `sdk-server` resources, that are relevant to rendering `GameServer` objects - as such, updating configuration relevant to how `GameServer` is rendered also needs to be considered.

# Requirements / Constraints
I propose the following requirements and constraints:

## UI/UX <a href="#ui-ux" id="ui-ux">#</a>

1. The system must function normally during an upgrade, from a user's perspective
   - Should not delete / interrupt existing `GameServers` or `Fleets`. 
   - Allocated `GameServers` must not be disrupted by an Agones upgrade
      - FleetAutoScaler buffers should be maintained, etc.
   - GameServer allocations must continue during an Agones upgrade

1. Upgrading via `kubectl apply` must be supported.
   - **Why?** Our end-to-end tests focus exclusively on `helm` installs, but our users often use a `kubectl apply` based lifecycle, using `helm` as the renderer (`skaffold` is a common flow for this as well).
   - Pruning using `kubectl apply` is not well supported, and we will need to monitor this carefully if we make deployment changes. We haven’t actually needed to prune between versions yet, but we should consider embracing something like [ApplySet](https://kubernetes.io/blog/2023/05/09/introducing-kubectl-applyset-pruning/). However, this can be a later follow-on and may be easier to implement in an operator.
   - Using `helm` will still be preferable, as it allows for e.g. rollback behavior if the new version fails to install. Here we are simply saying that `apply` is a requirement.

1. Helm is still the renderer of choice - in particular, modifying and rendering manifests should be done using helm, [e.g. our docs.](https://agones.dev/site/docs/installation/install-agones/yaml/#customizing-your-install)
   - In particular, just changing the version of the image in the manifest is not a recommended upgrade path - it may work _sometimes_, but the install should be re-rendered between versions. (Upgrading by applying our [released install.yaml](https://1-39-0.agones.dev/site/docs/installation/install-agones/yaml/) is supported - this is rendered every release.).
   - Relatedly, the arguments to our deployed binaries (e.g. `agones-controller`) are not a stable interface and may change between versions.
   - Why? We need to be able to make deployment changes - we did this with the [HA Agones](https://github.com/googleforgames/agones/issues/2797) work, for instance.

1. We support only two configurations of Agones running at the same time, i.e. `1.{N}` -> `1.{N+1}` or on rollback, `1.{N}` -> `1.{N-1}` (this express a version change, but feature gate or configuration changes are similar). Deploying a third configuration while there are still mixed objects of the other two is not supported. *This includes the sdk-server*, meaning it’s not supported to do rapid-fire upgrades without waiting for Fleets to be patched to the new version and all allocated GameServer sessions to complete.
   - Example:
     - 1.40 is installed
     - An upgrade to 1.41 fails, leaving the cluster in indeterminate state
     - It is not supported to attempt to upgrade to 1.42 without first rolling back 1.41.
   - We will have optional automation to reconcile all Fleets to the new configuration, plus documentation for how to do it manually.
   - **Why?** It’s hard enough to reason about two versions. Mixing three versions is [mass hysteria](https://www.youtube.com/watch?v=-9NMt42il4Q).

## Version Upgrades <a href="#version-upgrades" id="version-upgrades">#</a>

1. For each release after we support upgrades, we will publish a supported upgrade horizon for that version. We may move the upgrade horizon forward as necessary, but we will always support an upgrade horizon of at least one version. (Initially, we may only support a build horizon of one version, but at the very least we should get in the habit of publishing it.)
   - Larger horizons are more user friendly, but if upgrades are truly non-disruptive, it may encourage users to upgrade more frequently. We’ll need to find a balance between “complexity to maintainers” and “simplicity to users”.
   - Example: 
     - When 1.50 is released, we publish an upgrade horizon of 1.45, meaning that any version in [1.45, 1.50) can be upgraded to 1.50. 
     - However, to support an intricate migration, we may say that 1.51 has an upgrade horizon of 1.50.
     - In this example, to upgrade from 1.45 to 1.51 would require an upgrade from 1.45 -> 1.50, then 1.50 -> 1.51.
   - **Why?** Supporting an indefinitely long upgrade horizon often results in intractable complexity. Separately, forcing “landing pad” releases (such as 1.50 in the above example) solves some migration issues, e.g. you can write API-storage migration code in one release and then rip out the migration in a future release after moving the horizon forward.

## Feature Gates <a href="#feature-gates" id="feature-gates">#</a>

1. We must support enabling [Feature Gates](https://1-39-0.agones.dev/site/docs/guides/feature-stages/#feature-gates) during configuration changes.
   - Feature gate enablement may/will occur during a version upgrade (e.g. a feature is promoted to Beta in the version being upgraded to), or because the feature gate is explicitly enabled)
   - **Why?** This is table stakes for the feature, as we commonly promote feature gates between versions.
     - I considered requiring a flow where we upgrade versions first then upgrade feature gates, but unless/until we have an operator, a phased upgrade is complicated for users.

1. Upgrading a configuration running Alpha Feature Gates may result in lost configuration and/or disable the feature.
   - **Why?** “Support for this feature may be dropped at any time without notice. The API may change in incompatible ways in a later software release without notice.” [[source](https://1-39-0.agones.dev/site/docs/guides/feature-stages/#alpha)]. We will do our best to keep Alpha APIs consistent, but the safest way to change Alpha APIs is to abandon the original name/structure, resulting in a loss of the configuration. **We will not write CRD migration code for Alpha features.**
   - Breaking changes in Alpha features will be called out in upgrade documentation, alongside the build horizon, allowing operators to plan their upgrades.

1. Disabling a Beta Feature Gate may result in lost configuration and/or disable the feature.
   - **Why?** Similar to above, if a Beta gate is disabled, the feature capabilities and configuration may be lost entirely.
   - **We will write CRD migration code for Beta features if APIs change.** This is a change from the existing documentation that [says we won’t.](https://1-39-0.agones.dev/site/docs/guides/feature-stages/#beta) I am proposing that we raise the bar for Beta.

1. Newly enabled API features **may be unreliable** until an upgrade is fully rolled out to the Agones Control Plane. Some features may also rely on a full `sdk-server` rollout as well (Counts & Lists is a great example).
   - Newly enabled API features may “flap” as old-version controllers remove the unknown fields.
   - **Why?** A common problem with API-related features is that one replica knows about the API and another does not. The general user guidance is to avoid using new APIs until the upgrade is finished.
     - This constraint lets us be slightly lazy - otherwise we would need a consensus system before we used a new feature gate.
     - Note that this same problem exists in kube-apiserver, and e.g. GKE has ~30m regional cluster upgrades - it has not been a big issue.

1. Logic for newly enabled features must assume other controller replicas may not enforce the same logic until the upgrade is fully rolled out.
   - **Why?** Similar to the API case above, within a single replica you can assume consistency, but until an upgrade is complete, other replicas may run with different configurations. Even in the case of agones-controller features where the controller is leader-elected, you have to assume that the leader may be lost (eviction, crash, etc.)

## SDK proto / sdk-server <a href="#sdk-proto" id="sdk-proto">#</a>

This section is specifically about the [SDK proto definitions](https://github.com/googleforgames/agones/blob/main/proto/sdk/sdk.proto), the interchange between the [Client SDKs](https://agones.dev/site/docs/guides/client-sdks/) and the `sdk-server`. We discuss the [Client SDKs below](#client-sdks).

1. In order to allow compatibility between game server binaries and sdk-server, a game server binary using Beta and Stable SDK protos must remain compatible with a newer sdk-server. **Our SDK compatibility contract is:** If your game server uses a non-deprecated Stable API, your binary will be compatible for 10 releases (~1y) starting from the SDK version packaged - e.g. if the game server uses non-deprecated APIs in the 1.40 SDK, it will be compatible through the 1.50 sdk-server. Stable APIs will almost certainly be compatible beyond 10 releases, but 10 releases is guaranteed. Similarly, using a non-deprecated Beta API, your binary will be compatible for 5 releases (~6mo).
   - We expect deprecations of Stable APIs to be exceedingly rare and thoroughly documented. The deprecation timer will start when the new APIs are available in Stable, with migration guides from the old APIs - SDK proto message names for the old/new functionality will remain distinct as well.
   - Beta deprecations will occur on graduation to Stable, which will involve introducing the APIs to Stable and marking the Beta APIs as deprecated. This deprecation follows the guarantee above, meaning the graduated APIs will remain in both Stable and Beta for 5 releases (6mo).
   - **Why?** SDK proto compatibility is crucial to reducing toil on our users. We don’t want to require users to redeploy binaries every release (6w!), and like to encourage upgrades as much as possible.

1. Alpha SDK APIs/RPCs are subject to change between releases - a game server binary using Alpha SDKs may not be compatible with a newer sdk-server. In Alpha, incompatible changes retaining the same SDK proto message name are allowed. When we make incompatible Alpha changes, we will document the APIs involved.
   - We will emphasize in the Alpha APIs that any use of Alpha APIs may prevent upgrades - at the very least, it will require you to pay attention to releases.
   - Since we aren’t guaranteeing proto compatibility between releases for Alpha SDK protos, there is no need to overlap Alpha/Beta on graduation. When Alpha APIs are graduated to Beta, the Alpha APIs will be deleted, with no overlapping release.
   - **Why?** Alpha means Alpha. We love testers - that’s great! But we need to be able to change Alpha APIs/RPCs.

1. We will document the maturity level and history of each Stable/Beta proto message, i.e.: `Foo (Stable in 1.40+; Beta in 1.38, deprecated 1.40, removed 1.45)`
   - This is probably best presented as a table.
   - **Why?** This allows users to easily understand if they are adopting an API that can’t be rolled back from. For example, with Foo above, the API is Stable in 1.40 - which means it’s not present in the Stable surface at all before 1.40, meaning they are pinned to 1.40+, though they could use the Beta surface to retain compatibility in [1.38, 1.44].

## Client SDKs <a href="#client-sdks" id="client-sdks">#</a>
This section is specifically about the [Client SDKs](https://agones.dev/site/docs/guides/client-sdks/), the per-language SDK. We discuss the [SDK proto surface above](#sdk-proto).

1. Alpha Client SDK APIs are subject to change between releases at any time. Beta Client SDK APIs should be considered generally stable, but we may alter the parameters/returns if we need to. Stable SDK APIs require a 5 release (6mo) deprecation cycle to remove, using a similar procedure to Stable SDK protos (make new API available, document migration, start timer, remove deprecated API).
   - Example: #3738 changes the returns of certain functions in the Go SDK. This change is fine in Alpha, should be rare in Beta but still allowed. However, if this API had made it to Stable, we would instead introduce new functions, deprecate the old, and remove them after 6mo.
   - **Why?** The burden of API stability is different for the Client SDK - for one, the persona doing the update is different in that Client SDK updates are usually handled by a developer, whereas most of the rest of this doc is related to changes made by a cluster administrator. We don’t want to burden developers too much or they will never update their SDK - but we also need to be able to modify it.

1. Users are not required to update Client SDK in game server binaries except for [SDK proto](#ske-proto) deprecations, or breaking Alpha API changes. We foresee that updating Client SDKs in game server binaries will be yearly toil, if using Stable APIs, or semiannual toil, if using any Beta API; this toil can be as simple as verifying that there were no deprecations in the period involved. Given how stable our core APIs have been, it may be possible to go multiple years without updating the Client SDK.

## Configuration / Helm Values <a href="#config" id="config">#</a>

1. The names of feature gates will be usable indefinitely in Agones binaries, as long as they match the new forever-value. This is a change from today, where [we reject](https://github.com/googleforgames/agones/blob/main/pkg/util/runtime/features.go#L167) gates that are not “active”. We will keep a map of retired gate names and whether they were retired true (every Stable gate) or false (retracted Alpha gates). If we see an old gate name with the right value, we will log a warning and continue - if the value does not match, we still reject it.
   - **Why?** This will allow manifests that include old gates to remain working at new versions, as long as the enablement value matches. (We will continue to fail brightly in the case that something is wrong, though.)

1. Helm [values.yaml](https://github.com/googleforgames/agones/blob/main/install/helm/agones/values.yaml) structure may be altered, but within the upgrade horizon, the old structure should still be usable and implicitly migrated.
   - **Why?** We need to be able to refactor values.yaml as needs change, so it can’t be a forever contract. However, we can give users some time to migrate.
   - TBD: It looks like Helm still has no capability to warn. Filed https://github.com/helm/helm/issues/12937 to ask about this - there’s a suggested alternate route in that issue as well.

# Milestones <a href="#milestones" id="milestones">#</a>

The project will be split up into the following milestones, each described in separate issues.

## Alpha <a href="#alpha" id="alpha">#</a>
At Alpha support level, applying an upgrade will upgrade the Agones Control Plane. However, Fleets will still use the old render configuration (`sdk-server` version, etc.) until updated manually; we will publish instructions on how to update Fleets to the new sdk-server.

1. [**SDK Compatibility**](https://github.com/googleforgames/agones/issues/3767): Ensure that the sdk-server sidecar remains compatible within the upgrade horizon for Beta and Stable APIs.

1. [**Storage Compatibility**](https://github.com/googleforgames/agones/issues/3771): Ensure that a controller at one configuration can read that were stored objects of a controller in a different configuration. Usually this will be the result of a feature gate change between configurations, i.e. `FeatureFoo` adds the `Foo` field to GameServer - if one configuration has the gate enabled and the other does not, we need to reason about object storage / compatibility. We will also write developer policies regarding API feature progression.

1. [**Testing**](https://github.com/googleforgames/agones/issues/3795): We need automated testing before we should consider this Alpha.

## Beta <a href="#beta" id="beta">#</a>

1. **Versioned Configuration**: (probably not needed - nice to have, but let's wait and see) For any configuration relevant to how the controller renders Pods, we will move the configuration to a ConfigMap that also keeps a generation number. This will allow us to tie rendered elements below Fleet to the specific configuration they were rendered under. We will then (optionally) roll out the new rendered configuration across all Fleet objects, or provide a simple workflow to do so.

1. **JSON schema for helm**: We should add a [JSON schema](https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) for our Helm chart. If we've made a breaking change in our Helm charts, this will make it more obvious to our users. (TBD: It could also potentially give us a way to enforce version horizons, since we could intentionally change schema elements in a way that there would be an error if you tried to skip too far at once.)

### LINKED ISSUE #4093: JSON Schema Validation for Helm

igooch:
**What type of PR is this?**

/kind feature

Final piece in the in-place upgrades moving to beta.

Adds a simple Helm schema validation test that runs as part of `make test` in CI / CD.

Note that this test only tests that the Helm `values.yaml` file passes the schema validation. It does _not_ test if there have been changes to the Helm template that require changes to the validation schema.

Closes #4093
Closes #3766


lacroixthomas:
Can't approve, but LGTM, also tested locally

peterzhongyi (approved):

PR #4115, opened 2025-02-19
Fix: Ensure Buildx Builders Are Created or Used for ARM64 and Windows

0xaravindh:
**What type of PR is this?**

/kind cleanup

This PR ensures that Docker Buildx builders are either used if existing or created if missing for ARM64 and Windows platforms, preventing build failures.

```
ERROR: existing instance for "arm64-builder" but no append mode, specify the node name to make changes for existing instances
```

```
ERROR: existing instance for "windows-builder-ltsc2019" but no append mode, specify the node name to make changes for existing instances
```

Closes #

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -514,9 +514,8 @@ build-required-src-dist:
 # According to the docker documentation, to build the images for ARM64 arch CPU
 # we would need to use buildx command which uses builder https://docs.docker.com/desktop/multi-arch/
 # We primarily need this for M1 Macs, see: https://medium.com/geekculture/docker-build-with-mac-m1-d668c802ab96
-# Making this command optional (using - as prefix) because it will fail once built for the first time, so we ignore failures.
 create-arm64-builder:
-	-docker buildx create --name $(BUILDX_ARM64_BUILDER)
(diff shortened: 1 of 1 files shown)


build/Makefile
```diff
@@ -570,8 +569,8 @@ build-agones-sdk-binary-windows: $(ensure-build-image)
 	$(GO_BUILD_WINDOWS_AMD64) \
 		-o $(go_build_base_path)/cmd/sdk-server/bin/sdk-server.windows.amd64.exe $(go_rebuild_flags) $(go_version_flags) $(agones_package)/cmd/sdk-server
 
-ensure-windows-buildx-%:
-	-docker buildx create --name=$(BUILDX_WINDOWS_BUILDER)-$*
+ensure-windows-buildx:
```
gongmax, line 572: why remove the flexibility of using `-%`? I haven't checked but will this makes any of the existing make command doesn't work?

0xaravindh: No, the previous change with -% wouldn't impact any other commands. We can use the same buildx instance for both Windows LTSC 2019 and LTSC 2022 without issues. we simplify the process and reduce some build time as well

gongmax (approved):

PR #4122, opened 2025-03-05
Fix JSON Schema validation for ServiceAccount annotations

### LINKED ISSUE #4119: JSON Schema does not match helm chart structure for serviceaccount.*.annotations
**What happened**:

Since JSON Schema validation was added in the 1.47.0 release it is impossible to install that release if one uses serviceaccount annotations, e.g. 

```
serviceaccount:
  sdk: 
    annotations: 
       default: 
          "annotation": value                      
```

Because in the schema annotations are specified like this
```
 "annotations": {
       "type": "object",
        "additionalProperties": {
            "type": "string"
          }
 }
```

Which causes this error 

```
error: values don't meet the specifications of the schema(s) in the following chart(s):
    agones:
    - agones.serviceaccount.sdk.annotations.default: Invalid type. Expected: string, given: object
```

The annotations should work as specified, helm chart actually expects objects there where each set of annotations is keyed by a namespace.

See above.

I assume it may be just a matter of adding a new "layer" to the spec like this

```
 "annotations": {
       "type": "object",
        "additionalProperties": {
            "type": "object",
            "additionalProperties": {
               "type": "string"
             }
          }
 }
```

**Environment**:
- Agones version:  1.47.0
```
Client Version: v1.32.2
Kustomize Version: v5.5.0
Server Version: v1.31.5-gke.1068000
```

0xaravindh:
**What type of PR is this?**

 /kind bug

- Updated `serviceaccount.*.annotations` schema to allow key-value string pairs.
- Changed `additionalProperties` from `type: string` to `type: object`.
- Resolves issue preventing Helm chart installation when annotations are provided.

Closes #4119


install/helm/agones/values.schema.json
```diff
@@ -213,7 +213,7 @@
                 "annotations": {
                   "type": "object",
                   "additionalProperties": {
-                    "type": "string"
+                    "type": "object"
```
igooch, line 216:
```suggestion
                    "type": "object"
                    "additionalProperties": {
                        "type": "string"
                     }
```

igooch (commented):
For this PR we only want to change the `agones.serviceaccount.sdk.annotations`. From our [Helm install docs](https://agones.dev/site/docs/installation/install-agones/helm/) `agones.serviceaccount.sdk.annotations` is `A map of namespaces to maps of [Annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) added to the Agones SDK service account for the specified namespaces`, whereas the rest of the annotations are a simple map of Annotations, and not a map of a map.

gongmax:
@0xaravindh can you verify the fix by following the reproduce step in https://github.com/googleforgames/agones/issues/4119?

install/helm/agones/values.schema.json
igooch: The service account annotations for `sdk` is a map of maps. The service account annotations for `controller` and `allocator` are maps as defined in https://agones.dev/site/docs/installation/install-agones/helm/.

So a sample section of a values.yaml file might look like:

```
  serviceaccount:
    allocator:
      name: agones-allocator
      annotations:
        "foo": bar
    controller:
      name: agones-controller
      annotations:
        "baz": qux
    sdk:
      name: agones-sdk
      annotations:
        default:
          "annotation": value
```

In Kubernetes Annotations (and labels) are required to be strings https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/. So here we want to verify that the user only passes in string values.

install/helm/agones/values.schema.json
```diff
@@ -173,13 +173,13 @@
...
```
igooch, line 176:
```suggestion
                    "type": "string"
```

install/helm/agones/values.schema.json
```diff
@@ -173,13 +173,13 @@
...
                   }
                 },
                 "labels": {
                   "type": "object",
                   "additionalProperties": {
-                    "type": "string"
+                    "type": "object"
```
igooch, line 182:
```suggestion
                    "type": "string"
```

install/helm/agones/values.schema.json
```diff
@@ -196,7 +196,7 @@
                 "annotations": {
                   "type": "object",
                   "additionalProperties": {
-                    "type": "string"
+                    "type": "object"
```
igooch, line 199:
```suggestion
                    "type": "string"
```

igooch (changes requested):

0xaravindh:
@igooch please take a look!

igooch (approved):
LGTM

Issue #3955, opened 2024-08-19
Add logging to policies that return errors within Chain policies


indexjoseph:
**What happened**:

The current implementation of the chain policy doesn't log any errors within the fleetautoscalers unless it's last chain entry, right now if an error is seen the chain fleetautoscaler will just skip over the chain entry and go to the next entry. 

**What you expected to happen**:

Log any unexpected errors within the chain regardless of index of the entry.

markmandel:
We should definitely fix this, moving to `awaiting-maintainer`

indexjoseph:
I can start working on this.

PR #4131, opened 2025-03-21
Log Chain ID in Events When Applying ChainPolicy in FleetAutoscaler

### LINKED ISSUE #3953: Add chain ID status to fleetautoscaler event logger

Update the fleetautoscaler controller to log events for the chainpolicy - Would like to be able to see the chain id attached to the event that finally gets applied on change in one of the columns, then at least we'd know which chain entry was eventually applied.

Maybe in the from section, something like: `fleet-controller:chain:1`

**Additional context**
Related to #3008

0xaravindh:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Added logging of Chain ID to the event message when a ChainPolicy is applied in the FleetAutoscaler controller.

Closes #3953


vicentefb:
@indexjoseph fyi

0xaravindh:
Below is an example of the relevant event output:

```
karavindh@karavindh:~$ k describe fleetautoscaler chain-fleet-autoscaler 
Name:         chain-fleet-autoscaler
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  autoscaling.agones.dev/v1
Kind:         FleetAutoscaler
Metadata:
  Creation Timestamp:  2025-03-21T18:27:33Z
  Generation:          1
  Resource Version:    1742581683798655009
  UID:                 668b9aa1-3905-40c1-989d-28c256ca7b16
Spec:
  Fleet Name:  fleet-example
  Policy:
    Chain:
      Id:  weekday
      Schedule:
        Active Period:
          Duration:    5h
          Start Cron:  0 1 * * 0
          Timezone:    America/New_York
        Between:
          End:    2100-02-23T21:04:04Z
          Start:  2100-02-20T21:04:04Z
        Policy:
          Buffer:
            Buffer Size:   50
            Max Replicas:  2000
            Min Replicas:  100
          Type:            Buffer
      Type:                Schedule
      Id:                  weekend
      Schedule:
        Active Period:
          Duration:    7h
          Start Cron:  0 1 * * 0
          Timezone:    America/New_York
        Between:
          End:    2100-02-26T21:04:05Z
          Start:  2100-02-24T21:04:05Z
        Policy:
          Counter:
            Buffer Size:   10
            Key:           rooms
            Max Capacity:  1000
            Min Capacity:  500
          Type:            Counter
      Type:                Schedule
      Buffer:
        Buffer Size:   5
        Max Replicas:  2000
        Min Replicas:  100
      Id:              default
      Type:            Buffer
    Type:              Chain
  Sync:
    Fixed Interval:
      Seconds:  30
    Type:       FixedInterval
Status:
  Able To Scale:     true
  Current Replicas:  100
  Desired Replicas:  100
  Last Scale Time:   2025-03-21T18:27:33Z
  Scaling Limited:   true
Events:
  Type     Reason              Age                  From                        Message
  ----     ------              ----                 ----                        -------
  Normal   AutoScalingFleet    10m                  fleetautoscaler-controller  Scaling fleet fleet-example from 2 to 100
  Warning  ScalingLimited      9m43s (x2 over 10m)  fleetautoscaler-controller  Scaling fleet fleet-example was limited to minimum size of 100
  Normal   ChainPolicyApplied  13s (x21 over 10m)   fleetautoscaler-controller  FleetAutoscaler 'chain-fleet-autoscaler' applied ChainPolicy | ID: default | Type: Buffer
karavindh@karavindh:~$ 
```

indexjoseph:
I'm also curious if this bug can be consolidated with #3955.

0xaravindh:
> #3955



> I think this all looks good except for one thing - it seems like the current functionality we're logging on every sync cycle. I'm not sure if it matters that much but I think we should only log when there's a change within the chain. For instance, we'd log when the `FleetAutoscaler` switches from applying the `default` policy to applying the `weekend` policy.

https://github.com/0xaravindh/agones/blob/09ca94005be43ae0159631177d0f51f086418530/pkg/fleetautoscalers/fleetautoscalers.go#L465-L470

You want to stop logging the default state. right?

indexjoseph:
> > #3955
> 
> > I think this all looks good except for one thing - it seems like the current functionality we're logging on every sync cycle. I'm not sure if it matters that much but I think we should only log when there's a change within the chain. For instance, we'd log when the `FleetAutoscaler` switches from applying the `default` policy to applying the `weekend` policy.
> 
> https://github.com/0xaravindh/agones/blob/09ca94005be43ae0159631177d0f51f086418530/pkg/fleetautoscalers/fleetautoscalers.go#L465-L470
> 
> You want to stop logging the default state. right?

Not necessarily, I'm pointing out that the current implementation logs the applied policy and fleet size on every synchronization cycle of the FleetAutoscaler. This results in repetitive event entries, even when there's no actual change in the applied policy or the fleet's size.

My suggestion is to modify the logging behavior so that it only occurs when the FleetAutoscaler switches from applying one policy to another (e.g., from 'default' to 'weekend'). Although I'm unsure if this is possible without adding logic to hold state, there's no reliable way to compare the previous policy within a chain against the current policy and determine whether a new log entry is needed, it's up to you if you think it's necessary.

0xaravindh:
> > > #3955
> > 
> > 
> > > I think this all looks good except for one thing - it seems like the current functionality we're logging on every sync cycle. I'm not sure if it matters that much but I think we should only log when there's a change within the chain. For instance, we'd log when the `FleetAutoscaler` switches from applying the `default` policy to applying the `weekend` policy.
> > 
> > 
> > https://github.com/0xaravindh/agones/blob/09ca94005be43ae0159631177d0f51f086418530/pkg/fleetautoscalers/fleetautoscalers.go#L465-L470
> > You want to stop logging the default state. right?
> 
> Not necessarily, I'm pointing out that the current implementation logs the applied policy and fleet size on every synchronization cycle of the FleetAutoscaler. This results in repetitive event entries, even when there's no actual change in the applied policy or the fleet's size.
> 
> My suggestion is to modify the logging behavior so that it only occurs when the FleetAutoscaler switches from applying one policy to another (e.g., from 'default' to 'weekend'). Although I'm unsure if this is possible without adding logic to hold state, as of right now there's no way to compare the previous policy within a chain against the current policy and determine whether a new log entry is needed, it's up to you if you think it's necessary.

@indexjoseph I added a `LastAppliedPolicy` field to track the previously applied policy. Now, logs will only show when the policy actually changes (e.g., from default to weekend). This should prevent the repeated event entries you pointed out. Let me know if that works!

```
karavindh@karavindh:~$ kubectl describe fleetautoscaler chain-fleet-autoscaler
Name:         chain-fleet-autoscaler
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  autoscaling.agones.dev/v1
Kind:         FleetAutoscaler
Metadata:
  Creation Timestamp:  2025-04-16T07:11:28Z
  Generation:          1
  Resource Version:    1744787518525023009
  UID:                 8a835d61-5578-4a84-85e1-861570bd61ab
Spec:
  Fleet Name:  fleet-example
  Policy:
    Chain:
      Id:  weekday
      Schedule:
        Active Period:
          Duration:    5h
          Start Cron:  0 0 * * *
          Timezone:    America/Los_Angeles
        Between:
          End:    2025-07-01T07:00:00Z
          Start:  2025-04-01T07:00:00Z
        Policy:
          Buffer:
            Buffer Size:   50
            Max Replicas:  120
            Min Replicas:  30
          Type:            Buffer
      Type:                Schedule
      Id:                  weekend
      Schedule:
        Active Period:
          Duration:    5h
          Start Cron:  0 0 * * *
          Timezone:    America/Los_Angeles
        Between:
          End:    2025-07-01T07:00:00Z
          Start:  2025-04-01T07:00:00Z
        Policy:
          Counter:
            Buffer Size:   10
            Key:           rooms
            Max Capacity:  120
            Min Capacity:  30
          Type:            Counter
      Type:                Schedule
      Buffer:
        Buffer Size:   5
        Max Replicas:  100
        Min Replicas:  20
      Id:              default
      Type:            Buffer
    Type:              Chain
  Sync:
    Fixed Interval:
      Seconds:  30
    Type:       FixedInterval
Status:
  Able To Scale:        true
  Current Replicas:     50
  Desired Replicas:     50
  Last Applied Policy:  Chain:weekday:Schedule
  Last Scale Time:      2025-04-16T07:11:28Z
  Scaling Limited:      false
Events:
  Type    Reason            Age   From                        Message
  ----    ------            ----  ----                        -------
  Normal  ChainPolicy       20m   fleetautoscaler-controller  FleetAutoscaler 'chain-fleet-autoscaler' successfully applied ChainPolicy | ID: weekday | Type: Schedule
  Normal  AutoScalingFleet  20m   fleetautoscaler-controller  Scaling fleet fleet-example from 2 to 50
karavindh@karavindh:~$ 
```

indexjoseph:
> @indexjoseph I added a `LastAppliedPolicy` field to track the previously applied policy. Now, logs will only show when the policy actually changes (e.g., from default to weekend). This should prevent the repeated event entries you pointed out. Let me know if that works!

Yep, LGTM!

pkg/apis/autoscaling/v1/fleetautoscaler.go
```diff
@@ -301,6 +301,11 @@ type FleetAutoscalerStatus struct {
 	// ScalingLimited indicates that the calculated scale would be above or below the range
 	// defined by MinReplicas and MaxReplicas, and has thus been capped.
 	ScalingLimited bool `json:"scalingLimited"`
+
+	// LastAppliedPolicy is the ID of the last applied policy in the ChainPolicy.
+	// Used to track policy transitions for logging purposes.
+	// +optional
```
vicentefb, line 307: why is this optional ?

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -253,9 +263,36 @@ func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32,
...
 	}
 
+	loggerForFleetAutoscalerKey(fasLog.fas.ObjectMeta.Name, fasLog.baseLogger).Debugf(
+		"Fleet Autoscaler operation completed for fleet: %s, with BufferPolicy: %v", f.ObjectMeta.Name, b.BufferSize)
+
 	return replicas, scalingInLimited || scalingOutLimited, nil
 }
 
+// New function to call applyCounterOrListPolicy
+func applyCounterOrListPolicyWrapper(c *autoscalingv1.CounterPolicy, l *autoscalingv1.ListPolicy,
```
vicentefb, line 273: The logging in `applyCounterOrListPolicyWrapper` happens only if applyCounterOrListPolicy returns nil for the error, right? 

This means successful applications of the policy are logged, but failures within `applyCounterOrListPolicy` are not logged with the same "Fleet Autoscaler operation completed" message. You might want to log regardless of the error status to ensure all attempts are recorded (?)

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -214,13 +215,22 @@ func applyWebhookPolicy(w *autoscalingv1.WebhookPolicy, f *agonesv1.Fleet) (repl
...
+
 	if faResp.Response.Scale {
 		return faResp.Response.Replicas, false, nil
 	}
+
 	return f.Status.Replicas, false, nil
 }
 
-func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet) (int32, bool, error) {
+func applyBufferPolicy(b *autoscalingv1.BufferPolicy, f *agonesv1.Fleet, fasLog *FasLogger) (int32, bool, error) {
```
vicentefb, line 233: Just confirming and double checking, could there be a potential integer overflow ? This is out of scope for this PR but just wondering.

pkg/fleetautoscalers/fleetautoscalers.go
```diff
@@ -404,31 +444,50 @@ func applyChainPolicy(c autoscalingv1.ChainPolicy, f *agonesv1.Fleet, gameServer
 	replicas := f.Status.Replicas
 	var limited bool
 	var err error
+	var chainEntry autoscalingv1.FleetAutoscalerPolicyType
 
 	// Loop over all entries in the chain
 	for _, entry := range c {
```
vicentefb, line 450: The applyChainPolicy iterates through the policies in the chain. If an error occurs for a specific policy, the loop continues. Only if all policies in the chain fail will the function return an error. The logging and event emission in `applyChainPolicy` happen only if an error occurs or if there's a change in the applied policy. Is this interpretation correct @indexjoseph I might have forgotten.

vicentefb (commented):
I left some questions

install/helm/agones/templates/crds/fleetautoscaler.yaml
```diff
@@ -93,6 +93,8 @@ spec:
                   type: boolean
                 scalingLimited:
                   type: boolean
+                lastAppliedPolicy:
+                  type: string
```
igooch, line 97: Per the CRD policy, this field either needs to be optional, or have a default value. In this case it looks like the controller needs to have this field, so there should be a default value. https://github.com/googleforgames/agones/edit/main/install/helm/agones/README.md has the details.

```suggestion
                  type: string
                  default: ""
```

pkg/client/applyconfiguration/autoscaling/v1/fleetautoscalerstatus.go
```diff
@@ -77,3 +79,11 @@ func (b *FleetAutoscalerStatusApplyConfiguration) WithScalingLimited(value bool)
 	b.ScalingLimited = &value
 	return b
 }
+
+// WithLastAppliedPolicy sets the LastAppliedPolicy field in the declarative configuration to the given value
+// and returns the receiver, so that objects can be built by chaining "With" function invocations.
+// If called multiple times, the LastAppliedPolicy field is set to the value of the last call.
```
igooch, line 85: The first sentence of this comment is confusing. Could you reword?

igooch (changes requested):
Is there a unit or e2e test that checks that the last applied policy works as expected? And throws handles errors as expected?

pkg/fleetautoscalers/fleetautoscalers.go
0xaravindh: I haven't added any error loggin in the fleetautoscalers.go file. So, if something goes wrong, the default error handling will take care of it.

https://github.com/0xaravindh/agones/blob/490d764eaa0a78041209369097c6409bd48661fb/pkg/fleetautoscalers/controller.go#L343-L344

pkg/apis/autoscaling/v1/fleetautoscaler.go
0xaravindh: removed!

pkg/fleetautoscalers/fleetautoscalers.go
0xaravindh: We can’t pass too many parameters easily in a function, so I put them into a separate struct. what are the cons here? Could you explain a little bit more?

pkg/client/applyconfiguration/autoscaling/v1/fleetautoscalerstatus.go
0xaravindh: we can't edit this file. please correct me if i'm wrong

https://github.com/googleforgames/agones/blob/533ff3fc0ef29cd8424466012796bab45759ed19/pkg/client/applyconfiguration/autoscaling/v1/fleetautoscalerstatus.go#L15-L17

install/helm/agones/templates/crds/fleetautoscaler.yaml
0xaravindh: Updated now!

0xaravindh:
> Is there a unit or e2e test that checks that the last applied policy works as expected? And throws handles errors as expected?

Yes, I added unit and e2e tests to verify LastAppliedPolicy and error handling.

pkg/fleetautoscalers/fleetautoscalers.go
vicentefb: Yeah, maybe we can create an Issue to enhance logging so that we log regardless of the error status to ensure all attempts are recorded. Not sure if this could clog the log.

pkg/fleetautoscalers/fleetautoscalers.go
vicentefb: Nevermind, I was imagining a very extreme scenario. For example, if AllocatedReplicas = 2,000,000,000 and BufferSize (as int32) = 200,000,000, the sum 2,200,000,000 would overflow int32.

vicentefb (approved):
lgtm, i believe Ivy requested changes

test/e2e/fleetautoscaler_test.go
```diff
@@ -1301,6 +1301,20 @@ func TestCounterAutoscalerAllocatedMultipleNamespaces(t *testing.T) {
 				return fas.Status.CurrentReplicas > 0
 			})
 
+			// Wait for LastAppliedPolicy to be set to CounterPolicyType for fasA
+			framework.WaitForFleetAutoScalerCondition(t, fasA, func(log *logrus.Entry, fas *autoscalingv1.FleetAutoscaler) bool {
```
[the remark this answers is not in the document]
0xaravindh: I added couple of testcase for `TestChainAutoscaler` and `TestAutoscalerWebhook`

igooch (approved):
LGTM

PR #4138, opened 2025-03-31
Promote FeatureDisableResyncOnSDKServer to Stable

igooch:
**What type of PR is this?**

/kind feature

FeatureDisableResyncOnSDKServer has been in Beta, and on by default, for 11 months with no issues filed. This PR is to promote the feature to Stable, which means that the feature now cannot be turned off.

We maintain the feature flag name, so that on upgrade a new controller does not fail to parse the feature flag name.

Closes # 4137

This PR does not warn that setting a stable feature flag to false is a no-op.
This PR does not include instructions for upgrading a SDK feature from Beta -> GA.

Those should be handled as separate issues / PRs.


markmandel:
So `TestAllocatorWithDeprecatedRequired` has been consistently flaky for me too, and it was driving me up the wall. I'm glad (?) to see it wasn't just me 😁

igooch:
Looks like it's tired of being ignored 😁

markmandel:
I've noticed it seems to be particularly that whole suite -- the Fleet just won't spin up. It also seems to only be on Autopilot.

Now sure if the clusters have hit quota maybe?

gongmax (approved):

PR #4143, opened 2025-04-03
Various e2e test improvements

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

* Better eviction test
* Extra test for communication on shutdown
* Extra test logging and cleanup

N/A

N/A

### DIFF
--- a/test/e2e/allocator_test.go
+++ b/test/e2e/allocator_test.go
@@ -72,7 +72,8 @@ func TestAllocatorWithDeprecatedRequired(t *testing.T) {
 	} else {
 		flt, err = helper.CreateFleet(ctx, framework.Namespace, framework)
 	}
-	assert.NoError(t, err)
+	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


markmandel:
Checks passed on re-run. Haven't seen that unit test flake before though I don't think 🤔

test/e2e/allocator_test.go
```diff
@@ -72,7 +72,8 @@ func TestAllocatorWithDeprecatedRequired(t *testing.T) {
 	} else {
 		flt, err = helper.CreateFleet(ctx, framework.Namespace, framework)
 	}
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
gongmax, line 75: why change this to require?

markmandel: Because if it's `assert` the test will continue past this point, and the test ends up panicing because `flt` is nil... which means it won't retry.

Better to stop immediately if you can't create the Fleet, since without the Fleet, there is no point in running any more of the test.

gongmax (approved):

Issue #3642, opened 2024-02-12
SDK Server: Adopt Sidecar Containers


markmandel:
**Is your feature request related to a problem? Please describe.**

The current [Health Failure Strategy and lifecycle](https://agones.dev/site/docs/guides/health-checking/#health-failure-strategy) for the sdkserver and game server container works, but it has edge cases, and is eventually consistent - which can also be fun (but not always in a good way).

This means as new ways for Pods to fail come up we have build out new features to capture this, and it's not straightforward, and often puts more load on the K8s control plane.

**Describe the solution you'd like**

Once Agones supports Kubernetes 1.29+ (or maybe earlier behind a feature flag?) we could move the sdkserver to the new [Sidecar container](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) model.

Then we can set the Pod to `restart: Never` by default, and the sdkserver sidecar to `restart: Always`, which would simplify things greatly.

* Alot of https://github.com/googleforgames/agones/blob/main/pkg/gameservers/health.go would likely dissapear, especially around setting `GameServerReadyContainerIDAnnotation` to know if the container is ready or not.
* The sdkserver only shutting down on `Shutdown` state also goes away, since the sidecar container can be spun up before and stay alive for the entire duration of the Pod.

There is one thing that is definitely tricky here - the way we have things setup now, if a GameServer is before Ready, we let it restart. If it's after Ready, we do not. Maybe we should revisit this as a pattern we still need for simplification - especially if the Pod crashes before being Ready, a new one will be recreated 🤔 Although this breaks backward compatibility, and that is less than ideal.

**Describe alternatives you've considered**
Leave things as they are, they seem to mostly work!

**Additional context**

N/A

markmandel:
Just copying a comment from an offline conversation with a user -- "the atomic non-restart to Unhealthy is way more important to us than being able to restart a game server process before it moves to Ready".

swermin:
We encountered this problem recently and I do agree that simplifying this flow would be a tremendous help.
We have our pods set to `restart: Never` since we do want the game servers to cycle as fast as possible. But with this setup if the game server `terminates` before reaching the `Ready` state then that pod is forever in a state where we cannot re-use it.
[...]

PR #4146, opened 2025-04-07
Adoption of Sidecar Containers

markmandel:
**What type of PR is this?**
>

/kind breaking

> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

This is a breaking change for both how health checking works and to the SDK lifecycles, but behind the `SidecarContainer` feature flag.

1. This moves the sdkserver into a initContainer with a `RestartPolicy` of always. This means that it will be alive for the entirety of the main Pod's containers.
2. There is no waiting for the Shutdown signal any more for the sidecar container before shutting down since it now natively waits for the main Pod containers.
3. The RestartPolicy of the `gameserver` container defaults to `Never` (although it can be changed), and no longer restarts before the `Ready` state to simplify lifecycle health checking.
4. Various hacky health checks are now simplified to "is the Pod in a `Failed` state.", and therefore we no longer need to do as much text searching for edge cases.

Work on #3642

Next: Write the docs for this work.


markmandel:
@igooch @gongmax can you cycle the autopilot clusters please? 😄

markmandel:
Conflicts fixed. Gentle bump on review plz :)

peterzhongyi:
Thanks Mark! I'll take a look! Coincidentally, I just met with a potential customer who's evaluating Agones and they were specifically interested in GameServer's pod restart policy induced edge cases. So this PR might change a lot of things!

markmandel:
Looks like it's time to recycle the autopilot clusters please.

igooch:
> Looks like it's time to recycle the autopilot clusters please.

Updating now.

build/Makefile
```diff
@@ -334,7 +334,7 @@ test-e2e-integration: GO_E2E_TEST_INTEGRATION_ARGS ?=\
 test-e2e-integration: $(ensure-build-image)
 	echo "Starting e2e integration test!"
 ifdef E2E_USE_GOTESTSUM
-	$(GOTESTSUM) --packages=$(agones_package)/test/e2e -- $(go_test_args) -timeout=30m $(ARGS) -args \
+	$(GOTESTSUM) --packages=$(agones_package)/test/e2e -- $(go_test_args) -timeout=45m $(ARGS) -args \
```
markmandel, line 337: Revert back down to 30m?

pkg/gameservers/controller.go
```diff
@@ -955,11 +955,37 @@ func (c *Controller) syncGameServerRequestReadyState(ctx context.Context, gs *ag
...
+	if addressPopulated {
+		c.recorder.Event(gs, corev1.EventTypeNormal, string(gs.Status.State), "Address and port populated")
+	}
+	c.recorder.Event(gs, corev1.EventTypeNormal, string(gs.Status.State), "SDK.Ready() complete")
+	return gs, nil
+}
+
+// applyGameServerReadyContainerIDAnnotation updates the GameServer and its corresponding Pod with an annotation
+// indicating the ID of the container that is running the game server, once it's in a Running state.
+func (c *Controller) applyGameServerReadyContainerIDAnnotation(ctx context.Context, gsCopy *agonesv1.GameServer, pod *corev1.Pod) (*agonesv1.GameServer, error) {
```
peterzhongyi, line 978: Is this function modifying gsCopy at all? I seem to only see various kinds of reads on it.

pkg/gameservers/health.go
```diff
@@ -107,12 +107,17 @@ func NewHealthController(
 // isUnhealthy returns if the Pod event is going
 // to cause the GameServer to become Unhealthy
 func (hc *HealthController) isUnhealthy(pod *corev1.Pod) bool {
-	return hc.evictedPod(pod) || hc.unschedulableWithNoFreePorts(pod) || hc.failedContainer(pod)
+	return hc.evictedPod(pod) || hc.unschedulableWithNoFreePorts(pod) || hc.failedContainer(pod) || hc.failedPod(pod)
```
peterzhongyi, line 110: Can you explain more detailedly why unschedulableWithNoFreePorts and failedContainer checks can be omitted, and evictedPod can't be omitted with the addition of the failedPod? I know it's probably related to "not needing to know the GS container is ready or not", but it is not crystal clear to me why these specific checks can be omitted.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -788,11 +788,25 @@ func (gs *GameServer) Pod(apiHooks APIHooks, sidecars ...corev1.Container) (*cor
...
-	// Put the sidecars at the start of the list of containers so that the kubelet starts them first.
-	containers := make([]corev1.Container, 0, len(sidecars)+len(pod.Spec.Containers))
-	containers = append(containers, sidecars...)
-	containers = append(containers, pod.Spec.Containers...)
-	pod.Spec.Containers = containers
+	if runtime.FeatureEnabled(runtime.FeatureSidecarContainers) {
+		// make sure all sidecars have a restart policy of Always, so they are valid sidecar containers.
+		// https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#sidecar-containers-and-pod-lifecycle
+		always := corev1.ContainerRestartPolicyAlways
+		for i := range sidecars {
```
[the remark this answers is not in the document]
markmandel: I wrote this signature at the beginning, to set things up if we ever decided to add more than one sidecar. It was just a nice way to keep things open if we ever needed it.

pkg/gameservers/controller.go
markmandel: Doesn't look like it - would you prefer the signature to be `gs` rather than `gsCopy`. I don't mind either way.

pkg/gameservers/health.go
markmandel: Since Kubernetes is not restarting the container anymore, we don't have to look at error message and the like to see if the game server containers crashed, or any other kind of edge condition (which is what `unschedulableWithNoFreePorts` and `failedContainer` did). Now we just need to check if the Pod is failed.

I left in the check for the Evicted reason, mainly because I wasn't 100% sure if an Evicted pod would end up with a failed pod or not, and the check is so simple, if it was a double check, there would be no harm.

pkg/apis/agones/v1/gameserver.go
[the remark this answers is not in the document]
peterzhongyi: Sounds good

pkg/gameservers/health.go
peterzhongyi: Thanks for the explanation!

pkg/gameservers/controller.go
peterzhongyi: I was asking more about why the gsCopy was even returned by the function if it is just being read? In this case, maybe the function should just return error and exclude *agonesv1.GameServer?

Also, in the function description you mentioned the "applyGameServerReadyContainerIDAnnotation updates the GameServer" - maybe adjust the description if no game server is being updated?

markmandel: Ah, it does edit the `gsCopy` here:

https://github.com/googleforgames/agones/blob/bcd0c0b3fc0ba8d032fa132df1c4d766b3a747f6/pkg/gameservers/controller.go#L994

markmandel: Just gently bumping - I think this should be good to merge now, yeah?

peterzhongyi (approved):

Issue #3992, opened 2024-09-20
Poor packing with agones-allocator and Counter for high-capacity game servers


dmorgan-blizz:
**What happened**:

`Allocated` game servers are not efficiently packed, even though they have plenty of Counter space:
![game_players](https://github.com/user-attachments/assets/b424c2fe-2015-41b2-8647-1183b4ffd795)

As you can see, when making allocation requests, Agones starts to fill up the first server, but then, before that server is anywhere close to its 300 Counter capacity, it starts sending players to the `Ready` buffer server. Now there are multiple game servers active, but even though those servers also have plenty of space, Agones will still occasionally allocate another `Ready` buffer server.

This also occurred when scaling `agones-allocator` replicas down to 1 (from 3), and actually, the packing was _worse_ in this scenario, even with an allocation rate as low as 5/s.

**What you expected to happen**:
Agones should always pick an `Allocated` game server with available `Counter` capacity before choosing a `Ready` buffer server.

**How to reproduce it (as minimally and precisely as possible)**:
Game server settings like the following:
```
Kind: FleetAutoscaler
Spec:
  Policy:
    Type: Buffer
    Buffer:
      Buffer Size: 1
```

```
kind: Fleet
spec:
  scheduling: Packed
  template:
    spec:
      counters:
        players:
          capacity: 300
```

where a game server runs for a period of time and players continually cycle in and out (using SDK's `DecrementCounterAsync(string key, long amount)` upon game end);

and an allocation request like the following
```csharp
GameServerSelectors = {
    new GameServerSelector {
        GameServerState = GameServerSelector.Types.GameServerState.Allocated,
        MatchLabels = {
            { "version", "1.2.3" },
        },
        Counters = {
            {
                "players", new CounterSelector {
                    MinAvailable = 1,
                }
            },
        },
    },
    new GameServerSelector {
        GameServerState = GameServerSelector.Types.GameServerState.Ready,
        MatchLabels = {
            { "version", "1.2.3" },
        },
        Counters = {
            {
                "players", new CounterSelector {
                    MinAvailable = 1,
                }
            },
        },
    },
},
Counters = {
    {
        "players", new CounterAction {
            Action = "Increment",
            Amount = 1,
        }
    },
},
```

**Anything else we need to know?**:
agones-allocator game server state does not seem to be shared between replicas (which seems like it should also be fixed, maybe Redis or something?), but as mentioned, even with only one replica, the packing was as bad or worse

**Environment**:
[...]

PR #4159, opened 2025-04-17
Allocation: Re-cache allocated `GameServer`

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

We want the allocated `GameServer` back in the cache anyway for re-allocation, so might as well get it back in as soon as possible.

Work on #3992


igooch:
@dmorgan-blizz FYI this is one proposal for your issue #3992. Let us know how this performs.

markmandel:
> @dmorgan-blizz FYI this is one proposal for your issue #3992. Let us know how this performs.

There should be no downside to this change, it should help no matter what.

markmandel:
Any blocking reason this shouldn't be merged?

peterzhongyi (commented):
Should we worry about the informer giving a stale update on the game server that overwrites the game server record you write here? Seems like it might happen frequently given that the informer is almost always going to be slower but will still try to write the the cache?

markmandel:
🤔 I can't think of an example, since the Informer is only going to give you the latest version of the `GameServer` instance.

We could do a generation check on store, and if it's less than the latest version that is stored in there (which is possibly not a bad idea anyway), we reject it.

https://github.com/googleforgames/agones/blob/daafb8fc2c9bc76e0f15d2bffe77243fbdbe2e0e/pkg/gameserverallocations/cache.go#L32-L38

Actually, that's a pretty good idea anyway, I'll put that in 👍🏻 seems like thing that shouldn't happen often (ever?), but easy enough to make sure it doesn't.

peterzhongyi:
Sounds good. LGTM!

pkg/gameserverallocations/cache.go
```diff
@@ -29,12 +29,26 @@ type gameServerCache struct {
...
 	defer e.mu.Unlock()
 	if e.cache == nil {
 		e.cache = map[string]*agonesv1.GameServer{}
 	}
+
+	// Check if there's already a value for this key
+	if existingGS, ok := e.cache[key]; ok {
+		// If the incoming gs object's Generation is less than the existing one, ignore it, but also check the ResourceVersion
+		// saves us a deep copy if we don't need to
+		if gs.ObjectMeta.Generation < existingGS.ObjectMeta.Generation || gs.ObjectMeta.ResourceVersion == existingGS.ObjectMeta.ResourceVersion {
```
markmandel, line 47: I had forgotten that you can get `resourceVersion` change, but with no `Generation` change, so you have to account for it, and lean on eventual consistency when it's needed.

PR #4160, opened 2025-04-17
Creates Feature Flag to Prevent Removal of an Allocated Game Server from the Allocation Cache

igooch:
**What type of PR is this?**

/kind feature

Removing a game server from the allocation cache when using Counters and Lists causes the allocator to allocate more game servers than needed. This PR adds a feature flag to prevent the removal of an allocated game server from the cache. 

Because the game server is not removed from the cache, and the intent of packing a game server until a Counter or List reaches capacity involves _wanting_ to allocate to the same game server, there will be resource (write) conflicts. In the sample allocation request we use a short `allocationBatchWaitTime` time, as well as throttling of current requests to control the rate of allocation and reduce the number of resource conflicts. These numbers should be tweaked during testing to find the best throughput of allocations for a particular game without overloading the Kubernetes control plane.

Working on #3992

* This draft includes a sample fleet, sample allocation request, modification of the `allocationBatchWaitTime` of the dev installation, and debug statements that will not be included in the final PR.

### DIFF
--- a/pkg/gameserverallocations/allocator.go
+++ b/pkg/gameserverallocations/allocator.go
@@ -464,6 +464,15 @@ func (c *Allocator) allocate(ctx context.Context, gsa *allocationv1.GameServerAl
 
 	select {
 	case res := <-req.response: // wait for the batch to be completed
+		// TODO: Remove debug statements
+		if counter, ok := res.gs.Status.Counters["players"]; ok {
+			if counter.Count == 1 {
+				fmt.Println("FIRST ALLOCATION", res.gs.Name)
+			}
+			if counter.Count == counter.Capacity {
+				fmt.Println("LAST ALLOCATION", res.gs.Name)
+			}
+		}
 		return res.gs, res.err
 	case <-ctx.Done():
 		return nil, ErrTotalTimeoutExceeded
@@ -471,7 +480,7 @@ func (c *Allocator) allocate(ctx context.Context, gsa *allocationv1.GameServerAl
 }
 
 // ListenAndAllocate is a blocking function that runs in a loop
-// looking at c.requestBatches for batches of requests that are coming through.
+// looking at c.pendingRequests for batches of requests that are coming through.
 func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int) {
 	// setup workers for allocation updates. Push response values into
 	// this queue for concurrent updating of GameServers to Allocated
@@ -557,13 +566,16 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
 				req.response <- response{request: req, gs: nil, err: err}
 				continue
 			}
-			// remove the game server that has been allocated
-			list = append(list[:index], list[index+1:]...)
 
-			if err := c.allocationCache.RemoveGameServer(gs); err != nil {
-				// this seems unlikely, but lets handle it just in case
-				req.response <- response{request: req, gs: nil, err: err}
-				continue
+			if !runtime.FeatureEnabled(runtime.FeatureReuseAllocated) {
+				// remove the game server that has been allocated
+				list = append(list[:index], list[index+1:]...)
+
+				if err := c.allocationCache.RemoveGameServer(gs); err != nil {
+					// this seems unlikely, but lets handle it just in case
+					req.response <- response{request: req, gs: nil, err: err}
+					continue
+				}
 			}
 
 			updateQueue <- response{request: req, gs: gs.DeepCopy(), err: nil}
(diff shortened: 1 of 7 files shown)


pkg/gameserverallocations/allocator.go
```diff
@@ -557,13 +566,16 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
...
 				continue
 			}
-			// remove the game server that has been allocated
-			list = append(list[:index], list[index+1:]...)
 
-			if err := c.allocationCache.RemoveGameServer(gs); err != nil {
-				// this seems unlikely, but lets handle it just in case
-				req.response <- response{request: req, gs: nil, err: err}
-				continue
+			if !runtime.FeatureEnabled(runtime.FeatureReuseAllocated) {
```
markmandel, line 570: ...but if if you do this, the allocated game server will be out of date if you don't remove it, so you'll get _more_ conflicts rather than less.

Unless I'm missing something?

markmandel:
I think #4159 might at least be part of what you actually want.

pkg/gameserverallocations/allocator.go
igooch: This is the only way I found (without building out a new caching layer) to force bin packing of Counters or Lists. I.e. every single allocation request should go to the same game server. Which does result in write conflicts as the trade off.

markmandel: > Which does result in write conflicts as the trade off.

Which is why this is a really bad idea.

markmandel: I wonder if there's a way to put an Allocated game server back in the `list` after it is allocated.

It's never going to be perfect since there are multiple allocators each with their own list, but it will improve things somewhat.

markmandel (changes requested):

igooch:
FWIW the idea here is not to bring this feature to stable, it's to leave it as a dev or alpha feature flag (off by default) until, or if, there's a more involved solution around a shared allocator cache. This feature only makes sense if the user has a Counter/List, _and_ they want to fill it to capacity before allocating the next game server.

The point of not removing the game server from the list here is so that it will force all allocators to bin pack. If the game server is removed and then placed back in the list you still end up allocating additional unnecessary game servers. It's a trade off of a slower cap on allocation rate vs. not well bin packed.

markmandel:
> FWIW the idea here is not to bring this feature to stable, it's to leave it as a dev or alpha feature flag (off by default) until, or if, there's a more involved solution around a shared allocator cache. 

Yeah, but the premise isn't sound for what you are stating anyway, as least from my perspective.

> This feature only makes sense if the user has a Counter/List, _and_ they want to fill it to capacity before allocating the next 
> game server.

If you have multiple Allocators, this won't do that though. You're always going to hit the disparate cache issue anyway.

> The point of not removing the game server from the list here is so that it will force all allocators to bin pack. 

Yeah, but that won't work -- if you don't remove it from the list, if there is an attempt to allocate it again, it will fail with a conflict, and therefore won't bin pack.

> If the game server is removed and then placed back in the list you still end up allocating additional unnecessary game servers. > It's a trade off of a slower cap on allocation rate vs. not well bin packed.

Not if you (a) put back the freshly allocated game server and (b) put it back in the correct sorting order (binary search is pretty fast!).

This doesn't solve the multi-allocator problem, and there is likely some race conditions in here, but it actually gets you closer to your intended goal than what you have now.

markmandel:
Shall we close this and clean things up a little?

dmorgan-blizz:
Unfortunately, https://github.com/googleforgames/agones/pull/4159 did not solve our allocation problems ;( we see very similar behavior on 1.49 as we did on 1.44, where more gameservers are spun up than needed. For instance

image

We have not had a chance to test these additional changes, but it appears some solution that caches outside of the allocator process itself is needed (we're running what I believe is the standard min 3 allocators)

markmandel:
@dmorgan-blizz have you seen the conversations and work happening on #4190 ? That should replace this PR.

igooch:
This works as a simple bin packing fix when used with a significantly lower `allocationBatchWaitTime`. With with tweaks to the `allocationBatchWaitTime` and client-side throttling it can handle many requests (can't find my old test runs at the moment with exact numbers) with low conflicts. 

That being said, it looks like [miai10](https://github.com/miai10) has some time to work on a the complex solution. We can close, and if you need something sooner @dmorgan-blizz you can test on this PR even if it's closed, and reopen if needed.

dmorgan-blizz:
I hadn't seen https://github.com/googleforgames/agones/issues/4190 that sounds great! I'm fine with waiting for that :)

PR #4161, opened 2025-04-23
Adds explicit permissions for github workflows

igooch:
**What type of PR is this?**

/kind hotfix

GitHub is now enforcing explicit permissions in workflows, causing the labeler and pr_update workflows to fail.

This adds the permissions to write to a PR, using the same permissions as the existing labeler-pr workflow.

NA

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -20,6 +20,9 @@ on:
 jobs:
   labeler:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
     name: Label the PR size
     steps:
       - uses: codelytv/pr-size-labeler@v1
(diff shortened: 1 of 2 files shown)


markmandel (approved):
TIL!

markmandel:
Agreed - seems rather flaky. @igooch does it make sense to `cat` the log of the file on failure? Just so we can see what's up?

lacroixthomas:
> Agreed - seems rather flaky. @igooch does it make sense to `cat` the log of the file on failure? Just so we can see what's up?

I was looking at the code, from what I understand, it reuses an existing cluster to do the upgrade test, maybe there is some leftovers / issues from a previous setup or something on it, as it's specific to this clusters (the others looks fine with 1.30 and 1.32)

PR #4176, opened 2025-05-07
Feat: improve poor sessions packing using batch allocator

### LINKED ISSUE #3992: Poor packing with agones-allocator and Counter for high-capacity game servers

miai10:
**What type of PR is this?**

/kind feature

This PR improves the packing of sessions on game servers but batching multiple allocation into one update of the game server CR.

There are 2 major changes:
- the batch allocator implementation (and tests)
- enabling retries for ErrNoGameServer instead of quickly returning 400

Optionally, I have added 2 new metrics (gameserver_allocations_batch_size, gameserver_allocations_updates_duration_seconds) that I used for investigations but they can be discarded.

**The batch allocator**
The current allocator favors spreading the allocations from a batch on multiple servers by picking one and removing it from the availability list for the next allocation (see [this](https://github.com/googleforgames/agones/blob/main/pkg/gameserverallocations/allocator.go#L561)). This leads to many updates of the game servers CR, using multiple game servers sometimes leading to ErrNoGameServer if the pool is small. For high capacity game servers, this approach is not well suited, see [this discussions](https://github.com/googleforgames/agones/issues/3992). 

The difference between this and the original allocator is that this will apply the allocation to the local gs (still removing it from the cache but not from the local available servers list) and continue with the next allocation from the batch. When the batch is done, the update workers will try to update each newly allocated gs with the last state.

batchAllocationUpdateWorkers tries to update each newly allocated gs with the last state. If the update fails because of a version conflict, all allocations that were applied onto a gs will receive an error, thus being available for retries. If the update succeeds, all allocations that were applied onto a gs will succeed, and the gs with the updated state will be added back to the cache.

This should work fine even with the one session per server game servers. 

**No quick ErrNoGameServer**
Not having a game server to allocate is a transitory state of the system in high capacity servers as servers can be just out of the cache (while being updated), not necessary our of capacity. So it makes sense to continue with retries and then return ErrNoGameServer. The simple example is: with 2 available game servers and 3 pending allocations in a batch, the 3rd allocation will not get a game server and fail instantly, instead of retrying. See TestAllocatorAllocateNoQuickNoGameServerError.

**Results**
The tests were done with my setup: custom gs implementation, Agones 1.45, 3 allocators in cluster, 100ms batch time. For 20 allocations per second, 100 allocations total, there was ~5% allocations failures and a pattern for spread: one with >50% usage, second with 20-30% usage, the other with <10% usage and the 6th or greater with 1 or 0% usage.

With the *re-cache allocated game server* (that Mark just merged) and *no quick ErrNoGameServer* optimizations, 1 allocator and 50ms batch time, I was able to use only 2 game servers with 15 allocations per second at <5% failures. With 5 allocations per second there was 0% failures.

With the batch allocator and the above optimizations and setup (only the batch time went to 300ms, but still 1 allocator), I am able to have tests with 15 allocations per second and 0% failures and no spread (all allocations on one game server).

**Discussion**
The results improved greatly because of the batch allocator and using 1 allocator. Using many allocators generates a lot of CR update failures since allocators select the same most suited game server. I don't have a strategy for fixing that, having at least 2 allocators would be more HA...

Right now the bottleneck is the time it takes for the allocator to get the allocation acknowledgement from the game server implementation (not really an acknowledgment as it is not waiting for it but the gs does updates to the CR thus creating potential update failures on the allocator side).

Closes https://github.com/googleforgames/agones/issues/3992

The new metrics are optional, I used them for the investigations.
I favored implementing stuff in a new file so that it is easier to review. The changes can be merged into the allocator.go file if ifs all over the place.


pkg/gameserverallocations/metrics_additional.go
```diff
@@ -0,0 +1,57 @@
...
+var (
+	gameServerAllocationsUpdatesLatency = stats.Float64("gameserver_allocations/update_latency", "The duration of gameserver updates", "s")
+	gameServerAllocationsBatchSize      = stats.Int64("gameserver_allocations/batch", "The gameserver allocations batch size", "1")
+)
+
+func init() {
+
+	stateViews := []*view.View{
+		{
+			Name:        "gameserver_allocations_updates_duration_seconds",
```
lacroixthomas, line 22: Maybe worth adding it in the metrics doc https://github.com/googleforgames/agones/blob/main/site/content/en/docs/Guides/metrics.md ? Not sure if the dev feature flag should be in though ?

pkg/util/runtime/features.go
```diff
@@ -160,6 +163,7 @@ var (
 		FeatureSidecarContainers:             false,
 
 		// Dev features
+		FeatureAllocatorBatchesUpdates: false,
```
lacroixthomas, line 166: I'm not sure about the `dev` feature flags, but we might want to add it to these files with false as default values:

https://github.com/googleforgames/agones/blob/main/cloudbuild.yaml#L444
https://github.com/googleforgames/agones/blob/main/build/Makefile#L76

lacroixthomas: Actually, dev and alpha are disabled by default you can ignore this comment 😄

pkg/gameserverallocations/metrics_additional.go
miai10: Good catch: either I use the feature flag for those metrics or I make them available for the current allocator too. Or drop them since they are used only for performance investigations like I am doing right now. What do you suggest to do?

pkg/gameserverallocations/batch_allocator.go
```diff
@@ -0,0 +1,266 @@
...
+	batchResponsesPerGs := make(map[string]batchResponses)
+
+	flush := func() {
+		if requestCount > 0 {
+			metrics.recordAllocationsBatchSize(ctx, requestCount)
+		}
+
+		for _, batchResponses := range batchResponsesPerGs {
+			batchUpdateQueue <- batchResponses
+		}
```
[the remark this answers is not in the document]
miai10: Can you please explain this more, I don't think I understand the issue

igooch (commented):
Thanks for the PR! I ran a manual test and it appear to be working as intended. A couple comments, but generally looks good.

pkg/gameserverallocations/allocator.go
```diff
@@ -174,7 +174,11 @@ func (c *Allocator) Run(ctx context.Context) error {
 	}
 
 	// workers and logic for batching allocations
-	go c.ListenAndAllocate(ctx, maxBatchQueue)
+	if runtime.FeatureEnabled(runtime.FeatureAllocatorBatchesUpdates) {
+		go c.ListenAndBatchAllocate(ctx, maxBatchQueue)
```
markmandel, line 178: 👍🏻 on a whole new file. Makes this much easier to manage.

pkg/gameserverallocations/batch_allocator.go
```diff
@@ -0,0 +1,266 @@
...
+	batchUpdateQueue := make(chan batchResponses)
+
+	for i := 0; i < workerCount; i++ {
+		go func() {
+			for {
+				select {
+				case batchRes := <-batchUpdateQueue:
+					if len(batchRes.responses) > 0 {
+						// The last response contains the latest gs state
+						lastGsState := batchRes.responses[len(batchRes.responses)-1].gs
```
markmandel, line 41: One thing I'm trying to wrap my head around -- are previously allocated GameServers searched first for subsequent allocations (over other GameServers in the sorted list?)

If so, that might actually break packing, since some types of allocations could delete items from a list/counter, leaving a GameServer more open than it was previously.

markmandel (commented):
Sorry for the delay on this, I've been sick for the past couple of weeks.

I'd like to spend some time digging through this, but on surface level it's a very interesting approach! I like the idea of batching multiple changes to a GameServer into a single update on the API.

The original allocation code was written when we only allocated Ready GameServers, so some of that assumption definitely still bleeds through now - ending up with the scenario you are trying to solve now.

I would also like to be able to look at any differences in performance between the original implementation and this one. It's probably scenario dependent, but we should double check to be sure.

pkg/gameserverallocations/batch_allocator.go
miai10: The state of the game server is just updated in place and the position in the list remains the same (it is removed only from the cache). 

I think a chosen game server can only get "more allocated" after applying the allocation, I don't see how it can get "more open". Do you have an example in mind?

pkg/util/runtime/features.go
miai10: I made the feature alpha to get it better tests coverage

pkg/gameserverallocations/metrics_additional.go
miai10: I made the new metrics work with the regular allocator too because the agones_gameserver_allocations_updates_duration_seconds one is important for the performance of the system: it's a big part of the allocation latency and it influences how many available game servers you have in the list.

pkg/gameserverallocations/batch_allocator.go
markmandel: > The state of the game server is just updated in place and the position in the list remains the same (it is removed only from the cache).

Neat!

Yeah - if you delete items from a list, or decrement a counter, it could become more open. Probably not something that happens often, but technically it is possible.

Should each item actually change position in the list according to its new sorting position, it's prioritized appropriately?

pkg/gameserverallocations/allocator.go
```diff
@@ -693,11 +697,10 @@ func Retry(backoff wait.Backoff, fn func() error) error {
 			}
 		}
 
+		// No quick 400s, still do retries for ErrNoGameServer
 		switch {
 		case err == nil:
 			return true, nil
-		case err == ErrNoGameServer:
-			return true, err
```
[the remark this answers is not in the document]
markmandel: Yeah, I like this change -- it's a artifact of when all we had was Ready GameServers. If Allocated GameServer immediately go back in the cache like they do now, this makes even more sense - since there should be Allocated items in there.

pkg/gameserverallocations/batch_allocator.go
```diff
@@ -0,0 +1,266 @@
...
+						// Try to update with the latest gs state
+						updatedGs, updateErr := c.gameServerGetter.GameServers(lastGsState.ObjectMeta.Namespace).Update(ctx, lastGsState, metav1.UpdateOptions{})
+						if updateErr != nil {
+							metrics.recordAllocationUpdateFailure(ctx, time.Since(requestStartTime))
+
+							if !k8serrors.IsConflict(errors.Cause(updateErr)) {
+								// since we could not allocate, we should put it back
+								// but not if it's a conflict, as the cache is no longer up to date, and
+								// we should wait for it to get updated with fresh info.
+								c.allocationCache.AddGameServer(updatedGs)
```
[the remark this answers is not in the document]
markmandel: oh that's a really good point. The original GS should go back in. That's a bug.

Basically it's a "if the K8s control plane falls over, then put it back in the cache, because it's likely not changed".

markmandel (commented):
Other fun question I have - for this to work, do you _have_ to only use a single instance of the Allocation Service?

I feel like contention would be worse with this? (I haven't tried it, just in my head).

Edit, Just saw this:

> The results improved greatly because of the batch allocator and using 1 allocator. Using many allocators generates a lot of CR update failures since allocators select the same most suited game server. I don't have a strategy for fixing that, having at least 2 allocators would be more HA...

Yeah, this would be _bad_.

markmandel:
In theory, you could tied k8s cilent leader to the readiness probe for the allocation service, so only one is available behind the Service at any point and time?

pkg/gameserverallocations/batch_allocator.go
miai10: The only case where the list is not prioritized appropriately is when gs1.CompareCountAndListPriorities(gsa.Spec.Priorities, gs2) will return different once a new allocation has been applied. Basically when the server has less priority than its neighbor following an allocation. I could add this check if you think it is worth it.

miai10:
> In theory, you could tied k8s cilent leader to the readiness probe for the allocation service, so only one is available behind the Service at any point and time?

That is an interesting idea that is worth trying.

miai10:
> Other fun question I have - for this to work, do you _have_ to only use a single instance of the Allocation Service?
> 
> I feel like contention would be worse with this? (I haven't tried it, just in my head).
> 
> Edit, Just saw this:
> 
> > The results improved greatly because of the batch allocator and using 1 allocator. Using many allocators generates a lot of CR update failures since allocators select the same most suited game server. I don't have a strategy for fixing that, having at least 2 allocators would be more HA...
> 
> Yeah, this would be _bad_.

The batching will work with any batch of allocations the allocator receives, it's internal to the allocator. 

Having just 1 allocator just adds to better packing. Maybe having a main allocator with a replica standing by as you suggested will fix the HA issue.

pkg/gameserverallocations/batch_allocator.go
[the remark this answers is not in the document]
miai10: Can you please explain why the most recent copy of the gs (updatedGs) should not be placed in the cache? It's like having a free GET.

lacroixthomas:
> In theory, you could tied k8s cilent leader to the readiness probe for the allocation service, so only one is available behind the Service at any point and time?

The issue with the readiness probe is that we'll have a pod in `unready` / `unhealthy`, the helm install would fail after some time as it would expect a specific amount of ready pods

I'm still investigating on how we could keep the pod in ready without receiving requests (and also still looking into the readinessprobe, maybe I missed something too)

pkg/gameserverallocations/batch_allocator.go
markmandel: (1) binary sort insert is not very expensive (2) if we want to be true to packing, we probably should

markmandel:
> The issue with the readiness probe is that we'll have a pod in unready / unhealthy, the helm install would fail after some time as it would expect a specific amount of ready pods

ah fair. Also leader election is not exactly the fastest thing in the world, so switchover on failure wouldn't be instantaneous.

lacroixthomas:
> > The issue with the readiness probe is that we'll have a pod in unready / unhealthy, the helm install would fail after some time as it would expect a specific amount of ready pods
> 
> ah fair. Also leader election is not exactly the fastest thing in the world, so switchover on failure wouldn't be instantaneous.

It's just a wild idea and it's probably overkill for this, but we could maybe setup a crd for the service, some kind of stand-by/active service ? That would ensure only a single pod is exposed from this service crd ? (Something to add in the controller)

What do you think about it ? 

Otherwise we can set a label once a pod is elected leader and setting it from the service's selector, but the failover takes a bit of time as well

markmandel:
> It's just a wild idea and it's probably overkill for this, but we could maybe setup a crd for the service, some kind of stand-by/active service ? That would ensure only a single pod is exposed from this service crd ? (Something to add in the controller)

Might as well just have a single pod deployment - it'll restart when something goes down anyway 😄 

Also just a vague idea. I was poking at: https://github.com/olric-data/olric for a distributed data storage that could be embedded into the allocation system.

One thing that did look useful was having pubsub capabilities - which could be a handy way of distributing information about sorted list across instances (some hand waving here, haven't worked out the details).

Thoughts?

lacroixthomas:
This could also works with batch allocation, depending on how we manage the pub sub, we could have multiple pods and batching them from a gs subscription or something, I like the idea, shall see with the performances (we'd get rid of the conflicts and keep the batching/ avoiding too many CR updates)

markmandel:
oooh the worker queue's could listen to pubsub rather than the channels? 🤔 that could be neat.

The only other potentially really tricky part could be managing the bits that do GSA vs the Allocation endpoints -- need to go back and look at the code again. It may well be fine.

markmandel:
> Let's use this thread to get the batch allocator merged.

So I'll push back here a little. This code works really well under _certain circumstances_, and isn't an actual solution that is generally applicable (i.e. there is no HA for allocation API servers)

So I'm quite hesitant to merge it in its current form. It feels like something we would eventually end up deleting in favour for a different implementation that does work across the board.

miai10:
> > Let's use this thread to get the batch allocator merged.
> 
> So I'll push back here a little. This code works really well under _certain circumstances_, and isn't an actual solution that is generally applicable (i.e. there is no HA for allocation API servers)
> 
> So I'm quite hesitant to merge it in its current form. It feels like something we would eventually end up deleting in favour for a different implementation that does work across the board.

So if we fix this first: https://github.com/googleforgames/agones/issues/4190, we should be ok with the batch allocator too? Should they be under the same feature gate?

Tbh, I don't see why batching the calls to the API server for the update of one gs would ever be removed since making fewer calls should be part of any optimization. And it offers better packing too. And this is no matter how many allocator there are or how we process the allocations batches (with a channel or with a message queue).

markmandel:
> So if we fix this first: https://github.com/googleforgames/agones/issues/4190, we should be ok with the batch allocator too? Should they be under the same feature gate?

I think so, yes! It's all part of the same goal 👍🏻 

To be clear - love the idea of batching the operations! Totally a fan! But you just said something that contradicts what you said earlier (emphasis is me):

> Tbh, I don't see why batching the calls to the API server for the update of one gs would ever be removed since making fewer calls should be part of any optimization. And it offers better packing too. **And this is no matter how many allocator there are or how we process the allocations batches (with a channel or with a message queue).**

vs original comment of:

> Using many allocators generates a lot of CR update failures since allocators select the same most suited game server. I don't have a strategy for fixing that, having at least 2 allocators would be more HA...

That's my specific concern -- assuming there _is_ an issue with multiple allocators, which from my mental model, I think there would be, (and is the default installation at the moment), by merging this PR in its current form, we risk accepting a solution in the code base that does **not** work for HA allocators, which means we'd never move it to stable -- which seems like a risky thing to take onboard as a maintenance burden.

If we can solve the HA issue - love it, 100% in 🥳 

I hope that clears up my positioning.

Also, this is a gnarly problem - so big appreciation for tackling it.

pkg/gameserverallocations/batch_allocator.go
miai10: Done

pkg/gameserverallocations/batch_allocator.go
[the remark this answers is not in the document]
miai10: @markmandel : this is the last open thread (besides the HA policy that Thomas is working on in another MRs). I think we should trust the docs and even if we get an error we should put the returned gs in the cache.

0xaravindh:
Please resolve the lint issues. @miai10 

```
pkg/gameserverallocations/allocation_cache_test.go:874:39: unnecessary conversion (unconvert)
				strategy = apis.SchedulingStrategy(testScenario.packingStrategy)
				                                  ^
pkg/gameserverallocations/allocation_cache_test.go:617:2: commentFormatting: put a space between `//` and comment text (gocritic)
	//gs0 := agonesv1.GameServer{ObjectMeta: metav1.ObjectMeta{Name: "gs0", Namespace: defaultNs, UID: "0"},
	^
pkg/gameserverallocations/allocation_cache.go:333:28: `startegy` is a misspelling of `strategy` (misspell)
			c.baseLogger.WithField("startegy", strategy).
			                        ^
pkg/gameserverallocations/allocation_cache.go:353:29: `startegy` is a misspelling of `strategy` (misspell)
// Used for the distributed startegy.
                            ^
pkg/gameserverallocations/allocation_cache.go:370:24: `startegy` is a misspelling of `strategy` (misspell)
// Used for the packed startegy.
                       ^
make: *** [Makefile:509: lint] Error 1
```

lacroixthomas:
> > So if we fix this first: #4190, we should be ok with the batch allocator too? Should they be under the same feature gate?
> 
> I think so, yes! It's all part of the same goal 👍🏻
> 
> To be clear - love the idea of batching the operations! Totally a fan! But you just said something that contradicts what you said earlier (emphasis is me):
> 
> > Tbh, I don't see why batching the calls to the API server for the update of one gs would ever be removed since making fewer calls should be part of any optimization. And it offers better packing too. **And this is no matter how many allocator there are or how we process the allocations batches (with a channel or with a message queue).**
> 
> vs original comment of:
> 
> > Using many allocators generates a lot of CR update failures since allocators select the same most suited game server. I don't have a strategy for fixing that, having at least 2 allocators would be more HA...
> 
> That's my specific concern -- assuming there _is_ an issue with multiple allocators, which from my mental model, I think there would be, (and is the default installation at the moment), by merging this PR in its current form, we risk accepting a solution in the code base that does **not** work for HA allocators, which means we'd never move it to stable -- which seems like a risky thing to take onboard as a maintenance burden.
> 
> If we can solve the HA issue - love it, 100% in 🥳
> 
> I hope that clears up my positioning.
> 
> Also, this is a gnarly problem - so big appreciation for tackling it.

I was thinking, would that be alright if we move that to a dev feature gate for it to be merged ? This way we'll be able to also test with my dev feature gate about the HA issue ? What do you think about it ? @markmandel @miai10 @igooch 

If we agree with that, it means that there will be some minor changes to do on this PR to move it from alpha to dev 

I'm gonna create some PRs soon, I'm doing some last tests with it at the moment, almost ready ! 😄

miai10:
> > > So if we fix this first: #4190, we should be ok with the batch allocator too? Should they be under the same feature gate?
> > 
> > 
> > I think so, yes! It's all part of the same goal 👍🏻
> > To be clear - love the idea of batching the operations! Totally a fan! But you just said something that contradicts what you said earlier (emphasis is me):
> > > Tbh, I don't see why batching the calls to the API server for the update of one gs would ever be removed since making fewer calls should be part of any optimization. And it offers better packing too. **And this is no matter how many allocator there are or how we process the allocations batches (with a channel or with a message queue).**
> > 
> > 
> > vs original comment of:
> > > Using many allocators generates a lot of CR update failures since allocators select the same most suited game server. I don't have a strategy for fixing that, having at least 2 allocators would be more HA...
> > 
> > 
> > That's my specific concern -- assuming there _is_ an issue with multiple allocators, which from my mental model, I think there would be, (and is the default installation at the moment), by merging this PR in its current form, we risk accepting a solution in the code base that does **not** work for HA allocators, which means we'd never move it to stable -- which seems like a risky thing to take onboard as a maintenance burden.
> > If we can solve the HA issue - love it, 100% in 🥳
> > I hope that clears up my positioning.
> > Also, this is a gnarly problem - so big appreciation for tackling it.
> 
> I was thinking, would that be alright if we move that to a dev feature gate for it to be merged ? This way we'll be able to also test with my dev feature gate about the HA issue ? What do you think about it ? @markmandel @miai10 @igooch
> 
> If we agree with that, it means that there will be some minor changes to do on this PR to move it from alpha to dev
> 
> I'm gonna create some PRs soon, I'm doing some last tests with it at the moment, almost ready ! 😄

Sounds good to me. I'll do the mode to dev shortly!

markmandel:
> I was thinking, would that be alright if we move that to a dev feature gate for it to be merged ?

In head, this would be bundled with your new processor allocation code @lacroixthomas - not sure if you are saying the same thing or not?

lacroixthomas:
> > I was thinking, would that be alright if we move that to a dev feature gate for it to be merged ?
> 
> In head, this would be bundled with your new processor allocation code @lacroixthomas - not sure if you are saying the same thing or not?

You are totally right, could bundled it with the processor part directly, I had in mind that it could be 2 separated things, like the processor allocation working without the batch allocator for example, but the actual goal is for the batch allocator 😄 (slowly lost track)

I'll try to finish the processor allocation asap ! Will create some PRs this week

miai10:
Ok, what feature gate name should we use? AllocatorBatchesUpdates or ProcessorAllocator? How about SingleBatchAllocator?

miai10:
> > > I was thinking, would that be alright if we move that to a dev feature gate for it to be merged ?
> > 
> > 
> > In head, this would be bundled with your new processor allocation code @lacroixthomas - not sure if you are saying the same thing or not?
> 
> You are totally right, could bundled it with the processor part directly, I had in mind that it could be 2 separated things, like the processor allocation working without the batch allocator for example, but the actual goal is for the batch allocator 😄 (slowly lost track)
> 
> I'll try to finish the processor allocation asap ! Will create some PRs this week

@lacroixthomas, @markmandel: If the 2 parts will be under different feature gates (as they can be used independently) and we know that the HA policy will follow soon, do you think we can merge this in time for the next release? It would allow others to test the feature and give feedback.

markmandel:
I would say - keep it all under the signle new processor + batch allocator feature.

1. It's way easier to test for end users, rather than trying to work out which combination of features to enable and disable
2. It's easier for us to coordinate work - all the new efforts go towards this new processing allocation system -- which also enables the batch processing with HA at the same time. Otherwise it's increases our maintenance burden to keep everything running together.

Longer wait to get it all lined up, but I think it'll pay off in the long run.

lacroixthomas:
FYI: The work to [improve the HA of the allocator](https://github.com/googleforgames/agones/issues/4190) is still in progress, I don't want to give you a date, but it's getting there, hopefully in the next month or two we'll have it all ready behind a dev feature gate (`ProcessorAllocator`) and we'll be able to test with this PR !

miai10:
Thanks for the update! Let me know in advance if posible when I need to update this branch.

miai10:
Hi, @lacroixthomas! How do we proceed with this MR when the move of the Agones repo will happen?

markmandel:
> Hi, @lacroixthomas! How do we proceed with this MR when the move of the Agones repo will happen?

GitHub handles migrations really well - you could even not update your settings, and everything will still work. It does some very nice redirecting.

lacroixthomas:
Hey @miai10 !
To keep you posted, It's getting there, from now on, I will focus on this one - and avoid getting lost on other tasks ! 😄 
This PR got some conflicts, if you have some time to resolve them ? Otherwise I might have access to resolve them, as you prefer

markmandel:
@lacroixthomas does it make more sense for you to take this approach and merge it into your new code? Rather than the other way around?

lacroixthomas:
@markmandel You mean merging the feature gate to use only one ? `processorAllocator` or `AllocatorBatchesUpdates` but not both ?  
In term of changes from this PR, this should be alright as it's part of the allocation flow - kind of separated than the refactor of allocator + processor 
Or maybe I'm missing something ? 🤔

markmandel:
Yeah - taking this approach and pulling it into `processorAllocator` - do a whole new allocation system in there - - this was my original push too in this thread.

Also given how many conflicts there are, and the fact this can't work for multiple allocators - I think it makes far more sense to have it just within the `processorAllocator` feature.

lacroixthomas:
Sounds good 👌🏼 I'm in for that

miai10:
Hi, @lacroixthomas, long time no talk! Tbh, I haven't been following your work on the `processorAllocator` feature. Mark's suggestion is fine with me but let me know if you need help, I can chip in. I'll review the code ance you create an MR.

lacroixthomas:
Hi @miai10 ! 
Been quite some time already, times flies haha 
Hope you're doing well ! 😃 

I'll create the PR soon, will make a seperated PR for the metrics as i'll add other ones for the processor too

Will poke you when they are ready

lacroixthomas:
Just to keep you posted, I created a draft PR about it, I've done some manual tests with minikube with allocations in batch using counter and lists and also direct allocations, looking great so far 🤞🏼 ! 
With 3 replicas of allocators + extensions and 2 processor (one leader and one in standing by) 

Way more manual testing to do though - and probably minor code changes

markmandel:
I'm on a PR cleanup kick - @miai10 @lacroixthomas - can we close this PR now? 😄

lacroixthomas:
Yep all good to close this one, I bring back everything on this one: https://github.com/agones-dev/agones/pull/4559 
(Which still need some love, will focus on review for this week and will come back to it)

markmandel:
inspiration

An inspiration for us all. Mostly @lacroixthomas though 😁

PR #4213, opened 2025-06-30
Fix for TestSetupMetrics_StackdriverOnly_NoPanic

### LINKED ISSUE #4212: Test: TestSetupMetrics_StackdriverOnly_NoPanic fails when running locally
**What happened**:

Running `make test-go` and when running the test within my IDE, I get the following error.

```
=== RUN   TestSetupMetrics_StackdriverOnly_NoPanic
{"error":"error getting zone: Get \"http://169.254.169.254/computeMetadata/v1/instance/zone\": dial tcp 169.254.169.254:80: connect: no route to host","message":"error discovering monitored resource","severity":"warning","source":"metrics","time":"2025-06-29T21:28:41.88767958-07:00"}
{"error":"google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information","message":"Could not register Stackdriver exporter","severity":"fatal","source":"metrics","time":"2025-06-29T21:28:41.916038332-07:00"}
```

What is weird though, is that I don't see any assertions failing or a panic - it's just failing and I'm not sure why. Is there something capturing a panic and not outputting it somewhere? It's very confusing.

Test to pass everywhere.

`make test-go` or run the test through an IDE.

N/A

**Environment**:
- Agones version: dev 
- Kubernetes version (use `kubectl version`): N/A
- Troubleshooting guide log(s): See above.
- Others:

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Logging shouldn't be `Fatal` without a really good reason.

Closes #4212

N/A

### DIFF
--- a/pkg/metrics/exporter.go
+++ b/pkg/metrics/exporter.go
@@ -140,9 +140,10 @@ func SetupMetrics(conf Config, server *httpserver.Server) (healthcheck.Handler, 
 	if conf.Stackdriver {
 		sd, err := RegisterStackdriverExporter(conf.GCPProjectID, conf.StackdriverLabels)
 		if err != nil {
-			logger.WithError(err).Fatal("Could not register Stackdriver exporter")
+			logger.WithError(err).Error("Could not register Stackdriver exporter")
+		} else {
+			closer = func() { sd.Flush() }
 		}
-		closer = func() { sd.Flush() }
 	}
 
 	// Prometheus Metrics
@@ -150,7 +151,7 @@ func SetupMetrics(conf Config, server *httpserver.Server) (healthcheck.Handler, 
 		registry := prom.NewRegistry()
 		metricHandler, err := RegisterPrometheusExporter(registry)
 		if err != nil {
-			logger.WithError(err).Fatal("Could not register Prometheus exporter")
+			logger.WithError(err).Error("Could not register Prometheus exporter")
 		}
 		server.Handle("/metrics", metricHandler)
 		health = healthcheck.NewMetricsHandler(registry, "agones")


markmandel:
FYI: @0xaravindh

0xaravindh:
LGTM

pkg/metrics/exporter.go
```diff
@@ -140,17 +140,18 @@ func SetupMetrics(conf Config, server *httpserver.Server) (healthcheck.Handler,
...
 		}
-		closer = func() { sd.Flush() }
 	}
 
 	// Prometheus Metrics
 	if conf.PrometheusMetrics {
 		registry := prom.NewRegistry()
 		metricHandler, err := RegisterPrometheusExporter(registry)
 		if err != nil {
-			logger.WithError(err).Fatal("Could not register Prometheus exporter")
```
igooch, line 153: Do we want these to not throw fatal error in case the metrics cannot be registered? The SetupMetrics function is called as part of the all the Agones controllers setups. 

If we're just trying to fix a locally failing test the setup for the test may need to change instead.

markmandel: Fatal will crash the entire controller. Is that what you are saying, the Agones controller should enter crashloopbackoff if metrics can't be set?

igooch: Mostly saying that's the current behavior. At some point we made the decision that metrics were important enough to the controller to fail loudly without them. If we're making the decision that metrics are no longer critical to the controller that decision should be called out somewhere.

The issue for this PR is that a test is failing because the test registration doesn't work locally. Seems like we should fix the test so it works with registration or uses a mock registration.

0xaravindh: @igooch @markmandel  Mocked GCE metadata to fix the Stackdriver test #4215. Let me know if it looks good or needs changes.

markmandel: > Mostly saying that's the current behavior. At some point we made the decision that metrics were important enough to the controller to fail loudly without them. If we're making the decision that metrics are no longer critical to the controller that decision should be called out somewhere.

That is a fair point! I'll close this PR 👍🏻

PR #4214, opened 2025-07-02
fix: Adding a retry mechanism in case the addMoreGameServers function call fails.

txuna:
**What type of PR is this?**
/kind bug

### Abstract and Summary
`affected version <= 1.50.0`   
The core of my pull request is that I identified an issue where, if `all GameServers` are deleted and the addMoreGameServers function fails, no further GameServers are created. To resolve this, I added logic to reprocess the GameServerSet using `c.workerqueue.EnqueueAfter`.
> [!NOTE]  
> In the case of partial GameServer deletion (not all GameServers), the GameServer informer’s resync triggers an Update event. This leads to a call to the computeReconciliationAction function, which calculates the number of GameServers to add (numServersToAdd), followed by a call to the addMoreGameServers function.

Without this fix, since no GameServers exist, there are no further GameServer informer updates or resyncs triggered. As a result, the addMoreGameServers function is never called again, and no new GameServers are created.

### POC
Imagine that the Agones Controller and Agones Extension are deployed on top of Kubernetes, and there is one GameServerSet (test-gss) with two GameServers (test-gss-1 and test-gss-2).

Delete the two GameServers (test-gss-1 and test-gss-2) while the Agones Extension is in the TLS certs changing state.
> [!NOTE]  
> Delete all deployed GameServers at once.
```bash
# Agones Extension Pod
http: TLS handshake error from ip:port: EOF
http: TLS handshake error from ip:port: EOF
http: TLS handshake error from ip:port: EOF
[...]
```
```bash
# my cli terminal
kubectl delete pod test-gss-1 test-gss-2 -n default
```

In a normal scenario, the addMoreGameServers function is triggered when a GameServer transitions to the Shutdown state or when its Pod is deleted.

but we can see this error message when invoke `addMoreGameServers`
```json
{
  "error": "error creating gameserver for gameserverset test-gss: Internal error occurred: failed calling webhook \"mutations.agones.dev\": failed to call webhook: Post \"https://agones-controller-service.agones-system.svc:443/mutate?timeout=10s\": tls: failed to verify certificate: x509~~"
}
```
Due to the above error, the addMoreGameServers function is no longer invoked.

### Solution
The solution is simple: if the addMoreGameServers function call fails, it should be reprocessed using `c.workerqueue.EnqueueAfter`.
```go
	if numServersToAdd > 0 {
		if err := c.addMoreGameServers(ctx, gsSet, numServersToAdd); err != nil {
+			c.workerqueue.EnqueueAfter(gsSet, 1*time.Second) // retry!
			loggerForGameServerSet(c.baseLogger, gsSet).WithError(err).Warning("error adding game servers and will retry")
		}
	}

```

### Limitation
A potential drawback is that retries may continue indefinitely until success.

***If you have any suggestions for a better approach, I'd appreciate your comments.***

### DIFF
--- a/pkg/gameserversets/controller.go
+++ b/pkg/gameserversets/controller.go
@@ -356,6 +356,7 @@ func (c *Controller) syncGameServerSet(ctx context.Context, key string) error {
 
 	if numServersToAdd > 0 {
 		if err := c.addMoreGameServers(ctx, gsSet, numServersToAdd); err != nil {
+			c.workerqueue.EnqueueAfter(gsSet, 1*time.Second)
 			loggerForGameServerSet(c.baseLogger, gsSet).WithError(err).Warning("error adding game servers")
 		}
 	}


pkg/gameserversets/controller.go
```diff
@@ -356,6 +356,7 @@ func (c *Controller) syncGameServerSet(ctx context.Context, key string) error {
 
 	if numServersToAdd > 0 {
 		if err := c.addMoreGameServers(ctx, gsSet, numServersToAdd); err != nil {
+			c.workerqueue.EnqueueAfter(gsSet, 1*time.Second)
```
markmandel: 👀 this looks like a good find.  I'm kinda wondering why we didn't return the error here in the first place.

Rather than re-enqueue, the canonical way to re-queue the item, is to return the error after the log, which gets handled by the workerqueue as a retry mechanism.

Since we're in here, we should also do the same at line 365/366 because if the delete fails, we should also retry.

Nice catch!

pkg/gameserversets/controller.go
```diff
@@ -356,6 +356,7 @@ func (c *Controller) syncGameServerSet(ctx context.Context, key string) error {
...
 			loggerForGameServerSet(c.baseLogger, gsSet).WithError(err).Warning("error adding game servers")
```
markmandel:
```suggestion
			loggerForGameServerSet(c.baseLogger, gsSet).WithError(err).Warning("error adding game servers")
			return errors.Wrap(err, "error adding game servers"))
```

markmandel (approved):
Love it.

PR #4218, opened 2025-07-11
Feat: Improve allocator HA with new processor

### LINKED ISSUE #4190: [Feature Proposal] - Improve allocator HA to favor packing sessions into multi sessions servers
The current HA solution for the allocator is implemented with a service and at least 2 replicas of the allocator that process the received gsa at intervals of time.

This approach is not well suited for high allocations rate and high capacity servers because the allocator compete to update the same server CRD thus generating update failures and many retries leading eventually to spreading of the sessions on multiple servers and high allocations time. Even with one session per server, the allocators can compete on the same server.

Using one allocator (especially with batching enabled) with a fine tuned batch wait time gives better results but the HA policy is downgraded to just restarting the pod when needed.

The scenarios we need to support would be:
- one node goes down with the allocator on it
- planned maintenance when the allocator is restarted to be updated
- allocator crashes

The perfect solution would be to find a design where multiple allocators are running, sharing the load when needed.

Possible solutions:
1. master/slave allocators implemented using the readiness check of each pod and a leader election scheme
2. pub/sub instead of channel for consuming the allocations 
3. shared cache and list for servers state between allocators 

**Additional context**

None

**Discussion Link (if any)**
There have been discussions here: https://github.com/googleforgames/agones/pull/4176#issuecomment-2892822957

lacroixthomas:
**What type of PR is this?**
/kind feature

The PR is still in draft, it's "working" but definitively still in WIP
Will add a detailed explanation soon

TODO:
- [ ] Config (with default values)
- [ ] Metrics (might have broke some part)
- [ ] Ensuring the context part are working as expected for graceful shutdown or new leader etc.
- [ ] Cleanup (I did a first pass, but there is definitely some part that need some love, even for the logic)
- [ ] Documentation on processor and feature flag
- [ ] Load tests

Closes #4190 

It's in WIP, but if you have feedbacks on the whole flow don't hesitate 😄


markmandel:
We had a chat on Slack 😄 let's create a dev feature flag, and start to break this up into smaller PRs -- going to be much easier to review 👍🏻

PR #4222, opened 2025-07-20
Feat: Add new binary processor

lacroixthomas:
**What type of PR is this?**
/kind feature

Added scaffolding for the new binary "processor"

The processor only logs and have a default leader election (under dev feature flag ProcessorAllocator

Part of #4190 


markmandel:
Just checking before review - should this be in draft? or is it ready for review?

cmd/processor/main.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
markmandel:
```suggestion
// Copyright 2025 Google LLC All Rights Reserved.
```

cloudbuild.yaml
```diff
@@ -573,6 +573,7 @@ images:
   - ${_REGISTRY}/agones-sdk
   - ${_REGISTRY}/agones-ping
   - ${_REGISTRY}/agones-allocator
+  - ${_REGISTRY}/agones-operator
```
markmandel:
```suggestion
  - ${_REGISTRY}/agones-processor
```

(trying to decide if there is a better name, but it should at least be processor, right?)

cmd/processor/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM gcr.io/distroless/static-debian12:nonroot
```
markmandel, line 15: Apache licence please.

install/helm/agones/templates/processor-deployment.yaml
```diff
@@ -0,0 +1,164 @@
+# Copyright 2019 Google LLC All Rights Reserved.
```
markmandel:
```suggestion
# Copyright 2025 Google LLC All Rights Reserved.
```

markmandel:
```suggestion
# Copyright 2025 Google LLC All Rights Reserved.
```

lacroixthomas:
> Just checking before review - should this be in draft? or is it ready for review?

Yep, no need to review, still in draft, still need to do some update and test around the deployment.yaml, adding some default values and also checking the service account, either using the one from the controller or adding a new one, it will need permissions for the leases

markmandel:
Sweet - lemme know if you want to kick it through full CI.

lacroixthomas:
Perfect, will let you know, might be all good next week for review 
Thanks for the quick review though, will update that 😄

lacroixthomas:
@markmandel 
That should be good for review, there is some things as the health check part / metrics, that I would prefer to setup when integrating the allocator inside this agones-processor (from a following PR)

This PR is only to setup the scaffolding of this new binary (with leader election)

cmd/processor/main.go
```diff
@@ -0,0 +1,192 @@
...
+		cancelCtx()
+		time.Sleep(1 * time.Second)
+		os.Exit(0)
+	})
+
+	whenLeader(ctx, cancelCtx, logger, conf.LeaderElection, kubeClient, conf.PodNamespace, func(ctx context.Context) {
+		logger.Info("Starting processor work as leader")
+
+		// Simulate processor work (to ensure the leader is working)
+		ticker := time.NewTicker(5 * time.Second)
```
miai10, line 142: Do you think this value will affect the down time of not having a leader? If so, maybe we should have it configurable?

cmd/processor/main.go
```diff
@@ -0,0 +1,192 @@
...
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock: lock,
+		// IMPORTANT: you MUST ensure that any code you have that
+		// is protected by the lease must terminate **before**
+		// you call cancel. Otherwise, you could have a background
+		// loop still running and another process could
+		// get elected before your background loop finished, violating
+		// the stated goal of the lease.
+		ReleaseOnCancel: true,
+		LeaseDuration:   15 * time.Second,
```
miai10: Maybe we should have them configurable to be able to fine tune them in our tests

cmd/processor/main.go
```diff
@@ -0,0 +1,192 @@
...
+				logger.Info("Processor is active as leader")
+			}
+		}
+	})
+
+	logger.Info("Processor exited gracefully.")
+}
+
+func whenLeader(ctx context.Context, cancel context.CancelFunc, logger *logrus.Entry, doLeaderElection bool, kubeClient *kubernetes.Clientset, namespace string, start func(_ context.Context)) {
+	if !doLeaderElection {
```
miai10: So there will be an actual use case where the all processors process requests (without the leader election feature enabled)?

cmd/processor/main.go
lacroixthomas: This will not impact it, as it's temporary, it's just to simulate the processor work until this part is implemented

cmd/processor/main.go
lacroixthomas: There will still be only one processor doing the job, if someone wants only 1 replicas of the processor (without standby processor), It's probably not worth setting the leader election for it

cmd/processor/main.go
lacroixthomas: Will add that

install/helm/agones/values.yaml
```diff
@@ -282,6 +286,32 @@ agones:
     totalRemoteAllocationTimeout: 30s
     allocationBatchWaitTime: 500ms
     topologySpreadConstraints: []
+  processor:
```
markmandel: I am wondering if this should be a child of `agones.allocator` ? Since the processor is only actually part of allocation.

So it would be `agones.allocator.processor` - then it's contextual, and processor doesn't really make sense on it's own. WDYT?

install/helm/agones/templates/processor-deployment.yaml
```diff
@@ -0,0 +1,178 @@
+# Copyright 2025 Google LLC All Rights Reserved.
```
markmandel: This file could just be `processor.yaml` - "deployment" is a bit redundant.

cmd/processor/main.go
miai10: If we want to support only the case where with one replica there is no need for leader election, maybe we can just detect that from the replica count and not have a config switch for it.

miai10: You already did that, sorry :P

install/helm/agones/values.yaml
lacroixthomas: Yep totally make sense, will update that

lacroixthomas:
Rolled back the fix of the pipeline to keep it on a separate task

cmd/processor/main.go
```diff
@@ -0,0 +1,208 @@
...
+		logger.Info("Pod shutdown has been requested, failing readiness check")
+		cancelCtx()
+		time.Sleep(1 * time.Second)
+		os.Exit(0)
+	})
+
+	whenLeader(ctx, cancelCtx, logger, conf, kubeClient, func(ctx context.Context) {
+		logger.Info("Starting processor work as leader")
+
+		// Simulate processor work (to ensure the leader is working)
```
markmandel, line 140: (non blocking suggestion)

```suggestion
		// Simulate processor work (to ensure the leader is working)
		// TODO: implement processor work
```

Just to make it extra clear that more work is coming.

cmd/processor/main.go
```diff
@@ -0,0 +1,208 @@
...
+
+	go func() {
+		healthserver.Handle("/", health)
+		_ = healthserver.Run(context.Background(), 0)
+	}()
+
+	signals.NewSigTermHandler(func() {
+		logger.Info("Pod shutdown has been requested, failing readiness check")
+		cancelCtx()
+		time.Sleep(1 * time.Second)
```
markmandel: Do we need a minute sleep timer before shutdown? Is this to handle any remaining processing items that are mid flight? (do we care? maybe we do).

If so, should it be configurable like it is in the allocator?

install/helm/agones/templates/processor.yaml
```diff
@@ -0,0 +1,178 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{- $featureGates := include "agones.featureGates" . | fromYaml }}
+{{- if $featureGates.ProcessorAllocator }}
+apiVersion: apps/v1
```
markmandel, line 17: I'm assuming documentation comes last? (i.e. we don't want to include the Helm config docs now -- we'll wait until everything is all worked out and do it all at once?) Not against the idea, just wanted to check.

markmandel (commented):
Looking good - just some minor questions.

cmd/processor/main.go
lacroixthomas: I should probably get rid of it from this PR as it's just the scaffolding, it will indeed be configurable on the implementation of the allocator on it 😄

install/helm/agones/templates/processor.yaml
lacroixthomas: Yep exactly, I'll write the documentation at the end, I've did some diagrams at the issue level in the comments part: https://github.com/googleforgames/agones/issues/4190 

I could probably start writing the doc in another draft PR while implementing the allocator / processor / extension, this would give more feedback and a better understanding of the big picture, also ensure we don't miss anything ? WDYT ? 

Even though the doc would only be merged once moved to Alpha I guess ?

cmd/processor/main.go
lacroixthomas: Updated: [a9b200f](https://github.com/googleforgames/agones/pull/4222/commits/a9b200f583e14c0fc423b3c2b438e49de6c07e17)

cmd/processor/main.go
lacroixthomas: Updated: [a9b200f](https://github.com/googleforgames/agones/pull/4222/commits/a9b200f583e14c0fc423b3c2b438e49de6c07e17)

install/helm/agones/templates/processor.yaml
markmandel: I'm not super fussed on the documentation part - the config files are enough for the moment, and if we find something along the way to change, you can just do one set of docs at the end rather than constantly updating 👍🏻

markmandel (approved):

PR #4227, opened 2025-07-24
feat: add processor proto 

lacroixthomas:
**What type of PR is this?**
/kind feature

Add processor proto + update makefile

Part of #4190 


lacroixthomas:
This is still in work in progress, quite some stuff to fix / cleanup / doc

miai10:
If you have some time, you could create a sequence diagram with the messages exchange to improve the documentation of this new part of the project

lacroixthomas:
> If you have some time, you could create a sequence diagram with the messages exchange to improve the documentation of this new part of the project

I was thinking to write the doc at the end: https://github.com/googleforgames/agones/issues/4190#issuecomment-3090726203 
But I can link a mermaid diagram as a comment here if wanted ?

miai10:
> > If you have some time, you could create a sequence diagram with the messages exchange to improve the documentation of this new part of the project
> 
> I was thinking to write the doc at the end: [#4190 (comment)](https://github.com/googleforgames/agones/issues/4190#issuecomment-3090726203) But I can link a mermaid diagram as a comment here if wanted ?

I was thinking that any piece of design doc is valuable to be reviewed right now. 

For me it is a bit unclear how the pull batches system will work. For example how will the leader processor know from which allocator address to pull from? If one allocator is replaced, how is it autodetected?

lacroixthomas:
Understood, this PR is mainly for the pull effect streaming with gRPC for the proto, but all the logic around how it would communicate would be on another PR when ill implement a client and server for it

If you're fine with it, I can add more info / diagram on the issue direclty until I start the task for the doc and the implementation of the client server ? 

I added more info on the original issue: https://github.com/googleforgames/agones/issues/4190#issuecomment-3132610583

markmandel:
Shouldn't be affecting `submit-upgrade-test-cloud-build` - it's just being flaky as per usual.

proto/allocation/processor.proto
```diff
@@ -0,0 +1,65 @@
...
+  // The requests is a repeated field of AllocationRequest messages.
+  repeated allocation.AllocationRequest requests = 1;
+}
+
+// A BatchResponse is used to send a response to a batch of allocation requests.
+message BatchResponse {
+  // The responses is a repeated field of AllocationResponse messages.
+  repeated allocation.AllocationResponse responses = 1;
+
+  // The errors is a repeated field of error messages.
```
markmandel, line 63: (Non blocking question): I assume errors are index based? so an error at errors[4] aligns with reposnses[4], and an empty error string is no error?

markmandel (commented):
Looks good, just a non blocking question.

proto/allocation/processor.proto
lacroixthomas: Yep, exactly

PR #4242, opened 2025-08-12
fix: Template image for Helm Cleanup job due to bitnami updates

DaThumpingRabbit:
**What type of PR is this?**
/kind hotfix

Following this update https://github.com/bitnami/charts/issues/35164, it will not be possible to keep using the kubectl image from bitnami for all users
This PR allows to choose what image to use to run the post-delete cleanup job and default on the best alternative I could find at the moment [rancher/kubectl](https://hub.docker.com/r/rancher/kubectl)

Please let me know if bitnami should stay the default registry based on your guidelines

### DIFF
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -36,7 +36,7 @@ agones:
     cleanupJobImage:
       registry: rancher
       repository: kubectl
-      tag: 1.32.7
+      tag: v1.32.7
   serviceaccount:
     allocator:
       name: agones-allocator


markmandel:
🤔 seems like a reasonable change - just wondering why we want to make it configurable? Might as well just switch to the rancher version and call it a day.

Just a question of ongoign maintenance and documentation.

markmandel:
Just noted, there is an official kubectl image: https://www.reddit.com/r/kubernetes/s/1rDLvH7rtR

> We provide an official kubectl image these days:
registry.k8s.io/kubectl:v1.33.3

> Fair warning for production dependency on this host: https://registry.k8s.io#stability

> TLDR this is volunteer operated and you're not paying us for an SLA. Mirror if you need uptime guaranteed, docs provided for doing that.

> Also, kubectl is a single static go binary, so making an image for it is pretty trivial.

DaThumpingRabbit:
> Just noted, there is an official kubectl image: https://www.reddit.com/r/kubernetes/s/1rDLvH7rtR

I saw the image you mentioned but was a bit afraid about the stability warning  
I have never used it so I don't know if it will be safe enough



> 🤔 seems like a reasonable change - just wondering why we want to make it configurable? Might as well just switch to the rancher version and call it a day.
> 
> Just a question of ongoign maintenance and documentation.

About this, I thought that maybe if some customers are paying the new bitnami secure package, they might want to switch back to bitnami  
However if that makes more sense for you to just switch it, I am fine with that as well

DaThumpingRabbit:
@markmandel I have just made some tests and I don't think that this new image will be a good fit anyway since it does not come with a shell and won't work in that scenario

Only alternative to bitnami I found for now that comes with a shell is [alpine/k8s](https://hub.docker.com/r/alpine/k8s)  
It has way more things than is needed but did not find a better option for now

Please let me know if this would be a good alternative for this

markmandel:
Apologies - this totally fell through the cracks.

I did some pocking around - Looks like https://hub.docker.com/r/alpine/kubectl has an `sh` shell which is all we need, and it's nice and small.

Sound good?

markmandel:
Hope you don't mind - I ended up doing what I suggested in #4268

markmandel:
Closing, since I took care of this in https://github.com/googleforgames/agones/pull/4268

Thanks for the heads up though!

DaThumpingRabbit:
Yes sorry I also completely forgot about this  
This is perfect for me, thanks for handling it 👍

PR #4244, opened 2025-08-13
Update Helm option for spec.strategy.type for controller and extensio…

### LINKED ISSUE #4198: [Feature Proposal] Add Helm option for spec.strategy.type for controller and extensions-deployment

The current [Agones upgrading documentation](https://1-48-0.agones.dev/site/docs/installation/upgrading/) has the caveat:

```
Regardless of the type of installation, there will be a brief ~20-30 second period during upgrade when the controller service switches to the new controller endpoint that service is unable to connect to the new controller. The SDK servers are still functional during this time, however the controller will not be able to receive requests such as creating new game servers. Be sure to include retry logic with back-off in your application logic to account for the error Internal error occurred: failed calling webhook "mutations.agones.dev": failed to call webhook: Post "[https://agones-controller-service.agones-system.svc:443/mutate?timeout=10s](https://agones-controller-service.agones-system.svc/mutate?timeout=10s)": no endpoints available for service "agones-controller-service"" during this time.
```

This is due to the `Recreate` strategy for the controller and extensions-deployment.

https://github.com/googleforgames/agones/blob/eb69e7f6c92ffed70e8e69b66436986949666df5/install/helm/agones/templates/controller.yaml#L38-L39
https://github.com/googleforgames/agones/blob/eb69e7f6c92ffed70e8e69b66436986949666df5/install/helm/agones/templates/extensions-deployment.yaml#L34-L35

By the [Kubernetes definition](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/?content_ref=all+existing+pods+are+killed+before+new+ones+are+created+when+spec+strategy+type+recreate#recreate-deployment), `All existing Pods are killed before new ones are created when .spec.strategy.type==Recreate.`. Since all pods are killed, there is the brief ~20-30 second period where there are no endpoints, thus the webhook (used for creating or allocating new game servers) temporarily fails.

Add a flag, similar to what we have for the [`agones.allocator.updateStrategy`](https://1-48-0.agones.dev/site/docs/installation/install-agones/helm/)

https://github.com/googleforgames/agones/blob/eb69e7f6c92ffed70e8e69b66436986949666df5/install/helm/agones/templates/service/allocation.yaml#L154-L157

so that the user can specify the `spec.strategy.type` for either `Recreate` or `RollingUpdate`.

In the documentation we should specify that `RollingUpdate` for these controllers should not be used with the Helm generated TLS certificates (more details below), and only with cert-manager or self-signed certificates that are not rotated during upgrade.

When switching the controller and extensions from "Recreate" to "RollingUpdate" the "no endpoints available for service" error no longer appears. However instead there is the error:

```
Internal error occurred: failed calling webhook "mutations.agones.dev": failed to call webhook: Post "https://agones-controller-service.agones-system.svc:443/mutate?timeout=10s": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "admission-controller-ca")
```

Based on https://github.com/helm/helm/issues/10731, this is expected behavior from Helm "until the pods refresh the certificate, the kubernetes apiserver will get a TLS error calling the admissions web hook." This is for the _Helm generated TLS certificates_. If the user uses `cert-manager` or their own certificates, and does not rotate the certificates at the same time as the upgrade, this should not be an issue. This means that `RollingUpdate` should not be hardcoded, and should instead be an updatable field.

**Additional context**

**Discussion Link (if any)**

Sivasankaran25:
**What type of PR is this?**

> 
/kind feature

Closes #4198

### DIFF
--- a/install/helm/agones/values.schema.json
+++ b/install/helm/agones/values.schema.json
@@ -484,6 +484,9 @@
             "tlsKey": {
               "type": "string"
             },
+            "updateStrategy": {
+              "type": "object"
+            },
             "pdb": {
               "type": "object",
               "properties": {
@@ -1257,6 +1260,9 @@
             },
             "allocationBatchWaitTime": {
               "type": "string"
+            },
+            "updateStrategy": {
+              "type": "object"
             },
             "pdb": {
               "type": "object",


install/helm/agones/values.yaml
```diff
@@ -102,6 +102,7 @@ agones:
       timeoutSeconds: 1
     allocationBatchWaitTime: 500ms
     replicas: 2
+    updateStrategy: {}
```
igooch, line 105: Update the `agones/install/helm/agones
/values.schema.json` as well, similar to how it's currently done for ping and allocator. https://github.com/googleforgames/agones/blob/c355028b6106a0d50fbab1c7cadcfd7059c80776/install/helm/agones/values.schema.json#L1081-L1083

Sivasankaran25: @igooch  Changes is Updated

PR #4252, opened 2025-08-18
Updates the upgrade test to print any fatal error messages to the job pod termination log

igooch:
**What type of PR is this?**

/kind cleanup

The upgrade main.go is refactored to pass back any fatal errors, cancel context, and cleanup resources on a fatal error. This way we can see exactly what Fatal error caused the test to fail. This will not include all failure scenarios (i.e. our recent failure scenario of a finalizer not being removed from the ping service), but will still make it easier to debug.

Additional wait timeouts are added to prevent the test from hanging.

NA

### DIFF
--- a/build/e2e_upgrade_test.sh
+++ b/build/e2e_upgrade_test.sh
@@ -34,23 +34,37 @@ print_failure_logs() {
     local testClusterLocation=$2
     echo "ERROR: Upgrade test failed on cluster: $testCluster"
     gcloud container clusters get-credentials "$testCluster" --region="$testClusterLocation" --project="$PROJECT_ID"
-    job_pods=$(kubectl get pods -l job-name=upgrade-test-runner -o jsonpath="{.items[*].metadata.name}")
-    if [[ -z "$job_pods" ]]; then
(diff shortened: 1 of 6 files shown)


build/e2e_upgrade_test.sh
```diff
@@ -34,23 +34,37 @@ print_failure_logs() {
     local testClusterLocation=$2
     echo "ERROR: Upgrade test failed on cluster: $testCluster"
     gcloud container clusters get-credentials "$testCluster" --region="$testClusterLocation" --project="$PROJECT_ID"
-    job_pods=$(kubectl get pods -l job-name=upgrade-test-runner -o jsonpath="{.items[*].metadata.name}")
-    if [[ -z "$job_pods" ]]; then
-        echo "No pods found for job upgrade-test-runner. They might have failed to schedule or were deleted."
+
```
markmandel, line 37: (non blocking meta) - this bash script is at a point where it probably should be a go script 😄

markmandel (approved):
I trust your knowledge of this code way more than I trust mine -- I assume this is good to go 😄

PR #4256, opened 2025-08-25
Fix: Cleanup finalizer for k8s (e2e pipeline)

lacroixthomas:
**What type of PR is this?**
/kind bug

Cleanup the finalizer from the services before installing patched: https://github.com/kubernetes/cloud-provider-gcp/pull/887 and that the cluster used for e2e is upgraded 

Similar issue also happens on 1.32: https://console.cloud.google.com/cloud-build/builds/7146c171-ea46-4958-bf19-32083d84a50d?project=agones-images *

### DIFF
--- a/build/e2e-image/e2e.sh
+++ b/build/e2e-image/e2e.sh
@@ -18,21 +18,13 @@ FEATURES=$1
 CLOUD_PRODUCT=$2
 REGISTRY=$3
 
-K8S_MINOR=$(kubectl version -o json | jq -r '.serverVersion.minor' | sed 's/+//')
-
 echo $FEATURES
 echo $REGISTRY
-echo $K8S_MINOR
 set -e
 
-# Ensure we cleanup the finalizers from k8s 1.33 until they patch it
-if [ "$K8S_MINOR" = "33" ]; then
-    echo "GKE 1.33 detected: cleaning up stuck service finalizers"
-
-    kubectl get svc -n agones-system -o json \
-    | jq -r '.items[] | select(.metadata.finalizers | length > 0) | .metadata.name' \
-    | xargs -r -I {} kubectl patch svc {} -n agones-system -p '{"metadata":{"finalizers":null}}' --type=merge
-fi
+kubectl get svc -n agones-system -o json \
+| jq -r '.items[] | select(.metadata.finalizers | length > 0) | .metadata.name' \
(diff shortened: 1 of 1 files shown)


lacroixthomas:
Seems that the issue also happens on 1.32: https://console.cloud.google.com/cloud-build/builds/7146c171-ea46-4958-bf19-32083d84a50d?project=agones-images 

I've updated it to ensure it always cleanup, not only for 1.33, that might be something else for the 1.32, maybe us not uninstalling it properly or similar

lacroixthomas:
I think that it wont be enough, it seems that it will timeout again :(

build/e2e-image/e2e.sh
```diff
@@ -21,6 +21,11 @@ REGISTRY=$3
 echo $FEATURES
 echo $REGISTRY
 set -e
+
+kubectl get svc -n agones-system -o json \
```
igooch, line 25: The finalizer issue is with the Agones upgrades tests specifically, as these uninstall Agones after each run. The e2e simply re-install Agones over an existing revision, so the deletion issue shouldn't happen in those clusters. It's not clear at the moment what's causing the e2e test failures. @vicentefb is investigating.

For instance on the gke-standard-e2e-1-33 cluster:

```
$ helm list -A
NAME  	NAMESPACE    	REVISION	UPDATED                                	STATUS  	CHART            	APP VERSION
agones	agones-system	14      	2025-08-26 23:10:26.196217778 +0000 UTC	deployed	agones-1.52.0-dev	1.52.0-dev
$ bash print_service_age.sh
Checking service age in namespace: agones-system

Service Name                        | Age
----------------------------------- | ---
agones-allocator                    | 1d4h30m
agones-allocator-metrics-service    | 1d4h30m
agones-controller-metrics-service   | 1d4h30m
agones-controller-service           | 1d4h30m
agones-extensions-metrics-service   | 1d4h30m
agones-ping-http-service            | 1d4h30m
agones-ping-udp-service             | 1d4h30m
```

markmandel: Weird - it totally passed on my other PR. It was the upgrade test that stalled out.

lacroixthomas: Will close this draft PR, looks like it's back to normal ! 🙂

PR #4263, opened 2025-09-02
Feat: implement the processor server

lacroixthomas:
**What type of PR is this?**
/kind feature

Implementation of the processor server behind dev feature flag

TODO (some might be on another MR):
- [ ] New Metrics ? (Other PR)
- [ ] TLS ? Is it needed ? (Other PR)
- [x] Unit tests

Part of #4190 

Opening the draft for early feedbacks

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -73,7 +73,7 @@ BETA_FEATURE_GATES ?= "CountsAndLists=true&GKEAutopilotExtendedDurationPods=true
 
 
 # Enable all alpha feature gates. Keep in sync with `false` (alpha) entries in pkg/util/runtime/features.go:featureDefaults
-ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&WasmAutoscaler=true&Example=true&ProcessorAllocator=true"
+ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&WasmAutoscaler=true&Example=true"
 
 # Build with Windows support
 WITH_WINDOWS=1
--- a/build/Makefile
+++ b/build/Makefile
@@ -73,7 +73,7 @@ BETA_FEATURE_GATES ?= "CountsAndLists=true&GKEAutopilotExtendedDurationPods=true
 
 
 # Enable all alpha feature gates. Keep in sync with `false` (alpha) entries in pkg/util/runtime/features.go:featureDefaults
-ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&WasmAutoscaler=true&Example=true"
+ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&WasmAutoscaler=true&Example=true&ProcessorAllocator=true"
 
 # Build with Windows support
 WITH_WINDOWS=1
--- a/build/Makefile
+++ b/build/Makefile
@@ -73,7 +73,7 @@ BETA_FEATURE_GATES ?= "CountsAndLists=true&GKEAutopilotExtendedDurationPods=true
 
 
 # Enable all alpha feature gates. Keep in sync with `false` (alpha) entries in pkg/util/runtime/features.go:featureDefaults
-ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&WasmAutoscaler=true&Example=true"
+ALPHA_FEATURE_GATES ?= "PlayerAllocationFilter=true&PlayerTracking=true&WasmAutoscaler=true&Example=true&ProcessorAllocator=true"
 
 # Build with Windows support
 WITH_WINDOWS=1
@@ -799,7 +799,7 @@ build-processor-image-arm64: $(ensure-build-image) build-processor-binary create
 
 # Build the debug image for the processor service
 build-processor-debug-image: $(ensure-build-image) build-processor-debug-binary build-licenses build-required-src-dist
-	docker build $(agones_path)/cmd/processor/ --file $(agones_path)/cmd/processor/Dockerfile.debug --tag=$(processor_tag) $(DOCKER_BUILD_ARGS)
+	docker build $(agones_path)/cmd/processor/ --file $(agones_path)/cmd/processor/Dockerfile.debug --tag=$(processor_amd64_tag) $(DOCKER_BUILD_ARGS)
 
 # Pushes up the processor image
 push-processor-image: push-processor-image-amd64
--- a/cmd/processor/handler.go
+++ b/cmd/processor/handler.go
@@ -1,0 +1,347 @@
+// Copyright 2026 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Processor
+package main
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"agones.dev/agones/pkg/allocation/converters"
+	allocationpb "agones.dev/agones/pkg/allocation/go"
+	allocationv1 "agones.dev/agones/pkg/apis/allocation/v1"
+	"agones.dev/agones/pkg/client/clientset/versioned"
+	"agones.dev/agones/pkg/client/informers/externalversions"
+	"agones.dev/agones/pkg/gameserverallocations"
+	"agones.dev/agones/pkg/gameservers"
+
+	"github.com/heptiolabs/healthcheck"
+	"github.com/sirupsen/logrus"
+	"go.opencensus.io/plugin/ocgrpc"
+	rpcstatus "google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/status"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+)
+
+// allocationResult represents the result of an allocation attempt
+type allocationResult struct {
+	response *allocationpb.AllocationResponse
+	error    *rpcstatus.Status
+}
+
+// processorHandler represents the gRPC server for processing allocation requests
+type processorHandler struct {
+	allocationpb.UnimplementedProcessorServer
+	ctx                       context.Context
+	cancel                    context.CancelFunc
+	mu                        sync.RWMutex
+	allocator                 *gameserverallocations.Allocator
+	clients                   map[string]allocationpb.Processor_StreamBatchesServer
+	grpcUnallocatedStatusCode codes.Code
+	pullInterval              time.Duration
+}
+
+// newServiceHandler creates a new instance of processorHandler
+func newServiceHandler(ctx context.Context, kubeClient kubernetes.Interface, agonesClient versioned.Interface,
+	health healthcheck.Handler, config processorConfig, grpcUnallocatedStatusCode codes.Code) *processorHandler {
+	defaultResync := 30 * time.Second
+	agonesInformerFactory := externalversions.NewSharedInformerFactory(agonesClient, defaultResync)
+	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, defaultResync)
+	gsCounter := gameservers.NewPerNodeCounter(kubeInformerFactory, agonesInformerFactory)
+
+	allocator := gameserverallocations.NewAllocator(
+		agonesInformerFactory.Multicluster().V1().GameServerAllocationPolicies(),
+		kubeInformerFactory.Core().V1().Secrets(),
+		agonesClient.AgonesV1(),
+		kubeClient,
+		gameserverallocations.NewAllocationCache(agonesInformerFactory.Agones().V1().GameServers(), gsCounter, health),
+		config.RemoteAllocationTimeout,
+		config.TotalRemoteAllocationTimeout,
+		config.AllocationBatchWaitTime)
+
+	batchCtx, cancel := context.WithCancel(ctx)
+	h := processorHandler{
+		allocator:                 allocator,
+		ctx:                       batchCtx,
+		cancel:                    cancel,
+		grpcUnallocatedStatusCode: grpcUnallocatedStatusCode,
+		pullInterval:              config.PullInterval,
+	}
+
+	kubeInformerFactory.Start(ctx.Done())
+	agonesInformerFactory.Start(ctx.Done())
+
+	if err := allocator.Run(ctx); err != nil {
+		logger.WithError(err).Fatal("Starting allocator failed.")
+	}
+
+	return &h
+}
+
+// StreamBatches handles a bidirectional stream for batch allocation requests from a client
+// Registers the client, processes incoming batches, and sends responses
+func (h *processorHandler) StreamBatches(stream allocationpb.Processor_StreamBatchesServer) error {
+	var clientID string
+
+	// Wait for first message to get clientID
+	msg, err := stream.Recv()
+	if err != nil {
+		logger.WithError(err).Debug("Stream receive error on connect")
+		return err
+	}
+
+	clientID = msg.GetClientId()
+	if clientID == "" {
+		logger.Warn("Received empty clientID, closing stream")
+		return nil
+	}
+
+	h.addClient(clientID, stream)
+	defer h.removeClient(clientID)
+	logger.WithField("clientID", clientID).Debug("Client registered")
+
+	// Main loop: handle incoming messages
+	for {
+		msg, err := stream.Recv()
+		if err != nil {
+			logger.WithError(err).Debug("Stream receive error")
+			return err
+		}
+
+		payload := msg.GetPayload()
+		if payload == nil {
+			logger.WithField("clientID", clientID).Warn("Received message with nil payload")
+			continue
+		}
+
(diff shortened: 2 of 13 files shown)


install/helm/agones/templates/processor.yaml
```diff
@@ -113,13 +113,37 @@ spec:
...
+        - name: GCP_PROJECT_ID
+          value: {{ .Values.agones.metrics.stackdriverProjectID | quote }}
+        - name: STACKDRIVER_LABELS
+          value: {{ .Values.agones.metrics.stackdriverLabels | quote }}
+        - name: REMOTE_ALLOCATION_TIMEOUT
+          value: {{ .Values.agones.allocator.remoteAllocationTimeout | quote }}
+        - name: TOTAL_REMOTE_ALLOCATION_TIMEOUT
+          value: {{ .Values.agones.allocator.totalRemoteAllocationTimeout | quote }}
+        - name: ALLOCATION_BATCH_WAIT_TIME
+          value: {{ .Values.agones.allocator.allocationBatchWaitTime | quote }}
```
lacroixthomas: I'm not sure what would be the best, maybe having that under allocator.processor from the values ? Or keeping it under allocator directly ? 🤔  
Under allocator.processor might help once out of feature gate (to cleanup the allocator parts)

lacroixthomas: Might gonna duplicate it actually, because the `ALLOCATION_BATCH_WAIT_TIME` will be way less (around 50ms compared to the current 500ms) as the processor will pull every 200ms and process it every 50 ms (the time it could take between the pull of all "clients" and the time it receives it all)

lacroixthomas: I've only moved the allocationBatchWaitTime to allocator.processor. I'll check for the others, I'd say let's be consistent and move it all, but shall see

lacroixthomas:
I'll come back to this task in the next days / week, didn't forget about it !

build/Makefile
```diff
@@ -798,7 +798,7 @@ build-processor-image-arm64: $(ensure-build-image) build-processor-binary create
 
 # Build the debug image for the processor service
 build-processor-debug-image: $(ensure-build-image) build-processor-debug-binary build-licenses build-required-src-dist
-	docker build $(agones_path)/cmd/processor/ --file $(agones_path)/cmd/processor/Dockerfile.debug --tag=$(processor_tag) $(DOCKER_BUILD_ARGS)
+	docker build $(agones_path)/cmd/processor/ --file $(agones_path)/cmd/processor/Dockerfile.debug --tag=$(processor_amd64_tag) $(DOCKER_BUILD_ARGS)
```
lacroixthomas, line 802: Sneaky fix for the debug mode around the processor*

cmd/processor/handler.go
```diff
@@ -0,0 +1,358 @@
+// Copyright 2025 Google LLC All Rights Reserved.
```
lacroixthomas: TODO: update all the copyrights added from this PR to 2026 😄

lacroixthomas:
I'm waiting for the e2e to finish and I'll rollback the feature gate from alpha to dev, it's just to confirm that the e2e still works with the new processor (which seems to be alright now !) 

It's still in draft because I want to write some unit tests around the handler.go from the processor server, only added you as reviewers, if you want to have a look before the unit tests are done 😄 

I'm gonna create an other PR to add new metrics

lacroixthomas:
Note: Ensure that this issue is also part of this new design: https://github.com/googleforgames/agones/pull/4415 
About a request context cancelled, to not add it to the list to process or ensuring it wont be processed - more info on the issue

lacroixthomas:
I'll create another PRs for TLS, new metrics and doc

markmandel:
So good for review now?

lacroixthomas:
Yep all good 👍🏼

install/helm/agones/values.yaml
```diff
@@ -292,6 +292,8 @@ agones:
     processor:
       replicas: 2
       maxBatchSize: 100
+      pullInterval: 200ms
+      allocationBatchWaitTime: 50ms
```
markmandel, line 296: Just noting we're going smaller than the 500ms which was the original.

Not necessarily a bad thing, just noticing that we're doing that. Do we have a reason as to why? Guessng it's a combo deal with the pullInterval as well?

pkg/processor/client.go
markmandel, line 245: This is old, but got picked up.

The select/default pattern around stream.Recv() causes a tight CPU spin when there are no messages:
```go
for {
    select {
    case <-ctx.Done():
        return ctx.Err()
    default:
        msg, err := stream.Recv()  // blocks, but the select/default makes this a spin loop
```

stream.Recv() is a blocking call, so the default branch is always taken immediately, and the goroutine never yields to the scheduler in a meaningful way. This will peg a CPU core. The fix is to remove the select/default entirely and just call stream.Recv() directly in the loop, handling ctx.Done() via the stream's context:

```
for {
    msg, err := stream.Recv()
    if err != nil {
        if ctx.Err() != nil {
            return ctx.Err()
        }
        return errors.Wrap(err, "stream recv error")
    }
    // handle msg...
}
```

pkg/gameserverallocations/controller.go
```diff
@@ -266,3 +261,42 @@ func (c *Extensions) serialisation(r *http.Request, w http.ResponseWriter, obj k
...
+// convertProcessorError handles processor client errors and converts them to appropriate responses
+func (c *Extensions) convertProcessorError(err error, gsa *allocationv1.GameServerAllocation) (k8sruntime.Object, int) {
+	if st, ok := status.FromError(err); ok {
+		switch st.Code() {
+		case codes.ResourceExhausted:
+			gsa.Status.State = allocationv1.GameServerAllocationUnAllocated
+			return gsa, http.StatusCreated
+		case codes.Aborted:
+			gsa.Status.State = allocationv1.GameServerAllocationContention
+			return gsa, http.StatusCreated
```
markmandel, line 274: This looks wrong - HTTP 201 - Created for allocation unallocated and contention?

Is this what we do now?

pkg/processor/client.go
```diff
@@ -297,9 +320,19 @@ func (p *client) sendBatch(stream allocationpb.Processor_StreamBatchesClient, ba
 	if err := stream.Send(batchMsg); err != nil {
 		p.logger.WithError(err).Error("Failed to send batch")
 
-		// Re-add the request to the hot batch and pendingRequests for the next pull
+		// Re-add retryable requests to the hot batch
 		p.batchMutex.Lock()
```
markmandel, line 320: Noticing this map has a lot of locks and unlocks - just triple checking we're good with that and being careful we don't read a state and then have something else come along and write a state while it's running that makes things go all skewiff.

cmd/processor/handler.go
```diff
@@ -0,0 +1,354 @@
...
+type allocationResult struct {
+	response *allocationpb.AllocationResponse
+	error    *rpcstatus.Status
+}
+
+// processorHandler represents the gRPC server for processing allocation requests
+type processorHandler struct {
+	allocationpb.UnimplementedProcessorServer
+	ctx                       context.Context
+	cancel                    context.CancelFunc
```
markmandel: What's the `cancelFunc` doing - we don't call it 😄 or is this a placeholder "UnimplementedProcessorServer" ?

pkg/processor/client.go
markmandel: Also old - but magic number. Also why 20? 😄

cmd/processor/handler.go
```diff
@@ -0,0 +1,354 @@
...
+	var wg sync.WaitGroup
+	results := make([]allocationResult, len(requestWrappers))
+
+	for i, reqWrapper := range requestWrappers {
+		wg.Add(1)
+		go func(index int, requestWrapper *allocationpb.RequestWrapper) {
+			defer wg.Done()
+			results[index] = h.processAllocation(ctx, requestWrapper.Request)
+		}(i, reqWrapper)
+	}
```
markmandel:
```suggestion
for i, reqWrapper := range requestWrappers {
    wg.Go(func() {
        results[i] = h.processAllocation(ctx, reqWrapper.Request)
    })
}
```

New!
* https://pkg.go.dev/sync#WaitGroup.Go
* https://appliedgo.net/spotlight/go-1.25-waitgroup-go/

pkg/processor/client.go
```diff
@@ -528,3 +561,23 @@ func (p *client) registerClient(stream allocationpb.Processor_StreamBatchesClien
...
+
+	drainErr := status.Errorf(codes.Unavailable, "processor client shutting down")
+	for id, req := range p.requestIDMapping {
+		select {
+		case req.error <- drainErr:
+		default:
+		}
+		delete(p.requestIDMapping, id)
+	}
+
```
markmandel: Nit:
```suggestion
	}

clear(p.requestIDMapping)
```

Less operations

markmandel (commented):
Big PR!

I reviewed what I could - but mostly trusting that you remember the design in question 😄 I went over it again, and i think I remember.

Found some old things along the way as well I flagged as well.

markmandel:
> TLS ? Is it needed ? (Other PR)

We don't do TLS for within cluster gRPC calls anywhere else. So I don't think we need this (local SDK for example).

install/helm/agones/values.yaml
lacroixthomas: The flow might need more explanations, but basically this is the steps for the allocations::

1. Requests goes to the clients hot batch
2. Server sends a pull request to each clients (pullInterval 200 ms)
3. Client send the hotbatch to the server
4. Server calls the processallocation for each client batch requests (each requests is pushed into the allocator's internal queue, if there is new ones in the 50ms interval they will be batched / processed togethers)
5. Returns to each clients

I'll definitively need to add documentation about it, or maybe simplify some part if they don't make sense, or too much batch

pkg/processor/client.go
lacroixthomas: Ooh good point, will update that

pkg/gameserverallocations/controller.go
lacroixthomas: I will double check, but from what I remember, it's the same behaviour as what is currently happening, we return a 201 with the state as alloc contention, but might be wrong

cmd/processor/handler.go
lacroixthomas: That's a good question hehe, I was adding the context and it's cancel in case I needed it, but actually I don't need it (gonna get rid of both of them from the struct, I only need it from the newServiceHandler which send it to other part)

pkg/processor/client.go
lacroixthomas: Gonna switch that to 1, I added it for some manual tests / debugging, was to see how it would handle if the server would send multiple pull requests at the same time, will update it 😅

cmd/processor/handler.go
lacroixthomas: Oooh nice ! Gonna update it

pkg/processor/client.go
lacroixthomas: Oh indeed, I'll double check that, there is 8 locks within this file, might be a lot / will ensure they are needed and also checking that there is no race conditions or so 👌🏼

cmd/processor/handler.go
```diff
@@ -0,0 +1,347 @@
...
+			}).WithError(err).Error("Failed to send response")
+			continue
+		}
+	}
+}
+
+// StartPullRequestTicker periodically sends pull requests to all connected clients
+func (h *processorHandler) StartPullRequestTicker(ctx context.Context) {
+	go func() {
+		ticker := time.NewTicker(h.pullInterval)
```
markmandel: (non blocking, suggestion) - you could use https://pkg.go.dev/k8s.io/utils/clock/testing which could make unit testing easier, since you it comes with fake clocks.

cmd/processor/handler.go
```diff
@@ -0,0 +1,347 @@
+// Copyright 2026 Google LLC All Rights Reserved.
```
markmandel: We know what to put here now - so will need to switch the licences 😄

cmd/processor/handler_test.go
```diff
@@ -0,0 +1,772 @@
...
+}
+
+func TestStartPullRequestTickerRemovesFailingClient(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	h := newTestHandler(ctx, nil)
+	h.pullInterval = 20 * time.Millisecond
```
markmandel: As I said above - you could instead do a fake clock that ticks when you want it to, without having to overwrite the interval -- and then you don't need the `Eventually` either.

cmd/processor/main.go
```diff
@@ -206,3 +299,92 @@ func whenLeader(ctx context.Context, cancel context.CancelFunc, logger *logrus.E
...
+	// Access to the Agones resources through the Agones Clientset
+	agonesClient, err := versioned.NewForConfig(config)
+	if err != nil {
+		return nil, nil, errors.New("Could not create the agones api clientset")
+	}
+	return kubeClient, agonesClient, nil
+}
+
+// grpcCodeFromHTTPStatus converts an HTTP status code to the corresponding gRPC status code.
+func grpcCodeFromHTTPStatus(httpUnallocatedStatusCode int) codes.Code {
```
markmandel: Is this a copy paste from elsewhere? (I feel like it is?) if so, centralise it somewhere?

If not, feels a tad off to be in main, since it's shared in a few places in this pacakge - maybe put it in a /pkg/processor/processor.go ?

cmd/processor/handler.go
```diff
@@ -0,0 +1,347 @@
...
+	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/status"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+)
+
+// gameServerAllocator represent the interface to allocate game servers
+type gameServerAllocator interface {
```
markmandel: I just realised this is in `cmd/processor/` - it should probably follow the patterns of most other things and be in `/pgk/processor` , unless there is a good reason not to?

---

Probably doesn't matter, but I'll ask the question as I'm looking at it -- `/pkg/processor`, is that a good top level folder name (processor of what?)? I had the thought of `/pkg/gameserverallocations/processor` but I'm not 100% convinced. or `pkg/allocation/processor` ?

We have `/pkg/allocation` which is just the converters and the generated go code (which I haven't loved that it's just over there doing that weird thing in the corner).

Thank you for coming to my ted talk. Consider it non blocking, if the concern resonates with you as well let me know, otherwise ignore.

markmandel (commented):
Outside of previous comments, I think these are just nits really.

cmd/processor/handler.go
lacroixthomas: I'm gonna move the cmd/processor/handler to pkg/processor/handler (with the client etc.) 

But I can create another PR to move the pkg/processor to pkg/gameserverallocation/processor, IMO it would make sense, we already got the processor under allocator in the chart

markmandel: > But I can create another PR to move the pkg/processor to pkg/gameserverallocation/processor, IMO it would make sense, we already got the processor under allocator in the chart

Agreed - makes sense to move it in another PR 👍🏻

pkg/gameserverallocations/controller.go
lacroixthomas: Yep that seems correct 👌🏼 It's the current behavior

pkg/processor/client.go
lacroixthomas: Looks all good as well 🤞🏼 From what I've seen, there is no race condition or so

lacroixthomas:
Note to myself: Remember to rollback the feature gate to dev *

lacroixthomas:
I think I already seen this error :( 

```
Step #2 - "e2e-stable": VERBOSE: time="2026-04-23 01:28:14.554" level=info msg="checking if GameServer exists" error="client rate limiter Wait returned an error: context deadline exceeded" gs=game-serverppbms state= test=TestGameServerPodCompletedAfterCleanExit
Step #2 - "e2e-stable": VERBOSE:     gameserver_test.go:549: 
Step #2 - "e2e-stable": VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:549
Step #2 - "e2e-stable": VERBOSE:         	Error:      	Received unexpected error:
Step #2 - "e2e-stable": VERBOSE:         	            	client rate limiter Wait returned an error: context deadline exceeded
Step #2 - "e2e-stable": VERBOSE:         	Test:       	TestGameServerPodCompletedAfterCleanExit
Step #2 - "e2e-stable": VERBOSE: --- FAIL: TestGameServerPodCompletedAfterCleanExit (303.14s)
```

/gcbrun

lacroixthomas:
@markmandel  The e2e that were passing here were using the `processorAllocator=true` for the featureGate as Alpha 
https://github.com/agones-dev/agones/pull/4263#issuecomment-4308106108 

I just rolled that back to dev, should be all good now 😄 🤞🏼

cmd/processor/main.go
```diff
@@ -206,3 +322,57 @@ func whenLeader(ctx context.Context, cancel context.CancelFunc, logger *logrus.E
...
+	// Access to the Agones resources through the Agones Clientset
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, nil, errors.New("Could not create the kubernetes api clientset")
+	}
+
+	// Access to the Agones resources through the Agones Clientset
+	agonesClient, err := versioned.NewForConfig(config)
+	if err != nil {
+		return nil, nil, errors.New("Could not create the agones api clientset")
```
markmandel:
```suggestion
		return nil, nil, errors.Wrap(err,"Could not create the agones api clientset")
```

Just to maintain the context of the original error.

cmd/processor/main.go
```diff
@@ -206,3 +322,57 @@ func whenLeader(ctx context.Context, cancel context.CancelFunc, logger *logrus.E
...
+	go func() {
+		go func() {
+			<-ctx.Done()
+			grpcServer.GracefulStop()
+		}()
+
+		err := grpcServer.Serve(listener)
+		if err != nil {
+			logger.WithError(err).Fatal("Allocation service crashed")
+			os.Exit(1)
```
markmandel:
```suggestion
```

Since the fatal will already exit with a non zero code.

markmandel (approved):
Small nits, but merge when ready.

Issue #4258, opened 2025-08-27
[Feature Proposal] - End the game server allocation with the state GameServerAllocationUnAllocated when an update conflict occurs


miai10:
**Is your feature request related to a problem? Please describe.**
The Agones allocator has 2 possible errors that it can return (among others): no game server found or failed to update the game server CR. Both are depicting the same situation (basically no game server was found to make the allocation happen) but result in different outcomes from the public API which is confusing. The confusion comes from the fact that the API response will be nil and the error code will be translated into an internal error.

**Describe the solution you'd like**
Create ErrGameServerUpdateConflict and use it [here](https://github.com/googleforgames/agones/blob/release-1.51.0/pkg/gameserverallocations/allocator.go#L603) and treat the new error similarly to ErrNoGameServer [here](https://github.com/googleforgames/agones/blob/release-1.51.0/pkg/gameserverallocations/allocator.go#L277) thus ending the gsa with the state GameServerAllocationUnAllocated.

**Describe alternatives you've considered**
Failure to update the CR should result in the same outcome as no server found: returning ErrNoGameServer [here](https://github.com/googleforgames/agones/blob/release-1.51.0/pkg/gameserverallocations/allocator.go#L603)

**Additional context**
None

**Link to the Agones Feature Proposal (if any)**
None

**Discussion Link (if any)**
None

markmandel:
Not a bad idea at all - I am wondering if #4190 will render this issue obsolete though (or at least far more rare).

@lacroixthomas wdyt?

lacroixthomas:
> Not a bad idea at all - I am wondering if [#4190](https://github.com/googleforgames/agones/issues/4190) will render this issue obsolete though (or at least far more rare).
> 
> [@lacroixthomas](https://github.com/lacroixthomas) wdyt?

It should make it pretty rare indeed, but might still happens with the counter and lists though (if there is a lot of values in counters and lists + handling the end of sessions and GS state etc.), everything get's slower and conflicts could happens as the sdkserver will try to update the CRD, the controller and the processor

I think this issue is still relevant for this scenario (even though, having too many counters and lists might be wrong for many things, like if the server crashes, it will lose all the current active sessions) 

(About the SDKServer, we could "batch" the CRD update as well, to avoid updating the CRD for each call from the SDK (updating counter, updating list etc.), got this working: https://gist.github.com/lacroixthomas/031d8037fbf71ceb4abf2c661ae391b0 
[...]

PR #4267, opened 2025-09-05
Return GameServerAllocationUnAllocated when an game server update error occurs

miai10:
**What type of PR is this?**
>
> /kind breaking
/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

End the game server allocation with the state GameServerAllocationUnAllocated when an update error occurs

Closes https://github.com/googleforgames/agones/issues/4258

### DIFF
--- a/pkg/gameserverallocations/allocator.go
+++ b/pkg/gameserverallocations/allocator.go
@@ -67,6 +67,8 @@ var (
 	ErrConflictInGameServerSelection = errors.New("The Gameserver was already allocated")
 	// ErrTotalTimeoutExceeded is used to signal that total retry timeout has been exceeded and no additional retries should be made
 	ErrTotalTimeoutExceeded = status.Errorf(codes.DeadlineExceeded, "remote allocation total timeout exceeded")
+	// ErrGameServerUpdateConflict is returned when the game server selected for applying the allocation cannot be updated
+	ErrGameServerUpdateConflict = errors.New("Could not update the selected GameServer")
 )
 
 const (
@@ -273,10 +275,10 @@ func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocatio
 		return nil, err
 	}
 
-	switch err {
-	case ErrNoGameServer:
+	switch {
+	case goErrors.Is(err, ErrNoGameServer), goErrors.Is(err, ErrGameServerUpdateConflict):
 		gsa.Status.State = allocationv1.GameServerAllocationUnAllocated
-	case ErrConflictInGameServerSelection:
+	case goErrors.Is(err, ErrConflictInGameServerSelection):
 		gsa.Status.State = allocationv1.GameServerAllocationContention
 	default:
 		gsa.ObjectMeta.Name = gs.ObjectMeta.Name
@@ -600,7 +602,7 @@ func (c *Allocator) allocationUpdateWorkers(ctx context.Context, workerCount int
 							// we should wait for it to get updated with fresh info.
 							c.allocationCache.AddGameServer(gs)
 						}
-						res.err = errors.Wrap(err, "error updating allocated gameserver")
+						res.err = goErrors.Join(ErrGameServerUpdateConflict, err)
 					} else {
 						// put the GameServer back into the cache, so it's immediately around for re-allocation
 						c.allocationCache.AddGameServer(gs)
--- a/pkg/gameserverallocations/allocator_test.go
+++ b/pkg/gameserverallocations/allocator_test.go
@@ -496,7 +496,8 @@ func TestAllocatorAllocateOnGameServerUpdateError(t *testing.T) {
 	_, err := a.allocate(ctx, gsa.DeepCopy())
 	log.WithError(err).Info("allocate (private): failed allocation")
 	require.NotEqual(t, ErrNoGameServer, err)
-	require.EqualError(t, err, "error updating allocated gameserver: failed to update")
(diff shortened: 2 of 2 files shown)


miai10:
While doing the feature, I tried to understand [this](https://github.com/googleforgames/agones/blob/release-1.51.0/pkg/gameserverallocations/allocator.go#L279). 

ErrConflictInGameServerSelection seems similar to the new ErrGameServerUpdateConflict but it is used when removing an already removed server from the cache, [here](https://github.com/googleforgames/agones/blob/release-1.51.0/pkg/gameserverallocations/allocation_cache.go#L112). Maybe it is code that should have been cleaned up because I don't see a way to double remove. Also it is strange to translate that into contention.

pkg/gameserverallocations/allocator.go
```diff
@@ -600,7 +602,7 @@ func (c *Allocator) allocationUpdateWorkers(ctx context.Context, workerCount int
 							// we should wait for it to get updated with fresh info.
 							c.allocationCache.AddGameServer(gs)
 						}
-						res.err = errors.Wrap(err, "error updating allocated gameserver")
+						res.err = goErrors.Join(ErrGameServerUpdateConflict, err)
```
lacroixthomas: The goErrors.Join will add a `\n`, not sure we want this, what do you think about this or similar ?
```suggestion
						res.err = fmt.Errorf("%w: %v", ErrGameServerUpdateConflict, err)
```

pkg/gameserverallocations/allocator.go
```diff
@@ -67,6 +67,8 @@ var (
 	ErrConflictInGameServerSelection = errors.New("The Gameserver was already allocated")
 	// ErrTotalTimeoutExceeded is used to signal that total retry timeout has been exceeded and no additional retries should be made
 	ErrTotalTimeoutExceeded = status.Errorf(codes.DeadlineExceeded, "remote allocation total timeout exceeded")
+	// ErrGameServerUpdateConflict is returned when the game server selected for applying the allocation cannot be updated
+	ErrGameServerUpdateConflict = errors.New("Could not update the selected GameServer")
```
lacroixthomas: I always have a doubt about it.. but shouldn't that start with a lowercase ? (was checking the previous error message from the unit tests)

markmandel: Yes, lower case please!

pkg/gameserverallocations/allocator.go
markmandel: We have a legacy standard of using the `errors` library -- so stick to that please.

miai10: Applied Thomas' suggestion

pkg/gameserverallocations/allocator.go
```diff
@@ -600,7 +602,7 @@ func (c *Allocator) allocationUpdateWorkers(ctx context.Context, workerCount int
 							// we should wait for it to get updated with fresh info.
 							c.allocationCache.AddGameServer(gs)
 						}
-						res.err = errors.Wrap(err, "error updating allocated gameserver")
+						res.err = fmt.Errorf("%w: %v", ErrGameServerUpdateConflict, err)
```
markmandel: Please keep using the errors library we're been using for a while now (it gives us stack traces, and various things). It's pretty embedded now, so we'll stick with it. Surprised the `goerrors` package managed to slip in.

miai10: Sorry, it was not intentional, I just did not understand how to do it.

My goal was to have the error wrapping ErrGameServerUpdateConflict with the k8s err as details. That way I can use errors.Is later in the switch.

Would this work for you:
`
res.err = errors.Wrap(fmt.Errorf("%w: %v", ErrGameServerUpdateConflict, err), "error updating allocated gameserver")
`

markmandel: No worries at all - happens all the time.

Any reason not to return `ErrGameServerUpdateConflict` directly? There's no huge reason to wrap anyway.

miai10: The way I see it is:
- wrap adds the callstack
- ErrGameServerUpdateConflict is the base error and is mandatory to be able to use errors.Is
- err is the k8s error that adds more details.

Returning just ErrGameServerUpdateConflict is fine just that we are loosing context.

markmandel: True - but in this instance, the error type is a specific indicator of action - so we know what the context is. It's a very specific error coming from a very specific spot, that then gets consumed such that it's really never exposed to the outside world very much.

TL;DR - not sure the extra context is actually valuable in this particular instance.

pkg/gameserverallocations/allocator.go
```diff
@@ -273,10 +275,10 @@ func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocatio
 		return nil, err
 	}
 
-	switch err {
-	case ErrNoGameServer:
+	switch {
+	case goErrors.Is(err, ErrNoGameServer), goErrors.Is(err, ErrGameServerUpdateConflict):
```
markmandel:
```suggestion
	case ErrGameServerUpdateConflict:
```

same below, since we don't really use `goErrors` except to Join errors together below, and this is only needed on goErrors.Wrap(), and `pkg/errors` doesn't support `errors.Unwrap() error`

Maybe one day we'll do a big migration project? 😁

PR #4270, opened 2025-09-08
fix: broken websocket connection after upgrading github.com/grpc-ecosystem/grpc-gateway/v2

### LINKED ISSUE #4248: Calling WatchGameServer via websocket stopped working in v1.50
**What happened**:
We had issues using the *Unreal Engine SDK* after updating to v1.51. While debugging, we found out that calling the `/watch/gameserver` endpoint on the sdk-server via a websocket succeeds in v1.49 and returns a gameserver object. From version v1.50 onward the request times out. Using http, everything works fine.

That a request against the `/watch/gameserver` endpoint works via websockets and returns a gameserver object.

Checkout version v1.49 and start the sdk-server locally using `go run cmd/sdk-server/main.go --local` and fire up a request via curl using the following snippet:

```bash
  curl -iv \
       -H "Connection: Upgrade" \
       -H "Upgrade: websocket" \
       -H "Sec-WebSocket-Key: ExampleKey1234567890===" \
       -H "Sec-WebSocket-Version: 13" -X GET http://localhost:9358/watch/gameserver
```

Do the same after checking out v1.50 or v1.51 and the above command wont return.

- Agones version: v1.49, v1.50 & v1.51
- Kubernetes version (use `kubectl version`): none
- Cloud provider or hardware configuration: none
- Install method (yaml/helm): none

swermin:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

This PR introduces a change to the `tmc/grpc-websocket-proxy` to fix a deadlock when trying to establish a websocket connection to gRPC code that was generated by `github.com/grpc-ecosystem/grpc-gateway/v2` on version 2.26.2 and above.

`github.com/grpc-ecosystem/grpc-gateway/v2` on version 2.26.2 introduced this change [PR 5240](https://github.com/grpc-ecosystem/grpc-gateway/pull/5240) that causes request that has empty body specified to drain the body before proceeding with the call. This change was introduced back in April 2025 and the maintainer of the `tmc/grpc-websocket-proxy` has not been active for a while, making this breaking bug lingering and blocks projects that has websocket support as a requirement to be able to upgrade to newer agones versions.

As a workaround, due to critical functionality is broken, this PR introduces a fix in the vendor folder, so that we can unblock agones upgrades for projects that are depending on the websocket connection to work.

Closes #4248

This is just a fix for the watch gameserver websocket connection, as it only ever sends data from the sidecar to the gameserver executable. That is fine for this current setup. If we need bi-directional support whatsoever from the websocket connection then this change will not work as there is no functionality exposed to send data through the proxy.

### DIFF
--- a/build/includes/sdk.mk
+++ b/build/includes/sdk.mk
@@ -146,8 +146,10 @@ run-sdk-conformance-no-build: FEATURE_GATES ?=
 run-sdk-conformance-no-build: ensure-agones-sdk-image
 run-sdk-conformance-no-build: ensure-build-sdk-image
 	DOCKER_RUN_ARGS="--net host -e AGONES_SDK_GRPC_PORT=$(GRPC_PORT) -e AGONES_SDK_HTTP_PORT=$(HTTP_PORT) -e FEATURE_GATES='$(FEATURE_GATES)' $(DOCKER_RUN_ARGS)" COMMAND=sdktest $(MAKE) run-sdk-command & \
+	CONFORMANCE_TEST_PID=$$! && \
 	docker run -p $(GRPC_PORT):$(GRPC_PORT) -p $(HTTP_PORT):$(HTTP_PORT) -e "FEATURE_GATES=$(FEATURE_GATES)" -e "ADDRESS=" -e "TEST=$(TESTS)" -e "SDK_NAME=$(SDK_FOLDER)" -e "TIMEOUT=$(TIMEOUT)" -e "DELAY=$(DELAY)" \
-	--net=host $(sidecar_linux_amd64_tag) --grpc-port $(GRPC_PORT) --http-port $(HTTP_PORT)
+	--net=host $(sidecar_linux_amd64_tag) --grpc-port $(GRPC_PORT) --http-port $(HTTP_PORT) && \
+	wait $$CONFORMANCE_TEST_PID
 
 # Run SDK conformance test for a specific SDK_FOLDER
 run-sdk-conformance-test: TRIES=5
--- a/cmd/sdk-server/main.go
+++ b/cmd/sdk-server/main.go
@@ -252,7 +252,7 @@ func runGrpc(grpcServer *grpc.Server, grpcEndpoint string) {
 // runGateway runs the grpc-gateway
 func runGateway(ctx context.Context, grpcEndpoint string, mux *gwruntime.ServeMux, httpServer *http.Server) {
 	// nolint: staticcheck
-	conn, err := grpc.DialContext(ctx, grpcEndpoint, grpc.WithBlock(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	conn, err := grpc.NewClient(grpcEndpoint, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		logger.WithError(err).Fatal("Could not dial grpc server...")
 	}
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,10 @@ go 1.24.0
 
 toolchain go1.24.4
 
+// Hotfix to address websocket connection issues with github.com/grpc-ecosystem/grpc-gateway/v2 v2.26.2 and above.
+// See: https://github.com/grpc-ecosystem/grpc-gateway/issues/5326, https://github.com/googleforgames/agones/issues/4248
+replace github.com/tmc/grpc-websocket-proxy v0.0.0-20220101234140-673ab2c3ae75 => github.com/swermin/grpc-websocket-proxy v0.1.0
+
 require (
 	cloud.google.com/go/cloudbuild v1.15.1
 	cloud.google.com/go/compute/metadata v0.7.0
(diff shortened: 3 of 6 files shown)


markmandel:
> Final thing, while this should be pushed upstream it seems like there doesn't seem to be any traction from the maintainer, and since a lot of people (including me) depend on the websocket watch to work I added this change into the vendor directory for now.

Maybe silly question - but the watch gameserver websocket connection continues to work, no? (because the test checks this).

I thought this was just getting the GameServer through a websocket? Or maybe I misunderstood?

swermin:
Great question! I think I failed describing the issue properly.

The websocket connection is only used to listen on changes that happens on the gameserver CRD.
What happens is that when a connection is being established to the `/watch/gameserver` it hangs on the gateway side due to the introduction of `io.Copy(io.Discard, req.Body)`.
The websocket proxy sends a `io.Pipe()` as a body, meaning that the discard effectively hangs until the body closes.

The fix is to send an empty body. Firstly because we use a`GET` call, which does not expect a body at all, and secondly that will make the `io.Copy(io.Discard, req.Body)` succeed and continue the execution of the request. 
That means that the websocket connection is fully established and the receiving side will start getting the updates on the CRD as they happen.

The `ws-watch-test.go` doesn’t succeed at all without this change as far as I can tell.
It does four things:

- establish a websocket connection
- reserve a gamesever by calling `/reserve`
- Use http to get the gameserver crd and check if it is in `Reserved` state
- Poll the websocket connection for the CRD changes that shows the reserved message pushed.

The last point never executes because the connection never established, but hangs instead.

swermin:
Also, I don’t get why the gcburn failed 🤔

swermin:
Honestly though, the fix stared me right in the face. We don't need to fake the body, we just need to pass the request body to the proxy and that will work with both post/get calls

markmandel:
Looks like e2e tests are being extra flaky again.

lacroixthomas (commented):
LGTM ! Nice catch for the sdk Makefile ! 

I was wondering if we should add a comment on the code (as we updated the vendors directly ? 🤔)

markmandel:
Flake:

```
PASS test/e2e.TestFleetGSSpecValidation (9.81s)
=== RUN   TestSuperTuxKartGameServerReady
=== PAUSE TestSuperTuxKartGameServerReady
=== CONT  TestSuperTuxKartGameServerReady
    examples_test.go:83: 
        	Error Trace:	/go/src/agones.dev/agones/test/e2e/examples_test.go:83
        	            	waiting for {supertuxkart [{default  Dynamic <nil> 8080 0 UDP}] {false 60 0 30}  { 0 0} {{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []} {[] [] [{supertuxkart us-docker.pkg.dev/agones-images/examples/supertuxkart-example:0.19 [] []  [] [] [{ENABLE_PLAYER_TRACKING false nil}] {map[] map[] []} [] <nil> [] [] nil nil nil nil    nil false false false}] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] [] <nil> nil <nil> [] [] nil}} <nil> map[] map[] <nil>} GameServer instance readiness timed out (): waiting for GameServer 1758158444/supertuxkart-w7rfd to be Ready: GameServer reached terminal state Unhealthy
        	Test:       	TestSuperTuxKartGameServerReady
--- FAIL: TestSuperTuxKartGameServerReady (26.29s)
```

https://console.cloud.google.com/cloud-build/builds/9d80a8d5-68b1-4c41-85cf-64b4b2e002ac;step=1?authuser=1&project=agones-images

Blarg, this keeps failing on Autopilot for some reason. I do wonder why.

markmandel:
Quick gut check before I approve this for merge -

One issue with having this fix in the vendor directory is that any run of `go mod vendor` at this point will overview the patch.

Does it make sense for someone (@swermin ?) to maintain a fork of github.com/tmc/grpc-websocket-proxy with the fix in it, and we change our go.mod to point at the fork (probably with a `replace` declaration).

swermin:
I didn’t think of that, that a `go mod vendor` will revert the patch and then break things.
I don’t mind maintaining a fork with this change. I’ll update the PR tomorrow to reflect that change

swermin:
> LGTM ! Nice catch for the sdk Makefile !
> 
> I was wondering if we should add a comment on the code (as we updated the vendors directly ? 🤔)
> 
> /gcbrun

I did add a comment on the code, the part that needed a fix. Or do you mean on the other lines?

lacroixthomas:
> > LGTM ! Nice catch for the sdk Makefile !
> > 
> > I was wondering if we should add a comment on the code (as we updated the vendors directly ? 🤔)
> > 
> > /gcbrun
> 
> I did add a comment on the code, the part that needed a fix. Or do you mean on the other lines?

Oh indeed my bad, looks all good, or maybe in the go.mod if it uses a fork or something, otherwise all good

swermin:
I've updated to use a forked version of the module and used `replace` to highlight that it is for this hotfix

markmandel (approved):
oh yeah

swermin:
Woho! 🙌

markmandel:
Very nice work! Lemme know if I can do anything to help you be more active here 😃

swermin:
> Very nice work! Lemme know if I can do anything to help you be more active here 😃

Oh I’m constantly lurking in here, but I will try to be more engaging going forward! 😄

markmandel:
@swermin I went hunting for you in Slack 😄  could't find you.

swermin:
I just messaged you on slack! 🙌

PR #4272, opened 2025-09-09
Release v1.52.0

Sivasankaran25:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind release

Work On #4271

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -121,7 +121,7 @@ RUN mkdir -p /go/src/github.com/ahmetb && \
 
 # html checker
 RUN mkdir /tmp/htmltest && \
-    wget -O /tmp/htmltest/htmltest.tar.gz https://github.com/wjdp/htmltest/releases/download/v0.13.0/htmltest_0.13.0_linux_amd64.tar.gz && \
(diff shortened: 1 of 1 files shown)


site/htmltest.yaml
```diff
@@ -25,5 +25,6 @@ IgnoreURLs:
   - https://docs.unrealengine.com
   - https://github.com/googleforgames/agones/blob/main/docs/proposals
   - https://www.terraform.io/
+  - ^https?://
```
igooch: This means that our html test will not test any links. I'd recommend adjusting the `HTTPConcurrencyLimit` before resorting to effecting preventing this test from running.

Sivasankaran25: Updated

site/htmltest.yaml
```diff
@@ -18,11 +18,13 @@ OutputDir: /go/src/agones.dev/agones/site/tmp
...
 CheckExternal: true
 ExternalTimeout: 60
+HTTPConcurrencyLimit: 5
 IgnoreURLs:
   - http://localhost
   - https://twitter.com/agonesdev
   - https://www.youtube.com/playlist
   - https://docs.unrealengine.com
   - https://github.com/googleforgames/agones/blob/main/docs/proposals
+  - https://www.terraform.io/
```
markmandel, line 28: non blocking, but you could also `data-proofer-ignore` the link itself rather than skip it everywhere.

PR #4279, opened 2025-09-17
Update Add and Remove List Value in SDK Server

### LINKED ISSUE #4275: Incorrect response for list addValue and removeValue (agones SDK)
The docs say that the response for addValue and removeValue is the updated list, but I seem to always be getting { "name": "", "capacity": "0", "values": [] } as a response for both operations.

I didn't test the update method.

P.S.: this happens when communicating with the agones Sidecar via http requests

Sivasankaran25:
**What type of PR is this?**
>
> /kind breaking

/kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #4275


lacroixthomas:
Maybe we could add some unit tests to ensure there will be no regression in the future ? It's strange that no unit tests fails now that we returns the proper values (maybe we were not checking returned values at all)

Sivasankaran25:
> Maybe we could add some unit tests to ensure there will be no regression in the future ? It's strange that no unit tests fails now that we returns the proper values (maybe we were not checking returned values at all)

I have tested the test cases, and they are all passing. I also verified the behavior with the SDK, and it is working as expected.
image
image

pkg/sdkserver/sdkserver.go
```diff
@@ -1180,13 +1180,16 @@ func (s *SDKServer) AddListValue(ctx context.Context, in *beta.AddListValueReque
 			return nil, errors.Errorf("already exists. Value: %s already in List: %s", in.Value, in.Name)
 		}
 	}
+	if in.Value == "" {
```
igooch: Why this check? We don't have a requirement specifying that the string cannot be empty, only that a string must be unique.

pkg/sdkserver/sdkserver.go
```diff
@@ -1211,23 +1214,34 @@ func (s *SDKServer) RemoveListValue(ctx context.Context, in *beta.RemoveListValu
 	s.gsUpdateMutex.Lock()
 	defer s.gsUpdateMutex.Unlock()
 
-	// Verify value exists in the list
-	for _, val := range list.Values {
-		if in.Value != val {
-			continue
+	// Remove the first matching value directly from the in-memory list
```
igooch: We don't allow duplicate values, so there should only be one matching value in the list.

pkg/sdkserver/sdkserver.go
```diff
@@ -1180,13 +1180,16 @@ func (s *SDKServer) AddListValue(ctx context.Context, in *beta.AddListValueReque
 			return nil, errors.Errorf("already exists. Value: %s already in List: %s", in.Value, in.Name)
 		}
 	}
+	if in.Value == "" {
+		return nil, errors.Errorf("cannot add empty value to list: %s", in.Name)
+	}
 	list.Values = append(list.Values, in.Value)
 	batchList := s.gsListUpdates[in.Name]
 	batchList.valuesToAppend = list.Values
```
igooch: Looks like this is a pre-existing bug. The `valuesToAppend` shouldn't contain the entire list, only the `valuesToAppend` and the new value.
```suggestion
	batchList := s.gsListUpdates[in.Name]
	batchList.valuesToAppend = append(batchList.valuesToAppend, in.Value)
```
You might need to do the `append(batchList.valuesToAppend, in.Value)` in a separate variable first.

pkg/sdkserver/sdkserver.go
```diff
@@ -1211,23 +1214,34 @@ func (s *SDKServer) RemoveListValue(ctx context.Context, in *beta.RemoveListValu
...
+		newValues = append(newValues, v)
+	}
+
+	if !found {
+		return nil, errors.Errorf("not found. Value: %s not found in List: %s", in.Value, in.Name)
 	}
-	return nil, errors.Errorf("not found. Value: %s not found in List: %s", in.Value, in.Name)
+	list.Values = newValues
+
+	// Track this removal for batch persistence to K8s
```
igooch, line 1214: Because the order in which we update the list we first remove values, and then add values, we also as part of the `RemoveListValues` should check the `batchList.valuesToAdd`, and remove the value from that list if it exists.

igooch (commented):
Not explicitly mentioned in the issue for this PR is that the `RemoveListValues` and `AddListValues` are expected to be atomic. In general, I'm not sure what the benefit is of only holding the update diffs `gsCounterUpdates    map[string]counterUpdateRequest` and `gsListUpdates       map[string]listUpdateRequest` instead of keeping a full copy of the Counter or List that will be updated, and then on batch applying the copy of the Counter or List. If this List is large there might be some memory benefit.

pkg/sdkserver/sdkserver.go
```diff
@@ -1182,11 +1182,11 @@ func (s *SDKServer) AddListValue(ctx context.Context, in *beta.AddListValueReque
 	}
 	list.Values = append(list.Values, in.Value)
 	batchList := s.gsListUpdates[in.Name]
-	batchList.valuesToAppend = list.Values
+	batchList.valuesToAppend = append(batchList.valuesToAppend, in.Value)
 	s.gsListUpdates[in.Name] = batchList
 	// Queue up the Update for later batch processing by updateLists.
 	s.workerqueue.Enqueue(cache.ExplicitKey(updateLists))
-	return &beta.List{}, nil
+	return list, nil
```
igooch, line 1189: You'll also need to append the new value to the list before returning.

igooch (commented):
One last item, then looks good!

pkg/sdkserver/sdkserver.go
Sivasankaran25: Yes, I’ve already append the new value before returning

igooch: So it does, missed the unchanged line :)

Issue #4080, opened 2025-01-03
Feature: Script Based Autoscaler


markmandel:
I’ve had this idea floating around in my head for a while (I think it came up in a meeting?) and should really get it out 😆 and see if people like it.

**Is your feature request related to a problem? Please describe.**  
If there is a need for any kind of custom logic with an autoscaler, a webhook autoscaler is the only solution. The friction point there is that this needs to be a http service that is also hosted somewhere. Sometimes, that’s not an option, or at the very least, it would be handy to not have to deploy a whole extra service as well as Agones itself to a cluster \- but still have the ability to implement custom autoscaling logic with relative ease.

**Describe the solution you'd like**  
Be able to configure the math for autoscaling using a script that’s embedded in the `FleetAutoscaler` configuration.

My first thought is embedding Go into the yaml, via [https://github.com/traefik/yaegi](https://github.com/traefik/yaegi) , but there are a variety of scripting languages (linked below) , so something like this is a sacrificial draft:

```yaml
apiVersion: autoscaling.agones.dev/v1
kind: FleetAutoscaler
metadata:
  name: fleet-autoscaler-counter
spec:
  fleetName: fleet-example
  policy:
    type: Script
    script: |
       func Scale(r FleetAutoscaleRequest) (FleetAutoscaleResponse, error) {
           response := FleetAutoscaleResponse{
             UID = r.UID
             Scale = true
             Replicas = r.Status.AllocatedReplicas + 1
           }
           return response, nil
       }
```

Some extra options that might be needed/wanted:

* `function` \- the name of the function to use for scaling. Could have a default value of “Scale” (or similar), could also introspect the script and if there is only one function, and therefore use that one.

I do like yaegi though, since we can reuse the Go types we use inside Agones for the scripting language without as much need for interop layers.

**Describe alternatives you've considered**

Keep only the webhook autoscaler. A webhook does allow for custom logic as needed.

The only other downside to having this, is testability could be tricky. 

In theory, people could use the script library itself to write unit tests for the script contents, likely pulled from their YAML files. A *little* tricky, but definitely not impossible. We could provide an example (and we’d probably want to have one anyway for our unit tests).

**Additional context**

* [https://agones.dev/site/docs/getting-started/create-webhook-fleetautoscaler/](https://agones.dev/site/docs/getting-started/create-webhook-fleetautoscaler/)  
[...]

PR #4281, opened 2025-09-21
feat(autoscaling): CRDs for Wasm autoscaler policy

markmandel:
**What type of PR is this?**
>

/kind breaking

> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

- CRDs for WasmPolicy, with accompanying validation functions and tests
- Rename Go type WebhookPolicy to URLConfiguration so it makes more sens when reused across difference autoscaling needs and update codebase.

Work on #4080

This doesn't break older version of Agones Go libraries who use `WebhookPolicy`, as the backing structure of Yaml or JSON doesn't change. It's just if they are using Agones as a library to interact with Agones they will need to rename the type on upgrade. (this is something we've done before).


markmandel:
> LGTM !
> 
> Just got a question regarding your comment about the changes for the `webhookPolicy` to `URLConfiguration`, how would that be communicated on the release, is it something that needs to be added somewhere to not forget it ? 🤔

Because of the label `kind/breaking` the tooling that makes the release notes will automatically categorise it out as breaking, so it'll be included in the release notes and news.

Issue #4284, opened 2025-09-24
Patch Release 1.52.1


igooch:
# Patch Release 1.52.1

Modified release steps for creating a release patch version.

The cherry picked commit is specific to the fix for the Bitnami image #4270.

## Steps

- [x] Run `make shell` and run `gcloud config configurations activate agones-images`.
- [x] `git checkout release-1.52.0 && git pull --rebase upstream release-1.52.0`
- [x] Run `git cherry-pick 409a6fc3e96d82bda9a16fb10b6288d82c5fd356`
- [x] Run `make pre-build-release` to ensure all example images exist on agones-images/examples repository and to deploy the 1.52.1-1 service on GCP/App Engine/Services.
- [x] Manual update references to 1.52.0 to 1.52.1 (the script sdk-update-version does not currently have a function for patching.)
  - [x] "build/Makefile",
  - [x] "install/helm/agones/Chart.yaml",
  - [x] "install/yaml/install.yaml",
  - [x] "install/helm/agones/values.yaml",
  - [x] "sdks/nodejs/package.json",
  - [x] "sdks/nodejs/package-lock.json",
  - [x] "sdks/unity/package.json",
  - [x] "sdks/csharp/sdk/AgonesSDK.nuspec",
  - [x] "sdks/csharp/sdk/csharp-sdk.csproj",
  - [x] "sdks/rust/Cargo.toml"
- [x] Create a *draft* release with the [release template](https://github.com/googleforgames/agones/blob/main/docs/governance/templates/release.md).
    - [x] run `make release-example-image-markdown` to populate example images and append the output in `Images available with this release` section
    - [x] Draft a new release with [release draft](https://github.com/googleforgames/agones/releases). Update the `Tag version` and `Release title` with the release version and click on `Generate release notes` to generate the release notes with `Full Changelog` info for 1.52.1. Make sure to add the description. Include the `Images available with this release` section from the previous step that will be present after the `Full Changelog` and save the draft.
    - [x] copy the 1.52.1 release details from the `Full Changelog` and paste it on top of the CHANGELOG.md file
- [x] Site updated
    - [x] Create a new file named 1.52.1.md in `/site/content/en/blog/releases`. Copy the draft release content in this file (this will be what you send via email)[refer the previous release file].
    - [x] run `make site-server` frequently to make sure everything looks fine for the release in your localhost
[...]

PR #4286, opened 2025-09-25
Patch-releasev1.52.1

Sivasankaran25:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind release

Works on #4284

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -82,7 +82,7 @@ gcs_engine_id = "016691298986124624340:x7qv2dywdao"
 # current release branch. Never is rc.
 release_branch = "release-1.52.0"
 # the main version. Never is rc.
-release_version = "1.52.0"
+release_version = "1.52.1"
 
 # shown for production
 supported_k8s = ["1.31", "1.32", "1.33"]


site/config.toml
```diff
@@ -80,9 +80,9 @@ github_repo = "https://github.com/googleforgames/agones"
 gcs_engine_id = "016691298986124624340:x7qv2dywdao"
 
 # current release branch. Never is rc.
-release_branch = "release-1.52.0"
+release_branch = "release-1.53.0"
```
igooch: Looks like the script doesn't work for the patch version. This should still be release-1.52 as that is the release branch for the PR.

site/config.toml
```diff
@@ -80,9 +80,9 @@ github_repo = "https://github.com/googleforgames/agones"
...
 # the main version. Never is rc.
-release_version = "1.52.0"
```
igooch, line 85: This one should be updated to the patch version, 1.52.1.

site/config.toml
Sivasankaran25: Updated

site/config.toml
Sivasankaran25: Done

site/config.toml
```diff
@@ -80,9 +80,9 @@ github_repo = "https://github.com/googleforgames/agones"
 gcs_engine_id = "016691298986124624340:x7qv2dywdao"
 
 # current release branch. Never is rc.
-release_branch = "release-1.52.0"
+release_branch = "release-1.52.1"
```
igooch:
```suggestion
release_branch = "release-1.52.0"
```

We're using the release-1.52.0 branch, and not creating a new one, so this should stay the same.

igooch (commented):
A few last items.

Looks like there's two lingering references to 1.52.0 that should be updated to 1.52.1:

https://github.com/googleforgames/agones/blob/d68370de95ce2fc06faada76ebbb1c4b41a108ca/cloudbuild.yaml#L427

https://github.com/googleforgames/agones/blob/d68370de95ce2fc06faada76ebbb1c4b41a108ca/sdks/unity/package.json#L3

And then the one comment below that the name of the branch is still release-1.52.0.

site/config.toml
Sivasankaran25: Updated

igooch (approved):
LGTM

gongmax (approved):

Issue #4292, opened 2025-10-01
Patch Release 1.52.2


igooch:
# Patch Release 1.52.2

Modified release steps for creating a release patch version.

Includes the 1.52.1 patch release (the fix for the Bitnami image #4270), and the cherry picked commit fix for the hard coded main build in the release automation #4291.

Note: "If needed" means if this step only needs to be done if it was modified as part of the patch release or otherwise should be updated.

## Prerequisites

- [x] Editor-level access to the agones-images project.
- [x] Permission to publish new versions of the App Engine application.
- [x] Write access to the Agones GitHub repository.
- [x] git remote -v should show:
  - [x] An origin remote pointing to your personal Agones fork (e.g., git@github.com:yourname/agones.git).
  - [x] An upstream remote pointing to git@github.com:googleforgames/agones.git.
- [x] A gcloud config configurations configuration named agones-images pointing to the same project.
- [x] Edit access to the[ Agones Release Calendar](https://calendar.google.com/calendar/u/0?cid=Z29vZ2xlLmNvbV84MjhuOGYxOGhmYnRyczR2dTRoMXNrczIxOEBncm91cC5jYWxlbmRhci5nb29nbGUuY29t).
- [x] Approval for any modifications to the [Release Title and Description](https://docs.google.com/document/d/1bRZCxYB8lrVcrru41b6s5D_9uU0zS49vVGdBhg0yDIY/edit)

## Steps

- [x] Run `make shell` and run `gcloud config configurations activate agones-images`.
- [x] Create a new branch for the patch release, and base it off of the previous patch release. Note: In future releases this should be done with `make create-patch-release-branch PREVIOUS_VERSION=1.52.1 PATCH_VERSION=1.52.2`.
  - [x] Run `git remote update -p`
  - [x] Run `git fetch upstream --tags`
  - [x] Run `git checkout -b release-1.52.2 v1.52.1`
  - [x] Run `git status` to confirm you are on the expected branch name.
  - [x] Run `git log` to confirm the most recent commits are what you expect. (They should be the same as the release tagged v1.52.1.)
  - [x] Run `git push -u upstream release-1.52.2`
- Ensure all example images exist on agones-images/examples repository and to deploy the 1.52.0-1 service on GCP/App Engine/Services. Note: in future releases this should be done with `make pre-build-release`.
  - [x] Run `cd build/release`.
[...]

PR #4294, opened 2025-10-01
Patch Release v1.52.2

Sivasankaran25:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind release

Works on #4292


site/config.toml
```diff
@@ -80,9 +80,9 @@ github_repo = "https://github.com/googleforgames/agones"
 gcs_engine_id = "016691298986124624340:x7qv2dywdao"
 
 # current release branch. Never is rc.
-release_branch = "release-1.52.0"
+release_branch = "release-1.52.1"
```
igooch:
```suggestion
release_branch = "release-1.52.2"
```

igooch (changes requested):
Looks like we need to update `_BASE_VERSION: 1.52.1` to `_BASE_VERSION: 1.52.2`  in the cloudbuild.yaml. I'll update the patch instructions to include that step.

site/config.toml
Sivasankaran25: Updated

igooch (approved):
LGTM

igooch:
Looks like build is (correctly) failing due to artifacts not found for 1.52.1. I'd recommend updating the file site/content/en/blog/releases/1.52.1.md to remove most of the content and write something like "do not use this patch release; images are not correct".

PR #4298, opened 2025-10-06
Add Wasm autoscaler policy support to FleetAutoscaler CRDs YAML

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix
> /kind release

- Add 'Wasm' as a new policy type option for FleetAutoscaler
- Create reusable 'url.configuration' template for webhook and wasm URL configs
- Add wasm policy schema with function, config, from.url, and hash fields
- Refactor webhook policy to use shared url.configuration template
- Update generated install.yaml with corresponding CRD changes

Work on #4080

N/A

### DIFF
--- a/install/yaml/install.yaml
+++ b/install/yaml/install.yaml
@@ -6076,6 +6076,7 @@ spec:
                       - Webhook
                       - Counter
                       - List
+                      - Wasm
                       - Schedule
                       - Chain
                     buffer:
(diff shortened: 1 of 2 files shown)


install/yaml/install.yaml
```diff
@@ -6095,7 +6096,7 @@ spec:
                           anyOf:
                             - type: integer
                             - type: string
-                    webhook:
+                    webhook:
```
lacroixthomas, line 6099: Not sure why it added these spaces, I don't see them on the template

markmandel: Honestly - no idea. Code gen add them 🤷🏻 I can't see any good reason why.

PR #4319, opened 2025-10-30
Populate Env Vars for GameServer init containers

### LINKED ISSUE #4241: AGONES_SDK_GRPC_PORT and AGONES_SDK_HTTP_PORT should also be populated for init containers
**What happened**:
The environment variables ```AGONES_SDK_GRPC_PORT``` and ```AGONES_SDK_HTTP_PORT``` are not being injected into init containers of GameServer pods. These variables are correctly populated for the main game server container but are missing from any init containers defined in the GameServer specification.

Both ```AGONES_SDK_GRPC_PORT``` and ```AGONES_SDK_HTTP_PORT``` environment variables should be automatically injected into all containers within the GameServer pod, including init containers. This would allow init containers to interact with the Agones SDK during the initialization phase, enabling use cases such as:
- Pre-game configuration validation via SDK
- Health check setup before main container starts
- Game state initialization that requires SDK communication

1. Create a GameServer with an init container:
```yaml
apiVersion: agones.dev/v1
kind: GameServer
metadata:
  name: test-gameserver
spec:
  ports:
  - name: default
    containerPort: 7654
  template:
    spec:
      initContainers:
      - name: init-container
        image: alpine:latest
        command: ["/bin/sh", "-c"]
        args: 
        - |
          echo "AGONES_SDK_GRPC_PORT: $AGONES_SDK_GRPC_PORT"
          echo "AGONES_SDK_HTTP_PORT: $AGONES_SDK_HTTP_PORT"
          env | grep AGONES
      containers:
      - name: simple-game-server
        image: us-docker.pkg.dev/agones-images/examples/simple-game-server:latest
```
2. Apply the GameServer: ```kubectl apply -f gameserver.yaml```
3. Check init container logs: ```kubectl logs <pod-name> -c init-container```
4. Observe that AGONES_SDK_GRPC_PORT and AGONES_SDK_HTTP_PORT are empty or not present
5. Compare with main container: ```kubectl exec <pod-name> -c simple-game-server -- env | grep AGONES```
6. Notice that the main container has these variables properly set

- This issue prevents init containers from using the Agones SDK client libraries
- Workaround currently requires hardcoding the default ports (9357 for gRPC, 9358 for HTTP) in init containers
- The SDK sidecar is running and accessible during init container execution, but without the port environment variables, SDK clients cannot automatically connect

giornetta:
**What type of PR is this?**
/kind bug

This MR adds code to populate environment variables for init containers defined in a `GameServer` spec.

This allows init containers to communicate with the SDK server during initialization, enabling use cases where setup logic needs to interact with Agones APIs before the main game server container starts.

Closes #4241

### DIFF
--- a/pkg/fleetautoscalers/controller.go
+++ b/pkg/fleetautoscalers/controller.go
@@ -60,20 +60,25 @@ import (
 	"k8s.io/utils/clock"
 )
 
+// fasState is the state for each Fleet Autoscaler instance
+// expand this structure if extra state is needed for other autoscaler
+// strategies.
+type fasState struct {
+	wasmPlugin *extism.Plugin
+}
+
 // fasThread is used for tracking each Fleet's autoscaling jobs
 type fasThread struct {
 	cancel     context.CancelFunc
-	state      map[string]any
+	state      fasState
 	generation int64
 }
 
 // close cancels the context and cleans up any resources
 func (ft *fasThread) close(ctx context.Context) {
 	ft.cancel()
-	if plugin, ok := ft.state[wasmStateKey]; ok {
-		if p, ok := plugin.(*extism.Plugin); ok {
-			_ = p.Close(ctx) // Ignore any errors during cleanup
-		}
+	if ft.state.wasmPlugin != nil {
+		_ = ft.state.wasmPlugin.Close(ctx)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/gameservers/controller.go
```diff
@@ -832,31 +832,43 @@ func (c *Controller) addGameServerHealthCheck(gs *agonesv1.GameServer, pod *core
 }
 
 func (c *Controller) addSDKServerEnvVars(gs *agonesv1.GameServer, pod *corev1.Pod) {
+	sdkEnvVars := sdkEnvironmentVariables(gs)
+	if sdkEnvVars == nil {
+		// If a gameserver was created before 1.1 when we started defaulting the grpc and http ports,
```
markmandel, line 837: I wish I could remember why this is here 🤣

markmandel (commented):
Thanks, we needed this!

/gcbrun

PR #4324, opened 2025-11-04
Fix update counter to return correct values

### LINKED ISSUE #4220: Rest UpdateCounter not returning correct values
**What happened**:
api is returning wrong data for UpdateCounter

Get current count and capacity
but got:
```
{"name":"","count":"0","capacity":"0"}
```

```
curl -sSd '{"count": "5'", "capacity": "64"}' -H "Content-Type: application/json" -X PATCH http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/counters/players
```

both return the right data:
```
curl http://localhost:${AGONES_SDK_HTTP_PORT}/v1beta1/counters/players
```
->
```
{"name":"players","count":"39","capacity":"64"}
```

```
curl http://localhost:${AGONES_SDK_HTTP_PORT}/gameserver
```
->
```
    "ports": [
      {
        "name": "test",
        "port": 13743
      }
    ],
    "players": null,
    "counters": {
      "players": {
        "count": "58",
        "capacity": "64"
      }
    },
    "lists": {}

```

**Environment**:
- Agones version: 1.49.0

indurireddy-TF:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #4220


aditya-shantanu:
Add a test that validates this change ?

indurireddy-TF:
> Add a test that validates this change ?

Hi @aditya-shantanu I added a new test case to verify the changes.

pkg/sdkserver/sdkserver_test.go
```diff
@@ -1270,6 +1271,18 @@ func TestSDKServerUpdateCounter(t *testing.T) {
...
 		},
+		"projected state returned": {
+			counterName: "starfish",
+			requests: []*beta.UpdateCounterRequest{
+				{CounterUpdateRequest: &beta.CounterUpdateRequest{
+					Name:      "starfish",
+					CountDiff: 15,
+				}},
+			},
+			want:       agonesv1.CounterStatus{Count: int64(25), Capacity: int64(100)},
```
aditya-shantanu: nit: Change this to something other than the value before it so you are sure it got updated.

aditya-shantanu (approved):

pkg/sdkserver/sdkserver_test.go
indurireddy-TF: Updated CountDiff to 7. The initial counter value remains 10, consistent with existing test cases.

markmandel:
Thanks for this! I would like to take this for a spin when integrated with our simple gameserver container we use for e2e testing. 

We don't have a great way of automating testing this, so it may need to be manual 🙃

pkg/sdkserver/sdkserver_test.go
```diff
@@ -1328,11 +1341,15 @@ func TestSDKServerUpdateCounter(t *testing.T) {
 
 			// Update the Counter
 			for i, req := range testCase.requests {
-				_, err = sc.UpdateCounter(context.Background(), req)
+				resp, err := sc.UpdateCounter(context.Background(), req)
 				if testCase.updateErrs[i] {
 					assert.Error(t, err)
 				} else {
 					assert.NoError(t, err)
+					if test == "projected state returned" {
```
markmandel: 🤔 any reason to only do this for this test? Could we not do this for every test for extra test coverage?

indurireddy-TF: Thanks, Mark. I see your point completely on expanding the coverage.

I initially limited the assertion to a new, dedicated test case to ensure we validated the new return behavior without altering the legacy state assumptions of the older tests.

However, I absolutely agree that extending the assertion offers better coverage. I can proceed with refactoring to assert the full result object in every successful test case.

markmandel: Awesome, once that change comes in - then I think we are good to merge!

pkg/sdkserver/sdkserver.go
```diff
@@ -945,7 +945,25 @@ func (s *SDKServer) UpdateCounter(_ context.Context, in *beta.UpdateCounterReque
...
+func projectCounterState(name string, batchCounter counterUpdateRequest) *beta.Counter {
+	currentCapacity := batchCounter.counter.Capacity
+	if batchCounter.capacitySet != nil {
+		currentCapacity = *batchCounter.capacitySet
+	}
+	currentCount := batchCounter.counter.Count
+	if batchCounter.countSet != nil {
+		currentCount = *batchCounter.countSet
+	}
+	currentCount += batchCounter.diff
```
igooch, line 961: +1 to the bot's comment. The `CapacitySet` doesn't validate the count diff, so it's possible the `currentCount` could be outside the range [0,Capacity].

markmandel:
I think this is good to go. @igooch did you want to do another review?

igooch (approved):
LGTM, good work.

Issue #4321, opened 2025-10-31
Flaky: TestSuperTuxKartGameServerReady


markmandel:
**What happened**:

e2e test TestSuperTuxKartGameServerReady is flake.

What's _really_ weird, is that when it fails in one test, it continues to fail on subsequent tests. This also goes across standard and autopilot, although I _think_ it happens more on autopilot.

Example: https://console.cloud.google.com/cloud-build/builds/b71c42ca-43da-429b-83b8-7a497d518708;step=1?project=agones-images

Here are the logs and the test output.

```
VERBOSE: === RUN   TestSuperTuxKartGameServerReady
VERBOSE: === PAUSE TestSuperTuxKartGameServerReady
VERBOSE: === CONT  TestSuperTuxKartGameServerReady
VERBOSE: time="2025-10-31 20:43:08.178" level=info msg="GameServer created, waiting for Ready" gs=supertuxkart-4cchc test=TestSuperTuxKartGameServerReady
VERBOSE: time="2025-10-31 20:43:08.282" level=info msg="Waiting for states to match" awaitingState=Ready currentState=Starting gs=supertuxkart-4cchc test=TestSuperTuxKartGameServerReady
VERBOSE: time="2025-10-31 20:43:09.261" level=info msg="Waiting for states to match" awaitingState=Ready currentState=Starting gs=supertuxkart-4cchc test=TestSuperTuxKartGameServerReady
VERBOSE: time="2025-10-31 20:43:10.257" level=info msg="Waiting for states to match" awaitingState=Ready currentState=Scheduled gs=supertuxkart-4cchc test=TestSuperTuxKartGameServerReady
VERBOSE: time="2025-10-31 20:43:11.258" level=info msg="Waiting for states to match" awaitingState=Ready currentState=Scheduled gs=supertuxkart-4cchc test=TestSuperTuxKartGameServerReady
VERBOSE: time="2025-10-31 20:43:12.261" level=info msg="Waiting for states to match" awaitingState=Ready currentState=Scheduled gs=supertuxkart-4cchc test=TestSuperTuxKartGameServerReady
VERBOSE: time="2025-10-31 20:43:13.268" level=info msg="Waiting for states to match" awaitingState=Ready currentState=Scheduled gs=supertuxkart-4cchc test=TestSuperTuxKartGameServerReady
[...]

PR #4331, opened 2025-11-10
test: simplify autoscaler e2e tests for minikube compatibility

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

- Change ImagePullPolicy from PullAlways to PullIfNotPresent
- Remove unnecessary container port name 'autoscaler'
- Remove TargetPort reference to simplify service configuration

N/A

For some reason using a name for TargetPort failed on minikube, but this fixed it 🤷‍♂️

### DIFF
--- a/pkg/gameserverallocations/allocator.go
+++ b/pkg/gameserverallocations/allocator.go
@@ -67,6 +67,8 @@ var (
 	ErrConflictInGameServerSelection = errors.New("The Gameserver was already allocated")
 	// ErrTotalTimeoutExceeded is used to signal that total retry timeout has been exceeded and no additional retries should be made
 	ErrTotalTimeoutExceeded = status.Errorf(codes.DeadlineExceeded, "remote allocation total timeout exceeded")
+	// ErrGameServerUpdateConflict is returned when the game server selected for applying the allocation cannot be updated
+	ErrGameServerUpdateConflict = errors.New("could not update the selected GameServer")
 )
 
 const (
@@ -274,7 +276,7 @@ func (c *Allocator) allocateFromLocalCluster(ctx context.Context, gsa *allocatio
 	}
 
 	switch err {
-	case ErrNoGameServer:
+	case ErrNoGameServer, ErrGameServerUpdateConflict:
 		gsa.Status.State = allocationv1.GameServerAllocationUnAllocated
 	case ErrConflictInGameServerSelection:
 		gsa.Status.State = allocationv1.GameServerAllocationContention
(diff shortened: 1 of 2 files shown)


test/e2e/fleetautoscaler_test.go
```diff
@@ -771,7 +770,7 @@ func defaultAutoscalerWebhook(namespace string, fixedReplicasEnabled string) (*c
 			Ports: []corev1.ServicePort{{
 				Name:       "newport",
 				Port:       8000,
-				TargetPort: intstr.IntOrString{StrVal: "autoscaler"},
+				TargetPort: intstr.FromInt(8000),
```
lacroixthomas: I was doing some test locally, it seems to be working by using `intstr.FromString("autoscaler")` (and keeping the name: "autoscaler" L750

intstr.FromString add the type with it: `IntOrString{Type: String, StrVal: val}`, I've got it working locally, not sure how it works on the background, if you can also test on your side, otherwise it's good to me to keep it like this 😄

lacroixthomas: It might be by default to an int for the type and the validation expect it to be so and fails the creation of the service *
I'm not fully sure about it, but it's possible

markmandel: Good find! That worked!

As an interesting thing, checked the implementation:

```
// FromString creates an IntOrString object with a string value.
func FromString(val string) IntOrString {
	return IntOrString{Type: String, StrVal: val}
}
```

No `Type` value on the original implementation!

markmandel:
Something _weird_ happened to CI.

markmandel:
* #4321 again

markmandel:
```
docker run --rm -v /workspace/build//.config/gcloud:/root/.config/gcloud -v ~/.kube/:/root/.kube -v ~/.config/helm:/root/.config/helm -v ~/.cache/helm:/root/.cache/helm -v /workspace:/go/src/agones.dev/agones -v /workspace/build//.gomod:/go/pkg/mod -v /workspace/build//.gocache:/root/.cache/go-build -e "KUBECONFIG=/root/.kube/config" -e "GO111MODULE=on" -w /go/src/agones.dev/agones  agones-build:a8d1f85932 go test -mod=vendor -race agones.dev/agones/pkg/... \
    agones.dev/agones/sdks/... agones.dev/agones/cmd/...
FAIL	agones.dev/agones/sdks/rust/target/debug/deps/rmetajHKWjf [setup failed]
FAIL	agones.dev/agones/sdks/... [setup failed]
```

wha?

markmandel:
image
ohnno

PR #4332, opened 2025-11-12
feat(autoscaler)!: Remove caBundle requirement for HTTPS URLs

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix
> /kind release

BREAKING CHANGE: HTTPS webhook and WASM URLs no longer require caBundle when configured. The impact of this is that HTTPS URLs for a webhook or wasm package that are served via a trusted certificate authority will be processed successfully. This is particularly important for Wasm based autoscaling, as WASM packages will likely be served from Cloud storage providers, via HTTPS which is not controlled by the end user.

The caBundle field is now only required for URLs using custom or self-signed certificates, as they will fail when being called from the Fleet Autoscaler.

This simplifies configuration for standard HTTPS endpoints while maintaining security for custom certificate scenarios.

Changes:
- Updated URL validation logic to allow HTTPS without caBundle
- Modified HTTP client to be stored per Fleet Autoscaler, allowing for a custom http client per Fleet Autoscaler
- Updated tests to verify new behavior
- Adjusted validation error messages

Work on #4080

This will also make things easier if people want to host webhooks on a functions as a service, or serverless platform for ease of use -- since they often won't control the certificates in those environments.

### DIFF
--- a/examples/supertuxkart/Dockerfile
+++ b/examples/supertuxkart/Dockerfile
@@ -20,8 +20,8 @@ COPY main.go go.mod ./
 RUN go mod tidy &&\
     go build -o wrapper .
 
-# Installing STK
-FROM debian:bookworm
+# Installing STK - forky has 1.5, so we'll jump ahead a little.
+FROM debian:forky
 
 WORKDIR /home/supertuxkart
 RUN apt update && apt-get install -y supertuxkart && apt clean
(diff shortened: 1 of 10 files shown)


lacroixthomas:
Looks all good ! 👌🏼

PR #4350, opened 2025-11-26
test: improve TestMultiClusterAllocationFromLocal flakiness.

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Not sure that I fixed it, but made some improvements and added some extra logging as well:

- Change assert.NoError to require.NoError to stop test execution on allocation failure
- Add error wrapping with response body context in executeAllocation helper
- Fix createRequest to check error before setting headers
- Add Accept header to allocation requests

N/A

N/A


pkg/gameserverallocations/controller_test.go
```diff
@@ -806,11 +807,13 @@ func createRequest(gsa *allocationv1.GameServerAllocation) (*http.Request, error
...
 
 	r, err := http.NewRequest(http.MethodPost, "/", buf)
-	r.Header.Set("Content-Type", k8sruntime.ContentTypeJSON)
-
 	if err != nil {
 		return nil, err
 	}
+
+	r.Header.Set("Content-Type", k8sruntime.ContentTypeJSON)
+	r.Header.Set("Accept", k8sruntime.ContentTypeJSON)
```
igooch, line 815: Not sure we want this? Either the response is always json (even errors), and we never hit the `err = json.Unmarshal(jsn, ret)` on line 783, or we expect errors to be in a non-json format like a plain text stack trace, so we want those errors to fall through and get caught and wrapped and returned.

markmandel: > Not sure we want this? Either the response is always json (even errors), and we never hit the err = json.Unmarshal(jsn, ret) on line 783

It'll still fail, since the format of the error json won't match the struct - and even if it did go through, the result would end up not aligning with the expected result.

The flakiness _may_ be a result of the underlying K8s system thinking sometimes we don't want JSON -- so this is just to make sure that's not the actual issue.

PR #4353, opened 2025-11-26
Exclude wasm from example image check

igooch:
**What type of PR is this?**

/kind cleanup

When running `make check-example-versions` we get the error:

```
2025/11/26 22:31:51 Getting versions of autoscaler-wasm
2025/11/26 22:31:51 Could not get (local) version of example autoscaler-wasm: no version could be found
exit status 1
```

The example currently uses a path to the raw file, not to a Docker container in a registry like the rest of the examples, so there is no need to check for a Docker image.

https://github.com/googleforgames/agones/blob/0007c692714701a8d127567e40311ae5e6eec61a/examples/autoscaler-wasm/autoscaler.yaml#L39

NA


markmandel:
Good catch!

PR #4359, opened 2025-11-29
More fixes for SuperTuxKart example to attempt to fix flakiness.

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

I've run this several hundred times on my Minikube cluster, and so far it has failed to fail!

My theory is that the log file is doing something weird with the log file before it really starts pushing logs to it -- and therefore occasionally will assume there is no log at all.

To fix this I made the following changes:

- Bump version from 0.20 to 0.21 in Makefile, fleet.yaml, and gameserver.yaml
- Update Agones SDK from v1.51.0 to v1.53.0
- Replace deprecated hpcloud/tail with nxadm/tail v1.4.11
- Fix Dockerfile syntax: change 'as' to 'AS' for build stage
- Enhance log file monitoring with existence and size checks before tailing

Work on #4321

Like last time, will update the e2e test once this has merged and the image is up on the registry.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,19 @@ again.
 
 Participation in this project comes under the [Contributor Covenant Code of Conduct](code-of-conduct.md)
 
+## Finding Issues to Work On
+
+If you're new to the project and want to help, but don't know where to start, look for issues labeled with:
+
+- **[good first issue](https://github.com/googleforgames/agones/labels/good%20first%20issue)** - These are great for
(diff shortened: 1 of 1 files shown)


lacroixthomas:
The build has been pushed: us-docker.pkg.dev/agones-images/examples/supertuxkart-example:0.21 from what I see

PR #4375, opened 2025-12-04
feat: support init sidecar containers

### LINKED ISSUE #4358: [Feature Proposal] - Include InitContainer Sidecars in Port assignment
We would like to use `InitContainers` for sidecar containers, given that the lifecycle is better managed, but port assignment in Agones currently only looks for the containers in `Containers` meaning the sidecars cannot use external ports.

https://github.com/googleforgames/agones/blob/main/pkg/apis/agones/v1/gameserver.go#L731

Here, `InitContainers` with an `Always` restart policy should be considered.

None, there is no other way to do this.

**Additional context**

**Discussion Link (if any)**

nrwiersma:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

This allows port allocations on Init Container Sidecars. It also adds validation for init containers, which seems to be missing.

Init container sidecars offer better life cycle management, and is now fully supported in Kubernetes. In general Agones does not block the creation of init sidecars, but currently ports do not work on them.

Closes #4358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,51 @@
 # Changelog
 
+# [v1.56.0](https://github.com/googleforgames/agones/tree/v1.56.0) (2026-02-24)
+
+[Full Changelog](https://github.com/googleforgames/agones/compare/v1.55.0...v1.56.0)
+
+**Breaking changes**
+- Feat: Move SidecarContainers feature gate to Beta by @AryanBagade in https://github.com/googleforgames/agones/pull/4364
+- Fix Prometheus installation with helm chart v28.9.0. by @markmandel in https://github.com/googleforgames/agones/pull/4448
+
+**Implemented enhancements**
+- Build simple-game-server in CI for e2e tests by @IrvingMg in https://github.com/googleforgames/agones/pull/4401
+- Upgrade Hugo and Docsy by @Sivasankaran25 in https://github.com/googleforgames/agones/pull/4404
+- add optional agones.metrics.serviceMonitor.additonalLabels map to values.yaml in helm chart by @jessebot in https://github.com/googleforgames/agones/pull/4424
+- Added dark mode to the website by @thegalacticlor in https://github.com/googleforgames/agones/pull/4439
+- site: update homepage showcase and navbar styling by @markmandel in https://github.com/googleforgames/agones/pull/4438
+
+**Fixed bugs**
+- fix(sdk): add validation to SetAnnotation to prevent infinite retry loop by @indurireddy-TF in https://github.com/googleforgames/agones/pull/4417
+- Fixes post_cloudbuild release by explicitly passing variables as Makefile arguments by @igooch in https://github.com/googleforgames/agones/pull/4423
+- fix: Ensure the processclient have a correct timeout / fix flacky test by @lacroixthomas in https://github.com/googleforgames/agones/pull/4402
+- docs: fix alert shortcode syntax to process Markdown content by @markmandel in https://github.com/googleforgames/agones/pull/4437
+
+**Other**
+- Preparation for Release v1.56.0 by @Sivasankaran25 in https://github.com/googleforgames/agones/pull/4422
+- chore(sdk/nodejs): update dependencies and migrate to ESLint 9 config by @markmandel in https://github.com/googleforgames/agones/pull/4412
+- chore(sdk/nodejs): migrate linting from ESLint to Biome by @smazmi in https://github.com/googleforgames/agones/pull/4426
+- fix(docs): update external links in out-of-cluster dev server guide by @markmandel in https://github.com/googleforgames/agones/pull/4427
+- test: standardize runtime.ParseFeatures error handling and mutex usage by @markmandel in https://github.com/googleforgames/agones/pull/4416
+- Add Singularity 6 Palia architecture article to third party content by @markmandel in https://github.com/googleforgames/agones/pull/4430
+- Add CEDIL slides link for Sega Sonic Rumble CEDEC talk by @markmandel in https://github.com/googleforgames/agones/pull/4432
+- Add Google Doodle containerized games case study to third party content by @markmandel in https://github.com/googleforgames/agones/pull/4433
+- docs: add blog entry for "Making and Scaling a Game Server in Kubernetes using Agones" by @noetarbouriech in https://github.com/googleforgames/agones/pull/4428
+- Document Allocation batching by @markmandel in https://github.com/googleforgames/agones/pull/4440
+- build(deps): bump github.com/go-git/go-git/v5 from 5.16.4 to 5.16.5 in /build/scripts/example-version-checker by @dependabot[bot] in https://github.com/googleforgames/agones/pull/4442
+- test: fix flaky allocation metrics test server startup by @srpvpn in https://github.com/googleforgames/agones/pull/4445
+- docs: convert alert shortcodes to format for markdown rendering by @markmandel in https://github.com/googleforgames/agones/pull/4446
+- build: improve test-gen-api-docs failure message by @markmandel in https://github.com/googleforgames/agones/pull/4453
+
+**New Contributors**
+- @smazmi made their first contribution in https://github.com/googleforgames/agones/pull/4426
+- @IrvingMg made their first contribution in https://github.com/googleforgames/agones/pull/4401
+- @noetarbouriech made their first contribution in https://github.com/googleforgames/agones/pull/4428
+- @jessebot made their first contribution in https://github.com/googleforgames/agones/pull/4424
+- @AryanBagade made their first contribution in https://github.com/googleforgames/agones/pull/4364
+- @thegalacticlor made their first contribution in https://github.com/googleforgames/agones/pull/4439
+- @srpvpn made their first contribution in https://github.com/googleforgames/agones/pull/4445
+
 # [v1.55.0](https://github.com/googleforgames/agones/tree/v1.55.0) (2026-01-13)
 
 [Full Changelog](https://github.com/googleforgames/agones/compare/v1.54.0...v1.55.0)
--- a/build/e2e-image/Dockerfile
+++ b/build/e2e-image/Dockerfile
@@ -1,8 +1,15 @@
 FROM gcr.io/cloud-builders/gcloud
 
 RUN apt-get update && \
-    apt-get install -y wget psmisc make gcc jq zip && \
-    apt-get clean
+    apt-get install -y --no-install-recommends \
+    wget \
+    psmisc \
+    make \
+    gcc \
+    jq \
+    zip && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
 # install go
 WORKDIR /usr/local
--- a/build/includes/website.mk
+++ b/build/includes/website.mk
@@ -111,7 +111,8 @@ test-gen-api-docs: ensure-build-image
 	sort /tmp/generated.html > /tmp/generated.html.sorted
 	$(GEN_API_DOCS)
 	sort $(expected_docs) > /tmp/result.sorted
-	diff -bB /tmp/result.sorted /tmp/generated.html.sorted
+	diff -bB /tmp/result.sorted /tmp/generated.html.sorted || \
+		(echo "Error: API docs are out of date. Please run 'make gen-api-docs' to regenerate them since the CRD information in /pkg/apis has changed." && exit 1)
 
 # Remove feature expiry/publish version shortcodes update in site/content/en/docs
 feature-shortcode-update: ensure-build-image
--- a/install/helm/agones/Chart.yaml
+++ b/install/helm/agones/Chart.yaml
@@ -15,8 +15,8 @@
 # Declare variables to be passed into your templates.
 
 apiVersion: v1
-appVersion: "1.56.0-dev"
-version: 1.56.0-dev
+appVersion: "1.56.0"
+version: 1.56.0
 name: agones
 description: a library for hosting, running and scaling dedicated game servers on Kubernetes.
 keywords:
--- a/install/helm/agones/values.yaml
+++ b/install/helm/agones/values.yaml
@@ -321,7 +321,7 @@ agones:
         timeoutSeconds: 1
   image:
     registry: us-docker.pkg.dev/agones-images/release
-    tag: 1.56.0-dev
+    tag: 1.56.0
     controller:
       name: agones-controller
       pullPolicy: IfNotPresent
--- a/install/yaml/install.yaml
+++ b/install/yaml/install.yaml
@@ -59,7 +59,7 @@ metadata:
   namespace: agones-system
   labels:
     app: agones
-    chart: agones-1.56.0-dev
+    chart: agones-1.56.0
     release: agones-manual
     heritage: Helm
 ---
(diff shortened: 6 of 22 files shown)


pkg/apis/agones/v1/gameserver.go
```diff
@@ -726,7 +729,28 @@ func (gss *GameServerSpec) FindContainer(name string) (int, corev1.Container, er
 	return -1, corev1.Container{}, errors.Errorf("Could not find a container named %s", name)
 }
 
+// HasContainer determines if the GameServerSpec has a container with the specified name.
+// Init containers with RestartPolicy Always will also be considered.
```
markmandel: 🤔 any reason not to just check every init container?

pkg/gameservers/controller_test.go
```diff
@@ -155,6 +156,105 @@ func TestControllerSyncGameServer(t *testing.T) {
...
+
+	t.Run("Creating a new GameServer", func(t *testing.T) {
+		c, mocks := newFakeController()
+		updateCount := 0
+		podCreated := false
+		fixture := &agonesv1.GameServer{ObjectMeta: metav1.ObjectMeta{Name: "test", Namespace: "default"},
+			Spec: agonesv1.GameServerSpec{
+				Ports: []agonesv1.GameServerPort{
+					{ContainerPort: 7777},
+					{ContainerPort: 8888, Container: ptr.To("sidecar")},
```
markmandel, line 170: well that's a handy package I didn't know about! `k8s.io/utils/pt` going to steal that.

pkg/apis/agones/v1/gameserver.go
```diff
@@ -596,12 +596,15 @@ func (gss *GameServerSpec) Validate(apiHooks APIHooks, devAddress string, fldPat
 		}
 
 		if p.Container != nil && gss.Container != "" {
-			_, _, err := gss.FindContainer(*p.Container)
-			if err != nil {
+			if !gss.HasContainer(*p.Container) {
```
markmandel: Curious why not extend `gss.FindContainer()` to search init containers, rather than add another function?

markmandel (commented):
Looking good! Just a few questions.

Couple of things to round it out:
1. We would want an e2e test to test it on a cluster - should be able to use the simple-game-server to test like we do in our other `e2e/gameserver.go` tests.
2. We should write some docs to say you can use init containers. We could update the GameServer `containers` field as well as the GameServers.md reference file.

https://agones.dev/site/docs/contribute/documentation-editing-contribution/#within-a-page helps here as well.

pkg/apis/agones/v1/gameserver.go
nrwiersma: The main container should not be a sidecar, and `FindContainer` is used to find that. I really didnt find a good way to distinguish functions as to if we want to find a container or containers including sidecars. Maybe you have an idea here.

pkg/apis/agones/v1/gameserver.go
nrwiersma: You likely dont want to bind to a non-sidecar init container, cause it will be stopped before the main container can run.

pkg/apis/agones/v1/gameserver.go
markmandel: > The main container should not be a sidecar, and FindContainer is used to find that.

Not sure what you mean by "a sidecar container" - why not? Why not give people to choice to pick whichever container they want? Or can you expand further on your reasoning here, maybe I'm missing something.

In my head - gameserver containers (tend) to never restart (especially when `SidecarContainers` is enabled)  -- so if you do want a gameserver container as an init container, you have to make it Always under your code, which seems, a little odd.

Agones has always been philosophically a more open system that let's you do what you see fit, and if you shoot yourself in the foot, it's kinda on you 😄

nrwiersma: Fair enough, but `FindContainer` is still not a good fit, as it returns the index of the container, which becomes a bit meaningless, and it is a public API, so cannot really be changed.

markmandel:
Apologies - realised this slipped through with all the goings on. I need to sit down and re-read how init sidecar containers work to make sure I understand all the bits and piece of this before reviewing one more time.

pkg/apis/agones/v1/gameserver.go
markmandel: Ahhh this is the magic part of the docs I'd totally forgotten:

> If an init container is created with its restartPolicy set to Always, it will start and remain running during the entire life of the Pod. This can be helpful for running supporting services separated from the main application containers.

site/content/en/docs/Reference/gameserver.md
```diff
@@ -159,6 +159,14 @@ The `spec` field is the actual GameServer specification and it is composed as fo
 The GameServer resource does not support updates. If you need to make regular updates to the GameServer spec, consider using a [Fleet]({{< ref "/docs/Reference/fleet.md" >}}).
 {{< /alert >}}
 
+{{% feature publishVersion="1.55.0" %}}
+## Sidecar Containers
+
+Agones supports the use of [Sidecar Containers](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/).
```
markmandel: We've always supported having sidecar containers -- I think the difference you want to highlight here is that you can assign ports to sidecar containers, whereas you couldn't before?

pkg/apis/agones/v1/gameserver.go
markmandel: Oooh! For some reason in my head I thought you might want to use an sidecar container as the gameserver container. 

🤔 should we allow it?

site/content/en/docs/Reference/gameserver.md
markmandel, line 141: This isn't alpha anymore - shall we update it, since we're here?

This could also be the spot where we document "hey, you can also do either standard containers or sidecar containers - if you enable the correct feature flag. Bit trickier on formatting though.

We should definitely also update the reference:
https://github.com/googleforgames/agones/blob/1db6361cfff7d8a652e62b49992725f0bff8a17b/pkg/apis/agones/v1/gameserver.go#L296-L298

We've used `Alpha` and `Beta` tags in that kind of documentation to show they are feature flagged.

See https://github.com/googleforgames/agones/blob/db5946af038ac4b48940fcf78c1f124d83a2a629/pkg/apis/agones/v1/gameserver.go#L328-L330 for an example.

pkg/apis/agones/v1/gameserver.go
markmandel: I was doing some review - we only use FindContainer in this Validate function - so if we want to rewrite it, we should. Maybe something like `gss.HasContainer(container, includeSidecar)` rather than two separate functions?

markmandel:
This looks like a pretty legit e2e test failure:

```
VERBOSE: time="2026-01-08 06:38:21.182" level=info msg="Waiting for states to match" awaitingState=Ready currentState=Starting gs=game-serverfhhw4 test=TestSideCarCommunicatesWhileTerminating
VERBOSE: time="2026-01-08 06:38:21.789" level=info msg="Checking GameServerSet" fleet=simple-fleet-1.0jb4t2 gss0="{Replicas:1 ReadyReplicas:0 ReservedReplicas:0 AllocatedReplicas:1 ShutdownReplicas:0 Players:0xc000c54ae0 Counters:map[] Lists:map[]}" gss1="{Replicas:4 ReadyReplicas:3 ReservedReplicas:0 AllocatedReplicas:0 ShutdownReplicas:0 Players:0xc000c54880 Counters:map[] Lists:map[]}" test=TestFleetScaleWithDualAllocations
VERBOSE:     gameserver_test.go:832: Could not get a GameServer ready: creating {udp-server [{gameport  Dynamic <nil> 7654 0 UDP} {second-gameport  Dynamic 0xc000cd4320 5000 0 UDP}] {false 0 0 0}  { 0 0} {{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []} {[] [{sidecar-server us-docker.pkg.dev/agones-images/examples/simple-game-server:0.40 [] [-port 5000]  [] [] [] {map[] map[] []} [] <nil> [] [] nil nil nil nil   IfNotPresent nil false false false}] [{udp-server us-docker.pkg.dev/agones-images/examples/simple-game-server:0.40 [] []  [] [] [] {map[] map[] []} [] <nil> [] [] nil nil nil nil   IfNotPresent nil false false false}] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] [] <nil> nil <nil> [] [] nil}} <nil> map[] map[] <nil>} GameServer instances failed (): admission webhook "validations.agones.dev" denied the request: GameServer.agones.dev "udp-server59459" is invalid: spec.ports[1].container: Invalid value: "sidecar-server": VERBOSE: Container must be empty or the name of a container in the pod template
VERBOSE: --- FAIL: TestGameServerWithPortsMappedToInitSidecarContainers (3.40s)
VERBOSE: === CONT  TestGameServerWithPortsMappedToMultipleContainers
```

nrwiersma:
> This looks like a pretty legit e2e test failure:

Yeah, I could not get the e2e to run on my machine, so this was a first take. Will take a look at all of this next week when I am back at work.

markmandel:
@nrwiersma if you continue to struggle to get your e2e tests running locally, start a thread in #development in Slack and we'll get you up and running

pkg/apis/agones/v1/gameserver.go
nrwiersma: I personally would have avoided breaking the public API, but if you prefer that, happy to do it.

pkg/apis/agones/v1/gameserver.go
nrwiersma: I would have said no. It is just weird to have it function this way. Having the GameServer in a sidecar points towards having something else in the "main" container, which is just an odd situation IMO.

nrwiersma:
e2e is now running, but I get a strange non-agones related issue. I cannot seem to get traffic into the init sidecar container, even though a port is assigned and a hostport is assigned. I am not sure if I am running into k8s like we saw before, but I am using 1.35 and it still happens.

nrwiersma:
~Hostport on init containers is just ignored, and no traffic is ever forwarded. It is unclear if this is a CNI thing or k8s things (my money is on CNI). Kinda makes this a non-starter and init sidecars are just not going to work if they need external traffic.~

This works on cilium but not GKE or kind, so it is for sure a CNI issue. Not that that helps a whole bunch, but good to know.

markmandel:
@igooch do you have any insight into the GKE issue? Should hotport work on an init container?

igooch:
> [@markmandel](/markmandel)

Is this the issue that you're running into? https://docs.cloud.google.com/kubernetes-engine/docs/how-to/dataplane-v2#connectivity-issue-hostport

nrwiersma:
@igooch From my understanding the V2 Dataplane is based on Cilium, which have fixed the issue: https://github.com/cilium/cilium/issues/16191
We were running V1, and could not get the CNI to pass traffic to an initContainer sidecar with a `hostport`.

markmandel:
@nrwiersma what do we want to do with this PR? Sounds like it only works on certain network configurations which is sad -- but also means we can't have it be e2e tested on GKE. We could always release it with specific "hey this only works on these specific Kubernetes configs" but that seems icky 😬 but I can't think of anything better.

WDYT?

nrwiersma:
Tend to agree. The behaviour is correct, so I would remove the e2e and add a warning in the docs that it only works with come CNIs at the moment. Work for you?

markmandel:
Ah! I know what this is - `make gen-api-docs` needs to be run when CRDs are changed, even if it's just commends.

test/e2e/gameserver_test.go
```diff
@@ -781,6 +782,89 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
...
+		t.Fatalf("Could not get a GameServer ready: %v", err)
+	}
+	defer framework.AgonesClient.AgonesV1().GameServers(framework.Namespace).Delete(ctx, readyGs.ObjectMeta.Name, metav1.DeleteOptions{}) // nolint: errcheck
+	assert.Equal(t, readyGs.Status.State, agonesv1.GameServerStateReady)
+
+	interval := 2 * time.Second
+	timeOut := 60 * time.Second
+
+	expectedMsg1 := "Ping 1"
+	errPoll := wait.PollUntilContextTimeout(context.Background(), interval, timeOut, true, func(_ context.Context) (done bool, err error) {
```
markmandel: Looks like this is going to get bumped to next release -- sorry about the delay on this one!

Given this has asserts inside polls (which has had issues in the past), and we want to move to using `EventuallyWithT` overall (it's _so much better_), let's take the time to move to `EventuallyWithT` please.

https://pkg.go.dev/github.com/stretchr/testify/assert#EventuallyWithT

test/e2e/gameserver_test.go
```diff
@@ -781,6 +782,89 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
...
+			t.Logf("Could not message GameServer on %s: %v. Will try again...", firstPort, err)
+		}
+		return err == nil && strings.Contains(res, expectedMsg1), nil
+	})
+	if errPoll != nil {
+		assert.FailNow(t, errPoll.Error(), "expected no errors after polling a port: %s", firstPort)
+	}
+
+	expectedMsg2 := "Ping 2"
+	errPoll = wait.PollUntilContextTimeout(context.Background(), interval, timeOut, true, func(_ context.Context) (done bool, err error) {
```
markmandel: Same here, EventuallyWithT

site/content/en/docs/Reference/gameserver.md
```diff
@@ -159,6 +159,18 @@ The `spec` field is the actual GameServer specification and it is composed as fo
...
 {{% /alert %}}
 
+{{% feature publishVersion="1.56.0" %}}
+## Sidecar Containers
+
+Agones supports the use of [Sidecar Containers](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) with exposed ports.
+This requires the `RestartPolicy` to be set to `Always` and can be referenced in the `ports.*.container` field. Typically they used to run
+services that need to run alongside the GameServer container, such as log shippers, monitoring agents, allocators and other supporting services.
+
+{{< alert title="Warning" color="warning">}}
```
markmandel, line 169: Nice

pkg/apis/agones/v1/gameserver.go
```diff
@@ -720,6 +722,8 @@ func (gs *GameServer) IsActive() bool {
 
 // FindContainer returns the container specified by the name parameter. Returns the index and the value.
 // Returns an error if not found.
+//
+// Deprecated: use HasContainer instead.
```
markmandel: No need to deprecate. Just deleted it.

test/e2e/gameserver_test.go
```diff
@@ -781,6 +782,89 @@ func TestGameServerWithPortsMappedToMultipleContainers(t *testing.T) {
 	assert.NoError(t, errPoll, "expected no errors after polling a port: %s", secondPort)
 }
 
+func TestGameServerWithPortsMappedToInitSidecarContainers(t *testing.T) {
+	t.Skip("Skip test. Only works on some CNIs. The current e2e platform does not support ports on init containers.")
```
markmandel, line 786: Good call. Nice to have it if we can ever test it reliably.

markmandel (commented):
Let's go!

PR #4379, opened 2025-12-08
test: deferred cleanup for GameServer in TestSuperTuxKartGameServerReady

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Add a deferred function to properly delete the GameServer after the test completes, ensuring resources are cleaned up even if the test fails. Also update error handling to check if readyGs is not nil before logging events.

This makes rerunning the test hundreds of time to check for flakes much easier without a super large cluster.

N/A

N/A


lacroixthomas (approved):
Oooh good catch, not sure we clean up the ressources all the times even from other e2e in case of errors, could definitively be an issue 
LGTM !

PR #4382, opened 2025-12-10
Add Wasm Fleet Autoscaler quickstart documentation

### LINKED ISSUE #4080: Feature: Script Based Autoscaler

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix
> /kind release

This adds a new quickstart guide for creating a Fleet Autoscaler using Wasm modules.

Closes #4080

ALL DONE

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -69,11 +69,11 @@ KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 GS_TEST_IMAGE ?= us-docker.pkg.dev/agones-images/examples/simple-game-server:0.39
 
 # Enable all beta feature gates. Keep in sync with `true` (beta) entries in pkg/util/runtime/features.go:featureDefaults
-BETA_FEATURE_GATES ?= "CountsAndLists=true&GKEAutopilotExtendedDurationPods=true&PortPolicyNone=true&PortRanges=true&RollingUpdateFix=true&ScheduledAutoscaler=true"
+BETA_FEATURE_GATES ?= "CountsAndLists=true&GKEAutopilotExtendedDurationPods=true&PortPolicyNone=true&PortRanges=true&RollingUpdateFix=true&ScheduledAutoscaler=true&FleetAutoscaleRequestMetaData=true"
 
 
 # Enable all alpha feature gates. Keep in sync with `false` (alpha) entries in pkg/util/runtime/features.go:featureDefaults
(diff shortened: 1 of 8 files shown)


lacroixthomas:
Looking great ! 👌🏼 
This seems to be the last PR for the wasm fleet autoscaler ? 🎉

lacroixthomas (approved):

PR #4386, opened 2025-12-13
fix(sdk): validate labels in SetLabel to prevent retry loops

### LINKED ISSUE #4385: Error handling on client SDK feature "setLabel"
When attempting to submit an invalid label with the REST API. rather than returning a failure response the client SDK will enter an infinite loop of attempting to submit the label without any additional output unless debug logging is enabled, this also disables further creation of labels.

Simple failure responses in the form of 400 or 500 response codes with a simple JSON body stating success or failure along with the attempted label key and value.

I did briefly consider monitoring the labels but considering the loop knocks out the set label feature it would be a bit moot, could also force restart the client sdk sidecar somehow but I'm not sure if that will break anything upstream when considering health checking.

**Additional context**
Infinite label create loop log sample:
Image

NA

**Discussion Link (if any)**
https://agones.slack.com/archives/C9DGM5DS8/p1765412391189909

rajanarahul93:
/kind bug

This PR adds early validation to the SDK sidecar `SetLabel` RPC to prevent invalid Kubernetes labels from being enqueued and retried indefinitely.

Previously, `SetLabel` accepted any key/value pair and queued it for synchronization. When the label was invalid (for example, containing spaces), the Kubernetes API server rejected the update. Because the failure was deterministic, the SDK worker queue retried forever, resulting in a retry loop and blocking subsequent label updates.

This change validates label keys and values up front using Kubernetes validation rules and returns a clear `InvalidArgument` gRPC error to the client, avoiding unnecessary retries and improving debuggability.

Closes #4385

This PR intentionally limits its scope to `SetLabel`, as described in the linked issue.
A similar validation pattern may also apply to `SetAnnotation`, but that is left out here to keep the fix minimal and focused.

### DIFF
--- a/pkg/metrics/controller_test.go
+++ b/pkg/metrics/controller_test.go
@@ -26,7 +26,6 @@ import (
 	"agones.dev/agones/pkg/util/runtime"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.opencensus.io/metric/metricdata"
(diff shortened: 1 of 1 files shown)


pkg/sdkserver/sdkserver.go
```diff
@@ -575,6 +579,29 @@ func (s *SDKServer) Health(stream sdk.SDK_HealthServer) error {
...
+
+	if errs := validation.IsValidLabelValue(kv.Value); len(errs) > 0 {
+		return nil, status.Errorf(
+			codes.InvalidArgument,
+			"invalid label value %q for key %q: %s",
+			kv.Value,
+			kv.Key,
+			strings.Join(errs, ", "),
+		)
+	}
```
lacroixthomas, line 608: Even though that seems totally valid, I wonder if we should not use `errors.Errorf` instead, to be consistent with the rest of the code ? 🤔 @markmandel

lacroixthomas (commented):
Hi @rajanarahul93,
Thanks for the contribution ! 

I've put a minor change, but will need confirmation first

rajanarahul93:
Good question, thanks for raising it!

I used `status.Errorf(codes.InvalidArgument, ...)` intentionally here so that the SDK returns a proper gRPC `InvalidArgument` error, which is translated by the grpc-gateway into an HTTP 400 response for REST clients.

Using `errors.Errorf` would result in a generic gRPC error (`codes.Unknown`), which would map to an HTTP 500 and wouldn’t reflect a client-side validation failure.

Happy to adjust if there’s a preferred pattern for returning gRPC status errors elsewhere in the SDK.

pkg/sdkserver/sdkserver.go
markmandel: You know - I'm not against it. But it does mean this method could return a 4xx rather than a 5xx like in most (all?) other cases.

I'm almost wondering if we should create a ticket to go back and make our grpc errors proper gRPC errors. The only thing I can think it _might_ break is manual REST implementations that expect a 500 response in some way. WDYT?

lacroixthomas:
Hey ! @rajanarahul93  
I don't see any issues to return the status code from gRPC, it would indeed give more feedback for the user, would make things clearer 
I was checking the sdks (./sdks/*) and they seems to be all fine as well they already returns the error as expected (from what I see, nothing to change / no breaking changes)

I don't know what you think about it @markmandel  but I'm fine with it, LGTM, but if you're also good with it, I'll create some issues to ensure the other functions from the sdkServers also returns gRPC status code to be consistent (and also give more context / proper status code for the user) 

/gcbrun

markmandel:
> Hey ! @rajanarahul93  
> I don't see any issues to return the status code from gRPC, it would indeed give more feedback for the user, would make things clearer 
> I was checking the sdks (./sdks/*) and they seems to be all fine as well they already returns the error as expected (from what I see, nothing to change / no breaking changes)
> 

That's a fair point. I should go look, but I'm willing to bet that most SDK methods don't even return errors ever? Or if they do, it's very rare. Pretty much everything goes straight into a worker queue. So this shouldn't be a big breaking change really. 

(Totally on mobile and haven't looked at the code at all 🙂)

lacroixthomas:
From what I've seen, they only returns go errors (even for invalidArguments - which is why I think we can create issues to improve the other ones as well), but the SDKs either returns it directly or wrap the errors (depending of the language - most of them don't return any errors indeed), I can only see benefit from it, and in term of breaking changes, I only see improvements of the error handling, for the SetLabels from this PR, it was previously returning no errors at all 

(No pressure at all, was only going through some PR reviews 😄)

rajanarahul93:
Thanks a lot for the review and the discussion!

Glad to hear this approach makes sense and improves the user-facing error handling.  
I’m happy to keep this PR scoped to `SetLabel` for now, and it sounds great to track consistency improvements for other SDK server methods separately.

Let me know if you’d like me to do anything else here.

markmandel:
Let's get this merged! Shall we also do SetAnnotation() at some point? If someone could make an issue for it, that would be ace.

/gcbrun

PR #4412, opened 2026-01-09
chore(sdk/nodejs): update dependencies and migrate to ESLint 9 config

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Upgrade Node.js SDK dependencies to latest versions and migrate ESLint configuration from legacy .eslintrc.json to the new flat config format required by ESLint 9.

This was mostly just to clean up the dependabot PRs that have been there for ages - hopefully I did right by NodeJS code 😄 

@steven-supersolid did I do okay? 😄  (or anyone else with Node experience).

### DIFF
--- a/pkg/processor/client_test.go
+++ b/pkg/processor/client_test.go
@@ -22,6 +22,7 @@ import (
 
 	allocationpb "agones.dev/agones/pkg/allocation/go"
 	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
 	"google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/grpc/metadata"
 )
@@ -145,7 +146,7 @@ func TestProcessorClient_Allocate(t *testing.T) {
 			batchSize: 1,
 			setupResponse: func(_ *mockStream, _ []string) {
 				// Do not send any batch response, let it timeout
-				time.Sleep(100 * time.Millisecond)
+				time.Sleep(300 * time.Millisecond)
 			},
 			expectError: []bool{true},
 		},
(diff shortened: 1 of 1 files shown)


lacroixthomas (approved):
LGTM !

But it's been quite some time I've been doing some Node, I've been looking a bit on the changes with ESLint 9 etc. it's looking all good, but a second approval would be great though 😄

sdks/nodejs/package.json
```diff
@@ -4,14 +4,15 @@
         "url": "https://github.com/googleforgames/agones/issues"
     },
     "dependencies": {
-        "@grpc/grpc-js": "1.10.9",
-        "google-protobuf": "3.21.2"
+        "@grpc/grpc-js": "^1.14.3",
```
steven-supersolid: It may be safer to pin these versions, i.e. remove the ^
This way if there is a later vulnerability introduced in a dependency then it won't bubble up through the agones package
We leave dev dependencies unpinned as these will only be installed on the developer machine

sdks/nodejs/eslint.config.js
```diff
@@ -0,0 +1,37 @@
...
+	{
+		languageOptions: {
+			ecmaVersion: 2018,
+			globals: {
+				...globals.node,
+				...globals.es6,
+				...globals.jasmine
+			}
+		},
+		rules: {
```
steven-supersolid, line 29: This looks good but apparently these stylistic rules are no longer recommended and may not work. If `npm run lint` works then all is good for now

I recommend moving to biome as we can just use the standard rules with no overrides with the consequence quotes will change to double but this could be a separate task

markmandel: #4425 created.

sdks/nodejs/package.json
markmandel: Done!

PR #4415, opened 2026-01-11
Fix: propagate request context on allocation

### LINKED ISSUE #4408: Request context not propagated to allocator in allocation service
**What happened**:

When operating many GameServers in a single cluster, allocation latency can occur. In this situation, if the client sending the allocation request terminates due to a timeout, the agones-allocator does not stop the allocation process. Since requests are put into a queue and processed, once timeouts start piling up, the situation gets worse - the allocator continues processing requests that clients have already given up on, causing a cascading effect.

When a client cancels an allocation request (due to timeout or explicit cancellation):
  - The request should not be added to the queue, OR
  - If already in the queue, the allocation logic should be skipped when processing it

  1. Deploy Agones with many GameServers (enough to cause allocation latency)
  2. Send multiple allocation requests with short timeouts
  3. Observe that the allocator continues processing requests even after clients have timed out
  4. Notice that timed-out requests accumulate in the queue, worsening the overall latency

 Root cause: In cmd/allocator/main.go, the gRPC request context is not passed to the allocator in the standard (non-ProcessorAllocator) path. The workerCtx (server lifecycle context) is captured in a closure instead of using the per-request context.

```
  // line 521-523 in newServiceHandler
  allocationCallback: func(gsa *allocationv1.GameServerAllocation) (k8sruntime.Object, error) {
      return allocator.Allocate(ctx, gsa)  // ctx here is workerCtx, not request ctx
  },

  // line 755 in Allocate method
  resultObj, err := h.allocationCallback(gsa)  // request ctx is not passed
```
  `Note: The ProcessorAllocator path correctly passes the request context (line 742), but the standard path does not.`

**Environment**:
- Agones version: 1.45.0 (but it's the same as main also)
- Kubernetes version (use `kubectl version`): 1.31
- Install method (yaml/helm):
- Others:

lacroixthomas:
**What type of PR is this?**
/kind bug

Added the request context to the allocator allocation instead of the main context

Closes #4408 

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -29,7 +29,7 @@ RUN gem install fpm && \
 # install go
 WORKDIR /usr/local
 ENV GO_VERSION=1.25.0
-ENV GOPATH /go
+ENV GOPATH=/go
 ENV GO111MODULE=on
 RUN wget -q https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
     tar -xzf go${GO_VERSION}.linux-amd64.tar.gz && rm go${GO_VERSION}.linux-amd64.tar.gz && mkdir ${GOPATH}
@@ -45,14 +45,14 @@ RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.c
     echo "source /usr/share/google-cloud-sdk/completion.bash.inc" >> /root/.bashrc
 
 # update the path for go
-ENV PATH /usr/local/go/bin:/go/bin:$PATH
+ENV PATH=/usr/local/go/bin:/go/bin:$PATH
 
 # install go tooling for development, building and testing
 RUN go install golang.org/x/tools/cmd/goimports@latest && \
     go install golang.org/x/pkgsite/cmd/pkgsite@latest
 
 # the kubernetes version for the file
-ENV KUBERNETES_VER 1.33.5
+ENV KUBERNETES_VER=1.33.5
 
 # overwrite kubectl as we want a specific version
 RUN curl -LO https://dl.k8s.io/release/v${KUBERNETES_VER}/bin/linux/amd64/kubectl && \
@@ -65,8 +65,8 @@ RUN mkdir -p /go/src/k8s.io && cd /go/src/k8s.io && \
  git clone -b kubernetes-${KUBERNETES_VER} --depth=3 https://github.com/kubernetes/code-generator.git
 
 # install Helm package manager
-ENV HELM_VER 3.18.4
-ENV HELM_URL https://get.helm.sh/helm-v${HELM_VER}-linux-amd64.tar.gz
+ENV HELM_VER=3.18.4
+ENV HELM_URL=https://get.helm.sh/helm-v${HELM_VER}-linux-amd64.tar.gz
 RUN curl -L  ${HELM_URL} > /tmp/helm.tar.gz \
     && tar -zxvf /tmp/helm.tar.gz -C /tmp \
     && mv /tmp/linux-amd64/helm /usr/local/bin/helm \
@@ -96,7 +96,7 @@ RUN curl -s https://get.extism.org/cli | sh -s -- -v v1.6.2 -y && \
 #     \_/\_/ \___|_.__/|___/_|\__\___|
 #
 
-ENV HUGO_VER 0.152.2
+ENV HUGO_VER=0.152.2
 RUN mkdir /tmp/hugo && \
     wget -q -O /tmp/hugo/hugo.tar.gz https://github.com/gohugoio/hugo/releases/download/v${HUGO_VER}/hugo_extended_${HUGO_VER}_Linux-64bit.tar.gz && \
     tar -zxvf /tmp/hugo/hugo.tar.gz -C /tmp/hugo/ && \


lacroixthomas:
Keeping it in draft, will check if this could be cleaner and also need to add unit tests around it

pkg/gameserverallocations/allocator.go
```diff
@@ -459,7 +460,13 @@ func (c *Allocator) getClientCertificates(namespace, secretName string) (clientC
...
 	// creates an allocation request. This contains the requested GameServerAllocation, as well as the
 	// channel we expect the return values to come back for this GameServerAllocation
-	req := request{gsa: gsa, response: make(chan response)}
+	req := request{gsa: gsa, response: make(chan response), ctx: ctx}
+
+	// Avoid pushing request to the queue if the context is already done to prevent
+	// the pendingRequests reaching the max buffer size and blocking further allocations
+	if ctx.Err() != nil {
+		return nil, ErrTotalTimeoutExceeded
+	}
```
lacroixthomas, line 469: Might be overkill 🤔

markmandel:
@Reasonably - just checking if you had a chance to have a look at this implementation?

pkg/gameserverallocations/allocator.go
```diff
@@ -517,6 +524,12 @@ func (c *Allocator) ListenAndAllocate(ctx context.Context, updateWorkerCount int
 	for {
 		select {
 		case req := <-c.pendingRequests:
+			// Ensure the request context is still active before processing to avoid allocating unnecessarily
+			if req.ctx.Err() != nil {
+				req.response <- response{request: req, gs: nil, err: ErrTotalTimeoutExceeded}
```
Reasonably: If `allocate()` already returned via `ctx.Done()`, nobody is reading `req.response`. Since the channel is unbuffered, wouldn't this block forever?                                                                                                                                                                                                                         


Should this be a non-blocking send? Or how about using a buffered channel?

```go                                                                                                                                                                                                                                                                                                                                                                             
  if req.ctx.Err() != nil {                                                                                                                                                                                                                                                                                                                                                  
      select {                                                                                                                                                                                                                                                                                                                                                               
      case req.response <- response{request: req, gs: nil, err: ErrTotalTimeoutExceeded}:                                                                                                                                                                                                                                                                                    
      default:                                                                                                                                                                                                                                                                                                                                                               
      }                                                                                                                                                                                                                                                                                                                                                                      
      continue                                                                                                                                                                                                                                                                                                                                                               
  }

Comment (lacroixthomas) [reply pkg/gameserverallocations/allocator.go]: Oh that's a good point, nice catch ! - will try to have a look at that later this week

**Comment by lacroixthomas** [comment]
Didn't forget this one - will take some time tomorrow evening to address the comment, been quite busy recently

**Comment by lacroixthomas** [comment]
/gcbrun

Will double check the error, ive seen some timeout / cert issue maybe ? (On my phone, will look deeper once on my laptop)

File: pkg/gameserverallocations/allocator.go (hunk 2 above)
Comment (Reasonably) [reply pkg/gameserverallocations/allocator.go]: I checked the commit that changed to use a buffered channel.   
LGTM 👍

**Comment by lacroixthomas** [comment]
/gcbrun 
```
            	Test:       	TestControllerCreateGameServerPod/forbidden_pods_creation
            	Error:      	Max difference between 2026-02-01 00:21:32.000059604 +0000 UTC m=+12.717592485 and 2026-02-01 00:21:31 +0000 UTC allowed is 1s, but difference was 1.000059604s
            	Error Trace:	/go/src/agones.dev/agones/pkg/gameservers/controller_test.go:1437
 ```

Sadness, so close... 😄 0.000059604s too late unlucky

lacroixthomas:
Only failing for `gke-autopilot-e2e-test-cluster-1-34` : `TestAllocatorCrossNamespace` `ERROR rerun aborted because previous run had a suspected panic and some test may not have run` 
/gcbrun

Reasonably:
Are there any other actions I need to take besides approving with LGTM?

markmandel:
I'm at GDC but lemme know if this is ready for review 👍🏻

@Reasonably but would definitely like your LGTM before I review 👍🏻

markmandel:
@Reasonably just checking if you had a chance to review - would definitely save me some time 😄

Reasonably:
> @Reasonably just checking if you had a chance to review - would definitely save me some time 😄

LGTM 😄

pkg/gameserverallocations/allocator_test.go
```diff
@@ -627,6 +627,42 @@ func TestAllocatorRunLocalAllocations(t *testing.T) {
...
+			Spec: allocationv1.GameServerAllocationSpec{
+				Selectors: []allocationv1.GameServerSelector{{LabelSelector: metav1.LabelSelector{MatchLabels: map[string]string{agonesv1.FleetNameLabel: "thereisnofleet"}}}},
+			}}
+		gsa.ApplyDefaults()
+		errs := gsa.Validate()
+		require.Len(t, errs, 0)
+
+		reqCtx, reqCancel := context.WithTimeout(context.Background(), time.Millisecond)
+		reqCancel()
+		j1 := request{gsa: gsa.DeepCopy(), response: make(chan response), ctx: reqCtx}
```
markmandel:
```suggestion
		j1 := request{gsa: gsa.DeepCopy(), response: make(chan response, 1), ctx: reqCtx}
```

Since the implementation uses a buffered channel, we should here too.

Since tryRespondWithError does a non-blocking send. If the ListenAndAllocate goroutine happens to process the request before the main goroutine reaches `res1 :=  <-j1.response`, the non-blocking send sees no reader, goes to default:, and the test hangs forever on the receive.

In practicethe goroutine scheduling probably saves you here, but it's a race. `It should be make(chan response`, 1) to be consistent with the actual `allocate()` implementation and to make the test deterministically safe.

Otherwise, LGTM!

pkg/gameserverallocations/allocator_test.go
```diff
@@ -561,11 +561,11 @@ func TestAllocatorRunLocalAllocations(t *testing.T) {
 		require.Len(t, errs, 0)
 
 		// line up 3 in a batch
-		j1 := request{gsa: gsa.DeepCopy(), response: make(chan response)}
+		j1 := request{gsa: gsa.DeepCopy(), response: make(chan response), ctx: context.Background()}
```
markmandel: I saw there were changes, but shouldn't these all these channel creations in this test be `make(chan response, 1)` to match what happens in the implementing code?

```suggestion
		j1 := request{gsa: gsa.DeepCopy(), response: make(chan response, 1), ctx: context.Background()}
```

lacroixthomas: Oh I thought i updated them at the same time, might have dis something wrong with my rebase/squash, will update them as well

markmandel:
/gcbrun


This is all autopilot flakiness

PR #4416, opened 2026-01-13
test: standardize runtime.ParseFeatures error handling and mutex usage

### LINKED ISSUE #4407: Flaky:  TestControllerAllocator/successful_allocation
Another flaky Unit test:

```
--- FAIL: TestControllerAllocator (1.29s)
    --- FAIL: TestControllerAllocator/successful_allocation (1.25s)
        controller_test.go:117: 
            	Error Trace:	/go/src/agones.dev/agones/pkg/gameserverallocations/controller_test.go:117
            	            				/go/src/agones.dev/agones/pkg/gameserverallocations/controller_test.go:132
            	Error:      	Not equal: 
            	            	expected: 201
            	            	actual  : 422
            	Test:       	TestControllerAllocator/successful_allocation
```

https://console.cloud.google.com/cloud-build/builds/0fdbb16c-b47c-4589-8437-791600a3155d;step=19?invt=AcGWmA&project=agones-images

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Ensure all calls to runtime.ParseFeatures in unit tests follow consistent patterns:

- Wrap all calls with require.NoError instead of assert.NoError to fail fast on parse errors

- Add runtime.FeatureTestMutex locks with defer unlocks to all tests/subtests using ParseFeatures

- Fix mutex scope in pkg/sdkserver/sdk_test.go by moving locks from parent test to individual subtests

This ensures thread-safe feature flag manipulation and prevents test flakiness from concurrent feature state modifications.

This also includes some (probable) fixes for flaky tests, particularly closes #4407 as part of this work.

🤞🏻

### DIFF
--- a/site/content/en/docs/Advanced/out-of-cluster-dev-server.md
+++ b/site/content/en/docs/Advanced/out-of-cluster-dev-server.md
@@ -79,7 +79,7 @@ Obviously, this document does not know what every developer's specific network c
 The developer will need to figure out which steps are necessary for their specific configuration.
 
 If attempting to connect via the internet, the developer needs to set the `GameServer`'s `metadata:annotations:agones.dev/dev-address` field to their public IP.
-This can be found by going to [whatsmyip.org](https://www.whatsmyip.org/) or [whatismyip.com](https://www.whatismyip.com/) in a web browser.
+This can be found by going to [whatsmyip.org](https://www.whatsmyip.org/) or <a href="https://www.whatismyip.com/" data-proofer-ignore>whatismyip.com</a> in a web browser.
 
 The  `GameServer`'s `spec:ports:hostPort`/`spec:ports:containerPort` should be set to whichever port the game server binary's logic will bind to -- the port used by `simple-game-server` is 7654 (by default).
 The local network's router must also be configured to forward this port to the desired machine; allowing inbound external requests (from the internet) to be directed to the machine on the network that is running the game server.


pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -619,10 +620,9 @@ func TestApplyWebhookPolicyWithMetadata(t *testing.T) {
 	utilruntime.FeatureTestMutex.Lock()
 	defer utilruntime.FeatureTestMutex.Unlock()
 
-	err := utilruntime.ParseFeatures(string(utilruntime.FeatureFleetAutoscaleRequestMetaData) + "=true")
-	assert.NoError(t, err)
+	require.NoError(t, utilruntime.ParseFeatures(string(utilruntime.FeatureFleetAutoscaleRequestMetaData)+"=true"))
 	defer func() {
-		_ = utilruntime.ParseFeatures("")
+		require.NoError(t, utilruntime.ParseFeatures(""))
```
igooch: Is this needed? None of the other tests use a cleanup function for the ParseFeatures.

If we do need the cleanup function, should this be assert instead of require? The require inside of a defer cleanup function might mask a primary failure elsewhere.

pkg/gameserverallocations/controller_test.go
```diff
@@ -59,6 +60,9 @@ const (
 func TestControllerAllocator(t *testing.T) {
 	t.Parallel()
 
+	runtime.FeatureTestMutex.Lock()
+	defer runtime.FeatureTestMutex.Unlock()
+
```
igooch, line 65: The mutex forces this test to run serially as it waits for the global lock, even though there's t.Parallel, correct?

As an alternative, should we remove `t.Parallel` from all tests that change the state with `ParseFeatures` or otherwise, and run all other tests that do not need to change the state with `t.Parallel`? That way all the tests that need to lock run serially, and all remaining tests are run in parallel [with and only with other Parallel tests](https://pkg.go.dev/testing#T.Parallel).

pkg/gameserverallocations/controller_test.go
```diff
@@ -114,7 +116,9 @@ func TestControllerAllocator(t *testing.T) {
 			rec := httptest.NewRecorder()
 			err = c.processAllocationRequest(ctx, rec, r, "default")
 			require.NoError(t, err)
-			require.Equal(t, http.StatusCreated, rec.Code)
+			if !assert.Equal(t, http.StatusCreated, rec.Code, "Unexpected HTTP Code") {
+				require.Failf(t, "Unexpected HTTP Code Body: %s", rec.Body.String())
+			}
```
igooch: Nit:
```suggestion
require.Equal(t, http.StatusCreated, rec.Code, "Response body: %s", rec.Body.String())
```

pkg/gameserverallocations/controller_test.go
markmandel: The benefit of doing it this way is that other non-mutex'd tests can run in parallel with this one - so at least we can get some speed benefits of those running in parallel together.

This also gives us some testing of changes across unit tests of different feature gates -- but it can lead to flaky tests -- but it can find some potential issues around the way.

pkg/gameserverallocations/controller_test.go
markmandel: Oh yeah - that makes _way_ more sense. 🤦🏻

pkg/fleetautoscalers/fleetautoscalers_test.go
markmandel: Agreed - it was there when I was refactoring. Happy to remove.

pkg/gameserverallocations/controller_test.go
igooch: It would be interesting to see what the speed difference would be across all e2e tests if we made the change. Not needed for this PR though :)

markmandel:
On mobile atm, but there's a link that's not working anymore.

markmandel:
On mobile atm, but there's a link that's not working anymore.

PR #4417, opened 2026-01-13
fix(sdk): add validation to SetAnnotation to prevent infinite retry loop

### LINKED ISSUE #4409: fix(sdk): validate labels in SetAnnotation to prevent retry loops
This is the same as #4385 but for SetAnnotations vs SetLabel

When attempting to submit an invalid Annotation with the REST API. rather than returning a failure response the client SDK will enter an infinite loop of attempting to submit the label without any additional output unless debug logging is enabled, this also disables further creation of labels.

Simple failure responses in the form of 400 or 500 response codes with a simple JSON body stating success or failure along with the attempted Annotation key and value.

I did briefly consider monitoring the labels but considering the loop knocks out the set annotation feature it would be a bit moot, could also force restart the client sdk sidecar somehow but I'm not sure if that will break anything upstream when considering health checking.

indurireddy-TF:
**What type of PR is this?**
>
> /kind breaking

 /kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Closes #4409

### DIFF
--- a/pkg/sdkserver/sdkserver.go
+++ b/pkg/sdkserver/sdkserver.go
@@ -623,7 +623,8 @@ func (s *SDKServer) SetAnnotation(_ context.Context, kv *sdk.KeyValue) (*sdk.Emp
 	if kv == nil {
 		return nil, status.Error(codes.InvalidArgument, "annotation key/value cannot be nil")
 	}
-
+	// TODO(k8s-1.35): Replace validation.IsQualifiedName with the new
+	// qualified name validation function once we bump apimachinery to v0.35+
 	if errs := validation.IsQualifiedName(kv.Key); len(errs) > 0 {
 		return nil, status.Errorf(
 			codes.InvalidArgument,


pkg/sdkserver/sdkserver.go
```diff
@@ -620,6 +620,18 @@ func (s *SDKServer) SetLabel(_ context.Context, kv *sdk.KeyValue) (*sdk.Empty, e
 func (s *SDKServer) SetAnnotation(_ context.Context, kv *sdk.KeyValue) (*sdk.Empty, error) {
 	s.logger.WithField("values", kv).Debug("Adding SetAnnotation to queue")
 
+	if kv == nil {
+		return nil, status.Error(codes.InvalidArgument, "annotation key/value cannot be nil")
+	}
+
+	if errs := validation.IsQualifiedName(kv.Key); len(errs) > 0 {
```
igooch, line 628: FYI: `validation.IsQualifiedName` will be deprecated in k8s.io/apimachinery v0.35. We are currently on v0.33 to align with our middle supported version of Kubernetes 1.33. We should migrate to the new function once our middle supported version shifts to 1.35, as the new function is not available until [v0.35](https://pkg.go.dev/k8s.io/apimachinery@v0.35.0/pkg/api/validate/content).

igooch (commented):
LGTM. Waiting to officially approve and merge until after today's release.

pkg/sdkserver/sdkserver.go
indurireddy-TF: @igooch  Added TODO comments for post–Kubernetes 0.35 migration and will create a new issue to track this.

PR #4426, opened 2026-01-17
chore(sdk/nodejs): migrate linting from ESLint to Biome

### LINKED ISSUE #4425: Move Node SDK linting to `biome`
Move Node linting to "biome" - https://www.npmjs.com/package/@biomejs/biome

> This looks good but apparently these stylistic rules are no longer recommended and may not work. If `npm run lint` works then all is good for now
> 
> I recommend moving to biome as we can just use the standard rules with no overrides with the consequence quotes will change to double but this could be a separate task 

 _Originally posted by @steven-supersolid in [#4412](https://github.com/googleforgames/agones/pull/4412/changes#r2681816888)_

So there may be some tweaks to the SDk as well to align with the new style guides.

To test if the changes work as expected locally:  ` make run-sdk-conformance-test-node` is what you are looking for.

smazmi:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Migrates the Node.js SDK linting from ESLint to [Biome](https://biomejs.dev/), using Biome's standard recommended rules with no custom overrides.

Changes include:
- Replaced `eslint` and `globals` dependencies with `@biomejs/biome`
- Updated `lint` script from `eslint --fix` to `biome check --write`
- Applied Biome's default formatting (double quotes, consistent spacing)
- Fixed lint issues to comply with Biome's stricter rules:
  - Prefixed unused callback parameters with `_`
  - Replaced `forEach` with `for...of` to avoid implicit return values
  - Replaced TypeScript `any` types with `unknown`

Closes #4425

- Conformance tests pass: `make run-sdk-conformance-test-node`
- Unit tests pass: `npm test` (78 specs, 0 failures)
- This follows the suggestion in #4412 to use Biome's standard rules

### DIFF
--- a/sdks/nodejs/biome.json
+++ b/sdks/nodejs/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.3.11/schema.json",
+	"$schema": "https://biomejs.dev/schemas/2.3.11/schema.json",
 	"formatter": {
 		"enabled": true,
 		"indentStyle": "tab"


sdks/nodejs/biome.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://biomejs.dev/schemas/2.3.11/schema.json",
```
steven-supersolid: It looks like this line is indented with spaces and others with tabs

steven-supersolid:
Everything else looks great

sdks/nodejs/biome.json
smazmi: My bad. I added this from my work machine, and the IDE there has different tab/space rules. I’ll fix it.

PR #4434, opened 2026-01-26
SuperTuxKart example: Mark gameserver as allocated once the first player connected

D3luxee:
**What type of PR is this?**
/kind feature

This PR implements logic into the supertuxkart example to mark the gameserver as allocated once the first player connected to it.

### DIFF
--- a/examples/supertuxkart/main.go
+++ b/examples/supertuxkart/main.go
@@ -106,6 +106,7 @@ func main() {
 		}
 	}
 	defer t.Cleanup()
+	var isAllocated bool
 	for line := range t.Lines {
 		// Don't use the logger here. This would add multiple prefixes to the logs. We just want
 		// to show the supertuxkart logs as they are, and layer the wrapper logs in with them.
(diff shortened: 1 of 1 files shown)


markmandel:
_comes out 🤧 land_

🤔 how much do we want to be promoting this as a pattern 🤔 

Feels like it's has high potential for race conditions where a player connects, and while connecting the server could get shutdown.

markmandel:
What do you think about this @lacroixthomas I'm tempted to reject, as I don't think it's a pattern we really want to push?

lacroixthomas:
I agree that it could bring some race conditions, I know that it's in the example but still, I would probably reject it as well

According to the SDK documentation (https://agones.dev/site/docs/guides/client-sdks/#allocate ), GameServerAllocation is the preferred approach as `it gives Agones control over how packed GameServers are scheduled within a cluster, whereas with Allocate() you relinquish control to an external service`

The docs says that the SDK.Allocate() should mainly be used for the matchmaker registration pattern (https://agones.dev/site/docs/integration-patterns/matchmaker-registration/) where an external matchmaker requires servers to self register

I don't think we want to push this pattern in the examples either (also that it currently brings race conditions) 🤔

mikeseese:
I agree that this (calling `s.Allocate()` instead of creating a `GameServerAllocation`) falls under "advanced/specific usage" and not something to be promoted in an example. Further, generally speaking, players joining is not the correct time to mark a server as `Allocated`; if anything players shouldn't be able to join until after it's marked `Allocated`. See the other [SuperTuxKart docs](https://agones.dev/site/docs/examples/supertuxkart/#connect-to-the-game-server):

> After allocating a GameServer from the fleet and obtaining its status and IP, you’re ready to connect and play.

I agree that this PR should be rejected. Thanks for the PR @D3luxee though!

markmandel:
I'm going to close this just to clean things up - but appreciate the PR. If you want to continue the conversation on an issue, please feel free to open one 👍🏻

PR #4448, opened 2026-02-14
Fix Prometheus installation with helm chart v28.9.0.

### LINKED ISSUE #4436: Metrics Prometheus install with agones's yaml fails on current prometheus-community helm chart
**Link to page**: 
  /site/docs/guides/metrics/
  
  **Description of documentation issue**:
Starting with prometheus-community helm chart version 28.0.0, scrape configs are changed to map. Because of that current agones's prometheus.yaml is outdated and fails to install. I added commit link from prometheus-community helm chart repo below for reference.
https://github.com/prometheus-community/helm-charts/commit/f80e884cd56e920c5322d8be0b85b6de093bc547
  
  **What improvements to documentation would you like to see?**:
Either prometheus.yaml file should be updated, or helm command in docs should include --version 27.52.0

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Pin chart version and rename scrape jobs to avoid conflicts.

Also updated documentation on which version is currently tested.

Closes #4436

N/A

### DIFF
--- a/build/prometheus.yaml
+++ b/build/prometheus.yaml
@@ -29,26 +29,21 @@ serverFiles:
   prometheus.yml:
     rule_files: [/etc/config/rules, /etc/config/alerts]
     scrape_configs:
-      - job_name: prometheus
-        static_configs:
-          - targets:
-              - localhost:9090
-
       # A scrape configuration for running Prometheus on a Kubernetes cluster.
       # This uses separate scrape configs for cluster components (i.e. API server, node)
       # and services to allow each to use different authentication configs.
-      #
+
       # Kubernetes labels will be added as Prometheus labels on metrics via the
       # `labelmap` relabeling action.
 
       # Scrape config for API servers.
-      #
+
       # Kubernetes exposes API servers as endpoints to the default/kubernetes
       # service so this uses `endpoints` role and uses relabelling to only keep
       # the endpoints associated with the default/kubernetes service using the
       # default named port `https`. This works for single API server deployments as
       # well as HA API server deployments.
-      - job_name: kubernetes-apiservers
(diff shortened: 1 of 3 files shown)


build/prometheus.yaml
```diff
@@ -29,26 +29,21 @@ serverFiles:
   prometheus.yml:
     rule_files: [/etc/config/rules, /etc/config/alerts]
     scrape_configs:
-      - job_name: prometheus
-        static_configs:
```
markmandel, line 33: This section is duplicated from the default config, so taking it out as it causes issues if duplicated.

build/prometheus.yaml
```diff
@@ -29,26 +29,21 @@ serverFiles:
...
       # Scrape config for API servers.
-      #
+
       # Kubernetes exposes API servers as endpoints to the default/kubernetes
       # service so this uses `endpoints` role and uses relabelling to only keep
       # the endpoints associated with the default/kubernetes service using the
       # default named port `https`. This works for single API server deployments as
       # well as HA API server deployments.
-      - job_name: kubernetes-apiservers
+      - job_name: kubernetes-apiservers-agones
```
lacroixthomas, line 46: Should we tag this as a "breaking change" (maybe not as much, but maybe a note or similar), because it might affect users with existing custom Prometheus rules, alerts, or dashboards relying on the old job names (it's actually from the ./build/Makefile, but maybe for people who are testing it ? Not sure about it, WDYT 🤔 

Otherwise LGTM

markmandel:
Good call - tagged `kind/breaking`

PR #4469, opened 2026-03-12
fix: duplicate GameServerSets not being cleaned up under a Fleet

### LINKED ISSUE #4468: Fleet controller does not clean up duplicate GameServerSets with matching spec
**What happened**:
- I updated fleets yesterday, Some fleets goes up desired, When I scan the cluster, there are 2 valid GameServerSet in a Fleet.
- A cause of explosion of desired is faults that our custom webhook scaler has a safe scale-in logic seeing current replicas(cut off if target desired too lower than current, cut it)
- But 2 gss is still a problem. 
  -  1 is valid and 1 is orphaned(never deleted)

- If there are many GSS that meet fleet spec, the controller deletes old GSS even if they meet the spec of the fleet.

- I don't know how to reproduce it correctly, but [code](https://github.com/googleforgames/agones/blob/main/pkg/fleets/controller.go#L733-L746) shows the cause of the issue.
- It overrides if there are many matched GSS, and the overridden GSS never goes to rest.

Image

- A, B goes up because our webhook autoscaler has bugs(it sees current replicas of fleet) 
- Each reconcile loop don't guarantee sequence of gss. It can be applied to all spec-valid gss.

**Environment**:
- Agones version: v1.45.0
- Kubernetes version (use `kubectl version`): 1.33.6
- Cloud provider or hardware configuration: azure
- Others:

mnthe:
**What type of PR is this?**

/kind bug

Fixes a bug in `filterGameServerSetByActive()` where duplicate GameServerSets sharing the same template as the Fleet are silently dropped — placed in neither the `active` nor the `rest` slice.

When multiple GameServerSets match the Fleet's template, the current code overwrites the `active` variable on each match (last-write-wins). Earlier matches are lost entirely: they don't go into `rest`, so `applyDeploymentStrategy()` and `deleteEmptyGameServerSets()` never touch them.

This causes two problems:
1. **Orphaned GameServerSets** that are never scaled down or deleted.
2. **All duplicates converge to the Fleet's desired replica count**,
   - because the Kubernetes informer store iterates a Go map with non-deterministic ordering — each `syncFleet()` can pick a different GSS as "active" and update its `Spec.Replicas`.

**Screenshot — 3 duplicate GameServerSets under a single Fleet, all with the same Desired value:**

무제

The fix keeps the oldest GameServerSet (by `CreationTimestamp`) as active and places any duplicates into the `rest` slice, where the existing deployment strategy logic naturally scales them down and deletes them.

Closes #4468

- Root cause is in `filterGameServerSetByActive()` (`pkg/fleets/controller.go`), which has had last-write-wins behavior since its initial implementation and has never been modified.
- Added `TestControllerFilterGameServerSetByActiveDuplicates` covering:
  - 3 duplicate-template GSS → oldest selected as active, others in rest
  - Non-deterministic list ordering → same deterministic result
  - Mix of matching and non-matching templates
- All existing tests pass (`make test-go`).
- Minimal change: 17 lines added to the function, 55 lines of new tests.

**Does this PR introduce a user-facing change?**

```release-note
Fixed a bug where duplicate GameServerSets with the same template could
accumulate under a Fleet without being cleaned up, causing replicas to
multiply beyond the desired count.
```

### DIFF
--- a/pkg/fleets/controller.go
+++ b/pkg/fleets/controller.go
@@ -729,14 +729,29 @@ func (c *Controller) updateFleetStatus(ctx context.Context, fleet *agonesv1.Flee
 
 // filterGameServerSetByActive returns the active GameServerSet (or nil if it
 // doesn't exist) and then the rest of the GameServerSets that are controlled
-// by this Fleet
+// by this Fleet.
+// If multiple GameServerSets match the Fleet's template (e.g. due to informer
+// cache lag during creation), the oldest one by CreationTimestamp is kept as
+// active and the duplicates are placed in rest so they can be scaled down and
+// cleaned up by the normal deployment strategy.
 func (c *Controller) filterGameServerSetByActive(fleet *agonesv1.Fleet, list []*agonesv1.GameServerSet) (*agonesv1.GameServerSet, []*agonesv1.GameServerSet) {
 	var active *agonesv1.GameServerSet
 	var rest []*agonesv1.GameServerSet
 
 	for _, gsSet := range list {
 		if apiequality.Semantic.DeepEqual(gsSet.Spec.Template, fleet.Spec.Template) {
-			active = gsSet
+			if active == nil {
+				active = gsSet
+			} else {
+				// Multiple GameServerSets match the fleet template.
+				// Keep the oldest as active; treat the newer duplicate as rest.
+				if gsSet.ObjectMeta.CreationTimestamp.Before(&active.ObjectMeta.CreationTimestamp) {
+					rest = append(rest, active)
+					active = gsSet
+				} else {
+					rest = append(rest, gsSet)
+				}
+			}
 		} else {
 			rest = append(rest, gsSet)
 		}
--- a/pkg/fleets/controller_test.go
+++ b/pkg/fleets/controller_test.go
@@ -991,6 +991,61 @@ func TestControllerFilterGameServerSetByActive(t *testing.T) {
 	assert.Equal(t, []*agonesv1.GameServerSet{gsSet1, gsSet2}, rest)
 }
 
+func TestControllerFilterGameServerSetByActiveDuplicates(t *testing.T) {
+	t.Parallel()
+
+	f := defaultFixture()
+	c, _ := newFakeController()
+
+	now := time.Now()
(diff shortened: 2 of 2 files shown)


markmandel:
Sorry for the delay - GDC + kubecon + cncf migration is sucking time. Will definitely look.

pkg/fleets/controller.go
```diff
@@ -729,14 +729,29 @@ func (c *Controller) updateFleetStatus(ctx context.Context, fleet *agonesv1.Flee
...
 	for _, gsSet := range list {
 		if apiequality.Semantic.DeepEqual(gsSet.Spec.Template, fleet.Spec.Template) {
-			active = gsSet
+			if active == nil {
+				active = gsSet
+			} else {
+				// Multiple GameServerSets match the fleet template.
+				// Keep the oldest as active; treat the newer duplicate as rest.
+				if gsSet.ObjectMeta.CreationTimestamp.Before(&active.ObjectMeta.CreationTimestamp) {
+					rest = append(rest, active)
```
markmandel, line 749: This is a really elegant fix! If it does occur, then the old duplicate gsset ends up being treated like a rolling/replace update, and is slowly removed over time, and we end up in a good state in an eventually consistent manner -- which is perfect for Kubernetes!

markmandel:
```
Error: UPGRADE FAILED: release agones failed, and has been rolled back due to atomic being set: cannot patch "agones-controller-pdb" with kind PodDisruptionBudget: an error on the server ("Internal Server Error: \"/apis/policy/v1/namespaces/agones-system/poddisruptionbudgets/agones-controller-pdb?fieldManager=helm\": the server is currently unable to handle the request") has prevented the request from succeeding (patch poddisruptionbudgets.policy agones-controller-pdb)
make: *** [Makefile:451: install] Error 1

```

That's new. Let's try some percussive maintenance.

/gcbrun

markmandel:
https://gopkg.in is being flaky.

mnthe:
Thank you for the quick review! I appreciate you taking the time despite your busy CNCF schedule. :D

markmandel:
Thanks for the great fix!

PR #4488, opened 2026-03-27
Upgrade supported Kubernetes versions to 1.33, 1.34, 1.35

Sivasankaran25:
**What type of PR is this?**
>

/kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Works On #4487

### DIFF
--- a/test/e2e/gameserver_test.go
+++ b/test/e2e/gameserver_test.go
@@ -538,7 +538,7 @@ func TestGameServerPodCompletedAfterCleanExit(t *testing.T) {
 		}
 	}()
 
-	err = wait.PollUntilContextTimeout(ctx, 3*time.Second, 3*time.Minute, true, func(ctx context.Context) (bool, error) {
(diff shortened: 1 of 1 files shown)


examples/cpp-simple/fleet.yaml
```diff
@@ -30,5 +30,5 @@ spec:
         spec:
           containers:
             - name: cpp-simple
-              image: us-docker.pkg.dev/agones-images/examples/cpp-simple-server:0.24
+              image: us-docker.pkg.dev/agones-images/examples/cpp-simple-server:0.26
```
igooch, line 33: Double checking that this is the correct image? Looks like there was an image published ~10 hours ago with tag 0.26, and ~45 min ago with tag 0.25.

Sivasankaran25: Image 0.26 is the supported version; 0.25 is not required and can be ignored

markmandel:
Wow that was fast!!! Nice work!

PR #4513, opened 2026-04-14
feat: track allocations per fleet

### LINKED ISSUE #4452: [Feature Proposal] - Total Allocations Tracking
At no point does Agones track or expose the number of allocations (GameServers moving into `Allocated` state). This however is an extremely useful metric in general, but also specifically for autoscaling.

Add `TotalAllocations` to the `FleetStatus`, which is a `uint64` counter of the total number of allocations observed over time. This should not update the `Fleet` for every allocation observed, but rather rate limit changes to the `Fleet` for this counter to 1s. A prom metric should also be exposed with this data.

The `TotalAllocations` field should be provided in the `FleetAutoscaleRequest` to webhooks and WASM.

This allows users to reason about the number of allocations on a fleet and react to this data in a custom autoscaler, which should allow for better buffer size optimisation.

An example of the new `FleetStatus`:
```
type FleetStatus struct {
	Replicas int32 `json:"replicas"`
	ReadyReplicas int32 `json:"readyReplicas"`
	ReservedReplicas int32 `json:"reservedReplicas"`
	AllocatedReplicas int32 `json:"allocatedReplicas"`
	TotalAllocations uint64 `json:"totalAllocations"`
	Players *AggregatedPlayerStatus `json:"players,omitempty"`
	Counters map[string]AggregatedCounterStatus `json:"counters,omitempty"`
	Lists map[string]AggregatedListStatus `json:"lists,omitempty"`
}
```

It is possible to do this outside of Agones in a custom controller or webhook, and then inject the data where needed.

**Additional context**

The link to the AFP PR.

**Discussion Link (if any)**

nrwiersma:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
/kind feature
> /kind hotfix
> /kind release

This PR adds allocations tracking on a Fleet level to provide real allocation metrics to the Fleet Autoscaler. This provides better metrics to base custom autoscaling off of while also providing a good metric to the system

Closes #4452

Initially I thought to use a `RWMutex` and `atomic.Int64`, but the locking logic got really verbose and I dont think, even in active clusters, that the contention on a `Mutex` is likely to be a real issue, given how low the actual lock time will be. If this proves to no longer be true, it is simple to change it.

It also occurred to me that queuing the Fleet when an Allocation is observed is likely not needed, as this will always result in either a change in AllocatedReplicas, ReadyReplicas or Replicas on the `GameServerSet` which will itself queue the `Fleet`. I will run some manual tests to prove this.

### DIFF
--- a/pkg/fleets/controller.go
+++ b/pkg/fleets/controller.go
@@ -72,9 +72,9 @@ type Controller struct {
 	fleetLister         listerv1.FleetLister
 	fleetSynced         cache.InformerSynced
 	workerqueue         *workerqueue.WorkerQueue
-	allocsMu            sync.Mutex
-	allocs              map[string]int64
 	recorder            record.EventRecorder
+	allocs              map[string]int64
+	allocsMu            sync.Mutex
 }
 
 // NewController returns a new fleets crd controller
--- a/pkg/fleets/controller.go
+++ b/pkg/fleets/controller.go
@@ -18,7 +18,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"agones.dev/agones/pkg/apis/agones"
@@ -71,10 +73,9 @@ type Controller struct {
 	fleetGetter         getterv1.FleetsGetter
 	fleetLister         listerv1.FleetLister
 	fleetSynced         cache.InformerSynced
+	allocs              *allocTracker
 	workerqueue         *workerqueue.WorkerQueue
 	recorder            record.EventRecorder
-	allocs              map[string]int64
-	allocsMu            sync.Mutex
 }
 
 // NewController returns a new fleets crd controller
@@ -102,7 +103,7 @@ func NewController(
 		fleetGetter:         agonesClient.AgonesV1(),
 		fleetLister:         fleets.Lister(),
 		fleetSynced:         fInformer.HasSynced,
-		allocs:              map[string]int64{},
+		allocs:              newAllocTracker(),
 	}
 
 	c.baseLogger = runtime.NewLoggerWithType(c)
@@ -122,7 +123,7 @@ func NewController(
 		DeleteFunc: func(obj interface{}) {
 			fleet := obj.(*agonesv1.Fleet)
 
-			c.removeAllocations(fleet.ObjectMeta.Namespace, fleet.ObjectMeta.Name)
+			c.allocs.remove(fleet.ObjectMeta.Namespace, fleet.ObjectMeta.Name)
 		},
 	})
 
@@ -152,7 +153,7 @@ func NewController(
 				return
 			}
 
-			c.incAllocations(newGs.Namespace, fleet)
+			c.allocs.inc(newGs.Namespace, fleet, 1)
 		},
 	})
 
@@ -756,10 +757,18 @@ func (c *Controller) updateFleetStatus(ctx context.Context, fleet *agonesv1.Flee
 		}
 	}
 
-	fCopy.Status.Allocations += c.getAllocations(fleet.ObjectMeta.Namespace, fCopy.ObjectMeta.Name)
+	allocs := c.allocs.get(fleet.ObjectMeta.Namespace, fleet.ObjectMeta.Name)
+	fCopy.Status.Allocations += allocs
 
 	_, err = c.fleetGetter.Fleets(fCopy.ObjectMeta.Namespace).UpdateStatus(ctx, fCopy, metav1.UpdateOptions{})
-	return errors.Wrapf(err, "error updating status of fleet %s", fCopy.ObjectMeta.Name)
+	if err != nil {
+		return errors.Wrapf(err, "error updating status of fleet %s", fCopy.ObjectMeta.Name)
+	}
+
+	// The update was successful, the allocation count must be decremented to reflect this.
+	c.allocs.dec(fleet.ObjectMeta.Namespace, fleet.ObjectMeta.Name, allocs)
+
+	return nil
 }
 
 // filterGameServerSetByActive returns the active GameServerSet (or nil if it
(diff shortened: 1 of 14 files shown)


pkg/metrics/exporter_test.go
```diff
@@ -274,6 +275,24 @@ func setupGameServerPlayerConnect(t *testing.T, ctrl *fakeController) {
...
 }
 
+func setupGameServerAllocation(t *testing.T, ctrl *fakeController) {
+	gs := gameServerWithFleetAndState("test-fleet", agonesv1.GameServerStateCreating)
+	ctrl.gsWatch.Add(gs)
+	gs = gs.DeepCopy()
+	gs.Status.State = agonesv1.GameServerStateAllocated
+	ctrl.gsWatch.Modify(gs)
+
+	require.Eventually(t, func() bool {
```
markmandel: If you use https://pkg.go.dev/github.com/stretchr/testify@v1.11.1/require#EventuallyWithT you can check multiple assertions, and also get better error messages.

site/content/en/docs/Reference/fleetautoscaler.md
```diff
@@ -375,6 +375,8 @@ type FleetStatus struct {
 	ReservedReplicas int32 `json:"reservedReplicas"`
 	// AllocatedReplicas are the number of Allocated GameServer replicas
 	AllocatedReplicas int32 `json:"allocatedReplicas"`
+    // Allocations is a counter of the number of allocations observed.
```
markmandel, line 485: New docs will need feature shortcodes please:
https://agones.dev/site/docs/contribute/documentation-editing-contribution/

pkg/fleets/controller.go
```diff
@@ -125,6 +137,25 @@ func NewController(
 		},
 	})
 
+	_, _ = gsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
markmandel, line 143: You'll wan tto sync GameServers at, since you are now watching their events:

https://github.com/nrwiersma/agones/blob/30678d321529ca2fd3bdd27f58b8bd2a4b5c8e96/pkg/fleets/controller.go#L255

pkg/fleets/controller.go
```diff
@@ -760,6 +795,63 @@ func (c *Controller) filterGameServerSetByActive(fleet *agonesv1.Fleet, list []*
...
+
+	delete(c.allocs, key)
+}
+
+func (c *Controller) flushAllocations() {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	c.allocsMu.Lock()
+	defer c.allocsMu.Unlock()
```
markmandel: Nice one on making sure it gets processed on shutdown.

Probably doesn't make a huge difference, but holding the whole mutex over the network calls may cause some interesting long term locking over shutdown. Might be worth taking a snapshot - or the very least, moving to a RWLock, to avoid some potential contention.

pkg/fleets/controller.go
```diff
@@ -72,6 +73,8 @@ type Controller struct {
 	fleetSynced         cache.InformerSynced
 	workerqueue         *workerqueue.WorkerQueue
 	recorder            record.EventRecorder
+	allocs              map[string]int64
+	allocsMu            sync.Mutex
```
markmandel: I mention this during flush, but would this make more sense as a RWLock?

pkg/fleets/controller.go
```diff
@@ -760,6 +795,63 @@ func (c *Controller) filterGameServerSetByActive(fleet *agonesv1.Fleet, list []*
...
+
+	c.allocsMu.Lock()
+	defer c.allocsMu.Unlock()
+
+	count, ok := c.allocs[key]
+	if !ok {
+		c.allocs[key] = 1
+		return
+	}
+	c.allocs[key] = count + 1
```
markmandel:
```suggestion
	c.allocs[key]++
```
maps will 0 value here, just as a suggestion

pkg/fleets/controller.go
```diff
@@ -723,6 +756,8 @@ func (c *Controller) updateFleetStatus(ctx context.Context, fleet *agonesv1.Flee
 		}
 	}
 
+	fCopy.Status.Allocations += c.getAllocations(fleet.ObjectMeta.Namespace, fCopy.ObjectMeta.Name)
```
markmandel: If `UpdateStatus` fails, the worker queue re-enqueues the fleet, but on retry the counter is already 0 — those allocations are permanently lost. 

My suggestion would be to capture the delta, attempt the update, and only zero out the counter on success

markmandel (commented):
Thanks for doing this -- sorry it took a bit to get to it.

We'll need some docs at https://agones.dev/site/docs/guides/metrics/ -- but also will need to be feature shortcoded -- https://agones.dev/site/docs/contribute/documentation-editing-contribution/

pkg/fleets/controller.go
nrwiersma: I thought about that. I decided not to in the end as it does not stop the controller from working. We do not handle Add on the EventHandler (just Update), so the sync would basically block for no reason. Will gladly add it if you still want though.

pkg/fleets/controller.go
nrwiersma: It does, discussed this in the PR description. It makes the logic super complex for very little potential gain. Even at 100 allocs/s the contention on the lock would likely be very low. RWLock requires that you check under RLock, then check again after upgrading to Lock, so the code gets very verbose.

pkg/fleets/controller.go
nrwiersma: Good catch, had not seen that failure path.

pkg/fleets/controller.go
markmandel: Yeah that's fair - rescinded!

pkg/fleets/controller.go
markmandel: Yeah, I'd recommend it - I've seen some weirdness when not syncing -- it'll probably mean nothing much, but also means that if other people edit this later on, or do other controller things they aren't left wondering "when do I sync, or not sync" - better to always sync 👍🏻

nrwiersma: Fair. The global rule for me is, if it is in the `processNextItem` loop anywhere, it gets synced. But will add it in.

pkg/fleets/controller.go
```diff
@@ -760,6 +806,107 @@ func (c *Controller) filterGameServerSetByActive(fleet *agonesv1.Fleet, list []*
...
+	maps.Copy(counts, t.counts)
+	t.mu.Unlock()
+
+	for key, count := range counts {
+		if count.Load() == 0 {
+			continue
+		}
+
+		ns, fleetName, _ := cache.SplitMetaNamespaceKey(key)
+		fn(ns, fleetName, count.Load())
```
markmandel:
```suggestion
  v := count.Load()
  if v == 0 {
      continue
  }
  ns, fleetName, _ := cache.SplitMetaNamespaceKey(key)
  fn(ns, fleetName, v)
```

Just to avoid the value potentially changing between invocations.

site/content/en/docs/Reference/fleetautoscaler.md
```diff
@@ -252,6 +253,60 @@ type FleetStatus struct {
...
+	Request  *FleetAutoscaleRequest  `json:"request"`
+	Response *FleetAutoscaleResponse `json:"response"`
+}
+
+type FleetAutoscaleRequest struct {
+	// UID is an identifier for the individual request/response. It allows us to distinguish instances of requests which are
+	// otherwise identical (parallel requests, requests when earlier requests did not modify etc)
+	// The UID is meant to track the round trip (request/response) between the Autoscaler and the WebHook, not the user request.
+	// It is suitable for correlating log entries between the webhook and apiserver, for either auditing or debugging.
+	UID types.UID `json:"uid""`
```
markmandel:
```suggestion
	UID types.UID `json:"uid"`
```

But also, shouldn't this be:


```suggestion
	UID types.UID `json:"string"`
```

?

pkg/metrics/controller_test.go
```diff
@@ -428,6 +428,44 @@ func TestControllerGameServersTotal(t *testing.T) {
...
+	generateGsEvents(15, agonesv1.GameServerStateAllocated, "test", c.gsWatch)
+	generateGsEvents(19, agonesv1.GameServerStateAllocated, "", c.gsWatch)
+
+	expected := 34
+	assert.Eventually(t, func() bool {
+		list, err := c.gameServerLister.GameServers(gs.ObjectMeta.Namespace).List(labels.Everything())
+		if err != nil || list == nil {
+			return false
+		}
+		require.NoError(t, err)
```
markmandel: if you are going to use require.NoError within Eventually - use a https://pkg.go.dev/github.com/stretchr/testify/require#EventuallyWithT

markmandel (commented):
Nice! Getting close!

site/content/en/docs/Reference/fleetautoscaler.md
nrwiersma: I have no idea, that is pre-existing in the docs. Gladly change it if you like.

nrwiersma: wait, no, that is the json name, not its type.

markmandel:
/gcbrun

Sorry - I missed this one for a while!

nrwiersma:
No stress. When you happy to merge, ill update the versions again in the docs.

site/content/en/docs/Reference/fleetautoscaler.md
markmandel: duh yeah, ignore me.

site/content/en/docs/Guides/metrics.md
```diff
@@ -45,6 +45,7 @@ Follow the [Google Cloud Monitoring installation steps](#google-cloud-monitoring
 
 
 
+{{% feature expiryVersion="1.58.0" %}}
```
markmandel: Sorry to make you do this, but yeah - update the version to 1.59

markmandel (commented):
Looks good! Just need to update the version numbers in the docs, and this is good to go!

Thanks!

markmandel:
```
Release "agones" does not exist. Installing it now.
I0527 01:37:06.607611      89 warnings.go:110] "Warning: autopilot-default-resources-mutator:Autopilot updated Deployment agones-system/agones-ping: defaulted unspecified 'cpu' resource for containers [agones-ping] (see http://g.co/gke/autopilot-defaults)."
I0527 01:37:06.640118      89 warnings.go:110] "Warning: autopilot-default-resources-mutator:Autopilot updated Deployment agones-system/agones-controller: defaulted unspecified 'cpu' resource for containers [agones-controller] (see http://g.co/gke/autopilot-defaults)."
I0527 01:37:06.661004      89 warnings.go:110] "Warning: autopilot-default-resources-mutator:Autopilot updated Deployment agones-system/agones-extensions: defaulted unspecified 'cpu' resource for containers [agones-extensions] (see http://g.co/gke/autopilot-defaults)."
I0527 01:37:06.679747      89 warnings.go:110] "Warning: autopilot-default-resources-mutator:Autopilot updated Deployment agones-system/agones-allocator: defaulted unspecified 'cpu' resource for containers [agones-allocator] (see http://g.co/gke/autopilot-defaults)."
I0527 01:37:06.949030      89 warnings.go:110] "Warning: AdmissionWebhookController: mutated namespaceselector of the webhooks to enforce GKE Autopilot policies."
I0527 01:37:06.951059      89 warnings.go:110] "Warning: AdmissionWebhookController: mutated namespaceselector of the webhooks to enforce GKE Autopilot policies."
I0527 01:47:12.232464      89 warnings.go:110] "Warning: autopilot-default-resources-mutator:Autopilot updated Job agones-system/agones-delete-agones-resources: defaulted unspecified 'cpu' resource for containers [pre-delete-delete-agones-resources] (see http://g.co/gke/autopilot-defaults)."
Error: release agones failed, and has been uninstalled due to atomic being set: context deadline exceeded
make: *** [Makefile:446: install] Error 1
```

markmandel (approved):
Let's gooooo!

Issue #4514, opened 2026-04-15
Update Apache License Copyright Headers to CNCF Format


markmandel:
Agones is a CNCF project, and we need to update our source file copyright headers to follow the CNCF-recommended format. This ensures that the copyright is attributed to the project contributors under the Linux Foundation's umbrella.

As per the [CNCF Copyright Notices guidelines](https://github.com/cncf/foundation/blob/main/copyright-notices.md), the copyright notice should be updated to:
`Copyright Contributors to Agones a Series of LF Projects, LLC.`

**Tasks:**
- [ ] Identify all source files containing old copyright notices (e.g., `Copyright 2017 Google LLC`).
- [ ] Update headers in Go files (`*.go`).
- [ ] Update headers in Protocol Buffer files (`*.proto`).
- [ ] Update headers in Shell scripts (`*.sh`).
- [ ] Update headers in YAML/Configuration files where applicable.
- [ ] Ensure the `LICENSE` file and any `NOTICE` files are consistent with this change.

**Proposed Header Change:**

*Before:*
```go
// Copyright 2017 Google LLC
//
// Licensed under the Apache License, Version 2.0 (the "License");
// you may not use this file except in compliance with the License.
...
```

*After:*
```go
// Copyright Contributors to Agones a Series of LF Projects, LLC.
//
// Licensed under the Apache License, Version 2.0 (the "License");
// you may not use this file except in compliance with the License.
...
```

**References:**
- [CNCF Foundation - Copyright Notices](https://github.com/cncf/foundation/blob/main/copyright-notices.md)

lacroixthomas:
Huge PR incoming with this one 😄

markmandel:
Could also do this in chunks - doesn't _have_ to be done in one go.

mateenali66:
happy to take this on. saw the note about chunking, but one sweep across all file types (go/proto/sh/yaml) would avoid merge-conflict churn and keep the history clean. any objection to doing it in a single PR? if chunking is preferred I can start with `*.go` and follow up with the rest. could you assign it to me?

markmandel:
Some done! Bit more to go!

JeevaRamanathan:
I would like to work on `vendor/` directory

markmandel:
Just to repeat what I posted on the PR that I closed.

> All content from /vendor are third party dependencies. The licences of them are their own and shouldn't be changed.
>
> Appreciate the work, but I'm going to close this PR.

Plenty of other ones to do though!

[...]

PR #4527, opened 2026-04-22
chore: update copyright headers to CNCF format (install/helm)

mateenali66:
follow-up to #4522 covering the 30 YAML files under \`install/helm/agones/\`. regenerated \`install/yaml/install.yaml\` via \`make gen-install\` so the \`test-install-yaml\` check stays green.

one file (\`templates/extensions.yaml\`) is CRLF-terminated in the repo, so used perl to keep line endings intact rather than touching anything else in it. diff is +1/-1 on each of the 30 helm files.

Refs #4514


markmandel (approved):

markmandel (approved):

markmandel:
Thanks!

/gcbrun

Issue #2781, opened 2022-10-28
Allow graceful restarts of game server containers instead of shutdown


austin-space:
**Is your feature request related to a problem? Please describe.**
Pod churn(creating and deleting pods) is a relatively expensive operation. In cases where pod churn is particularly high, churn can get in the way of a fleet's ability to scale since both are contending for the same resources.

**Describe the solution you'd like**
Have some value in the game server health configuration like `AllowGracefulContainerRestart`. If this flag is set to true, instead of calling `Shutdown()` on the sdk, we'd just exit out of the game server process with a zero exit code.  The health controller would not mark this as unhealthy(as long as we had a non-zero exit code). The game server would then go back to the `Starting` state and would run normally from there. 

In the case where a game server exited with a non-zero exit code or exited before being allocated, that would indicate an unhealthy game server and the pod should be restarted.

This would remove the costly pod churn operation, while still allowing us to have a brand new game server container.

**Describe alternatives you've considered**
1. "Just calling Ready again": this is definitely a realistic option and is described in the docs. The issue is that it means that we have to be very diligent about ensuring we're actually clearing state between game server. Accidentally carrying over state can have privacy and security concerns. That seems like more of an optimization around the time to start a new game server being relatively high, which is not quite the case that this issue is trying to work around.
2. Wrapping the game server with another process that sits in the same container, and acts as a bridge between agones and our game server: also viable, but more difficult and burns more resources. It also leaves us in a similar situation where if we muck with the container's state in any way, so we mitigate some of the issue, but not all of it.

**Additional context**
This would be targeting a fairly specific case where the cost to start a new game server is very low, but the impact of starting a new pod is relatively high.

markmandel:
This is an interesting idea 🤔 and I can see the value!

For reference, this is where we track container failure and move containers to Unhealthy:
https://github.com/googleforgames/agones/blob/d70793c00abc4774265eecb4ff07bbcd3fde2250/pkg/gameservers/health.go#L135-L148

At first, I didn't think that the K8s API actually gave us access to status codes when a termination occurs, but I was wrong!

As we can see here:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#containerstateterminated-v1-core

There is an `exitCode` provided, so in theory, this does seem possible!

Additional thought/detail/questions:

[...]

PR #4528, opened 2026-04-23
Implement gameserver + cron support

### LINKED ISSUE #3907: Cron Job to restart the game server in-place
This is an extension of https://github.com/googleforgames/agones/issues/2781, because some game servers apparently have memory leak or game state lost (because game industry loved spaghetti code that worked quick'n'dirty with [cowboy programmers](https://en.wikipedia.org/wiki/Cowboy_coding) because it is a hyper-active industry) when they are running long enough, so we need a periodic time restart (say like during 4AM in midnight) to not let the game server die. Of course, I could always delete the game server externally, but that would also mean a lot of "pod-churning" to the scheduler. 

We also need to handle extra cases such as whether the game is still progressing, whether it still have players or not, because obviously it will impact ongoing players' experience. So, the cron is more like a soft-deadline, if either of the above two cases mentioned are true, of course we are not going to restart, but we will restart as soon as the game state is done. We can also introduce an optional hard deadline, say like even if the soft-deadline is triggered and awaiting a chance to restart, but if the soft-deadline is procrastinated long enough say like 2 days, then we will have to force kill it, since something might be wrong with the game server itself.

Relevant ideas: https://furiko.io/docs/execution/jobconfig/scheduling

Sivasankaran25:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature
> /kind hotfix
> /kind release

Closes #3907


igooch (commented):
It looks like this PR combines two distinct functional areas that should be discussed separately:

1. **In-place restart mechanism**: The fundamental logic within the SDK sidecar to trigger container restarts via `os.Exit(0)` when a specific annotation is detected. This is similar to the discussion in #2781 which looks like it will be unblocked as of [Kubernetes 1.35](https://github.com/kubernetes/kubernetes/issues/57291#issuecomment-3445011917).

2. **Cron-based scheduling**: The automation layer provided by the `RestartController` and the new `RestartPolicy` specification to manage periodic restarts and deadlines. If the foundation for #2781 is implemented, can this cron automation be simplified or addressed as a separate follow-up feature?

CC: @markmandel @stevefan1999-personal

markmandel:
Thanks for the ping @igooch - I've been meaning to jump into this.

I haven't checked the code yet (and I should), but I was going to push back on this whole PR pretty hard for a variety of reasons:

1. No design in the ticket -- what is the right approach for this? Why this approach. Our contributing guide says to do design first in an issue before developing as a general principle.
2. No broad consensus on the ticket. I know @stevefan1999-personal is super passionate about it (and we love you for that) - it would be good to get some general consensus from approvers that this is a feature and maintenance burden we want to take on (at first pass, it feels like something that people could do with a script inside the container?). Also would be ideal if more than one person said "yes, we want this" or if we (maintainers) think this will apply broadly. Just because one person wants something, doesn't mean it's something that we all agree on.
3. It's a huge PR -- which makes it hard to review, and also easy to miss issues. Assuming we're happy with (a) the design (b we agree it should be something we _want_ to do , then it should be broken up into smaller PRs to make it easy for review (this should also be part of our contributing guide as well -- I would like it to be when we review it as part of CNCF onboarding).

So I kinda want to see this closed, and have the conversation come back to the issue, see if we can reach consensus, and then we can talk plan and implementation.

markmandel:
I'll also make a semi side comment - can always post on an issue, but there is always #development on the slack server! I recently made it less noisy so that if anyone is ever wondering about best practices or talking about issues - that's a great place to have the conversations.

COME ON DOWN! 😄

stevefan1999-personal:
> It looks like this PR combines two distinct functional areas that should be discussed separately:
> 
> 1. **In-place restart mechanism**: The fundamental logic within the SDK sidecar to trigger container restarts via `os.Exit(0)` when a specific annotation is detected. This is similar to the discussion in #2781 which looks like it will be unblocked as of [Kubernetes 1.35](https://github.com/kubernetes/kubernetes/issues/57291#issuecomment-3445011917).
> 2. **Cron-based scheduling**: The automation layer provided by the `RestartController` and the new `RestartPolicy` specification to manage periodic restarts and deadlines. If the foundation for #2781 is implemented, can this cron automation be simplified or addressed as a separate follow-up feature?
> 
> CC: @markmandel @stevefan1999-personal

Surface area of this PR is good enough so split this into two, so I'm in favor of two separate PRs.

I need the container to restart to download the new dedicated server updates and within game the game server intrinsically detects if there is a new update right now, but I have code the SourceMod or AMXX plugins to detect if we need to restart the servers. 

However, sometimes you can't do that with self hosted game servers, those are usually like UE4 or UE5, so a more simple solution is to periodcally restart the server on, say like 4AM in the morning, since there are usually no people playing around that time, then the hours-long update downloading init container will kick in...until like 11AM where there are some people coming. That's the reason I needed cron support to chronically restart the servers on schedules...another benefit is to restart tactically to battle the memory leak problem over a long session of serving the games, as the memory gets more fragmented, so does the performance of the game server, I've seen it went down from 66 ticks per second down to 30-40 ticks per second, just because the game has too many memory allocations over 24 hours and the memory is scattered around different places, it can't keep up. But that's more like a plus reason, and not out of necessity.

Basically it would make so much more sense if some plugins that detects whether there is a new CS2 dedicated server update and tell Agones's "hey, don't mind me if I can serve more people, just kill me after the game is finished".

An even better solution for myself: code another program to run this on schedule, build a new docker image when there is a new dedicated server, push the image, and suspend the old game servers, only run new game servers with new image, Agones will phase them out one by one. Not sure if that's a better

markmandel:
> An even better solution for myself: code another program to run this on schedule, build a new docker image when there is a new dedicated server, push the image, and suspend the old game servers, only run new game servers with new image, Agones will phase them out one by one. Not sure if that's a better

This sounds like bash/python/go script you could write on your own. I don't know if it needs to be an Agones feature.

stevefan1999-personal:
> > An even better solution for myself: code another program to run this on schedule, build a new docker image when there is a new dedicated server, push the image, and suspend the old game servers, only run new game servers with new image, Agones will phase them out one by one. Not sure if that's a better
> 
> This sounds like bash/python/go script you could write on your own. I don't know if it needs to be an Agones feature.

That's why it is "an even better solution for myself" as it is not a general pattern

markmandel:
Shall we close this, given the discussion above? We can always come back to this on the issue -- but it sounds like something we may no take on.

Sivasankaran25:
> Shall we close this, given the discussion above? We can always come back to this on the issue -- but it sounds like something we may no take on.

Okay Sure

markmandel:
Closing to return to discussion in ticket (we may close the ticket too).

PR #4542, opened 2026-04-29
Add selectable fields to gameserver crd

swermin:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix
> /kind release

We have observers to make sure that our servers are being in the right state. Since we are running with multi servers it is hard to find specific servers with fields like `status` and `nodeName`. So, to facilitate the filtering we wanted to add those fields as selectable.

This is a nice to have feature, to be able to filter based on fields like state and node name.

### DIFF
--- a/site/content/en/docs/Installation/Install Agones/helm.md
+++ b/site/content/en/docs/Installation/Install Agones/helm.md
@@ -390,6 +390,17 @@ The following tables lists the configurable parameters of the Agones chart and t
 
 ### GameServers
 
+{{% feature expiryVersion="1.58.0" %}}
+| Parameter                              | Description                                                                                                                             | Default       |
+| -------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- | ------------- |
+| `gameservers.namespaces`               | a list of namespaces you are planning to use to deploy game servers                                                                     | `["default"]` |
+| `gameservers.minPort`                  | Minimum port to use for dynamic port allocation                                                                                         | `7000`        |
+| `gameservers.maxPort`                  | Maximum port to use for dynamic port allocation                                                                                         | `8000`        |
+| `gameservers.additionalPortRanges`     | Port ranges from which to do named dynamic port allocation. Example: <br /> additionalPortRanges: <br />&nbsp;&nbsp;game: [9000, 10000] | `{}`          |
+| `gameservers.podPreserveUnknownFields` | Disable [field pruning][pruning] and schema validation on the Pod template for a [GameServer][gameserver] definition                    | `false`       |
+| `gameservers.lists.maxItems`           | The maximum number of items that can be specified for a list.                                                                           | `1000`        |
+{{% /feature %}}
+{{% feature publishVersion="1.58.0" %}}
 | Parameter                              | Description                                                                                                                             | Default                                |
 | -------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------|
 | `gameservers.namespaces`               | a list of namespaces you are planning to use to deploy game servers                                                                     | `["default"]`                          |
@@ -397,11 +408,9 @@ The following tables lists the configurable parameters of the Agones chart and t
 | `gameservers.maxPort`                  | Maximum port to use for dynamic port allocation                                                                                         | `8000`                                 |
 | `gameservers.additionalPortRanges`     | Port ranges from which to do named dynamic port allocation. Example: <br /> additionalPortRanges: <br />&nbsp;&nbsp;game: [9000, 10000] | `{}`                                   |
 | `gameservers.podPreserveUnknownFields` | Disable [field pruning][pruning] and schema validation on the Pod template for a [GameServer][gameserver] definition                    | `false`                                |
-{{% feature publishVersion="1.58.0" %}}
 | `gameservers.selectableFields`         | spec fields available for querying [GameServer][gameserver] resources.                                                                  | `[".status.state", "status.nodeName"]` |
-{{% /feature %}}
 | `gameservers.lists.maxItems`           | The maximum number of items that can be specified for a list.                                                                           | `1000`                                 |
-
+{{% /feature %}}
 
 ### Helm Installation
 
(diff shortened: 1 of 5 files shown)


lacroixthomas:
Hello ! 
You might need to run `make gen-install`, this will update the `./install/yaml/install.yaml`

swermin:
Ah, good point! Let me do that

markmandel:
Is there any limit on the number of selectable fields we can apply?

swermin:
> Is there any limit on the number of selectable fields we can apply?

I don’t think so. What do you have in mind?

lacroixthomas:
Seems that there is indeed a limitation of 8 selectable fields per CRD: https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1#CustomResourceDefinitionVersion 

```
// selectableFields specifies paths to fields that may be used as field selectors.
	// A maximum of 8 selectable fields are allowed.
	// See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors
	//
	// +featureGate=CustomResourceFieldSelectors
	// +optional
	// +listType=atomic
	SelectableFields [][SelectableField](https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1#SelectableField) `json:"selectableFields,omitempty" protobuf:"bytes,9,rep,name=selectableFields"`
```
Might not be an issue, but we can keep an eye on it, I guess these selectable fields makes sense ?

markmandel:
I thought I read that there was a limit!

lacroixthomas:
> I thought I read that there was a limit!

And you were right 😄

markmandel:
I don't have any issues with these fields - but we should be careful with what we pick.

swermin:
> Seems that there is indeed a limitation of 8 selectable fields per CRD:

I cannot seem to read the manual 🤣

swermin:
> I don't have any issues with these fields - but we should be careful with what we pick.

I get that. These fields would help us make our maintenance cycle simpler, but that doesn’t mean that it is a right choice to put in here.
I defer to your judgement here if you want to merge it or not.

Our gameservers (for now) do not support signals. So we have `safe-to-evict: false` set.
Currently when maintenance is needed on a node then we need to use a combination of`awk + sed` to figure out which `Ready` gameservers to delete and let the `Allocated` one’s drain normally.

These selectors would make is so that we can run a more simple command, like:
```
Kubectl delete gs —field-selector status.state=Ready,status.nodeName=a-node-name
```

markmandel:
I have a fun idea - why not make selected fields helm configurable, and we can default to the two you have selected already.

We don't use them internally, and that way, anyone can change them should they need to.

WDYT?

swermin:
I was actually wondering if that was possible. Because it would make sense to give the option for others to set the ones they need (if needed that is)

swermin:
I’ll do some tinkering!

swermin:
I am also fine to keep the selectable fields empty if that is preferred!

markmandel:
/gcbrun

Love it.

Thinking about docs.

We'll definitely need docs here: https://agones.dev/site/docs/installation/install-agones/helm/#gameservers

But wondering where to put that fields are selectable in the docs -- maybe in here? https://agones.dev/site/docs/reference/gameserver/ ?

swermin:
Oh, I forgot about the documentation. Thanks for pointing me to the right ones 🙏

I’ve added accordingly, but documentation is not my strongest suit so please look at it with critical eyes!!

site/content/en/docs/Reference/gameserver.md
```diff
@@ -158,6 +158,17 @@ The `spec` field is the actual GameServer specification and it is composed as fo
 The GameServer resource does not support updates. If you need to make regular updates to the GameServer spec, consider using a [Fleet]({{< ref "/docs/Reference/fleet.md" >}}).
 {{% /alert %}}
 
+{{% feature publishVersion="1.58.0" %}}
+## Selectable fields
+
+Agones supports selectable fields for querying `GameServer` resources based on spec field values. This feature enables efficient server-side filtering of `GameServer` resources.
```
markmandel:
```suggestion
Agones supports [selectable fields](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) for querying `GameServer` resources based on spec field values. This feature enables efficient server-side filtering of `GameServer` resources.
```

site/content/en/docs/Reference/gameserver.md
```diff
@@ -158,6 +158,17 @@ The `spec` field is the actual GameServer specification and it is composed as fo
...
+
+The following spec fields are available for field selectors in `GameServer` resources:
```
markmandel:
```suggestion
By default, the following spec fields are available for field selectors in `GameServer` resources:
```

site/content/en/docs/Installation/Install Agones/helm.md
```diff
@@ -390,14 +390,17 @@ The following tables lists the configurable parameters of the Agones chart and t
...
+| `gameservers.namespaces`               | a list of namespaces you are planning to use to deploy game servers                                                                     | `["default"]`                          |
+| `gameservers.minPort`                  | Minimum port to use for dynamic port allocation                                                                                         | `7000`                                 |
+| `gameservers.maxPort`                  | Maximum port to use for dynamic port allocation                                                                                         | `8000`                                 |
+| `gameservers.additionalPortRanges`     | Port ranges from which to do named dynamic port allocation. Example: <br /> additionalPortRanges: <br />&nbsp;&nbsp;game: [9000, 10000] | `{}`                                   |
+| `gameservers.podPreserveUnknownFields` | Disable [field pruning][pruning] and schema validation on the Pod template for a [GameServer][gameserver] definition                    | `false`                                |
+{{% feature publishVersion="1.58.0" %}}
```
markmandel: I expect this will need to be put outside the table, because it'll render badly, but let's do a preview and see what happens.

site/content/en/docs/Reference/gameserver.md
```diff
@@ -158,6 +158,17 @@ The `spec` field is the actual GameServer specification and it is composed as fo
...
+- status.state - The lifecycle of the `GameServer` (See **GameServer State Diagram**)
+- status.nodeName - The node which the `GameServer` is running on
+
```
markmandel, line 169:
```suggestion

If you want to configure which fields are selectable, you can configure this through [Helm GameServer configuration variables]({% ref "docs/installation/install-agones/helm.md#gameservers )}

```

(I think that ref path is right - might want to check it with `make site-server`)

markmandel (commented):
Nice!

site/content/en/docs/Installation/Install Agones/helm.md
swermin: You were correct, the page did not render properly. I've updated the code accordingly

site/content/en/docs/Reference/gameserver.md
swermin: Nice addiction, did not think about linking to the actual configuration as well 🙏

swermin:
@markmandel thank for for showing the `make site-server` command, I didn't know it existed and it did help me test the site locally. I just assumed a gcburn was was needed

markmandel:
/gcbrun

Ah, I have a fix for TestGameServerTcpProtocol incoming.

markmandel (approved):
Nice!

PR #4546, opened 2026-05-03
Feat: Add PDB on new processor allocator

lacroixthomas:
**What type of PR is this?**
>
/kind feature

Add PDB on the new processor allocator (same way as the extensions - and also matching the latest fix around it with min and max)

Part of #4190 4190


markmandel (approved):
Noice

PR #4549, opened 2026-05-04
Fix SucceededController to handle pods with deletion timestamp

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

When Kubernetes sets a deletion timestamp on a pod simultaneously with it reaching PodSucceeded phase, the SucceededController was skipping the GameServer transition to Shutdown. Remove the DeletionTimestamp guard so a pod in PodSucceeded phase always triggers GameServer→Shutdown.

Update the test case 'pod is in terminating state' to reflect the corrected behaviour: expect updated=true and GameServerStateShutdown.

N/A

Should fix TestGameServerPodCompletedAfterCleanExit flake!


markmandel:
Totally valid failures. Need to think about this a bit more.

markmandel:
Closing while I think about this some more.

PR #4556, opened 2026-05-07
Update fleetautoscalers.go

### LINKED ISSUE #4555: FleetAutoscaler webhook and Wasm handlers panic when response body omits 'response' field
## Summary

`applyWebhookPolicy` and `applyWasmPolicy` in
`pkg/fleetautoscalers/fleetautoscalers.go` both dereference
`Response.Scale` without first checking if `Response` is nil.

When a webhook or Wasm policy returns a JSON body that omits the
`response` field (e.g., `{}`), `json.Unmarshal` succeeds with
`Response` as a nil pointer. The subsequent dereference panics,
crashing the controller.

Both HA controller replicas pull the same FleetAutoscaler from the
shared workqueue and crash on the same item, producing CrashLoopBackOff
on both replicas. FleetAutoscaler reconciliation halts cluster-wide
until the resource is removed by an admin.

## Affected code

`pkg/fleetautoscalers/fleetautoscalers.go`:
- `applyWebhookPolicy`: `faResp.Response.Scale` accessed without nil check
- `applyWasmPolicy`: `review.Response.Scale` accessed without nil check

## Reproduction

Tested against `helm install agones` release 1.57.0 in a kind cluster.

1. Deploy any minimal Fleet.
2. Deploy an HTTP server that returns `{}` to all POST requests:

```python
from http.server import HTTPServer, BaseHTTPRequestHandler

class H(BaseHTTPRequestHandler):
    def do_POST(self):
        self.send_response(200)
        self.send_header("Content-Type", "application/json")
        self.end_headers()
        self.wfile.write(b'{}')

HTTPServer(('', 8888), H).serve_forever()
```

3. Apply a FleetAutoscaler pointing at that server:

```yaml
kind: FleetAutoscaler
metadata:
  name: panic-poc
spec:
  fleetName: <fleet-name>
  policy:
    type: Webhook
    webhook:
      url: "http://<server-service>.<namespace>.svc.cluster.local:8888/"
```

4. Both `agones-controller` pods enter CrashLoopBackOff within 30s.

## Stack trace

agones.dev/agones/pkg/fleetautoscalers.applyWebhookPolicy(...)
.../pkg/fleetautoscalers/fleetautoscalers.go:355 +0x974
[recovered, repanicked]

## Fix

A nil check on `Response` before dereferencing at both sites.
A PR with the fix is forthcoming.

MusGaas:
/kind bug

applyWebhookPolicy and applyWasmPolicy both dereference Response.Scale
without checking if Response is nil. When a webhook or Wasm policy
returns a JSON body that omits the response field (e.g., {}),
json.Unmarshal succeeds with Response as a nil pointer. The subsequent
dereference panics, crashing both HA controller replicas via the shared
workqueue and halting cluster-wide FleetAutoscaler reconciliation until
an admin removes the resource.

Both sites now return an explicit error when Response is nil, allowing
the reconcile loop to handle the error gracefully rather than panicking.

Tested against helm install agones release 1.57.0 in a kind cluster.
Both controller pods entered CrashLoopBackOff before the fix. After
the fix, the error is returned cleanly with no panic.

Closes #4555

This fix was identified and reported via Google's OSS VRP (Agones
documented security intake at g.co/vulnz). Two sites fixed in the
same file — applyWebhookPolicy and applyWasmPolicy — same root cause,
same one-line nil check pattern.


MusGaas:
Hi, could a maintainer please run `/gcbrun` to trigger the CI build? Thank you.

markmandel:
We should add a unit test for this, to make sure it continues to not be a problem.

markmandel:
> This fix was identified and reported via Google's OSS VRP (Agones
> documented security intake at g.co/vulnz). Two sites fixed in the
> same file — applyWebhookPolicy and applyWasmPolicy — same root cause,
> same one-line nil check pattern.

Two things:

1. That's not our security policy anymore, and hasn't been for several weeks.
2. I don't think this is a security issue, since it's not something that is accessible from the outside world (unless someone configures their cluster badly - which we can't control). Unless I'm missing something here?

markmandel:
Gentle bump on the above?

markmandel:
Closing this, since I'm replacing it with https://github.com/agones-dev/agones/pull/4596 , since it's been a month with no response.

PR #4557, opened 2026-05-07
Fix extensions apiserver to authenticate requests via RequestHeader CA

anandppatil:
## What

Fixes an authentication bypass in the agones-extensions apiserver where
any in-cluster workload could directly call the aggregated APIService
endpoint and allocate GameServers in any namespace, bypassing Kubernetes RBAC.

## Why

The extensions apiserver serves `/apis/allocation.agones.dev/v1/...` as a
Kubernetes aggregated APIService but did not implement the client-side
authentication required by the [Kubernetes aggregation layer specification](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/).

## How

- Loads `extension-apiserver-authentication` ConfigMap from kube-system
- Configures TLS `ClientAuth: VerifyClientCertIfGiven` with the RequestHeader CA
- Verifies proxy client certificate CN against `requestheader-allowed-names`
- Extracts `X-Remote-User` / `X-Remote-Group` headers
- Performs `SubjectAccessReview` for `create gameserverallocations`

## Testing

- Unit tests for `LoadRequestHeaderConfig`, `AuthenticateRequest`, and
  `parseJSONStringArray` added in `pkg/util/apiserver/auth_test.go`
- Manual verification: deploy to a Kind cluster, confirm that direct
  `curl` to the service without a valid proxy client cert returns 401.

Fixes: https://issuetracker.google.com/issues/505221322

### DIFF
--- a/cmd/extensions/main.go
+++ b/cmd/extensions/main.go
@@ -144,6 +144,24 @@ func main() {
 	}
 	// https server and the items that share the Mux for routing
 	httpsServer := https.NewServer(ctlConf.CertFile, ctlConf.KeyFile, ctlConf.WebhookPort)
+
+	// Load the RequestHeader CA from the extension-apiserver-authentication ConfigMap.
+	// This enables the extensions apiserver to verify that requests come from the
+	// kube-apiserver aggregator, as required by the Kubernetes aggregation layer spec.
+	// See: https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/
+	reqHeaderConfig, err := apiserver.LoadRequestHeaderConfig(ctx, kubeClient)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


markmandel:
So I'm going to close this - it's pretty clearly AI slop.

- rolled own JSON parser
- included patch file?
- generated entire code base rather than using existing libraries.
- and more.

If you are serious about inspecting this as a potential issue - please file an issue and we'll discuss approaches and implementation details.

anandppatil:
Fair point on the PR shape — closing this. The underlying vulnerability (missing RequestHeader CA authentication on the aggregated APIService) is a real finding, but the fix should come from maintainers who know the codebase.

I'll open a focused issue with the vulnerability details and repro steps so the team can discuss the right implementation approach.

PR #4560, opened 2026-05-11
Implmemnted Feature Windows Gameserver Move Alpha to beta

### LINKED ISSUE #4523: Move Windows (windows/amd64) GameServer support from Alpha to Beta
## Community Question First

**Before committing to this work: is anyone actually running Agones GameServers on Windows nodes in production or at scale?**

Windows CI infrastructure has real ongoing cost. If there is no active user base, it may be more appropriate to deprecate Windows support rather than invest in graduating it. Please comment with your use case if you rely on Windows GameServers — that feedback will directly shape whether this issue moves forward.

---

## Summary

Windows GameServer support was added as Alpha and is documented as such in both the [installation matrix](https://agones.dev/site/docs/installation/) and the [Windows GameServers guide](https://agones.dev/site/docs/guides/windows-gameservers/). The missing piece for Beta graduation is **continuous integration on real Windows nodes** — without it we cannot assert correctness or catch regressions.

Notably, Windows support in Agones is intentionally narrow: **only the SDK sidecar (`sdk-server`) has Windows builds**. All control-plane components (controller, allocator, extensions, ping) remain Linux-only. This significantly limits the scope of CI needed.

## Criteria for Beta

Per the [feature stage definitions](https://agones.dev/site/docs/guides/feature-stages/):

- No known critical bugs.
- Automated testing covers the feature.
- Documentation is up to date.

## Proposed CI Approach

Add a **small Windows node pool to one of the existing GKE Standard e2e clusters**. Windows node pools in GKE require at least one Linux node pool for system pods — the existing clusters already satisfy this — so no new cluster is needed.

**Node pool spec (suggested):**
- Machine type: `n2-standard-4`
- Windows versions: `ltsc2022` (ltsc2019 as secondary)
- Node count: 2 nodes
- Added to one of the existing `build/terraform/e2e/gke-standard/` cluster configs

**Test subset:**

| Test | What it validates |
|---|---|
| `TestCreateConnect` (Windows variant) | GameServer schedules on a Windows node, SDK sidecar initialises, GameServer reaches Ready, UDP connectivity works |

The test uses a Windows-specific GameServer manifest (mirroring `examples/simple-game-server/gameserver-windows.yaml`) with `nodeSelector: kubernetes.io/os: windows` and the Windows `simple-game-server` image.

## Tasks

- [ ] **Community validation**: Confirm there is an active user base for Windows GameServers (see question above)
- [ ] Add a Windows node pool to the Terraform config for one of the existing e2e clusters
- [ ] Ensure the `simple-game-server` Windows image is built and pushed as part of the release pipeline
- [ ] Add a Windows `TestCreateConnect` variant (or parameterise the existing test with a Windows GameServer fixture)
- [ ] Add the Windows smoke test to `ci/e2e-test-cloudbuild.yaml`
- [ ] Run the smoke test consistently green across ≥ 2 weeks / several PRs
- [ ] Update `site/content/en/docs/Installation/_index.md`: `Alpha` → `Beta`
- [ ] Update `site/content/en/docs/Guides/windows-gameservers.md` to remove the Alpha caveat
- [ ] Add a blog/changelog entry noting the graduation

## Notes

- GKE Windows node pools require a Linux node pool — already satisfied by existing e2e clusters.
- Windows nodes cannot run the Agones control plane; this is by design and unchanged.
- A dedicated Windows smoke test is required even with Windows nodes present in the cluster. GKE taints Windows nodes with `node.kubernetes.io/os=windows:NoSchedule` by default, so regular e2e GameServers (Linux images, no node selector) will never schedule there. Only a GameServer spec that explicitly sets `nodeSelector: kubernetes.io/os: windows` and uses a Windows container image will exercise the Windows node pool.
- The smoke suite intentionally covers only the sidecar path — no allocator, fleet, or HA tests needed for Beta.

Sivasankaran25:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

 /kind feature
> /kind hotfix
> /kind release

Closes #4523


nikkisingh0204 (approved):

igooch (commented):
GKE Autopilot does not support Windows Nodes https://docs.cloud.google.com/kubernetes-engine/docs/concepts/windows-server-gke#limitations. Try running tests on Standard only.

markmandel:
Curious to hear how the work here is going? I can't see the clusters, but what is failing here?

Sivasankaran25:
> Curious to hear how the work here is going? I can't see the clusters, but what is failing here?

The `e2e-windows-smoke` test step failed because GKE Autopilot does not support [Windows](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/windows-server-gke#limitations) nodes. The test must be configured to run only on GKE Standard clusters. Please suggest possible approaches to fix this build error.

markmandel:
Could we use something like what we did here to skip autopilot?

https://github.com/agones-dev/agones/blob/00f9151d5df654cc60605860e7a52f773e4fd4b7/test/e2e/gameserver_test.go#L263

Sivasankaran25:
> Could we use something like what we did here to skip autopilot?
> 
> https://github.com/agones-dev/agones/blob/00f9151d5df654cc60605860e7a52f773e4fd4b7/test/e2e/gameserver_test.go#L263

Yes Sure I will try

examples/simple-game-server/gameserver-windows.yaml
```diff
@@ -21,11 +21,12 @@ spec:
     - name: default
       portPolicy: Dynamic
       containerPort: 7654
+      protocol: UDP
   template:
     spec:
       containers:
         - name: simple-game-server
-          image: us-docker.pkg.dev/agones-images/examples/simple-game-server:0.43
+          image: us-docker.pkg.dev/agones-images/examples/simple-game-server:0.42-windows_amd64-ltsc2019
```
markmandel: This shouldn't be necessary -- the manifest should load the windows image.

site/content/en/docs/Guides/windows-gameservers.md
```diff
@@ -7,9 +7,9 @@ description: >
   Run `GameServers` on Kubernetes nodes with the Windows operating system.
 ---
 
-{{% alert title="Warning" color="warning" %}}
-Running `GameServers` on Windows nodes is currently Alpha, and any feedback
-would be appreciated.
+{{% alert title="Note" color="info" %}}
```
markmandel: All doc changes need to be feature shortcoded - https://agones.dev/site/docs/contribute/documentation-editing-contribution/

site/content/en/docs/Installation/_index.md
```diff
@@ -33,7 +33,7 @@ The following container operating systems and architectures can be utilised with
 | --------- | ------------ | ---------- |
 | linux     | `amd64`      | **Stable** |
 | linux     | `arm64`      | Alpha      |
-| [windows] | `amd64`      | Alpha      |
+| [windows] | `amd64`      | Beta       |
```
markmandel, line 37: Same here - feature shortcoded.

test/e2e/windows_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// windowsGameServerImage is the published windows/amd64 simple-game-server.
+// Uses ltsc2019 — the currently available tag in agones-images registry.
+const windowsGameServerImage = "us-docker.pkg.dev/agones-images/examples/simple-game-server:0.42-windows_amd64-ltsc2019"
```
markmandel: I'd double check the manifest for this image - but pretty sure it's a multi-arch and should select the correct image by detecting the host arch

ci/e2e-test-cloudbuild.yaml
```diff
@@ -53,12 +51,32 @@ steps:
...
     id: e2e-stable
     waitFor: [e2e-feature-gates]
+  #
+  # Run Windows GameServer smoke test.
+  # Windows nodes are NOT supported on GKE Autopilot:
+  # https://cloud.google.com/kubernetes-engine/docs/concepts/windows-server-gke#limitations
+  # The test itself calls t.Skip() when CloudProduct == "gke-autopilot".
+  # Runs parallel with e2e-feature-gates to avoid adding wall-clock time.
+  #
+  - name: e2e-runner
```
markmandel: 🤔 if the test `windows_test.go` is in the e2e folder - why do we need this extra entry in this config? Shouldn't it just run in the standard e2e test suite?

markmandel: I'll reiterate this - we have tests that only run on autopilot, why do we need this extra build step? This should just be a single extra test that runs on standard GKE clusters, that tests the windows nodes.

test/e2e/windows_test.go
```diff
@@ -0,0 +1,55 @@
...
+func TestWindowsCreateConnect(t *testing.T) {
+	t.Parallel()
+
+	if framework.CloudProduct == "gke-autopilot" {
+		t.Skip("Windows nodes are not supported on GKE Autopilot")
+	}
+
+	gs := framework.DefaultGameServer(framework.Namespace)
+	gs.Spec.Template.Spec.NodeSelector = map[string]string{
+		"kubernetes.io/os": "windows",
```
markmandel, line 34: Looks good.

ci/e2e-test-cloudbuild.yaml
```diff
@@ -53,12 +51,30 @@ steps:
...
+    timeout: 900s
 tags:
   - e2e-test
   - cluster-${_TEST_CLUSTER_NAME}
   - location-${_TEST_CLUSTER_LOCATION}
   - commit-${_PARENT_COMMIT_SHA}
   - started-by-${_PARENT_BUILD_ID}
 timeout: 5400s  # 1.5h
-queueTtl: 7200s  # 2h // only one set of e2es should be running at once
-logsBucket: gs://agones-build-logs
```
markmandel, line 64: Don't remove the log bucket - that's why I can't see the logs.

ci/e2e-test-cloudbuild.yaml
```diff
@@ -53,12 +51,30 @@ steps:
...
+      - ${_FEATURE_WITHOUT_GATE}
+      - ${_CLOUD_PRODUCT}
+      - ${_TEST_CLUSTER_NAME}
+      - ${_TEST_CLUSTER_LOCATION}
+      - ${_REGISTRY}
+      - ${_GS_TEST_IMAGE}
+    id: e2e-windows-smoke
+    waitFor:
+      - build-e2e
+    timeout: 900s
```
markmandel:
```suggestion
    timeout: 900s
```

test/e2e/windows_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// TestWindowsCreateConnect is a smoke test that verifies a GameServer can be
+// scheduled onto a Windows node, reach Ready, and successfully respond to a
+// UDP connection. Windows GameServer support is Beta - see
+// https://agones.dev/site/docs/guides/windows-gameservers/ for details.
+func TestWindowsCreateConnect(t *testing.T) {
+	t.Parallel()
+
+	if framework.CloudProduct == "gke-autopilot" {
```
markmandel: Since you have this, it should run as part of the standard set of e2e tests -- no need for the extra build step.

test/e2e/windows_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+package e2e
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
```
igooch, line 23: `corev1` is needed for the toleration suggested below:

```suggestion
	agonesv1 "agones.dev/agones/pkg/apis/agones/v1"
	corev1 "k8s.io/api/core/v1"
```

test/e2e/windows_test.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	gs := framework.DefaultGameServer(framework.Namespace)
+	gs.Spec.Template.Spec.NodeSelector = map[string]string{
+		"kubernetes.io/os": "windows",
+	}
+	// framework.GameServerImage is multi-arch (linux/amd64 + windows/amd64),
+	// so no separate Windows-tagged image is required - the correct
+	// platform variant is selected automatically based on the node.
+	gs.Spec.Template.Spec.Containers[0].Image = framework.GameServerImage
```
igooch: 🔴 GKE taints Windows nodes `node.kubernetes.io/os=windows:NoSchedule`, but `DefaultGameServer` sets no toleration, so the pod can't schedule onto the Windows node. Also prefer the existing `SkipOnCloudProduct` helper, and drop the multi-arch comment — the e2e image is built with `WITH_WINDOWS=0` (see summary), so `GameServerImage` has no `windows/amd64` variant today.

```suggestion
	framework.SkipOnCloudProduct(t, "gke-autopilot", "Windows nodes are not supported on GKE Autopilot")

	gs := framework.DefaultGameServer(framework.Namespace)
	gs.Spec.Template.Spec.NodeSelector = map[string]string{
		"kubernetes.io/os": "windows",
	}
	gs.Spec.Template.Spec.Tolerations = []corev1.Toleration{{
		Key:      "node.kubernetes.io/os",
		Operator: corev1.TolerationOpEqual,
		Value:    "windows",
		Effect:   corev1.TaintEffectNoSchedule,
	}}
	gs.Spec.Template.Spec.Containers[0].Image = framework.GameServerImage
```

build/terraform/e2e/gke-standard/module.tf
```diff
@@ -65,3 +65,21 @@ module "gke_cluster" {
...
+    machine_type = "n2-standard-4"
+    image_type   = "WINDOWS_LTSC_CONTAINERD"
+    tags         = ["game-server"]
+  }
+
+  management {
+    auto_repair  = true
+    auto_upgrade = false
+  }
+}
```
igooch, line 91: 🟡 Set the Windows taint on the pool so the toleration above is actually exercised (GKE applies `node.kubernetes.io/os=windows:NoSchedule` to Windows nodes). This suggestion also restores the missing trailing newline that trips `terraform fmt`.

```suggestion
resource "google_container_node_pool" "windows_ltsc2022" {
  name       = "win-ltsc2022"
  cluster    = module.gke_cluster.name
  location   = var.location
  node_count = 2

  node_config {
    machine_type = "n2-standard-4"
    image_type   = "WINDOWS_LTSC_CONTAINERD"
    tags         = ["game-server"]

    taint {
      key    = "node.kubernetes.io/os"
      value  = "windows"
      effect = "NO_SCHEDULE"
    }
  }

  management {
    auto_repair  = true
    auto_upgrade = false
  }
}
```

igooch (commented):
Thanks for moving Windows GameServers to Beta! The docs and support-tier changes look right, and the Autopilot handling is correct — the test is skipped on Autopilot and the node pool is only added to the standard cluster.

However, **`TestWindowsCreateConnect` will fail consistently on the standard e2e cluster** (where it actually runs) for two reasons. Fixes are suggested inline, plus one blocker that lives outside this diff is that no Windows image is built for e2e:

`cloudbuild.yaml:188` builds and pushes the e2e `simple-game-server` image with `WITH_WINDOWS=0`, so its manifest is linux/amd64-only (`examples/simple-game-server/Makefile` only adds the `windows_amd64` layers when `WITH_WINDOWS=1`). The test comment claims `framework.GameServerImage` is multi-arch with a `windows/amd64` variant — that isn't true in CI, so a Windows-pinned pod has no runnable image and never reaches `Ready`. This needs `WITH_WINDOWS=1` for the e2e image build step:

```yaml
      - 'WITH_WINDOWS=1'
```

Inline suggestions

- Add the Windows node toleration and prefer the `SkipOnCloudProduct` helper (test file).
- Set the Windows taint on the node pool and restore the missing trailing newline (`module.tf`).

Nice work overall — once the e2e image includes a Windows variant and the pod can tolerate/schedule onto the Windows node, this should go green.

markmandel:
@igooch , @Sivasankaran25 - we just waiting to create windows nodes on the clusters for this, yeah?

site/content/en/docs/Installation/_index.md
```diff
@@ -29,11 +29,22 @@ This release has been tested against Kubernetes versions {{% k8s-version %}} on
 
 The following container operating systems and architectures can be utilised with Agones:
 
+{{% feature publishVersion="1.60.0" %}}
```
markmandel: Since we've done a release, let's update these versions.

test/e2e/windows_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+// TestWindowsCreateConnect is a smoke test that verifies a GameServer can be
+// scheduled onto a Windows node, reach Ready, and successfully respond to a
+// UDP connection. Windows GameServer support is Beta - see
+// https://agones.dev/site/docs/guides/windows-gameservers/ for details.
+func TestWindowsCreateConnect(t *testing.T) {
+	t.Parallel()
+
+	framework.SkipOnCloudProduct(t, "gke-autopilot", "Windows nodes are not supported on GKE Autopilot")
+	skipIfNoWindowsNodes(t)
```
markmandel: 🤔 why do we want this? This feels like the sort of thing that would hide issues with testing the windows machines.

TBH, I didn't look, and assumed the tests where passing. I'd recommend removing this, because if we don't have windows nodes on a specific gke standard version, these tests should fail.

PR #4563, opened 2026-05-12
flaky: TestMultiClusterAllocationFromLocal

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Looks like a feature flag switching in the middle of this test make it go bad. So locking it down.

N/A

N/A


lacroixthomas (approved):
I wonder if we could keep track of which tests use any specific feature gate when calling `runtime.ParseFeatures()`, then showing the unit test name / feature gates enabled on any error to show the state of the feature gates ?

It's just a random idea, just trying to find a way to ensure we lock/unlock properly for these tests, this might give an idea of the cause 🤔

PR #4567, opened 2026-05-13
fix: simple-game-server ports are bound before calling SDK Ready()

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Previously, tcpListener and udpListener were launched as goroutines and s.Ready() was called immediately after, with no guarantee that either goroutine had reached its net.Listen/net.ListenPacket call. On fast nodes (e.g. GKE Autopilot with a warm NAP node and cached image), the game server could become Ready in ~3 seconds, allowing the e2e test to attempt a TCP connection before the listener goroutine had been scheduled and bound its port — resulting in a flaky "connection refused" failure in TestGameServerTcpProtocol.

Fix this by using a sync.WaitGroup: each listener goroutine calls ready.Done() immediately after net.Listen/net.ListenPacket succeeds (the moment the OS has the port open and will buffer incoming connections/datagrams), and main() calls listenersReady.Wait() before invoking s.Ready(), guaranteeing the ports are bound before the GameServer is ever marked Ready.

N/A

N/A

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -66,7 +66,7 @@ KIND_PROFILE ?= agones
 KIND_CONTAINER_NAME=$(KIND_PROFILE)-control-plane
 
 # Game Server image to use while doing end-to-end tests
-GS_TEST_IMAGE ?= us-docker.pkg.dev/agones-images/examples/simple-game-server:0.42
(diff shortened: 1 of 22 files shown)


igooch (approved):
/lgtm

markmandel:
omg yay

please work please work please work

PR #4569, opened 2026-05-15
feat: Set Go runtime hints for SDK sidecar

### LINKED ISSUE #4568: Optimize SDK sidecar Go runtime defaults for unbounded Kubernetes pods
## Problem

The Agones SDK sidecar can run without explicit CPU or memory limits. In those environments, language runtimes may size scheduler parallelism or memory behavior from host/node-visible resources rather than the small amount of resources the sidecar is expected to use.

This can be more visible on Windows process-isolated containers, where CPU limits are enforced by HCS/Job Objects but may not be reflected as a reduced logical CPU count inside the runtime.

## Proposed Approach

Use the SDK sidecar's requested resources as runtime resource hints:

- CPU: derive the runtime CPU allocation hint from the sidecar CPU request.
- Memory: derive the runtime memory allocation hint from the sidecar memory request.
- If a request is not configured, do not provide an additional runtime hint for that resource.

This should not change Kubernetes resource requests or limits. It should only help the SDK sidecar runtime align its internal scheduling and memory behavior with the resources Agones asks Kubernetes to reserve for it.

mnthe:
/kind feature

### **What this PR does / Why we need it**:

This PR sets Go runtime resource hints for the GameServer SDK
sidecar:

- `GOMAXPROCS=1`
- `GOMEMLIMIT=<90% of sidecar memory request>` when
`agones.image.sdk.memoryRequest` is configured

The SDK sidecar is expected to be lightweight and should not need more than one vCPU. In pods without CPU limits, the Go runtime may otherwise size scheduler parallelism from host- visible CPU resources, which can be inefficient for the sidecar workload.

For memory, this derives the runtime memory hint from the configured sidecar memory request. If no sidecar memory request is configured, `GOMEMLIMIT` is not set.

This does not change Kubernetes CPU or memory requests or limits. It only provides runtime hints to the SDK sidecar process.

### **Which issue(s) this PR fixes**:

Closes #4568

### **Special notes for your reviewer**:

`GOMAXPROCS` is fixed at `1` because the SDK sidecar is not expected to need more than one vCPU, even when the CPU request is unset. If there are expected SDK sidecar workloads where using more than one vCPU is important, I can change this to derive `GOMAXPROCS` from the configured CPU request instead.

`GOMEMLIMIT` is set to 90% of the memory request to leave headroom for memory outside Go runtime-managed memory. The 90% ratio is a heuristic; if this does not leave enough headroom, I can adjust it to a more conservative value such as 80%.

The SDK sidecar is also expected to be lightweight in memory usage. However, in some Windows dedicated server deployments, the sidecar may be run in a Guaranteed QoS pod and can temporarily consume more memory than expected. For that case, this PR derives `GOMEMLIMIT` from the configured memory request so the Go runtime has a soft memory hint aligned with the scheduled sidecar resources.

Validation:
- `git diff --check`
- `go test -mod=vendor ./pkg/gameservers -run TestControllerSidecarGoRuntimeResourceHints -count=1`


pkg/gameservers/controller.go
```diff
@@ -64,7 +64,13 @@ const (
 	sdkserverSidecarName  = "agones-gameserver-sidecar"
 	grpcPortEnvVar        = "AGONES_SDK_GRPC_PORT"
 	httpPortEnvVar        = "AGONES_SDK_HTTP_PORT"
+	goMaxProcsEnvVar      = "GOMAXPROCS"
+	goMemLimitEnvVar      = "GOMEMLIMIT"
 	passthroughPortEnvVar = "PASSTHROUGH"
+
+	sidecarGoMaxProcs                  = "1"
+	sidecarGoMemLimitRequestPercentage = 90
```
markmandel, line 72: These seem like good defaults, but I wonder if this should be configurable, just in case. 🤔 WDYT/

markmandel:
Is that a legit test failure? It looks like one

markmandel:
/gcbrun

Checking if the error is the same. Curious to get this in, and see it in production.

markmandel:
Test failure is consistent unfortunately 😢

markmandel:
@mnthe This is an interesting addition - did you want to continue with this work?

markmandel:
Hey - so I'm going to close this to keep things clear. Please feel free to reopen or re-run if you want to revisit.

PR #4570, opened 2026-05-16
build: pin codelytv/pr-size-labeler to commit SHA in pull_request_target workflow

XananasX7:
/kind cleanup

  **What type of PR is this?**

  /kind cleanup

  Pins `codelytv/pr-size-labeler` to full commit SHA instead of mutable `v1` tag in the `labeler.yml` workflow.

  This workflow uses `pull_request_target` trigger which runs with write permissions on every PR from forks. Using a mutable tag reference means any compromise of the upstream action repository (tag move,
  account takeover) would execute arbitrary code with write access to this repository.

  Pinning to SHA ensures the action content is immutable and prevents supply chain attacks via tag manipulation.

  N/A — proactive security hardening

  This follows GitHub's security best practice of pinning actions to full SHA in privileged workflows. See:
  https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -25,7 +25,7 @@ jobs:
       pull-requests: write
     name: Label the PR size
     steps:
-      - uses: codelytv/pr-size-labeler@v1
+      - uses: codelytv/pr-size-labeler@56070d7fa4db29f9f14e8002a08ed8e42412d4dd # v1
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           xs_max_size: '10'


XananasX7:
I agree to the DCO (Developer Certificate of Origin) for this contribution.

XananasX7:
## Summary

This PR pins `codelytv/pr-size-labeler` to a full commit SHA in the `pull_request_target` workflow.

### Why this matters
The `pull_request_target` trigger gives this workflow base-repository context. The `codelytv/pr-size-labeler` action uses a mutable version tag, making it a viable supply chain attack vector — a compromised tag could run arbitrary code with repository write access on every PR.

### Fix
Pinned to the exact commit SHA with a version comment. Minimal change, no logic impact. DCO signed. Ready for review.

markmandel:
https://github.com/CodelyTV/pr-size-labeler/commit/56070d7fa4db29f9f14e8002a08ed8e42412d4dd also not a commit. Closing and banning account due to security concerns.

PR #4596, opened 2026-06-01
fix: nil pointer panic in webhook and wasm FleetAutoscaler handlers

### LINKED ISSUE #4555: FleetAutoscaler webhook and Wasm handlers panic when response body omits 'response' field

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

applyWebhookPolicy and applyWasmPolicy both dereferenced Response.Scale without checking if Response was nil. When a webhook or Wasm handler returns a JSON body that omits the 'response' field, json.Unmarshal succeeds with Response as a nil pointer, causing a panic that crashes controller replicas and halts FleetAutoscaler reconciliation temporarily.

Both sites now return an explicit error when Response is nil. Adds a scaleEmpty export to the autoscaler-wasm example (testing only) and regression tests for both code paths.

Closes #4555

Replaces #4556 since it's been a month.


lacroixthomas (approved):
LGTM !

PR #4600, opened 2026-06-01
pin GitHub Actions to commit SHAs

XiDianJiale:
What type of PR is this?
/kind cleanup

What this PR does / Why we need it:
Updates the GitHub Action references across the .github/workflows/ directory to use immutable commit SHAs instead of mutable tags (e.g., @v4, @main). Inline comments with the original version tags (like # v4) are preserved for easier future bumps.

Why: To ensure deterministic CI executions and fully align Agones with CNCF software supply chain security standards (specifically the OpenSSF Pin-Dependencies check).

Which issue(s) this PR fixes:
None.

Special notes for your reviewer:
Hi team, my studio is currently using Agones to orchestrate dedicated game servers for our multiplayer titles in production.

While conducting a recent internal software supply chain audit and referencing Agones' CI setups for our own custom SDK server build pipelines, we noticed that a few workflows (such as the fossa scanner, label-pr, and stale issues actions) were still using floating tags.

Since CNCF projects generally enforce strict SHA pinning to prevent tag-hijacking, and we rely heavily on the stability of this framework, I wanted to contribute this patch upstream to help tighten our shared infrastructure. All SHAs were resolved directly via git ls-remote from their upstream remotes. Thanks for the awesome project!

### DIFF
--- a/.github/workflows/close.yaml
+++ b/.github/workflows/close.yaml
@@ -29,7 +29,7 @@ jobs:
     permissions:
       issues: write
     steps:
-      - uses: actions/stale@v8
+      - uses: actions/stale@1160a2240286f5da8ec72b1c0816ce2481aabf84 # v8
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           days-before-issue-stale: 30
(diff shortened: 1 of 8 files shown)


markmandel:
Closing, as it looks like there's malicious commits in here. I'm going to do this for real, as this is the second time.

PR #4621, opened 2026-06-15
Issue-4589: Promote RollingUpdateFix feature gate from Beta to Stable

### LINKED ISSUE #4589: Promote `RollingUpdateFix` Feature Gate from Beta to Stable
## What does this feature do?

`RollingUpdateFix` corrects the fleet controller's rolling update logic. Without it, `cleanupUnhealthyReplicas` and `rollingUpdateRestFixedOnReady` calculate scale-down capacity using `fleet.Spec.Replicas` as the baseline, which can prematurely scale down old GameServerSets. With it enabled (the fixed path, implemented in `pkg/fleets/controller_rollingupdatefix.go`):
- Scale-down capacity is calculated from `Status.ReadyReplicas` rather than `Spec.Replicas`
- All replica states (Allocated, Reserved, Shutdown) are considered when determining how many can safely be removed
- Old GameServerSets are only scaled down once new ones are healthy and ready (`Spec.Replicas == Status.Replicas`)

**Current state:** Beta (enabled by default since 1.50.0)

## Acceptance Criteria / Tasks

- [ ] **Manual audit:** Before starting, grep the repo for `FeatureRollingUpdateFix` and `RollingUpdateFix` to confirm all usage sites — the list below may not be exhaustive
- [ ] `pkg/fleets/controller.go` — Remove the `FeatureRollingUpdateFix` gate at line 598; delete the old (pre-fix) code paths entirely; route unconditionally to the fix implementations
- [ ] `pkg/fleets/controller_rollingupdatefix.go` — Rename functions (drop the `RollingUpdateFix` suffix) and inline/move them into `controller.go`; delete this file once done
- [ ] `pkg/fleets/controller_test.go` — **Delete** the tests that validate old (pre-fix) behavior and skip when the feature is enabled (lines ~126, ~205, ~1163). Remove the disabled-feature test case (~line 1721) and de-gate the enabled-feature cases (~lines 1763, 1778, 1793)
- [ ] `test/e2e/fleet_test.go` — Simplify the conditional assertion (~line 1889) to always expect 2 GameServerSets (the "feature enabled" branch)
- [ ] `pkg/util/runtime/features.go` — Move `FeatureRollingUpdateFix` from `// Beta features` to `// Stable features` (keep in `featureDefaults` as `true` so old gate strings still parse)
- [ ] `install/helm/agones/defaultfeaturegates.yaml` — Move `RollingUpdateFix` from Beta to Stable section
- [ ] `build/Makefile` — Remove `RollingUpdateFix` from `BETA_FEATURE_GATES`
- [ ] `cloudbuild.yaml` — Remove `RollingUpdateFix=false` from the inverted e2e-runner config
- [ ] `site/content/en/docs/Guides/feature-stages.md` — Move row from Beta table to Stable table
- [ ] `test/upgrade/versionMap.yaml` — Add entry showing RollingUpdateFix moves to stable gates in this release

## Verification

- `make lint test-go` passes; no tests skip on `RollingUpdateFix`
- `test/e2e/fleet_test.go` rolling-update test no longer has a conditional assertion path
- Setting `RollingUpdateFix=true` or `RollingUpdateFix=false` in the feature gate string must not cause a parse error (covered by keeping the const in `featureDefaults`)

Oliver-ke:
**What type of PR is this?**

/kind cleanup

Move the `RollingUpdateFix` feature gate from Beta to Stable

Closes #4589

Y - Used it to easily navigate the codebase

File `TestRollingUpdateOnReady` was deleted. It never referenced the feature gate, but because it didn't initialize the feature map it was silently exercising the old (pre-fix) behavior, so it failed once the old code path was removed. The fixed behavior is already covered by `TestControllerRollingUpdateDeployment`

Testing: "make test-go" completes successfully


pkg/fleets/controller.go
```diff
@@ -595,12 +597,9 @@ func (c *Controller) rollingUpdateRestFixedOnReady(ctx context.Context, fleet *a
 	if len(rest) == 0 {
 		return nil
 	}
-	if runtime.FeatureEnabled(runtime.FeatureRollingUpdateFix) {
-		return c.rollingUpdateRestFixedOnReadyRollingUpdateFix(ctx, fleet, active, rest)
-	}
```
lacroixthomas, line 600: We might want to do the way around, keeping only this and removing the rest of it

pkg/fleets/controller_rollingupdatefix.go
```diff
@@ -1,163 +0,0 @@
-package fleets
```
lacroixthomas, line 1: We most probably want to keep this file, it's used from the controller.go

pkg/fleets/controller.go
```diff
@@ -559,19 +559,21 @@ func (c *Controller) cleanupUnhealthyReplicas(ctx context.Context, rest []*agone
...
 			continue
 		}
-		if gsSet.Spec.Replicas == gsSet.Status.ReadyReplicas {
+		if gsSet.Spec.Replicas <= healthy {
 			// no unhealthy replicas found, no scaling required.
 			continue
 		}
 
-		scaledDownCount := int32(integer.IntMin(int(maxCleanupCount-totalScaledDown), int(gsSet.Spec.Replicas-gsSet.Status.ReadyReplicas)))
+		scaledDownCount := int32(integer.IntMin(int(maxCleanupCount-totalScaledDown), int(gsSet.Spec.Replicas-healthy)))
```
lacroixthomas: Not certain about why these are changed - also other part on this file ? 🤔 
We should keep it as it is - as it's already been in beta and is working well

pkg/fleets/controller_test.go
```diff
@@ -1533,110 +1356,6 @@ func TestControllerRollingUpdateDeploymentGSSUpdateFailedErrExpected(t *testing.
 	assert.EqualError(t, err, "error updating gameserverset inactive: random-err")
 }
 
-func TestRollingUpdateOnReady(t *testing.T) {
```
lacroixthomas, line 1536: We should be able to keep this unit tests - not related to the featuregate RollingUpdateFix disabled

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -52,6 +51,7 @@ The current set of `alpha`, `beta` and `stable` feature gates include:
 |------------------------------------------------------------------------------------------|----------------------------|---------|----------|--------|
 | [DisableResyncOnSDKServer](https://github.com/agones-dev/agones/issues/3377)             | `DisableResyncOnSDKServer` | Enabled | `Stable` | 1.49.0 |
 | [Support Passthrough on GKE Autopilot](https://github.com/agones-dev/agones/issues/3721) | `AutopilotPassthroughPort` | Enabled | `Stable` | 1.54.0 |
+| [Rolling Update Fixes](https://github.com/agones-dev/agones/issues/3688)                 | `RollingUpdateFix`         | Enabled | `Stable` | 1.50.0 |
```
lacroixthomas: About this part, you will need to add shortcodes around it, as it will only be available as stable starting from next release but not in Beta since the next one as well
https://agones.dev/site/docs/contribute/documentation-editing-contribution/#within-a-page
We usually copy the whole table and wrap them with shortcodes (< 1.59.0 and > 1.59.0 for example)

As an example, it's in another PR as well:
https://github.com/agones-dev/agones/pull/4609/changes#diff-6a76655b1e397b2f3b9955ebe19b8263aa78cd04aed336f6a737d9e408eac30c
For the file `site/content/en/docs/Guides/feature-stages.md`

lacroixthomas (commented):
Hi thanks for the contribution ! 
I left a couple of feedbacks, let me know what you think about it

site/content/en/docs/Guides/feature-stages.md
Oliver-ke: Cool, check the docs, I'd add that

pkg/fleets/controller_test.go
Oliver-ke: Good callout, and I went back to double-check this one because I had the same instinct at first.

Here is the thing,

This test never initializes the feature gates (no ParseFeatures call, and there's no TestMain in the package), so featureGates is a nil map, and FeatureEnabled(RollingUpdateFix) reads back false. That means before this PR, the test was actually running the old, pre-fix branch the whole time, just that it never mentions the gate by name, so it's easy to miss.

Let me know what you think. I'm happy to keep the coverage if you'd prefer. Just need to rewrite these cases against the fixed behavior.

pkg/fleets/controller.go
markmandel: +1, I looked at the final code:

https://github.com/Oliver-ke/agones/blob/9f733dfcf1aa711068e3e267ffae83e71e33398a/pkg/fleets/controller.go#L554-L564

```
func (c *Controller) rollingUpdateRestFixedOnReady(ctx context.Context, fleet *agonesv1.Fleet, active *agonesv1.GameServerSet, rest []*agonesv1.GameServerSet) error {
	if len(rest) == 0 {
		return nil
	}
	return c.rollingUpdateRestFixedOnReadyRollingUpdateFix(ctx, fleet, active, rest)
}

// rollingUpdateRest applies the rolling update to the inactive GameServerSets
func (c *Controller) rollingUpdateRest(ctx context.Context, fleet *agonesv1.Fleet, active *agonesv1.GameServerSet, rest []*agonesv1.GameServerSet) error {
	return c.rollingUpdateRestFixedOnReady(ctx, fleet, active, rest)
}
```

Which when you look at it, you can see how there is unnecessary layering of functions that can go away.

pkg/fleets/controller_test.go
markmandel: If there's no feature gate set - the test should run successfully under whatever flags are set -- so it should be kept for sure.

markmandel:
/gcbrun

First look - seems comprehensive!

markmandel:
/gcbrun

Autopilot flakes.

site/content/en/docs/Guides/feature-stages.md
```diff
@@ -52,6 +53,28 @@ The current set of `alpha`, `beta` and `stable` feature gates include:
...
+| [Extend Webhook autoscaler to send fleet metadata with the request](https://github.com/agones-dev/agones/issues/3951)  | `FleetAutoscaleRequestMetaData`    | Enabled | `Beta` | 1.54.0 |
+| [Sidecar Containers](https://github.com/agones-dev/agones/issues/3642)                                                 | `SidecarContainers`                | Enabled | `Beta` | 1.56.0 |
+
+### Stable Features
+
+| Feature Name                                                                             | Gate                       | Default | Stage    | Since  |
+|------------------------------------------------------------------------------------------|----------------------------|---------|----------|--------|
+| [DisableResyncOnSDKServer](https://github.com/agones-dev/agones/issues/3377)             | `DisableResyncOnSDKServer` | Enabled | `Stable` | 1.49.0 |
+| [Support Passthrough on GKE Autopilot](https://github.com/agones-dev/agones/issues/3721) | `AutopilotPassthroughPort` | Enabled | `Stable` | 1.54.0 |
+| [Rolling Update Fixes](https://github.com/agones-dev/agones/issues/3688)                 | `RollingUpdateFix`         | Enabled | `Stable` | 1.50.0 |
```
markmandel:
```suggestion
| [Rolling Update Fixes](https://github.com/agones-dev/agones/issues/3688)                 | `RollingUpdateFix`         | Enabled | `Stable` | 1.59.0 |
```
Since that's when it'll actually go to stable.

markmandel (commented):
Found one small thing to fix, and otherwise this is good to go.

site/content/en/docs/Guides/feature-stages.md
Oliver-ke: Updating now..

Oliver-ke: Updated

markmandel (approved):
Nice!

markmandel:
Oh no conflicts! @Oliver-ke - can you resolve them, then this is good to merge.

Oliver-ke:
@markmandel conflict resolved

markmandel (approved):
/gcbrun

Let's go.

PR #4625, opened 2026-06-18
fix: Fix flapping with list-based FleetAutoscaler

rhysperry111:
### **What type of PR is this?**

/kind bug

### **What this PR does / Why we need it**:

There is an issue caused by inconsistencies in FleetAutoscaler upscale logic and downscale logic, resulting in flapping when using a List based policy. This was reported [here](https://agones.slack.com/archives/C9DGM5DS8/p1779029384013789) on Slack.

---

The root cause of the problem is due to the way that downscale decisions are made. The downscale logic:
1. Gets the list of current GameServers
2. Simulates a deletion
3. Checks whether doing such would cause a server to need to be added again according to the list policy
 
The first two steps were implemented in a way that does not match how Fleets will actually choose to delete a server, as it doesn't exclude GameServers that would normally be considered undeletable. This inconsistency between the simulated deletion and actual deletion logic is what leads to constant upscaling and downscaling (flapping).

---

The first commit on this PR implements a test to reproduce the issue described, and to help check for regressions in the future.

The second commit on the PR implements the logic in FleetAutoscaler to correctly check which servers would be deleted on a downscale.

### **Which issue(s) this PR fixes**:

Addresses the issue raised on Slack [here](https://agones.slack.com/archives/C9DGM5DS8/p1779029384013789).

### **Did you use AI tools in preparing this PR?**:

No

### **Special notes for your reviewer:**

A build of this code was tested by the user that originally reported the issue on Slack. They are able to confirm that it fixes the original issue.

### DIFF
--- a/pkg/fleetautoscalers/fleetautoscalers.go
+++ b/pkg/fleetautoscalers/fleetautoscalers.go
@@ -694,17 +694,23 @@ func isScheduleActive(s *autoscalingv1.SchedulePolicy, currentTime time.Time) bo
 	return false
 }
 
-// getSortedGameServers returns the list of Game Servers for the Fleet in the order in which the
+// getSortedDeletableGameServers returns the list of Game Servers for the Fleet in the order in which the
 // Game Servers would be deleted.
-func getSortedGameServers(f *agonesv1.Fleet, gameServerNamespacedLister listeragonesv1.GameServerNamespaceLister,
+func getSortedDeletableGameServers(f *agonesv1.Fleet, gameServerNamespacedLister listeragonesv1.GameServerNamespaceLister,
 	nodeCounts map[string]gameservers.NodeCount) ([]*agonesv1.GameServer, error) {
 	gsList, err := fleets.ListGameServersByFleetOwner(gameServerNamespacedLister, f)
 	if err != nil {
 		return nil, err
 	}
 
-	gameServers := gssets.SortGameServersByStrategy(f.Spec.Scheduling, gsList, nodeCounts, f.Spec.Priorities)
-	return gameServers, nil
+	deletable := make([]*agonesv1.GameServer, 0, len(gsList))
+	for _, gs := range gsList {
+		if gs.IsDeletable() {
+			deletable = append(deletable, gs)
+		}
+	}
+
+	return gssets.SortGameServersByStrategy(f.Spec.Scheduling, deletable, nodeCounts, f.Spec.Priorities), nil
 }
 
 // isLimited indicates that the calculated scale would be above or below the range defined by
@@ -738,7 +744,7 @@ func scaleDownLimited(f *agonesv1.Fleet, gameServerNamespacedLister listeragones
 	nodeCounts map[string]gameservers.NodeCount, key string, isCounter bool, replicas int32,
 	aggCapacity, maxCapacity int64) (int32, bool, error) {
 	// Game Servers in order of deletion on scale down
-	gameServers, err := getSortedGameServers(f, gameServerNamespacedLister, nodeCounts)
+	gameServers, err := getSortedDeletableGameServers(f, gameServerNamespacedLister, nodeCounts)
 	if err != nil {
 		return 0, false, err
 	}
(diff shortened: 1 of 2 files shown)


markmandel:
/gcbrun

Let's run it!

pkg/fleetautoscalers/fleetautoscalers.go
lacroixthomas: Neat, just to update the comment to `getSortedDeletableGameServers` hehe 😄 

```
// Exit early if we're already at MinCapacity to avoid calling getSortedDeletableGameServers if unnecessary
```

rhysperry111: Good spot, thanks :D

rhysperry111:
It passed yesterday and all that's changed since is a rebase and a comment :sob:

markmandel:
/gcbrun

That's a new flake - haven't seen that before. Let's run it again, while I run this in the background.

lacroixthomas (approved):
LGTM

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -2464,6 +2464,109 @@ func TestApplyListPolicy(t *testing.T) {
 	}
 }
 
+// TestApplyListPolicyFlapping reproduces a List based FleetAutoscaler flapping bug caused by inconsistencies in upscale/downscale logic
+func TestApplyListPolicyFlapping(t *testing.T) {
+	require.NoError(t, utilruntime.ParseFeatures(string(utilruntime.FeatureCountsAndLists)+"=true"))
```
markmandel, line 2474:
```suggestion

  utilruntime.FeatureTestMutex.Lock()
  defer utilruntime.FeatureTestMutex.Unlock()

	require.NoError(t, utilruntime.ParseFeatures(string(utilruntime.FeatureCountsAndLists)+"=true"))
```

Since we're setting feature flags, we don't want them to flipflop mid-test.

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -2464,6 +2464,109 @@ func TestApplyListPolicy(t *testing.T) {
...
+	m.AgonesClient.AddReactor("list", "gameservers", func(_ k8stesting.Action) (bool, runtime.Object, error) {
+		return true, &agonesv1.GameServerList{Items: gsList}, nil
+	})
+	informer := m.AgonesInformerFactory.Agones().V1()
+	_, cancel := agtesting.StartInformers(m, informer.GameServers().Informer().HasSynced)
+	defer cancel()
+	lister := informer.GameServers().Lister().GameServers("default")
+
+	// Scale up from 2 -> 3 to satisfy the buffer. This is correct and expected.
+	replicas, _, err := applyCounterOrListPolicy(nil, lp, scaleUpFleet, lister, nc)
```
markmandel: Would be nice to assert the `limited` value that comes back, rather than skip it with _. Especially since you have edited the logic on limited.

Might also want to add a test that covers limited true/false values if we don't cover it already.

pkg/fleetautoscalers/fleetautoscalers_test.go
```diff
@@ -2464,6 +2464,109 @@ func TestApplyListPolicy(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, int32(3), replicas, "should scale up to 3 to satisfy the buffer")
+
+	// At 3 servers the fleet should be stable. The bug instead scales back down to
+	// 2 by accounting for the un-deletable Allocated server, which causes flapping.
+	replicas, _, err = applyCounterOrListPolicy(nil, lp, scaleDownFleet, lister, nc)
```
markmandel: Would be nice to asset the `limited` value that comes back, rather than skip it with _.

markmandel (commented):
Found a couple of small things - but otherwise, LGTM.

pkg/fleetautoscalers/fleetautoscalers_test.go
rhysperry111: Ooh yeah good shout - thanks <3

pkg/fleetautoscalers/fleetautoscalers_test.go
rhysperry111: Will assert that value 🫡

And yeah I'll get another test written up that'll hit the limited branch as well.

pkg/fleetautoscalers/fleetautoscalers_test.go
rhysperry111: Will assert that value 🫡

rhysperry111:
Reworked the tests to be a bit more data-driven (with some test cases that also hit the limited paths). Also covers asserting the limited response. Thanks Mark and Thomas for the pointers :D

markmandel (approved):

rhysperry111:
Yippee!

PR #4645, opened 2026-07-04
fix(util/https): redact sensitive headers before logging requests

herdiyana256:
## Summary
Reported via OSS VRP (Issue 531238137).

LogRequest() attached the complete, unredacted http.Header map to every log entry it produced. ErrorHTTPHandler calls it for any non-nil error returned by a wrapped handler, including the plain error processAllocationRequest returns on ordinary capacity exhaustion, so this fired on routine conditions rather than only on crafted or malicious requests.

The resource handler pattern registered in apiserver.go for GameServerAllocation, one of Agones' security sensitive endpoints, is wired through this exact path, so a caller's Authorization header or any API key header ended up in cleartext logs whenever an allocation request failed for any reason, which happens under ordinary capacity exhaustion in a busy fleet.

## Fix

Added a small redaction step in pkg/util/https/https.go: known sensitive header names (Authorization, Proxy-Authorization, Cookie, Set-Cookie, X-Api-Key) get their values replaced with a fixed placeholder before the header map is attached to the log entry. Header names are preserved so the log is still useful for seeing which headers were present on a request, only the values that commonly carry credentials are redacted.

## Testing

Added a test that sends a request with Authorization and X-Api-Key headers through LogRequest() and asserts both are redacted while an unrelated header (Content-Type) passes through unchanged. Verified locally with a small program exercising the exact wiring used for the GameServerAllocation handler (ErrorHTTPHandler wrapping a handler that returns the same plain-error condition as an allocation failure): before this change, the captured log output contained both secret values in full; after this change, both appear as REDACTED and the rest of the log entry is unaffected. Existing tests in the package continue to pass.


markmandel:
There are none of these heads that come across the APIServer ... so I'm wondering why this matters?

herdiyana256:
Thanks @markmandel. :)

You're right that requests routed through the Kubernetes API aggregation layer don't carry the caller's Authorization header, since kube-apiserver strips it and replaces it with X-Remote-User and X-Remote-Group before forwarding to the extensions server. That path is indeed not affected in the way I originally described.

That said, there are two paths where the header logging still reaches real credential values:
1. The agones-allocator service: cmd/allocator/main.go exposes a gRPC and HTTP 
   endpoint that game clients call directly with bearer tokens and mTLS client 
   certificates. Requests on that path arrive at LogRequest with the full, 
   unredacted r.Header including Authorization. ErrorHTTPHandler is used there 
   too, so any error on that endpoint writes the header map to the log.

2. The 404 handler in https.go (line 67): LogRequest is also called directly 
   in the default handler for unrecognized paths, unconditionally on every 404 
   regardless of whether the request came via aggregation or directly.

Even scoped only to the aggregation-layer path, the X-Remote-User, X-Remote-Group, and X-Remote-Extra-* headers forwarded by kube-apiserver are themselves identity information. Redacting Authorization while logging those in cleartext is a narrower issue but not a non-issue.

The fix itself is a three-line helper that costs nothing at runtime and degrades gracefully: unknown headers still pass through unchanged, so there is no risk of over-redacting useful debug information. If the allocator path alone is considered sufficient justification, I'm happy to keep the scope there and update the PR description accordingly.

markmandel (approved):

PR #4663, opened 2026-07-28
fix: mark GameServer Unhealthy when the game container terminates and the Pod stays Running

GrowlyX:
**What type of PR is this?**

/kind bug

I have a cluster with a Fleet that provisiones GameServers with a game container and several other sidecar containers. We had an issue where our main game container OOMKilled, but the pod stayed and was not replaced by Agones.

In `failedContainer`, when SidecarContainers is enabled, then it falls back to the Kubernetes behavior which requires every container in the pod to fail before the pod is marked as Failed. 

This PR makes it so `failedContainer` reports the pod as Failed when the primary game container has terminated.

N/A

Used for codegen & verifying the fix works

N/A


Reasonably:
We hit this exact case in production with Agones v1.57 and `SidecarContainers` enabled.

The container referenced by `spec.container` terminated with exit code 137. Another regular container remained running, so the Pod stayed `Running` and the GameServer remained `Allocated`.

With `restartPolicy: Never`, a failed liveness probe can terminate the game container, but Kubernetes will not restart it. Once the container has terminated, its `/gshealthz` probe also stops. In our case the SDK sidecar recovered afterward, but there was no longer a running game container or probe to drive the health lifecycle.

Since `failedContainer()` returned false and the Pod never reached `Failed`, Agones had no remaining path to move the GameServer to `Unhealthy`.

Using the designated game container as the lifecycle signal is the behavior we expected. A terminated game container cannot recover under `restartPolicy: Never`, even if the rest of the Pod is still running.

markmandel:
Sorry about not catching this PR -- was head deep in the final parts of CNCF migration.

So if I get this right -- it's basically because if you are running a non-sidecar container in your Pod, if the gameserver dies, since there is still a running container, K8s keeps everything running, and that's not what we want!

My only thought on this would be - it would be great to have a test for this in /test/e2e -- run the simple-game-server with some kind of generic container (a pause container?) also in the pod, cause the simple-game-server to crash, and it should show the problem happening.

Not seeing anything super huge jump out at me otherwise with the changes outlined though. Makes sense to me. Sad we still have to look for this stuff 😄

markmandel:
/gcbrun 

Since we should check, see how it goes.

markmandel:
/gcbrun

We fixed that issue - let's run again.

GrowlyX:
> Sorry about not catching this PR -- was head deep in the final parts of CNCF migration.
> 
> So if I get this right -- it's basically because if you are running a non-sidecar container in your Pod, if the gameserver dies, since there is still a running container, K8s keeps everything running, and that's not what we want!
> 
> My only thought on this would be - it would be great to have a test for this in /test/e2e -- run the simple-game-server with some kind of generic container (a pause container?) also in the pod, cause the simple-game-server to crash, and it should show the problem happening.
> 
> Not seeing anything super huge jump out at me otherwise with the changes outlined though. Makes sense to me. Sad we still have to look for this stuff 😄

Sounds good, I'll get that added.

markmandel:
Please fix DCO 👍🏻 need to sign all your commits.

test/e2e/gameserver_test.go
```diff
@@ -489,6 +489,69 @@ func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
...
+
+	gsClient := framework.AgonesClient.AgonesV1().GameServers(framework.Namespace)
+	defer gsClient.Delete(ctx, readyGs.ObjectMeta.Name, metav1.DeleteOptions{}) // nolint: errcheck
+
+	address := fmt.Sprintf("%s:%d", readyGs.Status.Address, readyGs.Status.Ports[0].Port)
+
+	// keep crashing, until we move to Unhealthy. Solves potential issues with controller Informer cache
+	// race conditions in which it has yet to see a GameServer is Ready before the crash.
+	var stop int32
+	defer func() {
```
markmandel: Rather than this - let's do an https://pkg.go.dev/github.com/stretchr/testify/require#EventuallyWithT - which is what we do elsewhere.

Less fancy and much easier to reconcile.

markmandel:
ooh, so this changes some behaviour here, which is interesting

```
VERBOSE: time="2026-08-10 23:14:47.933" level=info msg="Sending CRASH 0" test=TestGameServerPodCompletedAfterCleanExit
VERBOSE: time="2026-08-10 23:14:49.896" level=info msg="checking if GameServer exists" deletionTimestamp="<nil>" error="<nil>" gs=game-server7dz8h state=Unhealthy test=TestGameServerPodCompletedAfterCleanExit
VERBOSE:     gameserver_test.go:605: 
VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:608
VERBOSE:         	            				/usr/local/go/src/runtime/asm_amd64.s:1771
VERBOSE:         	Error:      	Should be true
VERBOSE:     gameserver_test.go:605: 
VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:605
VERBOSE:         	Error:      	Condition never satisfied
VERBOSE:         	Test:       	TestGameServerPodCompletedAfterCleanExit
```

That now fails.

https://console.cloud.google.com/cloud-build/builds/5afbdf30-bfe3-43f9-b4ef-d9b04493596d;step=2?project=agones-images

Clean exit (exit 0), shouldn't go to unhealthy.

test/e2e/gameserver_test.go
GrowlyX: Gotcha, will do this

GrowlyX:
> ooh, so this changes some behaviour here, which is interesting
> 
> ```
> VERBOSE: time="2026-08-10 23:14:47.933" level=info msg="Sending CRASH 0" test=TestGameServerPodCompletedAfterCleanExit
> VERBOSE: time="2026-08-10 23:14:49.896" level=info msg="checking if GameServer exists" deletionTimestamp="<nil>" error="<nil>" gs=game-server7dz8h state=Unhealthy test=TestGameServerPodCompletedAfterCleanExit
> VERBOSE:     gameserver_test.go:605: 
> VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:608
> VERBOSE:         	            				/usr/local/go/src/runtime/asm_amd64.s:1771
> VERBOSE:         	Error:      	Should be true
> VERBOSE:     gameserver_test.go:605: 
> VERBOSE:         	Error Trace:	/go/src/agones.dev/agones/test/e2e/gameserver_test.go:605
> VERBOSE:         	Error:      	Condition never satisfied
> VERBOSE:         	Test:       	TestGameServerPodCompletedAfterCleanExit
> ```
> 
> That now fails.
> 
> https://console.cloud.google.com/cloud-build/builds/5afbdf30-bfe3-43f9-b4ef-d9b04493596d;step=2?project=agones-images
> 
> Clean exit (exit 0), shouldn't go to unhealthy.

ahhh, will fix this yeah

markmandel:
Actually maybe a silly question (I'm trying real hard to not have to check container statuses and terminaton states, the more I look at this code -- I'd love to not have to worry about that, it makes things so much easier).

Is there a reason why the failing health check doesn't end up terminating the pod, and thereby the gameserver? (I assume it fails, right?)

markmandel:
Maybe another fun question:

> I have a cluster with a Fleet that provisiones GameServers with a game container and several other sidecar containers.

When you say sidecar container, do you mean sidecar container or _sidecar container_ 😁 

But seriously, is this configured as an init container with a restartPolcy?

I am also kinda wondering if we don't do this implementation, and be explicit that you should use initContainers for other sidecar workloads -- because then shutdown semantics are basically the same as Kubernetes.

But it's possible I'm not understanding a specific use case or architecture pattern that requires a mainline container vs an initContainer setup? Definitely love to learn more here.

pkg/gameservers/health.go
```diff
@@ -182,11 +187,11 @@ func (hc *HealthController) failedContainer(pod *corev1.Pod) bool {
 
 		for _, cs := range pod.Status.ContainerStatuses {
 			if cs.Name == container {
-				if cs.State.Terminated != nil {
+				if cs.State.Terminated != nil && cs.State.Terminated.ExitCode != 0 {
```
markmandel, line 177: ooh nice find!

I was looking for this! This isn't so bad 😄

markmandel:
/gcbrun

Also to double check - https://agones.dev/site/docs/guides/health-checking/#health-failure-strategy -- what you have here is what is actually described in the docs 👍🏻 

I still am curious about my open questions - but with the tweak to only look at non zero exit codes, I think this is pretty good to go.

GrowlyX:
> Actually maybe a silly question (I'm trying real hard to not have to check container statuses and terminaton states, the more I look at this code -- I'd love to not have to worry about that, it makes things so much easier).
> 
> Is there a reason why the failing health check doesn't end up terminating the pod, and thereby the gameserver? (I assume it fails, right?)

I just did a deep dive into our logs + metrics (because the original issue I described was occurring a lot this morning for us). Turns out the failing health check was marking the game server as unhealthy, but it was severely delayed. We have a 60-second health check period, so the 3 failures required to mark the game server unhealthy took 3 minutes. Since in this particular scenario we had all of the game servers on this fleet OOM at once (😭), it took almost 3 minutes for our users to be able to access this specific fleet again because agones never provisioned new game servers.

This PR would bring that way down if someone else also has a large health check period like us.

> Maybe another fun question:
> 
> > I have a cluster with a Fleet that provisiones GameServers with a game container and several other sidecar containers.
> 
> When you say sidecar container, do you mean sidecar container or _sidecar container_ 😁
> 
> But seriously, is this configured as an init container with a restartPolcy?
> 
> I am also kinda wondering if we don't do this implementation, and be explicit that you should use initContainers for other sidecar workloads -- because then shutdown semantics are basically the same as Kubernetes.
> 
> But it's possible I'm not understanding a specific use case or architecture pattern that requires a mainline container vs an initContainer setup? Definitely love to learn more here.

I actually did not have it set up as an init container with a restartPolicy, will definitely do that. I think it should be specified in the docs that other sidecar workloads should be marked as the init container, it would definitely help people when setting up their workloads properly.

I do think this PR is necessary for those who have `health.disabled` set to true. In my case, the SDK caught the failed container after 3 minutes, but with health checking disabled there's no liveness probe attached, so `healthy` returns true. That would mean `failedContainer` is the only system able to catch a dead game container for health.disabled fleets.

GrowlyX:
I just resolved DCO as well!

markmandel:
> I do think this PR is necessary for those who have health.disabled set to true. In my case, the SDK caught the failed container after 3 minutes, but with health checking disabled there's no liveness probe attached, so healthy returns true. That would mean failedContainer is the only system able to catch a dead game container for health.disabled fleets.

No I agree - we should have the PR in it for sure -- I was just curious about the rest of it played out -- and also, waiting minutes for a gameserver be removed should only happen in extreme circumstances. We should be proactive. And this also aligns with what we have in the docs that we say actually happens.

At the very least, we should link out to K8s sidecar docs on how that affects lifecycles as well. If you want to take a stab in this PR, that would be ace (waiting on 1.60.0 to land anyway), if not - I can do once this merges.

GrowlyX:
> > I do think this PR is necessary for those who have health.disabled set to true. In my case, the SDK caught the failed container after 3 minutes, but with health checking disabled there's no liveness probe attached, so healthy returns true. That would mean failedContainer is the only system able to catch a dead game container for health.disabled fleets.
> 
> No I agree - we should have the PR in it for sure -- I was just curious about the rest of it played out -- and also, waiting minutes for a gameserver be removed should only happen in extreme circumstances. We should be proactive. And this also aligns with what we have in the docs that we say actually happens.
> 
> At the very least, we should link out to K8s sidecar docs on how that affects lifecycles as well. If you want to take a stab in this PR, that would be ace (waiting on 1.60.0 to land anyway), if not - I can do once this merges.

Just pushed the documentation updates and updated this branch as well!

pkg/gameservers/health.go
```diff
@@ -156,13 +156,50 @@ func (hc *HealthController) evictedPod(pod *corev1.Pod) bool {
...
+	//
+	// GameServer Pods are created with RestartPolicy: Never (see
+	// agonesv1.GameServer.Pod()), so a terminated game container can never come back.
+	// Report it as failed rather than waiting for a Pod phase transition that will not
+	// happen.
+	//
+	// Only a non-zero exit code counts. A game container that exits cleanly is a normal
+	// shutdown, and is handled by the SucceededController moving the GameServer to
+	// Shutdown - marking it Unhealthy here would win that race and, since Unhealthy is a
+	// terminal state, permanently block the move to Shutdown.
```
markmandel: Nit: This is a huge AI comment - don't mind a smaller one, but for what is pretty self explanatory code, I think this is overkill 😄

site/content/en/docs/Guides/health-checking.md
```diff
@@ -47,10 +47,32 @@ The following is the process for what happens to a `GameServer` when it is unhea
 
 1. The Pod is set to `restartPolicy: Never` by default.
 2. The SDK server sidecar container is set to `restartPolicy: Always`.
-3. If main containers within the Pod fail, the `GameServere` will move to an `Unhealthy` state.
-4. The SDK server sidecar container stays alive for the entire duration of the Pod, and therefore SDK functionality 
-   is always available.
-5. If the SDK sidecar fails, then it will be restarted, assuming the `restartPolicy` remains the default.
+3. If a main container within the Pod fails, the `GameServer` will move to an `Unhealthy` state.
+4. Since the Pod is `restartPolicy: Never`, a terminated game server container can never be restarted, so this happens as soon as the game server container exits with a non-zero exit code, rather than waiting for the Pod to reach the `Failed` phase.
```
markmandel:
```suggestion
4. Assuming the Pod is `restartPolicy: Never`, a terminated game server container can never be restarted, so this happens as soon as the game server container exits with a non-zero exit code.
```

Depends on what people set it as. Also the "rather" isn't needed, since we don't have to compare it to the previous version.

site/content/en/docs/Guides/health-checking.md
```diff
@@ -47,10 +47,32 @@ The following is the process for what happens to a `GameServer` when it is unhea
...
+   * If the game server container exits cleanly (exit code `0`), it is treated as a normal shutdown rather than a failure, and the `GameServer` moves to `Shutdown` once the Pod completes.
+   * The SDK server sidecar container stays alive for the entire duration of the Pod, and therefore SDK functionality is always available.
```
markmandel: Thought: Should these be sub items or are they better as a 5 and 6?

site/content/en/docs/Guides/health-checking.md
```diff
@@ -47,10 +47,32 @@ The following is the process for what happens to a `GameServer` when it is unhea
...
+7. If the SDK sidecar fails, then it will be restarted, assuming the `restartPolicy` remains the default.
+
+### Running Additional Workloads Alongside the Game Server
+
+If you run supporting workloads in the same Pod as your game server, such as log shippers, metrics agents or crash
+dump uploaders, declare them as
```
markmandel: Nit"

```suggestion
dump uploaders, we recommend declaring them as
```

Just switching language to be best practices, rather than prescribed.

site/content/en/docs/Guides/health-checking.md
```diff
@@ -47,10 +47,32 @@ The following is the process for what happens to a `GameServer` when it is unhea
...
+[terminated automatically](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#sidecar-containers-and-pod-lifecycle)
+once the main containers have exited. Declaring supporting workloads as sidecars therefore keeps the Pod's shutdown
+semantics tied to the lifetime of the game server container:
+
+* the game server container exiting is enough to end the Pod, and
+* each sidecar is sent `SIGTERM` *after* the game server container has exited, and has the remainder of
+  `terminationGracePeriodSeconds` to finish any work still in flight.
+
+If you declare these workloads in `containers` instead, a long-lived one will hold the Pod in the `Running` phase after
+the game server container has died. Agones will still move the `GameServer` to `Unhealthy` as described in rule 3
```
markmandel:
```suggestion
if the game server container has crashed or exited unexpectantly. Agones will still move the `GameServer` to `Unhealthy` as described in rule 3
```

Correct me if I'm wrong, but this is primarily an issue in the above scenario? If someone calls `SDK.Shutdown()` then everything works correctly, yeah?

markmandel (commented):
Looks good - some nits on the docs 👍🏻 and you have to remember to sign your commits on every update 😁 but otherwise, looks good to go.

markmandel:
Ooh, new linter found some fun stuff.

markmandel:
/gcbrun

Likely lint issues are already already fixed.

GrowlyX:
Will get these done tomorrow morning 😊

test/e2e/gameserver_test.go
```diff
@@ -490,6 +490,62 @@ func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
...
+	gsClient := framework.AgonesClient.AgonesV1().GameServers(framework.Namespace)
+	defer gsClient.Delete(ctx, readyGs.ObjectMeta.Name, metav1.DeleteOptions{}) // nolint: errcheck
+
+	address := fmt.Sprintf("%s:%d", readyGs.Status.Address, readyGs.Status.Ports[0].Port)
+
+	// keep crashing, until we move to Unhealthy. Solves potential issues with controller Informer cache
+	// race conditions in which it has yet to see a GameServer is Ready before the crash.
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		func() {
+			conn, err := net.Dial("udp", address)
```
markmandel: I'm on mobile, but I think we have a utility function that does all this for you.

GrowlyX:
@markmandel should be good now!

test/e2e/gameserver_test.go
```diff
@@ -490,6 +490,51 @@ func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
...
+	// keep crashing, until we move to Unhealthy. Solves potential issues with controller Informer cache
+	// race conditions in which it has yet to see a GameServer is Ready before the crash.
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		// the game server exits on CRASH without replying, so an error waiting for a reply is expected.
+		_, _ = framework.SendGameServerUDP(t, readyGs, "CRASH")
+
+		current, err := gsClient.Get(ctx, readyGs.ObjectMeta.Name, metav1.GetOptions{})
+		if !assert.NoError(c, err) {
+			return
+		}
```
markmandel:
```suggestion
   require.NoError(c, err) 
```

Does it for you.

test/e2e/gameserver_test.go
```diff
@@ -490,6 +490,51 @@ func TestGameServerUnhealthyAfterReadyCrash(t *testing.T) {
...
+	gs.Spec.Template.Spec.Containers = append(gs.Spec.Template.Spec.Containers, corev1.Container{
+		Name:            "generic",
+		Image:           "registry.k8s.io/pause:3.10",
+		ImagePullPolicy: corev1.PullIfNotPresent,
+	})
+
+	readyGs, err := framework.CreateGameServerAndWaitUntilReady(t, framework.Namespace, gs)
+	if err != nil {
+		t.Fatalf("Could not get a GameServer ready: %v", err)
+	}
```
markmandel:
```suggestion
   require.NoError(t, err)
```

markmandel:
/gcbrun

Minor test nits, but otherwise this is good to go.

Thanks for all the work!

GrowlyX:
> /gcbrun
> 
> Minor test nits, but otherwise this is good to go.
> 
> Thanks for all the work!

Yup, no problem! Just pushed the test fix

markmandel:
>   Get "https://1-56-0.agones.dev": x509: certificate has expired or is not yet valid: current time 2026-08-22T04:58:33Z is after 2026-08-17T20:09:35Z --- site/docs/third-party-content/articles/index.html --> https://1-56-0.agones.dev
  Get "https://1-54-0.agones.dev": x509: certificate has expired or is not yet valid: current time 2026-08-22T04:58:33Z is after 2026-08-17T20:09:35Z --- site/docs/third-party-content/articles/index.html --> https://1-54-0.agones.dev
  Get "https://1-53-0.agones.dev": x509: certificate has expired or is not yet valid: current time 2026-08-22T04:58:33Z is after 2026-08-17T20:09:35Z --- site/docs/third-party-content/articles/index.html --> https://1-53-0.agones.dev
  Get "https://1-52-0.agones.dev": x509: certificate has expired or is not yet valid: current time 2026-08-22T04:58:33Z is after 2026-08-17T20:09:35Z --- site/docs/third-party-content/articles/index.html --> https://1-52-0.agones.dev
site/docs/third-party-content/podcasts/index.html

Gah, not your fault - the certs expired on the old versions of the docs.

GrowlyX:
@markmandel - anything else needed here?

markmandel:
/gcbrun

Sorry! Gamescom!

markmandel:
/gcbrun

We have a new set of linters - let's see if there are any issues.

GrowlyX:
@markmandel looks good here!

markmandel (approved):
Let's go!

PR #4593, opened 2026-06-01
chore: upgrade website to Docsy 0.15.0

Cloud-Architect-Emma:
Upgrades the Agones website from Docsy v0.14.3 to v0.15.0 and bumps the Hugo build image version to match the new requirement.
Changes
site/go.mod : Bump github.com/google/docsy from v0.14.3 to v0.15.0
site/go.sum : Regenerated via cd site && go mod tidy
build/build-image: Dockerfile  Bump HUGO_VER from 0.152.2 to 0.157.0
Breaking Changes Reviewed

Version menu markup : no custom CSS or layout overrides targeting the old version menu classes found in the Agones site
Card shortcode : no card shortcode usage found in the site content
Multilingual links :  not applicable, Agones site is English-only

New Features in Docsy 0.15.0

Agent support (experimental): generates llms.txt and Markdown page output for AI tools
Doc-rooted sites (experimental): publish docs at root URL
Enhanced version menu: supports headings, separators, per-entry page-link behaviour and kind-specific styling

Testing

 cd site && go mod tidy ran successfully with no errors
 Site renders correctly
 Version menu renders correctly on desktop and mobile viewports
 Card shortcode pages render correctly

Closes #4586

PR #4669, opened 2026-08-01
chore: upgrade website to Docsy 0.16.0 and restore the version lock

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix
> /kind release

PR #4593 removed the docsy require block from site/go.mod, leaving the theme unpinned. When Docsy 0.16.0 shipped, site builds floated onto it and broke with:

    unknown output format "llms" for kind "home"

Fixes include:

- pin github.com/google/docsy/theme v0.16.0 in site/go.mod
- point the module import at github.com/google/docsy/theme and drop the docsy/dependencies import; Bootstrap and Font Awesome now come from npm
- bump Hugo to 0.164.0 (Docsy 0.16.0 requires extended >= 0.160.1)
- add a site-deps target running "hugo mod npm pack && npm install", required by both site-server and site-static
- drop the postcss-cli/autoprefixer installs; Docsy 0.16 only invokes PostCSS for RTL languages or when a postcss.config.* is present
- add security.allowContent for text/html; Hugo 0.162 denies HTML content files by default and the site has two
- rename languages.en.languageName to label, deprecated in Hugo 0.158
- set $navbar-padding-x so the navbar brand lines up with page content; Bootstrap gives .navbar no horizontal padding and our navbar.html override lacks Docsy's inner .container-fluid

N/A

Oh yeah.

This will unblock CI!


markmandel:
/gcbrun

Let's try again 🤷🏻 

..what weird thing is going on with the 1.33 clusters. Not having access to the project is very frustrating 😬  (ping @igooch - please work on the migration! )

markmandel:
FUN CONFLICTS

site/go.mod
```diff
@@ -4,7 +4,4 @@ go 1.26.5
 
 require gopkg.in/yaml.v2 v2.4.0
 
-require (
-	github.com/google/docsy v0.15.0 // indirect
-	github.com/google/docsy/dependencies v0.7.2 // indirect
-)
\ No newline at end of file
+require github.com/google/docsy/theme v0.16.0 // indirect
```
igooch: The lock is going to get deleted again the next time someone tidies.

`site/` is a real Go module (`main.go`, `handler.go`) that *also* carries the Hugo module requires, and `go mod tidy` drops any require no Go file imports. I ran it on this branch to be sure:

```console
$ go mod tidy && cat go.mod
module github.com/agones/agones/site

go 1.26.5

require gopkg.in/yaml.v2 v2.4.0
```

The docsy line is simply gone. And our own Go-upgrade checklist does exactly that, twice — `.github/ISSUE_TEMPLATE/golang_version_upgrade.md` lines 27 and 59 run `find . -name 'go.mod' -not -path '*/\.*' -execdir go mod tidy \;`, which walks straight into `site/`. That's a very plausible account of how it came unpinned in the first place, and re-pinning alone won't stop it recurring.

Could we close the loop while we're here? A note in this file plus scoping the checklist commands (`-not -path './site/*'`) would do it:

```gomod
// Hugo theme module: required by site/config.toml, not by any Go import.
// Do NOT run `go mod tidy` in this directory — it will delete the line below
// and unpin the Docsy theme. Use `hugo mod get` / `hugo mod tidy` instead.
require github.com/google/docsy/theme v0.16.0 // indirect
```

site/config.toml
```diff
@@ -168,14 +168,16 @@ desc = "Chat with other project users in #users"
 # 	icon = "fa fa-envelope"
 #         desc = "Discuss development issues around the project"
 
+[security]
+  # Hugo >= 0.162 denies text/html content files by default. Agones has two:
+  # content/en/_index.html and the generated content/en/docs/Reference/agones_crd_api_reference.html
+  allowContent = ['^text/markdown$', '^text/html$']
```
igooch, line 176: Nit, but worth a conscious decision: Hugo's default for `allowContent` is a *deny-only* list (`['! ^text/html$']`), i.e. "allow everything except HTML". Replacing it with an allow-only list flips the semantics to "allow only these two", so asciidoc / org / rst / pandoc content would now be actively rejected rather than merely unused.

Harmless today — `site/content` is 201 `.md` plus the 2 `.html` files you mention, and the build passes — just noting it's a tighter change than "let HTML through", and someone adding a new content format later would hit a fairly opaque failure. A half-line in the comment saying that's deliberate would help.

site/assets/scss/_variables_project.scss
```diff
@@ -13,6 +13,11 @@ $grey:  #888 !default;
 $orange: $secondary;
 $td-sidebar-tree-root-color: #222 !default;
 
+// Bootstrap gives .navbar no horizontal padding; upstream Docsy supplies it via an
+// inner .container-fluid, which our navbar.html override doesn't use. Match the
+// container gutter (--bs-gutter-x: 1.5rem) so the brand lines up with page content.
+$navbar-padding-x: 0.75rem;
```
igooch, line 19: Noting the alternative: Docsy already ships `.td-navbar .td-navbar-container.navbar-is-overflowing { padding: 0 }`, which our markup never triggers because `navbar.html` has no `.td-navbar-container`. Adding that wrapper to the override would keep us tracking upstream rather than carrying a Bootstrap variable override — bigger change though, and this is fine as-is.

igooch (commented):
One note, two nits. Otherwise LGTM.

site/assets/scss/_variables_project.scss
markmandel: Yeah, that was my thought too - may as well just do a small change.

site/go.mod
markmandel: I have a solution that stops it from dissapearing on `go mod tidy` - a tools.go!

site/config.toml
markmandel: Done!

markmandel:
/gcbrun

https://github.com/agones-dev/agones/issues/4686

igooch (approved):
/lgtm

Issue #4421, opened 2026-01-13
Checklist - Moving Agones to CNCF


thisisnotapril:
## REQUIRED BEFORE PROCEEDING WITH ONBOARDING

A "Project Contribution Agreement" must be completed and any existing trademarks **MUST** be transferred to the Linux Foundation **BEFORE** the CNCF staff onboarding tasks can be completed. Other tasks can be done by projects themselves in the meantime.

- [x] Review and understand the [CNCF IP Policy](https://github.com/cncf/foundation/blob/main/charter.md#11-ip-policy). Ensure you are using a CNCF compatible license; inbound projects must use the Apache 2.0 license. Licenses for dependencies are covered separately below.
- [x] Review and understand the [CNCF Third Party License Policy](https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md#cncf-allowlist-license-policy). This policy governs the licenses used by third party libraries in your project. CNCF FOSSA or CNCF Snyk are configured to check that projects are in compliance with this policy. Let us know which service you would prefer to use.
- [x] Review and understand the [LF trademark guidelines](https://www.linuxfoundation.org/legal/trademark-usage). Let the TOC know if you plan to change your project name.
- [ ] Transfer any [trademark and logo assets to the Linux Foundation](https://github.com/cncf/foundation/tree/main/agreements) via the Contribution Agreement. CNCF staff will send this document to the contact emails listed in the Sandbox application.

---

## Review and understand other documents

- [ ] The [Technical Leadership Principles](https://github.com/cncf/toc/blob/main/PRINCIPLES.md#technical-leadership-principles) that outline the expected behavior for any maintainer in a leadership role.
- [ ] The [project proposal process and requirements](https://github.com/cncf/toc/blob/main/process/README.md).
- [ ] The [services available for your project at the CNCF](https://contribute.cncf.io/resources/project-services/).
- [ ] The [online program guidelines](https://github.com/cncf/foundation/blob/main/policies-guidance/online-programs-guidelines.md).
- [ ] The [telemetry data collection and usage policy](https://www.linuxfoundation.org/legal/telemetry-data-policy).
- [ ] Optional: [Book time with CNCF staff](http://project-meetings.cncf.io) to understand CNCF resources available to you, to work through onboarding tasks together, or ask other questions.

## Contribute and transfer other materials

[...]

PR #4676, opened 2026-08-09
build: upgrade golangci-lint to v2.12.2 and migrate config

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

The pinned golangci-lint was v1.64.8. The v1 line is end-of-life, so it receives no new linters or bugfixes. Upgrade to v2.12.2 and migrate .golangci.yml to the v2 schema, keeping the effective linter set the same.

Work on OpenSSF work, as per #4421

Y/N

Config migration (via `golangci-lint migrate`, comments restored by hand):
- `gosimple` is merged into `staticcheck` and is no longer a separate linter.
- `goimports` moves to the new top-level `formatters` section.
- `issues.exclude-rules` -> `linters.exclusions.rules`, `exclude-dirs` -> `linters.exclusions.paths`, `linters-settings` -> `linters.settings`.
- `exclude-use-default: false` is dropped; v2 defaults to no exclusion presets, which is the same behaviour.
- errcheck, govet, ineffassign, staticcheck and unused are no longer listed explicitly, as v2's default `standard` set already enables them.

The upgrade surfaced several latent problems:

- `golangci-lint run ./examples/...` in build/Makefile has never linted anything. `exclude-dirs-use-default` was never set, so it defaulted to true, and v1's built-in skip-dirs include `examples$`. Confirmed by running the v1 binary directly: it produced no output for that target. Under v2 the exclusion no longer applies, so the pass now works and its 16 issues are fixed here.

- `run.deadline` and `run.skip-files` had already been removed in later v1 releases and were being silently ignored, so the configured 5m timeout was never in effect. `golangci-lint migrate` refuses to run until they are cleaned up.

- The default issue caps (50 per linter, 3 per identical message) were hiding the true backlog; each fix revealed more. Both are now set to 0.

Fixes for genuine defects found along the way:

- examples/simple-genai-server: `defer resp.Body.Close()` sat after an early return on the io.ReadAll error, leaking the body on read failure.
- examples/simple-genai-server: `defer ln.Close()` before a `log.Fatalf` in the accept loop could never run.
- test/sdk/go: a duplicated `if err != nil` block, unreachable because the real check above it already calls log.Fatalf (govet nilness).
- pkg/sdkserver: `==` on time.Time compares the monotonic clock and location rather than the instant; use Equal.
- test/e2e: three `fmt.Sprintf("%s:%d")` addresses passed to net.Dial do not work with IPv6 (new Go 1.26 govet hostport analyzer); use net.JoinHostPort.

Two staticcheck checks are disabled with rationale in the config: QF1008 (dropping embedded fields from selectors fights Kubernetes convention) and QF1001 (the controllers deliberately write guards as a negated positive precondition). goconst gains `ignore-tests`, since table-driven tests repeat literals by nature.


examples/simple-genai-server/main.go
```diff
@@ -156,11 +159,11 @@ func main() {
...
 					log.Printf("=== New player %s (id %d) ===", name, fid)
 
-					log.Printf("Creating GenAI Client at endpoint %s (from_id=%d, to_id=%d)", *genAiEndpoint, fid, *toId)
-					genAiConn := initClient(*genAiEndpoint, *genAiContext, "GenAI", *genAiNpc, fid, *toId)
+					log.Printf("Creating GenAI Client at endpoint %s (from_id=%d, to_id=%d)", *genAiEndpoint, fid, *toID)
+					genAiConn := initClient(*genAiEndpoint, *genAiContext, "GenAI", *genAiNpc, fid, *toID)
 
 					log.Printf("%s: Creating client at endpoint %s, sending prompt: %s", name, *simEndpoint, *prompt)
-					simConn := initClient(*simEndpoint, *simContext, name, *simNpc, *toId, *toId)
+					simConn := initClient(*simEndpoint, *simContext, name, *simNpc, *toID, *toID)
```
lacroixthomas, line 166: It's not related to your changes really, but I wonder if we didn't have an issue here, I guess it should be `*fromID, *toID` based on the `initClient(.., fromID int, toID int)` 🤔 
I don't know much about this `simple-genai-server` example, ignore me if I'm wrong hehe

test/sdk/go/sdk-client-test.go
```diff
@@ -106,9 +106,6 @@ func main() {
 	if err != nil {
 		log.Fatalf("Could not set label: %s", err)
 	}
-	if err != nil {
-		log.Fatalf("Error received on watch gameserver %s", err)
-	}
```
lacroixthomas, line 111: That's a fun one 😄  
Double error check hehe

lacroixthomas (approved):

examples/simple-genai-server/main.go
markmandel: You're probably not wrong. That beings said, I don't know if we really care about the example.

Realistically this is an API call from game server 🤷🏻 it's nothing particularly special. We could delete it if we wanted to.

PR #4664, opened 2026-07-29
Upgrade supported Kubernetes versions to 1.34, 1.35, 1.36

Sivasankaran25:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>

/kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

**What this PR does / Why we need it**:

**Which issue(s) this PR fixes**:

Works On #4655

**Did you use AI tools in preparing this PR?**:

Y/N

**Special notes for your reviewer**:

PR #4683, opened 2026-08-11
Release-1.60.0

Sivasankaran25:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

/kind release

Closes #

Y/N


igooch (commented):
agones_crd_api_reference.html:6 and :4972 still carry expiryVersion="1.60.0" / publishVersion="1.60.0", added by #4664. Looks like `make feature-shortcode-update` only looks at *.md files, so this was skipped. Please update manually, and then we can update the `make` command in a separate issue / PR.

igooch (commented):
Thanks for the `agones_crd_api_reference.html` cleanup — I verified it on the preview build and it looks right.

One thing to sort out before merge:

`6824b8dc7` merges `main` into the branch, which pulls in `40410b7aa` (#4677, "add pkg/util/errors"). That landed on `main` after the release branch was cut, so it shouldn't ride along in 1.60.0.

`5cbfa86f5` already sits directly on the cut point, so dropping the merge is:

```
git reset --hard 5cbfa86f5
git cherry-pick 26e415130
git push --force-with-lease
```

This updates #4682 at the same time, since both PRs share this branch as their head and differ only in base. #4682's diff won't change in substance (#4677 already cancels against `main`) — it just loses the merge from its commit list.

Two things to avoid until #4683 merges:

- Please don't click **Update branch** on #4682 — that's what created `6824b8dc7`, and the banner will reappear after the force-push.
- Don't enable auto-merge on #4682 — `pr_update.yml` merges `main` into approved auto-merge PRs on every push to `main`, which lands us right back here.

Merge order is #4683 first, then #4682.

I cut a follow-up issue #4684 (not this PR):

Filing an issue to extend `make feature-shortcode-update` to `.html` so this step doesn't
have to be manual next release — `build/scripts/feature-shortcode-update/main.go:45` skips
anything that isn't `.md`. Same issue covers shortcodes that name an already-released
version and so never match a future cleanup (`rust.md` still has a `1.52.0` pair,
`fleetautoscaler.md` a `1.54.0` pair).

igooch (approved):
/lgtm

vicentefb (approved):
/lgtm

Issue #4510, opened 2026-04-11
Discussion: Migrate from `github.com/pkg/errors` to stdlib `errors` package?


markmandel:
## Overview

Agones currently depends on [`github.com/pkg/errors`](https://github.com/pkg/errors) v0.9.1 for error handling and wrapping. This package predates Go 1.13 (released 2019), which introduced native error wrapping and unwrapping via `fmt.Errorf("...: %w", err)`, `errors.Is`, and `errors.As`.

`pkg/errors` is in [maintenance mode](https://github.com/pkg/errors#readme) — the author considers it feature-complete and defers to the stdlib for new work. Should we migrate to the modern stdlib approach?

This issue is a **community discussion** — please share your thoughts before we commit to any direction.

---

## Current Usage

- **55 source files** in `pkg/`, `cmd/`, and `sdks/go/` import `github.com/pkg/errors`
- Key functions in use:

| `pkg/errors` function | Approximate count | Notes |
|---|---|---|
| `errors.Wrapf(err, "msg %s", arg)` | ~118 | Wraps with formatted message + stack trace |
| `errors.Errorf("msg %s", arg)` | ~118 | Like `fmt.Errorf` but with stack trace |
| `errors.Wrap(err, "msg")` | ~59 | Wraps with message + stack trace |
| `errors.New("msg")` | ~58 | Like stdlib `errors.New` but with stack trace |
| `errors.WithStack(err)` | ~4 | Adds stack trace to an existing error |
| `errors.Cause(err)` | ~3 | Unwraps to the root cause |
| `errors.StackTrace` interface | ~1 | Used in `pkg/util/runtime/runtime.go` for logging |

The stack trace interface is used in [`pkg/util/runtime/runtime.go`](pkg/util/runtime/runtime.go) to extract and log stack traces from errors, and `errors.Cause` is used in [`pkg/util/workerqueue/workerqueue.go`](pkg/util/workerqueue/workerqueue.go) and [`pkg/gameserverallocations/allocator.go`](pkg/gameserverallocations/allocator.go) to inspect wrapped Kubernetes conflict errors.

---

## Pros of Migrating to stdlib

- **One fewer dependency** — removes `github.com/pkg/errors` from `go.mod`/`go.sum` and the `vendor/` tree
- **Idiomatic modern Go** — stdlib error wrapping (`%w`) is now the community standard; contributors familiar with modern Go won't need to learn `pkg/errors` idioms
[...]

PR #4691, opened 2026-08-18
feat: use new errors pkg in pkg/gameservers/

lacroixthomas:
**What type of PR is this?**
/kind cleanup

Use the new error package (https://github.com/agones-dev/agones/blob/main/pkg/util/errors/errors.go) on the pkg/gameservers/* files 

Work on #4510 

Y

Many others PRs will follow

Also, not sure about the unit tests, the expected errors are fine, but matching the exact error message hmm not fully sure we want to keep that

### DIFF
--- a/pkg/gameservers/controller.go
+++ b/pkg/gameservers/controller.go
@@ -286,17 +286,17 @@ func (ext *Extensions) creationMutationHandler(review admissionv1.AdmissionRevie
 
 	newGS, err := json.Marshal(gs)
 	if err != nil {
-		return review, ext.errs.Wrap(err, fmt.Sprintf("error marshalling default applied GameServer %s to json", gs.ObjectMeta.Name))
+		return review, ext.errs.Errorf("error marshalling default applied GameServer %s to json: %w", gs.ObjectMeta.Name, err)
 	}
 
 	patch, err := jsonpatch.CreatePatch(obj.Raw, newGS)
 	if err != nil {
-		return review, ext.errs.Wrap(err, fmt.Sprintf("error creating patch for GameServer %s", gs.ObjectMeta.Name))
+		return review, ext.errs.Errorf("error creating patch for GameServer %s: %w", gs.ObjectMeta.Name, err)
 	}
 
 	jsonPatch, err := json.Marshal(patch)
 	if err != nil {
-		return review, ext.errs.Wrap(err, fmt.Sprintf("error creating json for patch for GameServer %s", gs.ObjectMeta.Name))
+		return review, ext.errs.Errorf("error creating json for patch for GameServer %s: %w", gs.ObjectMeta.Name, err)
 	}
 
 	pt := admissionv1.PatchTypeJSONPatch
@@ -325,7 +325,7 @@ func (ext *Extensions) creationValidationHandler(review admissionv1.AdmissionRev
 	gs := &agonesv1.GameServer{}
 	err := json.Unmarshal(obj.Raw, gs)
 	if err != nil {
-		return review, ext.errs.Wrap(err, fmt.Sprintf("error unmarshalling GameServer json after schema validation: %s", obj.Raw))
+		return review, ext.errs.Errorf("error unmarshalling GameServer json after schema validation: %s: %w", obj.Raw, err)
 	}
 
 	loggerForGameServer(gs, ext.baseLogger).WithField("review", review).Debug("creationValidationHandler")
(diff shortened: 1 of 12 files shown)


pkg/gameservers/controller.go
```diff
@@ -282,17 +286,17 @@ func (ext *Extensions) creationMutationHandler(review admissionv1.AdmissionRevie
 
 	newGS, err := json.Marshal(gs)
 	if err != nil {
-		return review, errors.Wrapf(err, "error marshalling default applied GameServer %s to json", gs.ObjectMeta.Name)
+		return review, ext.errs.Wrap(err, fmt.Sprintf("error marshalling default applied GameServer %s to json", gs.ObjectMeta.Name))
```
markmandel:
```suggestion
		return review, ext.errs.Errorf(err, "error marshalling default applied GameServer %s to json: %w", gs.ObjectMeta.Name, err)
```

I was wondering to myself... do we need an `errors.Wrapf` - but then I realised, we have `Errorf` - so we don't need it!

Same for the rest.

lacroixthomas: Ooooh indeed, will update that 👌🏼

lacroixthomas: Fun thing, I did that pretty much everywhere 😄 (the way around), but that will be way cleaner with Errof - also won't need fmt on some of them

lacroixthomas: I have something locally, but I will come back to them one by one, because the errors.Wrap from before was sometimes returned directly (if the errors was nil, it would have return nil, but if not it would have returned the error) - I need to double check them all first before pushing something else, will poke you back once all good

lacroixthomas: Actually, if you're fine with it I'll add the `errors.Wrapf` - if we want to keep them without checking the err != nil, there is around 80+ of them on the repos, I don't mind one or the other actually
Either just checking for err != nil then doing `Errorf`, or adding a `Wrapf` and keeping them as is

WDYT ? @markmandel 

I'm actually fine with checking the error and using the Errorf, feels easier to follow, but I'm easy on this one

markmandel: Yeah that's a good point on `errors.Wrapf` - if err is nil it does nothing -- yeah, I think let's add it! I like it!

lacroixthomas: Sounds good, i'll add that 👌🏼

pkg/util/errors/errors.go
```diff
@@ -122,6 +122,14 @@ func (e *Errors) Wrap(err error, msg string) error {
 	return fmt.Errorf("%s: %w", e.prefix(msg), err)
 }
 
+// Wrapf returns an error annotating err with a formatted msg, prefixed as described on
+// Errors. The returned error wraps err, so errors.Is and errors.As traverse it.
+//
+// Wrapf returns nil if err is nil.
+func (e *Errors) Wrapf(err error, format string, a ...any) error {
+	return e.Wrap(err, fmt.Sprintf(format, a...))
```
lacroixthomas: Could also be 
```
	if err == nil {
		return nil
	}
	return fmt.Errorf("%s: %w", e.prefix(fmt.Sprintf(format, a...)), err)
```
Might be clearer with the "prefix" part

pkg/gameservers/controller_test.go
```diff
@@ -678,7 +678,7 @@ func TestControllerCreationValidationHandler(t *testing.T) {
 
 		_, err = ext.creationValidationHandler(review)
 		if assert.Error(t, err) {
-			assert.Equal(t, `error unmarshalling GameServer json after schema validation: "WRONG DATA": json: cannot unmarshal string into Go value of type v1.GameServer`, err.Error())
+			assert.Contains(t, err.Error(), `error unmarshalling GameServer json after schema validation: "WRONG DATA": json: cannot unmarshal string into Go value of type v1.GameServer`)
```
markmandel: Just being nit picky on the tests 😄 

```suggestion
			assert.ErrorContains(t, err, `error unmarshalling GameServer json after schema validation: "WRONG DATA": json: cannot unmarshal string into Go value of type v1.GameServer`)
```

Same for the rest.

pkg/gameservers/health_test.go
```diff
@@ -250,7 +250,9 @@ func TestHealthControllerSkipUnhealthyGameContainer(t *testing.T) {
 			result, err := hc.skipUnhealthyGameContainer(gs, pod)
 
 			if len(v.expected.err) > 0 {
-				require.EqualError(t, err, v.expected.err)
+				if assert.Error(t, err) {
+					assert.Contains(t, err.Error(), v.expected.err)
+				}
```
markmandel:
```suggestion
       require.ErrorContains(t, err, v.expected.err) 
```

I assume it needs to be a substring?

pkg/util/errors/errors.go
markmandel: I think I like the nil check first - since if the error is nil, then there is no expansion of the Sprintf string - whereas in the it's current format it always gets translated.

pkg/util/errors/errors_test.go
```diff
@@ -140,3 +140,19 @@ func TestWrap(t *testing.T) {
 	// matches github.com/pkg/errors.Wrap: nil in, nil out
 	assert.NoError(t, pkg.Wrap(nil, "no capacity"))
 }
+
+func TestWrapf(t *testing.T) {
+	sentinel := errors.New("boom")
```
markmandel, line 145: boom.

markmandel (approved):

PR #4701, opened 2026-08-31
feat: use new errors pkg in pkg/gameserverallocations/*

lacroixthomas:
**What type of PR is this?**
/kind cleanup

Use the new error package (https://github.com/agones-dev/agones/blob/main/pkg/util/errors/errors.go) on the pkg/gameserverallocations/* files 

Work on #4510 

Y

### DIFF
--- a/pkg/gameserverallocations/allocation_cache.go
+++ b/pkg/gameserverallocations/allocation_cache.go
@@ -24,11 +24,11 @@ import (
 	informerv1 "agones.dev/agones/pkg/client/informers/externalversions/agones/v1"
 	listerv1 "agones.dev/agones/pkg/client/listers/agones/v1"
 	"agones.dev/agones/pkg/gameservers"
+	"agones.dev/agones/pkg/util/errors"
 	"agones.dev/agones/pkg/util/logfields"
 	"agones.dev/agones/pkg/util/runtime"
 	"agones.dev/agones/pkg/util/workerqueue"
 	"github.com/heptiolabs/healthcheck"
-	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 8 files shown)


pkg/gameserverallocations/allocation_cache.go
```diff
@@ -260,7 +262,7 @@ func (c *AllocationCache) syncCache() error {
 	// build the cache
 	gsList, err := c.gameServerLister.List(labels.Everything())
 	if err != nil {
-		return errors.Wrap(err, "could not list GameServers")
+		return c.errs.Wrap(err, "could not list GameServers")
```
markmandel, line 265: Loving how easy this is with the wrapper.

pkg/gameserverallocations/doc.go
```diff
@@ -15,3 +15,7 @@
 // Package gameserverallocations handles management of
 // allocating Ready GameServers via selectors
 package gameserverallocations
+
+import "agones.dev/agones/pkg/util/errors"
```
markmandel: Interesting. Why in doc.go? I'd assume doc.go would just be.... docs 😄

I don't know if I see a better spot, but ... maybe we rename the file then?

markmandel (approved):
Question around docs.go (maybe a rename on the file?) - but approving - merge when you've worked out what you want to do about that nit.

pkg/gameserverallocations/doc.go
markmandel: Or create a gameserverallocations.go - which seems to be our pattern for generic code for a package.

lacroixthomas: Gonna do that, I put that in doc.go to have something centralised for the package, but indeed, I would also expect it to contains doc and not errors hehe 
Will update that 😄

PR #4716, opened 2026-09-03
feat: use new errors pkg in pkg/util/*

lacroixthomas:
**What type of PR is this?**
/kind cleanup

Use the new error package (https://github.com/agones-dev/agones/blob/main/pkg/util/errors/errors.go) on the pkg/util/* files 

Work on #4510 

Y


pkg/util/httpserver/httpserver.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package httpserver
+
+import "agones.dev/agones/pkg/util/errors"
+
+var errs = errors.FromPackage()
```
markmandel, line 19: Nit: Given there is only one file in this package, any reason to not just throw this in server.go?

markmandel (approved):
One small nit that I don't think is worth blocking on - change if you think you should 👍🏻

pkg/util/httpserver/httpserver.go
lacroixthomas: That's true indeed, was doing that by consistency but for a single file in the package, probably not worth, I'm gonna put that back to server.go, shall see how this will be handled within the Agent.md

lacroixthomas:
> **Build Failed 😭**
> 
> _Build Id: c2462ccf-16b3-4267-a035-d693d1844e34_
> 
> Status: FAILURE
> 
> * [Cloud Build view](https://console.cloud.google.com/cloud-build/builds/c2462ccf-16b3-4267-a035-d693d1844e34?project=258182270954)
> * [Cloud Build log download](https://storage.googleapis.com/agones-build-logs/log-c2462ccf-16b3-4267-a035-d693d1844e34.txt)
> 
> To get permission to view the Cloud Build view, join the [agones-discuss](https://groups.google.com/forum/#!forum/agones-discuss) Google Group.

http://cmake.org/ still down - Shall see we we want to skip it or if they fix it soon

markmandel:
if this fails again, let's just ignore it.

PR #4720, opened 2026-09-06
feat: use new errors pkg in pkg/apis/*

lacroixthomas:
**What type of PR is this?**
/kind cleanup

Use the new error package (https://github.com/agones-dev/agones/blob/main/pkg/util/errors/errors.go) on the pkg/apis/* files 

Work on #4510 

Y

### DIFF
--- a/pkg/apis/agones/v1/gameserver.go
+++ b/pkg/apis/agones/v1/gameserver.go
@@ -760,7 +760,7 @@ func (gs *GameServer) ApplyToPodContainer(pod *corev1.Pod, containerName string,
 			return nil
 		}
 	}
-	return errs.Errorf("failed to find container named %s in pod spec", containerName)
+	return gameserverErrors.Errorf("failed to find container named %s in pod spec", containerName)
 }
 
 // Pod creates a new Pod from the PodTemplateSpec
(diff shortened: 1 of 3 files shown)


pkg/apis/agones/v1/gameserver.go
```diff
@@ -761,7 +760,7 @@ func (gs *GameServer) ApplyToPodContainer(pod *corev1.Pod, containerName string,
 			return nil
 		}
 	}
-	return errors.Errorf("failed to find container named %s in pod spec", containerName)
+	return errs.Errorf("failed to find container named %s in pod spec", containerName)
```
lacroixthomas: It's "not" * from within a struct and it's using a package level error, but it's because it's the pkg/apis/

pkg/apis/agones/v1/v1.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import "agones.dev/agones/pkg/util/errors"
+
+var errs = errors.FromPackage()
```
lacroixthomas: I'm not sure if we want to keep that consistent and setting it on the v1.go file, or maybe that for this one we don't really care - only used by the gameserver.go 🤷🏼 I'm easy on this one, as you feel

pkg/apis/agones/v1/gameserver.go
markmandel: 🤔 wait, why not put the errors library in the GameServer struct?

We don't want to recreate it on each invocation? (we could create a global one to reuse across gameservers that gets created on package creation on an empty GameServer struct?)

pkg/apis/agones/v1/v1.go
markmandel: I'm fine with it -- consistent with what we'e done elsewhere.

pkg/apis/agones/v1/gameserver.go
lacroixthomas: Youps my comment is wrong, it's the way around, we dont use the error at struct level but at global / package level *

lacroixthomas: And I didn't put that at struct level as there is no constructor, users will use it ditectly as the exported struct, which is why I set it as a global *

lacroixthomas: Oh hmm, I think I just understood, let me check

lacroixthomas: As in using a global one as: `var gameserverErrors= errors.FromStruct(&GameServer{})` or `var gameserverErrors= errors.FromStruct[GameServer](nil)`

lacroixthomas: Updated: af24164

pkg/apis/agones/v1/v1.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import "agones.dev/agones/pkg/util/errors"
+
+var gameserverErrors = errors.FromStruct[GameServer](nil)
```
markmandel, line 19: Nice use of a generic!

markmandel (approved):

PR #4729, opened 2026-09-08
fix: plumb gameservers.lists.maxItems through to the SDK server

### LINKED ISSUE #4680: UpdateList SDK REST endpoint fails with "no gameservers for fleet" even though the GameServer exists
## What happened

Calling the Beta `UpdateList` endpoint on the SDK sidecar (e.g. `PATCH /v1beta1/lists/{name}?updateMask=capacity`) reliably fails with:

```
HTTP 500
{"code":2, "message":"no gameservers for fleet", "details":[]}
```

...even though the calling GameServer exists, is `Ready`, and carries the correct `agones.dev/fleet` label. `GET`/`AddListValue`/`RemoveListValue` on the same List, and `UpdateCounter` on a Counter, all work fine against the same sidecar/GameServer - only `UpdateList` fails.

## Agones version

1.58.0 (Helm chart `agones-1.58.0`), Kubernetes 1.29+, sidecar image `us-docker.pkg.dev/agones-images/release/agones-sdk:1.58.0`.

## Steps to reproduce

1. Deploy a Fleet with a declared list, e.g.:
   ```yaml
   spec:
     template:
       spec:
         lists:
           players:
             capacity: 20
             values: []
   ```
2. From inside the GameServer pod, call:
   ```
   curl -X PATCH "http://localhost:9358/v1beta1/lists/players?updateMask=capacity" \
     -H "Content-Type: application/json" -d '{"capacity":"20"}'
   ```
3. Observe `500 {"code":2,"message":"no gameservers for fleet"}`.
4. `GET http://localhost:9358/v1beta1/lists/players` on the same GameServer succeeds and returns the correct list.

## Root cause (found by reading `pkg/sdkserver/sdkserver.go`)

`UpdateList` calls `GsListsMaxItems` (introduced in #4326 / closes #4282) to look up the Fleet's declared list capacity:

```go
gs, _ := s.gameServerGetter.GameServers(s.namespace).Get(ctx, s.gameServerName, metav1.GetOptions{})
labelsName = gs.ObjectMeta.Labels[agonesv1.FleetNameLabel]
```

The error from `Get(...)` is discarded (`_`). If that lookup doesn't return the expected object (e.g. informer cache not populated/synced for this lookup path), `gs.ObjectMeta.Labels[...]` silently yields `""`, so the subsequent

```go
selector := labels.Set{agonesv1.FleetNameLabel: labelsName}
gsList, err := s.gameServerGetter.GameServers(s.namespace).List(ctx, metav1.ListOptions{LabelSelector: selector.String()})
```

searches for `agones.dev/fleet=` (empty value) instead of the real fleet name, matches zero GameServers, and hits:

```go
if len(gsList.Items) == 0 {
    return fmt.Errorf("no gameservers for fleet")
}
```

This reproduces consistently, not just immediately after GameServer startup (tested several minutes in), so it doesn't look like a one-off cache-sync race.

## Expected behavior

`UpdateList` should succeed the same way `GET`/`AddListValue`/`RemoveListValue`/`UpdateCounter` already do, using the list capacity already available on the GameServer object the sidecar already watches/holds - `GsListsMaxItems`'s extra `List()`-by-fleet-label lookup (and the silently swallowed `Get()` error) looks unnecessary and is the likely source of the bug.

## Related

- #4282 (the issue #4326 fixed) was about a different symptom (hardcoded 1000-item limit) - this is a regression introduced by the same fix (#4326), not previously reported.

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

The `gameservers.lists.maxItems` Helm value was only ever consumed by the CRD templates, so no Go binary received it. Lacking a configured limit, `UpdateList` called `SDKServer.GsListsMaxItems`, which tried to discover one at runtime by listing the GameServers in the fleet. That implementation discarded the error from its `Get` of the current GameServer, so a failed lookup left the fleet label empty, selected `agones.dev/fleet=`, matched nothing and returned "no gameservers for fleet" - failing every `UpdateList` while `GET`, `AddListValue` and `RemoveListValue` on the same List kept working. It also hardcoded the list name "players" and the fleet label "fleet-example" for the `default` namespace, and did an N+1 List+Get from every sidecar.

Separately, `agonesv1.ListMaxCapacity` hardcoded 1000 as the cap enforced during allocation, a second unsynchronised copy of a number `values.yaml` already owns: raise `maxItems` above 1000 and the CRD accepted a capacity the allocator rejected; lower it below 1000 and the allocator accepted one the API server then refused to persist.

The configured value is now carried to every binary that enforces it, as a `MAX_LIST_ITEMS` env var with the same name at each hop, and injected through constructors rather than package globals:

    values.yaml gameservers.lists.maxItems
      |-> controller  -> sidecar env MAX_LIST_ITEMS -> sdk-server -> NewSDKServer
      |-> extensions  -> NewExtensions -> NewAllocator
      |-> allocator   -> newServiceHandler -> NewAllocator
      `-> processor   -> NewAllocator

Closes #4680

Y

Operations implemented:

- Delete `GsListsMaxItems`, `GsLocalListsMaxItems`, the `GameServerListMaxCapacity` package global and the `agonesv1.ListMaxCapacity` constant.
- `NewSDKServer`, `NewLocalSDKServer`, `NewController`, `NewAllocator` and `NewExtensions` take the limit as a constructor parameter; `UpdateListCapacity` and `ListActions` take it as an argument. The out-of-range errors now report the configured limit rather than a hardcoded 1000.
- Add `minimum: 1` to `gameservers.lists.maxItems` in `values.schema.json`, so Helm rejects a non-positive limit at install time.

Also fixes a pre-existing mismatch found en route: the controller sets the sidecar env var `REQUESTS_RATE_LIMIT`, but cmd/sdk-server declared its flag as `request-rate-limit`, which viper maps to `REQUEST_RATE_LIMIT`. The names never matched, so `agones.sdkServer.requestsRateLimit` never reached the sidecar and it always fell back to the 500ms default. The flag is renamed to `requests-rate-limit`; the env var the controller emits is unchanged.

### DIFF
--- a/cmd/allocator/main.go
+++ b/cmd/allocator/main.go
@@ -95,6 +95,7 @@ const (
 	apiServerBurstQPSFlag            = "api-server-qps-burst"
 	logLevelFlag                     = "log-level"
 	allocationBatchWaitTime          = "allocation-batch-wait-time"
+	maxListItemsFlag                 = "max-list-items"
 	readinessShutdownDuration        = "readiness-shutdown-duration"
 	httpUnallocatedStatusCode        = "http-unallocated-status-code"
 	processorGRPCAddress             = "processor-grpc-address"
@@ -117,6 +118,7 @@ func parseEnvFlags() config {
 	viper.SetDefault(totalRemoteAllocationTimeoutFlag, 30*time.Second)
 	viper.SetDefault(logLevelFlag, "Info")
 	viper.SetDefault(allocationBatchWaitTime, 500*time.Millisecond)
+	viper.SetDefault(maxListItemsFlag, 1000)
 	viper.SetDefault(httpUnallocatedStatusCode, http.StatusTooManyRequests)
 	viper.SetDefault(processorGRPCAddress, "agones-processor.agones-system.svc.cluster.local")
 	viper.SetDefault(processorGRPCPort, 9090)
@@ -136,6 +138,7 @@ func parseEnvFlags() config {
 	pflag.Duration(totalRemoteAllocationTimeoutFlag, viper.GetDuration(totalRemoteAllocationTimeoutFlag), "Flag to set total remote allocation timeout including retries.")
 	pflag.String(logLevelFlag, viper.GetString(logLevelFlag), "Agones Log level")
 	pflag.Duration(allocationBatchWaitTime, viper.GetDuration(allocationBatchWaitTime), "Flag to configure the waiting period between allocations batches")
+	pflag.Int64(maxListItemsFlag, viper.GetInt64(maxListItemsFlag), "Flag to set the maximum Capacity a GameServer List may be set to during allocation. Can also use MAX_LIST_ITEMS env variable")
 	pflag.Duration(readinessShutdownDuration, viper.GetDuration(readinessShutdownDuration), "Time in seconds for SIGTERM/SIGINT handler to sleep for.")
 	pflag.Int32(httpUnallocatedStatusCode, viper.GetInt32(httpUnallocatedStatusCode), "HTTP status code to return when no GameServer is available")
 	pflag.String(processorGRPCAddress, viper.GetString(processorGRPCAddress), "The gRPC address of the Agones Processor service")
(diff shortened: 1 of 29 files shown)


lacroixthomas (approved):
Nice catch for the `request-rate-limit`, but mainly for this maxItems hardcoded 😅 
The PR is quite big, don't know if we want a second review on it maybe ? 
Otherwise LGTM !

PR #4737, opened 2026-09-13
fix: move GameServer to Shutdown when the game container exits cleanly and the Pod stays Running

### LINKED ISSUE #4728: Health Checks not running when having "Completed" containers in Pods
**What happened**:
We had a gameserver container that exited with exit code 0 due to an issue and a second container which was responsible to register in some backend. The issue happened right at startup so it never entered Ready or any other state and was still in "Scheduled".

The gameserver container in the pod entered the state completed and despite having healthchecks configured and nothing handling them the gameserver never entered Unhealthy state and kept running forever.

I was able to reproduce this with simple ubuntu images and shell.

Gameservers move into Unhealthy and get cleaned up.

Create the gameserver object, it will start a pod where one container enters the completed state via exit 0. The other container keeps running and while healthchecks are configured agones does not move the Gameserver Object into Unhealthy state and it keeps running.
```
apiVersion: agones.dev/v1
kind: GameServer
metadata:
  name: bug-reproduction
spec:
  container: default
  eviction:
    safe: Never
  health:
    failureThreshold: 3
    initialDelaySeconds: 10
    periodSeconds: 5
  immutableReplicas: 1
  sdkServer:
    grpcPort: 9357
    httpPort: 9358
    logLevel: Info
  template:
    spec:
      containers:
      - args:
        - exit 0
        command:
        - sh
        - -c
        image: ubuntu:resolute
        imagePullPolicy: IfNotPresent
        name: default
        resources:
          requests:
            cpu: 5m
            memory: 8Mi
        securityContext:
          runAsGroup: 1000
          runAsNonRoot: true
          runAsUser: 1000
      - image: ubuntu:resolute
        imagePullPolicy: IfNotPresent
        name: allocator-sidecar
        args:
        - sleep infinity
        command:
        - sh
        - -c
        resources:
          requests:
            cpu: 5m
            memory: 8Mi
        securityContext:
          runAsGroup: 1000
          runAsNonRoot: true
          runAsUser: 1000
      securityContext:
        fsGroup: 1000
      terminationGracePeriodSeconds: 30
```

**Environment**:
- Agones version:  1.59.0
- Kubernetes version (use `kubectl version`): v1.35.x (across rke2, gke, eks, aks)
- Cloud provider or hardware configuration: Various

GrowlyX:
**What type of PR is this?**

/kind bug

The `SucceededController` only moved a `GameServer` to `Shutdown` once the Pod reached the `Succeeded` phase.

A Pod only reaches `Succeeded` once *every* container in `containers` has terminated, so a long-lived non-sidecar container holds the Pod in `Running` after the game server container exits `0`, and the `GameServer` runs forever.

The controller now also treats the Pod as completed when the game server container has terminated with exit code `0`, on a Pod with more than one entry in `containers` and `RestartPolicy: Never` — the conditions under which something other than the game server can hold the Pod in `Running`, and a terminated container is final. #4663 covered the non-zero exit code path in the `HealthController`; this is the exit code `0` counterpart.

**Docs**

- `health-checking.md`: rule 5 of the `SidecarContainers` list said the `GameServer` moves to `Shutdown` "once the Pod completes", which no longer holds. It now matches rule 4's wording and fires as soon as the game server container exits.
- `health-checking.md`: the "additional workloads" section now says Agones watches the game server container directly, so a long-lived container no longer keeps a finished `GameServer` alive (gated behind `publishVersion="1.61.0"`).
- `gameserver.md` / `health-checking.md`: the existing "some CNIs may not support ports on init containers" warning is replaced with the actual cause. The kubelet only passes `hostPort` mappings from regular `containers` to the CNI when it creates the Pod sandbox ([kuberuntime_sandbox.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go#L121)), so a `hostPort` on a native sidecar never reaches the `portmap` plugin. This matters for #4728 because "make it a native sidecar" is the suggested workaround, and it is not available to a container that needs a `GameServer` port (as @nrwiersma pointed out). This is also why `TestGameServerWithPortsMappedToInitSidecarContainers` is skipped.

Closes #4728

Y

Since the first push:

- Dropped the `SidecarContainers` feature gate check from the new code path, per @nrwiersma. The gate only moves the SDK server into `initContainers` and says nothing about whether the user's own containers are sidecars. It was also redundant — `RestartPolicy: Never` is only defaulted inside that gate today — and it wrongly excluded anyone who sets `restartPolicy: Never` themselves.
- Fixed the new e2e test, which failed the last build. It expected an `ACK` from `CRASH 0`, but `handleCrash` calls `os.Exit` before replying, so `SendGameServerUDP` always timed out. It now ignores the reply, as `TestGameServerUnhealthyAfterReadyCrashWithGenericContainer` already does.

The e2e test still skips without `SidecarContainers`, since `DefaultGameServer` only gets `RestartPolicy: Never` via that gate.


pkg/gameservers/succeeded.go
```diff
@@ -181,6 +185,35 @@ func (c *SucceededController) syncGameServer(ctx context.Context, key string) er
...
+// A Pod only reaches the Succeeded phase once *every* container in `containers` has
+// terminated, so a long-lived non-sidecar container can hold the Pod in the Running phase
+// after the game server has finished. With SidecarContainers enabled the Pod is
+// RestartPolicy: Never, so a cleanly exited game server container is done regardless of
+// what the rest of the Pod is doing. The non-zero exit code case is handled by the
+// HealthController.
+func gameServerContainerCompleted(pod *corev1.Pod) bool {
+	if !runtime.FeatureEnabled(runtime.FeatureSidecarContainers) || pod.Spec.RestartPolicy != corev1.RestartPolicyNever {
+		return false
+	}
```
nrwiersma: You can run into the issue when you have `runtime.FeatureSidecarContainers` enabled. Enabling this just moves the Agones SDK to init containers, not all sidecars. You could instead check if the Pod has multiple containers though.

GrowlyX: Fair point

pkg/gameservers/succeeded_test.go
```diff
@@ -211,6 +212,126 @@ func TestSucceededControllerSyncGameServer(t *testing.T) {
...
+
+	// A Pod with only the game server container reaches Succeeded on its own.
+	single := pod.DeepCopy()
+	single.Spec.Containers = single.Spec.Containers[:1]
+	assert.False(t, gameServerContainerCompleted(single))
+
+	// The Succeeded phase is still a completion on its own.
+	succeeded := running.DeepCopy()
+	succeeded.Status.Phase = corev1.PodSucceeded
+	assert.True(t, podCompleted(succeeded))
```
nrwiersma: Nitpick: You would get a clearer signal on these by using `t.Run` on each with the description on the name.

site/content/en/docs/Guides/health-checking.md
```diff
@@ -70,9 +70,27 @@ semantics tied to the lifetime of the game server container:
...
 if the game server container has crashed or exited unexpectedly. Agones will still move the `GameServer` to
 `Unhealthy` as described in rule 3 above, but the Pod's own phase will no longer reflect the state of your game server.
+{{% /feature %}}
+{{% feature publishVersion="1.61.0" %}}
+If you declare these workloads in `containers` instead, a long-lived one will hold the Pod in the `Running` phase
+after the game server container has exited. Agones does not depend on the Pod phase for this: it watches the game
+server container directly, and will move the `GameServer` to `Unhealthy` on a non-zero exit code (rule 4) or to
+`Shutdown` on a clean exit (rule 5) as soon as the game server container terminates. The Pod's own phase, however,
+will no longer reflect the state of your game server, and the other containers keep running until the `GameServer`
+is deleted.
```
nrwiersma: Nitpick: `deleted` implies the Pod exists longer than the GameServer, which it doesnt.

```suggestion
will no longer reflect the state of your game server, and the other containers keep running until the `GameServer`
is shutdown.
```

site/content/en/docs/Reference/gameserver.md
```diff
@@ -172,9 +172,14 @@ Agones supports the use of [Sidecar Containers](https://kubernetes.io/docs/conce
 This requires the `RestartPolicy` to be set to `Always` and can be referenced in the `ports.*.container` field. Typically they used to run
 services that need to run alongside the GameServer container, such as log shippers, monitoring agents, allocators and other supporting services.
 
-{{< alert title="Warning" color="warning">}}
-Some CNIs (Container Network Interface) may not support the use of init containers with host ports. Please check CNI for compatibility.
-{{< /alert >}}
+{{% alert title="Warning" color="warning" %}}
+Ports on sidecar containers do not work on most clusters. When the kubelet creates the Pod sandbox it only passes the
```
nrwiersma: Nitpick: `most clusters` makes assumptions about how clusters are setup. Perhaps:

```suggestion
Ports on sidecar containers are not mapped by Kubernets. When the kubelet creates the Pod sandbox it only passes the
```

nrwiersma (approved):

pkg/gameservers/succeeded.go
```diff
@@ -181,6 +185,42 @@ func (c *SucceededController) syncGameServer(ctx context.Context, key string) er
...
+
+	// a Pod only reaches Succeeded once *every* container in `containers` has terminated, so only a
+	// Pod with more than one of them can be held in Running by something other than the game server.
+	if len(pod.Spec.Containers) < 2 {
+		return false
+	}
+
+	// only RestartPolicy: Always brings a container that exited 0 back.
+	if pod.Spec.RestartPolicy != corev1.RestartPolicyNever && pod.Spec.RestartPolicy != corev1.RestartPolicyOnFailure {
```
markmandel:
```suggestion
	if pod.Spec.RestartPolicy == corev1.ContainerRestartPolicyAlways {
```

Wouldn't that be easier?

site/content/en/docs/Guides/health-checking.md
```diff
@@ -49,7 +49,7 @@ The following is the process for what happens to a `GameServer` when it is unhea
 2. The SDK server sidecar container is set to `restartPolicy: Always`.
 3. If a main container within the Pod fails, the `GameServer` will move to an `Unhealthy` state.
 4. Assuming the Pod is `restartPolicy: Never`, a terminated game server container can never be restarted, so this happens as soon as the game server container exits with a non-zero exit code.
-5. If the game server container exits cleanly (exit code `0`), it is treated as a normal shutdown rather than a failure, and the `GameServer` moves to `Shutdown` once the Pod completes.
+5. If the game server container exits cleanly (exit code `0`), it is treated as a normal shutdown rather than a failure, and the `GameServer` moves to `Shutdown` as soon as the game server container exits.
```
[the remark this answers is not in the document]
markmandel: Fair comment - also, should be feature shortcoded so it only changes on next release please.

site/content/en/docs/Reference/gameserver.md
```diff
@@ -172,9 +172,14 @@ Agones supports the use of [Sidecar Containers](https://kubernetes.io/docs/conce
 This requires the `RestartPolicy` to be set to `Always` and can be referenced in the `ports.*.container` field. Typically they used to run
 services that need to run alongside the GameServer container, such as log shippers, monitoring agents, allocators and other supporting services.
 
-{{< alert title="Warning" color="warning">}}
-Some CNIs (Container Network Interface) may not support the use of init containers with host ports. Please check CNI for compatibility.
-{{< /alert >}}
+{{% alert title="Warning" color="warning" %}}
+Ports on sidecar containers are not mapped by Kubernetes. When the kubelet creates the Pod sandbox it only passes the
```
markmandel: This feels slightly contradictory - it says the Kubenretes doesn't support sidecar containers for ports, but then says that some CNIs support it. Which is it? 😄

markmandel (commented):
Looks good - just some docs things I wanted to poke at 👍🏻

markmandel (approved):

PR #4738, opened 2026-09-13
test: call t.Helper() in shared test helpers

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

Deliberately not a linter change. thelper flags 112 sites, but 67 of those are the per-case closures in table-driven tests, where t.Helper() makes the output strictly worse: those closures each have exactly one caller -- the dispatch line in the test loop -- so marking them as helpers collapses every case's failure onto that single line instead of the assertion that actually failed. In allocation_cache_test.go all nine cases would report line 192.

thelper has no setting that distinguishes a shared helper from a table closure (its only knobs are test.first/name/begin), so enabling it would mean either accepting those 67 or writing 67 suppressions. Neither is worth it for a rule that is a debugging convenience rather than a correctness gate.

So - we're not enabling thelper linting, but we'll merge the good parts in.

Work on #4421

Y/N

N/A

### DIFF
--- a/cmd/controller/main_test.go
+++ b/cmd/controller/main_test.go
@@ -147,6 +147,7 @@ func TestControllerConfigValidationMaxListItems(t *testing.T) {
 }
 
 func errorsContainString(t *testing.T, errs []error, expected string) {
+	t.Helper()
 	found := false
 	for _, v := range errs {
 		if strings.Contains(v.Error(), expected) {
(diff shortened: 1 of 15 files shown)


nrwiersma (approved):

lacroixthomas (approved):
LGTM

PR #4739, opened 2026-09-13
fix: scope test-go to sdks/go to avoid rust SDK build race

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix
> /kind release

test-go passed agones.dev/agones/sdks/... to go test, which walks every subdirectory under sdks/, including sdks/rust. That races with the concurrently-running cargo build/test for the Rust SDK (same -j test run), which creates and removes transient directories under sdks/rust/target/debug/build/*/rmeta*. When go test's package-pattern expansion caught one of those mid-existence, it flaked with:

  FAIL agones.dev/agones/sdks/rust/target/debug/build/agones-<hash>/rmeta<rand> [setup failed]
  cannot find package

sdks/go is the only actual Go package under sdks/, so scope the pattern there instead of relying on the wildcard.

N/A

Y


igooch (approved):
LGTM

PR #4740, opened 2026-09-14
fix: tolerate an existing allocator client secret when copying

GrowlyX:
## What

Make `CopyDefaultAllocatorClientSecret` tolerate a secret that is already present in the target namespace.

## Why

`TestCounterAutoscalerAllocatedMultipleNamespaces` fails in CI with (Observed in Cloud Build `706c4d35-04e9-4267-85c9-bad7c1f01a7c`): 

```
helper_func.go:89: Could not copy default allocator client default/allocator-client.default secret to namespace <ns>: secrets "allocator-client.default" already exists
```

The helper `t.Fatalf`'d on *any* `Create` error, including `AlreadyExists`. Both callers deliberately copy the secret into the **shared** `framework.Namespace`:

- `test/e2e/allocator_test.go` — `TestAllocatorCrossNamespace`
- `test/e2e/fleetautoscaler_test.go` — `TestCounterAutoscalerAllocatedMultipleNamespaces`

both with the comment `// let's reuse an existing one`. So whichever runs second hits the collision, and so does any `--rerun-fails` retry of either.

## Also

Added the missing `t.Helper()`

## Testing

- `go vet ./test/e2e/...` clean
- `gofmt -l` clean
- e2e test binary compiles


markmandel:
/gcbrun

Flake? TestAllocatorAfterDeleteReplica

lacroixthomas (approved):
LGTM

PR #4743, opened 2026-09-15
fix: stop htmltest checking cmake.org

GrowlyX:
## What

Add `https://cmake.org` to `IgnoreURLs` in `site/htmltest.yaml`.

## Why

`make site-test` currently fails on every PR. The `hugo-test` target retries htmltest 20 times and all 20 attempts fail on a single external link:

```
site/docs/guides/client-sdks/cpp/index.html
  Non-OK status: 403 --- https://cmake.org/cmake/help/v3.13/policy/CMP0048.html
make[1]: *** [includes/website.mk:64: hugo-test] Error 1
make: *** [includes/website.mk:73: site-test] Error 2
```

The link comes from `site/content/en/docs/Guides/Client SDKs/cpp.md`, where `CMP0048` is cited for the `-Wno-dev` CMake option.

## Testing

- `curl -A "<the UA from site/htmltest.yaml>" https://cmake.org/cmake/help/v3.13/policy/CMP0048.html` → `403`


markmandel:
Looks like we're good on cmake now, but thanks!

PR #4745, opened 2026-09-15
chore: add new rule for errors in linter and agent

lacroixthomas:
**What type of PR is this?**
/kind cleanup

Added a new rule on the golangci lint to ensure we don't use the previous error package

Added rule in agent on how to use the new package and handle errors 

Work on #4510 

Y


.golangci.yml
```diff
@@ -56,6 +57,13 @@ linters:
     - unparam
     - wastedassign
   settings:
+    depguard:
```
markmandel, line 60: Does the affect tests as well? Because we likely don't care much about tests. WDYT?

lacroixthomas: It might indeed affect tests, even though it would make things consistent between code and unit tests, but indeed it might be too strict hehe

lacroixthomas: Updated  090c4ea

markmandel (approved):

